PR #1794, opened 2012-05-31
don't use XDG path on OS X

minrk:
This should never have been used on OS X.

A few relevant tests adjusted and added appropriately.

It would rarely come up, but occasional bad OS X citizens would create a ~/.config dir, and new OS X installs of IPython after this point would use this dir.

It's entirely possible that zero users have been affected by this so far.

### DIFF
--- a/IPython/utils/path.py
+++ b/IPython/utils/path.py
@@ -227,13 +227,13 @@ def get_home_dir(require_writable=False):
 def get_xdg_dir():
     """Return the XDG_CONFIG_HOME, if it is defined and exists, else None.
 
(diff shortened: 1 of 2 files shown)


jcrben:
I will use the workaround - however - XDG is not a macOS spec spec, but if a macOS user went through the trouble of setting the environment variable, then they probably want it to be used. Git, for example, uses XDG_CONFIG_HOME on macOS if it is set.

minrk:
@jcrben good point. Respecting the environment variable when set and using XDG defaults aren't the same thing. We could still respect the env when set while having platform-appropriate defaults when not set.

PR #2464, opened 2012-10-03
Locate URLs in text output and convert them to hyperlinks.

embray:
The subject of this PR came up in a discussion on the Astropy mailing list, in which we were discussing including URLs to help pages in some exception messages.  The question came up as to how different terminals handle URLs, and which ones render them as hyperlinks.

The question also came up in the context of IPython Notebook. I knew that Markdown cells already detect and render URLs, but that is not true for the output from normal code cells, so this fixes that.  Is there any good reason _not_ to do this?  I've tried several cases and this seems to work pretty well.

### DIFF
--- a/IPython/frontend/html/notebook/static/js/outputarea.js
+++ b/IPython/frontend/html/notebook/static/js/outputarea.js
@@ -379,8 +379,10 @@ var IPython = (function (IPython) {
     OutputArea.prototype.append_text = function (data, element, extra_class) {
         var toinsert = $("<div/>").addClass("box-flex1 output_subarea output_text");
         // escape ANSI & HTML specials in plaintext:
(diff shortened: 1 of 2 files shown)


anonderpling:
This is actually pretty annoying. I'm not sure if this is a colab "enhancement" or comes directly from colab, but this also links paths...including paths in `` `backticks` ``, paths in `<pre>preformatted blocks</pre>` and other methods. [I can add `<wbr>` in the middle](https://stackoverflow.com/a/64023611) like `/<wbr>home/anonderpling/.gitconfig`, however that breaks once I put the path in backticks, as it no longer inserts a wordbreak. I can enter a zero-width space to do so, but then the path can no longer be copied and used, as it now contains an invisible character.

PR #2846, opened 2013-01-25
notebook: Added new short key for all-cell execution

wking:
Actions exposed in the `Cell` menu should have keyboard shortcuts. Now you can
type:

  Ctrl-m . Enter Ctrl-m Enter

To restart the kernel (Ctrl-m .), accept the restart (Enter), and rerun all
the cells (Ctrl-m Enter) to straighten out your prompt numbers.

### DIFF
--- a/IPython/frontend/html/notebook/static/js/notebook.js
+++ b/IPython/frontend/html/notebook/static/js/notebook.js
@@ -128,6 +128,10 @@ var IPython = (function (IPython) {
             } else if (event.which === key.ENTER && event.ctrlKey) {
                 that.execute_selected_cell({terminal:true});
                 return false;
+            } else if (event.which === key.ENTER && that.control_key_active) {
+                that.execute_cell_range(0, that.ncells());
+                that.control_key_active = false;
+                return false;
             } else if (event.which === 77 && event.ctrlKey && that.control_key_active == false) {
                 that.control_key_active = true;
                 return false;
--- a/IPython/frontend/html/notebook/static/js/quickhelp.js
+++ b/IPython/frontend/html/notebook/static/js/quickhelp.js
@@ -28,6 +28,7 @@ var IPython = (function (IPython) {
         var shortcuts = [
             {key: 'Shift-Enter', help: 'run cell'},
             {key: 'Ctrl-Enter', help: 'run cell in-place'},
+            {key: 'Ctrl-m Enter', help: 'run all cells'},
             {key: 'Alt-Enter', help: 'run cell, insert below'},
             {key: 'Ctrl-m x', help: 'cut cell'},
             {key: 'Ctrl-m c', help: 'copy cell'},


lexelby:
I too wanted to be able to quickly restart the kernel and rerun all of my code blocks.  I've found a solution that I'm pretty happy with, and it requires no modification of ipython's code.

First off, I noted that restarting the kernel takes a long time, and that annoyed me, so I found a different way to do what I wanted.  I just want to be able to modify code buried in modules and then rerun my test code entirely, reimporting everything to get the latest version of all of my modules.  A kernel restart will do that, but it turns out it's not necessary.  Instead, we can blow away all modules which will cause them to be reloaded by the next `import` statement.  I do this by carefully manipulating `sys.modules` and `locals()`.  I put this in `~/.ipython/<profile>/startup/restart.py`:

```
def _iprestart():
    import sys
    for module in sys.modules.keys():
        if module not in _initial_modules:
            del sys.modules[module]

    scope_pointer = _scope_pointer
    scope_copy = _scope_copy
    scope_pointer.clear()
    scope_pointer.update(scope_copy)
    scope_pointer['_scope_pointer'] = scope_pointer
    scope_pointer['_scope_copy'] = scope_copy

import sys
_initial_modules = sys.modules.keys()
_scope_pointer = locals()
_scope_copy = locals().copy()
```

The code here is rather bizarre because I'm modifying a scope that `_iprestart()` is in the middle of using.  It boils down to:

* Save a copy of sys.modules after ipython starts up.
* Also save a copy of the contents of locals().
* When _iprestart is called:
 1. remove all modules from sys.modules that were added after startup
 1. erase the interpreter's locals()
 1. replace the interpreter's locals() with the saved copy of locals()
 1. replace references that we'll need next time _iprestart() is called

Convoluted, but it works.

Next, we need to bind this to a shortcut in IPython Notebook.  Add this to `static/custom/custom.js` in your profile directory:

```
var rerun = function (event) {
        IPython.notebook.kernel.execute('_iprestart()');
        IPython.notebook.execute_cell_range(0, IPython.notebook.ncells());
        return false;
    };

$([IPython.events]).on('app_initialized.NotebookApp', function(){
    IPython.keyboard_manager.command_shortcuts.add_shortcut('Ctrl-r', rerun);
    IPython.keyboard_manager.edit_shortcuts.add_shortcut('Ctrl-r', rerun);
    IPython.toolbar.add_buttons_group([
        {
             'label'   : 'reload and rerun',
             'icon'    : 'icon-refresh',
             'callback': rerun
        }
    ]);

    IPython.keyboard_manager.command_shortcuts.remove_shortcut('Ctrl-k');
});
```

The removal of Ctlr-k is just because I hate that ipython eats chrome's ctrl-k, which is used to focus the location bar in search mode.

Now you'll have ctrl-R which will wipe the interpreter's slate clean and rerun all cells.  The shortcut can be used anywhere (command mode or edit mode).  A toolbar button is also added.

Victory!

Carreau:
That’s nice. 

Did you had a look at deep-reload (in IPython.lib) as well as `%reset` and `%reset_selective` magics ? 
— 
m

lexelby:
I considered dreload, but it's not really the solution I wanted because it requires me to manually reload each module I use every time I want to test my code.  %reset is interesting, but it doesn't include unloading modules, so I'd still have to do that.  Plus it would blow away any attempt I made to save the initial module list (unless I stuffed it in sys or something).

My use case is that I want the convenience of ipython (having every line run in a REPL so that I can experiment) with the ability to design and run a test case on the fly.  A test script isn't really what I want because I want the REPL.  My solution works perfectly for my use case -- it's really the missing piece of Python development that I've been searching for.

PR #3291, opened 2013-05-08
Add the option to skip output cells when saving.

filmor:
To allow notebooks to be put under proper version control it is very sensible to remove the variable but “non-source” parts, the output cells (i.e. massive base64-encoded pngs) and prompt numbers.


elgehelge:
@tkf Did you find a solution to this? I am looking for a way to keep the output, but still show nice diffs.

PR #3560, opened 2013-07-06
Remove max_msg_size

minrk:
The max_msg_size safety limit is just removed.

<del>to 1MB. 64k is probably too conservative (1MB may be as well, I'm not sure of the value of this configurable, really).

Also warn about the dropped messages, so that there is a hint about the cause of strange behavior.
</del>

closes #3124


andycjw:
long multiline input takes forever to run and i'm using 3.1.0-cbccb68.
the string is only 100000 in char length, and it's not working for me

PR #3673, opened 2013-07-17
enable comment/uncomment selection

minrk:
with `(cmd|ctrl)-/`

closes #1540


Dalar:
I support that it should be in drop down menus. Very useful feature (also for us Europeans with odd keyboards!)

aloctavodia:
Hello, there is a workaround for people without a US keyboard? (Latin American Keyboard here). Thanks!

PR #4302, opened 2013-09-29
Add an Audio display class

davidosterberg:
This is a first attempt at addressing issue #4241.
In fact it is my first attempt at contributing back so please bear with me for a bit :)


nils-werner:
I'm happy to see my initial attempt was picked up and brought to a state in which it could get merged 👍

PR #4369, opened 2013-10-09
Allow the output of code cells to be individually selectable when making slides

sixpi:
Pull request for #3841.


yasirroni:
So, where is the extensions? Matplotlib by default print the output even when is not asked. Make us to control where the ouput shown will really help.

PR #4457, opened 2013-10-29
Use ~/.ipython as default config directory

takluyver:
As discussed at the dev meeting, we're returning to .ipython on Linux and similar platforms.

The tests for IPython.utils.path could do with some more work to make them properly independent of one another, I think. How do we feel about adding a dependency on mock for Python 2.7? It's in the stdlib in 3.3.

### DIFF
--- a/IPython/core/usage.py
+++ b/IPython/core/usage.py
@@ -45,10 +45,9 @@ Usage
 
     This file is typically installed in the `IPYTHONDIR` directory, and there
     is a separate configuration directory for each profile. The default profile
-    directory will be located in $IPYTHONDIR/profile_default. For Linux users,
-    IPYTHONDIR defaults to `$HOME/.config/ipython`, and for other Unix systems
-    to `$HOME/.ipython`.  For Windows users, $HOME resolves to C:\\Documents
-    and Settings\\YourUserName in most instances.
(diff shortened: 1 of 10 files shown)


flying-sheep:
hi. i’m a linux user and i expect every configuration (except for historical things like vim and bash) to be in `$XDG_CONFIG_HOME`, not polluting my `$HOME`.

just like every windows user expects things to be in in `%APPDATA% and mac users

look [here](http://www.chromium.org/user-experience/user-data-directory) for how you’re supposed to do things. (approximately: while `XDG_CONFIG_HOME=~/.config/` by default, *the user* can configure it to point elsewhere. but those users of course know what they’re doing)

takluyver:
We've been over this plenty of times. We feel it's more important to be consistent across platforms than to try to comply with XDG specs. It's not that we don't know about the specs or how to comply with them, it's that we've weighed up the options and decided to do it with `~/.ipython`.

If it helps, IPython is now some 14 years old, so you can consider it historical.

ivanov:
Please read [Thomas' explanation](https://github.com/ipython/ipython/pull/4457#issuecomment-27737942) for why we ended up doing this (even though we were using XDG directories, we weren't complying to the standard).

Issue #1977, opened 2012-06-18
Add an 'ignore exceptions' mode to run commands in notebook


fperez:
I've been finding myself with notebooks where I do want to keep an exception at a certain point (typically for demo/illustration purposes), but those prove to be really annoying to re-run whole.  One can use 'run all' up to the exception, but from there onwards the only solution is to run them cell-by-cell.

I think it would make sense to have an 'ignore exceptions' toggle in the menu so users can choose to run things simply ignoring any cell that causes an exception.

We also will eventually want more fine-grained 'run' controls, such as 'run from here' and 'run until here', but that's for another issue.

dlukes:
What's the status on this? :) I know there's a [workaround extension for the classic notebook](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/runtools), and as @takluyver points out [here](https://github.com/ipython/ipython/pull/4993#issuecomment-369552684),

> Since notebook 5.1 (PR jupyter/notebook#2549), you can also tag a cell with raises-exception to indicate that an exception is expected; then execution will continue through that cell.

So effectively, the behavior *can* be achieved with stock Jupyter Notebook, although it takes a while to figure it out. As for the new JupyterLab interface, these features are also being discussed on its own issue tracker (https://github.com/jupyterlab/jupyterlab/issues/4496, https://github.com/jupyterlab/jupyterlab/issues/2412). Does this mean this issue can be closed?

PR #4993, opened 2014-02-02
Add a new mode: Run All (skip exceptions)

astrofrog:
This is a prototype implementation of a mode discussed in #1977 to run all the cells in a notebook and skip over any exceptions.

I am new to Javascript and the IPython Notebook code, so this may not be the correct approach, but I would be interested in initial feedback (or suggestions for other approaches).

cc @fperez @minrk

Remaining to-dos:
- [ ] Documentation
- [ ] Changelog
- [ ] Tests (how are tests implemented for the notebook?)

### DIFF
--- a/IPython/html/static/notebook/js/codecell.js
+++ b/IPython/html/static/notebook/js/codecell.js
@@ -302,7 +302,12 @@ var IPython = (function (IPython) {
      * Execute current code cell to the kernel
      * @method execute
      */
-    CodeCell.prototype.execute = function () {
+    CodeCell.prototype.execute = function (skip_exceptions) {
+
+        if(typeof skip_exceptions === 'undefined') {
+          skip_exceptions = false;
+        }
+
         this.output_area.clear_output();
         
         // Clear widget area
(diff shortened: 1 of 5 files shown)


AlJohri:
For anyone that comes here from google, I found doing this in the notebook works well enough:

```
import traceback

try:
    torch.from_numpy(np_array_new)
except RuntimeError as e:
    traceback.print_exc()
```

Usually the point is while teaching to show that doing x y z will throw an exception. This will probably work for everything other than a SyntaxError.

takluyver:
Since notebook 5.1 (PR https://github.com/jupyter/notebook/pull/2549), you can also tag a cell with `raises-exception` to indicate that an exception is expected; then execution will continue through that cell.

fperez:
On Thu, Mar 1, 2018 at 2:45 AM, Thomas Kluyver <notifications@github.com>
wrote:

> Since notebook 5.1 (PR jupyter/notebook#2549
> <https://github.com/jupyter/notebook/pull/2549>), you can also tag a cell
> with raises-exception to indicate that an exception is expected; then
> execution will continue through that cell
>

I didn't know that. Brilliant, thanks :)

PR #5194, opened 2014-02-22
require latex titles to be ascii

minrk:
otherwise the ucs package dies, telling you to enable a 'combine' option, whose only effect seems to be to remove all unicode characters.

see #5133


michaelpacer:
Now that we use XeLaTeX we should be able to handle unicode in titles. Maths can be a different issue, but that's often because of coverage in the math font itself. 

Were we to try to change this, what would the minimal test case need to cover and would it be better to change it here or in the nbconvert repo directly?

Issue #5987, opened 2014-06-14
Allow to embed local videos using html5


dawehner:
As shown in http://nbviewer.ipython.org/github/ipython/ipython/blob/1.x/examples/notebooks/Part%205%20-%20Rich%20Display%20System.ipynb you can embed videos into ipython not only using Youtube or Vimeo but also local files.

Is there a reason why there is no simple way of an Video display which allows you to pass in the filepath?

mpenkov:
This looks like an easy starting issue, so I'll start working on it and send a PR during the week.

mpenkov:
It looks like #5992 got merged, so perhaps this issue should be marked as completed?

Carreau:
Sure, thanks. 

You should be able to close it by mentioning our helper bot as follow:

@meeseeksdev[bot] close

PR #5992, opened 2014-06-15
Add support to embed videos.

dawehner:
http://nbviewer.ipython.org/gist/dawehner/36e7b62883358f57db3e
is the test notebook.

Note: There are a couple of TODOs in the patch of functions I don't really
know what to do about.

This solves https://github.com/ipython/ipython/issues/5987

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -21,6 +21,7 @@ from __future__ import print_function
 
 import os
 import struct
+import mimetypes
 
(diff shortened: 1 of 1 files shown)


JhonnyBillM:
Hi, since this is kind of old, I would to know what can I do to help here?

takluyver:
I think this is all done. Try `from IPython.display import Video`.

PR #6221, opened 2014-07-27
Switch to CodeMirror 4.6.0

Carreau:
Update to CodeMirror 4

That mean multiple cursor ! Yeaaah. 

<del>On the dark side, we have to painfully recode ipythongfm and ipython mode (I'll appreciate help) .</del> (Done)

A few things have changed, in particular `event.stop()` does not exist, and I had to comment the `goLineUp` and `goLineDown` in kerboardmanager as returning `true`/`false`seem not to be the way to go anymore to prevent CM from handeling the events ( @ellisonbg your input on that )

<del>Completer does not intercept up/down anymore...</del>

Todo: 
- [x] update to cm 4.6
- [x] fix completer
- [x] search for "somethingSelected" add "or MultipleCursors" (like completion)
- [x] fix shift tab 4 time select previsous cell
- [x] fix tests.
- [x] remove deprecated logic since CM 2.x
- [x] ~~completer : insert at all cursors.~~
- [x] autohighlight (Carreau/ipython#10)

But globally it seem to work.

### DIFF
--- a/IPython/html/static/base/js/dialog.js
+++ b/IPython/html/static/base/js/dialog.js
@@ -4,7 +4,8 @@
 define([
     'base/js/namespace',
     'jquery',
-], function(IPython, $) {
+    'codemirror/lib/codemirror',
+], function(IPython, $, CodeMirror) {
     "use strict";
     
     var modal = function (options) {
--- a/IPython/html/static/base/js/keyboard.js
+++ b/IPython/html/static/base/js/keyboard.js
@@ -1,5 +1,12 @@
 // Copyright (c) IPython Development Team.
 // Distributed under the terms of the Modified BSD License.
+/**
+ *
+ *
+ * @module keyboard
+ * @namespace keyboard
+ * @class ShortcutManager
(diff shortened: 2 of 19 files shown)


spencerlyon2:
@jdfreder, I tried adding this inside `$([IPython.events]).on('app_initialized.NotebookApp', function(){` in my custom.js, but had no luck.

Am I supposed to put is somewhere different?

Carreau:
> `$([IPython.events])`

Will not be define you need to 
```
require([
        'base/js/namespace',
        'base/js/events'
    ], function(IPython, events) {

.....

});
````

minrk:
Yes, though it's a bug that `$([IPython.events])` isn't defined.

jdfreder:
That snippet sits in my custom.js just as I pasted it, no reference to the events object.  Require.js is enough.

spencerlyon2:
Excellent. I have put it at the top level of my custom.js and it works.

Thanks.

PR #6290, opened 2014-08-11
Deprecate rmagic

takluyver:
The R magic is in released versions of rpy2 now, and we're starting to see test failures because of changes in rpy2. This loads rmagic from rpy2 if possible, and removes our copy of the tests and docs.

Ping @davclark


nihakue:
So now that rmagic is deprecated, what is the kosher way to access that functionality? I really liked being able to %load_ext rmagic and then go on my merry way.

davclark:
Since it doesn't appear clear from the above, you now do this with `%load_ext rpy2.ipython`. Note that the magic used to depend on rpy2, so you needed to install that anyway. Also - the conversion paths have been in flux in rpy2, so the auto-conversion is (for now) disabled in the magic. I'm planning to get back to this in a month or two.

PR #6351, opened 2014-08-23
Multiple Checkpoint Support and Beginnings of Google Drive Support

jhemmelg:
These changes implement multiple checkpoint support for notebooks.  There is a new menu item "Delete a Checkpoint" which allows for checkpoints to be deleted.  Checkpoints are now serialized with a simple number.  New checkpoints get a serial number one higher than the highest existing serial number.

Google Drive support can be experimented with by substituting drive_contentmanager.js for contentmanager.js.


KesterTong:
@jhemmelg, In case you were interested in taking a look, we have finished the first version of Drive integration at https://github.com/jupyter/jupyter-drive.  Thanks again for your contributions in the early stages of this project.

PR #6708, opened 2014-10-15
include mathjax in components

minrk:
increases IPython install size by 2 MB (11->13MB), and wheel size by 1.3MB (2.9->4.2MB)


minrk:
Closing, because we won't do this until after the repo split.

PR #6751, opened 2014-10-21
Added --url option to iptest

jdfreder:
Allows `iptest js` to be used against remote servers.

Example usage: `iptest js --url='http://localhost:8888'`

Pinging @rgbkrk who requested this.


jasongrout:
As a note, if you are testing with phantomjs, you'll need to also export a variable to tell IPython to use old websockets: `export IPYTHON_ALLOW_DRAFT_WEBSOCKETS_FOR_PHANTOMJS=1`

jdfreder:
Thanks @jasongrout

PR #6879, opened 2014-11-07
Add matplotlib cell magic

francisco-dlp:
In my workflow I often activate the matplotlib inline backend to execute one cell and I go back to the previous backend afterwards. I use the maplotlib line magic for that but I find that this could be done more conveniently if there was a matplotlib cell magic. This pull request modifies the matplotlib line magic to make it work also as a cell magic.

### DIFF
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -16,7 +16,7 @@ from __future__ import print_function
 # Our own packages
 from IPython.config.application import Application
 from IPython.core import magic_arguments
-from IPython.core.magic import Magics, magics_class, line_magic
+from IPython.core.magic import Magics, magics_class, line_magic, line_cell_magic
 from IPython.testing.skipdoctest import skip_doctest
 from IPython.utils.warn import warn
 from IPython.core.pylabtools import backends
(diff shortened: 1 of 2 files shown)


takluyver:
From the dev meeting last week: we'd prefer this to be a context manager, perhaps even in matplotlib (see e.g. [switch_backend()](http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.switch_backend)), but enabling and disabling the inline backend may require hooks after the cell's execution, in which case it couldn't be implemented as a context manager, and we'd consider this approach.

However, at this point, we'd like to leave it until after IPython 3.0, which we are trying to get ready for release. Setting the milestone to 4.0.

ellisonbg:
This one has sat idle for a while with no activity. Any plans to work on this soon? If not, do you mind if we close this until work starts again - it may be the the work needs to be done in matplotlib as well. We are splitting our repo and all existing PRs are going to bit-rot quickly...

Carreau:
I'm closing for inactivity, but did open an Issue reminding us that this was here.

francisco-dlp:
Sorry to be slow to reply. This feature is important for us and we will be
happy to resume work on this in the close future.

2015-04-03 0:29 GMT+02:00 Matthias Bussonnier <notifications@github.com>:

Issue #5327, opened 2014-03-11
Allow to select multiple in select widgets


dawehner:
It would be kinda helpful for some usecases to be able to configure a SelectWidget to have multiple selections instead of just one

Maybe something like this could work
```
def helper(a):
    print a

interact(helper, a=SelectWidget(values = ['a', 'b', 'c'], multiple=True))
```

Thank you in advance
PS: is there any reason to append tags to new issues?

bollwyvl:
Fixed by #6890!

Here's the [new API](http://nbviewer.ipython.org/github/ipython/ipython/blob/8cd71d912654c02d2d3678bb3501caad4b924de8/examples/Interactive%20Widgets/Widget%20List.ipynb#Selection-widgets)!

bobbysan:
Selecting multiple items using the SelectMultiple widget does not work in some cases for me. If the selected options are highlighted using only the mouse (and are contiguous) it works fine. If selecting only apples and bananas (using Shift and Ctrl) in the example below it does not work. It will highlight/select the last item clicked. Please advise. Thanks.

```python
w = widgets.SelectMultiple(
    description="Fruits",
    options=['Apples', 'Oranges', 'Pears', 'Grapes', 'Peaches', 'Bananas'])
display(w)
```

My setup:
ipython-verson: 3.1.0-dev & 3.0.0 (tried both)
platform: windows 7

bollwyvl:
Confirmed.

It is actually, really briefly, correct, but then the super `change`
handler fires and clobbers the new, correct value.

jdfreder:
I'll copy this over to an issue so we can track it.

PR #6890, opened 2014-11-09
SelectMultiple widget

bollwyvl:
Adds a minimal multiple select widget, with `_MultipleSelection` as baseline for additional multiple select widgets: i.e. `ToggleButtonsMultiple`.

Unlike the suggestion of #5327 to have `multiple=True`, this adopts a whole separate widget due to  downstream incompatibilities: some of the `_Selection` subclasses cannot accept multiple values (Radios). Perhaps adding a `multiple` traitlet while allowing implementations to opt-out would make more sense.

### DIFF
--- a/IPython/html/static/widgets/js/widget_selection.js
+++ b/IPython/html/static/widgets/js/widget_selection.js
@@ -5,8 +5,9 @@ define([
     "widgets/js/widget",
     "base/js/utils",
     "jquery",
+    "underscore",
     "bootstrap",
-], function(widget, utils, $){
+], function(widget, utils, $, _){
 
     var DropdownView = widget.DOMWidgetView.extend({
         render : function(){
@@ -52,19 +53,19 @@ define([
             /**
              * Update the contents of this view
              *
-             * Called when the model is changed.  The model may have been 
+             * Called when the model is changed.  The model may have been
              * changed by another view or by a state update from the back-end.
              */
 
             if (options === undefined || options.updated_view != this) {
-                var selected_item_text = this.model.get('value_name');
+                var selected_item_text = this.model.get('selected_label');
                 if (selected_item_text.trim().length === 0) {
                     this.$droplabel.html("&nbsp;");
                 } else {
                     this.$droplabel.text(selected_item_text);
                 }
                 
-                var items = this.model.get('_value_names');
+                var items = this.model.get('_options_labels');
                 var $replace_droplist = $('<ul />')
                     .addClass('dropdown-menu');
                 // Copy the style
@@ -150,7 +151,7 @@ define([
              * Calling model.set will trigger all of the other views of the 
              * model to update.
              */
-            this.model.set('value_name', $(e.target).text(), {updated_view: this});
+            this.model.set('selected_label', $(e.target).text(), {updated_view: this});
             this.touch();
 
             // Manually hide the droplist.
@@ -188,7 +189,7 @@ define([
              */
             if (options === undefined || options.updated_view != this) {
                 // Add missing items to the DOM.
-                var items = this.model.get('_value_names');
+                var items = this.model.get('_options_labels');
                 var disabled = this.model.get('disabled');
                 var that = this;
                 _.each(items, function(item, index) {
@@ -209,7 +210,7 @@ define([
                     }
                     
                     var $item_element = that.$container.find(item_query);
-                    if (that.model.get('value_name') == item) {
+                    if (that.model.get('selected_label') == item) {
                         $item_element.prop('checked', true);
                     } else {
                         $item_element.prop('checked', false);
@@ -263,7 +264,7 @@ define([
              * Calling model.set will trigger all of the other views of the 
              * model to update.
              */
-            this.model.set('value_name', $(e.target).val(), {updated_view: this});
+            this.model.set('selected_label', $(e.target).val(), {updated_view: this});
             this.touch();
         },
     });
@@ -305,7 +306,7 @@ define([
              */
             if (options === undefined || options.updated_view != this) {
                 // Add missing items to the DOM.
-                var items = this.model.get('_value_names');
+                var items = this.model.get('_options_labels');
                 var disabled = this.model.get('disabled');
                 var that = this;
                 var item_html;
@@ -328,7 +329,7 @@ define([
                             .on('click', $.proxy(that.handle_click, that));
                         that.update_style_traits($item_element);
                     }
-                    if (that.model.get('value_name') == item) {
+                    if (that.model.get('selected_label') == item) {
                         $item_element.addClass('active');
                     } else {
                         $item_element.removeClass('active');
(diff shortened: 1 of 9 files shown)


takluyver:
For the naming, what about something along the lines of `options`, `selected_option`, `selected_options`?

jdfreder:
+1 to @takluyver 's suggestion

ellisonbg:
+1 - I like that too

On Wed, Jan 21, 2015 at 1:37 PM, Jonathan Frederic <notifications@github.com
> wrote:

> +1 to @takluyver <https://github.com/takluyver> 's suggestion
>
> —
> <https://github.com/ipython/ipython/pull/6890#issuecomment-70926666>.
>



-- 
Brian E. Granger

bollwyvl:
Sorry, a bit out of the loop the last few days :)
Just to clarify, are we talking changing the name of `value` to `selected_option` in `_Selection` and all subclasses, or just for the new `_MultipleSelection`?

jdfreder:
I prefer the consistency across all of _Selction widgets.

takluyver:
Does that conflict with things that expect widgets to have a `value` trait representing the user's input?

I don't have strong feelings either way, just thinking through the implications.

jdfreder:
Oh, hmm.  That'll be a problem with `interact`.  Perhaps each could implement an alias property named `value` that returns `selected_option` for the single selection variations  of the widget and `selected_options` for the multi-selection ones.

takluyver:
Thinking about this some more, maybe we will preserve `value` and rename the other things. Currently, the selection widgets have public traits `values` (the available choices), `value_name` (the text of the selected item) and `value` (the Python object associated with that), and private traits `_values_dict`, `_value_names`, and, magnificently, `_value_values`.

I propose leaving `value` unchanged for consistency with other widgets, calling the available choices 'options' rather than 'values', and the text displayed for them 'labels' instead of 'names'. In full, the six traitlets would become:

* `value`
* `selected_label` (`selected_labels` for the multiple-selection case)
* `options`
* `_options_dict`
* `_options_labels`
* `_options_values` (I'd like to get rid of 'values' here, but I can't think of a better word right now)

I'd also like to have a go at simplifying the logic for this, but that's a separate question.

bollwyvl:
...and SelectMultiple would still only expose `value`, it would just always be a list?

also, should the renaming to `options` hit as a separate PR before this?

jdfreder:
@bollwyvl I think it's okay for it to be in this PR, since this too is targeted for 3.0.

takluyver:
Correct regarding `value` - all of the stateful widgets have a `value` trait reflecting the current state, which is used by `interact()`, so we should preserve that for SelectMultiple.

I think it's OK to do the renaming in this PR, but let @jdfreder weigh in on my proposal before you start changing things.

jdfreder:
@bollwyvl I like @takluyver 's proposal.  Go ahead and implement that in this PR and once the ducks are in order we can merge. :)

bollwyvl:
Changed `values` to `options` throughout, and the family of `_dict`, `_labels`, `_values` that go with it: this included tests, example notebooks, etc. Hopefully that's as far as the creep goes :octopus:

bollwyvl:
Also, i left `selected_label` for SelectMultiple... since `value` stayed the same, it seemed more natural to leave it be instead of pluralizing.

takluyver:
I think I'd be inclined to pluralise even so, and just have value as a
special case. Jon, thoughts?

jdfreder:
I agree with @takluyver - I think value should be the only exception, and other variables should be named logically (selected_labels here).

bollwyvl:
okiedoke. I thought it would cause more problems that it did on the front end, but it appears to work just fine with only minor modification. rebased and changed.

also updated the example notebook.

jdfreder:
Thanks @bollwyvl , I'm taking a look at this now.

bollwyvl:
Hm... as <kbd>shift</kbd>/<kbd>ctrl</kbd> is part of the gimmick of mutliple select (potentially with the keyboard), the `handle_click` is a little insufficient... is there a reason to not use `on("input", ...)` on the listbox itself? I've added it to SelectMultipleView, but it might be appropriate on SelectView as well...

IPython/html/static/widgets/js/widget_selection.js
```diff
@@ -43,28 +44,28 @@ define([
                 this.update_button_style();
             }, this);
             this.update_button_style('');
-            
+
```
takluyver: When working on IPython, can you configure your editor not to strip trailing whitespace? It causes unnecessary git conflicts, and makes the diffs harder to review. Thanks.

IPython/html/static/widgets/js/widget_selection.js
```diff
@@ -506,18 +507,48 @@ define([
...
-    
+
+
+    var SelectMultipleView = SelectView.extend({
+        render: function(){
+            SelectMultipleView.__super__.render.apply(this);
+            this.$el.removeClass('widget-select')
+              .addClass('widget-select-multiple');
+            this.$listbox.attr('multiple', true)
+              .on('input', $.proxy(this.handle_click, this));
```
takluyver: I can't find docs for an `input` event on select widgets, but `change` sounds like what we want: http://api.jquery.com/change/

takluyver:
It looks to me like `on('change', ...` instead of `on('input', ...` is what we want to use.

@jdfreder, up to you whether we accept this with the whitespace changes or ask @bollwyvl to refactor them out, since it's most likely to cause conflicts with stuff you know about.

bollwyvl:
gah, sorry about the whitespace, bane of all developers. switched to atom
when i switched machines, and it strips by default. would a squash merge,
then interactive merge without whitespace work? is there a magic rebase
configuration to not add whitespace changes?

as to events, looks like `change` and `input` both fire: (events logged in
console)

http://codepen.io/anon/pen/PwKBjz

i'm used to `change` being pretty late-binding on text inputs (basically
`blur`) which seems un-widgetlike, but for selects, each atomic click
appears to be counted, so it should indeed be fine.

takluyver:
No worries, these things happen. git rebase does have a `--ignore-whitespace` flag, maybe that will do the trick.

I only see 'change' firing in your example, with Firefox on Linux, not 'input'. It seems to be firing at all the right times, both with mouse and keyboard input, though.

bollwyvl:
That was really unpleasant :facepunch: reminds me of the bad old days of big CVS merges! Does anyone in the project have

Changed the event to `change` for both, and yes: seeing the same thing on FF, though chromium is giving me `input`.

I've taken screenshots, and still can't figure why the tests are failing, as the one right after the rebase worked... will keep investigating...

bollwyvl:
The tests were using `$el.click`, which _doesn't_ actually fire the `change` event. Learn something every day. We could use `casper.sendEvent` to be a little more realistic, but adding this case seemed reasonable for now.

Now handling that little gem, I think this is basically ready to go... unless I should do yet more event investigating. Is keyboard support a thing that needs testing?

jdfreder:
> Is keyboard support a thing that needs testing?

It's not tested now so I wouldn't worry about it.  In the future we may be changing test frameworks anyways.  I restarted Travis, because it looked like the last failure was sporadic.

jdfreder:
Thanks @bollwyvl !  I'm sorry this took so long to merge.  Tests are passing, I looked over it one more time, and it looks good to me.

PR #6936, opened 2014-11-14
Don't specify profile-dir for kernels

minrk:
we are passing `--profile-dir` to non-IPython kernels. This avoids specifying it to _any_ kernels, which includes IPython, which may be premature.

It does clean up something in the console frontends, which blended frontend and kernel args. Now the console frontends  no longer need to understand anything about kernel flags, and kernel args can only be passed after `--`, e.g.:

```
   ipython qtconsole -- --matplotlib inline

   ipython qtconsole --kernel julia -- --some-julia-kernel-flag
```

closes #6626

### DIFF
--- a/IPython/consoleapp.py
+++ b/IPython/consoleapp.py
@@ -18,20 +18,13 @@ from IPython.config.application import boolean_flag
 from IPython.core.profiledir import ProfileDir
 from IPython.kernel.blocking import BlockingKernelClient
 from IPython.kernel import KernelManager
-from IPython.kernel import tunnel_to_kernel, find_connection_file, swallow_argv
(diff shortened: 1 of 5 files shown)


minrk:
rebased

Carreau:
+1

Carreau:
Rhaa.. should have merge before split. 

@minrk quick rebase and merge ?

minrk:
Rebase applied cleanly. No conflicts.

Carreau:
Ok, github was giving grey button unclickable... I suppose rebase can be smarter than merge.

minrk:
Yup, I've had it happen a few times when it's just files moving around.

Carreau:
Travis is greenish. 
I'll read once more and push.

PR #7008, opened 2014-11-21
Theme generator.

Carreau:
Work in progress, will probably end up in tools or something like that.
Problem is that less.js APi is not documented so I have to digg in the source.
It also requires less 2.0.

Allow you to easy generate a theme for the notebook. It should even be able to watch for file change and recompile on the fly.

With 

```
i= setInterval( 
  function(){
    $('head>link:nth(4)').attr('href',"/static/style/style.min.css?v="+Date()); console.log(Date())}
  , 4000)
```

it allows the css to be live reloaded in the notebook, without refreshing the page, which is nice to play with css..

### DIFF
--- a/IPython/html/static/style/custom.less
+++ b/IPython/html/static/style/custom.less
@@ -1,0 +1,64 @@
+// @border_radius : 0;
+// @text-color: #0d0104;
+// 
+// @border-radius-base:    0px;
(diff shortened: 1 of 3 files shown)


Carreau:
Closing as #7931 reference that. 

Anybody feel free to take over.

PR #7074, opened 2014-12-01
make deprecation more explicit

Carreau:
Just playing around a with decorator to mark deprecated feature and have an option that woudl actually make them Warn or raise if used : 

```
In [3]: ip = get_ipython();ip.config._merge({'z':3})

In [4]: from IPython.utils import deprecate; deprecate.nocompat=True

In [5]: ip = get_ipython();ip.config._merge({'z':3})
DeprecationWarning                        Traceback (most recent call last)
<ipython-input-5-4b0aa8f46ac3> in <module>()
----> 1 ip = get_ipython();ip.config._merge({'z':3})

/usr/local/lib/python3.4/site-packages/IPython/utils/deprecate.py in _raise(*args, **kwargs)
    130             def _raise(*args,**kwargs):
    131                 if _comp(version) or nocompat:
--> 132                     raise DeprecationWarning('This is a deprecated feature, it will eb removed after %s ' % str(version))
    133                 else:
    134                     return func(*args,**kwargs)

DeprecationWarning: This is a deprecated feature, it will eb removed after (4, 0, 0)
```

Also once we bump IPython version above the limit for the deprecation. The thing will raise even earlier to remind us that there is something to clean up. 

From a discussion with @ogrisel yesterday in the train. And about the fact that numpy turn their deprecation warning as real error during testing to ensure they don't usetheir own deprecated apis.

### DIFF
--- a/IPython/config/loader.py
+++ b/IPython/config/loader.py
@@ -17,6 +17,7 @@ from IPython.utils import py3compat
 from IPython.utils.encoding import DEFAULT_ENCODING
 from IPython.utils.py3compat import unicode_type, iteritems
 from IPython.utils.traitlets import HasTraits, List, Any
+from IPython.utils import deprecate
 
 #-----------------------------------------------------------------------------
 # Exceptions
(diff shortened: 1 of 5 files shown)


takluyver:
This has been inactive for over a month, and it's not in a form we can merge, so closing for now.

PR #7087, opened 2014-12-03
Enable the IPython Notebook to display messages from other clients.

nheijermans:
This feature was discussed in #6123, but it doesn't look like anything was ever incorporated into the IPython Notebook.

Here's a brief overview of the changes:
- Display of messages from other clients can be toggled on and off from within a notebook, either using the `<M-m>e` keyboard shortcut in the web UI, or through the option in the `Kernel` drop-down menu.
- `notebook.js` controls whether messages are displayed through a callback that is invoked from `kernel.js` when no callbacks are available for a message.
- The UI displays `execute_input` messages originating from an other clients in new cells at the end of the notebook. Output messages (`execute_result` et al.) will only be displayed if a cell exists with a matching message ID.

Pending design questions:
- Should each `execute_input` message cause a new cell to be created?
- Should new cells be placed at the end of the notebook, or elsewhere? If the latter, what criteria should be followed?


pablocobelli:
Enabling jupiter notebooks to display messages from other clients would be a great feature. From reading the comments I don't get whether this was implemented or not. Is it possible to do it? If yes, how would one accomplish that? Thanks in advance.

pablocobelli:
I'll answer my own question since I found the answer (it was just a matter of properly reading the previous exchange, though). This capability was developed by nheijermans as an extension for the Jupyter Notebook; it's available for download at https://github.com/nheijermans/nbexts. Thanks again and sorry for the noise.

PR #7132, opened 2014-12-08
Split js tests into subgroups

Carreau:
js test are the longer ~8 minutes vs ~3 for Python, and fails the more often.
Hopefully this will speedup the build and have less test relaunched when we
have a random faillure.  Splitting in 3 should make the js test a bit faster
and gave us some rooms without increasing the number of groups.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,10 @@ python:
     - 2.7
     - 3.3
 env:
-    - GROUP=js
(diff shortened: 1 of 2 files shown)


damianavila:
>I think we can merge this and see how it goes.

After rebase :+1:

Carreau:
rebased.

PR #7175, opened 2014-12-11
Asynch InProcessKernel

marcus-oscarsson:
The InProcessKernel can be really usefull, however the evaluation of the code is blocking which cuases
the application that embeds the kernel to freeze. This is a simple attempt to create a AsyncInProcessKernel and AsyncInProcessKernelManager that facilitates asynchronous evaluation of the code so that the client can continue to process events.

Thanks for all the great work you guys are putting in to this !

### DIFF
--- a/IPython/kernel/inprocess/asyncipkernel.py
+++ b/IPython/kernel/inprocess/asyncipkernel.py
@@ -1,7 +1,6 @@
 """An asynchronous in-process kernel"""
 import threading
 import time
-import gevent
 
 from .ipkernel import InProcessKernel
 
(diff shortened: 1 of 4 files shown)


damianavila:
>We discussed this at the dev meeting today, and concluded that if you need changes in the existing code to make this possible, we're happy to merge those changes, but we'd prefer you to maintain the new subclasses for the async in-process kernel as a standalone project. Thanks!

@marcus-oscarsson, do you have plans to request/made entry points using this same PR? 
If not, I think we can close this one... there is no real possibility to get this merged, so there is no need to keep it opened.

marcus-oscarsson:
Sorry for taking sometime to answer your comments, I have been away for quite some time. I will take a closer look at #7180 to see what changes that needs to be done. I need the feature for one of my own projects so I will probably continue to maintain this solution, if I don't find anything better/easier. Ill try to implement something and Ill get back to you guys if I have something I think might be worth merging. I will close this pull request and perhaps open a new one later.

Happy new year !

damianavila:
Thanks for the feedback @marcus-oscarsson... happy new year for you too!

PR #7184, opened 2014-12-11
various latextools fixes

minrk:
- support unicode text, not just bytes
- remove unused math_to_image copied from mpl
- load global config, instead of instantiating in shell (fixes config load in qtconsole)
- don't unwrap/rewrap math in qtconsole - this broke display of all non-math latex
- mpl math_to_png doesn't support display math, coerce to inline when found


Carreau:
Had a quick read through. +1, merging.

PR #7278, opened 2014-12-19
don't enforce .ipynb extension in URLs

### LINKED ISSUE #7275: Allow arbitrary notebook filename extensions
(... or allow custom handlers)

I first described my use case in #7222, it has slightly changed since then. [GitHub repo here](https://github.com/rossant/ipymd).

Basically I want to write technical documents/books in Markdown, and use the notebook UI to write/execute code. This brings the benefits of Markdown with the benefits of the notebook UI for interactive computing. When writing a book, I don't want to save/version-control the code outputs. I can always build a full document with the outputs eventually.

Technically, I can achieve this by writing a custom `FileContentsManager`, and implementing my own serialization/deserialization process for notebooks (basically `nbcontents (dict) <==> markdown` converters).

The only thing that doesn't work nicely for now is the file extension. Using `.ipynb` works, but it's probably not a good idea to use the same extension for two different file formats. When using `.md`, it looks like I cannot use the notebook UI because of the way handlers are implemented. AFAIU there's a `.ipynb` regex that redirects those paths to the notebook UI. For now, I could patch IPython myself, and ask users to clone my own fork/branch.

I can see at least two solutions:

* create a parameter with a list of file extensions that redirect to the notebook UI (basically changing the regex)
* create a parameter with a list of extra handlers for the Web server

What do you think?

minrk:
let ContentsManager decide what's a notebook

closes #7275

### DIFF
--- a/IPython/html/base/handlers.py
+++ b/IPython/html/base/handlers.py
@@ -464,7 +464,13 @@ class TrailingSlashHandler(web.RequestHandler):
 
 class FilesRedirectHandler(IPythonHandler):
     """Handler for redirecting relative URLs to the /files/ handler"""
-    def get(self, path=''):
+    
+    @staticmethod
+    def redirect_to_files(self, path):
+        """make redirect logic a reusable static method
+        
+        so it can be called from other handlers.
+        """
         cm = self.contents_manager
         if cm.dir_exists(path):
             # it's a *directory*, redirect to /tree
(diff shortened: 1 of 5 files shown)


rossant:
Indeed, thanks!

Carreau:
Cyrille is happy, looking one last time and merging.

PR #7285, opened 2014-12-22
Menu actions

Carreau:
Draft/Proof of concept. Done in the car while moving from one flat to another (no I was not driving).

So what are you suppose to see here. 
Purely declarative menu form: 

```
"Edit",[
                'ipython.cut-selected-cell',
                'ipython.copy-selected-cell',
                'ipython.paste-cell-before',
                'ipython.paste-cell-after',
                '<divider>',
                'ipython.delete-cell',
                '<divider>',
                ['Submenu',[
                    'Action 1',
                    'Action 2',
                    'Action 3'
                    ]
                ]
```

And automatically show icon/shortcut if associated with action. 

capture d ecran 2014-12-22 a 18 29 49

Still highly work in progress.

### DIFF
--- a/IPython/html/static/notebook/js/menubar.js
+++ b/IPython/html/static/notebook/js/menubar.js
@@ -92,46 +92,41 @@ define([
     //      - submenu : list of 2 element [submenu name, list of action] (ie this)
     //   the pseudo action depends on specific object, or can be <separator> in the menu
     MenuBar.prototype.addMenu = function(name, list, element, submenu){
-//        <li class="dropdown"><a href="#" class="dropdown-toggle" data-toggle="dropdown">File</a>
-        var theli = $('<li/>')
-                .addClass('dropdown')
-            if(submenu){
-                theli.addClass('dropdown-submenu')
-                }
-            theli.append(
-                $("<a/>").addClass('dropdown-toggle').attr('data-toggle',"dropdown").text(name)
-            )
+        var theli = $('<li/>').addClass('dropdown')
+        if(submenu){
+            theli.addClass('dropdown-submenu')
+            }
+        theli.append(
+            $("<a/>").addClass('dropdown-toggle').attr('data-toggle',"dropdown").text(name)
+        )
         var theul = $('<ul/>').addClass('dropdown-menu');
         theli.append(theul);
         for(var index in list){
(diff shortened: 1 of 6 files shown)


minrk:
@Carreau planning to finish this one? I started to work on something, and realized it would be better for this to be merged first.

IPython/html/static/notebook/js/menubar.js
```diff
@@ -39,6 +39,101 @@ define([
...
         this.save_widget = options.save_widget;
         this.quick_help = options.quick_help;
+        this.actions = options.actions;
+        if(!options.noinit){
+            this.init()
+        }
+        Object.seal(this);
+        
+
+        this.addMenu("Edit",[
```
minrk: so we're using camelCase now?

Carreau:
Still moving, and heading to us embassy tomorrow morning so limited bandwidth. So feel free to move forward. 

Envoyé de mon iPhone

Carreau:
Marking pessimistically as 4.0.

minrk:
I'll pick this one up, you focus on visa stuff.

IPython/html/static/notebook/js/menubar.js
jdfreder: :-1:

minrk:
@Carreau how were you planning to handle the fact that most things in the menu are not actions? Were you planning to add a bunch of actions, or have a bunch of pseudo_actions here, like in the toolbar?

Carreau:
I was planning to move some things to actions when it make sens (toggle all outputs... ) and add pseudo action for the rest (for now) something like `<#aname>` could be special cased to only append the div, with the `aname` id, and let the rest of the code works for now.

takluyver:
From the dev meeting: this is a nice thing, but not critical, and it's tricky to check exhaustively for bugs that it may introduce, so we're punting it to 4.x in the interests of actually making a 3.0 release.

Carreau:
I'll close and open an issue to redo it.

IPython/html/static/notebook/js/menubar.js
jdfreder: heh, my opinion has since changed :+1:

PR #7306, opened 2014-12-24
css spinner

Carreau:
@gibiansky said: 

> [...] Another user played around with the notebook, then accidentally set off a very long-running computation. He then was very confused as to why no more cells were evaluated – In[_] turns out to be a mediocre way to signal "I'm busy", and he just thought that everything had broken and he had to restart the application and that it was buggy. I think that at least might be easily fixed by letting cells have some sort of spinner indicating business instead of In[_], people are more used to that.

This is my attempt at fixing it with pure css : 

capture d ecran 2014-12-24 a 14 06 38

Smooth fade in/out and height change, rotate while the cell as the `.running` class.

technically I would prefer something like [that](http://css-spinners.com/#/spinner/throbber/) but it require to inject a placeholder for the spinner.

Design is of course to be discussed.

### DIFF
--- a/IPython/html/static/style/style.less
+++ b/IPython/html/static/style/style.less
@@ -112,28 +112,119 @@
     margin-bottom: 2em;
 }
 
-.inner_cell::after {
-  transition: all 0.5s;
-  content:'';
-  -webkit-animation: spinner 1500ms infinite linear;
-  -moz-animation: spinner 1500ms infinite linear;
-  -ms-animation: spinner 1500ms infinite linear;
-  -o-animation: spinner 1500ms infinite linear;
-  animation: spinner 1500ms infinite linear;
+
+
+@-webkit-keyframes blink {
+    @cin : orange;
+    @cout : lightblue;
+    0%   { box-shadow: 10px 00px 0px rgba(0,0,0,0),
+                       20px 00px 0px @cin,
+                       30px 00px 0px @cout,
+                       40px 00px 0px @cout,
+                       50px 00px 0px @cout,
+                       60px 00px 0px @cout
+                       ;}
+    
+    30%{ box-shadow:   20px 00px 0px @cin,
+                       30px 00px 0px @cin,
+                       40px 00px 0px @cout,
+                       50px 00px 0px @cout,
+                       60px 00px 0px @cout,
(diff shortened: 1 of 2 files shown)


damianavila:
>I can do that. Though it is using box-shadow, so it is not super CPU friendly.
>We should consider using not-pure css. 

I think we need to maintain all the icon styling congruent... so although you did a nice job with the pure css animation I think we need to go the font-awesome ones...

Carreau:
Something is wrong even with fa-spinner. Th CPU climb by 5-10% usage as soon as you have a few cells with spinner. (and increase with number of 'running cells'):
capture d ecran 2015-01-08 a 11 06 28

is ~2% on iddle. 

Figured it out because my laptop fan noise.

damianavila:
mmm... that's too much... not nice, I agree...

ellisonbg:
So what are our options that are CPU friendly? None at all?

On Thu, Jan 8, 2015 at 7:19 AM, Damián Avila <notifications@github.com>
wrote:

> mmm... that's too much... not nice, I agree...
>
> —
> <https://github.com/ipython/ipython/pull/7306#issuecomment-69193327>.
>



-- 
Brian E. Granger

Carreau:
css spinner should be CPU friendly from my understanding. 
They seem not to be, but I don't know why. Gif are an alternative. 
I found https://github.com/FortAwesome/Font-Awesome/issues/701 but they say "Chrome Bug".

Or actually js rotation with a fixed fps framerate.

Carreau:
http://stackoverflow.com/questions/13176746/css-keyframe-animation-cpu-usage-is-high-should-it-be-this-way is a bit old, but we can try;

damianavila:
I guess the gif could be a better alternative... there is still high CPU activity with the link you posted above...

ellisonbg:
Added to the dev meeting agenda for next Tuesday 9:30 am PST.

jdfreder:
@Carreau CPU issues aside, I really like your prompt spinner! :)

Carreau:
> @Carreau CPU issues aside, I really like your prompt spinner! :)

Thanks !

Carreau:
closing in favor of issue #7577 !

PR #7216, opened 2014-12-14
some code quality cleanup.

Carreau:
do to some report of static analysis tool, unused variable (mainly that)
a few function re-declarations, a python2 print statement...

cf https://www.codacy.com/public/bussonniermatthias/ipython/issues

PR #7315, opened 2014-12-26
miscellaneous cleanup

Carreau:
Change pattern that are prone to error, like function redefinition
and other.

Same as #7216  but twice smaller. 

I stil believe we should do something like that sooner than later as the huge disparity of code style make it really difficult to use tools like jshint that report hundreds of potential mistakes. 

If this is still controversial, I can do almost each fix independently.


damianavila:
I think this is ready to go...

PR #7324, opened 2014-12-27
DEV: Refactor checkpoint logic out of FileContentsManager into a separate class.

ssanderson:
The purpose of this change is to provide an API for users to replace just the checkpointing behavior of a ContentsManager.  This is achieved by delegating checkpoint management to a dedicated `CheckpointManager`.

 The impetus for this change came after I completed a full Contents API implementation in [pgcontents](https://github.com/quantopian/pgcontents) and realized that in many cases I actually wanted the files managed by my ContentsManager to be present in the filesystem.  (This was most notably true when uploading data files.)  After playing around with alternatives for a while, I realized that the "Checkpoints API" can be implemented independently of the rest of the Contents API.  This is valuable because it makes it possible to separate the storage backend for the current working version of a file from the storage of previous versions.  This significantly lowers the barrier to entry for doing remote storage of checkpoints as well as things like managing checkpoints through version control.

**Changes:**
- Add a `CheckpointManager` base class and infrastructure for creating a
  `checkpoint_manager` instance attribute on `ContentsManager`.
- Provide default implementations of `delete` and `rename` in the base
  `ContentsManager` class.  **`ContentsManager` subclasses are now
  required to implement `delete_file` and `rename_file`.**  These methods
  no longer need to manage checkpoints.
- Move checkpoint-related functionality from `FileContentsManager` to a
  dedicated `FileCheckpointManager` subclass.
- Move shared filesystem interaction logic into `FileManagerMixin`
  used by both `FileContentsManager` and `FileCheckpointManager`.
- Minor tweaks to ContentsManager tests to get methods from the right
  object.


ssanderson:
@minrk I removed the unncessary notary calls you pointed out.  As long as trust information is preserved by round-tripping through nbformat.read and nbformat.write, there should be no change in trust behavior.

ssanderson:
I've also added support for making non-notebook text files first-class citizens in the checkpoint API.

Previously checkpointing of non-notebooks worked, (and, in fact, happened on save because we ensured that there was always a checkpoint defined for **any** file), but was left out of the Contents specification.  Checkpoints for non-notebooks seems a lot more useful now that the text editor has landed, so I figured I might as well make them actually supported.

ssanderson:
Pushed a minor tweak to the API here: `get_checkpoint` returns a model dict suitable for passing directly to `ContentsManager.save` now.

ssanderson:
Also added an analog of `os.walk` for arbitrary ContentsManagers.  I was using this to implement synchronization between contents backends and realized that it could be written entirely in terms of the contents API.  Normally I'd put that on a separate PR, but I'm basing an internal testing deployment on the PR branch here.

@minrk any more thoughts on this direction for the Contents API?

minrk:
@ssanderson why did you add walk? It's not used anywhere.

minrk:
I think I'm okay with splitting Contents and Checkpoints, if it really helps. I really don't like the fact that there's a mixin required for the most basic case. I think we use far too many of them already, but I'm not sure of a better way other than leaving it as one class, since I think the two really are deeply coupled.

minrk:
ContentsManagers are not generic APIs for performing all kinds of actions on any files, they are abstractions for use in the REST API, so there shouldn't be any functionality in ContentsManager that's not used by the REST API. With that in mind, I wouldn't 

@ssanderson is there a way for the basic case to still use shutil copy, move actions? I still think that's the right thing to do when both are on the filesystem. There doesn't need to be an abstraction separating the two, it just gets in the way.

ssanderson:
@minrk I'm using `walk` in pgcontents to implement synchronization between different contents backends.  The pseudocode is roughly:
```
def sync(mgr_1, mgr_2):
    """Sync contents managed by mgr_1 to mgr_2."""
    for dirname, subdir, files in mgr_1.walk():
        for path in files:
            # In reality this is slightly more complicated because there's some munging
            # of models.
            content = mgr_1.get(file, content=True)
            mgr_2.save(content)
```
I'm happy to pull out `walk` and have it live in pgcontents if you'd rather not support it here.  It seemed worthwhile to add since it's a useful operation that can be written entirely in terms of the existing API.

ssanderson:
As far as the existing mixin implementation goes, that could be removed by renaming that class to something like `FileSystemIO` and putting an instance of it on each Manager instead of making it a mixing.  That would help with the inheritance problem of "where the hell is this method defined?".

ssanderson:
> @ssanderson is there a way for the basic case to still use shutil copy, move actions? I still think that's the right thing to do when both are on the filesystem. There doesn't need to be an abstraction separating the two, it just gets in the way.

In my mind, the essential benefit of splitting ContentsManager from CheckpointsManager is that, if you write ContentsManager without making any assumptions about the storage backend used by your CheckpointsManager, then you're guaranteed that your implementation is still correct if you swap in a different CheckpointsManager.  Any implementation that uses `shutil.copy` or similar would require either that ContentsManager know where to move files for use by CheckpointManager, or that CheckpointManager know where to move files for use by ContentsManager, which would effectively negate the benefit of the separation.

You could of course rewrite `FileCheckpointManager.create_checkpoint` to something like:
```
def create_checkpoint(self, path):
    if isinstance(self.checkpoint_manager, FileCheckpointManager):
        # Do stuff that makes assumptions about shared filesystem.
    else:
        super(FileContentsManager, self).create_checkpoint(path)
```
My worries with this would be that: 1.) You lose the benefits of abstraction outlined above, and 2.) The `else` path there becomes code that's likely to be unused in normal usage, which makes it easier to introduce bugs in future changes.

minrk:
@ssanderson the issue with shutil vs model seems pretty generic to abstraction - using higher abstractions, it can be correct and reusable in more situations, but it eliminates the ability to do the simpler or preferable thing when the two storage mechanisms are the same - e.g. both on the filesystem or both in the same database. You have to read the whole notebook even when you could have created the checkpoint in an entirely remote operation without any potentially expensive reads.  This is why I think we *should* make assumptions about the ContentsManager for a given checkpoint mechanism.

For that reason, I think maybe changing the API to take just a path, and letting the CheckpointsManager decide how to get the data from ContentsManager will allow Contents and Checkpoints to be implemented in a more intimate way, which I expect to be the way to go most of the time. I would expect the 'get the generic model' route to be takenl only when the CheckpointsManager is a poor choice for use with a given ContentsManager.

As for walk, when we ship stuff we don't use, it tends to go stale and break without us knowing about it (same applies to your point about special-casing FileContentsManager with one branch taken in ~all cases, so I'm arguing against myself a bit here). This happens all the time (in-process kernel, pure Python kernel, etc.), and I would like to avoid it when possible. So it would be great if you shipped walk with pgcontents where you actually use it, rather than here where we don't, even though it's a generically useful operation.

ssanderson:
> the issue with shutil vs model seems pretty generic to abstraction - using higher abstractions, it can be correct and reusable in more situations, but it eliminates the ability to do the simpler or preferable thing when the two storage mechanisms are the same

@minrk yep, wholeheartedly agree that this is an inherent tradeoff.  It's also not unexpected that, as the author of a library that's interested in abstracting the differences between storage backends, I'm inclined to advocate for the interface that's more abstract (and consequently more implementation-constraining), whereas as the maintainer of the default implementation, you're inclined to advocate for the interface that makes the most common case more expedient.

>  I would expect the 'get the generic model' route to be taken only when the CheckpointsManager is a poor choice for use with a given ContentsManager.

I disagree with this characterization a bit.  The generic model route is going to be reasonable any time you're storing checkpoints in a different backend than the base contents.  In particular, I think it's appropriate for any setup where you want to use checkpoints as a remote backup system, which is essentially what I'm doing with PostgresCheckpointManager.  

Part of the difficulty here is that there are very few cases where you'd want to use `FileCheckpointsManager` with anything but `FileContentsManager`, because if you have a compelling reason to not store notebooks as local files (e.g. because your filesystem is in the cloud and the cloud is made of lies), you probably don't want to store your checkpoints locally either.  More generally, you probably want your Checkpoint storage to be at least as durable as your Contents storage.

> For that reason, I think maybe changing the API to take just a path, and letting the CheckpointsManager decide how to get the data from ContentsManager will allow Contents and Checkpoints to be implemented in a more intimate way

I had tried this in an earlier revision and felt like it made the `CheckpointsManager` implementation, though I think that was when I still thought the CheckpointManager would have to care about notary signing, so it might be cleaner than I had originally thought.

ssanderson:
Moved `walk` into a pgcontents utilities directory.

ssanderson:
@minrk adba75fe40d2bd8dedd0b98cfb5a4da6cb2772ee implements your proposal for an API that does a direct copy when possible.  I'm not sure I totally buy that it's an improvement; it's trading a performance gain on `create_checkpoint` and `restore_checkpoint` for code that's more complex (all the logic that was there before is still there, but now there's also an additional "fast path" that does a direct copy) and more tightly coupled (FileCheckpointsManager now has to make stronger assumptions about the attributes available on FileContentsManager).  Moreover, the perf win probably isn't noticeable unless you have really big notebooks, but if that's the case you're still pulling the whole thing into memory and parsing it on every save/load, so the additional cost of doing that one more time on create probably isn't going to be the difference between usable and unusable.

For the record, from the perspective of `pgcontents`, I don't have much of a preference between this implementation and the previous one; the code I already have will continue to work just the same.  My gut feeling is that the added complexity here isn't worth the performance improvement, but I'll defer to your judgment on that.

minrk:
There's yet another option - rather than having both the generic implementation and fast path in the default, we could only ship a CheckpointManager that *does* assume a FileContentsManager. It would still be possible to make a generic CheckpointManager that only uses models, but that doesn't mean the one we ship has to be generic. And a GenericFileCheckpointManager ought to be a very small subclass of AssumingFileContentsFileCheckpointManager.

I'm still not sure what the best approach is, thanks for humoring me. I really appreciate your help on this stuff. At this point in the evening, I'm slightly inclined to either require FileContentsManager in the one we ship or your original proposal of just the model version. Implementing both with an if statement seems creaky.

ssanderson:
I think it makes sense to split the implementations, and after staring at http://en.wikipedia.org/wiki/Liskov_substitution_principle for longer than I care to admit, I agree that Specific > Generic is the right subtyping relationship.  The generic one essentially accepts ContentsManager in its method input signatures, whereas the specific one accepts FileContentsManger, and method inputs are *contra*variant.

Any chance I could convince you to keep the `GenericFileCheckpointsManager` in IPython proper?  It seems a shame to do all the design work to make it possible to use truly generic backends, but then put the generic FileCheckpointManager in a library dedicated to a specific backend.

minrk:
> Any chance I could convince you to keep the GenericFileCheckpointsManager in IPython proper?

I think that would be fine. You deserve at least that for doing all the work.

minrk:
It also serves as an illustrative example, so there's a good reason for having it around.

ssanderson:
@minrk this now implements `FileCheckpoints` and `GenericFileCheckpoints` (I dropped manager because I kept misreading ContentsManager vs CheckpointsManager) as separate classes, with the latter being a subtype of the former.

As of this change, the "Checkpoints API" is the following:

- `create_checkpoint`
- `restore_checkpoint`
- `delete_checkpoint`
- `rename_checkpoint`
- `list_checkpoints`

All checkpoints managers will implement `delete`, `rename`, and `list` "by hand".

`create_checkpoint` and `restore_checkpoint` are special because they transgress boundaries between the checkpoints storage backend and the contents storage backend.  You can implement these methods directly, which allows you to make assumptions about the storage backend of your ContentsManager (this is what's done by `FileCheckpointManager`), or you can use `GenericCheckpointsMixin` and implement handlers that take models instead:

- `get_file_checkpoint`
- `get_notebook_checkpoint`
- `create_file_checkpoint`
- `create_notebook_checkpoint`

One nice thing about this architecture is that you can make a portable version of any non-portable CheckpointManager by subclassing the non-portable version, adding GenericCheckpointsMixin, and implementing the four additional methods.  This is how `GenericFileCheckpointManager` is implemented, and it's likely that I'll do something similar in pgcontents so that I can do fewer roundtrips when I know I'm in postgres for both contents and checkpoints.

ssanderson:
I also moved the checkpoints classes and the FileManagerMixin into their own files since they were starting to get a bit long.

ssanderson:
I think this now meets all of my needs for pgcontents, and having the two APIs outlined above feels like this strikes a good balance between making it easy to do the portable thing and making it easy to do the non-portable-but-more-efficient thing.

minrk:
@ssanderson excellent, thanks.

PR #7346, opened 2014-12-29
Use page backdrop color in terminal page

### LINKED ISSUE #7295: Page backdrop color for terminal page

jdfreder:
This was supposed to make it in with #7242 , but it looks like I forgot to push the commit.

closes #7295

### DIFF
--- a/IPython/html/static/terminal/less/terminal.less
+++ b/IPython/html/static/terminal/less/terminal.less
@@ -1,23 +1,36 @@
-.terminal {
-  float: left;
-  font-family: monospace;
-  color: white;
-  background: black;
-  padding: @code_padding;
-  border-radius: @border-radius-base;
-  .box-shadow(@global-shadow-dark);
+body.terminal-app {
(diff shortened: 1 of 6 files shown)


IPython/html/static/terminal/less/terminal.less
```diff
@@ -1,23 +1,36 @@
...
-  &, dummy-screen {
-    line-height: 1em;
-    font-size: @notebook_font_size;
+  #header {
+    background: @body-bg;
+    .box-shadow(@global-shadow);
+
+    .header-bar {
+      margin-bottom: -1px;
+    }
```
damianavila: Probably it has to be applied broadly...

jdfreder:
Comments addressed :)

IPython/html/static/terminal/less/terminal.less
jdfreder: Ah got it, should be changed in latest commit.

PR #7357, opened 2014-12-31
When executing a cell, clear output and ignore queued animations.

### LINKED ISSUE #5892: Traceback overlay next cell if was widget.
capture decran 2014-05-23 a 10 22 10

Pretty simple to reproduce here, but I might not be on last master though. 
```
from IPython.html.widgets import interactive

def fun(a=0):
    return a
#w._  # comment this line,  run once uncomment rerun 
w = interactive(fun)
display(w)
```

jdfreder:
closes #5892

### DIFF
--- a/IPython/html/static/notebook/js/codecell.js
+++ b/IPython/html/static/notebook/js/codecell.js
@@ -416,7 +416,7 @@ define([
             return;
         }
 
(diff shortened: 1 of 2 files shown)


Carreau:
+1

PR #7362, opened 2015-01-02
Allow completion on the numpy struct itself

mbyt:
Removes the constraint to only allow completion on arrays of numpy structs, but not on numpy structs itself. Current master:

```
In [2]: import numpy as np
In [3]: dt_head = np.dtype([('dt', '>u4'), ('df', '>u4'), ('Fs', '>u4')])
In [4]: dt = np.dtype([('head', dt_head), ('data', '>u4', 10)])
In [5]: x = np.zeros(2, dtype=dt)
In [6]: x[<TAB> =>
x['data']  x['head']  
In [6]: y = x[0]
In [7]: y[<TAB>=> "nothing happens"
```

This pulls request allows completion for `In [7]`.

P.S.: Just happily discovered that dict and numpy structured array completion are supported in current master. For me, this is a killer feature.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -863,7 +863,8 @@ class IPCompleter(Completer):
                     return list(obj.keys())
                 except Exception:
                     return []
-            elif _safe_isinstance(obj, 'numpy', 'ndarray'):
+            elif _safe_isinstance(obj, 'numpy', 'ndarray') or\
+                 _safe_isinstance(obj, 'numpy', 'void'):
                 return obj.dtype.names or []
             return []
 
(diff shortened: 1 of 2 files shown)


mbyt:
As far as I understand, the test failure 10320.1 https://travis-ci.org/ipython/ipython/builds/45667961 is unrelated.

Carreau:
> As far as I understand, the test failure 10320.1 on https://travis-ci.org/ipython/ipython/builds/45667961 is unrelated.

Sure, relaunched.

Carreau:
Seem sensible to me. 

+1

PR #7363, opened 2015-01-02
fix widget description can not display when use --no-mathjax

dongweiming:
when i use `--no-mathjax`: 

``` python
ipython notebook --no-mathjax
```

i found widget's description not work. the div with class `widget-label` is empty.


jdfreder:
@dongweiming thanks!

PR #7373, opened 2015-01-04
store widget state in single key

### LINKED ISSUE #7371: WrappedError when saving notebook
When trying to save a notebook I get the following error in the follwing message in the Javascript console:
error-saving

System:
- current master
- Anaconda Python 3
- Virtualbox Windows 7
- Chrome

minrk:
rather than in a dict, which doesn't work without reserializing _all_ stored widget states (of all notebooks), since localStorage only supports storing strings.

closes #7371

ping @jdfreder


jdfreder:
Thanks, sorry for the dumb mistake. :(  I thought it was a simple fix, and I didn't pay enough attention to what I was doing.  :+1: to merge

PR #7375, opened 2015-01-04
ensure data, metadata are set in execute_results

minrk:
some circumstances such as `_ipython_display_` callers, can result in execute_result mimebundle data not being set, which confuses the javascript.

Added handling of this case to the Javascript, in case other kernels have similar bugs.

### DIFF
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -243,11 +243,17 @@ define([
         'text/plain'
     ];
 
-    OutputArea.prototype.validate_mimebundle = function (json) {
-        /**
-         * scrub invalid outputs
-         */
-        var data = json.data;
+    OutputArea.prototype.validate_mimebundle = function (bundle) {
+        /** scrub invalid outputs */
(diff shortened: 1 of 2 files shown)


IPython/html/static/notebook/js/outputarea.js
```diff
@@ -243,11 +243,17 @@ define([
...
+        /** scrub invalid outputs */
+        if (typeof bundle.data !== 'object') {
+            console.warn("mimebundle missing data", bundle);
+            bundle.data = {};
+        }
+        if (typeof bundle.metadata !== 'object') {
+            console.warn("mimebundle missing metadata", bundle);
+            bundle.metadata = {};
+        }
+        var data = bundle.data;
```
damianavila: why not define data at the top... and the do `typeof data...`

minrk: doesn't really make a different, but this avoids assigning data twice if it's undefined.

damianavila: >  this avoids assigning data twice

ok... this reason is enough... LGTM

PR #7377, opened 2015-01-04
make stop_on_error behavior optional

juhasch:
Followup to https://github.com/ipython/ipython/pull/6521

Rebased for current master and addressed comments by @jhamrick:
- added additional js test
- added kernel side tests

### DIFF
--- a/IPython/kernel/zmq/kernelbase.py
+++ b/IPython/kernel/zmq/kernelbase.py
@@ -348,6 +348,11 @@ class Kernel(SingletonConfigurable):
             self.log.error("%s", parent)
             return
         
+        if u'skip_exceptions' in content and content[u'skip_exceptions'] is True:
+            skip_exceptions = True
+        else:
+            skip_exceptions = False
+
         md = self._make_metadata(parent['metadata'])
         
         # Re-broadcast our input for the benefit of listening clients, and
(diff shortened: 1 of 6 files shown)


juhasch:
Travis is failing in the JS tests because it looks like in case of an execption when executing a Python command, the kernel is killed. This used to work before. 

Obviously if testing for exceptions kills the kernel, this can't be tested. Is this intended or am I doing something wrong here ?

minrk:
Killing the kernel is definitely not the normal behavior for exceptions in the kernel.

I do think that it makes more sense for this to be a 'stop on exception' key that defaults to true, rather than a 'skip exceptions' that defaults to false.

minrk:
@juhasch we discussed this at our dev meeting. Can you rename the flag `stop_on_error` instead (and invert the bool logic, accordingly)? Hopefully we can finally get this one in. Sorry for the wait!

IPython/kernel/zmq/kernelbase.py
```diff
@@ -348,6 +348,11 @@ def execute_request(self, stream, ident, parent):
             self.log.error("%s", parent)
             return
         
+        if u'skip_exceptions' in content and content[u'skip_exceptions'] is True:
```
minrk: this can be `stop_on_error = content.get('stop_on_error', True)`

minrk:
Thanks. I found the problem with the test - it's waiting for output that never arrives. I've added a commit that fixes that, and rebased to remove the components commit and the merges. Are you comfortable pulling [my branch]( and pushing it here? It would go something like this:

(in your skip-exceptions2 branch)

```bash
git remote add minrk https://github.com/minrk/ipython.git
git fetch minrk
git reset minrk/stop_on_error --hard # make your local branch match mine
git push -f # forcefully replace the remote branch (this PR) with the change
```

If not, I can open another PR from my branch and we can merge it.

juhasch:
Done. Thanks for helping me.

minrk:
Thanks again for your patience on this one.

PR #7379, opened 2015-01-05
Js tests

juhasch:
The call is `IPython.notebook.insert_cell_at_index(type, index )`


damianavila:
LGTM

Carreau:
How could the test be passing with that ? 

+1

damianavila:
>How could the test be passing with that ?

Javascript :P

minrk:
wat

PR #7382, opened 2015-01-05
Highlight cell if in edit mode

valhallasw:
Currently, there are very few visual clues to inform the user which
mode the notebook is currently in: a thin line around the cell, and
a small 'edit' icon in the toolbar. This change adds a background
color to the cell, which makes the visual clue much clearer.

### DIFF
--- a/IPython/html/static/notebook/less/cell.less
+++ b/IPython/html/static/notebook/less/cell.less
@@ -8,9 +8,11 @@ div.cell {
 
     &.selected {
         border-color: @border_color;
-        /* Don't border the cells when printing */
+        background-color: lightgreen;
+        /* Don't border or background the cells when printing */
         @media print {
             border-color: transparent;
+            background-color: transparent;
         }
     }
 
--- a/IPython/html/static/notebook/less/cell.less
+++ b/IPython/html/static/notebook/less/cell.less
@@ -7,9 +7,8 @@ div.cell {
     border-style: solid;
 
     &.selected {
-        border-color: @border_color;
-        background-color: lightgreen;
-        /* Don't border or background the cells when printing */
+        background-color: #efefef;
+        /* Don't show the background when printing */
         @media print {
             border-color: transparent;
             background-color: transparent;
@@ -17,10 +16,10 @@ div.cell {
     }
 
     .edit_mode &.selected {
-        border-color: green;
-        /* Don't border the cells when printing */
+        background-color: #ded;
+        /* Don't show the background when printing */
         @media print {
-            border-color: transparent;
+            background-color: transparent;
         }
     }
 


Carreau:
Lightgreen maybe a bit strong, but I agree that another visual indication of the mode would be nice. 
I'm not sure everybody would agree though. 

Implementation wide, I would not hardcode a color but link to a variable so that theming is easier, 
but that's a detail. 

Nice touch to though of forcing transparent on media-print.

valhallasw:
Yeah, I'm not completely sure about the color. On the one hand, it has to have enough contrast with white to be visible, and on the other hand, it has to match the border color. 

Some options for colors:

lightgreen: (indeed a bit strong)
image

palegreen: (better)
image

lightyellow: (not enough contrast to the white surroundings, doesn't match well with border color)
image

lightblue:
image

Of these, the last might actually the best, especially if the border is also changed to a blue-ish hue. That also matches the IPython logo better, I'd say?

(I also just realized I actually implemented this in the wrong place -- it's in .selected, while it should of course be in .edit_mode. I'll fix that after there is some consensus on the actual style)

Carreau:
Yeah, but IPython logo is going away :-) we can also make the full body colored in edit mode. :-P

minrk:
This is similar to what I do in my custom css. I use `#ded` for edit mode and `#efefef` for command mode:

screen shot 2015-01-05 at 11 49 03

screen shot 2015-01-05 at 11 47 27

valhallasw:
@minrk: Removing the border, and just having the background definitely looks good.

I just noticed variables.less already has a @cell_selected_background variable, but it doesn't seem to be in use anywhere, other than to define border_color and lighter_border_color?

ellisonbg:
Trying this out now...

ellisonbg:
Here is the look of the `#efefef` and `#ded` colors with our current borders and the new document grey background:

screen shot 2015-01-09 at 12 07 11 pm

screen shot 2015-01-09 at 12 07 35 pm

On the grey one, I think we are getting into a situation where we have too many subtle different greys on the page, which makes them all less effective. I think we do want to minimize our usage of large shaded areas on the page. So not sure about the grey one.

For the green one, I think we would need to pick a very subtle treatment. The stronger colors are just too much on the eye. I think the `#ded` version is in the right direction and definitely gives more emphasis to edit mode.

Honestly, this is exactly the type of design work though that is best when two people can sit down in person and try different things in real time. @jdfreder and I could do that next Wed.

Adding to the dev meeting for next week to get other ideas.

Carreau:
I used to use only `left-margin : thick solid <color>` as in  toward the last markdown cell.

It avoids the big colored area and i fell convey the notion of "cursor" much more.

damianavila:
>I used to use only left-margin : thick solid <color> as in...

I use something similar but I guess it could be a little bit subtle for the common user...

ellisonbg:
I like your treatment @Carreau, but I don't think is quite makes sense for
the default.

On Sun, Jan 11, 2015 at 8:41 AM, Damián Avila <notifications@github.com>
wrote:

> I used to use only left-margin : thick solid as in...
>
> I use something similar but I guess it could be a little bit subtle for
> the common user...
>
> —
> <https://github.com/ipython/ipython/pull/7382#issuecomment-69500703>.
>



-- 
Brian E. Granger

ellisonbg:
I will explore a few more options.

ellisonbg:
We all agreed to leave command mode as is, but to explore the different variants of the green background.

Carreau:
I'll close in favor of #7696 that references here.

PR #7383, opened 2015-01-05
tests for ScriptExporter

### LINKED ISSUE #7378: No encoding declared on Python output of nbconvert
I run *ipython nbconvert my_notebook.ipynb --to python* and what I get is a Python file which can't run due to the following error:
```Python
SyntaxError: Non-ASCII character '\xd7' in file my_notebook.py on line 212, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```

One of my cells output contains a non-ascii character, and when I add the encoding comment to the first line of my script:
```Python
# -*- coding: UTF-8 -*-
```

the script is usable again.

However, IPython nbconvert does not include a command line flag that allows me to do either of the following two options:
* Supply an encoding string that will be use in the *coding:* comment to the first line,
* Convert to Python code without the cells output.

Any solution will be appreciated.

BTW, I also didn't find a way to make Python run a .py script using a command-line specified encoding... But that's another deal.

minrk:
verify that it delegates to PythonExporter if specified in language_info

closes #7378

### DIFF
--- a/IPython/nbconvert/exporters/tests/test_python.py
+++ b/IPython/nbconvert/exporters/tests/test_python.py
@@ -1,23 +1,11 @@
 """Tests for PythonExporter"""
 
-#-----------------------------------------------------------------------------
-# Copyright (c) 2013, the IPython Development Team.
-#
+# Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
-#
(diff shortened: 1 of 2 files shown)


damianavila:
I don't think the Travis failure is related... code looks good to me...

PR #7387, opened 2015-01-05
allow system-wide installtion of nbextensions

minrk:
- like kernelspecs, default install is `/usr/local/share/jupyter/nbextensions` (PROGRAMDATA on Windows)
- user install goes to .ipython/nbextensions
- unlike kernelspecs, arbitrary destinations are allowed, because the nbextension path is already configurable

### DIFF
--- a/IPython/html/nbextensions.py
+++ b/IPython/html/nbextensions.py
@@ -25,6 +25,8 @@ from IPython.utils.path import get_ipython_dir, ensure_dir_exists
 from IPython.utils.py3compat import string_types, cast_unicode_py2
 from IPython.utils.tempdir import TemporaryDirectory
 
+class ArgumentConflict(ValueError):
+    pass
 
 # Packagers: modify the next block if you store system-installed nbextensions elsewhere (unlikely)
 SYSTEM_NBEXTENSIONS_DIRS = []
@@ -140,7 +142,7 @@ def install_nbextension(files, overwrite=False, symlink=False, user=False, prefi
         set verbose=2 for more output, or verbose=0 for silence.
     """
     if sum(map(bool, [user, prefix, nbextensions_dir])) > 1:
-        raise ValueError("Cannot specify more than one of user, prefix, or nbextensions_dir.")
+        raise ArgumentConflict("Cannot specify more than one of user, prefix, or nbextensions_dir.")
     if user:
         nbext = pjoin(get_ipython_dir(), u'nbextensions')
     else:
@@ -296,10 +298,6 @@ class NBExtensionApp(BaseIPythonApplication):
         help="Verbosity level"
     )
     
-    def check_install():
-        if sum(map(bool, [user, prefix, nbextensions_dir])) > 1:
-            raise TraitError("Cannot specify more than one of user, prefix, or nbextensions_dir.")
-    
     def install_extensions(self):
         install_nbextension(self.extra_args,
             overwrite=self.overwrite,
@@ -318,8 +316,11 @@ class NBExtensionApp(BaseIPythonApplication):
                     for ext in os.listdir(nbext):
                         print(u"    %s" % ext)
         else:
-            self.check_install()
-            self.install_extensions()
+            try:
+                self.install_extensions()
+            except ArgumentConflict as e:
+                print(str(e), file=sys.stderr)
+                self.exit(1)
 
 
 if __name__ == '__main__':
--- a/IPython/html/nbextensions.py
+++ b/IPython/html/nbextensions.py
@@ -83,15 +83,15 @@ def _safe_is_tarfile(path):
         return False
 
 
-def check_nbextension(files, nbextensions=None):
+def check_nbextension(files, nbextensions_dir=None):
     """Check whether nbextension files have been installed
     
     files should be a list of relative paths within nbextensions.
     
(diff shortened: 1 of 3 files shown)


IPython/html/nbextensions.py
```diff
@@ -25,6 +26,34 @@
...
+
+if os.name == 'nt':
+    programdata = os.environ.get('PROGRAMDATA', None)
+    if programdata: # PROGRAMDATA is not defined by default on XP.
+        SYSTEM_NBEXTENSIONS_DIRS = [pjoin(programdata, 'jupyter', 'nbextensions')]
+    prefixes = []
+else:
+    prefixes = ['/usr/local', '/usr']
+
+# add sys.prfix at the front
```
damianavila: `six.prefix`

IPython/html/nbextensions.py
```diff
@@ -96,16 +127,29 @@ def install_nbextension(files, overwrite=False, symlink=False, ipython_dir=None,
...
-    ipython_dir : str [optional]
-        The path to an IPython directory, if the default value is not desired.
-        get_ipython_dir() is used by default.
+    user : bool [default: False]
+        Whether to install to the user's .ipython/nbextensions directory.
+        Otherwise do a system-wide install (e.g. /usr/local/share/jupyter/nbextensions).
+    prefix : str [optional]
+        Specify install prefix, if it should differ from default (e.g. /usr/local).
+        Will install to prefix/share/jupyter/nbextensions
+    nbextensions : str [optional]
```
damianavila: Just a minor, but I find the `nbextensions` arg name a little confusing... could not be modified to something pointing to the abs path idea?

damianavila:
Just really minor comments, otherwise, LGTM...

IPython/html/nbextensions.py
Carreau: `sys.prefix`, someone will get it right !

IPython/html/nbextensions.py
Carreau: +1

IPython/html/nbextensions.py
```diff
@@ -238,25 +289,36 @@ class NBExtensionApp(BaseIPythonApplication):
...
+    user = Bool(False, config=True, help="Whether to do a user install")
+    prefix = Unicode('', config=True, help="Installation prefix")
+    nbextensions = Unicode('', config=True, help="Full path to nbextensions (probably use prefix or user)")
     verbose = Enum((0,1,2), default_value=1, config=True,
         help="Verbosity level"
     )
     
+    def check_install():
+        if sum(map(bool, [user, prefix, nbextensions])) > 1:
+            raise TraitError("Cannot specify more than one of user, prefix, or nbextensions.")
```
Carreau: Duplicated logic of line 142 ?

Carreau: And missing `self` ?

dsblank:
Thanks for this! The prefix option is also one that I suggested as an option for installing kernelspecs:

http://mail.scipy.org/pipermail/ipython-dev/2014-December/015673.html

If someone could answer that question, it would be useful.

IPython/html/nbextensions.py
minrk: self is missing, yes. This is duplicated because the ValueError is treated as a crash, where TraitError isn't. I think I can make it better without duplicating the check.

IPython/html/nbextensions.py
minrk: for consistency, I'll make it nbextensions_dir

IPython/html/nbextensions.py
```diff
@@ -25,6 +26,34 @@
...
+# Packagers: modify the next block if you store system-installed nbextensions elsewhere (unlikely)
+SYSTEM_NBEXTENSIONS_DIRS = []
+
+if os.name == 'nt':
+    programdata = os.environ.get('PROGRAMDATA', None)
+    if programdata: # PROGRAMDATA is not defined by default on XP.
+        SYSTEM_NBEXTENSIONS_DIRS = [pjoin(programdata, 'jupyter', 'nbextensions')]
+    prefixes = []
+else:
+    prefixes = ['/usr/local', '/usr']
```
jasongrout: `pjoin(os.sep, 'usr', 'local')`

IPython/html/nbextensions.py
```diff
@@ -25,6 +26,34 @@
...
+
+# add sys.prfix at the front
+if sys.prefix not in prefixes:
+    prefixes.insert(0, sys.prefix)
+
+for prefix in prefixes:
+    nbext = os.path.join(prefix, 'share', 'jupyter', 'nbextensions')
```
jasongrout: `pjoin`

IPython/html/nbextensions.py
minrk: removed the duplicate check

IPython/html/nbextensions.py
minrk: done

IPython/html/nbextensions.py
minrk: done

minrk:
review should be addressed

Carreau:
+1, Merging.

PR #7391, opened 2015-01-06
Base Docker image on debian:jessie

rgbkrk:
Switches our base image over to `debian:jessie`, as suggested by @ogrisel in https://github.com/ipython/docker-notebook/pull/33#issuecomment-68837395.

Note: The tests are failing due to encoding issues, I'll post logs in a comment below.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,10 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # Not essential, but wise to set the lang
 # Note: Users with other languages should set this in their derivative image
-RUN apt-get update && apt-get -y install locales && locale-gen en_US.UTF-8 && dpkg-reconfigure locales
+RUN apt-get update && \
+    apt-get -y install locales \
+    && echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen \
+    && dpkg-reconfigure locales
 ENV LANGUAGE en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,13 +9,10 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # Not essential, but wise to set the lang
 # Note: Users with other languages should set this in their derivative image
-RUN apt-get update
+RUN apt-get update && apt-get -y install locales && locale-gen en_US.UTF-8 && dpkg-reconfigure locales
 ENV LANGUAGE en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
-
-RUN locale-gen en_US.UTF-8
-RUN dpkg-reconfigure locales
 
 # Python binary dependencies, developer tools
 RUN apt-get update && apt-get install -y -q \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Installs IPython from the current branch
 # Another Docker container should build from this one to get services like the notebook
 
-FROM ubuntu:14.04
+FROM debian:jessie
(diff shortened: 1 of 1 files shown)


minrk:
@rgbkrk can you cherry-pick https://github.com/minrk/ipython/commit/d033ff7c6ff01b0d69357b42454d92b2ec6c2db8 ? I think it fixes the locale. It built successfully here.

rgbkrk:
That'll be my first time cherry picking, I'll try it out!

rgbkrk:
Well that was easy. I'll rebuild now.

rgbkrk:
Working great, including when its with the merge commit (locally at least). This will trigger some work for us to do down in the rest of the images I imagine.

I'll go ahead and reach out on the dev list about this.

rgbkrk:
Rebased too so its more update and so that I can trigger the overall build, let it run on tmpnb and see how that goes.

rdhyee:
How much breakage should I expect from the older ipython/* images that use Ubuntu as the base image?  (As someone who uses ubuntu regularly but not Debian, I don't have a feel for how different the two distributions are.  My understanding is that Ubuntu is based on Debian and therefore I might hope for minimal breakage.)

rgbkrk:
@rdhyee Package names can differ, as well as how they're packaged. In our case, we've noticed issues with how CRAN wants packages put together and how R is built on Ubuntu (but not on Debian). That's one of the factors that led to the shift. The Debian builds of openBLAS also use the `DYNAMIC_ARCH` option that we're doing manually in the scipystack images. This will help us move to that overall setup.

The debian images are a bit smaller by the way.

rdhyee:
Thanks @rgbkrk for your answer, which is very helpful.

FWIW, don't let my dependencies stand in the way of making this change.  I should be able to fix any problems I encounter.  I'm guessing I should also be able to pin my images to older versions of the IPython/* images if I need to do that.

rgbkrk:
If you're able to figure out how to pin to older images on Docker Hub, I'd love to know. I've yet to figure that one out. It doesn't seem to be documented anywhere.

Thanks for chiming in no matter what. I figured hitting up the mailing list was a good idea since this would certainly trigger changes.

rdhyee:
@rgbkrk -- oh yeah, pinning to older images on Docker Hub might not be so easy.  I got stuck on what I think is a related issue: correlated Build IDs on Docker Hub with Image IDs (which I see, for example, with `docker images`).  If I read the answer at https://groups.google.com/forum/#!searchin/docker-user/image$20id$20hub.docker.com/docker-user/YAITMBmS8jU/U9pNLT-POP8J correctly, the answer seems to be that from the point of view of a user (not builder) of an image on Docker Hub, you're currently out of luck.

If I figure how to pin my Dockerfile to an older version of IPython/* I'll let you know.  Please do the same if  you figure it out too. :-)

minrk:
We can tag the last build right before merging this as `:ubuntu`, and leave that up for a little while, right? Maybe with a `docker push`?

rgbkrk:
You can't `docker push` automated builds. We can make a tag (or branch) on github and tag on Docker though.

minrk:
From chat, holding this until after 3.0.

Carreau:
Now that we are post 3.x can we merge ?

rgbkrk:
We may be post 3.x, but where this Dockerfile will belong after the big split is likely in question.

minrk:
I think it will follow the notebook server repo.

minrk:
We can leave ipython/ipython as a single-line `FROM jupyter/notebook` Dockerfile.

rgbkrk:
> We can leave ipython/ipython as a single-line FROM jupyter/notebook Dockerfile.

That sounds great. I'll just close this PR.

PR #7393, opened 2015-01-07
un-nest the edit_mode div.cell.selected css selector

jasongrout:
If we tried nesting this into the div.cell rule by doing
`.edit_mode &.selected`, we would not be able to nest the IPython
css into another selector (since & would then be the full parent,
not just div.cell).

ping @Carreau.  I also tried to regenerate the css, but I ended up with a ton of changes that were not related to these changes.  Since basically all this should do is add a comment, I'll leave off the generating of the css.

### DIFF
--- a/IPython/html/static/notebook/less/cell.less
+++ b/IPython/html/static/notebook/less/cell.less
@@ -14,19 +14,23 @@ div.cell {
         }
     }
 
-    .edit_mode &.selected {
-        border-color: green;
-        /* Don't border the cells when printing */
-        @media print {
-            border-color: transparent;
-        }
-    }
-
     width: 100%;
     padding: 5px 5px 5px 0px;
     /* This acts as a spacer between cells, that is outside the border */
     margin: 0px;
     outline: none;
+}
+
+
+/* we put this edit_mode selector outside of the above div.cell rule because & is the full
+   parent (not just div.cell), so `.edit_mode &.selected` prevents us from nesting this rule
+   within another selector */
+.edit_mode div.cell.selected {
+    border-color: green;
+    /* Don't border the cells when printing */
+    @media print {
+        border-color: transparent;
+    }
 }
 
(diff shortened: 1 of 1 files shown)


Carreau:
a few other potential problematic things: 

```
$ grep -r '&' **/*.less | grep -v node_modules | cut -f2 -d:| sed 's/^[ \t]*//;s/[ \t]*$//' | sort |uniq | grep -v glyph | grep -v -e '^&'
-webkit-box-shadow
.btn & {
.btn-group-lg > .btn { &
.btn-group-sm > .btn { &
.btn-group-xs > .btn { &
.btn-xs & {
.container &,
.container-fluid & {
.edit_mode &.selected {
.list-group-item > & + & {
.list-group-item > & {
.list-group-item.active > &,
.nav-pills > .active > a > & {
.nav-pills > li > a > & {
.navbar > .container &,
.navbar > .container-fluid & {
.navbar-fixed-bottom & {
.navbar-fixed-top &,
.navbar-static-top &,
.open .dropdown-toggle& {
.open > .dropdown-toggle& {
.progress-striped & {
// Bordered & Pulled
// Emphasis & misc
// Rotated & Flipped Icons
a&
a& {
a&.active {
able&  { display
body&.in { display
button& {
d&     { display
d& { display
extarea& {
extarea&,
fieldset[disabled] & {
h&,
r&     { display
r&,
r&.in    { display
select& {
select[multiple]& {
src
```

jasongrout:
Hmmm...this is indicating that I should maybe find a different way to solve my problem...

Carreau:
I was pointing potential problèmes. They may not be...



Envoyé de mon iPhone

jasongrout:
Well, they point out that this problem is deeper than the original fix, and as such, should be designed around instead of affecting the code.  So now I compile the less in two stages.

PR #7398, opened 2015-01-08
Split apart the body attributes and classes in the notebook templates

jasongrout:
This makes it easier to selectively add classes and take advantage of template inheritance to inherit classes on the body element.

ping @Carreau and @ellisonbg


Carreau:
Looks fine to me.

jasongrout:
CC @ssanderson, since I think he may use custom templates...

Issue #7399, opened 2015-01-09
New notebook with same kernel


mathieu1:
Using a notebook with a non-default kernel, shouldn't `File > New` create a new notebook with the *same kernel* as the current one? (Current behaviour is that it creates a notebook with the default kernel)

PR #7401, opened 2015-01-09
create new notebook with same kernel

mathieu1:
Here's a quick implementation of the alternative behavior proposed [here](https://github.com/ipython/ipython/issues/7399) : starting with an existing notebook, creating a new notebook will have the _same kernel_ as the existing notebook.

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -35,7 +35,6 @@ define([
         var change_kernel_submenu = $("#menu-change-kernel-submenu");
         var new_notebook_submenu = $("#menu-new-notebook-submenu");
         
-        // Create the change kernel submenu
         var keys = Object.keys(data.kernelspecs).sort(function (a, b) {
             // sort by display_name
             var da = data.kernelspecs[a].spec.display_name;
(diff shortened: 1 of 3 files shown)


Carreau:
What about new being a submenu that list the kernels availlable ?
(potentially sorting the current kernel as the first of the submenu))

mathieu1:
There we go. I haven't yet put the current kernel at the top (see technical issue below)
Anyhow, does this look like the way to go? Here's what it looks like : 
newnotebook
-------
Technical issue for putting current kernel at the top of the list : at the moment the menu is created `notebook.kernel` is still `null`, so 
* either the menu has to be created at another moment
* or menu entries have to be rearranged after the kernel is set up
* or maybe there's another way to get the kernel name which i've overlooked

dsblank:
+1 Good idea; this will be useful, regardless of details.

mathieu1:
Ok, I went for the second solution. So the current kernel is at the top of the list.
newnotebook
There's only an embryonic attempt at catching errors (non-existing kernel?), should this be delved into further?

Carreau:
Me se only one thing: Scala and spark kernel.

Error handling would be nice, but I guess we can assume kernel exist if they are listed.

mathieu1:
Fine, I'll leave it like that, there are other things to worry about with missing kernels at the moment anyway.

ellisonbg:
I am +1 on this idea and putting the current kernel at the top of the list.

ellisonbg:
The code looks good and I have tested it. As @Carreau brought up we may have to improve the error handling later, but let's merge this and see how it goes.

PR #7405, opened 2015-01-09
Make sure tests aren't included twice as js and python

jhamrick:
I'm using iptest for nbgrader javascript tests, which works by simply appending `js/` to the beginning of the path of my javascript tests. Currently, my javascript tests get detected as both javascript and python, because they are not in the list of tests returned by `all_js_groups()`. This PR fixes that so that python tests are only those tests which are not javascript tests.

As far as I'm aware, there aren't any javascript tests that don't start with the `js/` prefix, so this shouldn't actually change how IPython tests are run.


takluyver:
Looks good to me, thanks @jhamrick

PR #7408, opened 2015-01-09
Fix ultratb when there are non-ascii filenames present in a traceback

ccordoba12:
This solves the simple problem of running a file with non-ascii chars in
its name that contains an error of any kind in it.

Example: A file called `á.py` with this content

```
foo('a')
```

run it with `%run á.py` and see a horrible error.

This was motivated by this SO post:

http://stackoverflow.com/questions/27834477/unicode-decode-error-report-in-spyder

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -248,7 +248,7 @@ def fix_frame_records_filenames(records):
         # Look inside the frame's globals dictionary for __file__,
         # which should be better. However, keep Cython filenames since
         # we prefer the source filenames over the compiled .so file.
-        filename = py3compat.cast_unicode(filename, "utf-8")
+        filename = py3compat.cast_unicode_py2(filename, "utf-8")
         if not filename.endswith(('.pyx', '.pxd', '.pxi')):
             better_fn = frame.f_globals.get('__file__', None)
             if isinstance(better_fn, str):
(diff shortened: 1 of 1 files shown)


IPython/core/ultratb.py
```diff
@@ -248,6 +248,7 @@ def fix_frame_records_filenames(records):
         # Look inside the frame's globals dictionary for __file__,
         # which should be better. However, keep Cython filenames since
         # we prefer the source filenames over the compiled .so file.
+        filename = py3compat.cast_unicode(filename, "utf-8")
```
takluyver: I suspect that all of these can probably be `cast_unicode_py2` - the problem doesn't occur on Python 3, does it?

ccordoba12: I don't know. I used `cast_unicode` because it's the only one used in `ultratb.py`. I'm going to check with PY3.

ccordoba12: Yep, you're right. This error doesn't happen in Python 3. Should I change to `cast_unicode_py2` then?

takluyver: Please. That way, when we eventually drop support for Python 2, it will be clear that the call can be removed.

ccordoba12: Do I need to do it just for the calls I introduced for all calls to `cast_unicode` in this file

takluyver: Just the ones you introduced for now. Thanks.

Carreau:
> Replace cast_unicode calls with cast_unicode_py2 ones

Thanks, merging.

PR #7416, opened 2015-01-11
Prevent jump when clicking on action in menu.

### LINKED ISSUE #7367: Clicking on menu actions in the notebook makes it jump to the beginning
Fox example, if you click on `Insert > Insert Cell Above` or `View > Toggle Header`, the notebook jumps immediately to the beginning. It seems the same happens with all actions (although I haven't tried them all).

I'm using 3.0.0-dev, revision 4b40dd4

**Edit**: I tried it in Firefox 33 and Chrome 34

Carreau:
To do so prevent even default action;

Closes #7367

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -60,7 +60,10 @@ define([
             ks = this.kernelspecs[keys[i]];
             ks_submenu_entry = $("<li>").attr("id", "kernel-submenu-"+ks.name).append($('<a>')
                 .attr('href', '#')
-                .click($.proxy(this.set_kernel, this, ks.name))
+                .click(function(){
+                        $.proxy(this.set_kernel, this, ks.name)
+                        event.preventDefault();
+                    })
                 .text(ks.spec.display_name));
             change_kernel_submenu.append(ks_submenu_entry);
         }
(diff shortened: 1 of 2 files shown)


mathieu1:
Don't forget the actions in `kernelselector.js` (new notebook and change kernel)

Carreau:
> Don't forget the actions in kernelselector.js (new notebook and change kernel)

Indeed, commit on its way.

Carreau:
And rebased.

IPython/html/static/notebook/js/kernelselector.js
```diff
@@ -60,7 +60,10 @@ define([
             ks = this.kernelspecs[keys[i]];
             ks_submenu_entry = $("<li>").attr("id", "kernel-submenu-"+ks.name).append($('<a>')
                 .attr('href', '#')
-                .click($.proxy(this.set_kernel, this, ks.name))
+                .click(function(){
+                        $.proxy(this.set_kernel, this, ks.name)
+                        event.preventDefault();
```
minrk: shouldn't this be `return false`?

Carreau: Nop, otherwise menu does not get dismissed.

minrk: gotcha, that's the other half of `return false`. AOK, then.

PR #7420, opened 2015-01-11
Fix "TypeError: data is undefined" when loading a notebook in Firefox

### LINKED ISSUE #7328: Error loading notebook
Using Firefox I get an error with the following message when loading some notebooks:
`TypeError: ks is undefined`

The javascript log output is:
```
TypeError: ks is undefined
var new_mode_url = 'kernelspecs/'+ks.name+'/kernel';
``` 
in Lines 2093 and also 2100 of `notebook.js`.

This is with current master using Anaconda and Firefox (tested with Linux and Windows).
Example notebook: https://gist.github.com/juhasch/67e38a443ce220d7247c

juhasch:
Fixes https://github.com/ipython/ipython/issues/7328 for me.

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -126,6 +126,7 @@ define([
             return;
         }
         var ks = this.kernelspecs[kernel_name];
+        if (ks === undefined) return;        
         
         try {
             this.notebook.start_session(kernel_name);


minrk:
Maybe this is a Firefox async issue. I'm pretty sure returning early from that function without taking any action is not the right thing to do.

juhasch:
I am sure you are correct on both accounts. On Chrome this is no issue.

I don't believe this to be the correct solution, it merely allows me to use Firefox again.
Strangely, just returning from the call works.

minrk:
closing in favor of #7432

Issue #6744, opened 2014-10-20
error in examples / Notebook / Trapezoid Rule.ipynb


nholtz:
The Tapezoidal Rule notebook example has an error that invalidates its math.  The 
URL for the example is:

http://nbviewer.ipython.org/github/ipython/ipython/blob/2.x/examples/Notebook/Trapezoid%20Rule.ipynb

and is pointed to by this page:
http://nbviewer.ipython.org/github/ipython/ipython/blob/2.x/examples/Notebook/Index.ipynb

The 6th cell has this:

a, b = 1, 9
xint = x[np.logical_and(x>=a, x<=b)][::30]
yint = y[np.logical_and(x>=a, x<=b)][::30]

but that doesn't properly integrate over the range 1,9.  In fact, the example as given
only integrates over 1.005,8.543 (because of the selection of every 30th point).
Simpler and better to replace cell with:

a, b = 1, 9
xint = np.linspace(a,b,6)
yint = f(xint)

but then, unfortunately, it won't illustrate what the original author was probably 
trying to illustrate.

andram:
In this example a cubic function f(x) is integrated from x=1 to x=9, however by coincidence it happens that the derivatives on both boundaries are the same. In this case, the [Maclaurin formula](https://en.wikipedia.org/wiki/Euler%E2%80%93Maclaurin_formula) tells us that f(x) can be exactly integrated through the trapezoidal rule. Therefore this example is not suitable to demonstrate the difference between the two. The claimed difference is not down to the trapezoidal rule, but due to the error the points are chosen in the interval [1,9]. 

This is fixed with pull request #7426 by integrating from x=1 to x=8. The implementation of the trapezoid rule is also added for illustration.

minrk:
closed by #7426

PR #7426, opened 2015-01-12
use better integration boundaries

andram:
Because this example intends to illustrate the difference between
trapezoid rule and high-order integration, choose an appropriate
integration interval where such a difference exists.

Also implement the trapezoid rule directly for pedagogical purposes,
instead of relying on a canned version from scipy.


andram:
This fixes issue #6744

minrk:
thanks!

PR #7432, opened 2015-01-12
use promises to wait for kernelspecs on notebook load

### LINKED ISSUE #7328: Error loading notebook

minrk:
should avoid race condition that causes attempt to load kernelspec prior to loading the kernelspec list

should close #7328

@juhasch, @twiecki can you check if this actually fixes the problem?

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -126,28 +126,27 @@ define([
         }
     };
 
-    KernelSelector.prototype.change_kernel = function (kernel_name) {
-        /**
-         * TODO, have a methods to set kernel spec directly ?
-         **/
+    KernelSelector.prototype.set_kernel = function (kernel_name) {
+        /** set the kernel by name, ensuring kernelspecs have been loaded, first */
+        var that = this;
+        return this.loaded.then(function () {
+            that._set_kernel(kernel_name);
+        });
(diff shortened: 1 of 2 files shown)


Carreau:
+1 once travis happy.

juhasch:
Sorry, doesn't look like it helps.

juhasch:
It's this line no. 1773 in notebook.js that makes the trouble:
`this.kernel_selector.change_kernel(this.metadata.kernelspec.name);`

minrk:
This now cleans up the kernelselector logic more substantially. I'll write up a description of the various event sequences, so we can be sure everything's correct.

juhasch:
Hey, what did you do with that nice error message ?
It's gone. Thanks guys.

minrk:
@juhasch thanks for checking

takluyver:
This looks good to my eyes. Merging.

PR #7436, opened 2015-01-13
fix access to png width and height in nbconvert html basic template

jfeist:
When converting a notebook to html with `ipython nbconvert --to html test.ipynb`, I got an error, with the relevant part of the long error message being

```
  File "/Users/feist/Downloads/self_compile/ipython/IPython/nbconvert/exporters/../templates/html/basic.tpl", line 134, in <module>
    width={{output.metadata['png']['width']}}
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/jinja2/environment.py", line 378, in getitem
    return obj[argument]
jinja2.exceptions.UndefinedError: 'IPython.nbformat.notebooknode.NotebookNode object' has no attribute 'png'
```

As far as I can tell, `basic.tpl` should instead use `output.metadata['image/png']`. With that change (which this pull request provides), everything works.


takluyver:
I think this looks good. Is it easy to add a test that would have caught the problem you fixed?

jfeist:
Good question. At first I though it happened with any png (and was surprised it hadn't been caught yet...). I checked again now and realized it only occurs if the width and height are explicitly saved as metadata, which at least seems to happen for matplotlib plots if you are using the `retina` figure format.
So to test it, a notebook with a single cell is enough:
```python
%matplotlib inline
%config InlineBackend.figure_formats = set(['retina'])
import matplotlib.pyplot as plt
plt.plot([0,1],[1,0])
```
If you run and save this notebook, running `ipython nbconvert --to html` on it gives the error.

takluyver:
Great - can you add an example of that as an automated test in this file:

https://github.com/ipython/ipython/blob/master/IPython/nbconvert/exporters/tests/test_html.py

jfeist:
Thanks for the hint, I've added a test now. Running `nosetest` in the directory gives the error if I don't have the patch, and does not give it with the patch, so it seems to be ok.

takluyver:
Great, thank-you. I'll merge once Travis has run the tests successfully.

Carreau:
Thanks, as per @takluyver comment, merging.

PR #7440, opened 2015-01-13
Fix kernel change breaking "new notebook" list

mathieu1:
The ordering of the `File > New Notebook` list (introduced in #7401) becomes broken if the kernel is changed. Here's how it looks like after a few kernel changes : 
fix-new-notebook
This PR fixes that behavior.


Carreau:
Oops, thanks.

PR #7441, opened 2015-01-13
fix indentation of raw cells

### LINKED ISSUE #3659: Running a cell will open up output totally
I don't know if it's considered as an issue for others, but I find it annoying that each time I rerun a cell, the output gets opened fully even if it was in "scrollable window" mode (or closed mode).

I am currently working with long (text) outputs, and I often "run all" => this has the effect of totally scrambling my vision of the page each time.

Yet, if this default behavior is considered good, probably the best would be an option to keep outputs in scrollable windowed mode if we want to.

### LINKED ISSUE #6262: Raw cell indentation problems with nbconvert
I use notebooks and nbconvert as the basis for the [seaborn](http://stanford.edu/~mwaskom/software/seaborn/) sphinx website. My basic workflow is to write notebooks with a mixture of python code cells and sphinx markup in raw text cells. See [here](https://github.com/mwaskom/seaborn/blob/master/doc/tutorial/tools/nb_to_doc.py) for the actual build script. This worked fine with 1.0, but I just built my docs (I think for the first time with 2.0; I use [runipy](https://github.com/paulgb/runipy) in the build chain and there was a lag in it supporting 2.0).

I have encountered a problem related to how nbconvert is handling the raw text. The first line of any raw text cell that follows code or image output ends up heavily indented, e.g.:

```
.. code:: python

    sns.lmplot("total_bill", "tip", tips, x_bins=8);


.. image:: quantitative_linear_models_files/quantitative_linear_models_21_0.png


                Faceted linear model plots
^^^^^^^^^^^^^^^^^^^^^^^^^^
                
                The :func:`lmplot` function is built on top of a :class:`FacetGrid`. That means it's easy to visualize how this relationship changes in different subsets of your dataset
```

As you would expect, this screws up the sphinx markup. I can fix it by manually entering a blank line at the top of all my raw text cells, but that's annoying.

Does this reflect something that has changed in nbconvert for a reason?

A related issue (though one that existed in 1.0) is that nbconvert-ing to .rst doesn't add new lines after code cells, which causes a bunch of sphinx warnings. It might be worth addressing that, since it's related to the issue here (but seemingly separate).

minrk:
a matter of missing '{%-'

closes #6262


minrk:
fixed related issue reference

PR #7445, opened 2015-01-14
left-aligning text

bollwyvl:
Does as #7239 asks.

Here is an example, maybe not the most demonstrative... perhaps someone can suggest a better notebook?

Before:
screenshot from 2015-01-13 18 57 42

After:
screenshot from 2015-01-13 18 57 00

Here is some light reading:
- [Nature](http://www.nature.com/srep/authors/submit.html): don't justify... manuscripts. 
- [Readability Do's and Don'ts](http://wmich.edu/writing/readability): meh... maybe use if it's long?
- [SO](http://graphicdesign.stackexchange.com/questions/16137/readability-and-appeal-of-justified-text): use justify. maybe.
- [Readability: Justified vs. Ragged](http://www.textproof.com/chronicle/Readability__Justified_vs__Ragged.html): justify if the lines are short
  - references [Readability of Computer-Generated Fill-Justified Text. Human Factors: The Journal of the Human Factors and Ergonomics Society April 1986 28: 159-163,](http://hfs.sagepub.com/content/28/2/159.full.pdf): behind paywall, don't have library proxy set up :)
  - [Tufte](http://www.edwardtufte.com/bboard/q-and-a-fetch-msg?msg_id=0001Au): don't justify... _and use serifs_

Line length is mentioned several times. While I have no idea what the metric should be, here's one from the michigan link above:

>  line length should be between 1.5 and two times the lowercase alphabet in the type and point size being used

For the normal text (14px) that would put the line length between 260 and 350. Well, that's not going to happen! However, it does raise the question of at some point providing a "responsive readability" mode (think flipboard, etc) that really went after the readability metrics, especially for devices, which are becoming such a popular way to read content.

### DIFF
--- a/IPython/html/static/notebook/less/renderedhtml.less
+++ b/IPython/html/static/notebook/less/renderedhtml.less
@@ -73,7 +73,7 @@
     th {font-weight: bold;}
     * + table {margin-top: 1em;}
 
-    p {text-align: justify;}
+    p {text-align: left;}
     * + p {margin-top: 1em;}
 
     img {
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -1100,7 +1100,7 @@ div.output_unrecognized a:hover {
   margin-top: 1em;
 }
 .rendered_html p {
-  text-align: justify;
+  text-align: left;
 }
 .rendered_html * + p {
   margin-top: 1em;
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -9548,7 +9548,7 @@ div.output_unrecognized a:hover {
   margin-top: 1em;
 }
 .rendered_html p {
-  text-align: justify;
+  text-align: left;
 }
 .rendered_html * + p {
   margin-top: 1em;


Carreau:
Thanks for doing that. I'm biased as I hate non-justified text, so I can know my answer even without looking at the pictures. 

So code wise +1, 
and I'll avoid voting on this one.

damianavila:
My impression... non-justified text makes more evident the unbalance between the left and right sides of the document (because of the prompt number there is an unbalance distribution).

minrk:
Do we want to put the justify switch on a media query, so it's justified when lines are short? I don't have a strong opinion either way, so I'm fine with this as-is.

ellisonbg:
Thanks for the PR.

For wide screens, the line lengths in the notebook are waaaaaaay longer than the usual typographical recommandations. But as you state, that is probably something we have to live with. Although @minrk has custom CSS that adds a symmetic cell margin to the R side (can you post a screen shot Min?) so he has shorter lines. That might be something to think about.

But even with that, we will still have long lines in the notebook on modern wide screens...

The traditional "justification" for left justify is this:

1. For longer lines of text, fill justify makes it difficult to detect where your eyes are in reading the paragraph. The left justify provides just enough irregularity to guide the eye as it moves down the lines vertically.
2. Browsers still aren't really good at implementing fill justify from a typographical perspective (word spacing, hyphenation, kerning).

The media query solves 1 but not 2. I am probably still +1 on this change always.

Here is a great link that goes into the algorithmic details of justification in the browser and shows a nice example of how the implementation in browsers doesn't cut it:

http://www.bramstein.com/projects/typeset/

damianavila:
>Although @minrk has custom CSS that adds a symmetic cell margin to the R side (can you post a screen shot Min?) so he has shorter lines. That might be something to think about.

I am completely +10 to this idea... not having symmetrical sides cause a big unbalance between the left and right side of the notebook...

minrk:
My current style on a wide screen:

screen shot 2015-01-15 at 15 06 52

ellisonbg:
@minrk thanks for the screen shot. Do you have other css that makes your page area narrower as well?

minrk:
One thing I don't like about the right margin and narrower max-width is that the maximum columns for an input cell is only 66 characters. I like it for prose, but it's not good enough for code. I wouldn't propose it for the default style as-is, and I don't think we should be doing any more significant style changes for 3.0.

ellisonbg:
66 chars is definitely not enough for code. And this is really to crux of
it. Code wants to be wide, text narrow.

On Thu, Jan 15, 2015 at 3:13 PM, Min RK <notifications@github.com> wrote:

> One thing I don't like about the right margin and narrower max-width is
> that the maximum columns for an input cell is only 66 characters. I like it
> for prose, but it's not good enough for code. I wouldn't propose it for the
> default style as-is, and I don't think we should be doing any more
> significant style changes for 3.0.
>
> —
> <https://github.com/ipython/ipython/pull/7445#issuecomment-70180909>.
>



-- 
Brian E. Granger

takluyver:
Do we want to do more review of this before merge, or shall we land it and see how we get on with it in master?

ellisonbg:
Let's put it on the agenda for tomorrows meeting and make a decision there.

takluyver:
From the dev meeting: let's merge this and see what it's like in use.

Carreau:
Ok, so my opinion after a bit of use, I don't like. 
I prefer justify,and would gladely add a right margin to be symmetrical with prompt area.
I even thinking this right area could be used for (some) controls, like futurs handle to show comments on cell and a like.

damianavila:
>I prefer justify,and would gladely add a right margin to be symmetrical with prompt area.

+ 10

>I even thinking this right area could be used for (some) controls, like futurs handle to show comments on cell and a like.

This is something I do in my custom setup :wink:

PR #7446, opened 2015-01-14
Set name/date for served notebooks

### LINKED ISSUE #7360: Notebook menu File -> Download as -> PDF title should be notebook title
If you select File -> Download as -> PDF then every notebook has the title of "Notebook":

screenshot from 2014-12-31 19 37 49

This is very confusing when making handouts for students... they all have the title "Notebook"! The PDF should get the title of the notebook, whatever it is.

bollwyvl:
This fixes #7360 by explicitly providing `resources` when calling `from_notebook_node` in the nbconvert handlers. It sets the name and modified date in `metadata`, but none of the other stuff which didn't seem relevant in this case:

``` python
{
"unique_key": "Stuff",
"output_files_dir": "Stuff_files",
"metadata": defaultdict(None, {"name": "Stuff", "modified_date": "January 13, 2015"}),
"profile_dir": "/home/weg/.ipython/profile_default"
} 
```

This is normally populated by `from_filename`, but calling that didn't seem appropriate for a `model` coming off the contents manager.

Also uncovered a mis-construction of `ResourcesDict` (a subclass of `defaultdict`), using the pattern found immediately above... was getting the ever-helpful `first argument must be callable` error, as it should be the default value for missing. Another option would be to change the constructor to be like `dict`, but this seemed less surprising.

### DIFF
--- a/IPython/html/nbconvert/handlers.py
+++ b/IPython/html/nbconvert/handlers.py
@@ -6,7 +6,6 @@
 import io
 import os
 import zipfile
-import collections
 
 from tornado import web
 
--- a/IPython/html/nbconvert/handlers.py
+++ b/IPython/html/nbconvert/handlers.py
@@ -6,6 +6,7 @@
 import io
 import os
 import zipfile
+import collections
 
 from tornado import web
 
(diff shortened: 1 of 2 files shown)


Carreau:
Seem sensible to me.

damianavila:
:+1:

PR #7453, opened 2015-01-14
Check for latex commands to improve error messages when commands are missing

takluyver:
Replaces #6884.

### DIFF
--- a/IPython/nbconvert/exporters/pdf.py
+++ b/IPython/nbconvert/exporters/pdf.py
@@ -6,8 +6,8 @@
 import subprocess
 import os
 import sys
-import shutil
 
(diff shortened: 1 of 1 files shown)


damianavila:
I am trying to test this one, but component is pointing to https://github.com/ipython/ipython-components/commit/f8f0c3b8958b1f91a888b5f15d2b17404a476016, which gives me an error about mistune which does not make any sense because it is already updated...

takluyver:
Not sure why components would affect mistune anyway.

damianavila:
>Not sure why components would affect mistune anyway.

Don't know neither... but in a copied branch, I rebase on master to catch up the correct component and it is working as expected, so :+1:

PR #7455, opened 2015-01-14
fix new-notebook and change-kernel menus

minrk:
recent PR made all of these no-ops

also include missing dialog module


Carreau:
will conflict with #7458

PR #7458, opened 2015-01-14
remove superfluous event.preventDefault

minrk:
from #7416

only one of the additions from that PR was necessary to address the issue

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -65,7 +65,6 @@ define([
                         .attr('href', '#')
                         .click( function () {
                             that.set_kernel(ks.name);
-                            event.preventDefault();
                         })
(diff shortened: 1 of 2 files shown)


Carreau:
Hum, indeed, I should have seen that.

Carreau:
Need rebase. Sorry.

minrk:
No problem. I only found out that some of these were unnecessary by accident. I was a bit surprised, and don't fully understand, but I'm happy about it.

minrk:
Rebased. If you can double-check to confirm that the issue is still fixed after this change, I would appreciate it. I'm never confident with event changes like this.

Carreau:
Can you be confident with js ?

Carreau:
I tried most of edit, view, insert, quickhelp, file download as... all seem fine. 

Merging.

PR #7464, opened 2015-01-14
discard invalid stream output

### LINKED ISSUE #6852: OutputArea.js text can be null
Outputarea.js:467 Uncaught TypeError: Cannot read property 'replace' of undefined

`if(!text.replace)` but `text` can be `null` in `json`. Either test for it, or be sure on the wire that json.text is not null but empty string.

minrk:
If a stream output is missing text, log an error and discard the broken message so it doesn't get into the notebook.

closes #6852


Carreau:
+1

PR #7465, opened 2015-01-14
update name, docstring for warn_format_error

### LINKED ISSUE #7072: warn_format_error no longer makes it clear that there was a problem formatting the output
The recent changes to `warn_format_error` seem to have made it useless: one can no longer see the difference between a failed command and the failure to print the output.

Currently, if formatting fails, `warn_format_error` shows the exception with traceback which looks indistinguishable to a normal traceback when a command failed.

At the very least, the name `warn_format_error` no longer fits what it does.

minrk:
since it shows regular tracebacks now, rather than special warnings

closes #7072

### DIFF
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -218,8 +218,8 @@ class FormatterWarning(UserWarning):
     """Warning class for errors in formatters"""
 
 @decorator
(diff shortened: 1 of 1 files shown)


Carreau:
+1.

PR #7466, opened 2015-01-14
allow interact(instancemethod)

### LINKED ISSUE #6278: interact doesn't work with instance methods
This code:

```python
from IPython.html.widgets import interact

class Foo(object):
    def show(self, x):
        print x
        
f = Foo()

interact(f.show, x=(1,10))
```

produces this exception:

```
<ipython-input-58-b03b8685dfc0> in <module>()
      7 f = Foo()
      8 
----> 9 interact(f.show, x=(1,10))

/home/fperez/usr/lib/python2.7/site-packages/IPython/html/widgets/interaction.pyc in interact(__interact_f, **kwargs)
    235         f = __interact_f
    236         w = interactive(f, **kwargs)
--> 237         f.widget = w
    238         display(w)
    239         return f

AttributeError: 'instancemethod' object has no attribute 'widget'
```

minrk:
objects that can't have `.widget` attribute assigned are wrapped in a lambda

closes #6278

### DIFF
--- a/IPython/html/widgets/interaction.py
+++ b/IPython/html/widgets/interaction.py
@@ -1,16 +1,7 @@
 """Interact with functions using widgets."""
 
-#-----------------------------------------------------------------------------
-# Copyright (c) 2013, the IPython Development Team.
-#
+# Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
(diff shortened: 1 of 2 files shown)


Carreau:
+1

PR #7469, opened 2015-01-15
preserve umask in atomic_writing

### LINKED ISSUE #7317: atomic write umask
Reported on gitter.im/ipython/ipython by @bigzachattack 

Sadly no one was on the chat at that time, too busy with a big bearded men 
dressed in red trying to smuggle things in our houses through the cheminee. 

```
I noticed today working in master, that any new notebook I create or copy has permissions
0o600 and ignores my umask. In IPython.utils.io.atomic_writing() uses mkstemp() to create
a temporary file for the atomic write. According to the python docs, mkstemp creates
files as 0o600.

After the write succeeds to the tmp file, _copy_metadata is called to copy the metadata
from the original file to destination file. It will throw an exception if there is no
source file. Thus when the notebook is copied into the notebook dir, it has
permissions 0o600.

Is this desired behavior, temporary, or a bug? I work in an environment where are default
permissions are 0o660 to allow for users to easily share information, so defaulting new
notebooks to 0o600 seriously inhibits this ability.
```

minrk:
by using mkdtemp instead of mkstemp

closes #7317

### DIFF
--- a/IPython/utils/io.py
+++ b/IPython/utils/io.py
@@ -267,17 +267,18 @@ def atomic_writing(path, text=True, encoding='utf-8', **kwargs):
         path = os.path.join(os.path.dirname(path), os.readlink(path))
 
     dirname, basename = os.path.split(path)
-    handle, tmp_path = tempfile.mkstemp(prefix=basename, dir=dirname)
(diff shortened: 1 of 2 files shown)


Carreau:
:shipit:

PR #7473, opened 2015-01-15
Smooth notebook scrolling on pager closing.

Carreau:
avoid brutal jump of notebook on closing pager, when notebook is scroll
to max

### DIFF
--- a/IPython/html/static/notebook/less/notebook.less
+++ b/IPython/html/static/notebook/less/notebook.less
@@ -70,6 +70,7 @@ p {
 
 .end_space {
     min-height: 100px;
+    transition: height .2s ease;
 }
 
 .notebook_app #header {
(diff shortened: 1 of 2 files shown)


minrk:
@Carreau I'm not quite clear on what this fixes. I can't seem to make it misbehave. What does 'scroll to max' mean?

Carreau:
Oh, it does not really "fix" anything. 

- Make many cells.
- `numpy?` (open pager)
- focus last cell
- scroll up as much as you can. 
- esc to close pager.

Notebook use to 'jump' back in place when end-space with go suddently from XXX->100px.
now it just slowly go from XX to 100 px

takluyver:
in current master, when I try to follow your instructions, it doesn't scroll the focused cell back into view at all: the pager closes without causing a scroll.

Carreau:
The framerate does not help, but here it is:
Master :
master

PR:

pr

PR #7476, opened 2015-01-15
Do not leave toolbar element focused.

### LINKED ISSUE #7475: Focus stays on toolbar buttons after action
I ran into the following:
Step 1 : delete a cell (click the cut button)
Step 2 : run a few cells with shift + enter a bunch of times
Ooops, many cells just got cut!

Much more harmless, but might be useful for debugging: the copy button also keep focus, and so do the cell-up (cell-down) buttons when activated on the top (bottom) cell.

By the way, mistakenly cutting more than one cell happens regardless of this issue (human error). Am I correct in thinking that in such a situation, the only thing that can be done currently is to shutdown the notebook before it auto-saves (or not rely on auto-saves but checkpoints instead)?

Carreau:
Can lead to enter re-trigerring them
(ok, can be a feature sometimes I guess)

closes #7475

### DIFF
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -102,7 +102,10 @@ define(function(require){
             icon: 'fa-cut',
             help_index : 'ee',
             handler : function (env) {
+                var index = env.notebook.get_selected_index();
                 env.notebook.cut_cell();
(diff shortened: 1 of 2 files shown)


Carreau:
Cut, Paste and dropdown give focus back to the selected cell now.

PR #7489, opened 2015-01-16
Scroll overflow

parente:
I included @joelkim's fix to the less for issue #5967 and spent time trying to find a fix for the scrollbars overflowing the `output_area` on collapse/expand of the cell. I was only able to reproduce the bug in webkit browsers (Safari/Chrome) but not Firefox. From googling and experimentation, it appears webkit has some known problems in this area. Hacks abound on stackoverflow (http://stackoverflow.com/questions/3485365/how-can-i-force-webkit-to-redraw-repaint-to-propagate-style-changes) and the only rendition I got working is the one you see in the PR (insert an extra node, wait, remove it).

It's gross.

I'm not recommending it as the final fix, just something that seems to work in case all else fails.

### DIFF
--- a/IPython/html/static/notebook/less/outputarea.less
+++ b/IPython/html/static/notebook/less/outputarea.less
@@ -92,7 +92,7 @@ div.output_area pre {
 /* This class is for the output subarea inside the output_area and after
    the prompt div. */
 div.output_subarea {
-    padding: @code_padding @code_padding 0.0em @code_padding;
+    padding: @code_padding;
     .box-flex1();
 }
 
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -867,7 +867,7 @@ div.output_area pre {
 /* This class is for the output subarea inside the output_area and after
    the prompt div. */
 div.output_subarea {
-  padding: 0.4em 0.4em 0em 0.4em;
+  padding: 0.4em;
   /* Old browsers */
   -webkit-box-flex: 1;
   -moz-box-flex: 1;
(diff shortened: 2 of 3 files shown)


IPython/html/static/notebook/less/outputarea.less
```diff
@@ -92,7 +92,7 @@ div.output_area pre {
 /* This class is for the output subarea inside the output_area and after
    the prompt div. */
 div.output_subarea {
-    padding: @code_padding @code_padding 0.0em @code_padding;
+    padding: @code_padding @code_padding @code_padding @code_padding;
```
minrk: When they are all the same, this can be collapsed to a single `@code_padding`

parente: Indeed. I amended the commit.

minrk:
This doubles the space between outputs, which I think it undesirable. We may need a different fix. Can you try moving the 0 padding from the bottom to the top, and adding a `:first` selector to add the padding to the top only on the first element? That should keep the relevant change and the space between outputs the same.

minrk:
The timeout bit doesn't appear to do anything for me. I'd be inclined to take the increased padding as a mitigation, leave out the timeout, and try to report the browser bug upstream. What do you think, @parente?

parente:
@minrk Regarding the timeout, do you mean it doesn't fix the scrollbar overlap problem after hiding and then re-showing the output? It does for me, but only with a significant delay so that the extra div remains in the DOM for around 1s or so. Anything shorter, and it didn't fix the problem for me.

I'm fine with merging the padding fix alone to make sure the cells render properly most of the time on notebook load. The add/remove div hack makes my skin crawl.

Want me to submit anew with the CSS changes only?

minrk:
> Want me to submit anew with the CSS changes only?

You can update this PR in-pace by pushing to the same branch. It does need a rebase in order to merge properly, though. If you are more comfortable doing that as a new PR based on master, that's fine, too.

parente:
Dropped the hack from the branch. Pull request should be OK to merge when travis goes green.

PR #7491, opened 2015-01-17
wait for terminado process to terminate

minrk:
this needs to wait for a terminado release after takluyver/terminado#3 is merged,
at which point we need to bump the terminado version dependency.


takluyver:
I've pushed a new release of terminado, 0.3.3, with your PR.

minrk:
Updated to require terminado >= 0.3.3. I also promoted the "no terminals" message to warn, and demoted it to debug on Windows. Does it make sense to even try to initialize terminals on Windows?

Carreau:
+1

PR #7492, opened 2015-01-17
handle missing stream name in nbformat conversion

### LINKED ISSUE #7490: nbformat v2 notebook with missing 'stream' key in stream outputs
@stefanv just sent me an nbformat 2 notebook which was missing the 'stream' keys in stream outputs, and failed to open in current master because of this.

@minrk , can you remember if that was ever a valid option in the notebook format? If so, we should support it. If not, we can write it off as having been saved by some broken development version of IPython.

minrk:
It was sometimes possible for this key to be absent. Even though that's never been officially valid, it happened, so might as well handle it.

closes #7490


takluyver:
Thanks, that works.

Issue #7486, opened 2015-01-16
Protect modal body from untrusted HTML


Carreau:
`IPython.dialog.modal({'body':"<script>alert('bar')</script>"})` does what you expect. 

Which is annoying as `load_notebook_error` get the message from the rejected promise, whose `error.message` might be remote from some weird unexpected things. 

proposal: 

in our dialog implementation, use a `unsafe_body` key when we know the body come from a trusted source and contain html. Use it if necessary,  and sanitize/escape otherwise.

minrk:
I agree that we should escape html from untrusted sources, but I disagree that we should use an 'unsafe_body' key, since we would always set it to true, which defeats its purpose. We should just be wary of html from untrusted sources. The load_notebook_error message does come from a trusted source, so it's not an example of a problem.

Carreau:
> The load_notebook_error message does come from a trusted source, so it's not an example of a problem

On jupyterdrive I am not completely certain it comes from a trusted source. (Agreed this is paranoia)

I think I miss expressed myself with unsafe_body:

Imho :
`modal({ body: '<script>alert('foo')</script>'})` should show `<script>alert('foo')</script>` and `modal({ unsafe_body: '<script>alert('foo')</script>'})` should execute the js, or whatever name we give it. 

unsafe usage  of modal should be explicit, safety should be default.

minrk:
Every single case we have right now would use unsafe_body, so I'm not sure how that helps.

jdfreder:
I think it's completely fine for this helper function to accept HTML.  I think wherever you're uncertain if the content is safe or not (you mention jupyterdrive) you should sanitize it before passing it into the dialog method.

minrk:
Absolutely. There's no way dialog should stop accepting HTML. I've retitled the issue, since I don't think @Carreau is proposing that.

Carreau:
Sorry for confusing title. 
I ment that if the function receive text it should not be considered trusted and turned into html.

takluyver:
This was addressed by #7493.

PR #7493, opened 2015-01-17
Sanitize dialog text-body.

Carreau:
Mitigate 7486, could probably be improved, but
treat text as pure text, but assume that when received a dom
element it actually come from a secure source.

Sanitizing all we get, including Dom Element break json-metadata editor
and a few others.

### DIFF
--- a/IPython/html/static/base/js/dialog.js
+++ b/IPython/html/static/base/js/dialog.js
@@ -8,6 +8,35 @@ define(function(require) {
     var IPython = require('base/js/namespace');
     var $ = require('jquery');
     
+    /**
+     * A wrapper around bootstrap modal for easier use
+     * Pass it an option dictionary with the following properties:
+     *
+     *    - body : <string> or <DOM node>, main content of the dialog
+     *            if pass a <string> it will be wrapped in a p tag and
+     *            html element escaped, unless you specify sanitize=false
+     *            option.
+     *    - title : Dialog title, default to empty string.
+     *    - buttons : dict of btn_options who keys are button label.
+     *            see btn_options below for description
+     *    - open : callback to trigger on dialog open.
+     *    - destroy:
+     *    - notebook : notebook instance
+     *    - keyboard_manager: keyboard manager instance.
+     *
+     *  Unlike bootstrap modals, the backdrop options is set by default 
+     *  to 'static'.
+     *
+     *  The rest of the options are passed as is to bootstrap modals. 
+     *
+     *  btn_options: dict with the following property:
+     *  
+     *    - click : callback to trigger on click
+     *    - class : css classes to add to button.
+     *
+     *
+     *
+     **/
     var modal = function (options) {
 
         var modal = $("<div/>")
@@ -20,7 +49,7 @@ define(function(require) {
         var dialog_content = $("<div/>")
             .addClass("modal-content")
             .appendTo(dialog);
-        if(typeof(options.body) === 'string'){
+        if(typeof(options.body) === 'string' && options.sanitize !== false){
             options.body = $("<p/>").text(options.body)
         }
         dialog_content.append(
@@ -35,7 +64,7 @@ define(function(require) {
                 ).append(
                     $("<h4/>")
                         .addClass('modal-title')
-                        .text(options.title || "No Title")
+                        .text(options.title || "")
                 )
         ).append(
             $("<div/>").addClass("modal-body").append(
--- a/IPython/html/static/base/js/dialog.js
+++ b/IPython/html/static/base/js/dialog.js
@@ -1,12 +1,12 @@
 // Copyright (c) IPython Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-define([
-    'base/js/namespace',
-    'jquery',
-    'codemirror/lib/codemirror',
-], function(IPython, $, CodeMirror) {
+define(function(require) {
     "use strict";
+
+    var CodeMirror = require('codemirror/lib/codemirror');
+    var IPython = require('base/js/namespace');
+    var $ = require('jquery');
     
     var modal = function (options) {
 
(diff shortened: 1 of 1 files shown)


IPython/html/static/base/js/dialog.js
```diff
@@ -20,6 +21,9 @@ define([
         var dialog_content = $("<div/>")
             .addClass("modal-content")
             .appendTo(dialog);
+        if(typeof(options.body) === 'string'){
+            options.body = $("<p/>").text(security.sanitize_html(options.body))
```
minrk: There's no reason to use `.text` *and* `sanitize_html`, since `.text` escapes everything.

minrk: But I agree that using `.text` is a good idea, here.

Carreau: Don't see what you are talking about. ( :eyes: wistle while looking away )

IPython/html/static/base/js/dialog.js
```diff
@@ -32,7 +35,7 @@ define([
                 ).append(
                     $("<h4/>")
                         .addClass('modal-title')
-                        .text(options.title || "")
+                        .text(options.title || "No Title")
```
minrk: I think actually having no title is better than giving the dialog a title called "No Title"

IPython/html/static/base/js/dialog.js
```diff
@@ -20,6 +20,9 @@ define([
         var dialog_content = $("<div/>")
             .addClass("modal-content")
             .appendTo(dialog);
+        if(typeof(options.body) === 'string'){
+            options.body = $("<p/>").text(options.body)
```
jdfreder: Doesn't this break the rename function of the savewidget ?

jdfreder:
I'm not sure why we need to sanitize the dialog method.  i think this will break the logic we use to rename notebooks, where we pass an input control into the body to make the dialog accept text.  I'm planning on using this same technique elsewhere in IPython, and I think it's an appropriate use of the dialog helper function (I'd rather not have to re-implement this everywhere I need it just because the base only accepts text).  Instead, maybe it's better sanitize the content before it is passed into this method.

IPython/html/static/base/js/dialog.js
jdfreder: Oh I see, JQuery selectors pass over this.

minrk: Yup, it only sanitizes string dialogs. HTML dialogs aren't touched, and won't be.

minrk:
> maybe it's better sanitize the content before it is passed into this method.

yes, definitely. It is absolutely up to the builder of the dialog HTML to make sure it's safe.

Carreau:
if you pass it a dom object it will behave as expected. so you can alcually say `modal({body:$('p').append(text)})` it will work. The problem is that in some places, I had the dialog sow error/response message from server, which is a potential security flaw to inject js. It is better to sanitize by default everywhere and be explicit when you want safety off than the other way around.

jdfreder:
> The problem is that in some places, I had the dialog sow error/response message from server, which is a potential security flaw to inject js. It is better to sanitize by default everywhere and be explicit when you want safety off than the other way around.

I disagree with this.  This dialog method should be treated like a `.html` method.  The error messages should be sanitized before being passed to the dialog method.

Alternatively, you could create a convenience flag on this method, i.e. `sanitize: true` which triggered sanitation to happen.  I think that would be better than having sanitation triggered implicitly by argument type.

IPython/html/static/base/js/dialog.js
jdfreder: Agreed

Carreau:
that would be fine with me also, but sanitation should then be true by default, that was my first attempt at this PR. Though, the sanitizing library does work only on string, so this path would be more invasive.

minrk:
> error/response message from server, which is a potential security flaw to inject js

Since the vulnerability would have to be in the same code that serves the sanitization js, I'm not sure I see the issue here.

A disadvantage of sanitizing string bodies means that you have to be sure that the caller  *doesn't* sanitize the string body, otherwise you will get double-escapes. I still think this probably makes sense to do, and we can make a pass on the HTML inputs, to see if there are issues. For instance, the error messages from the server all pass through a single ajax error utility, which should probably call `escape_html(message)`, so that we know that `error.message` is safe text everywhere we use it.

Carreau:
I'm more concern by the fact that we expose this as an API that is used by extensions. 

I know extensions are not our problem and that security is their problem, but providing an unsecure default has historically always be a bad choice.

minrk:
I think it's easy to be surprised by the plain-text being able to execute HTML, which is why I still think this PR is a good idea. If they are passing HTML, I don't think it's an issue *at all*, because they could be putting that same HTML on the page in any way they want, and it would always be applied as-is. That's why I think this PR, plus a pass on our own HTML dialogs, closes #7486.

jdfreder:
@Carreau what do you think about

```js
if (options.sanitize===undefined || options.sanitize) {
    options.body = $("<p/>").text(options.body)
}
```

?

Carreau:
`options.sanitize===undefined || options.sanitize` -> `options.sanitize!==false` to avoid things like `sanitize='sdkjgb'` to pass through would suite me.

Carreau:
Updated to @jdfreder comment. 
`p.text()` will be used only if string is passed and sanitize is not set to `false`.

jdfreder:
Thanks! :)

takluyver:
From the dev meeting: we're happy with what this does at present (escape strings passed to dialog, don't sanitise pre-constructed DOM elements, because whatever constructs them is responsible for doing so safely).

takluyver:
Travis passed. Who wants to do final review and/or press the button?

PR #7496, opened 2015-01-18
copy timeit.Timer.timeit from CPython 3.4

### LINKED ISSUE #6172: %timeit magic does work on latest pypy anymore
The problem is the use of the undocumented semi-internal API `timer.inner`.
More info [here](https://bitbucket.org/pypy/pypy/issue/1820/ipython-timeit-magic-not-working-anymore)

minrk:
We were relying on self.inner for %timeit, which is an undocumented implementation detail of Python, and no longer shared by PyPy.

This copies the CPython implementation, so it is no longer an undocumented attribute.

It will likely not be optimal for some PyPy timings, but at least it works.

closes #6172

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1,25 +1,19 @@
 # -*- coding: utf-8 -*-
-"""Implementation of execution-related magic functions.
-"""
+"""Implementation of execution-related magic functions."""
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
 from __future__ import print_function
-#-----------------------------------------------------------------------------
-#  Copyright (c) 2012 The IPython Development Team.
-#
-#  Distributed under the terms of the Modified BSD License.
-#
-#  The full license is in the file COPYING.txt, distributed with this software.
-#-----------------------------------------------------------------------------
 
-#-----------------------------------------------------------------------------
-# Imports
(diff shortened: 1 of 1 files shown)


IPython/core/magics/execution.py
```diff
@@ -115,6 +108,34 @@ def visit_For(self, node):
...
+        a number of times, as a float measured in seconds.  The
+        argument is the number of times through the loop, defaulting
+        to one million.  The main statement, the setup statement and
+        the timer function to be used are passed to the constructor.
+        """
+        it = itertools.repeat(None, number)
+        gcold = gc.isenabled()
+        gc.disable()
+        try:
+            timing = self.inner(it, self.timer)
```
takluyver: In CPython timeit, `self.inner` is set by `Timer.__init__`. Don't we also need to copy the `__init__` method? I'm assuming that if Pypy doesn't use `.inner`, it won't set it either.

minrk: CPython sets it in `__init__`, we override it in `%timeit`, it's used in `.timeit`. This patch changes nothing at all for CPython. PyPy doesn't set it in `__init__`, we set it after, and this class ensures that it's used in `.timeit`.

takluyver: Ah, I missed that we were setting it as well. Then I think this looks fine.

PR #7497, opened 2015-01-19
don't let frontend readline split tokens in `ipython console`

### LINKED ISSUE #6442: Latex tab completions leave extra backslash in terminal console
In `ipython console`:

```
\alp<TAB> -> \\alpha
\alpha<TAB> -> \α
```

In both cases, there's a backslash left over that shouldn't be there. It works in the notebook and the Qt console.

minrk:
Can cause weird misalignment of completions when the tokenizing doesn't match with the kernel.

This tells readline to only work on whole lines, and expands completions to match.

I haven't kept track, but it seems like this might fix a few other issues people may have had with completions in `ipython console`.

closes #6442

### DIFF
--- a/IPython/terminal/console/completer.py
+++ b/IPython/terminal/console/completer.py
@@ -29,13 +29,18 @@ class ZMQCompleter(IPCompleter):
         self.shell = shell
         self.client =  client
         self.matches = []
-        
+        # don't do any splitting client-side,
+        # rely on the kernel for that
+        self.splitter.delims = '\r\n'
+        if self.readline:
+            self.readline.set_completer_delims('\r\n')
+    
     def complete_request(self, text):
         line = readline.get_line_buffer()
         cursor_pos = readline.get_endidx()
         
         # send completion request to kernel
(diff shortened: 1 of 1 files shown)


takluyver:
I confirm that this fixes #6442 for me.

IPython/terminal/console/completer.py
```diff
@@ -43,7 +48,8 @@ def complete_request(self, text):
         
         msg = self.client.shell_channel.get_msg(timeout=self.timeout)
         if msg['parent_header']['msg_id'] == msg_id:
-            return msg["content"]["matches"]
+            cursor_start = msg['content']['cursor_start']
+            return [ line[:cursor_start] + m for m in msg['content']['matches'] ]
```
blink1073: @minrk, you may want to use what I had started [here](https://github.com/ipython/ipython/pull/6714/files), to account for `cursor_end`.

takluyver:
Merging this, as it does fix real problems. Other potential improvements can go in separate PRs.

PR #7498, opened 2015-01-19
Traitlet support for functools.partial as callback

johanfforsberg:
When generating widgets programmatically, it can be convenient to be able to use functools.partial for callbacks, binding arguments to more generalised functions. However, traitlets do not support this due to inspect.getargspec not working on partials. This PR enables that functionality by adding some code taken from Sphinx - which presumably is battle-tested.

The functionality could of course also be imported from Sphinx but I guess it's desirable to keep Sphinx as a soft dependency.

### DIFF
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -280,6 +280,86 @@ def dlink(source, *targets):
     """Shorter helper function returning a directional_link object"""
     return directional_link(source, *targets)
 
+
+# Alternative getargspec implementations taken from Sphinx
+if sys.version_info >= (3, 0):
+
+    from functools import partial
+
+    def getargspec(func):
+        """Like inspect.getargspec but supports functools.partial as well."""
+        if inspect.ismethod(func):
+            func = func.__func__
+        if type(func) is partial:
+            orig_func = func.func
+            argspec = getargspec(orig_func)
+            args = list(argspec[0])
+            defaults = list(argspec[3] or ())
+            kwoargs = list(argspec[4])
+            kwodefs = dict(argspec[5] or {})
+            if func.args:
+                args = args[len(func.args):]
+            for arg in func.keywords or ():
+                try:
+                    i = args.index(arg) - len(args)
+                    del args[i]
+                    try:
+                        del defaults[i]
+                    except IndexError:
+                        pass
+                except ValueError:   # must be a kwonly arg
+                    i = kwoargs.index(arg)
+                    del kwoargs[i]
+                    del kwodefs[arg]
+            return inspect.FullArgSpec(args, argspec[1], argspec[2],
+                                       tuple(defaults), kwoargs,
+                                       kwodefs, argspec[6])
+        while hasattr(func, '__wrapped__'):
+            func = func.__wrapped__
+        if not inspect.isfunction(func):
+            raise TypeError('%r is not a Python function' % func)
+        return inspect.getfullargspec(func)
+elif sys.version_info >= (2, 5):
+
+    from functools import partial
+
+    def getargspec(func):
+        """Like inspect.getargspec but supports functools.partial as well."""
+        if inspect.ismethod(func):
+            func = func.im_func
+        parts = 0, ()
+        if type(func) is partial:
+            keywords = func.keywords
+            if keywords is None:
+                keywords = {}
+            parts = len(func.args), keywords.keys()
+            func = func.func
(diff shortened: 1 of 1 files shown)


Carreau:
Hi, 

thanks for the PR !

Usually for soft dependency, we put things in [external](https://github.com/ipython/ipython/tree/master/IPython/external) where they do a conditional import (here would be of sphinx). It is easier to maintain and diff. 

That being said, I just want to warn that as we are getting closer to release, this might not be merged **right now**. 

Otherwise this look like a reasonable request.

IPython/utils/traitlets.py
```diff
@@ -280,6 +280,86 @@ def dlink(source, *targets):
...
+                    del kwodefs[arg]
+            return inspect.FullArgSpec(args, argspec[1], argspec[2],
+                                       tuple(defaults), kwoargs,
+                                       kwodefs, argspec[6])
+        while hasattr(func, '__wrapped__'):
+            func = func.__wrapped__
+        if not inspect.isfunction(func):
+            raise TypeError('%r is not a Python function' % func)
+        return inspect.getfullargspec(func)
+elif sys.version_info >= (2, 5):
```
minrk: This is always True. IPython doesn't run on 2.5

takluyver:
I don't think we want to put the whole of Sphinx in external just to use one utility function. Sticking it in traitlets for now seems reasonable, though I'd like to ensure that it's not documented as being there, in case we want to move it into a separate module later.

Carreau:
> I don't think we want to put the whole of Sphinx in external just to use one utility function. Sticking it in traitlets for now seems reasonable, though I'd like to ensure that it's not documented as being there, in case we want to move it into a separate module later.

No I mean just the function, with a comment "extracted from Sphinx/..." so that it's easy to compare fo upstream. Of course not requiring full sphinx.

takluyver:
We usually put single extracted functions in utils, though, not external. e.g. TemporaryDirectory and signature backported from recent Pythons, and with_metaclass from six.

IPython/utils/traitlets.py
johanfforsberg: Ah, didn't consider that. The code is all verbatim from Sphinx, I suppose there's a value in keeping it as is, though...

minrk:
@johanfforsberg getargspec appears to work on functools.partial on Python 3. Can you provide examples where the `sys.version_info > (3,0)` branch has an effect? Maybe it only affects versions less than 3.3, which is the earliest Python 3 supported by IPython.

johanfforsberg:
I'm sorry I've been ignoring this PR for a while.

@minrk: I'll look into it. Should I remove the parts from the PR that aren't relevant to ipython because of version incompatibility? I was thinking it might be a better idea to keep the code as is from Sphinx from a future upgrade perspective, but I'll do whichever you guys think makes more sense.

minrk:
@johanfforsberg keeping it verbatim is a good idea from a maintenance perspective. That said, maybe we should be using the Python 3 signature stuff that we already have backported. Does that support functools.partial? If not, it's a moot point and we should go ahead with this in its own `utils.getargspec` file.

johanfforsberg:
I'm unsure what you mean when you talk about backported stuff. I found two different versions of "getargspec" in the IPython sources, but one of them has the same issue and the other one does not seem to work quite like the inspect one.

In any case, I've tried inspect.getargspec under python 3.4 and it does not work on partials either.

Carreau:
> I'm unsure what you mean when you talk about backported stuff. I found two different versions of "getargspec" in the IPython sources, but one of them has the same issue and the other one does not seem to work quite like the inspect one.

Ouh, that seem nasty. Will have a look.

Carreau:
> Ouh, that seem nasty. Will have a look.

I seem to find only one in `IPython/core/oinspect.py` excluding test dirs...

johanfforsberg:
Ah, you are right, I was looking too quickly. Anyway, the "oinspect" one does not work on partials.

minrk:
@johanfforsberg Thanks for checking on oinspect, I think we want this code as-is, then. Can you move the argspec to a new `utils.getargspec.py` file and import it into traitlets?

minrk:
I've opened #8117 to move the function to utils.getargspec

johanfforsberg:
Ah, sorry for not replying. Thanks for finishing it!

takluyver:
Closing this in favour of #8117. Thanks @johanfforsberg.

PR #7499, opened 2015-01-19
fix find_connection_file exclude notebook_cookie_secret

dongweiming:
i has start a notebook server. but when i use `find_connection_file`. i found this error:

``` python
In [1]: from IPython.kernel import find_connection_file

In [2]: find_connection_file('*')
Out[2]: u'/Users/dongweiming/.ipython/profile_default/security/notebook_cookie_secret'

In [3]: find_connection_file('')
Out[3]: u'/Users/dongweiming/.ipython/profile_default/security/notebook_cookie_secret'

In [4]: ls /Users/dongweiming/.ipython/profile_default/security
nbserver-92188.json     notebook_cookie_secret
```

You can see the result is a cookie secret file. so need exclude it.

``` python
$python -c "import IPython; print(IPython.sys_info())"
{'commit_hash': u'e2778c5',
 'commit_source': 'repository',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/dongweiming/ipython/IPython',
 'os_name': 'posix',
 'platform': 'Darwin-13.4.0-x86_64-i386-64bit',
 'sys_executable': '/usr/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '2.7.5 (default, Mar  9 2014, 22:15:05) \n[GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)]'}

```

### DIFF
--- a/IPython/kernel/connect.py
+++ b/IPython/kernel/connect.py
@@ -220,6 +220,8 @@ def find_connection_file(filename, profile=None):
         # accept any substring match
         pat = '*%s*' % filename
     matches = glob.glob( os.path.join(security_dir, pat) )
+    # exclude notebook_cookie_secret
+    matches = [m for m in matches if not m.endswith('notebook_cookie_secret')]
     if not matches:
         raise IOError("Could not find %r in %r" % (filename, security_dir))
     elif len(matches) == 1:


takluyver:
I don't think the fix is quite right: there are several different kinds of files that may live in the security directory, and we may add more in future, so excluding specific filenames doesn't seem viable.

I'm not really sure why this function takes a glob pattern rather than just a kernel ID. @minrk? Maybe the default pattern should be `kernel-*.json`?

minrk:
The glob pattern is used for the `--existing` flag, which connects to the most recent kernel. It should be `kernel-*.json`. That function hasn't been updated since connection files were the only file in there.

minrk:
Note that `find_connection_file` is never meant to be called with effectively no arguments, so the calling case that fails here is not a valid one.

takluyver:
See #7500.

PR #7502, opened 2015-01-19
Add Handling of cursor_end in Console Completer

### LINKED ISSUE #6690: ipython console ignores cursor_start and cursor_end on complete_reply messages

blink1073:
Closes #6690.  Extends #7497 and implements #6714.

### DIFF
--- a/IPython/terminal/console/completer.py
+++ b/IPython/terminal/console/completer.py
@@ -45,11 +45,16 @@ class ZMQCompleter(IPCompleter):
             code=line,
             cursor_pos=cursor_pos,
         )
-        
+    
         msg = self.client.shell_channel.get_msg(timeout=self.timeout)
         if msg['parent_header']['msg_id'] == msg_id:
-            cursor_start = msg['content']['cursor_start']
-            return [ line[:cursor_start] + m for m in msg['content']['matches'] ]
+            content = msg['content']
+            cursor_start = content['cursor_start']
+            matches = [ line[:cursor_start] + m for m in content['matches'] ]
+            if content["cursor_end"] < cursor_pos:
+                extra = line[content["cursor_end"]: cursor_pos]
+                matches = [m + extra for m in matches]
+            return matches
         return []
     
     def rlcomplete(self, text, state):


blink1073:
Any idea how to get rid of the change in `static.components`?  I tried a `git submodule update`.

takluyver:
This appears to break all completions in `ipython console` for me.

blink1073:
Ha, that will happen when you forget to return the matches.

takluyver:
Pulled the branch again, still not getting any completions.

Not sure how the components change got in there - don't add that change when you commit, and it shouldn't show up in the PR.

blink1073:
Okay, completions are working on my end now.  I still can't shake the submodules change.

blink1073:
Note that the #7497 merge still suffers from part of the problem that stumped me in #6714, in that the whole line is shown for completions (at least its not the whole buffer anymore).

takluyver:
I don't know that we can easily avoid that given the way readline works.

This seems to be working now, but you'll still need to rebase and remove the components change.

blink1073:
Done!

takluyver:
Thanks, I think this looks good now, but I'll let someone else have a look over it before merging.

minrk:
@blink1073 can you provide an example this improves?

minrk:
This makes sense to me, I just can't think of a case where cursor_end would be less than cursor_start. That would mean triggering a completion that changes some part of the line that doesn't involve the cursor, leaving the area where the completion was triggered unmodified. Maybe a non-Python kernel does this?

Carreau:
> I just can't think of a case where cursor_end would be less than cursor_start

julia ? where `foo.method()` is `method(foo)` and you might want `foo<tab>` ?

minrk:
@Carreau I don't think that's an example, unless I'm misunderstanding. cursor_end and cursor_start bound the part of the *current* cell that are to be replaced. So only if the cursor is *entirely outside* the replacement should this be the case.

blink1073:
@minrk, I don't have a real world example, this was about implementing the full API and aligning with the behaviour of the QtConsole and the Notebook.   @dsblank, can you think of a good example where we'd want to set `cursor_end` to a value less than `cursor_pos`?

minrk:
Makes sense, anyway. Merging away.

PR #7504, opened 2015-01-19
Fixes #7503 - Cell magics fail to switch the language

### LINKED ISSUE #7503: Cell magics fail to switch the language
To reproduce, create a notebook with a cell like this:

```
%%sh
a=${HOME}
echo $a
```

Select File / Print Preview.  Observe the red boxes around `$`'s.

```html
..
<span class="n">a</span>
<span class="o">=</span>
<span class="err">$</span>
<span class="p">{</span>
<span class="n">HOME</span>
<span class="p">}</span>
..
```

I am using Python 3.4.2 and IPython 3.0.0-dev.

abalkin:
Recognize "ipython3" as a default language to be replaced
in the presence of cell magic.

### DIFF
--- a/IPython/nbconvert/filters/highlight.py
+++ b/IPython/nbconvert/filters/highlight.py
@@ -113,7 +113,7 @@ def _pygments_highlight(source, output_formatter, language='ipython', metadata=N
 
     # If the cell uses a magic extension language,
     # use the magic language instead.
-    if language == 'ipython' \
+    if language.startswith('ipython') \
         and metadata \
         and 'magics_language' in metadata:
 


IPython/nbconvert/filters/highlight.py
```diff
@@ -113,7 +113,7 @@ def _pygments_highlight(source, output_formatter, language='ipython', metadata=N
 
     # If the cell uses a magic extension language,
     # use the magic language instead.
-    if language == 'ipython' \
+    if language in ('ipython', 'ipython3') \
```
takluyver: Looking at the code a few lines below, I think this should accept 'ipython2' as well.

abalkin: I assumed that it worked in python2, but was too lazy to test.  Let me test now.

abalkin: The patched version works fine with python2:
 
screen shot 2015-01-19 at 4 03 15 pm

takluyver: Referring to line 122 in this file, and to [this line](https://github.com/ipython/ipython/blob/master/IPython/kernel/zmq/ipkernel.py#L114), I'm pretty sure that 'ipython2' is a valid value that may be passed to this function, even if it isn't in the case you've tested.

abalkin: Would `if language.startswith('ipython')` test be too permissive?

abalkin: I am hesitant to make more changes than necessary to fix the bug.  It is possible that in the cases where `python2` is passed in as language, it should not be replaced with the magics language.

takluyver: I think I'd be happy with `language.startswith('ipython')`.

I'm working on the principle that whatever behaviour is desired should be symmetrical for Python 2 and 3. It's possible that there's some situation where it's not right, but if it's the same for both Python versions, we have the best chance of understanding that issue and fixing it in a way that doesn't introduce more issues.

abalkin:
@takluyver - I made the change that you suggested.

takluyver:
Great, thanks.

PR #7505, opened 2015-01-19
Dashboard checkbox style interface

### LINKED ISSUE #7300: Actions dropdown on dashboard
Also for folders...

### LINKED ISSUE #7364: File Management UI supports creation of directories, but not their deletion
On multiple occasions I've accidentally clicked the Folder option in the new dropdown for creating notebooks, causing me to create an "Untitled Folder" directory that I now can't delete... :crying_cat_face: .

jdfreder:
closes #7300
closes #7364

MVP contents API GUI for 3.0.  Ping @ellisonbg who requested this.

### DIFF
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -518,6 +518,11 @@ define([
                 Delete : {
                     class: "btn-danger",
                     click: function() {
+                        // Shutdown any/all selected notebooks before deleting 
+                        // the files.
+                        that.shutdown_selected();
+
+                        // Delete selected.
                         that.selected.forEach(function(item) {
                             that.contents.delete(item.path).then(function() {
                                     that.notebook_deleted(item.path);
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -264,6 +264,7 @@ define([
         if (selectable !== undefined) {
             checkbox = $('<input/>')
                 .attr('type', 'checkbox')
+                .attr('title', 'Click here to rename, delete, etc.')
                 .appendTo(item);
         }
 
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -329,7 +329,7 @@ define([
 
     NotebookList.prototype._selection_changed = function() {
         var selected = [];
-        var has_notebook = false;
+        var has_running_notebook = false;
         var has_directory = false;
         $('.list_item :checked').each(function(index, item) {
             var parent = $(item).parent().parent();
@@ -339,7 +339,8 @@ define([
                 type: parent.data('type')
             });
 
-            has_notebook = has_notebook || parent.data('type') == 'notebook';
+            has_running_notebook = has_running_notebook || 
+                (parent.data('type') == 'notebook' && that.sessions[parent.data('path')] !== undefined);
             has_directory = has_directory || parent.data('type') == 'directory';
         });
         this.selected = selected;
@@ -352,7 +353,7 @@ define([
         }
 
         // Shutdown is only visible when one or more notebooks are visible.
-        if (has_notebook) {
+        if (has_running_notebook) {
             $('.shutdown-button').css('display', 'inline-block');
         } else {
             $('.shutdown-button').css('display', 'none');
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -597,12 +597,13 @@ define([
         /**
          * Remove the deleted notebook.
          */
+        var that = this;
         $( ":data(path)" ).each(function() {
             var element = $(this);
-            if (element.data("path") == path) {
+            if (element.data("path") === path) {
                 element.remove();
                 events.trigger('notebook_deleted.NotebookList');
-                this._selection_changed();
+                that._selection_changed();
             }
         });
     };
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -87,6 +87,11 @@ define([
                     that.load_list();
                 });
             });
+
+            $('.rename-button').click($.proxy(this.rename_selected, this));
+            $('.shutdown-button').click($.proxy(this.shutdown_selected, this));
+            $('.duplicate-button').click($.proxy(this.duplicate_selected, this));
+            $('.delete-button').click($.proxy(this.delete_selected, this));
         }
     };
 
@@ -111,7 +116,7 @@ define([
                 // read non-notebook files as binary
                 reader.readAsArrayBuffer(f);
             }
-            var item = that.new_item(0);
+            var item = that.new_item(0, true);
             item.addClass('new-file');
             that.add_name_input(f.name, item, file_ext == '.ipynb' ? 'notebook' : 'file');
             // Store the list item in the reader so we can use it later
@@ -218,7 +223,7 @@ define([
         var path = this.notebook_path;
         var offset = n_uploads;
         if (path !== '') {
-            item = this.new_item(offset);
+            item = this.new_item(offset, false);
             model = {
                 type: 'directory',
                 name: '..',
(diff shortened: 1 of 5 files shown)


jdfreder:
Scrolled down the list a bit:
capture

minrk:
Should shutdown be under the actions list? If it's only outside because it's the only indicator that a notebook is running, perhaps we can add a different indicator instead.

jdfreder:
> If it's only outside because it's the only indicator that a notebook is running, perhaps we can add a different indicator instead.

Precisely.

One idea, which I hesitate to try, is to color the notebooks if they are running - maybe green.
Another, use a different icon, like an open notebook or the logo of the kernel.
Another, display a second icon, either or the left or right side of the row, which indicates the notebook is running.  This icon could also be the kernel logo, to additionally indicate what type of kernel is running.

ellisonbg:
Is there a good font awesome icon to indicate "running" state? We could use
that next to the Action menu as a static indicator and put Shutdown inside
the menu.

On Mon, Jan 19, 2015 at 1:19 PM, Jonathan Frederic <notifications@github.com
> wrote:

> If it's only outside because it's the only indicator that a notebook is
> running, perhaps we can add a different indicator instead.
>
> Precisely.
>
> One idea, which I hesitate to try, is to color the notebooks if they are
> running - maybe green.
> Another, use a different icon, like an open notebook or the logo of the
> kernel.
> Another, display a second icon, either or the left or right side of the
> row, which indicates the notebook is running. This icon could also be the
> kernel logo, to additionally indicate what type of kernel is running.
>
> —
> <https://github.com/ipython/ipython/pull/7505#issuecomment-70562626>.
>



-- 
Brian E. Granger

jdfreder:
capture

jdfreder:
Hey @minrk and @ellisonbg and other listeners,

I realized that it's barely any more effort to use a checkbox layout, which is our end goal anyways.  I've updated this PR so it's now a checkbox style GUI, similar to GMAIL.  See attached screenshot:

screen shot 2015-01-20 at 4 21 24 pm

ellisonbg:
Cool, will start to test this!

IPython/html/static/tree/js/notebooklist.js
```diff
@@ -229,34 +234,84 @@ define([
...
-            $("<a/>").addClass("item_link").append(
-                $("<span/>").addClass("item_name")
-            )
-        ).append(
-            $('<div/>').addClass("item_buttons  pull-right")
-        ));
+    /**
+     * Creates a new item.
+     * @param  {integer} index
+     * @param  {boolean} [selectable] - tristate, undefined: don't draw checkbox,
```
minrk: What's the third state for?

jdfreder: The running tab uses 'undefined' - don't render checkboxes.
The bread crumbs use 'false' - css visibility = hidden (pad as if the checkbox was rendered)
The directories, files, and notebooks use 'true' - show checkbox

Carreau:
> I realized that it's barely any more effort to use a checkbox layout, which is our end goal anyways. I've updated this PR so it's now a checkbox style GUI, similar to GMAIL. See attached screenshot:

Like !

I would separate Rename/duplicate (need single notebook checked) from shutdown/delete (can select multiple), from new/refresh (need nothing selected).
Separation could be something like a vertical bar. Or just a bigger space. 

Icon selection that might be useful: 

  -  http://fortawesome.github.io/Font-Awesome/icon/trash/ 
  -  http://fortawesome.github.io/Font-Awesome/icon/trash-o/ 
  -  http://fortawesome.github.io/Font-Awesome/icon/recycle/

IPython/html/static/tree/js/notebooklist.js
minrk: Okay. I think that suggests we might be using one function for too many things, but it's okay, for now. I like the result, in general.

minrk:
Bugs noticed while playing with it:

- Shutdown is listed as an action, even when selected notebook is not running.
- On delete, I get `TypeError: undefined is not a function`.
- Shutdown is listed as an action if a notebook is selected, even if a directory is also selected (actions should only be the subset that applies to all selected).

non-bug issues:

- Until checked, there's no indication of why I would check a box. Maybe add hover text to the checkboxes?
- Since the buttons are on the same scrolling element as the list, it's quite tedious to delete an item late in the list because it means scrolling down to the item, and scrolling all the way back to click delete.
- Since there appears to be no good icon for it, perhaps use the text "Running" to indicate that a notebook is running.

jdfreder:
Thanks for the feedback !

> Shutdown is listed as an action if a notebook is selected, even if a directory is also selected (actions should only be the subset that applies to all selected).

Indeed the shutdown logic is not correct.  However, I think it should display if atleast one of the items selected are a notebook that is running.  That way, when we add a way for the user to select all X (i.e. notebooks), the shutdown button will work- it will shutdown all of the notebooks it can. 

> On delete, I get TypeError: undefined is not a function.

Whoops, last minute change I did not check before committing, will fix.

jdfreder:
Ug, stupid close and comment button...

jdfreder:
> On delete, I get TypeError: undefined is not a function.

After playing with it a bit more, I can't seem to find out how to reproduce this... Does it happen under some circumstances and not with others?

jdfreder:
@minrk here's what it looks like with the text instead of the power off icon:

screen shot 2015-01-21 at 11 50 29 am

minrk:
This removes the ability to take any actions on the Running tab.

minrk:
@jdfreder can you investigate making the item list scrollable, so the action buttons stay on the screen?

ellisonbg:
@minrk - we talked about this yesterday. I completely agree with you that
in the end it will be much nicer. At the same time, trying to get the page
the not to scroll when the Files tab is open, but have the page scroll when
the Running or Cluster tab open will be a total hack/pain. Obviously, the
right way to do this is to have the content areas of each tab scroll (page
never scrolls), but I was a little worried about that being a rabbit hole.
What about this:

* Let's merge this with the existing rough set of functionality (but proper
review and testing) soon so more people can try it out.
* Let's gather feedback and see what other changes we want to make before
we release, and in the meantime explore how we can implement the scrolling
in a clean way.
* One example of an additional change would be to have a dropdown at the
top that quickly let's a user select different sets of things like {all
notebooks, all running notebooks, all files, all directories}. But I think
that can go into a later PR as well.


On Fri, Jan 23, 2015 at 11:18 AM, Min RK <notifications@github.com> wrote:

> @jdfreder <https://github.com/jdfreder> can you investigate making the
> item list scrollable, so the action buttons stay on the screen?
>
> —
> <https://github.com/ipython/ipython/pull/7505#issuecomment-71248490>.
>



-- 
Brian E. Granger

minrk:
That works for me. MVP, iterate in master, and all that. It's only an issue on long file lists, anyway.

> One example of an additional change would be to have a dropdown at the
top that quickly let's a user select different sets of things like {all
notebooks, all running notebooks, all files, all directories}. But I think
that can go into a later PR as well.

Yup, that can go in the theoretical major dashboard redesign that keeps being on the horizon.

ellisonbg:
At least for now, I think it is more reasonable for us to get to that "major dashboard redesign" with these types of incremental improvements, rather than starting over from scratch.

jdfreder:
@minrk I re-added a shutdown button to the running group :)

jdfreder:
Also, I prototyped a scrolling notebook list yesterday, and it looked and behaved nice**-ish**.  There were some small issues which is one of the reasons why @ellisonbg and I think it could be a rabbit hole.

minrk:
@jdfreder still working, or ready for another review pass?

jdfreder:
> still working, or ready for another review pass?

Depends on the decision about the shutdown button's behavior.  Right now I have it showing `if (has_running_notebook) {`.  Should that be changed to `if (only_running_notebooks) {` or  `if (only_notebooks && has_running_notebook) {` ?  I'm in favor of the current behavior, but I think you favored one of the other two?

minrk:
I would only have actions available that apply to all selected items. That would suggest that shutdown should only be available if everything selected is at least a notebook, and possibly only if they are all running. I don't have a strong preference between the two, but I do think shutdown should not be an option if directories or non-notebook files are selected.

takluyver:
I'd intuitively side with actions that apply to any selected items. Though I could imagine being unsure if clicking shutdown with a directory selected would shut down any notebooks inside that folder, which is a point against my intuitive preference.

ellisonbg:
I think Thomas makes a really good point about Shutdown when a directory is
selected. Because of this potential for confusion, I do think Min's idea of
only allowing action that apply to all selected items makes the most sense.

On Fri, Jan 23, 2015 at 3:47 PM, Thomas Kluyver <notifications@github.com>
wrote:

> I'd intuitively side with actions that apply to any selected items. Though
> I could imagine being unsure if clicking shutdown with a directory selected
> would shut down any notebooks inside that folder, which is a point against
> my intuitive preference.
>
> —
> <https://github.com/ipython/ipython/pull/7505#issuecomment-71286410>.
>



-- 
Brian E. Granger

Carreau:
Agreed with Min too, what if I have notebooks selected that are both with and without kernels and click delete ? Action is ambiguous at best. Make some kernels unstoppable at best.

minrk:
Needs a rebase, now.

jdfreder:
Hey guys, rebased and behavior changed (see last commit).

Now the logic is to show shutdown if **only notebooks are selected AND at least one running notebook is selected**.  This will make it easy for people to (in the future) select all notebooks, and click shutdown.  Also, when are range of notebooks are selected, one doesn't have to worry about deselecting the kernel-less ones in order to see the shutdown button.

minrk:
@jdfreder I'm still getting `TypeError: undefined is not a function` when I delete any file.

minrk:
@jdfreder it was a missing `that = this`, see [here](https://github.com/minrk/ipython/compare/jdfreder:actions...checkbox-dashboard).

jdfreder:
Ahh sorry @minrk .  That's for the super convenient commit- saves me some keystrokes :)  I pulled it in.

jdfreder:
@minrk earlier you asked if this was ready for another review pass, now I believe it is.

IPython/html/static/tree/js/notebooklist.js
```diff
@@ -332,104 +431,179 @@ define([
...
-                            click: function() {
-                                notebooklist.contents.delete(path).then(
-                                    function() {
-                                        notebooklist.notebook_deleted(path);
+    NotebookList.prototype.delete_selected = function() {
+        var message;
+        if (this.selected.length == 1) {
+            message = 'Are you sure you want to permanently delete: ' + this.selected[0].name + '?';
+        } else {
+            message = 'Are you sure you want to permanently delete the ' + this.selected.length + ' files selected?';
```
takluyver: Should this say something like '... files/folders selected?' I don't think of 'files' as including folders.

takluyver:
I've done a quick review pass, but someone more familiar with this stuff should look over it too.

IPython/html/static/tree/js/notebooklist.js
jdfreder: Maybe something more generic, like items.

takluyver: 'items' is a bit too generic for my taste, but I could live with it.

jdfreder: files/folders it is!

jdfreder:
Review comment addressed

minrk:
After playing with it, it seems to behave sensibly, at this point. :+1: to merge, and test in master.

takluyver:
OK, let's land this.

dsblank:
Nice work, all!

PR #7506, opened 2015-01-19
handle unicode issues in `ipython console` completions

minrk:
- readline.get_endidx gives byte offset, not character offset
- readline talks str, not unicode

symptom on master: any completion after a multi-byte character will insert extra spaces, e.g.

```
α <TAB>=
```

becomes

```
α =|=
```

### DIFF
--- a/IPython/terminal/console/completer.py
+++ b/IPython/terminal/console/completer.py
@@ -11,6 +11,7 @@ except ImportError:
 
 from IPython.config import Configurable
 from IPython.core.completer import IPCompleter
+from IPython.utils.py3compat import str_to_unicode, cast_bytes, unicode_to_str
 from IPython.utils.traitlets import Float
 import IPython.utils.rlineimpl as readline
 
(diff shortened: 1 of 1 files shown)


IPython/terminal/console/completer.py
```diff
@@ -36,8 +37,17 @@ def __init__(self, shell, client, config=None):
...
+        byte_cursor_pos = readline.get_endidx()
+        
+        # get_endidx is a byte offset
+        # account for multi-byte characters to get correct cursor_pos
+        cursor_pos = byte_cursor_pos
+        i = 0
+        while i < cursor_pos:
+            bytelen = len(cast_bytes(line[i]))
+            cursor_pos -= (bytelen-1)
+            i += 1
```
takluyver: Can we not do this without the loop? Something like this:

```python
byte_cursor_pos = readline.get_endidx()
bytes_before_cursor = cast_bytes(line)[:byte_cursor_pos]
unicode_cursor_pos = len(cast_unicode(bytes_before_cursor))
```

It probably has weird behaviour if readline reports we're in the middle of a multibyte character, but in that case weird behaviour seems likely no matter what we do.

minrk: that makes sense. If encoding and/or decoding fails, you might get off by one, but I'm not too worried about that.

minrk: Switched to simpler computation of the offset

PR #7508, opened 2015-01-20
Don't use widget.description as kwarg key

### LINKED ISSUE #7507: description keyword for interactive widget
When I use the keyword description with interact

```python
from IPython.html import widgets
def simpleFunction(a):
    print a

i = interact(simpleFunction, 
                  a=widgets.FloatSliderWidget(values=(0,10), description='A label for my widget'))
```
I get this error
```
TypeError: simpleFunction() got an unexpected keyword argument 'A label for my widget'
```

minrk:
in interact.

description can be set manually, so it cannot be relied upon to store the kwarg key. Add a `widget._kwarg` attribute instead.

closes #7507

### DIFF
--- a/IPython/html/widgets/interaction.py
+++ b/IPython/html/widgets/interaction.py
@@ -155,6 +155,7 @@ def _widgets_from_abbreviations(seq):
         widget = _widget_from_abbrev(abbrev, default)
         if not widget.description:
             widget.description = name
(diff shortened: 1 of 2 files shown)


takluyver:
:+1: 

I think this warrants a test case or an example that we can run manually.

minrk:
Test added that fails on master.

PR #7513, opened 2015-01-20
Don't link file again if it's already a correct hard link

### LINKED ISSUE #6450: Temporary files in FilesWriter
I am not sure if this is a bug or not, but it was unexpected behavior for me, so I figured I'd report it.

When using `FilesWriter` with nbconvert, if you are converting multiple notebooks at once that rely on the same set of supplementary files, they do get copied to the build directory, but so do temporary files (one version per extra notebook). For example, if I have the following directory/file structure:

```
$ ls -R
.:
bar.ipynb  files/  foo.ipynb  ipython_nbconvert_config.py

./files:
file1.txt  file2.txt
```

Where `ipython_nbconvert_config.py` is:

```python
from glob import glob

c = get_config()

c.NbConvertApp.export_format = 'notebook'
c.NbConvertApp.notebooks = glob('*.ipynb')
c.NbConvertApp.writer_class = 'FilesWriter'

c.FilesWriter.build_directory = 'build'
c.FilesWriter.files = glob('files/*')
```

Then, if I run nbconvert, the `build/files` directory contains:

```
$ ls build/files/
file1.txt  file1.txt-temp-77A5  file2.txt  file2.txt-temp-CE50
```

I traced the root of the problem to `IPython.utils.path.link_or_copy`, which says "Because ``os.link`` does not overwrite files, a unique temporary file will be used if the target already exists, then that file will be moved into place." That's fine, except it seems like the `os.rename(new_dst, dst)` line isn't actually removing `new_dst` (which is the temporary file).

takluyver:
Closes #6450

### DIFF
--- a/IPython/utils/path.py
+++ b/IPython/utils/path.py
@@ -554,6 +554,12 @@ def link_or_copy(src, dst):
 
     link_errno = link(src, dst)
     if link_errno == errno.EEXIST:
+        if os.stat(src).st_ino == os.stat(dst).st_ino:
+            # dst is already a hard link to the correct file, so we don't need
+            # to do anything else. If we try to link and rename the file
+            # anyway, we get duplicate files - see http://bugs.python.org/issue21876
+            return
+
         new_dst = dst + "-temp-%04X" %(random.randint(1, 16**4), )
         try:
             link_or_copy(src, new_dst)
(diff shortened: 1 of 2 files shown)


minrk:
Confirmed this fixed #6450 for me. @jhamrick, want to confirm this one?

jhamrick:
Yep, fixes it for me as well!

Issue #5570, opened 2014-04-09
User Interface tour: too fast, too furious


jakevdp:
If you cycle through the user interface tour too fast, the tour windows do not go away and make the notebook unusable.

To replicate, open Help->User Interface Tour in the notebook, and then press the ``→`` (right arrow) key several times very quickly.

PR #7515, opened 2015-01-20
Fix user interface tour

### LINKED ISSUE #7418: User interface tour broken.
Don't start at all;  raise as undefined not a function deep in bootstrap.

jhamrick:
This fixes #7418 
- updates the version of bootstrap tour
- removes usage of the backdrop, since it seems to be broken
- fix various element ids so the tour points to the right element

### DIFF
--- a/IPython/html/static/notebook/js/tour.js
+++ b/IPython/html/static/notebook/js/tour.js
@@ -40,13 +40,11 @@ define([
             }, {
                 element: $("#menus").parent(),
                 placement: 'bottom',
-                backdrop: true,
                 title: "Notebook Menubar",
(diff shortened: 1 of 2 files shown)


jhamrick:
Ideally it would be nice to have tests to catch when the tour breaks, but I remember trying to write tests for it a while back and failing to properly get the keyboard events working (or something along those lines). Unfortunately, I don't think I'll have time to try to get them working now, though I might in a few weeks.

jhamrick:
I think this also fixes #5570

minrk:
I merged ipython/ipython-components#52, so you can update the components ref to master.

jhamrick:
Cool, thanks.

minrk:
Thanks!

PR #7521, opened 2015-01-21
change keyboard help to use `esc` and adds key symbol table

amcdawes:
Work in progress, this adds a key symbol table for MacOS as per #7322 

Should automate the creation of the symbol table. I've put the symbol table in the alert box at the top, this is a design choice that needs review but I feel it is on par with the modal information that is already there.

### DIFF
--- a/IPython/html/static/notebook/js/quickhelp.js
+++ b/IPython/html/static/notebook/js/quickhelp.js
@@ -206,7 +206,7 @@ define([
                     //'⌫	Delete back'+
                     //'⌦	Delete forward'+
                     //'<div class="col-md-4">col3</div>'
-            );
+                    );
         }
         element.append(doc);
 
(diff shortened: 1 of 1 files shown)


amcdawes:
Should be ready for review now.

Carreau:
Looks nice !
capture d ecran 2015-01-22 a 15 53 54
 
And code looks fine. 

I guess there might be a better way to structure that like:
`short-name , long-name, symbol` and have only one mapping, 
but I think that a project for later.

Carreau:
Side note (not for this PR). Should we really make the yellow part closable ? 

The closing is not persisted, and I don't see any reason to close it.

Otherwise +1 to this, I'lll let other give their thought

IPython/html/static/notebook/js/quickhelp.js
```diff
@@ -203,6 +207,34 @@ define([
...
 
+    QuickHelp.prototype.build_key_names = function () {
+       var key_names_mac =  [{ shortcut:"⌘", help:"Command" },
+                    { shortcut:"⌃", help:"Control" },
+                    { shortcut:"⌥", help:"Option" },
+                    { shortcut:"⇧", help:"Shift" },
+                    { shortcut:"↩", help:"Return" },
+                    { shortcut:"␣", help:"Space" },
+                    { shortcut:"⇥", help:"Tab forward" },
+                    { shortcut:"⇤", help:"Tab back" }];
```
minrk: I don't think "Tab back" is a key. It's certainly not a key we use, so it should probably be removed and "Tab forward" should be "Tab".

amcdawes: I was curious about that too, meant to delete earlier, fixed now.

amcdawes:
I'm going to need help understanding why Travis failed this. I don't see anything that pertains to this PR listed in the failed tests... but maybe I'm misreading the results.

Carreau:
> I'm going to need help understanding why Travis failed this. I don't see anything that pertains to this PR listed in the failed tests... but maybe I'm misreading the results.

Weird, I tried to relaunch travis.

minrk:
@amcdawes there are a few intermittent failures that can be caused by timing issues, looks like this was one of those.

minrk:
Thanks, @amcdawes!

PR #7522, opened 2015-01-21
remember and persist manual scroll state.

### LINKED ISSUE #2105: don't collapse output cells as the data arrives (chromium)
Here's a snippet, that, as the data arrives, will cause the cell to collapse, but clicking to expand it gets ignored as new data arrives and the cell is once again collapsed.

```python
import time
for x in range(500):
    print 1
    print 2
    print 3
    print 4
    print 5
    time.sleep(.1)
```

this one will do it, too

```python
%pylab inline
import time
for x in range(100):
    plt.figure()
    plt.plot(range(10))
    plt.show()
    time.sleep(.1)
```

I'm only seeing this in Chromium at the moment, but I think that's just because this auto-collapse after reaching some threshold functionality was disabled in Firefox (is that right?). FF keeps the state (without toggling) as the data arrives (which seems like the correct behavior to me).

(sorry, I poked around `codecell.js` and `outputarea.js` but without much luck, and figured someone who knows that side of things better would be able to do it much quicker and cleaner)

### LINKED ISSUE #2172: Auto-scrolling of long output should be configurable in the UI
We've had on-list, in person and here (#2154) various requests for different thresholds or behaviors regarding auto-scrolling.  Ultimately there's no one good default that fits all cases, so this should have a UI control for users to enable/disable the feature as well as changing the threshold.

### LINKED ISSUE #3657: Notebook cell output state is not preserved by save
The output state (full, scroll, collapsed) of a cell in the notebook should be preserved by saving (and carried over to nbconvert output), but it currently opens to full output on upon close and halt followed by reloading.

Closely related to #3659

### LINKED ISSUE #3659: Running a cell will open up output totally

minrk:
When user toggles the scroll state, the choice is remembered (in OutputArea.scroll_state) and persisted (in Notebook.metadata.scrolled).

There are three states:
- 'auto' (default, only state in master)
- true (always scroll if above minimum_scroll_threshold)
- false (never scroll)

true or false is persisted in cell.metadata.scrolled. "auto" may be persisted according to the schema, but isn't currently because it's the default state.

There is no UI for setting the value back to 'auto', and I'm not sure there's a good one, nor is there a need for one. That said, the scroll state can be cleared by collapsing and expanding the output.

closes #3657
closes #3659
closes #2105
closes #2172

### DIFF
--- a/IPython/html/static/notebook/js/codecell.js
+++ b/IPython/html/static/notebook/js/codecell.js
@@ -624,14 +624,7 @@ define([
             }
             this.set_input_prompt(data.execution_count);
             this.output_area.trusted = data.metadata.trusted || false;
-            this.output_area.fromJSON(data.outputs);
-            if (data.metadata.collapsed !== undefined) {
-                if (data.metadata.collapsed) {
-                    this.collapse_output();
-                } else {
-                    this.expand_output();
-                }
-            }
+            this.output_area.fromJSON(data.outputs, data.metadata);
         }
     };
 
@@ -649,6 +642,11 @@ define([
         data.outputs = outputs;
         data.metadata.trusted = this.output_area.trusted;
         data.metadata.collapsed = this.output_area.collapsed;
+        if (this.output_area.scroll_state === 'auto') {
+            delete data.metadata.scrolled;
+        } else {
+            data.metadata.scrolled = this.output_area.scroll_state;
+        }
         return data;
     };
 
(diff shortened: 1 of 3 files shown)


takluyver:
This looks good, though I haven't reviewed it in detail. I think you got all the debugging statements.

Carreau:
+1

PR #7523, opened 2015-01-21
Fix the output widget

### LINKED ISSUE #7509: Output widget broken
Caused by the grand asynchronous movement of 2014.  Specifically, the asynchronous code in the comm that calls back the user registered message handling function.

The output assumes (now incorrectly) that the widget custom messages are received by the front-end in the same order as the stdout/display messages. 

@takluyver @jasongrout :horse:

jdfreder:
Changes the output widget's behavior so the output capturing occurs kernel side.

closes #7509

Ping @jasongrout who seemed interested.

### DIFF
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -212,11 +212,9 @@ define([
             json.name = content.name;
         } else if (msg_type === "display_data") {
             json.data = content.data;
-            json.output_type = msg_type;
             json.metadata = content.metadata;
         } else if (msg_type === "execute_result") {
             json.data = content.data;
-            json.output_type = msg_type;
             json.metadata = content.metadata;
             json.execution_count = content.execution_count;
         } else if (msg_type === "error") {
--- a/IPython/html/static/widgets/js/widget_output.js
+++ b/IPython/html/static/widgets/js/widget_output.js
@@ -46,30 +46,14 @@ define([
         /**
          * Handles re-routed iopub messages.
          */
-        _handle_route_msg: function(content) {
-            if (content) {
-                var msg_type = content.type;
-                var json = {
-                    output_type: msg_type
-                };
-
-                var data = content.args[0];
+        _handle_route_msg: function(msg) {
+            if (msg) {
+                var msg_type = msg.msg_type;
                 if (msg_type=='clear_output') {
-                    this.output_area.clear_output(data.wait || false);
-                    return;
-                } else if (msg_type === "stream") {
-                    data = content.kwargs.content;
-                    json.text = data.text;
-                    json.name = data.name;
-                } else if (msg_type === "display_data") {
-                    json.data = data.data;
-                    json.metadata = data.metadata;
+                    this.output_area.handle_clear_output(msg);
                 } else {
-                    console.log("unhandled output message", msg);
-                    return;
+                    this.output_area.handle_output(msg);
                 }
-
-                this.output_area.append_output(json);    
             }
         },
     });
--- a/IPython/html/static/widgets/js/widget_output.js
+++ b/IPython/html/static/widgets/js/widget_output.js
@@ -9,18 +9,18 @@ define([
     'use strict';
 
     var OutputView = widget.DOMWidgetView.extend({
+        /**
+         * Public constructor
+         */
         initialize: function (parameters) {
-            /**
-             * Public constructor
-             */
             OutputView.__super__.initialize.apply(this, [parameters]);
             this.model.on('msg:custom', this._handle_route_msg, this);
         },
 
+        /**
+         * Called when view is rendered.
+         */
         render: function(){
-            /**
-             * Called when view is rendered.
-             */
             this.output_area = new outputarea.OutputArea({
                 selector: this.$el, 
                 prompt_area: false, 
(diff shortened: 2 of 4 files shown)


IPython/html/widgets/widget_output.py
```diff
@@ -38,13 +38,30 @@ def clear_output(self, *pargs, **kwargs):
...
+        self._original_send = send
+        self._session = session
+
+        def send_hook(stream, msg_or_type, *args, **kwargs):            
+            if stream is kernel.iopub_socket and msg_or_type in ['clear_output', 'stream', 'display_data']:
+                msg = {'type': msg_or_type, 'args': args, 'kwargs': kwargs}
+                self.send(msg)
+            else:
+                send(stream, msg_or_type, *args, **kwargs)
+                return
```
takluyver: I think this return is superfluous.

IPython/html/widgets/widget_output.py
```diff
@@ -38,13 +38,30 @@ def clear_output(self, *pargs, **kwargs):
...
         self._flush()
-        self.send({'method': 'push'})
+        kernel = get_ipython().kernel
+        session = kernel.session
+        send = session.send
+        self._original_send = send
+        self._session = session
+
+        def send_hook(stream, msg_or_type, *args, **kwargs):            
+            if stream is kernel.iopub_socket and msg_or_type in ['clear_output', 'stream', 'display_data']:
```
takluyver: Should this also handle `msg_or_type` being an already composed message? This looks like it assumes that `msg_or_type` is only the type.

jdfreder:
@ellisonbg

IPython/html/widgets/widget_output.py
jdfreder: I don't think so, but I may not understand all of the different entry points to this method.  The time which I saw a full message being passed in was for the repr/display hook, which shouldn't be captured in the context anyways.

jdfreder:
@minrk I noticed the data fed into Python `session.send` does not map 1 to 1 to Javascript `OutputArea.prototype.handle_output = function (msg) {`.  This results in me having to maintain my own map between the two, which is not ideal.  Do you know if there is a built-in method in Python to compile the `msg` as expected by Javascript `OutputArea.prototype.handle_output`?  If not, where should I be hooking other than sessions.send that does create/have access to the correct `msg` to forward over the widget comm?

IPython/html/widgets/widget_output.py
jdfreder: Yup

IPython/html/static/widgets/js/widget_output.js
```diff
@@ -43,14 +43,33 @@ define([
...
+                var msg_type = content.type;
+                var json = {
+                    output_type: msg_type
+                };
+
+                var data = content.args[0];
+                if (msg_type=='clear_output') {
+                    this.output_area.clear_output(data.wait || false);
+                    return;
+                } else if (msg_type === "stream") {
```
jasongrout: Wouldn't it be nicer to have a switch statement?

IPython/html/widgets/widget_output.py
jasongrout: Perhaps, just in case,  you should throw an error if msg_or_type isn't what you expect.

jdfreder: I can't throw an error because messages that I don't handle must be handled by the default mechanism.

takluyver: I'd be inclined to handle the possibility anyway, because the method you're replacing allows that possibility, and even if none of the code currently uses it, it may in the future. And it should be an easy couple of lines to handle it.

IPython/html/static/widgets/js/widget_output.js
minrk: yes

IPython/html/static/notebook/js/outputarea.js
```diff
@@ -212,11 +212,9 @@ define([
             json.name = content.name;
         } else if (msg_type === "display_data") {
             json.data = content.data;
-            json.output_type = msg_type;
```
minrk: why have you removed setting output_type?

minrk:
@jdfreder I don't fully understand your question, but `Session.send` returns the message it built. If you need the message dict before sending, you can use `Session.msg`.

IPython/html/static/notebook/js/outputarea.js
jdfreder: It's redundant, see line 208 which also sets it.

jdfreder:
@minrk thanks!  That's exactly what I was looking for.

All comments addressed, I think this is done the "right way" now.

IPython/html/widgets/widget_output.py
```diff
@@ -38,13 +39,41 @@ def clear_output(self, *pargs, **kwargs):
...
+                msg_type = msg_or_type
+                msg = session.msg(msg_type, content=content, parent=parent, 
+                    header=header, metadata=metadata)
+
+            # If this is a message type that we want to forward, forward it.
+            if stream is kernel.iopub_socket and msg_type in ['clear_output', 'stream', 'display_data']:
+                self.send(msg)
+            else: 
+                send(stream, msg_or_type, content=content, parent=parent, ident=ident,
+                    buffers=buffers, track=track, header=header, metadata=metadata)
```
takluyver: Is it more efficient to use the `msg` that is already prepared, rather than potentially making session build it again? i.e. call this as `send(stream, msg, ident=ident, buffers=buffers, track=track)`?

takluyver:
Other than that very minor point, I think this looks good, and probably the best course is to land it soon so that there's time to discover any bugs in the new implementation.

IPython/html/widgets/widget_output.py
jdfreder: Good point!

jdfreder:
@takluyver your comment is addressed.

takluyver:
OK, then I will merge this tomorrow unless someone wants to do some more pre-merge review.

PR #7526, opened 2015-01-21
Explicitly size codemirror editor in Edit app

### LINKED ISSUE #7298: Edit page codemirror scroll
We can't use the autosize feature of codemirror because it leads to extremely inefficient rendering.

jdfreder:
closes #7298

@ellisonbg

### DIFF
--- a/IPython/html/static/edit/js/main.js
+++ b/IPython/html/static/edit/js/main.js
@@ -93,5 +93,5 @@ require([
     window.onresize = _handle_resize;
 
     // On document ready, resize codemirror.
-    $(_handle_resize);
+    $(document).ready(_handle_resize);
 });
--- a/IPython/html/static/edit/js/main.js
+++ b/IPython/html/static/edit/js/main.js
@@ -2,6 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 require([
+    'jquery',
     'base/js/namespace',
     'base/js/utils',
     'base/js/page',
@@ -14,6 +15,7 @@ require([
     'edit/js/notificationarea',
     'custom/custom',
 ], function(
+    $,
     IPython,
     utils,
     page,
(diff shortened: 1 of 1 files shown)


IPython/html/static/edit/js/main.js
```diff
@@ -75,4 +77,14 @@ require([
         }
     };
 
+    // Make sure the codemirror editor is sized appropriatley.
+    var _handle_resize = function() {
+        var header = $('#header');
+        var padding = header.outerHeight(true) - header.innerHeight();
+        $('div.CodeMirror').height(window.innerHeight - header.height() - 2*padding);
```
minrk: Why is it 2x?

IPython/html/static/edit/js/main.js
```diff
@@ -75,4 +77,14 @@ require([
...
+    };
+    window.onresize = _handle_resize;
+
+    // On document ready, resize codemirror.
+    $(_handle_resize);
```
takluyver: I hadn't even realised that the jQuery constructor took a function, but it's there [in the docs](http://api.jquery.com/jQuery/#jQuery3)...

minrk: That's doing one too many things (or closer to 9 too many things) with one function. Sheesh.

IPython/html/static/edit/js/main.js
jdfreder: Top & bottom

jdfreder:
@minrk I commented the sizing logic.  It should be easier to understand now.

IPython/html/static/edit/js/main.js
takluyver: So this uses the padding from the header as the padding for the Codemirror widget? That's probably worth an explanatory comment if so, because it's not obvious why that would happen.

IPython/html/static/edit/js/main.js
jdfreder: Yeah, admittedly the first time I saw the JQuery constructor used this way I was confused too.  I'm probably just passing that confusion along by using it like that again, but it's ~15 or so less characters to type ;)

jdfreder: But seriously, if it would help I could use another method like `document.onready`.  I figured the comment was enough though...

minrk: I think document.ready is better.

jdfreder: Ok should be changed now

PR #7530, opened 2015-01-22
Bump less to @<3.0 so setup.py css works

### LINKED ISSUE #7529: "lessc too old" in Dockerfile
Maybe it's not the intended purpose, but I tried using the Dockerfile in the root of the project to setup a containerized dev environment using boot2docker on my Mac. The image builds fine, but I get the following trying to rebuild the CSS:

```
root@f48642bc1719:/srv/ipython# python setup.py css
running css
error: lessc too old: 1.7.5 < 2.0 (npm install -g 'less@<3.0')
```

parente:
closes #7529

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN apt-get update && apt-get install -y -q \
     npm
 
 # In order to build from source, need less
-RUN npm install -g less@1.7.5
+RUN npm install -g 'less@<3.0'
 
 RUN pip install invoke
 


Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN apt-get update && apt-get install -y -q \
     npm
 
 # In order to build from source, need less
-RUN npm install -g less@1.7.5
+RUN npm install -g less@2.2.0
```
minrk: should be `'less@<3.0'`

parente: Yea, oscillated between that and fixed version. Will switch.

parente:
Separate but related topic: would it be helpful to have doc somewhere (README? Wiki?) about how to use boot2docker + a built image from that Dockerfile as a dev environment? Kinda nice to keep it contained and runnable in a few commands.

Carreau:
> Separate but related topic: would it be helpful to have doc somewhere (README? Wiki?) about how to use boot2docker + a built image from that Dockerfile as a dev environment? Kinda nice to keep it contained and runnable in a few commands.

Yes, sure wiki is open if you like to add a page. 

I guess if we provide a 3.x image one it's released we might have a link to wiki in main docs.

rgbkrk:
Thanks @parente!

Issue #6939, opened 2014-11-14
Specify where install_nbextension installs extensions


jasongrout:
It would be great from the packaging standpoint if we could specify where nbextensions were installed.

```
    files : list(paths or URLs)
         One or more paths or URLs to existing files directories to install.
         These will be installed with their base name, so '/path/to/foo'
        will install to 'nbextensions/foo'.
```
It would be great if we could give a dictionary of paths/URLs.  For example, `{'myext1': '/path/to/first/extension', 'myotherext': '/path/to/next/extension'}` would install the first extension to `nbextensions/myext1` and the second extension to `nbextensions/myotherext`, regardless of the basename of the paths.

This solves an awkward directory structure problem in distributing extensions.  Currently, if you want to distribute the javascript in a python package, it makes sense to store it inside the package, often with the same name as the package.  If we have to have the basename match the package name (so that the nbextension is in `nbextensions/foo` for a package `foo`, then a directory structure like `foo/nbextension/foo/*` is awkward.  It would be nicer if the directory structure in the package was `foo/nbextension/*`, with `nbextensions/foo` pointing to `foo/nbextension` in the package, and you could do that with the above change by doing something like `install_nbextension({'foo': os.path.join(__file__, 'nbextension'})`.

jasongrout:
This was implemented in #7537.

PR #7537, opened 2015-01-22
Configurable destinations for nbextensions.

jasongrout:
This allows for a directory in a package to be called 'nbextensions', but be installed to an nbextension named after the package.  This addresses #6939.

### DIFF
--- a/IPython/html/nbextensions.py
+++ b/IPython/html/nbextensions.py
@@ -11,6 +11,7 @@ import shutil
 import sys
 import tarfile
 import zipfile
+import uuid
 from os.path import basename, join as pjoin
 
 # Deferred imports
@@ -117,10 +118,11 @@ def install_nbextension(files, overwrite=False, symlink=False, user=False, prefi
     Parameters
     ----------
     
-    files : list(paths or URLs)
+    files : list(paths or URLs) or dict(install_name: path or URL)
         One or more paths or URLs to existing files directories to install.
-        These will be installed with their base name, so '/path/to/foo'
-        will install to 'nbextensions/foo'.
+        If given as a list, these will be installed with their base name, so '/path/to/foo'
+        will install to 'nbextensions/foo'.  If given as a dict, such as {'bar': '/path/to/foo'},
+        then '/path/to/foo' will install to 'nbextensions/bar'.
         Archives (zip or tarballs) will be extracted into the nbextensions directory.
     overwrite : bool [default: False]
         If True, always install the files, regardless of what may already be installed.
@@ -158,8 +160,20 @@ def install_nbextension(files, overwrite=False, symlink=False, user=False, prefi
     if isinstance(files, string_types):
         # one file given, turn it into a list
         files = [files]
+    if isinstance(files, (list,tuple)):
+        # list given, turn into dict
+        _files = {}
+        for path in map(cast_unicode_py2, files):
+            if path.startswith(('https://', 'http://')):
+                destination = urlparse(path).path.split('/')[-1]
+            elif path.endswith('.zip') or _safe_is_tarfile(path):
+                destination = str(uuid.uuid4()) # ignored for archives
+            else:
+                destination = basename(path)
+            _files[destination] = path
+        files = _files
     
-    for path in map(cast_unicode_py2, files):
+    for dest_basename,path in (map(cast_unicode_py2, item) for item in files.items()):
         
         if path.startswith(('https://', 'http://')):
             if symlink:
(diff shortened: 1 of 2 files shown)


takluyver:
LGTM, but I'd like to see a test in `IPython.html.tests.test_nbextensions`.

jasongrout:
There are a few bugs that I'm working through.  I'll fix those and put a few tests in.

jasongrout:
Did you know that install_nbextension actually copies the files over every single time on python 2.7.9 on a modern linux?  I think it's related to this bug: http://bugs.python.org/issue12904.  Basically, getting the times gets them in nanosecond precision, but setting them only sets them in microsecond precision.  Here is an example, `s` is a filename.
```python
In [43]: oldtimes=(os.stat(s).st_atime, os.stat(s).st_mtime)

In [44]: os.utime(s, (oldtimes))

In [45]: newtimes=(os.stat(s).st_atime, os.stat(s).st_mtime)

In [46]: oldtimes==newtimes  
Out[46]: False

In [47]: oldtimes[0]-newtimes[0]
Out[47]: 9.5367431640625e-07

In [48]: oldtimes[1]-newtimes[1]
Out[48]: 9.5367431640625e-07
```

jasongrout:
I'm going to add a fudge factor to checking the mtime to get around this issue.

jasongrout:
@takluyver - I got the last doctest working because I changed the names to not use weird unicode characters.  I'm not sure what is going wrong on python 2 with the unicode characters (see https://travis-ci.org/ipython/ipython/jobs/47975940#L340).  Do you happen to see what is going on and how to fix the test (or the code) in the case where we have a weird unicode filename?

takluyver:
I suspect that you just need the filename to be a unicode literal. i.e. `u'ƒile.js'` instead of `'ƒile.js'`.

jasongrout:
Ah, right, of course.  We'll see if Travis is happy now...

jasongrout:
Looks like things are good.  @takluyver, I think this is ready for review.

I noticed that there is a bit of code complexity because you can specify a list of extensions.  My guess is that the vast majority of the time, you're only installing one extension.  In the rare exception that you'd be installing multiple extensions, I don't think it's that onerous to have people call the function multiple times.  Do you mind if I simplify the function to just handle installing one extension?

jasongrout:
(simplify to handle just one extension on another PR.  I think this PR is feature-frozen at this point)

takluyver:
Thanks, will review later this afternoon.

I'd say you're free to make a PR simplifying this function to only install one extension at a time, then we'll discuss whether we prefer the extra functionality or the technical simplicity.

IPython/html/nbextensions.py
```diff
@@ -74,8 +77,10 @@ def _maybe_copy(src, dest, verbose=1):
     if _should_copy(src, dest, verbose):
         if verbose >= 1:
             print("copying %s -> %s" % (src, dest))
-        shutil.copy2(src, dest)
-
+        try:
+            shutil.copy2(src, dest)
+        except IOError as e:
+            print(str(e), file=sys.stderr)
```
takluyver: Why is this error being printed but suppressed?

takluyver:
Other than that, :+1:

IPython/html/nbextensions.py
jasongrout: It is printed for informational purposes, but suppressed to try to get as much of the extension as possible copied over.  Would you rather that we didn't suppress the error?

jasongrout:
As for the install-single-extension PR, why don't we decide which we'd rather have before I spend the time doing it.

IPython/html/nbextensions.py
takluyver: Yes. Exceptions tell the calling code that something is wrong, which may be able to do something about it. Additionally, if copying one file has failed:

1. It's very likely that copying the next one will fail, so for a large extension, the user will see a long line of intimidating errors.
2. It's very likely that the extension won't work correctly, so there's no use pretending to the caller that everything's fine.

takluyver:
Don't feel you have to write it, but if we're weighing up the relative merits of simpler code versus more features, and we already have the more featureful version implemented, then it's easier to make the case for simplification if it's clear how much simpler the code would be.

IPython/html/nbextensions.py
jasongrout: Fair enough.  Those weren't the case for my example extension, but I agree that those are the more likely cases.  I'll take out the error trapping.

jasongrout:
@takluyver: I fixed the trapping error issue - can you take a look at this again?

takluyver:
Looks OK to me; just re-running a Travis test group that failed, but I'm pretty sure it's unrelated.

PR #7538, opened 2015-01-22
Setting union traitlet metadata correctly

SylvainCorlay:
Rebinding union traitlet metadata to the one of the successfully validated trait type. This is useful for rebinding to the right version of `to_json` and `from_json` for example.

### DIFF
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -1069,7 +1069,9 @@ class Union(TraitType):
     def validate(self, obj, value):
         for trait_type in self.trait_types:
             try:
-                return trait_type._validate(obj, value)
+                v = trait_type._validate(obj, value)
+                self._metadata = trait_type._metadata
+                return v
             except TraitError:
                 continue
         self.error(obj, value)


IPython/utils/traitlets.py
```diff
@@ -1069,7 +1069,9 @@ def instance_init(self, obj):
     def validate(self, obj, value):
         for trait_type in self.trait_types:
             try:
-                return trait_type._validate(obj, value)
+                v = trait_type._validate(obj, value)
+                self._metadata = v._metadata
```
minrk: I don't think this is correct

minrk: it sets metadata from *each* of the union traits, so Union(A, B) results in different metadata than Union(B, A).

minrk: `trait_type._metadata` here

SylvainCorlay: sure, wrong merge.

SylvainCorlay:
@minrk done.

PR #7534, opened 2015-01-22
Alternative fix for notebook help links

Carreau:
Alternative to #7533. 

keep the `href`s, but explicitely prevent default on all needed links.

Closes #7532

PR #7533, opened 2015-01-22
do not prevent defualt, but remove anchor to #

Carreau:
closes #7532 

Since html 5 `<a>` without href is valid. 

Alternate fix in preparation.

PR #7541, opened 2015-01-22
remove preventDefault on menubar

### LINKED ISSUE #7532: Notebook Help links don't work
In current master, none of the Help links in the Help menu that open URLs are working. For example, clicking on menu -> Help -> Markdown doesn't go to URL (nothing happens).

I'm on Chrome on Linux.

minrk:
fixes help links

based on #7517
closes #7532

It was a workaround for problems caused by the fixed header, and no longer needed after #7517.

supersedes #7533 and #7534

### DIFF
--- a/IPython/html/static/base/less/page.less
+++ b/IPython/html/static/base/less/page.less
@@ -14,6 +14,7 @@ body {
     right: 0px;
     top: 0px;
     bottom: 0px;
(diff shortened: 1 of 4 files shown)


takluyver:
#7517 is now merged. The actual change here is very simple, so merging now.

PR #7542, opened 2015-01-22
add eurosym package to define \euro symbol

### LINKED ISSUE #7511: PDF generation from notebook fails if euro symbol (€) is present
Hi, I found that if the euro currency symbol is in the markup of a Notebook, "Download as PDF" and `ipython nbconvert --to=pdf` fail, since LaTeX doesn't know `\euro`. The LaTeX headers need an appropriate `\usepackage`, perhaps `\usepackage{eurosym}` (which works but some users find ugly).

amcdawes:
fixes #7511


Carreau:
+1. 

Waiting on travis to be green.

Thanks a lot for helping !

PR #7551, opened 2015-01-23
fix left-margin on xxs screen

Carreau:
On xxs screen spacing between border and codemirror was 0 for left. 
this fixes that by making 5px all around (and reducing prompt with a bit more (by 1ex instead of 5px), which still does not prevent 3 digit prompt. 

capture d ecran 2015-01-23 a 19 08 06


Carreau:
oops, commited submodule. Will fix.

PR #7554, opened 2015-01-23
Use Output widget with Interact

### LINKED ISSUE #6534: Same widget does not update all interact.

jdfreder:
closes #6534
alternative to #7482

Ping @jasongrout who was also interested in this solution.

Also this fixes a small bug in which`get_ipython` is not defined in the context where the Output widget is `__enter__`ed.

### DIFF
--- a/IPython/html/widgets/tests/test_interaction.py
+++ b/IPython/html/widgets/tests/test_interaction.py
@@ -80,7 +80,8 @@ def check_widgets(container, **to_check):
     # build a widget dictionary, so it matches
     widgets = {}
     for w in container.children:
-        widgets[w.description] = w
+        if hasattr(w, 'description'):
+            widgets[w.description] = w
     
     for key, d in to_check.items():
         nt.assert_in(key, widgets)
@@ -138,7 +139,7 @@ def test_single_value_float():
 def test_single_value_int():
     for a in (1, 5, -3):
         c = interactive(f, a=a)
-        nt.assert_equal(len(c.children), 1)
+        nt.assert_equal(len(c.children), 2)
         w = c.children[0]
         check_widget(w,
             cls=widgets.IntSlider,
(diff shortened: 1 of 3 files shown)


IPython/html/widgets/widget_output.py
```diff
@@ -34,6 +34,10 @@ class Output(DOMWidget):
             print('prints to output widget')"""
     _view_name = Unicode('OutputView', sync=True)
 
+    def __init__(self, *args, **kwargs):
+        super(Output, self).__init__(*args, **kwargs)
+        self._ip = get_ipython()
```
minrk: This file is missing `from IPython import get_ipython`

minrk:
Travis failures look real. Missing `description`?

jdfreder:
@minrk the test problem *should* be fixed in the latest commit.  The interact tests erroneously assumed every widget in the interact-produced container had a description property.  Standing by to see what Travis thinks...

minrk:
Tests are failing with a few off-by-one errors, but it looks like the tests need updating rather than an actual bug, since an extra widget is added to the collection.

jdfreder:
Looks like this is passing now.

IPython/html/widgets/tests/test_interaction.py
```diff
@@ -80,7 +80,8 @@ def check_widgets(container, **to_check):
     # build a widget dictionary, so it matches
     widgets = {}
     for w in container.children:
-        widgets[w.description] = w
+        if hasattr(w, 'description'):
+            widgets[w.description] = w
```
minrk: This means that widgets without a description will be excluded from the dict. Is there a problem with doing that?

jdfreder: It shouldn't be a problem, widgets without descriptions cause this code (without the check) to barf, so they can't be included as-is.

minrk: ok

PR #7556, opened 2015-01-23
Issue 7548: Better handling of line magics in nbconvert

abalkin:


IPython/nbconvert/utils/lexers.py
```diff
@@ -49,10 +49,14 @@
 line_re = re.compile('.*?\n')
 
 ipython_tokens = [
-  (r'(\%+)(\w+)\s+(\.*)(\n)', bygroups(Operator, Keyword,
+  (r'(%)(sx|sc|system)(.*)(\n)', bygroups(Operator, Keyword,
                                        using(BashLexer), Text)),
-  (r'(\%+)(\w+)\b', bygroups(Operator, Keyword)),
```
takluyver: I think the `%+` should be preserved on this line, to handle cell magics as well.

takluyver:
Other than that, I think this is good. I'm not especially familiar with pygments lexer code, but I don't think we have anyone who is.

minrk:
Thanks. We are in bugfix-only release mode, so holding all new features until after we finish shipping 3.0.

abalkin:
> We are in bugfix-only release mode, 

@minrk - is there any part that you would consider a bug-fix?  I think a red error box around `!` and `$` in an html-converted notebook would qualify as a bug.

Consider this screenshot:

screen shot 2015-01-23 at 7 02 15 pm

takluyver:
If that's readily reproducible, I'd be happy with accepting a simple fix for it.

takluyver:
And I can easily reproduce it. Marking this as potentially 3.0 again, for discussion.

abalkin:
> If that's readily reproducible, ..

To reproduce - just enter 

```
x = !echo $HOME
```

in a cell and  select File > Print Preview from the menu.

IPython/nbconvert/utils/lexers.py
abalkin: I believe cell magics are handled at a higher level in [filters](https://github.com/ipython/ipython/blob/3.0.0-dev/IPython/nbconvert/filters/highlight.py#L118) and we don't get to this code if the cell starts with `%%`, but I don't see any harm in being defensive here.  

The original regexp is plain wrong.  The backslash is redundant and it matches any number of `%`'s.  What would you say about `(%%?)`?

takluyver: `(%%?)` would be fine, there's no need to handle more than two %. But I think we do need to handle it here:

1. These lexers are not only used in highlighting notebooks. Despite being in nbconvert, this module is imported by `IPython.sphinxext.ipython_console_highlighting`, which is used for IPython syntax in Sphinx docs.
2. I don't think the code you linked to catches all cell magics, only a subset that have a highlighting language specifically associated with them: https://github.com/ipython/ipython/blob/3.0.0-dev/IPython/nbconvert/preprocessors/highlightmagics.py#L39

abalkin:
Here are some more before and after demos:

|master|PR|
|---|---|
|before|after|

IPython/nbconvert/utils/lexers.py
abalkin: > These lexers are not only used in highlighting notebooks.

Please take a look at #7558.  I will make it much easier to test (and use) ipython lexers if they are exposed as pygments entry points.  (For example, pygments comes with a testcase formatter that converts plain code examples to unittests.)

abalkin: I think cell magics will require a separate rule because they should extend the match to the end of the cell and not terminate on `\n`.  I'll add some test cases for that.

abalkin:
I think this now behaves the way I want:

screen shot 2015-01-23 at 10 29 25 pm

Please let me know if you have any more comments and if not I'll rebase this into one commit.

dsblank:
Keep in mind other kernels that don't support magics, or that use different syntax to invoke their own versions of them. It isn't clear to me how this will affect other kernels/languages, if at all. When/where is `ipython_tokens` used?

abalkin:
@dsblank - this patch should not affect other languages kernels.  First, it is limited to nbconvert, so it will only affect html, pdf, slides, etc. output generated by the backend.  Other kernels will not see this output.  Second, if `language` argument is given to the highlighter and it is different from `ipython`, IPythonLexer is not used.

> When/where is `ipython_tokens` used?

They are used by IPythonLexer (and its companion IPython3Lexer) only.  It should really be a class variable, but for some reason IPythonLexer and IPython3Lexer are created dynamically without a class statement.

dsblank:
@abalkin Be aware that there are hybrid kernels (wrapperkernels) that do use the IPython infrastructure even those languages may shell-out to a sub-process (eg, Octave). All kernels would still use the nbconvert utilities for notebook conversions to html, pdf, etc. right?

But if what you say about `language` is correct, then this isn't an issue because none would use `ipython`.

abalkin:
>  But if what you say about language is correct, ..

If it is not, I would say there is bug elsewhere.  The job of the IPythonLexer is to parse IPython code, not the code destined for non-ipython kernels.  The docstring for `build_ipy_lexer` says:

>    The lexer inherits from an appropriate Python lexer and then adds
    information about IPython specific keywords (i.e. magic commands,
    shell commands, etc.)

This patch simply fixes the bugs in the lexer logic.  It does not introduce new features.  The lexer already treats `%`, and `!` specially and otherwise expects python syntax.

Carreau:
We **might** have assumption (python language == IPython kernel) and that would definitively be a bug. 
Though I doubt that for now there are python kernel that would use a similar syntax for magic and would not want this highlighting.

abalkin:
> I doubt that for now there are python kernel that would use a similar syntax for magic and would not want this highlighting.

In the long run, I would like to replace the regex-based lexer with the tokenize-based one similar to what is used in core inputtransformer.  This should allow sharing the logic between core and nbconvert.  If this approach is successful alternative kernels can do the same and guarantee that highlighting follows the same rules as evaluation.

takluyver:
Yes, this code is only used when the ipython pygments lexer is in use,
which it won't be for notebooks created with other kernels.

Carreau:
This seem like a net improvement, and we have few chances of having complain and/or user testing without merging. 

So +1, thanks a lot, this will make a lot of people happy on nbviewer !

Merging.

PR #7557, opened 2015-01-23
don't use flexbox to size `#site`

minrk:
The approach in #7517 didn't work on Safari (as in the entire page was invisible).

This is the same mechanism used to size the CodeMirror div on the edit page, and should work more reliably than the flexbox approach.

Adds translateZ(0) to avoid repaints of `#site` on scroll.

### DIFF
--- a/IPython/html/static/base/js/page.js
+++ b/IPython/html/static/base/js/page.js
@@ -10,14 +10,20 @@ define([
 
     var Page = function () {
         this.bind_events();
-
-        // When the page is ready, resize the header.
-        var that = this;
-        $(function() { that._resize_header(); });
     };
 
     Page.prototype.bind_events = function () {
-        events.on('resize-header.Page', $.proxy(this._resize_header, this));
+        // resize site on:
+        // - window resize
+        // - header change
+        // - page load
+        var _handle_resize = $.proxy(this._resize_site, this);
+        
+        $(window).resize(_handle_resize);
+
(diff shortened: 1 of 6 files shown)


minrk:
ping @jdfreder for review and sanity check. I'd like to merge quickly, since it's a pretty bad failure.

Thanks to the translateZ(0) bit, this now has better repaint characteristics than any iteration so far, even the fixed-header. Only the scrollbar itself is repainted on scroll.

michaelaye:
works fine here as well.

Carreau:
So we get back to initial implementation (almost) :-( , 
But if we have to.

jdfreder:
> So we get back to initial implementation (almost) :-( , 
> But if we have to.

Well that was 50 SLOC, this is 8 SLOC, so this is much simpler and helps with the repaint issues.  @minrk does this butcher the ctrl-p print support? :(

minrk:
>  @minrk does this butcher the ctrl-p print support? :(

100%, I'll work on another fix. I hate this stuff so much...

jasongrout:
FYI, sometimes we have fixed position elements that are descendants of #site for stuff we are doing.  Because #site now has a translateZ(0), #site acts as "a containing block for fixed positioned descendants." (http://www.w3.org/TR/css3-transforms/#transform-property). This means that fixed-position elements inside of #site no longer are positioned relative to the viewport, but instead relative to the #site element.

PR #7562, opened 2015-01-23
Indicate that terminals are unavailable

### LINKED ISSUE #7019: Make terminals more discoverable

minrk:
reissue of #7552 that got borked.  Waiting for #7505, closes closes #7019.


minrk:
This should have waited for #7505. Now that one needs a rebase.

jdfreder:
Not a problem.  I think the only reason it is conflicting with that is because I was messing with terminals tab visibility too.

PR #7563, opened 2015-01-23
document restarting, dead kernel status messages

### LINKED ISSUE #6603: Document kernel execution state of restarting/dead
In the message spec, the kernel can be 'busy', 'idle', or 'starting', but it can *also* be 'restarting' or 'dead' -- these states aren't part of the message spec because the kernel can't say it is dead if it is dead, for example, but they should probably be documented somewhere (because those states are currently not mentioned in the message spec documentation).

minrk:
that are only sent over the websocket channel, not from the kernel.

closes #6603


takluyver:
The changes from #7552 are showing up in the diff for some reason, but that was already merged. The actual change here looks fine.

PR #7565, opened 2015-01-24
enable keyboard manager when codemirror is focused

minrk:
Clicking on an input area should always re-enable the keyboard manager. This should always be true, and mitigate focus-related keyboard bugs, such as #7555.

It shouldn't ever be necessary, but some bugs (ours or browsers') can cause problems, and when there are weird keyboard-related issues, clicking a cell is often my first instinct.


takluyver:
I've confirmed that the mitigation works here, and I agree that this is both a good idea in general, and sufficient if we can't come up with a proper fix for #7555. Clicking in a cell's input area is also my first thought when I find keyboard shortcuts not working.

PR #7575, opened 2015-01-24
allow requesting contents without body

minrk:
adds `?content=0` to only fetch the metadata in the model.

The REST API now exposes all arguments to ContentsManager.get.

I'll open a separate PR that uses this to check if a file has been modified to avoid clobber on save.

### DIFF
--- a/IPython/html/services/contents/handlers.py
+++ b/IPython/html/services/contents/handlers.py
@@ -117,13 +117,19 @@ class ContentsHandler(IPythonHandler):
         format = self.get_query_argument('format', default=None)
         if format not in {None, 'text', 'base64'}:
             raise web.HTTPError(400, u'Format %r is invalid' % format)
-
-        model = yield gen.maybe_future(self.contents_manager.get(path=path, type=type, format=format))
-        if model['type'] == 'directory':
+        content = self.get_query_argument('content', default='1')
+        if content not in {'0', '1'}:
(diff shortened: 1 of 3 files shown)


Carreau:
is `?content=0` rest compliant, or is there a http methods or header that would be better?

minrk:
Are you asking about how to express boolean params, or whether params are okay at all? Params are definitely fine, we are already using them for other args on the same target, and many other REST services do the same. I don't think we should use a header for one arg and params for the rest.

If it's about 0/1 vs true,false, I can switch it to true/false to be clearer.

Carreau:
No, it was just query param vs something else. If you say it's the way to do it, then fine.

takluyver:
Min and I discussed doing this with a HEAD request, which is semantically similar in that it only asks for metadata, not content. However, the response to a HEAD request typically (and I assume by specification) conveys metadata as HTTP headers, and doesn't have content. Here we're supplying metadata as JSON in the HTTP content, so we decided that a GET request with a parameter was more appropriate.

Carreau:
> Min and I discussed doing this with a HEAD request, which is semantically similar in that it only asks for metadata, not content. However, the response to a HEAD request typically (and I assume by specification) conveys metadata as HTTP headers, and doesn't have content. Here we're supplying metadata as JSON in the HTTP content, so we decided that a GET request with a parameter was more appropriate.

Ok, merging then.

PR #7580, opened 2015-01-25
indicate unsaved changes in text editor

Carreau:
Add an indicator near the file name:
capture d ecran 2015-01-25 a 14 27 40

Change to a √ when all is saved.

### DIFF
--- a/IPython/html/static/edit/js/editor.js
+++ b/IPython/html/static/edit/js/editor.js
@@ -25,12 +25,16 @@ function($,
     var Editor = function(selector, options) {
         var that = this;
         this.selector = selector;
+        this.clean = false;
         this.contents = options.contents;
         this.events = options.events;
         this.base_url = options.base_url;
         this.file_path = options.file_path;
         this.config = options.config;
         this.codemirror = new CodeMirror($(this.selector)[0]);
+        this.codemirror.on('changes', function(cm, changes){
+            that._clean_state();
+        });
         this.generation = -1;
         
         // It appears we have to set commands on the CodeMirror class, not the
@@ -49,7 +53,11 @@ function($,
             );
             that._set_codemirror_options(cmopts);
             that.events.trigger('config_changed.Editor', {config: that.config});
+            that._clean_state();
         });
+        this.clean_sel = $('<div/>');
+        $('.last_modified').before(this.clean_sel);
+        this.clean_sel.addClass('dirty-indicator-dirty');
     };
     
     // default CodeMirror options
@@ -78,6 +86,7 @@ function($,
                 that.save_enabled = true;
                 that.generation = cm.changeGeneration();
                 that.events.trigger("file_loaded.Editor", model);
+                that._clean_state();
             }).catch(
             function(error) {
                 that.events.trigger("file_load_failed.Editor", error);
@@ -147,6 +156,7 @@ function($,
                 that.file_path = model.path;
                 that.events.trigger('file_renamed.Editor', model);
                 that._set_mode_for_model(model);
+                that._clean_state();
             }
         );
     };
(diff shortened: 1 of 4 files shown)


dsblank:
Notification of unsaved changes would be handy too on the notebook, right?

Carreau:
The dirty state of the notebook is a little more complicated, people tend to disagree on what the dirty state is, like EG, when a text has been changed, but the codecell no re-run. Is it dirty or not ? 

I don't remember the all discussion, but in the end we decided not to do it in the notebook for now. 
On a text-file its unambiguous as you don't have an extra state space.

damianavila:
I like the idea... not sure about the icon... I mean, I don't know if it is necessary...

Carreau:
Well technically the icon the the only visual indication.

IPython/html/static/edit/js/editor.js
```diff
@@ -49,7 +49,11 @@ function($,
             );
             that._set_codemirror_options(cmopts);
             that.events.trigger('config_changed.Editor', {config: that.config});
+            that._clean_state();
         });
+        this.clean_sel = $('<div/>')
```
minrk: semicolon

IPython/html/static/edit/js/editor.js
```diff
@@ -49,7 +49,11 @@ function($,
...
+        $('.last_modified').after(this.clean_sel);
+        this.clean_sel.addClass('fa fa-check')
```
minrk: semicolon

IPython/html/static/edit/js/editor.js
```diff
@@ -164,8 +170,18 @@ function($,
...
         return this.contents.save(this.file_path, model).then(function(data) {
             that.events.trigger("file_saved.Editor", data);
+            that._clean_state();
         });
     };
+
+    Editor.prototype._clean_state = function(){
+        var clean = this.codemirror.isClean(this.generation);
+        if(clean){
+            this.clean_sel.attr('class','fa fa-check').attr('title','all changes saved');
```
minrk: I would use "up to date" or "no changes to save"

IPython/html/static/edit/js/editor.js
```diff
@@ -164,8 +170,18 @@ function($,
...
+            that._clean_state();
         });
     };
+
+    Editor.prototype._clean_state = function(){
+        var clean = this.codemirror.isClean(this.generation);
+        if(clean){
+            this.clean_sel.attr('class','fa fa-check').attr('title','all changes saved');    
+        } else {
+            this.clean_sel.attr('class','fa fa-remove').attr('title','unsave changes');
```
minrk: "unsaved changes"

IPython/html/static/edit/js/editor.js
```diff
@@ -176,6 +192,10 @@ function($,
             }
             codemirror.setOption(opt, value);
         });
+        var that = this;
+        codemirror.on('changes', function(cm, cahnges){
```
minrk: `s/cahnges/changes/`, or leave out ignored args altogether.

minrk: Also, this is probably not the right place to set this, since this method is called on any update to config (e.g. toggling line numbers). This should probably be called when the CodeMirror object is instantiated.

Carreau:
@minrk and on the idea in general ? Design, position of x/√ ?

damianavila:
>Well technically the icon the the only visual indication.

Well... rethinking this one... maybe is the position I did not like too much... it too close to the text.

dsblank:
If you put the indicator before the saved text, then it won't be moving all over the line as the length of the text changes. I have used `*` (like emacs) before in such a position to indicate dirty, and nothing when saved.

Carreau:
> If you put the indicator before the saved text, then it won't be moving all over the line as the length of the text changes. I have used * (like emacs) before in such a position to indicate dirty, and nothing when saved.

Seem to be a good idea. I'll try. the `*` could potentially be used in tab title.

minrk:
@Carreau general idea makes sense, I don't think x is a good indicator of unsaved changes, but √ is fine. I'd give @dsblank's idea a try. Putting it in the document title seems like a good idea as well.

Carreau:
Updated with feedback on design, and cleanup implementation.
capture d ecran 2015-01-26 a 22 05 40
capture d ecran 2015-01-26 a 22 05 53

takluyver:
@Carreau this has a merge conflict now, so it will need a rebase.

Carreau:
rebased and squashed.

Issue #7582, opened 2015-01-25
Horizontal scroll not working


damianavila:
In Chrome Versión 39.0.2171.65 Ubuntu 14.04 (64-bit), hscroll is hidden no matter what...

![2015-01-25-130202_1366x768_scrot](https://cloud.githubusercontent.com/assets/1640669/5892076/64e9a76a-a492-11e4-84dd-d5ffd982e676.png)

I think it is related with this hack: https://github.com/ipython/ipython/blob/master/IPython/html/static/notebook/less/codemirror.less#L53-L58

minrk:
Closed by #7583

PR #7583, opened 2015-01-25
Fix the hidden x scrolling.

damianavila:
This PR removing the lines showed in #7582 seems to fix the issue...

2015-01-25-130446_1366x768_scrot

I think this is inside the 3.0 milestone...


Carreau:
Seeing the comment in the text file and that we want to remove the hack, I'm going to merge to have user testing.

Thanks @damianavila

Issue #7548, opened 2015-01-23
Syntax highlighting in presence of line magics


abalkin:
IPython notebook knows how to switch language for (some) cell magic directives, but line magics and the ! directive seem to be completely ignored.

minrk:
That's correct. Only cell magics trigger language switching. There isn't any mode switching for lines. Disabling Python highlighting on line magics should probably be added to the IPython codemirror mode.

abalkin:
The issue is the same on the "server" side.  This should probably be handled in the IPythonLexer.

abalkin:
The best solution would be to allow line/cell magic objects to provide the necessary pieces such as a pygments lexer and a codemirror mode.

takluyver:
Possibly, but the rendering code doesn't load the magic commands, so I'm
not sure how the information would be conveyed.

abalkin:
I was thinking of just providing the name of pygments lexer which in most cases will be the same as the matching cm mode, but for better interoperability the mime types are probably better.  A magic object can define a mimetypes list and other IPython components may pick the best match.

Both pygments and codemirror specify mimetypes, but since those are non-standard they don't always match.  For example, pygments defines bash lexer mimetypes as

```
    mimetypes = ['application/x-sh', 'application/x-shellscript']
``` 

but codemirror uses text/x-sh.  The shell magic can specify all three and highlighters will pick the one they understand.

abalkin:
Interestingly, I was only able to find one IANA-registered MIME type for a computer language - [application/sql](http://www.iana.org/assignments/media-types/application/sql) for SQL.  And neither pygments nor cm uses it.

Pygments:

```python
    mimetypes = ['text/x-sql']
```

Codemirror: text/x-sql, text/x-mysql, text/x-mariadb, text/x-cassandra, text/x-plsql, text/x-mssql, text/x-hive.

takluyver:
Yep, but the code rendering the notebook doesn't load the magic functions,
so we'd have to come up with some way to convey the highlighting
information to the frontend and store it in the notebook.

abalkin:
I am new to IPython development, so I would like to proceed in baby steps:

1. Write tests for the current behavior.
2. Implement no highlighting between the line magic operator and the end of the line.
3. Figure out how to convey magics' language information to the lexer.

takluyver:
If the current behaviour isn't desired, I'm not sure it's worth writing tests for it.

[...]

PR #7585, opened 2015-01-25
Additional fixes for the ipython lexer

abalkin:
Two more cases added for #7548:
1. Don't confuse != with a shell escape.
2. Recognize leasing ? or ?? as starting a help line.


takluyver:
Thanks, make sense.

PR #7586, opened 2015-01-25
re-hook notification area for editor.

### LINKED ISSUE #7569: Editor notification
I think we should add the notification area (at least for save) on the Editor. 
Maybe also a dirty indicator.

Carreau:
- re-order element to be the same as in notebook,
- rename events correctly.
- Add event on savign to show that save is in progress, that get
  dismissed on file save

Closes #7569

capture d ecran 2015-01-25 a 20 00 14

### DIFF
--- a/IPython/html/static/edit/js/notificationarea.js
+++ b/IPython/html/static/edit/js/notificationarea.js
@@ -19,9 +19,6 @@ define([
         var that = this;
         var savew = this.new_notification_widget('save');
         
-        this.events.on("file_saving.Editor", function() {
-            savew.set_message("Saving File...");
-        });
         this.events.on("file_saved.Editor", function() {
             savew.set_message("File saved", 2000);
         });
--- a/IPython/html/static/edit/js/notificationarea.js
+++ b/IPython/html/static/edit/js/notificationarea.js
@@ -17,10 +17,13 @@ define([
      */
     EditorNotificationArea.prototype.init_notification_widgets = function () {
         var that = this;
-        var enw = this.new_notification_widget('editor');
+        var savew = this.new_notification_widget('save');
         
-        this.events.on("save_succeeded.TextEditor", function() {
-            enw.set_message("File saved", 2000);
+        this.events.on("file_saving.Editor", function() {
+            savew.set_message("Saving File...");
+        });
+        this.events.on("file_saved.Editor", function() {
+            savew.set_message("File saved", 2000);
         });
     };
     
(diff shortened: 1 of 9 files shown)


IPython/html/static/edit/js/notificationarea.js
```diff
@@ -17,10 +17,13 @@ define([
      */
     EditorNotificationArea.prototype.init_notification_widgets = function () {
         var that = this;
-        var enw = this.new_notification_widget('editor');
+        var savew = this.new_notification_widget('save');
         
-        this.events.on("save_succeeded.TextEditor", function() {
-            enw.set_message("File saved", 2000);
+        this.events.on("file_saving.Editor", function() {
+            savew.set_message("Saving File...");
```
takluyver: Do we really need a notification for this? It seems more likely to cause an annoying flicker than to ever be useful.

takluyver:
When did 'Plain Text' get added to the editor toolbar? Does that ever show anything else? What's the rationale behind that? (*Please* don't say it's just to make it match the notebook...)

IPython/html/static/edit/js/notificationarea.js
Carreau: With G-Drive, the saving can take a few second, and I was like "No wont save, [re-press Cmd-S a few time]"

takluyver: Does it make any sense to be using our text editor with G-Drive anyway?

Carreau:
> When did 'Plain Text' get added to the editor toolbar? Does that ever show anything else? What's the rationale behind that? (Please don't say it's just to make it match the notebook...)

No, it show the mode of codemirror (here plain text), but it can say Python, LaTeX, Markdown...

> Does it make any sense to be using our text editor with G-Drive anyway?

Maybe not, but it does not mean it will be the only slow backend.

takluyver:
Ah, OK, codemirror mode makes sense.

I'd leave the 'saving' notification out for now, unless/until there's a way to delay a notification briefly without having it replace one that appears in the meantime (i.e. more sophisticated than setTimeout). Then we could have the 'saving' notification show up if it takes more than e.g. 250ms to finish the save.

Carreau:
> I'd leave the 'saving' notification out for now

ok


> unless/until there's a way to delay a notification briefly without having it replace one that appears in the meantime (i.e. more sophisticated than setTimeout). Then we could have the 'saving' notification show up if it takes more than e.g. 250ms to finish the save.

And then the second event triggers after 260 ms and you get the flashing for 10 ms ? 
I don't think it is solvable with out current understanding of causality.

Carreau:
Removed. I have left the event though.

Issue #7488, opened 2015-01-16
Markdown output is not sanitized by nbconvert


ssanderson:
In the live notebook, Markdown output is always sanitized before being displayed to the user. However, the static html produced by nbconvert is unsanitized.  This isn't an XSS vulnerability per se, since the rest of the rendered static html is also allowed to have javascript, but It seems odd that markdown renders differently in the different context.

It would also be nice if it were easy to configure nbconvert to spit out escaped markdown.

minrk:
What do you mean by "spit out escaped markdown"?

You can set the display_priority config to exclude markdown altogether. Since sanitization happens in js, I'm not sure how we can add sanitization to nbconvert without including a massive js library.

ssanderson:
By "spit out escaped markdown", I mean "render markdown with the same escaping/sanitization semantics as are applied with the live notebook".  

A more fancy version of this would be something like adding an HTMLExporter option to configure a whitelist of tags to render as HTML, with any other being removed and/or escaped.  (If this route were considered the issue probably wouldn't be Markdown-specific so much as it would be about making it possible to configure the escaping behavior of nbconvert). I've had some success playing with [Bleach](https://pypi.python.org/pypi/bleach) locally to accomplish this.  I can probably make this work with my own local subclass of `HTMLExporter`, but I suspect that other people might benefit from having this be configurable in the default distribution.

minrk:
We should probably consider adding Bleach in the future. Marking as 4.0.

ssanderson:
@minrk it turns out it's actually really straightforward to do this nicely with an nbconvert Preprocessor.  Would you guys be interested in a PR to add it to the existing preprocessor library?

Carreau:
> @minrk it turns out it's actually really straightforward to do this nicely with an nbconvert Preprocessor. Would you guys be interested in a PR to add it to the existing preprocessor library?

Sure, we can always review it, even if it it not merged, at least it is reviewed and easy to make standalone.

minrk:
I think it makes sense, but we can review it after we get 3.0 out the door.

ssanderson:
Proposal for this lives at #7595

Carreau:
Closing as this issue is not in IPython itself and if still problematic and relevant should be opened on the right repository.  This will allow to keep the number of opened issue on the IPython repo under control. 

Feel free to keep commenting or reopen if necessary.

Thanks.

PR #7595, opened 2015-01-26
DEV: Add preprocessor for sanitizing NBConvert outputs.

ssanderson:
Addresses https://github.com/ipython/ipython/issues/7488

### DIFF
--- a/IPython/nbconvert/preprocessors/sanitize.py
+++ b/IPython/nbconvert/preprocessors/sanitize.py
@@ -100,7 +100,8 @@ class SanitizeHTML(Preprocessor):
         runs `sanitize_html_tags` over 'text/html'.
         """
         for output in outputs:
-            if output['output_type'] != 'display_data':
+            # These are always ascii, so nothing to escape.
+            if output['output_type'] in ('stream', 'error'):
                 continue
             data = output.data
             to_remove = []
--- a/IPython/nbconvert/preprocessors/sanitize.py
+++ b/IPython/nbconvert/preprocessors/sanitize.py
@@ -1,0 +1,133 @@
+"""
+NBConvert Preprocessor for sanitizing HTML rendering of notebooks.
+"""
+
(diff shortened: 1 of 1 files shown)


ssanderson:
@minrk @Carreau this is my proposal for a sanitizing preprocessor for NBConvert.  Credit to @TimShawver for the initial concept and implementation.

minrk:
Thanks, I think this makes sense. We can investigate it further once we get 3.0 out the door.

Carreau:
Make sens also. 
Travis need to be updated to install bleach, and setup.py also. 
Or we make it a optional dependency.

takluyver:
@ssanderson are you still interested in this?

ssanderson:
@takluyver sure.  We're already running an implementation of this internally at quantopian, so I guess the question is whether it makes more sense to have this in the main repository (which would require taking on bleach as a dependency) or whether it should live in our repo, in which case the work here is probably just a link or two in the docs somewhere.

takluyver:
We do want to ensure nbconvert is easily extensible to add things like this, so if you're willing to be guinea pigs for that, +1 to leaving it in your repository.

minrk:
This is the only PR touching nbconvert. Are we going to give this a try as an example of custom behavior, rather than inside nbconvert?

Carreau:
+1 as standalone example/lib. Could even I guess be it's own package.

ssanderson:
I probably won't have time to prep our repo to be open sourced for at least a week or so.  As long as that's fine with you all I think having this as an external lib is reasonable.

Carreau:
We can also have a jupyter-contrib package which is throw in for many one time/one file scripts/extensions. We have that for for notebook extensions....

minrk:
@ssanderson thanks, and no rush. This PR can serve as an example if anyone needs it in the time between now and when you get the chance to set it up as a plugin.

PR #7596, opened 2015-01-26
Make Lorenz differential equations example compatible with Python 3

takluyver:
Thanks to @franktoffel for pointing this out in #7350


jenshnielsen:
There also seems to be a Lorentz/Lorenz typo in the doc string that might be worth correcting?

takluyver:
Thanks, fixed.

PR #7598, opened 2015-01-26
don't execute empty cells

### LINKED ISSUE #7591: Numbers in notebook cells are duplicated, incorrect
The input cell number in the UI is wrong. The `In` list of inputs kernel side is correct.

numbers

minrk:
reorder execute a little, so that clearing old state all happens before setting new state

closes #7591

### DIFF
--- a/IPython/html/static/notebook/js/codecell.js
+++ b/IPython/html/static/notebook/js/codecell.js
@@ -405,20 +405,26 @@ define([
         this.widget_subarea.height('');
         this.widget_area.height('');
         this.widget_area.hide();
-
-        this.set_input_prompt('*');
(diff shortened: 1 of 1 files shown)


minrk:
The use of `trim` means that one can't write a [whitespace](http://en.wikipedia.org/wiki/Whitespace_%28programming_language%29) kernel. If I leave out trim, it would cover most real cases.

takluyver:
Technically, you can write a whitespace kernel, the user would just need to ensure that each cell contains at least one comment, i.e. any non-whitespace characters.

Carreau:
Travis seem happy, merging.

PR #7599, opened 2015-01-26
remove duplicated border in indicator area

minrk:
looks like it was added accidentally, when doing some refactoring in #7586.

master:

screen shot 2015-01-26 at 10 56 51

this PR:

screen shot 2015-01-26 at 10 56 30


Carreau:
My Bad.

minrk:
@Carreau not a problem, every CSS change seems to have unforseen consequences.

PR #7600, opened 2015-01-26
protect websocket against errant messages

minrk:
- message handler being triggered after stream is closed (no channels)
- messages sent to invalid channels

reported in chat.


Carreau:
+1, kicked Travis.

PR #7602, opened 2015-01-26
Validation on the python side

SylvainCorlay:
This is a simple fix for  #6814. In general, whenever a widget model may be modified by both the backend _and_ the frontend, if there is a validation stage that transform the set values, the logic should be implemented on both sides. 

On this regard, but outside of the scope of this small PR: 
- I don't think that what we currently have for the validation of values in widgets is good enough, because the validation happens AFTER we set the value, and it does not prevent the assignment. 
- Ideally, we should be able to hook up the custom validation logic in the traitlet validation, via the traitlet type metadata, so that in case of error, the assignment does not actually happen. 
  - EDIT: See #7603 for the early validation  hook

### DIFF
--- a/IPython/html/widgets/widget_float.py
+++ b/IPython/html/widgets/widget_float.py
@@ -38,13 +38,28 @@ class _BoundedFloat(_Float):
     def __init__(self, *pargs, **kwargs):
         """Constructor"""
         super(_BoundedFloat, self).__init__(*pargs, **kwargs)
-        self._validate('value', None, self.value)
-        self.on_trait_change(self._validate, ['value', 'min', 'max'])
+        self._value_changed('value', None, self.value)
+        self._min_changed('min', None, self.min)
+        self._max_changed('max', None, self.max)
 
-    def _validate(self, name, old, new):
-        """Validate value, max, min."""
+    def _value_changed(self, name, old, new):
+        """Validate value."""
         if self.min > new or new > self.max:
             self.value = min(max(new, self.min), self.max)
+
+    def _max_changed(self, name, old, new):
+        """Make sure the min is always <= the max."""
+        if new < self.min:
+            raise ValueError("setting max < min")
+        if new < self.value:
+            self.value = new
+
+    def _min_changed(self, name, old, new):
+        """Make sure the max is always >= the min."""
+        if new > self.max:
+            raise ValueError("setting min > max")
+        if new > self.value:
+            self.value = new
 
 
 @register('IPython.FloatText')
--- a/IPython/html/widgets/widget_float.py
+++ b/IPython/html/widgets/widget_float.py
@@ -21,7 +21,7 @@ from IPython.utils.warn import DeprecatedClass
 # Classes
 #-----------------------------------------------------------------------------
 class _Float(DOMWidget):
-    value = CFloat(0.0, help="Float value", sync=True) 
+    value = CFloat(0.0, help="Float value", sync=True)
     disabled = Bool(False, help="Enable or disable user changes", sync=True)
     description = Unicode(help="Description of the value this widget represents", sync=True)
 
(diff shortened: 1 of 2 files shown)


jdfreder:
Thanks @SylvainCorlay .  I see you opened #7603 too, I still think this should get merged regardless, for 3.0 .

SylvainCorlay:
Yes I think too. I doubt the other logic will replace this one for IPython 3.0.

jdfreder:
@SylvainCorlay I've checked out this branch and I'm testing it now.  The code looks a-okay to me.

jdfreder:
@SylvainCorlay this does fix the problem @ssunkara1 reported, but I did find another problem.  I haven't check if it exists in master too or if it's a result of this fix.

```python
a.min = 20
a.value = 50
```

```python
print(a.value)
```

50

```js
%%javascript
console.log(a.get('value'));
```

50

```
a.value = 10
```

```python
print(a.value)
```

20

```js
%%javascript
console.log(a.get('value'));
```

50

The Javascript value is out of sync with the Python value.  It should log 20...

jdfreder:
My IPython logging convenience function:
```js
%%javascript
/**
 * Log to a notebook cell for easy reading.
 *
 * Usage:
 * %%javascript
 * element.log('hello world!');
 */
element.constructor.prototype.log = function() {
    for (var i = 0; i < arguments.length; i++) {
        this.append($('<div/>').text(String(arguments[i])));
    }
};
```

SylvainCorlay:
The problem also exists in `Master`. I can fix it tonight.
However, as long as we don't validate before assignment (as in #7603), any exception happening before the sync will lead to an inconsistent case.

IPython/html/widgets/widget_float.py
```diff
@@ -38,14 +38,32 @@ class _BoundedFloat(_Float):
...
+        self.on_trait_change(self._handle_min_changed, ['min'])
 
-    def _validate(self, name, old, new):
-        """Validate value, max, min."""
+    def _validate_value(self, name, old, new):
+        """Validate value."""
         if self.min > new or new > self.max:
             self.value = min(max(new, self.min), self.max)
 
+    def _handle_max_changed(self, name, old, new):
```
minrk: Why not use the `_max_changed` name instead of defining the method and registering it manually?

SylvainCorlay: I was simply replicating the logic from `_BoundedInt` in `_BoundedFloat`.

jasongrout:
+1 to the idea.  I think it would make sense to make the HasTraits validation come after the traitlet validation.  For example, the traitlet validation may cast a value to a specific type, etc.  That can be useful to do before the HasTraits validation function runs.

For example, suppose I have a CDate function that understands lots of string date formats and casts them to standard python date formats.  It's next to useless to have a HasTraits validation run before the traitlet validation since the HasTraits validation doesn't understand the string formats.  However, once the traitlet validation has run, there are some guarantees about what is in the value, and the HasTraits validation function can rely on that.

IPython/html/widgets/widget_float.py
SylvainCorlay: Now using `_max_changed`, `_min_changed` and `_value_changed`.

jdfreder:
Thanks @SylvainCorlay !

jdfreder:
It looks like there's a legitimate test failure.

jdfreder:
I think now that you're using the _X_changed syntax, validation is occurring earlier and that's causing the failure.  Strange, the default values should pass validation...

SylvainCorlay:
I can reset to the previous commit if it is fine with you. It is likely that the validation happens even before all the default values are set.

jdfreder:
It seems that may be a better idea than using a lock to prevent validation until after initialization.  I'm going to take a quick look...

jdfreder:
@SylvainCorlay ah it was just a typo: https://github.com/SylvainCorlay/ipython/pull/1

minrk:
If there's a good reason for it to not use the `_foo_changed`, that's fine. I just didn't see the reason.

takluyver:
Test failures look real.

jdfreder:
Sorry for the trouble @SylvainCorlay , I thought those typos were the problem.  At this point I think it's best to revert back to 200775b because I think my original evaluation of the problem was right (validation is occurring too early now).

SylvainCorlay:
ok, let me do that - and rebase.

jdfreder:
@SylvainCorlay looks like the commit got credited to me even though I haven't done anything :P .  Do you want to fix that before we Travis&merge?

SylvainCorlay:
@jdfreder I rebased the whole thing to 2 commits: 
1. use `_**_change` and do the correct validation, amended as yours. 
2. switch to other regular handlers.

SylvainCorlay:
It is funny how an old inline comment on an outdated diff is displayed again.

SylvainCorlay:
@jdfreder green button?

jdfreder:
:green_apple:

PR #7609, opened 2015-01-27
Change install_nbextension to take install only a single nbextension, with an optional destination

jasongrout:
See #7537 for discussion of this.

### DIFF
--- a/IPython/html/nbextensions.py
+++ b/IPython/html/nbextensions.py
@@ -241,7 +241,7 @@ def install_nbextension(path, overwrite=False, symlink=False, user=False, prefix
                     _maybe_copy(src, dest_file, verbose)
         else:
             src = path
-            _maybe_copy(src, dest, verbose)
+            _maybe_copy(src, full_dest, verbose)
 
 #----------------------------------------------------------------------
 # install nbextension app
--- a/IPython/html/nbextensions.py
+++ b/IPython/html/nbextensions.py
@@ -134,7 +134,7 @@ def check_nbextension(files, user=False, prefix=None, nbextensions_dir=None):
     return all(os.path.exists(pjoin(nbext, f)) for f in files)
 
 
-def install_nbextension(files, overwrite=False, symlink=False, user=False, prefix=None, nbextensions_dir=None, verbose=1):
+def install_nbextension(path, overwrite=False, symlink=False, user=False, prefix=None, nbextensions_dir=None, destination=None, verbose=1):
     """Install a Javascript extension for the notebook
     
     Stages files and/or directories into the nbextensions directory.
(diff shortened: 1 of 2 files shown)


jasongrout:
Bonus of having this: --destination now is a command-line argument for the ipython install-nbextension command.

IPython/html/nbextensions.py
```diff
@@ -304,9 +293,9 @@ class NBExtensionApp(BaseIPythonApplication):
     
     Usage
     
-        ipython install-nbextension file [more files, folders, archives or urls]
```
takluyver: Square brackets usually mean optional arguments. I would just put `path/url`

IPython/html/nbextensions.py
```diff
@@ -304,9 +293,9 @@ class NBExtensionApp(BaseIPythonApplication):
...
+        ipython install-nbextension [file, folder, archive, or url]
     
-    This copies files and/or folders into the IPython nbextensions directory.
```
takluyver: 'and/or' is no longer relevant, it's just 'or'.

IPython/html/nbextensions.py
```diff
@@ -173,75 +175,61 @@ def install_nbextension(files, overwrite=False, symlink=False, user=False, prefi
...
-                raise ValueError("Cannot symlink from archives")
-            if verbose >= 1:
-                print("extracting %s to %s" % (path, nbext))
-            archive.extractall(nbext)
-            archive.close()
-            continue
-        
-        dest = pjoin(nbext, dest_basename)
-        if overwrite and os.path.exists(dest):
+        else:
```
takluyver: I don't think this else branch can be reached.

jasongrout:
All very good points.  Fixed!

IPython/html/nbextensions.py
```diff
@@ -173,75 +175,59 @@ def install_nbextension(files, overwrite=False, symlink=False, user=False, prefi
...
-            elif path.endswith('.zip') or _safe_is_tarfile(path):
-                destination = str(uuid.uuid4()) # ignored for archives
-            else:
-                destination = basename(path)
-            _files[destination] = path
-        files = _files
+    if isinstance(path, (list, tuple)):
+        raise TypeError("path must be a string pointing to a single extension to install; call this function multiple times to install multiple extensions")
     
-    for dest_basename,path in (map(cast_unicode_py2, item) for item in files.items()):
```
takluyver: The previous code called `cast_unicode_py2` on both the path and the destination name; it looks like the new code only calls it on the path.

jasongrout: Yes, another good point!  Fixed!

takluyver:
I think this looks OK now, but I'll give it a day or two so others can have a look.

jasongrout:
Thanks!

PR #7610, opened 2015-01-27
make some target _blank configurable

Carreau:
there seem to be "only" 2 places where this make sens, the other
being href in mdcell, detach pager, help link in new windows.

set IPython._target = '_self' for try.jupyter.org shoudl work in
custom.js

@rgbkrk ?

### DIFF
--- a/IPython/html/static/base/js/namespace.js
+++ b/IPython/html/static/base/js/namespace.js
@@ -5,5 +5,6 @@ var IPython = IPython || {};
 define([], function(){
     "use strict";
     IPython.version = "3.0.0-dev";
+    IPython._target = '_blank';
     return IPython;
 });
--- a/IPython/html/static/edit/js/menubar.js
+++ b/IPython/html/static/edit/js/menubar.js
@@ -49,7 +49,7 @@ define([
         
         //  File
         this.element.find('#new-file').click(function () {
-            var w = window.open();
+            var w = window.open(undefined, IPython._target);
             // Create a new file in the current directory
             var parent = utils.url_path_split(editor.file_path)[0];
             editor.contents.new_untitled(parent, {type: "file"}).then(
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -272,7 +272,7 @@ define([
 
     KernelSelector.prototype.new_notebook = function (kernel_name) {
         
-        var w = window.open();
+        var w = window.open(undefined, IPython._target);
         // Create a new notebook in the same path as the current
         // notebook's path.
         var that = this;
--- a/IPython/html/static/notebook/js/menubar.js
+++ b/IPython/html/static/notebook/js/menubar.js
@@ -76,7 +76,7 @@ define([
             notebook_path
         ) + "?download=" + download.toString();
         
-        var w = window.open();
+        var w = window.open(undefined, IPython._target);
         if (this.notebook.dirty) {
             this.notebook.save_notebook().then(function() {
                 w.location = url;
@@ -101,7 +101,7 @@ define([
         
         this.element.find('#open_notebook').click(function () {
             var parent = utils.url_path_split(that.notebook.notebook_path)[0];
-            window.open(utils.url_join_encode(that.base_url, 'tree', parent));
+            window.open(utils.url_join_encode(that.base_url, 'tree', parent), IPython._target);
         });
         this.element.find('#copy_notebook').click(function () {
             that.notebook.copy_notebook();
--- a/IPython/html/static/notebook/js/notebook.js
+++ b/IPython/html/static/notebook/js/notebook.js
@@ -2033,7 +2033,7 @@ define(function (require) {
     Notebook.prototype.copy_notebook = function () {
         var that = this;
         var base_url = this.base_url;
-        var w = window.open();
+        var w = window.open(undefined, IPython._target);
         var parent = utils.url_path_split(this.notebook_path)[0];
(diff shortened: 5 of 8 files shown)


Carreau:
the "configurability" is limitted, and pretty hidden, could be made better later, but stop gap for now.

rgbkrk:
All of custom.js would then just read:

```JavaScript
IPython._target = '_self'
```

?

rgbkrk:
What was the original decision to use `_blank` by default?

minrk:
> What was the original decision to use _blank by default?

Two separate decisions, I think:

1. dashboard to notebook - it's nice to keep the dashboard open. This one's an ambiguous matter of taste, I think.
2. notebook links - it's super annoying to trigger onbeforeunload warning or possibly lose unsaved changes because you clicked a link. It's quite rare that links clicked in a notebook should logically close the notebook itself.

rgbkrk:
Now that we auto save, are we any closer to losing the `onbeforeunload` warning?

minrk:
> Now that we auto save, are we any closer to losing the onbeforeunload warning?

Not at all. Default autosave interval is two minutes, not on every keystroke. Even when we have cursor-based differential updates, there will still be a finite time during which the browser state hasn't been persisted, which needs to warn on departure.

rgbkrk:
To me, I wish it was all a one page app. When I want a link to open in a new page I usually make this decision myself, holding the right key for the platform I'm on or right clicking (open in new tab or window).

My issue with onbeforeunload is probably more about the behavior when you try to close, say, the Nature demo page after you made some changes. To a new person, they don't realize there's a saving flow.

About the dashboard links, there is some inconsistency leading to tab sprawl:

target_blank

jdfreder:
> dashboard to notebook - it's nice to keep the dashboard open. This one's an ambiguous matter of taste, I think.

I agree with @rgbkrk , I think the dashboard links should target self unless the user explicitly wants a new tab.

> notebook links - it's super annoying to trigger onbeforeunload warning

I think a better behavior would be to save without asking.  That way, when you click back on your browser the notebook is in the same state as when you left it.  This wouldn't work for onbeforeunload, but we could manually handle the links on the page, saving before navigation.

takluyver:
> we could manually handle the links on the page, saving before navigation.

-1 if that means replacing plain HTML links with something that does the navigation from JS.

jdfreder:
> -1 if that means replacing plain HTML links with something that does the navigation from JS.

We could leave the plain links, but capture the click event and do the navigation ourselves.  Here's an example I made, http://codepen.io/anon/pen/KwvvjK . I added a one second delay to simulate the round trip save.

takluyver:
OK, that's better, but I think from a user perspective, I'd still prefer links to open in a new tab instantly than open in the same tab after a noticeable delay while it saves the document.

rgbkrk:
Maybe I should rewind and assume that if they haven't explicitly saved or allowed an auto save, they're ready to exit and we should just let them. No thinking involved on the users part.

minrk:
@rgbkrk I think it would be fine to disable onbeforeunload in tmpnb, just not the 'real' notebook.

minrk:
From @jdfreder adding this argument to our various `window.open` calls for new, etc. should make things consistent.

rgbkrk:
Would it be possible to stick with a distinct set of window names instead of one global [target attribute](https://developer.mozilla.org/en-US/docs/Web/API/Window.open#Do_not_use_target.3D.22_blank.22)?

takluyver:
I don't think that paragraph is in sync with the way people are used to the web working. It's not uncommon for a site like gmail to open links in new tabs. It's quite rare for a set of links to open in a separate but common tab. The only case where I recall that behaviour is things like a help window which is reused as you click for help on different things. That's not really applicable to our situation.

ellisonbg:
Did we make any progress is seeing what our options on to make the configuration part of this work in a consistent way across the different links in the dashboard and live notebook page?

Carreau:
> Would it be possible to stick with a distinct set of window names instead of one global target attribute of _blank?

So you want each target to be configurable independently, so that you can potentially have a name ?

The problem with having a defined name, is that it will actually re-use another tab. Example if a bing `File -> copy` to `target_copy_notebook` , it means that I actually cant click twice on the menu. 

potential target of windows.open:
 - editor new
 - editor download
 - new notebook (strangely in kernelselector.js)
 - open, download, nbconvert notebook
 - copy notebook 
 - open pager in new window
 - new notebook (from dashboard) 
 - notebook list links
 - terminal list links
 - Jupyter logo.

Which one(s) should have configurability indivudually or globally ?

minrk:
I think the only logical values are `_self` and `_blank`. reusing window names is an old pattern that mostly results in annoyance and confusion that I don't really see anymore.

As for the list, I think there should only be either one or two values:

1. everything is the same
2. output/markdown links from user content are handled separately

I updated all of the `window.open` calls [here](https://github.com/minrk/ipython/compare/Carreau:_blank...target) when we were talking about it earlier, and I think we can treat user links in the notebook as one of:

- use the same config value
- use a separate config value
- leave it hardcoded as `_blank`

Carreau:
I've merge Min's branch in.

minrk:
ok, this should now allow you to open any of our UI links in a single window by setting `IPython._target = '_self'`. 'user' links in markdown and output remain hardcoded to `_blank`. We can add a separate configurable for those values later, if we need to.

minrk:
@rgbkrk want to set `IPython._target = '_self'` on tmpnb?

rgbkrk:
Thank you!

Issue #7613, opened 2015-01-28
Dashboard UI: it is not clear which button corresponds to which notebook


kmike:
Hi,

This is how a dashboard looks like now on a reasonably small window (less than 1200px wide):

![dashboard](https://cloud.githubusercontent.com/assets/107893/5933866/6cf75f08-a6e8-11e4-8a80-28474f542075.png)

I think that it is hard to tell which notebook will be deleted/stopped when you hover Delete or Shutdown buttons because notebook names are in far left.

Some ideas about how to fix it:

1. move buttons closer to text;
2. make the table less wide by default;
3. highlight the current table line.

(1) and (2) fix this particular issue better, but they have their own problems; (3) is less invasive.

mathieu1:
This has been addressed since merging of #7505 (new UI). 

![New UI](https://cloud.githubusercontent.com/assets/3292874/5843785/ce28d610-a163-11e4-92bf-395469ac5fcf.png)

The only difficulty that remains is to see exactly which notebooks are running. There were some nice ideas (which I believe were somehow postponed in favor of more important discussions) like:
> One idea, which I hesitate to try, is to color the notebooks if they are running - maybe green.
> Another, use a different icon, like an open notebook or the logo of the kernel.
> Another, display a second icon, either or the left or right side of the row, which indicates the notebook is running. This icon could be the kernel logo, to additionally indicate what type of kernel is running.

To me having the kernel logo (or an open notebook icon in absence of a logo) would be a great indicator of a notebook "running".

kmike:
Wow, thanks! It turns out my IPython master was outdated by ~16 hours, I didn't expect the issue to be fixed during this time :)

Carreau:
> Wow, thanks! It turns out my IPython master was outdated by ~16 hours, I didn't expect the issue to be fixed during this time :)

Wow, 16h ago master is soooooo Yesterday :-)

Thanks for feedback though !

PR #7616, opened 2015-01-28
Add running indication for notebook icon in dashboard

mathieu1:
While the new dashboard UI #7505 mostly addresses the issues raised in #7613, it remains that on large screens, it is hard to determine which notebook is running from the current indicator.

This PR mitigates this by adding an extra indication that a notebook is running by changing the icon color like so:
running-icon
Maybe a different icon would be desirable, but the best alternative I could find is `fa-play`...
alt-icon
It has also been suggested that the notebook's kernel logo be used to indicate running notebooks (and possibly fall back to the above behavior in case of missing logo).

### DIFF
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -8668,6 +8668,7 @@ ul.breadcrumb span {
 }
 .list_header {
   font-weight: bold;
+  background-color: #eeeeee;
 }
 .list_container {
   margin-top: 4px;
@@ -8689,6 +8690,9 @@ ul.breadcrumb span {
 }
 .list_item a {
   text-decoration: none;
+}
+.list_item:nth-child(odd) {
+  background-color: #f6f6f6;
 }
 .action_col {
   text-align: right;
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -8691,8 +8691,8 @@ ul.breadcrumb span {
 .list_item a {
   text-decoration: none;
 }
-.list_item:nth-child(odd) {
-  background-color: #f6f6f6;
+.list_item:hover {
+  background-color: #fafafa;
 }
 .action_col {
   text-align: right;
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -8794,6 +8794,23 @@ input.engine_num_input {
 .notebook_icon:before.pull-right {
   margin-left: .3em;
 }
+.running_notebook_icon:before {
+  display: inline-block;
+  font: normal normal normal 14px/1 FontAwesome;
+  font-size: inherit;
+  text-rendering: auto;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  transform: translate(0, 0);
+  content: "\f02d";
+  color: #5cb85c;
+}
+.running_notebook_icon:before.pull-left {
+  margin-right: .3em;
+}
+.running_notebook_icon:before.pull-right {
+  margin-left: .3em;
+}
 .file_icon:before {
   display: inline-block;
   font: normal normal normal 14px/1 FontAwesome;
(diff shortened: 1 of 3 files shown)


minrk:
ping @ellisonbg for design review. `fa-play` seems like the opposite of the right indicator, since its presence indicates that something is *not* running, but clicking it will make it run.

ellisonbg:
I agree that the idea of a play button is that you can click it to make something play - not an indication that it is already playing. I do agree that it is difficult to visually connect the name of a notebook horizontally with the "Running" text on the R side.

I think that changing the "book" icon to green is a reasonable thing to try. Let me do some testing of it to see what I think...

jdfreder:
Additionally I think it would be nice to add zebra striping to the dashboard- i.e http://getbootstrap.com/css/#tables-striped .

mathieu1:
Here's how the three tabs look with the zebra striping:
stripes-files
stripes-running
stripes-clusters

I added the grey background for the headers in the Files and Clusters tabs (matching what was already there in the Running tab). The striping uses a slightly lighter shade of grey. Another option would be to use the same grey for stripes and headers.

Carreau:
+1 to stripes, and slight highlight on hover ?

ellisonbg:
Not sure about the zebra stripes...will try it out soon...

On Thu, Jan 29, 2015 at 1:35 AM, Matthias Bussonnier <

> +1 to stripes, and slight highlight on hover ?
>
> —
> <https://github.com/ipython/ipython/pull/7616#issuecomment-71994714>.
>



-- 
Brian E. Granger

jdfreder:
+1 to the stripes too

fperez:
my vote is to just keep the green icon, no stripes.

takluyver:
From the dev meeting: we're not convinced by the stripes aesthetically, and we think they're superfluous once the green icon indicates which notebooks are running.

mathieu1:
Good, away with the stripes then. I also tested slight highlight on hover following @Carreau's suggestion. Stripes + hover highlight looks disgusting (at least the way I tested it), but hover highlight alone has a nice sober look and might be a worthwile addition to the green icon. Moreover, it invites the user to click anywhere on the line, which does toggle the checkbox (as opposed to necessarily have to reach for the checkbox itself) 

Here's how it looks with hover highlight alone:
hover

takluyver:
I like the subtle hover effect.

minrk:
Current state looks good to me, thanks @mathieu1.

ellisonbg:
I also like the hover effect, but let's try to pick the grey in a semi-non
random, so we don't have yet another shade of grey in our UI...otherwise we
will have to code name 3.0 "50 shades of grey" and do lots of explaining...

On Thu, Jan 29, 2015 at 4:30 PM, Min RK <notifications@github.com> wrote:

> Current state looks good to me, thanks @mathieu1
> <https://github.com/mathieu1>.
>
> —
> <https://github.com/ipython/ipython/pull/7616#issuecomment-72132751>.
>



-- 
Brian E. Granger

mathieu1:
:smiley:  Only reasonable shade of grey I could find is that of headers and tab hover (`@page-backdrop-color`), but that makes the highlighting less subtle.
std_color
A very minor issue now is that in the running tab, the line separating each header from the first row is also of that color, which makes sense when "**There are no nb/term running**" but is slightly strange when hovering over the first running nb/term.
std_color_running

takluyver:
Yeah, I don't think that looks right. I'm going to merge it with the very light shade of grey (i.e. leaving out the last commit). We can bikeshed over the precise colour once it's there. I'm not sure why it's so important to try to do everything with the same few greys.

ellisonbg:
+1

On Fri, Jan 30, 2015 at 9:34 AM, Thomas Kluyver <notifications@github.com>
wrote:

> Yeah, I don't think that looks right. I'm going to merge it with the very
> light shade of grey (i.e. leaving out the last commit). We can bikeshed
> over the precise colour once it's there. I'm not sure why it's so important
> to try to do everything with the same few greys.
>
> —
> <https://github.com/ipython/ipython/pull/7616#issuecomment-72238607>.
>



-- 
Brian E. Granger

jdfreder:
I'm happy with how this turned out!  Thanks @mathieu1

PR #7620, opened 2015-01-28
add missing error dialogs to tree actions

### LINKED ISSUE #7527: Misc permissions edge cases
Playing around with some permissions things, I have discovered a few more edge cases:

* Currently, if you try to delete a file, create a new folder, or duplicate a file, but don't have permission, it fails silently from the dashboard, though it appears to be handled correctly on the backend, e.g.

```
[W 21:39:21.046 SingleUserNotebookApp] delete /Problem Set 0/0 - Index.ipynb
[W 21:39:21.046 SingleUserNotebookApp] Permission denied: Problem Set 0/0 - Index.ipynb
```

* If you try to create a new file but don't have write permission on the directory, just it opens a new, blank tab. Again, relevant log messages show up suggesting it's being handled properly on the back end.

* If the directory is `chmod -w` but the user does have write permission to the *file*, IPython still complains that the file is read only.

minrk:
- new file/folder, copy were not catching errors
- made error messages more consistent

closes #7527

### DIFF
--- a/IPython/html/static/tree/js/newnotebook.js
+++ b/IPython/html/static/tree/js/newnotebook.js
@@ -85,16 +85,20 @@ define([
                     url += "?kernel_name=" + kernel_name;
                 }
                 w.location = url;
-            },
-            function (error) {
-                w.close();
-                dialog.modal({
-                    title : 'Creating Notebook Failed',
-                    body : "The error was: " + error.message,
-                    buttons : {'OK' : {'class' : 'btn-primary'}}
-                });
-            }
-        );
+        }).catch(function (e) {
+            w.close();
+            dialog.modal({
+                title : 'Creating Notebook Failed',
+                body : $('<div/>')
+                    .text("An error occurred while creating a new notebook.")
+                    .append($('<div/>')
+                        .addClass('alert alert-danger')
+                        .text(e.message || e)),
+                buttons: {
+                    OK: {'class' : 'btn-primary'}
+                }
+            });
+        });
     };
     
     return {'NewNotebookWidget': NewNotebookWidget};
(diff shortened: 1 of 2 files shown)


IPython/html/static/tree/js/newnotebook.js
```diff
@@ -85,16 +85,20 @@ define([
...
-            function (error) {
-                w.close();
-                dialog.modal({
-                    title : 'Creating Notebook Failed',
-                    body : "The error was: " + error.message,
-                    buttons : {'OK' : {'class' : 'btn-primary'}}
-                });
-            }
-        );
+        }).catch(function (e) {
```
takluyver: Does this need to be `...then(func).catch(errfunc)` instead of `...then(func, errfunc)`, or do you just prefer that style? The error should be coming from the original promise, not the then() section, right?

jdfreder: MDN says they're the same- "The catch() method returns a Promise and deals with rejected cases only. It behaves the same as calling Promise.prototype.then(undefined, onRejected)."  Although, looking closer, using .catch after the .then will additionally handle rejections from the preceding then...  More clearly said `P.then(s).catch(e)`, `e` will handle `s`'s rejections in addition to `P`'s whereas with `P.then(s, e)` it's not clear that `e` will handle `s`'s rejection at all (if s is a promise itself).

minrk: I think we want then().catch in general, since it catches errors in the then block as well, rather than only in the initial request.

jdfreder: Looks like what I said was right.  `P.then(s, e)` leads to an `Unhandled promise rejection Promise`:
http://jsfiddle.net/fq9z10ap/

Whereas `P.then(s).catch(e)` works as intended:
http://jsfiddle.net/4j54cyjy/

takluyver:
OK, then I think this is fine. @jdfreder, can you take a look if you haven't already, and merge when you're happy?

jdfreder:
Thanks @minrk !

PR #7625, opened 2015-01-28
Fix error when overwriting a bad symbolic link installing an nbextension

jasongrout:
The unit test fails before this fix and works after this fix.
Basically, if the already-installed symlink is a bad link, .exists()
returns False (so the existing link is not removed), while .lexists()
will return True (so the existing link is removed).

@takluyver or @minrk - can you look at this?  It's a simple bug to fix, and I provided a unit test to illustrate the bug and fix.


minrk:
Makes sense to me.

takluyver:
And me. One of us will merge once Travis clears it.

jasongrout:
Thanks!

PR #7630, opened 2015-01-29
remove translateZ hack

minrk:
while it does improve repaint performance on _some_ setups (OS X+Chrome+Low-DPI), it seems to have no effect on most (OS X+FF, OS X+Chrome+High-DPI, Linux+Chrome), and the opposite effect on others (Linux+FF).

There is also apparently a bug in most browsers regarding the layout of `position:`ed elements whose parents have this property.

ping @jasongrout

### DIFF
--- a/IPython/html/static/base/less/page.less
+++ b/IPython/html/static/base/less/page.less
@@ -79,12 +79,6 @@ body {
 }
 
 #site {
-    // avoid repaints on size with translateZ(0)
-    -webkit-transform: translateZ(0);
-    -moz-transform: translateZ(0);
-    -ms-transform: translateZ(0);
-    -o-transform: translateZ(0);
-    transform: translateZ(0);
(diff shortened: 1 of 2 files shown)


jasongrout:
The layout of `position:fixed` descendants is not a bug.  It's in the standard (http://www.w3.org/TR/css3-transforms/#transform-property).  It does break some assumptions, though, that fixed position elements are positioned relative to the viewport.

jasongrout:
So all-in-all, +1 on this.

minrk:
Seems to violate the definition of fixed positioning, and the spec itself seems to have some arguing about why on earth it would have that effect, when it doesn't make any sense, but yes, it does appear to be in the spec, apparently as an unintended, but resigned-to side effect of implementation challenges, rather than by design. The behavior is listed as a bug on Chrome, at least.

PR #7631, opened 2015-01-29
fix initial collapsed state on load

### LINKED ISSUE #7624: Cell output is sometimes invisible until toggled
From time to time after few recalculations of a cell (and possibly remote connection restorations), it's output does not appear. Cell identifier goes through regular steps `In [ ] -> In [*] -> In[4]` but the output is invisible. The workaround that works for me is to use `Cell -> Current Output -> Toggle`. After that `.btn.output_collapsed` appears and I can expand the cell output.

This is not critical, but is completely mysterious for non-advanced users of my public notebook server.

```bash
{'commit_hash': '4693727',
 'commit_source': 'repository',
 'default_encoding': 'UTF-8',
 'ipython_path': '/var/www/jupyter/ipython/IPython',
 'os_name': 'posix',
 'platform': 'Linux-3.13.0-32-generic-i686-with-Ubuntu-14.04-trusty',
 'sys_executable': '/var/www/jupyter/env/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.4.0 (default, Apr 11 2014, 13:05:18) \n[GCC 4.8.2]'}
```

minrk:
OutputArea.collapsed should only be assigned inside expand/collapse methods

setting `collapsed = false` when it actually is collapsed prevents expand from having any effect, resulting in apparently invisible output.

closes #7624

### DIFF
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -908,9 +908,10 @@ define([
             this.append_output(outputs[i]);
         }
         if (metadata.collapsed !== undefined) {
-            this.collapsed = metadata.collapsed;
             if (metadata.collapsed) {
                 this.collapse();
+            } else {
+                this.expand();
             }
         }
         if (metadata.scrolled !== undefined) {


takluyver:
Does this address #7624?

minrk:
@takluyver yes, I didn't realize there was an Issue open.

minrk:
specific reproduction steps in master:

1. new notebook
2. new cell with `1/0` or other code that will fail (either don't run it, or clear output)
3. reload
4. run the failing cell

jdfreder:
Thanks min, I've confirmed this solves my problem.  :+1:

Carreau:
second option would be to make collapsed a property :-P 

+1, merging.

PR #7632, opened 2015-01-30
add terminado as [notebook] dep on non-Windows

minrk:
It's not a dependency for the notebook to work, but neither is nbconvert and we express that as a dependency for [notebook]. This makes it ever so slightly simpler to get everything IPython uses.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -258,6 +258,9 @@ extras_require = dict(
     nbconvert = ['pygments', 'jinja2', 'mistune>=0.3.1']
 )
 
+if not sys.platform.startswith('win'):
+    extras_require['notebook'].append('terminado>=0.3.3')
+
 if sys.version_info < (3, 3):
     extras_require['test'].append('mock')
 
--- a/setup.py
+++ b/setup.py
@@ -264,11 +264,6 @@ if sys.version_info < (3, 3):
 extras_require['notebook'].extend(extras_require['nbformat'])
 extras_require['nbconvert'].extend(extras_require['nbformat'])
 
-everything = set()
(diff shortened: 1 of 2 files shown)


ssanderson:
I'm not sure if anyone else has my unique security constraints, but this would prevent me from using [notebook] for installation.  There isn't a straightforward way to say "install notebook dependencies **except** tornado", is there?

takluyver:
Maybe that still points to leaving it out. It is, after all, easy enough to tell people to `pip install ipython[notebook] terminado` if they want that.

minrk:
I"m fine with leaving it out.

ssanderson:
I don't want to oversell my position: I'm fairly sure that the intersection of "people who want to let strangers run IPython on their servers" and "people who don't want to let strangers run shell commands on their servers" is a pretty small set.  All the same, my (entirely selfish) preference would be for these to remain separate :)

rgbkrk:
Regardless of this PR, for your stringent requirements @ssanderson I'd say that you should explicitly list the dependencies you want to use to limit potential escape vectors.

ssanderson:
@rgbkrk that's a fair point.  Terminado is a bit unique in that it happens to provide an alternative means of executing code on the server that's outside the IPython kernel.  But it's probably true that this PR being accepted shouldn't create a security hole for me either way.  So actually, I rescind my previously stated preference, assuming IPython plans to continue to support terminado-less notebooks for at least the near future.

takluyver:
We decided we would go ahead with this. @minrk needs to add a line to make the dependency work with wheels.

PR #7638, opened 2015-01-30
minor correction: string "Fin.">"End of tour"

franktoffel:
I've found a Spanish word at the end of the IPython Notebook User Interface Tour. It appeared within the title of the last window.

Since I guess this was unintended, I have changed it.


minrk:
I think this was intended.

franktoffel:
Great! I wasn't sure of it :+1:

Issue #7643, opened 2015-01-31
Dict traitlet type default value cannot be None


SylvainCorlay:
1. One cannot set the default value of a Dict traitlet to `None`. 

```python
from IPython.utils.traitlets import *
class Foo(HasTraits):
    bar = Dict(None, allow_none=True)  # allow_none is True by default anyway for Dict. 
print Foo().bar  # yields {} instead of None
```

2. The documentation for the Dict traitlet says that 

*The default value is created by doing dict(default_value), which creates a copy of the default_value.*

while dict(None) is undefined. 

I would expect the default value to be None.

PR #7644, opened 2015-01-31
Allow for None default value in Dict traitlets

SylvainCorlay:
Meant to solve #7643. 

By the way, I think that it would be interesting if less trait types had `allow_none=True` (explicit is better than implicit).

### DIFF
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -1635,14 +1635,14 @@ class Tuple(Container):
 class Dict(Instance):
     """An instance of a Python dict."""
 
-    def __init__(self, default_value=None, allow_none=True, **metadata):
+    def __init__(self, default_value={}, allow_none=True, **metadata):
         """Create a dict trait type from a dict.
 
         The default value is created by doing ``dict(default_value)``,
         which creates a copy of the ``default_value``.
         """
         if default_value is None:
-            args = ((),)
+            args = None
         elif isinstance(default_value, dict):
             args = (default_value,)
         elif isinstance(default_value, SequenceTypes):
@@ -1657,15 +1657,15 @@ class Dict(Instance):
 class EventfulDict(Instance):
     """An instance of an EventfulDict."""
 
-    def __init__(self, default_value=None, allow_none=True, **metadata):
+    def __init__(self, default_value={}, allow_none=True, **metadata):
         """Create a EventfulDict trait type from a dict.
 
-        The default value is created by doing 
-        ``eventful.EvenfulDict(default_value)``, which creates a copy of the 
+        The default value is created by doing
+        ``eventful.EvenfulDict(default_value)``, which creates a copy of the
         ``default_value``.
         """
         if default_value is None:
-            args = ((),)
+            args = None
         elif isinstance(default_value, dict):
             args = (default_value,)
         elif isinstance(default_value, SequenceTypes):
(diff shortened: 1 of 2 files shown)


IPython/utils/traitlets.py
```diff
@@ -1635,14 +1635,14 @@ def validate_elements(self, obj, value):
 class Dict(Instance):
     """An instance of a Python dict."""
 
-    def __init__(self, default_value=None, allow_none=True, **metadata):
+    def __init__(self, default_value={}, allow_none=True, **metadata):
```
jdfreder: This PR looks great to me.  I would have merged it quickly for line 1645 below, but I want to make sure the others are okay with the default value change here.  I will wait for one or a few of the others who I think are familiar with traitlets to respond, @ellisonbg @minrk @takluyver and/or @jasongrout ?

ssanderson: Won't this will use the same dictionary as the default value of **ALL** Dict instances?

ssanderson: Consider, for example:
```
class Foo(object):
    def __init__(self, default_value={}):
        self.value = default_value

Foo_1 = Foo()
Foo_2 = Foo()

Foo_1.value['key'] = 'value'
print Foo_2.value
```
which prints:
```
{'key': 'value'}
```

minrk: Yes. It needs to make sure a *new* dictionary is created. There's an `Undefined` singleton used to indicate this elsewhere.

jdfreder: @ssanderson good point.  That suggests the default in master is wrong and @SylvainCorlay 's suggested default of None is the way to go.

jdfreder:
Will merge in a day if no one objects...

IPython/utils/traitlets.py
takluyver: I think the actual default value will end up being `dict({})`, which should create a new copy for each, see this line of code: https://github.com/ipython/ipython/blob/master/IPython/utils/traitlets.py#L891

But I'm glad you prompted me to go and check that.

takluyver: @jdfreder I think that should be fine, the default value should be an empty dict either way, so the only thing this should change is the behaviour if you explicitly call it as `Dict(None)`, and I think that's OK.

minrk:
> Will merge in a day if no one objects...

If the above comments re: using `Undefined` are addressed, sure. It's not mergeable in its current state.

IPython/utils/traitlets.py
```diff
@@ -1635,14 +1635,14 @@ def validate_elements(self, obj, value):
...
         """Create a dict trait type from a dict.
 
         The default value is created by doing ``dict(default_value)``,
         which creates a copy of the ``default_value``.
         """
         if default_value is None:
-            args = ((),)
+            args = None 
         elif isinstance(default_value, dict):
```
minrk: this is the key line that ensures the result is a *copy* of the default_value

minrk:
After @SylvainCorlay adds a test that verifies that passing a dict as the default value doesn't result in a shared dict, :+1: to this. Since we didn't already have a test for the default_value copying behavior that is required for this change to be safe.

SylvainCorlay:
I added a test checking that the default empty dict is actually copied. Let me know what you think.

minrk:
Not exactly the test I was thinking of, but works just the same. Thanks.

PR #7646, opened 2015-01-31
Kernel test tool - A small tool to test kernels from a script, useful for testing wrapper kernels

hari-allamraju:
I was building a simple redis kernel (on the list of kernels in the docs) and found no way to test the kernel for all the message spec messages. After looking at the test_message_spec.py test case, and some clarifications on the gitter channel I came up with this small tool that can be used to test any kernel.

For this tool, I refactored the test_message_spec.py file to move all the message spec validation classes to a common file in testing. Then I created a kerneltest.py under IPython/testing which has a method for reading a json file that specifies the message to test and the data to use and execute the test.

The json file can be written in the format - 

{
        "execute":{
                "test_code":"get a"
        }
}

This is still WIP and I need to add the last bits to read the file and be able to run this with command line arguments. I am creating this PR so that I can get an idea on whether this is the right way to go so far and make any changes based on the comments


dsblank:
This sounds useful! I'd be glad to try it by testing a variety of wrapperkernel-based kernels. Could probably then work this into our own kernel Travis tests.

ellisonbg:
This is very cool, we have been wanting to have something like this for a while. Ideally, we could test kernels for confirmity with the message spec.

hari-allamraju:
Thank you for the comments - I just updated the code with parameter checks and ability to run from a command line. The tool is now complete in all features I wanted to add. So please take a look when you get a chance and let me know any suggestions for improvements. I will try to put in some comments and documentation in the code during the week

hari-allamraju:
A simple test script for my redis kernel - we can put descriptive comments for each test in the json data and it will be ignored by the tool

https://gist.github.com/hari-allamraju/33714ef474e9ebcd8a0f

Carreau:
I added a link to this PR on [the wiki page about creating a kernel](https://github.com/ipython/ipython/wiki/IPython-kernels-for-other-languages). Feel free to update !

dsblank:
Would be handy to have a flag to set the version of the IPython message spec you want to check (eg, 4 or 5).

hari-allamraju:
@dsblank - can you elaborate more about the flag? Do you mean fetch different versions of the spec based on this flag? At the moment for example the code does not validate the kernel info reply and just passes it back to the tester to validate. So the protocol version would need to be visually checked

if you mean import different versions of the message spec based on the version, then we can do that by having multiple messagespec.py files for each version and then import one of those based on the command line param or a field in the json test script

takluyver:
From the dev meeting today: we like this idea, but we want to spend some time working out the details of how kernel authors should use it, so we won't aim to merge it before we release 3.0. It can probably go into 3.1.

hari-allamraju:
@takluyver - noted, I could not spend time on this in the week, will try to get some examples or docs on the wiki soon

minrk:
@hari-allamraju this is great! I was supposed to do something like this a while ago, and you beat me to it. We're getting 3.0 out the door, then we can get back to finalizing this proposal.

hari-allamraju:
Added a wiki page on testing with the kerneltest tool. Also added the script to the setup scripts to add to the path at installation

[Dev: Testing kernels against message specification](https://github.com/ipython/ipython/wiki/Dev:-Testing-kernels-against-message-specification)

hari-allamraju:
I have added the ability to specify and import different versions of message spec (code can be added as new versions are released) as long as each spec module implements the validate_message method. At the moment it only supports version 5.

Carreau:
> Added a wiki page on testing with the kerneltest tool. Also added the script to the setup scripts to add to the path at installation

Awesome ! Thanks !

takluyver:
We're going to work out how to make this a standalone tool - we're just about to split IPython up into multiple smaller repos. So we won't actually merge this into this repo, but we will use the code for a general kernel test tool.

hari-allamraju:
hi, just one question - when do you plan to have the split repos? will start following those in case I get any more ideas for this

takluyver:
We will be creating the new repos in the next couple of weeks, I expect.

takluyver:
OK, I'm digging into this now. I'm planning to pull it out into its own repository under the jupyter organisation.

ellisonbg:
Great!

On Mon, Apr 27, 2015 at 3:04 PM, Thomas Kluyver <notifications@github.com>
wrote:

> OK, I'm digging into this now. I'm planning to pull it out into its own
> repository under the jupyter organisation.
>
> —
> <https://github.com/ipython/ipython/pull/7646#issuecomment-96834860>.
>



-- 
Brian E. Granger

takluyver:
OK, we're up and running at https://github.com/jupyter/kerneltest. I'll work on it some more later this afternoon.

hari-allamraju:
Nice! Will take a look at the repo 

Thanks,
Hari




On Mon, Apr 27, 2015 at 3:21 PM -0700, "Thomas Kluyver" <notifications@github.com> wrote:












OK, we're up and running at https://github.com/jupyter/kerneltest. I'll work on it some more later this afternoon.

takluyver:
I talked over with @minrk a bit what the interface should look like, and we reckoned that it should have a Python class that you can subclass and extend with your own tests. Then we might also want a declarative format for filling in values for the standard tests - like the JSON specifications you wrote - but we're less sure about that. It might be simpler just to always have people write Python files with a few lines of boilerplate.

Based on what I've added so far, you use it like this (testing the R kernel here):

```python
import unittest
import kerneltest

class IRkernelTests(kerneltest.KernelTests):
    kernel_name = "ir"

    language_name = "R"

    code_hello_world = "print('hello, world')"

if __name__ == '__main__':
    unittest.main()
```

You can just run that file with Python, or use something like nose or py.test if you want to include it in a larger Python test suite. I've already discovered one [bug](https://github.com/IRkernel/IRkernel/commit/ab3893b71f8065222aca816acd182a75e4034ecd) in IRkernel by doing this. :-)

bollwyvl:
Whether the test itself is captured as a python class, having the tests generate a machine-readable compliance report will make things like [our WIP kernel registry](https://github.com/jupyter/try.jupyter.org/issues/7) much more sane.

takluyver:
What sort of machine readable output are you thinking of? I was just aiming for a standard testing approach with pass/fail, primarily for kernel authors to use before release.

bollwyvl:
Here's an example from the JSON-LD project: this [format](https://github.com/digitalbazaar/pyld/blob/master/tests/runtests.py#L346), is used to generate something like [this report](http://json-ld.org/test-suite/reports/). Of course, we can't reuse tests like they do, as every kernel's inputs and outputs will be different, but having a shared vocabulary for "it does the thing" is good.

As I mention over on that proposal, I think it's important that when a kernel author cuts a new release, it is a matter of submitting a PR to the registry with a single changed value (version, url, sha, tag, script, whatever) from which we can reproduce the tests, etc. and present which features are tested (and therefore implemented). There's really no way to know that now other than reading [each kernel home page](https://github.com/ipython/ipython/wiki/IPython-kernels-for-other-languages).

This is akin to the way the current base IPython Dockerfile tests itself as the last step before running: this is great for categorically knowing _the tests pass_, but doesn't give any information as to _what features were tested_, which is the important thing to manage across kernels.

minrk:
I strongly disagree that we should be able to run the tests, or that we should even be running the tests of any given kernel. I think that should be 100% up to the kernel authors.

PR #7650, opened 2015-01-31
don't prevent spec_changed from firing on load

### LINKED ISSUE #7645: Kernel not being set on notebook load
The notebook kernel is not being set on notebook load, which can cause some confusion. 

I pulled 3.0.0b1 and immediately uploaded an existing notebook.  The first cell is `!pip install redis`, which ran fine, but my subsequent `import redis` failed.  I sat frozen for a bit before it hit me that in IPy3 I have to think about which kernel I'm using.  I'm not sure if new/uploaded notebooks default to a Python 3 kernel, as the indicator just said, 'Kernel'.

Long story short, I had to make sure my kernel selection matched my pip/pip3 install to get it to work properly.

kernel_not_set

minrk:
Fixes check for the case where kernelspec info is coming from the kernel, rather than notebook metadata.

This case was getting skipped by a bad check on `_session_starting`.

closes #7645

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -218,7 +218,8 @@ define([
                 return;
             }
         }
-        if (this.notebook._session_starting) {
+        if (this.notebook._session_starting &&
+            this.notebook.session.kernel.name !== ks.name) {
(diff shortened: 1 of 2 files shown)


takluyver:
Looks sensible to me.

PR #7651, opened 2015-02-01
Add some spacing below h3 headers in HTML docs

takluyver:
Before, images immediately below an h3 header overlapped it:

sphinx_header_overlap

Fixed:

sphinx_header_fixed


minrk:
I think this was leftover from heading cells, to account for the spacing between cells.

PR #7656, opened 2015-02-01
Split out AnsiCodeProcessor from qt.console

bfredl:
Right now the kernel protocol uses ansi sequences for color and formatting. I would like to reuse the parsing code used by the qt client, for a third party client (nvim-ipy, prototype branch [here](https://github.com/bfredl/nvim-ipy/tree/highlight) ) without requiring the PyQt dependency.

This PR splits out the non-qt dependent parts of `qt/console/ansi_code_processor.py` in a separate file that doesn't import `PyQt` (or similar) . There should be no "real" changes to the code, just moving around stuff.


takluyver:
That makes sense, but I think we should leave this kind of refactoring for after 3.0 at this point. Marking as for 4.0.

bfredl:
If I remember correctly 4.0 is going to be the repo split release?
Ok, I can just cherry-pick this file into nvim-ipy until then.

takluyver:
Yep, that's a good point - it would either have to go in the Qt console
repo, or something that that could depend on.

If you want to maintain the ansi code processing logic in a separate
package, so we can make the Qt console depend on it for 4.0, that might
work out nicely.

bfredl:
Oh, I somehow believed that there just was going to be 2 repos (Jupiter + IPython)... But I figure the generic code could go wherever the shared code for the kernel consoles (console and qtconsole) are (like `KernelManager/Client`, which a third-party python client also might use), and the qt specific code in the qt console repo, as you said.

takluyver:
Nope, there's going to be about 9 repos, and we haven't yet worked out where some of the shared bits will be going. This is going to be interesting.

So if AnsiCodeProcessor is a neatly self contained bit of code, we'd probably be grateful if someone were to make it a small package of its own.

bfredl:
You mean a separate pip-package? I suppose I could make one in time for 4.0.

takluyver:
Yep, exactly. 4.0 won't be for several months, and for a first pass it would be fine to just put the existing code in a module and write a simple setup.py, so I'm sure you'd have no problem getting it done in time. I haven't looked at the code in detail to see if that makes sense, but if you could move it from one module to another within IPython, that suggests it's pretty well isolated.

bfredl:
Yes, the separated module has absolutely no dependencies outside python standard library.

takluyver:
Looking some more at that code and how it's used, I don't think it's clearly separated enough from the Qt console to be worth releasing as a separate package. I'd advise you to either use an existing separate library like [pyte](http://pyte.readthedocs.org/en/latest/index.html), or to copy the `AnsiCodeProcessor` class into your own codebase and use and improve it as you go - potentially releasing it as a separate package when you're happy with how it's working.

At some point post-split, this functionality could be discarded from the Qt console in favour of relying on one of those packages.

takluyver:
I'm closing this so we can start splitting up the repository. Feel free to copy this code into your own if you prefer it to a separate library.

Carreau:
> Feel free to copy this code into your own if you prefer it to a separate library.

If you make it a separate library ping us we might start to depend on it !

bfredl:
> Looking some more at that code and how it's used, I don't think it's clearly separated enough from the Qt console to be worth releasing as a separate package. 

True right now, but TBH `QtAnsiCodeProcessor` looks weird design-wise. It inherits from `AnsiCodeProcessor`, but their concerns are different; i e one parses raw terminal data into (still ANSI) attribute sets, and the other translates ansi attributes into qt attributes, which looks more like functional composition.

> If you make it a separate library ping us we might start to depend on it ! 
I will probably just copy the relevant parts to `nvim-ipy` for now, but later on I might publish library and send PR to refactor in the new QtClient repo :) 

I looked quickly into pyte but it looked wrong level of abstraction, it keeps an entire virtual screen state, instead of just the formatting state machine.

takluyver:
From my brief look over pyte, I think you could use it just for this part by implementing a 'screen' that ignores the other pieces like cursor movement, and just handles the formatting and output. But I didn't spend very long thinking about it, so it may well be harder than that.

PR #7660, opened 2015-02-02
Reverse hscrollbar min-height hack on OS X

### LINKED ISSUE #7623: CodeMirror hscrollbar blocks the pointer events
(Originally reported in Jupyterhub repo and redirected here)

Hello,

Currently there is a bug in a Jupyterhub markup that makes it very annoying to edit long single-line cells. I encounter it in Chrome 39 (OSX). MirrorCode horizontal scrollbar element hangs in front of the cell with `min-height: 18px` and captures all pointer events. To move the cursor around the cell, I have to use a small slit at the very top.

Proofpic

Seems that removing the `min-height` attribute should fix the problem

minrk:
OS X has optional behavior to only draw scrollbars during scroll, which causes problems for CodeMirror's scrollbars.  CodeMirror's solution is to set a minimum size for their scrollbars, which is always present.

The trade is that the container overlays most of the last line, creating a "dead zone" that swallows click events when there is scrolling to do, even when no scrollbar is visible.

This reverses the trade, recovering the click events at the expense of never showing the horizontal scrollbar on OS X when this option is enabled.

closes #7623

### DIFF
--- a/IPython/html/static/notebook/js/cell.js
+++ b/IPython/html/static/notebook/js/cell.js
@@ -21,7 +21,22 @@ define([
 ], function(IPython, $, utils, CodeMirror, cm_match, cm_closeb, cm_comment) {
     // TODO: remove IPython dependency here 
     "use strict";
-
+    
+    var overlayHack = CodeMirror.scrollbarModel.native.prototype.overlayHack;
+    
+    CodeMirror.scrollbarModel.native.prototype.overlayHack = function () {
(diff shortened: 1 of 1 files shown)


jdfreder:
> This reverses the trade, recovering the click events at the expense of never showing the horizontal scrollbar on OS X when this option is enabled.

I definitely prefer this.  Maybe I'm biased because the cursor dead zone has been bothering me.

PR #7661, opened 2015-02-02
only show prompt overlay if there's a prompt

### LINKED ISSUE #7659: Output toggler overlaps the output in presence of html.widgets
Happens on the latest master. By no means critical

Without dev tools

With dev tools

```shell
{'commit_hash': '4abc89a',
 'default_encoding': 'UTF-8',
 'ipython_path': '/var/www/jupyter/env/lib/python3.4/site-packages/IPython',
 'os_name': 'posix',
 'platform': 'Linux-3.13.0-32-generic-i686-with-Ubuntu-14.04-trusty',
 'sys_executable': '/var/www/jupyter/env/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.4.0 (default, Apr 11 2014, 13:05:18) \n[GCC 4.8.2]'}
```

minrk:
on which to overlay

widgets don't have prompts, so the overlay doesn't make sense

closes #7659

### DIFF
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -138,7 +138,9 @@ define([
         if (this.collapsed) {
             this.collapse_button.hide();
             this.element.show();
-            this.prompt_overlay.show();
+            if (this.prompt_area) {
+                this.prompt_overlay.show();
+            }
             this.collapsed = false;
             this.scroll_if_long();
         }


jdfreder:
Thanks @minrk .  A pleasant side effect of using the output widget with interact is that it gets more testing. :)

jdfreder:
Testing this now...

jdfreder:
Works

PR #7662, opened 2015-02-02
Prevent dashboard list items from being deselected on redraw.

### LINKED ISSUE #7657: [Dashboard] Deactivate autorefresh on something selected
Otherwise the autorefresh make the selection lost. 

Optimistically tagging for 3.0

jdfreder:
This is an alternative solution to @Carreau 's suggestion.  The selected items are remembered (by path) before redraw and they are reselected (if they still exist) after redraw.

closes #7657

### DIFF
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -219,6 +219,9 @@ define([
     var type_order = {'directory':0,'notebook':1,'file':2};
 
     NotebookList.prototype.draw_notebook_list = function (list, error_msg) {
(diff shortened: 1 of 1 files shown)


minrk:
I think this is the preferable fix.

Carreau:
>> This is an alternative solution to @Carreau 's suggestion. The selected items are remembered (by path) before redraw and they are reselected (if they still exist) after redraw.

> I think this is the preferable fix.

Yeah this is better, I just though it might be too complex to implement for now. 

But I forgot about the mighty @jdfreder .

PR #7663, opened 2015-02-02
Don't remove edit mode indicator for command mode, hide it instead.

### LINKED ISSUE #7652: #modal_indicator constant width ?
the edit/command mode indicator, even if show "nothing" on command mode, should have the same width to avoid notification messages to move around.

jdfreder:
closes #7652

@Carreau and @ellisonbg for design review.

### DIFF
--- a/IPython/html/static/notebook/less/notificationarea.less
+++ b/IPython/html/static/notebook/less/notificationarea.less
@@ -30,12 +30,16 @@
     .indicator_area();
 }
 
-.edit_mode .modal_indicator:before {
+.modal_indicator:before {
     .icon(@fa-var-pencil)
 }
 
+.edit_mode .modal_indicator:before {
+    visibility: inherit;
+}
+
 .command_mode .modal_indicator:before {
-    .icon(' ');
+    visibility: hidden;
 }
 
 .kernel_idle_icon:before {
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -8925,6 +8925,7 @@ ul#new-menu {
   text-rendering: auto;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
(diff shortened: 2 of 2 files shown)


IPython/html/static/style/style.min.css
```diff
@@ -8925,6 +8925,7 @@ ul#new-menu {
   text-rendering: auto;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  transform: translate(0, 0);
```
minrk: Are these supposed to be happening?

jdfreder: No, not intentionally.  I'll update my less.

jdfreder: @minrk  - Ah it's a change introduced by https://github.com/ipython/ipython-components/pull/53 .  It looks like this is the first CSS recompile since that was merged.

minrk: ok, that's fine, then.

IPython/html/static/notebook/less/notificationarea.less
```diff
@@ -30,12 +30,16 @@
...
     .icon(@fa-var-pencil)
 }
 
+.edit_mode .modal_indicator:before {
+    visibility: inherit;
+}
+
 .command_mode .modal_indicator:before {
-    .icon(' ');
+    visibility: hidden;
```
minrk: This isn't how we should do it. Command mode has no icon, it isn't invisible (I have a non-empty command mode icon, and this would hide it).

Carreau:
Seem sensible.

IPython/html/static/notebook/less/notificationarea.less
jdfreder: `fa-fw` is probable a better idea

jdfreder:
Closed in favor of my alternative: #7665

Issue #7647, opened 2015-01-31
Can't reconnect to running Python2 notebook


peter-ch:
I installed the current git-master version of IPython, started a notebook server on a remote machine and then I made a notebook running a long calculation. It periodically outputs results (with clear_output(wait=True) and print()). So I left it running, saved it, and closed the browser. But when I returned to it, it didn't work. It says "Kernel starting" and it stays like that forever. The icon shows that the notebook is busy, the CPU usage is 100% as expected, but I can't see any output. Reconnect Kernel doesn't help. 
![screenshot from 2015-01-31 20 22 34](https://cloud.githubusercontent.com/assets/2143723/5989194/a906e6fc-a987-11e4-95ef-9979428c4ada.png)

takluyver:
I think we made it wait for the kernel to be idle before it decided that the kernel had started. Which is problematic in the case you describe.

jdfreder:
@takluyver does the kernel run in a separate process (or real thread) than the web server?  If so, I'd think we could make it so the web server knew that the kernel exists and is busy- then it could relay that to the front-end which should fix this problem.

takluyver:
@jdfreder you haven't completely forgotten our architecture, have you? Yes, kernels are separate processes from the web server.

As I remember, we made a deliberate decision that the frontend should wait for the kernel to be connected and idle before showing that it had started, so we could be sure it was ready to receive messages. In most cases, that makes sense, but it doesn't if you have a long-running function and you reconnect to a notebook.

@minrk probably recalls this stuff better than I do.

jdfreder:
Maybe @minrk can shed some light on why keeping track of the kernel's state in the web server is a bad idea.  It seems like it would be an 'easy' solution...

Carreau:
The server never disconnect from the kernel and act only as a zmq-websocket bridge at this stage. 
adding knowledge of kernel state to the server would break the all API/design.

The other thing that it won't help with is that disconnection browser and reopen won't forward output. 
All that should be solve once we have notebook state on the server, OT, and browser beeing only a view of the server model.

minrk:
I think there are two separate ways in which we wait for kernel_info:

1. on the server, we wait for kernel_info reply before opening the websocket. We only do this the first time someone connects to the kernel. We do this one, because we need to wait to do msg spec adaptation before any messages can be processed.
2. In the Javascript, we use a kernel_info ping to ensure that the initial busy state is correct.

[...]

PR #7664, opened 2015-02-02
don't trigger kernel_starting after kernel_connected

minrk:
trigger instead at the beginning of Kernel.start

if a websocket is fully connected, the kernel has finished starting

addresses notification confusion when connecting to busy kernel in #7647

### DIFF
--- a/IPython/html/static/services/kernels/kernel.js
+++ b/IPython/html/static/services/kernels/kernel.js
@@ -180,6 +180,7 @@ define([
             url = url + "?" + qs;
         }
 
+        this.events.trigger('kernel_starting.Kernel', {kernel: this});
         var that = this;
         var on_success = function (data, status, xhr) {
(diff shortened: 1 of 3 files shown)


jdfreder:
When you said subtle changes you weren't kidding!  The code looks okay, but I don't understand the order of events here enough to know what consequences may arise from reordering them.

minrk:
The main difference that results from this:

## Before

We show 'Kernel starting...' after completing a connection to a kernel, while waiting for a kernel_info reply that will trigger the `kernel_ready` event.

## After

We show 'Kernel starting...' when we ask for a kernel to be created, until we finish connecting to it.


The change is possible now because the websocket does not finish connecting until a kernel has finished starting. Prior to the server-side wait, it wasn't possible for the frontend to distinguish between a kernel that hasn't replied because it's busy working or busy starting.

jdfreder:
Awesome, thanks for the details.  :+1: from me.

PR #7665, opened 2015-02-02
Fix the mode indicator width.

### LINKED ISSUE #7652: #modal_indicator constant width ?

jdfreder:
Use fa-fw to set the fixed width of the icon.

closes #7652 
alternative to #7663

@minrk who wanted an alternative.
@Carreau and @ellisonbg for design review.

### DIFF
--- a/IPython/html/static/notebook/less/notificationarea.less
+++ b/IPython/html/static/notebook/less/notificationarea.less
@@ -1,3 +1,4 @@
+
 #notification_area {
     .pull-right();
     z-index: 10;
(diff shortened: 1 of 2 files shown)


minrk:
I think this is better, thanks.

minrk:
Confirmed this has the desired behavior, thanks. Did you mean to include the "Remove accident" commit?

jdfreder:
Yeah, it just removes an added blank line which was accidental.  Since I commit `invoke css` separate to my other changes, I was unable to quickly `git commit --amend` to fix the first commit.

minrk:
@jdfreder that's what the `fixup` command in `git rebase -i` is for! It's like `--amend` for commits other than the most recent.

minrk:
In any case, :+1: to merge. You don't need to rewrite history.

jdfreder:
> that's what the fixup command in git rebase -i is for!

I've never used that command before :P  Next time...

PR #7666, opened 2015-02-03
add download link to editor

### LINKED ISSUE #7642: Provide download link in the editor view
The `File` menu in the editor has New, Save, and Rename but no way to download the files.

I'm guessing that adding the `/files/{filename}` endpoint is simple enough. Where it seems to fail for me is in detecting the content type.

For context, I'm trying to use JupyterHub with a Chromebook to see what the experience is like.

minrk:
closes #7642


rgbkrk:
Cool.

takluyver:
Tested, looks good.

PR #7668, opened 2015-02-03
Add KernelSpecManager.whitelist

minrk:
set of allowed kernels by name

Mainly useful for excluding the 'native' python2/3 kernel, but it's generic.

This was requested somewhere, but I can't find an Issue for it, so I'm going to assume it was on gitter / mailing list.


takluyver:
Seems OK. I'll give it a day for others to look at it, then merge.

Issue #7673, opened 2015-02-03
Selection widget reverse lookup with numpy arrays


SylvainCorlay:
With selection widgets, a reverse lookup occurs when setting a new value to check whether the new valid is among the available options. This lookup uses the `==` operator, which is overloaded for numpy.array. 

We should give the possibility to the user to provide his own comparison predicate on construction of the selection widgets.

minrk:
We use equality to trigger change events on traitelts, don't we? Doesn't this need to be fixed at the lowest traitlets level as well?

minrk:
Apparently the equality check has been removed in underlying traitlets (which treats all numpy arrays as unequal, even identical ones).

PR #7674, opened 2015-02-03
Allow predictate to be used for comparison in selection widgets

SylvainCorlay:
Meant to solve #7673. Let me know what you think.


SylvainCorlay:
In the case of a numpy array, one can use `np.array_equal`.

minrk:
Isn't there a much larger issue related to the trait change events and numpy arrays?

SylvainCorlay:
@minrk done. I don't think that the failed test is related with my change.

minrk:
@SylvainCorlay no, definitely not related. I'll wait for Travis to double check, then merge. Thanks!

minrk:
Thanks

PR #7678, opened 2015-02-03
Convert cpasted code to unicode on Python 2

### LINKED ISSUE #7676: cpaste magic error with logging history
I am having a similar problem to issue #2412 upgrading from ipython==2.1.0 to ipython==2.3.1. Here is [my setup](https://gist.github.com/tlmaloney/ba7663022678da201808), I have just one startup script that logs. And [here is the error](https://gist.github.com/tlmaloney/1cb609c12ee864166c42) I'm getting. If I remove the `05_log.py` file from the startup folder, I get no errors when using `%cpaste`.

takluyver:
Closes gh-7676

No test, because this is very awkward code to test, but I can't reproduce the error manually after this change.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -66,7 +66,7 @@ def get_pasted_lines(sentinel, l_input=py3compat.input, quiet=False):
         prompt = ""
     while True:
         try:
-            l = l_input(prompt)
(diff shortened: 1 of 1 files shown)


tlmaloney:
Confirmed these changes work, I made them against master in a virtualenv, thanks a ton!

takluyver:
No worries, thanks for confirming the fix.

PR #7680, opened 2015-02-04
Fix widget view persistence.

### LINKED ISSUE #7679: Widget persistence bug
When reloading a notebook that has multiple widget states saved in the browser's local storage, the last widget instantiated is rendered all the widget output areas. 

This should probably be solved before 3.0

jdfreder:
Corrects a variable's scope.

closes #7679

@sylvaincorlay

### DIFF
--- a/IPython/html/static/widgets/js/manager.js
+++ b/IPython/html/static/widgets/js/manager.js
@@ -381,16 +381,18 @@ define([
                         };
 
                         // Get the views that are displayed *now*.
-                        model_promises.push(utils.resolve_promises_dict(model.views).then(function(model_views) {
-                            for (var id in model_views) {
-                                if (model_views.hasOwnProperty(id)) {
-                                    var view = model_views[id];
-                                    if (view.options.cell_index) {
-                                        state[model_id].views.push(view.options.cell_index);
+                        (function(local_state) {
+                            model_promises.push(utils.resolve_promises_dict(model.views).then(function(model_views) {
+                                for (var id in model_views) {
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
Checked out the branch and it does fix the issue.
(One needs to re-execute the cells once first, because whatever was in the local storage before your fix was probably wrong. )

minrk:
@jdfreder thanks for the fix, @SylvainCorlay thanks for the report and confirmation of the fix.

@jdfreder is this something that can be reasonably tested?

SylvainCorlay:
`phantomjs` apparently supports local storage.

jdfreder:
@minrk last commit adds a test.

minrk:
It wasn't possible to test the return value of `get_state` directly? That wasn't incorrect?

SylvainCorlay:
In general, now that you are using the local storage for the widgets, Phantomjs' local storage should probably be cleared before running any test suite.
http://stackoverflow.com/questions/18408868/disabling-localstorage-on-phantomjs-for-clean-testing

jasongrout:
for what it's worth, @jdfreder's analysis and fix of the problem seems appropriate.

SylvainCorlay:
Yeah, :+1:  on this.

jdfreder:
@minrk yeah, the `get_state` call is what was returning an incorrect value.  I implemented the test the way I did so it tests the entire feature.

minrk:
Makes sense, you just said it was ugly, and I thought there might be a way for it to be less so. Thanks!

PR #7667, opened 2015-02-03
Dashboard selector dropdown.

jdfreder:
closes #7637

~~Before anyone asks, I tried hard to find a quick/easy way to make the checkbox a tristate (including manually overlaying fa-minus, etc...).  There are no good ways to do this.  We will either need our own control or another third party component- which I don't think we should do at this point (maybe in our next major release).~~

@ellisonbg for design review.

Issue #7628, opened 2015-01-29
Dashboard : Number of selected items 


Carreau:
I think a number of selected items near the buttons would be nice. 

Or at least a indication one/many. I tend to scroll and switch windows before going back to top to click on the action, and I'm not sure how many items are selected; or if something was selected from a previous time I was on the windows, and I did nothing.

PR #7681, opened 2015-02-04
Streamline tree-selector menu

mathieu1:
Streamlines the menu from #7667 with
- Shorter text, with icons.
- Main checkbox no longer opens the dropdown.
  Instead it just selects all/none.
- Fewer menu items but more actions possible by adding checkboxes
  in the menu
- Menu stays open until clicking outside.

Here's how it looks:
tree-selector

Of course, this needs some CSS work for it to look right, but let's see what you guys think about the mechanics first.
JS comments / CSS help and suggestions warmly welcome!

With more CSS tweaking, we could make that menu more compact so that this could be a dropup instead of a dropdown to avoid hiding the first items in the list (currently, the top of the menu is hidden behing Jupyter, see below)
dropup

### DIFF
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -499,12 +499,17 @@ define([
             if (selected_nums[i] === 0) {
                 $('#'+checkbox_ids[i])[0].indeterminate = false;
                 $('#'+checkbox_ids[i]).prop('checked', false);
-            } else if (selected_nums[i] === total_nums[i]) {
-                $('#'+checkbox_ids[i])[0].indeterminate = false;
-                $('#'+checkbox_ids[i]).prop('checked', true);
+                $('#badge-'+checkbox_ids[i]).text('');
             } else {
-                $('#'+checkbox_ids[i]).prop('checked', false);
-                $('#'+checkbox_ids[i])[0].indeterminate = true;
+                // Update badge
+                $('#badge-'+checkbox_ids[i]).text(selected_nums[i]);
+                if (selected_nums[i] === total_nums[i]) {
+                    $('#'+checkbox_ids[i])[0].indeterminate = false;
+                    $('#'+checkbox_ids[i]).prop('checked', true);
+                } else {
+                    $('#'+checkbox_ids[i]).prop('checked', false);
+                    $('#'+checkbox_ids[i])[0].indeterminate = true;
+                }
             }
         }
     };
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -122,12 +122,9 @@ define([
             $('.delete-button').click($.proxy(this.delete_selected, this));
 
             // Bind events for selection menu buttons.
-            $('#tree-selector .select-all').click($.proxy(this.select_all, this));
-            $('#tree-selector .select-notebooks').click($.proxy(this.select_notebooks, this));
-            $('#tree-selector .select-running-notebooks').click($.proxy(this.select_running_notebooks, this));
-            $('#tree-selector .select-files').click($.proxy(this.select_files, this));
-            $('#tree-selector .select-directories').click($.proxy(this.select_directories, this));
-            $('#tree-selector .deselect-all').click($.proxy(this.deselect_all, this));
+            $('.tree-selector').change(function(){that.select($(this).attr('id'),$(this).is(':checked'))});
+            // Do not propagate click for the menu to prevent the menu from closing
+            $('#tree-selector-menu').click(function(event){event.stopPropagation();})
         }
     };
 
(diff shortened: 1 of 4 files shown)


mathieu1:
One could easily address #7628 using badges... (will fix the UI quirks later if this stays)
badges

dsblank:
If the counts could be right-justified on the menu row, that would look nice.

jdfreder:
@mathieu1 I like this idea a lot!  

Few of points:
- I don't think that we should make the dropdown a drop-up.  I don't think it's that big of deal if the first few rows are not visible because the dropdown blocks them.
- I feel a little uneasy about the "stickyness" of the menu - the dropdown not hiding when you select an item.
- I think the count is a good idea.  If we decide not for all the items, I think we should at least show it for the running notebooks.

jdfreder:
@ellisonbg design review.

jasongrout:
Wow, the demos above look really nice!

mathieu1:
@jdfreder I don't have strong feelings regarding stickyness, but here are two good things about it:
- a priori it doesn't slow down common actions (like select all (running) notebooks and shutdown), and can significantly speed up rarer ones (like select idle notebooks).
- more importantly, when clicking on an "undetermined" checkbox, the user usually doesn't know the outcome of clicking (whether it's going to select all or none), and actually expects to be able to click twice in case he doesn't get the result he wants.

dsblank:
Another use for the counts is to actually count the items (rather than using it to select them). I have an extension allowing students to turn in their notebooks for homework, and I'd like to know how many have/have not quickly. (Not critical, of course, because I can pop over to a terminal... but this would be handy).

mathieu1:
@dsblank Maybe selecting all notebooks is easy enough to see how many there are.

dsblank:
Yes, that is all that I mean... I would like the counts for notebooks... I
like this, with the counts for notebooks +1 (and other counts, too).

mathieu1:
@dsblank Ok thanks, I thought you wanted another counter.

Rebased to include #7687, which solves refreshing issues.

Also improved the global experience:
- Disable checkboxes when there is nothing to select
- Avoid selecting the parent folder `..`
- right-alignment of badges in the menu
- HTML and CSS improvements for better look (still needs reworking)
Yet another screenshot:
disabled

Carreau:
1 - Awesome increase in usability.
2 - Stil a big code change for 3.0, so would like more votes. 
3 - small request/question to others : badge in the 'select all'  to  show the total number of selected items.

mathieu1:
About @Carreau's 3rd suggestion, the difficult question is really about the positioning of the badge. I don't quite see where to put it so that it doesn't look ugly... Current look below:
select-all-badge

(BTW, for some reason all my checkboxes turned grey (using Chrome) all of a sudden, *even on master*, and I haven't been able to figure out why (cleaning cache didn't help, nor rebooting). Does anyone else have the same issue?)

takluyver:
Would it look nicer as a plain number without the circle? Also, if we do go for that, we should certainly reserve enough width for at least two digits so the things to the right of it aren't moving as you select stuff?

As to the numbers in the dropdown: I would find it more natural if they were constant, showing the total number of folders, notebooks, running notebooks or files in the current view, rather than changing as things are selected.

Carreau:
> As to the numbers in the dropdown: I would find it more natural if they were constant, showing the total number of folders, notebooks, running notebooks or files in the current view, rather than changing as things are selected.

While I disagree, I think you need to know how many thing you have selected mostly I want to know if I have 1 or many thing selected when I want to take an action like deleting. 

> (BTW, for some reason all my checkboxes turned grey (using Chrome) all of a sudden

That would't surprise me from the chrome team to change the appearance of elements.

mathieu1:
A selection of counter styles:
badge-style

mathieu1:
Cleaned up a few remaining quirks. Already got some testing on FF/chrome/safari.

- Ready for review / testing
- Awaiting decision(s) on the counters 
- Further suggestions still welcome

jdfreder:
Since it seems like we are still undecided on some of the things in this PR, I added it to this week's meeting agenda to decide what release we will target it to.

ellisonbg:
I am +1 on fixing the following in 3.1:

* select all/none behavior of main checkbox
* Styling the different menu items (with icons).

I am -1 on the complexity of the item count tags, the sticky menu and the
checkboxes in the menu. Feels like over design.

On Mon, Feb 9, 2015 at 12:17 PM, Jonathan Frederic <notifications@github.com
> wrote:

> Since it seems like we are still undecided on some of the things in this
> PR, I added it to this week's meeting agenda to decide what release we will
> target it to.
>
> —
> <https://github.com/ipython/ipython/pull/7681#issuecomment-73581749>.
>



-- 
Brian E. Granger

takluyver:
We discussed this in the dev meeting today, and more or less agreed with @ellisonbg 's comment above - we want to keep a couple of small enhancements, but we don't want to add lots of features to this until it's clear which ones people actually want.

mathieu1:
Updated according to @ellisonbg's comments. Specifically:

Kept changes:
- fewer, shorter menu items (too cryptic?), with icons.
- main checkbox does not open menu, but selects all/none.
- total selection count next to main checkbox (unsure if this is to be
kept. It does address #7628).

Removed changes:
- sticky menu.
- checkboxes and counters in menu.

jdfreder:
I think now that the behavior has changed minimally, this can be retargeted for 3.1.  I'm excited about these changes.

IPython/html/static/tree/js/notebooklist.js
```diff
@@ -382,78 +386,26 @@ define([
...
-        this.deselect_all();
-        $('.list_item').each(function(index, item) {
-            if ($(item).data('type') === 'file') {
-                $(item).find('input[type=checkbox]').prop('checked', true);
+            // For each item, determine if the state should be set, depending on the selection_type that triggered select
+            var set_state = (selection_type === "select-all");
+            set_state = set_state || (selection_type === "select-folders" && $(item).data('type') === 'directory');
+            set_state = set_state || (selection_type === "select-notebooks" && $(item).data('type') === 'notebook');
+            set_state = set_state || (selection_type === "select-running-notebooks" && $(item).data('type') === 'notebook' && that.sessions[$(item).data('path')] !== undefined);
+            set_state = set_state || (selection_type === "select-files" && $(item).data('type') === 'file');
```
jdfreder: I have a minor style request here.  These lines are a little long, partially because you keep fetching `$(item).data('type')`.  Could you put that in a variable, like `type` or `item_type`?

IPython/html/static/tree/js/notebooklist.js
```diff
@@ -382,78 +386,26 @@ define([
...
+            if (set_state) {
+                $(item).find('input[type=checkbox]').prop('checked', state);
             }
```
jdfreder: I'm not sure that this behavior makes sense anymore.  Why not remove the `state` function argument, renamed `set_state` to `state`, and get rid of the conditional?

mathieu1:
@jdfreder Thanks for your comments, I just addressed them. The second suggestion changes a bit the behavior : selecting "all notebooks" for exemple will now also deselect everything else (before, it was just adding all notebooks to the selection, but not changing the rest). The current behavior is exactly like Gmail's.

Reminder to self/all: if merged for 3.x, this also fixes #7737

jdfreder:
Travis failure was sporadic.  I restarted the test.  @Carreau @ellisonbg or @minrk want to review and merge this?  I'm okay with it in it's current state :+1:

ellisonbg:
It would be good to get some more detailed review, but overall I like the current state. Thanks for the work on this! The only part of it that I am 50/50 on is the overall selected count display. I can go either way on this depending on what other's think.

minrk:
Looks good to me. The only thing I might change is the padding around the number, so that it doesn't jump when it changes. But I think it's good enough to merge right now, and we can continue with smaller improvements later.

mathieu1:
@minrk Do you mean adding even more space so that, say, 3-digit numbers would fit? Anyhow, merging and keeping smaller fixes for later is fine with me as well.

jdfreder:
Thanks @mathieu1

minrk:
@mathieu1 I would pick a balance - at least two digits shouldn't change width, but maybe three would be okay. It's a tradeoff of how much space it should take up. Alternately, it could be placed somewhere else, such that expanding doesn't push other elements around.

PR #7683, opened 2015-02-04
fill #site height with notebook containers

### LINKED ISSUE #7677: DropDown widget getting cut off
I am playing around with the widgets in iPython 3.0.0b1 and found that the DropDown widget gets cutoff if there aren't a sufficient number of cells underneath the cell that's displaying the widget. It's better explained using the two images attached. As you can see, in the first image, the dropdown menu is cut-off without showing all of the options. In the second image -- where another cell has been added below the current cell -- the drop down shows one more option but still gets cut off. 

For a drop down with 10-15 options, this seems like a less than ideal way to get the DropDown widget to display all its options. Isn't there a way to set the `overflow-y` property or something to get it to scroll instead?

img1

img2

minrk:
avoids cutoff of overlay elements

closes #7677

### DIFF
--- a/IPython/html/static/notebook/less/notebook.less
+++ b/IPython/html/static/notebook/less/notebook.less
@@ -8,12 +8,14 @@
 
 #ipython-main-app {
     .border-box-sizing();
+    height: 100%;
 }
 
 div#notebook_panel {
     margin: 0px;
     padding: 0px;
     .border-box-sizing();
+    height: 100%;
 }
 
 #notebook {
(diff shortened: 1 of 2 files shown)


jdfreder:
@minrk the code looks goods.  Testing functionality...

minrk:
@jdfreder ha, I think the code is horrible and I hate it, but it seems to work.

jdfreder:
:P Well I wish I could tell you I knew of a better way to do it.

Verified the problem and then verified that this fixes it.  Thanks!

PR #7686, opened 2015-02-04
Make shift-tab work as "indent-less" operation, too

juhasch:
Currently IPython shadows the `shift-tab` operation of codemirror. This PR copies some of the logic of handling the `tab` key a few lines below in the code to "shift-tab", too.

Now you can still get the tooltip using `shift-tab` _and_ use it for indent-less.


takluyver:
That seems reasonable.

Carreau:
+1

juhasch:
Thanks.

PR #7687, opened 2015-02-04
Defined selected in notebooklist constructor

jdfreder:
@jasongrout


jasongrout:
This seems to fix the error I was seeing, so +1.

PR #7694, opened 2015-02-05
bump pyzmq version dependency to 13

minrk:
pyzmq < 13 doesn't work with tornado >= 3, which we also require.

This results in clearer error messages when pyzmq is too old.


Carreau:
+1

Carreau:
Travis is happy. Pressing the green button.

PR #7700, opened 2015-02-05
Don't stop propagation in dashboard

mathieu1:
Stopping click propagation should be avoided when used as a hack
for something else, as it interacts poorly with the open/close
mechanics of Bootstrap menus (amongst others).
(more [here](http://css-tricks.com/dangers-stopping-event-propagation/))

This code preserves the intended behaviour (toggling the checkbox when
clicking on the row, except for the link) without resorting to `stopPropagation()`


jdfreder:
Thanks @mathieu1 , I wasn't aware that this was considered a bad practice.  Good article.

PR #7705, opened 2015-02-06
Remove powershell from script magic defaults

etgalloway:
The powershell script magic does not execute the cell (powershell.exe does not read from stdin, unlike python, bash, etc.).  Typing `%%powershell` has the same affect as typing `!powershell` -- it starts up powershell within IPython.


takluyver:
Thanks for letting us know!

PR #7706, opened 2015-02-06
make Comm publishing threadsafe

minrk:
If a Comm tries to send a message from another thread, schedule it to be called ASAP in the main thread via IOLoop.add_callback.

ping @jdfreder

### DIFF
--- a/IPython/kernel/comm/comm.py
+++ b/IPython/kernel/comm/comm.py
@@ -3,7 +3,10 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import threading
 import uuid
+
+from zmq.eventloop.ioloop import IOLoop
 
 from IPython.config import LoggingConfigurable
 from IPython.kernel.zmq.kernelbase import Kernel
@@ -13,7 +16,7 @@ from IPython.utils.traitlets import Instance, Unicode, Bytes, Bool, Dict, Any
 
 
 class Comm(LoggingConfigurable):
-    
+    """Class for communicating between a Frontend and a Kernel"""
     # If this is instantiated by a non-IPython kernel, shell will be None
     shell = Instance('IPython.core.interactiveshell.InteractiveShellABC',
                      allow_none=True)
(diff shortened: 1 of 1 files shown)


minrk:
For 4.0, we can consider making a similar check inside Session.send or use an IOLoop-aware Session subclass in the kernel process, but this should address the basic issue of Widgets updating in threads.

jdfreder:
Thanks @minrk .  Today when @jgoad is in the lab, we'll give this a test with the Crazyflie code.  For now, I'll test it against the test case I created.

jdfreder:
Woohoo!  This work great against my test case: http://nbviewer.ipython.org/gist/jdfreder/7051ca96984863d13dd7  I'll have to wait to see how it works against the lab code.

IPython/kernel/comm/comm.py
```diff
@@ -63,6 +66,10 @@ def __init__(self, target_name='', data=None, **kwargs):
     
     def _publish_msg(self, msg_type, data=None, metadata=None, buffers=None, **keys):
         """Helper for sending a comm message on IOPub"""
+        if threading.current_thread().name != 'MainThread' and IOLoop.initialized():
+            # make sure we never send on a zmq socket outside the main IOLoop thread
+            IOLoop.instance().add_callback(lambda : self._publish_msg(msg_type, data, metadata, buffers, **keys))
```
jdfreder: @minrk is there a chance that the callback list will overflow if you are adding callbacks to it quickly?  In other words, does the IOLoop instance do something like this:

```python
def tick(self):
    callback = self.callbacks.pop()
    callback()
```

or does it clear the whole buffer each iteration?

```python
def tick(self):
    while len(self.callbacks) > 0:
        callback = self.callbacks.pop()
        callback()
```

minrk: all callbacks fire each eventloop iteration. For threadsafety reasons, it's more like:

```python
with callback_lock:
    new_callbacks = []
    callbacks = self.callbacks
    self.callbacks = new_callbacks
for cb in callbacks:
    cb() # can schedule new callbacks, which won't fire until the next iteration
```

I don't know what you mean by overflow since it's a Python list, but I don't think there's any problem.

jdfreder: Overflow was a poor choice of words, sorry.  I meant expand until >100% memory usage, but it sounds like that isn't a problem.

minrk: Ah, gotcha. Yeah, it would be easy for callbacks that register callbacks to result in starvation, so IOLoop is careful to avoid that.

PR #7708, opened 2015-02-07
allow_none=False for all trait types but Instance and Type

### LINKED ISSUE #7670: Traitlets: allow_none default value for trait types
For a number of trait types the default value for `allow_none` is `True`, such as `Enum`, `Dict`, `List`, `Instance`. 

It probably makes sense for `Type`, but as explicit is often better than implicit, could be interesting to change this before traitlets are separated out.

This would be a divergence with respect to the `traits` library though. (still I think that given the popularity of IPython, traitlets are much more used).

SylvainCorlay:
Closes #7670 .

### DIFF
--- a/IPython/kernel/tests/test_message_spec.py
+++ b/IPython/kernel/tests/test_message_spec.py
@@ -106,7 +106,7 @@ class MimeBundle(Reference):
 
 class ExecuteReply(Reference):
     execution_count = Integer()
-    status = Enum((u'ok', u'error'))
+    status = Enum((u'ok', u'error'), allow_none=True)
     
     def check(self, d):
         Reference.check(self, d)
@@ -139,7 +139,7 @@ class ArgSpec(Reference):
 
 
 class Status(Reference):
-    execution_state = Enum((u'busy', u'idle', u'starting'))
+    execution_state = Enum((u'busy', u'idle', u'starting'), allow_none=True)
 
 
 class CompleteReply(Reference):
@@ -165,7 +165,7 @@ class KernelInfoReply(Reference):
 
 
 class IsCompleteReply(Reference):
-    status = Enum((u'complete', u'incomplete', u'invalid', u'unknown'))
+    status = Enum((u'complete', u'incomplete', u'invalid', u'unknown'), allow_none=True)
     
     def check(self, d):
         Reference.check(self, d)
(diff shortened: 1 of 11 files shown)


SylvainCorlay:
Regarding the places where I had to add `allow_none=True`. it is interesting to check whether it was expected.

minrk:
Good point. I bet it's not all of them.

Thanks for this!

IPython/kernel/tests/test_message_spec.py
```diff
@@ -187,7 +187,7 @@ class ExecuteInput(Reference):
 
 
 class Stream(Reference):
-    name = Enum((u'stdout', u'stderr'))
+    name = Enum((u'stdout', u'stderr'), allow_none=True)
```
minrk: I think all of these here should all have `allow_none=False`. I'm guessing that means a default_value must be specified explicitly.

SylvainCorlay: Yeah, I was just allowing None wherever my change was causing issues. I will add default values.

SylvainCorlay:
Done, I have replaced the allow_none=True with default values. The chosen default value is the first value proposed in the Enum. Travis is happy.

SylvainCorlay:
When #7770 is merged, it will become sensible to have `allow_none=False` for `Instance` and `Type` as well, since unused default values will not be instantiated. 

Then, it would be nice to not have any `allow_none=True` by default in any trait type, but I will make it a different PR because there are much more cases in IPython where it will be required to forcibly set `allow_none=True` for Type or Instance trait type.

SylvainCorlay:
Removed some redundant `allow_none=False` now that it is the default.

takluyver:
I think this might be causing some failures in IPython.parallel: "The 'stop_data' trait of a LocalEngineSetLauncher instance must be a dict, but a value of class 'NoneType' (i.e. None) was specified."

minrk:
Thanks, I'll make sure the parallel tests pass and open a PR with the few remaining changes. This was one part, but there's another as well.

SylvainCorlay:
Sorry I missed this one. On this regard, after  the default value validation PR is merged, I would like to do the same with `Type` and `Instance` (have `allow_none=False` by default).
I will make sure to run the parallel tests.

PR #7709, opened 2015-02-08
JSController cleanup: support case where no server is started

SylvainCorlay:
When a URL is provided to the constructor of `JSController`, no server is starter. Although the `cleanup` function attempts to shut it down resulting in an error.


takluyver:
Looks fine to me.

PR #7718, opened 2015-02-09
Disable output contents from overflowing page width.

### LINKED ISSUE #7701: Image overflow in IPython Notebook 3 (beta)

code
overflow

plot

jdfreder:
A scrollbar will appear when the output is too wide.

closes #7701
alternative to #7721

### DIFF
--- a/IPython/html/static/notebook/less/outputarea.less
+++ b/IPython/html/static/notebook/less/outputarea.less
@@ -92,6 +92,8 @@ div.output_area pre {
 /* This class is for the output subarea inside the output_area and after
    the prompt div. */
 div.output_subarea {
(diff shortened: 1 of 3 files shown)


jdfreder:
An alternate solution, which I attempted but didn't have much luck with, would be to make the page width resize to the output contents width.  This isn't as easy to do as it sounds because the page uses the flexible box model while the output contents doesn't.

jdfreder:
TODO: Test with ctrl-p print.

jdfreder:
Closing in favor of #7721

jdfreder:
Reopened since #7721 was unmerged.

PR #7720, opened 2015-02-09
Readonly indicator.

### LINKED ISSUE #7629: Make read-only clearer.
Not to annoy user, I suggest a warning on first edit.

I got surprised using google drive where the file was read-only a few time. 
Especially since autosaved is disabled in read-only, so you don't especially notice that you might 
loose work.

jdfreder:
screen shot 2015-02-09 at 11 11 45 am

screen shot 2015-02-09 at 11 11 55 am

closes #7629

### DIFF
--- a/IPython/html/templates/notebook.html
+++ b/IPython/html/templates/notebook.html
@@ -62,7 +62,7 @@ data-notebook-path="{{notebook_path}}"
               <span class="kernel_indicator_name">Kernel</span>
               <i id="kernel_indicator_icon"></i>
             </p>
-            <i id="readonly-indicator" class="navbar-text">
+            <i id="readonly-indicator" class="navbar-text" title='This notebook is read-only'>
                 <span class="fa-stack">
                     <i class="fa fa-save fa-stack-1x"></i>
                     <i class="fa fa-ban fa-stack-2x text-danger"></i>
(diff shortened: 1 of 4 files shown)


jdfreder:
@ellisonbg design review.  @Carreau who reported the issue.

jdfreder:
@Carreau this PR doesn't add a one time warning dialog though...  You may still want that.

Carreau:
> this PR doesn't add a one time warning dialog though... You may still want that.

No I think the icon is visually enough. If we show a warning that should be on first modif, which would be painful to do I guess. 

+1

IPython/html/templates/notebook.html
```diff
@@ -62,6 +62,12 @@
               <span class="kernel_indicator_name">Kernel</span>
               <i id="kernel_indicator_icon"></i>
             </p>
+            <i id="readonly-indicator" class="navbar-text">
```
Carreau: `title='read-only notebook'` ?

jdfreder: Yeah good point.

jdfreder:
@Carreau tooltip added.

minrk:
Makes sense. @jdfreder can you make the indicator start hidden, so it isn't shown while notebooks are loading?

jdfreder:
> . @jdfreder can you make the indicator start hidden, so it isn't shown while notebooks are loading?

Yeah, I can do that.  I actually thought of that and intentionally didn't do it since you actually can't save during that period since the notebook hasn't loaded completely.  But I have objection to changing it, seeing that it could be annoying to see that icon the 99% of time that you open a notebook (since opening a notebook that is actually read-only is probably rare).

jdfreder:
@minrk last commit hides it by default.

minrk:
Thanks, :+1:

PR #7725, opened 2015-02-10
Adding font-awesome icon to button and toggle button

SylvainCorlay:
Simply adding a font-awesome icon to button and toggle button.

### DIFF
--- a/IPython/html/static/widgets/js/widget_bool.js
+++ b/IPython/html/static/widgets/js/widget_bool.js
@@ -129,10 +129,11 @@ define([
 
                 var description = this.model.get('description');
                 this.$el.attr("title", this.model.get("tooltip"));
-                if (description.trim().length === 0) {
+                this.$el.text(description);
+                var icon = this.model.get("icon");
+                $('<i class="fa"></i>').prependTo(this.$el).addClass(icon);
+                if (description.trim().length === 0 && icon.trim().length ===0) {
                     this.$el.html("&nbsp;"); // Preserve button height
-                } else {
-                    this.$el.text(description);
                 }
             }
             return ToggleButtonView.__super__.update.apply(this);
--- a/IPython/html/static/widgets/js/widget_bool.js
+++ b/IPython/html/static/widgets/js/widget_bool.js
@@ -59,10 +59,9 @@ define([
             this.$checkbox.prop('checked', this.model.get('value'));
 
             if (options === undefined || options.updated_view != this) {
-                var disabled = this.model.get('disabled');
-                this.$checkbox.prop('disabled', disabled);
+                this.$checkbox.prop("disabled", this.model.get("disabled"));
 
-                var description = this.model.get('description');
+                var description = this.model.get("description");
                 if (description.trim().length === 0) {
                     this.$label.hide();
                 } else {
@@ -113,7 +112,7 @@ define([
             /**
              * Update the contents of this view
              *
-             * Called when the model is changed.  The model may have been 
+             * Called when the model is changed. The model may have been 
              * changed by another view or by a state update from the back-end.
              */
             if (this.model.get('value')) {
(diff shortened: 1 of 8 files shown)


SylvainCorlay:
@jdfreder is it simple enough for 3.0?

jdfreder:
Even though I do think it's simple, I think like @fperez said in the meeting, we should really avoid adding anything new so close to our RC.

jdfreder:
Although it doesn't hurt to ping others to see if they think it should make it in 3.0 , @minrk @Carreau @takluyver @ellisonbg @jasongrout ?

takluyver:
+0.5 to leaving it for now. Even simple changes can introduce unforseen problems.

SylvainCorlay:
I would use it right away: I am separating out our work into separately installable nbextensions, and some packages have these very simple widgets as their only common dependencies.

jasongrout:
This is a very simple change.  If it's tested, +1.

IPython/html/static/widgets/js/widget_bool.js
```diff
@@ -128,11 +128,12 @@ define([
                 this.$el.prop('disabled', disabled);
 
                 var description = this.model.get('description');
+                var faclass = this.model.get("icon");
                 this.$el.attr("title", this.model.get("tooltip"));
-                if (description.trim().length === 0) {
+                this.$el
+                   .html('<i class="fa '.concat(faclass, '"></i>', description));
```
minrk: .html calls should never take a user-defined value. Always use $.text. Also need to make sure that faclass. This adds two separate places for users to create incorrect HTML (faclass and description).

SylvainCorlay: ok, I will use `$.text` and `$.addClass - $.removeClass.`

minrk:
While this seems very simple, even a small change like this can introduce problems, especially in js where we aren't as good at spotting what side effects things can have. This one, for instance, disables HTML escaping of description text.

fperez:
We're trying to get an RC out, so I'm inclined to push back on this.

Unless @minrk thinks we really need a *beta* instead of an RC, then it's a different story. But if we really want to call it an RC, we shouldn't be touching code we don't absolutely *have* to...

Sorry @SylvainCorlay, I know it's not what you want to hear :) But as release day approaches, the job of the release team becomes making everyone unhappy ;)

minrk:
I also marked it as 3.1 instead of 4.0. I think little things like this are fine in minor releases, especially the early ones.

SylvainCorlay:
OK, too bad for 3.0. Thanks for putting these small changes for a minor release. Addressed your comment @minrk .

IPython/html/static/widgets/js/widget_bool.js
```diff
@@ -129,10 +129,11 @@ define([
...
                 var description = this.model.get('description');
                 this.$el.attr("title", this.model.get("tooltip"));
-                if (description.trim().length === 0) {
+                this.$el.text(description);
+                var icon = this.model.get("icon");
+                $('<i class="fa"></i>').prependTo(this.$el).addClass(icon);
+                if (description.trim().length === 0 && icon.trim().length ===0) {
                     this.$el.html("&nbsp;"); // Preserve button height
-                } else {
-                    this.$el.text(description);
```
minrk: It seems like this else block still makes sense. This change currently does $el.text unconditionally, even though the if statement here will override the result.

IPython/html/static/widgets/js/widget_bool.js
```diff
@@ -129,10 +129,11 @@ define([
 
                 var description = this.model.get('description');
                 this.$el.attr("title", this.model.get("tooltip"));
-                if (description.trim().length === 0) {
+                this.$el.text(description);
+                var icon = this.model.get("icon");
```
minrk: Is there any chance that `icon` will be undefined or null instead of a string? If so, that may need to be handled below, where it's assumed that `icon` is always a string.

minrk: I see from the Python side that icon can't reasonably be anything other than a string, so this assumption should be okay.

IPython/html/static/widgets/js/widget_bool.js
jdfreder: @SylvainCorlay could you just move lines 132-134 into the else?  And `var icon = this.model.get("icon") || '';` just to be defensive?  Then I think this is good to go!

IPython/html/static/widgets/js/widget_button.js
```diff
@@ -32,12 +32,12 @@ define([
...
             var description = this.model.get('description');
             this.$el.attr("title", this.model.get("tooltip"));
-            if (description.length === 0) {
+            this.$el.text(description);
+            var icon = this.model.get("icon");
+            $('<i class="fa"></i>').prependTo(this.$el).addClass(icon);
+            if (description.trim().length === 0 && icon.trim().length ===0) {
                 this.$el.html("&nbsp;"); // Preserve button height
-            } else {
-                this.$el.text(description);
```
jdfreder: Same here :)

SylvainCorlay:
Done. Travis seems happy.

jdfreder:
Thanks @SylvainCorlay !  Cool idea-

PR #7727, opened 2015-02-10
Toggle viewing files/folders/notebooks

mathieu1:
Intended as a complement or replacement for #7681:
add toggles in File tab of Dashboard for viewing/hiding each type in the list.
- Selected items that become hidden get deselected.
- Can coexist with a simple select all/none button (no dropdown menu needed,
  the only lost functionality would be to select running notebooks)
- No persistance of toggle state

Going forward with this, remains TODO:
- [ ] The select all/none/... code should be edited so as not to select
  hidden elements.
- [ ] Decide on best positionning of the buttons

Possible extras:
- search button, revealing a "search" text field (e.g. to find ".py" files)

Behavior:
toggle-view

### DIFF
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -128,6 +128,9 @@ define([
             $('#tree-selector .select-files').click($.proxy(this.select_files, this));
             $('#tree-selector .select-directories').click($.proxy(this.select_directories, this));
             $('#tree-selector .deselect-all').click($.proxy(this.deselect_all, this));
+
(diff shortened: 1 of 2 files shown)


Carreau:
For info, Need rebase.

mathieu1:
I'm fine with rebasing, but is there actually interest in this feature? It was more or less meant as a kind of substitute for #7681 which got merged (although this PR might also appeal to the 2.x nostalgic who are bothered by files showing up in the list).

takluyver:
That makes sense. I'll close this for now, we can reopen if we decide we want this in addition to #7681.

PR #7728, opened 2015-02-10
Stay in edit mode if tooltip is open

### LINKED ISSUE #7723: Caret disappears after escaping tooltip
There is a small glitch after triggering a tooltip  in which pressing Esc to dismiss the tooltip also escapes the caret that was placed after the written text. Example:

1. Write in a cell the word "map", then press `Shift + Tab` (a tooltip will appear and the caret remains)
2. Press `Esc` ( tooltip disappears and also the caret after the word "map") 
3. Now we are in "command mode".

Ipython 2 didn't have this behavior and intuitively makes sense to press  `Esc` to remove the tooltip and return to the text.

This is happening in the master branch using Chrome (40.0.2214.111) and Firefox (35.0.1).

mathieu1:
Do not go to command mode on pressing 'Esc' if tooltip is open (same behaviour as pressing 'Esc' while completion list is open)
closes #7723

### DIFF
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -65,6 +65,13 @@ define(function(require){
                 env.notebook.command_mode();
             }
         },
+        'close-tooltip-or-go-to-command-mode': {
+            help    : 'command mode',
+            help_index : 'aa',
+            handler : function (env) {
+                env.notebook.close_tooltip_or_command_mode();
+            }
+        },
         'split-cell-at-cursor': {
             help    : 'split cell',
             help_index : 'ea',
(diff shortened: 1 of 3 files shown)


minrk:
I really don't like the idea of a single action that's "do one action or maybe another." Is there a reason the tooltip can't solve this in the same way as the completion?

Carreau:
Completion get focus, tooltip does not.

mathieu1:
Right. While it is certainly out of the question to change the focus behaviour, IMO:
- from the user standpoint, focus is irrelevant. 
- It's not really "do an action or maybe another", there's an intuitively clear stack of things that `Esc` gets you out of: first the tooltip, then the edit mode.

minrk:
I opened #7733, which is a simpler fix for this. We already had code to capture this event, it just wasn't preventing it from bubbling to the keyboard handler properly.

minrk:
Closed in favor of #7733. Thanks, @mathieu1!

PR #7734, opened 2015-02-11
scrolling workaround for Firefox

### LINKED ISSUE #7546: Code cells can end up wider than the white 'page' background
Made my window a bit narrower, and this happened:

screenshot from 2015-01-22 16 59 52

### LINKED ISSUE #7724: Long text overflows in Firefox
I'm sure this has been reported before but I can't find the exact same issue.

When you write a very long text in a cell in the Ipython Notebook  (3 dev and 2.3.1) in Firefox (Firefox 35.0.1 in Ubuntu 14.10), this happens:

screenshot from 2015-02-09 22 14 01

In Chrome, it works as it should:

screenshot from 2015-02-09 22 12 08

This is similar to the output given by rossant in the issue #5695, although in that case, it is the output that overflows.

minrk:
setting overflow-x: hidden on Firefox prevents long input from growing the cell ever wider on Firefox.

CSS is only applied on Firefox

workaround is c/o @rsmith31415

closes #7724
closes #7546

### DIFF
--- a/IPython/html/static/notebook/less/cell.less
+++ b/IPython/html/static/notebook/less/cell.less
@@ -54,6 +54,14 @@ div.inner_cell {
     .box-flex1();
 }
 
+@-moz-document url-prefix() {
+    div.inner_cell {
+        // hack around FF bug causing cell to expand when lines are long
+        // instead of scrolling
+        overflow-x: hidden;
+    }
+}
+
 /* input_area and input_prompt must match in top border and margin for alignment */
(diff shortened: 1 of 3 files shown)


takluyver:
Should also close #7546.

minrk:
Thanks, I was looking for that one but couldn't find it.

takluyver:
Confirmed I see expected behaviour with this branch.

PR #7740, opened 2015-02-11
fix Radio margin

dongweiming:
after:

### DIFF
--- a/IPython/html/static/widgets/less/widgets.less
+++ b/IPython/html/static/widgets/less/widgets.less
@@ -266,3 +266,7 @@
         margin-top: 0px;
     }
 }
+
+.widget-radio {
+    margin-left: 20px;
+}


Carreau:
Screenshot seem legit. Any reason for 26px, or eyeballed ?

@jdfreder thought ? 

Side question: What's "Expand" thing ?

jdfreder:
I can verify that something has gone awry, I'm trying to figure out why this is happening.  It looks like the specificity got screwed up somehow.

jdfreder:
It may be related to the Bootstrap 3.3 upgrade- https://github.com/ipython/ipython-components/commit/39d3b273c7c16012f46bc3310e85d92d6d00788f#diff-724cc5a431bf2e1a8da1d19c22d21d2dL207 .

dongweiming:
@Carreau
` .widget-area .widget-subarea`'s `margin-left` is 6px;
but, `.radio input[type="radio"]`'s `margin-left` is -20px;

so this need 26px

dongweiming:
@Carreau 'Expand' is a custom extension in my [IPython-notebook-extensions](https://github.com/dongweiming/IPython-notebook-extensions/blob/add_toggle_full_textarea/usability/toggle_full_textarea/main.js)

minrk:
I'm not sure why the subarea's 6px should be taken into account, since this element is already inside the subarea. Shouldn't it just be 20px, then, to reverse the left-shift of the radio button? That seems to align better with other elements.

dongweiming:
@minrk yes. you are right

dongweiming:
@minrk how about this? if i only reverse the left-shift of the radio button. the content in `input[type="radio"]` will be overwritten

minrk:
@dongweiming thanks!

SylvainCorlay:
Shouln't there be the generated css committed with the less?

minrk:
@SylvainCorlay yes, crap. Should have noticed that. Fixed in master.

PR #7743, opened 2015-02-11
DEV: Add toggles for reraising failures during extension loading

ssanderson:
When deploying an IPython extension for use by other people, I'd rather have the server fail loudly instead of ignoring the error with a warning and carrying on.  This adds a `exit_on_extension_load_failure` flag to InteractiveShellApp that makes it barf if any extensions fail to load.

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -139,7 +139,13 @@ class InteractiveShellApp(Configurable):
     extra_extension = Unicode('', config=True,
         help="dotted module name of an IPython extension to load."
     )
-    
+
+    reraise_ipython_extension_failures = Bool(
+        False,
+        config=True,
+        help="If True, exit on failure to load any extensions.",
+    )
+
     # Extensions that are always loaded (not configurable)
     default_extensions = List(Unicode, [u'storemagic'], config=False)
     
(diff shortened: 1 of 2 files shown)


ssanderson:
Suggestions for less verbose names welcome.

ssanderson:
I don't *think* the test failures here are mine.

ssanderson:
I added an analogous toggle for server extensions and renamed both to be of the form: `reraise_*_extension_failures`.

takluyver:
I think these names are fine, but this needs a rebase now.

I'll merge this in a day or two unless anyone objects (assuming rebase and tests go fine).

ssanderson:
Rebased (conflict was whitespace-only).

takluyver:
Thanks @ssanderson . Tests passing.

IPython/core/shellapp.py
```diff
@@ -139,7 +139,18 @@ class InteractiveShellApp(Configurable):
     extra_extension = Unicode('', config=True,
         help="dotted module name of an IPython extension to load."
     )
-    
+
+    def _extra_extension_changed(self, name, old, new):
```
ssanderson: I think I picked up an extraneous diff in my rebase...not sure where this came from.

ssanderson: Will fix this in the morning, looks like I accidentally reverted 34a0e6b8867a4fd448c721d006e8253e8cec1089

takluyver: OK, ping us when you've cleared it up.

ssanderson:
@takluyver I think this is in the correct state now.

Issue #7742, opened 2015-02-11
how do I convert a "pure" Ijulia notebook v3 in a notebook v4 'Julia' ?


stonebig:
Hi,
I'm on ipython3.0b1, with kernels python3 and Julia.
I want to play with this notebook "ijulia-made":
http://nbviewer.ipython.org/github/JuliaLang/Interact.jl/blob/master/doc/notebooks/02-Widgets%20Overview.ipynb

But I don't see how to pre-convert it in "v04" 'Julia' with the Ipython GUI....
 
==> it opens in Python3 kernel ... shame.

(it doesn't seem to work if I copy/paste by hand, maybe Ijulia interact is not compatible with Ipython3)

takluyver:
Open it and go to the kernel menu, and select the Julia kernel. That should
change the kernel, the highlighting, and the notebook metadata.

stonebig:
aouh ! so simple. thanks !

stonebig:
still, i'm worried Julia 'Interact' module doesn't work anymore. Could you have a look on the given notebook ?

Carreau:
Julia notebooks do have `language=julia` in their metadata. I thought we were migrating that ?

takluyver:
It should be easy to write a script to do so, but I don't think we have done that.

@stonebig widgets may have changed quite a bit, possibly IJulia will need to update things for that to work. @jdfreder should be able to give more info.

Carreau:
> It should be easy to write a script to do so, but I don't think we have done that.

I was thinking it was either in JS, or in Nbformat.
We could maybe use the dialog when there is no kernelspec matching to prompt for a kernel ?

jdfreder:
> @jdfreder should be able to give more info.

https://github.com/ipython/ipython/wiki/IPython-3.0-comm-and-widget-migration-document

Carreau:
Can you try #7747 for automatically recognizing Julia notebooks ?

minrk:
closed by #7747

PR #7747, opened 2015-02-11
Select language based on old notebook meta.lang.

### LINKED ISSUE #7744: kernel_selector.set_kenel('string') miss behaving.
`set_kernel`, is supposed to be able to get a string but if the string does not match a spec, then the spec not found missbehave and show "could not find kernel _undefined_" as it tries to get "display_name" that is in fact undefined. 

Minor, but optimistically tagging 3.0

Carreau:
Also fix UI when selecting kernels by string.

Fix #7744 and #7742

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -169,7 +169,8 @@ define([
         var that = this;
         if (typeof selected === 'string') {
             selected = {
-                name: selected
+                name: selected,
+                language: selected
             };
         }
         if (this._loaded) {
@@ -246,7 +247,7 @@ define([
         
         var body = $("<form>").addClass("form-inline").append(
             $("<span>").text(
-                "I couldn't find a kernel matching " + (data.selected.display_name || data.name) + "." +
+                "I couldn't find a kernel matching " + (data.selected.display_name || data.selected.name) + "." +
                 " Please select a kernel:"
             )
         ).append(select);
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -169,8 +169,7 @@ define([
         var that = this;
         if (typeof selected === 'string') {
             selected = {
-                name: selected,
-                language: selected
+                name: selected
             };
         }
         if (this._loaded) {
--- a/IPython/html/static/notebook/js/notebook.js
+++ b/IPython/html/static/notebook/js/notebook.js
@@ -2209,13 +2209,20 @@ define(function (require) {
         } else if (this.nbformat_minor < nbmodel.nbformat_minor) {
             this.nbformat_minor = nbmodel.nbformat_minor;
         }
-        
+
         if (this.session === null) {
             var kernel_name = utils.get_url_param('kernel_name');
             if (kernel_name) {
                 this.kernel_selector.set_kernel(kernel_name);
             } else if (this.metadata.kernelspec) {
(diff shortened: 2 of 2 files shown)


IPython/html/static/notebook/js/kernelselector.js
```diff
@@ -169,7 +169,8 @@ define([
         var that = this;
         if (typeof selected === 'string') {
             selected = {
-                name: selected
+                name: selected,
+                language: selected
```
minrk: These shouldn't have the same value, should they?

IPython/html/static/notebook/js/notebook.js
```diff
@@ -2209,13 +2209,20 @@ define(function (require) {
...
         if (this.session === null) {
             var kernel_name = utils.get_url_param('kernel_name');
             if (kernel_name) {
                 this.kernel_selector.set_kernel(kernel_name);
             } else if (this.metadata.kernelspec) {
                 this.kernel_selector.set_kernel(this.metadata.kernelspec);
+            } else if (this.metadata.language) {
+                // compat with IJulia, IHaskell, and other early kernels
+                // adopters that where seting a language metadata.
+                this.kernel_selector.set_kernel(this.metadata.language);
```
minrk: When a single string is used, it must be the kernelspec name, not the language. Since this is the language, it should be passed as:

```javascript
set_kernel({
  name: "Unknown", // or something indicating that it's not a value that makes sense
  language: this.metadata.language,
});
```

minrk: Or "Unknown [language] kernel"

IPython/html/static/notebook/js/kernelselector.js
Carreau: Well, what would you do ?  There are assumption that selected has 'name' and 'language' a few places further in the code, and would require lots of refactoring.  I assume that if you say `set_kenel('julia')` you can mean both language and name ?

minrk: Answered below. When given as a single string, this should only be the kernelspec name. If you want to specify different information, use a dict.

IPython/html/static/notebook/js/notebook.js
Carreau: Ok, will try this fix.

Carreau:
Done.

Hopefully Travis will be happy this time.

IPython/html/static/notebook/js/notebook.js
```diff
@@ -2209,13 +2209,23 @@ define(function (require) {
...
+
         if (this.session === null) {
             var kernel_name = utils.get_url_param('kernel_name');
             if (kernel_name) {
                 this.kernel_selector.set_kernel(kernel_name);
             } else if (this.metadata.kernelspec) {
                 this.kernel_selector.set_kernel(this.metadata.kernelspec);
+            } else if (this.metadata.language) {
+                // compat with IJulia, IHaskell, and other early kernels
+                // adopters that where seting a language metadata.
```
minrk: setting

IPython/html/static/notebook/js/notebook.js
```diff
@@ -2209,13 +2209,23 @@ define(function (require) {
...
+                this.kernel_selector.set_kernel({
+                    name: "(No name)",
+                    language: this.metadata.language
+                  });
+                // this shoudl be stored in kspec now, delete it.
```
minrk: should

Carreau:
Fix `settings` and `should`.

minrk:
Thanks, will merge when Travis gives the green light.

PR #7748, opened 2015-02-11
add no-op every 5s on Windows

### LINKED ISSUE #7741: Notebook: Long delay before closing/server stop on Windows
I'm on Windows with ipython 2.4.0

The console says "Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).". But doing so takes over a minute on my machine to actually exit! Just measured and it was 102 seconds.

It's faster to just kill python.exe. Which is ugly and can't be the expected behavior. This is not new for me, but I couldn't find related issues, so what's up with that?

minrk:
signals don't interrupt the inner loop on Windows, so pump the eventloop every 5s

closes #7741

### DIFF
--- a/IPython/html/notebookapp.py
+++ b/IPython/html/notebookapp.py
@@ -1061,6 +1061,11 @@ class NotebookApp(BaseIPythonApplication):
                 threading.Thread(target=b).start()
         
         self.io_loop = ioloop.IOLoop.current()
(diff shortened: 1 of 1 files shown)


Carreau:
If it helps, +1.

Carreau:
According to comment on issues, merging.

PR #7752, opened 2015-02-11
minor doc improvement on nbextension

Carreau:
I think that's more explicit of what this does.


IPython/html/nbextensions.py
```diff
@@ -159,8 +159,8 @@ def install_nbextension(path, overwrite=False, symlink=False, user=False, prefix
         Whether to install to the user's .ipython/nbextensions directory.
         Otherwise do a system-wide install (e.g. /usr/local/share/jupyter/nbextensions).
     prefix : str [optional]
-        Specify install prefix, if it should differ from default (e.g. /usr/local).
```
takluyver: I agree with the change on the line below, but -0.5 to it on this line.

Carreau: > I agree with the change on the line below, but -0.5 to it on this line.

Which change on which line ? :-P

PR #7754, opened 2015-02-11
disable screenKeys in term.js

### LINKED ISSUE #7753: `Ctrl-a` keybinding should go to term.js terminal
The `C-a` keybinding is the standard bash 'go to start of line' keybinding.  I don't know how easy it is to make sure we redirect it so the terminal (in the notebook environment) sees it. I suspect it's being captured by the browser as 'select all'.

I'm flagging this as 3.0 as I think it's a big usability issue. BUT, if it's hard to do, let's punt and retarget for later. I only think we should fix now if the fix is very straightforward.

ps - we probably should have a label for the notebook terminal component.  Not sure what we want to name it (same for the notebook editor)

minrk:
these capture ctrl-a and other keyboard shortcuts

closes #7753

### DIFF
--- a/IPython/html/static/terminal/js/terminado.js
+++ b/IPython/html/static/terminal/js/terminado.js
@@ -6,7 +6,7 @@ define ([], function() {
         var term = new Terminal({
           cols: size.cols,
           rows: size.rows,
-          screenKeys: true,
+          screenKeys: false,
           useStyle: false
         });
         ws.onopen = function(event) {


fperez:
Great, confirming the fix works. Merging now.

Thanks!

kylekyle:
Without `screenKeys` it can be really difficult to copy and paste to and from the web terminal. For people relegated to using the web terminal, no copy and paste can make it really hard to use.

PR #7756, opened 2015-02-11
pass common config to contents manager

### LINKED ISSUE #7755: Content manager should get passed comon config.
Right now it get passed notebook, or tree config depending on the cases.

Carreau:
Closes #7755

### DIFF
--- a/IPython/html/static/edit/js/main.js
+++ b/IPython/html/static/edit/js/main.js
@@ -27,15 +27,19 @@ require([
     savewidget,
     notificationarea
     ){
+    "use strict";
     page = new page.Page();
 
     var base_url = utils.get_body_data('baseUrl');
     var file_path = utils.get_body_data('filePath');
-    contents = new contents.Contents({base_url: base_url});
     var config = new configmod.ConfigSection('edit', {base_url: base_url});
     config.load();
     var common_config = new configmod.ConfigSection('common', {base_url: base_url});
     common_config.load();
+    contents = new contents.Contents({
+        base_url: base_url,
+        common_config: common_config
+    });
     
     var editor = new editmod.Editor('#texteditor-container', {
         base_url: base_url,
--- a/IPython/html/static/notebook/js/main.js
+++ b/IPython/html/static/notebook/js/main.js
@@ -75,9 +75,10 @@ require([
     var save_widget = new savewidget.SaveWidget('span#save_widget', {
         events: events, 
         keyboard_manager: keyboard_manager});
-    var contents = new contents.Contents($.extend({
-        events: events, config:config_section},
-        common_options));
+    var contents = new contents.Contents({
+          base_url: common_options.base_url,
+          common_config: common_config
+        });
     var notebook = new notebook.Notebook('div#notebook', $.extend({
         events: events,
         keyboard_manager: keyboard_manager,
--- a/IPython/html/static/tree/js/main.js
+++ b/IPython/html/static/tree/js/main.js
@@ -50,13 +50,14 @@ require([
     common_options.config = cfg;
     var common_config = new config.ConfigSection('common', common_options);
     common_config.load();
-    
+
     var session_list = new sesssionlist.SesssionList($.extend({
-        events: events}, 
-        common_options));
-    var contents = new contents_service.Contents($.extend({
         events: events},
         common_options));
+    var contents = new contents_service.Contents({
+        base_url: common_options.base_url,
+        common_config: common_config
+    });
     var notebook_list = new notebooklist.NotebookList('#notebook_list', $.extend({
         contents: contents,
         session_list:  session_list}, 
@@ -98,16 +99,16 @@ require([
         if (terminal_list) {
             terminal_list.load_terminals();
         }
-    }
+    };
 
     var enable_autorefresh = function(){
         /**
          *refresh immediately , then start interval
          */
-        var now = new Date()
+        var now = new Date();
 
         if (now - _last_refresh < IPython.min_delta_refresh*1000){
-            console.log("Reenabling autorefresh too close to last tree refresh, not refreshing immediately again.")
+            console.log("Reenabling autorefresh too close to last tree refresh, not refreshing immediately again.");
         } else {
             _refresh_list();
         }


Carreau:
Oops, overcommited, shoudl be fixed.

Carreau:
I would like that to move https://github.com/jupyter/jupyterdrive forward if possible.

minrk:
If common_config is a thing, it makes sense to me for it to be part of the common options we pass around.

IPython/html/static/edit/js/main.js
```diff
@@ -31,11 +31,14 @@ require([
 
     var base_url = utils.get_body_data('baseUrl');
     var file_path = utils.get_body_data('filePath');
-    contents = new contents.Contents({base_url: base_url});
     var config = new configmod.ConfigSection('edit', {base_url: base_url});
     config.load();
     var common_config = new configmod.ConfigSection('common', {base_url: base_url});
     common_config.load();
+    contents = new contents.Contents({base_url:base_url, 
+                                        config:config,
```
takluyver: Is there any reason to pass `config` as well? Our own contents class doesn't use any config, and in general I don't want to be passing config to an API module except where it's really necessary.

Carreau: it is already passed in `tree/js/main` so I made it consistent. I can try not to pass local config, but pass common one.

takluyver: As I mentioned, I'm pretty sure that's unintentional, just a side effect of reusing the common_options object. The standard IPython contents module that talks to our REST API does not use any config at all, so if you're not using the other config object, there's no reason for it to be there.

minrk: Passing just common config also seems reasonable - contents shouldn't be configured in an application-specific way.

Carreau: Ok, I'l uniformize the other way by using `common_option.config = common_config` and introduce common_option also in `editor/js/main`

takluyver: `common_option.config = common_config` may cause problems for the other things we pass common_options to. AIUI, `common_options` is just a convenience thing for 'here are some things we need to pass to several constructors', so if it doesn't make sense to use those things for the Contents constructor, let's just make a separate options dict for it.

minrk: No, not common_options.confg, that should still be the application config. contents is a special case that it's config should be cross-application.

Carreau:
Done through `common_option.common_config = common_config`.

Carreau:
nop. Editor don't load.

Carreau:
> nop. Editor don't load.

removed `Editor({config:config})` by mistake, should be fixed.

IPython/html/static/edit/js/main.js
```diff
@@ -29,22 +29,29 @@ require([
...
-    var base_url = utils.get_body_data('baseUrl');
-    var file_path = utils.get_body_data('filePath');
-    contents = new contents.Contents({base_url: base_url});
-    var config = new configmod.ConfigSection('edit', {base_url: base_url});
+    var common_options = {
+        ws_url : utils.get_body_data("wsUrl"),
+        base_url : utils.get_body_data("baseUrl"),
+        file_path : utils.get_body_data('filePath'),
+        notebook_path : utils.get_body_data("notebookPath"),
+        notebook_name : utils.get_body_data('notebookName')
```
minrk: Why did notebook_etc get added to the edit options? This doesn't make sense to me.

IPython/html/static/edit/js/main.js
```diff
@@ -29,22 +29,29 @@ require([
...
+    };
+
+    // var base_url = utils.get_body_data('baseUrl');
+    // var file_path = utils.get_body_data('filePath');
+    var config = new configmod.ConfigSection('edit', common_options);
```
takluyver: I'm not sure this is the direction we should be going in - common_options is a convenient thing, but it's better to pass constructors only the set of options they use where that's practical. So I'd like to use common_options less, not more.

takluyver: (e.g. if we hadn't been passing common_options to Contents, we wouldn't have run into this confusion in the first place)

IPython/html/static/edit/js/main.js
Carreau: Because for whatever reason it is used in `tree` (and `notebook`, but it make sens on `notebook`) , so I guessed common_options should be consistent in all the pages.

Carreau:
> If common_config is a thing, it makes sense to me for it to be part of the common options we pass around.

> I'm not sure this is the direction we should be going in - common_options is a convenient thing, but it's better to pass constructors only the set of options they use where that's practical. So I'd like to use common_options less, not more.

So which one of the two ? I'm fine with either.

IPython/html/static/edit/js/main.js
takluyver: common_options is *not* an API thing. There's no need for it to be consistent. It is purely a convenient thing when we want to reuse a few keyword arguments.

This especially doesn't make sense because the editor won't have a notebook path or name.

minrk: I don't agree with that. Adding unused arguments that should never be used just for consistency with other pages doesn't seem like a sound reason.

takluyver:
The things that only take one parameter (like ConfigSection) should definitely get an explicit `{base_url: base_url}`, not common_options. I spoke to Min about this, he agrees.

Things that make no sense on that page, like notebook_path on an editor page, definitely shouldn't be there. There's no need for it to be consistent between pages.

For other stuff, it's just a collection of keyword arguments. If it makes sense to pass the same object to several things, as it might for common_config, it can go in.

I'm now considering getting rid of common_options entirely, because it seems to be causing too much confusion, especially where we pass options to things that don't need or use them. But that's probably not a 3.0 change.

Carreau:
OK, redone from scratch.
 - Remove passing `events`, which is unused, 
 - and added just `common_config: common_config` where necessary.
 -  Don't pass `config: config_section` to `Contents`
 -  `"use strict";` and semicolon.

IPython/html/static/notebook/js/main.js
```diff
@@ -76,7 +76,8 @@ require([
         events: events, 
         keyboard_manager: keyboard_manager});
     var contents = new contents.Contents($.extend({
-        events: events, config:config_section},
+          common_config: common_config
```
minrk: did you want to make these contents constructors only take base_url, common_config like above?

IPython/html/static/notebook/js/main.js
```diff
@@ -76,7 +76,8 @@ require([
...
+        },
```
takluyver: Let's explicitly pass `base_url` and `common_config` rather than extending common_options here. It's the same number of lines, and explicit > implicit.

IPython/html/static/tree/js/main.js
```diff
@@ -55,7 +55,8 @@ require([
         events: events}, 
         common_options));
     var contents = new contents_service.Contents($.extend({
-        events: events},
+          common_config: common_config
+        },
         common_options));
```
takluyver: Ditto.

takluyver:
Other than that, looks good.

Carreau:
Ok, constuctor just take to positionnal arguments now.

IPython/html/static/services/contents.js
```diff
@@ -7,7 +7,7 @@ define(function(require) {
     var $ = require('jquery');
     var utils = require('base/js/utils');
 
-    var Contents = function(options) {
+    var Contents = function(base_url, common_config) {
```
takluyver: No, I think it should stay as keyworks arguments (i.e. a dictionary). I just meant pass them explicitly.

Carreau:
> No, I think it should stay as keyworks arguments (i.e. a dictionary). I just meant pass them explicitly.

OK, my mistake, I missunderstood.
Fixed.

takluyver:
OK, I think this looks good now (sorry for the runaround). Will merge once Travis is done.

minrk:
:+1:

PR #7762, opened 2015-02-12
No default value creation for element traits of a List

### LINKED ISSUE #7760: Parameterized traitlets and default value validation
When using a List traitlet with validated elements, the default value of the elements is validated, like in this example:

```Python

class Bar():
    pass

class Foo(HasTraits):
    bars = List(Instance(Bar, allow_none=False))

foo = Foo()
```
which yields a `TraitError` because the default value, None, is invalid. 

A fix would be to do `bars = List(Instance(Bar, tuple(), allow_none=False))`, i.e. to create a  default value. But somehow, I think that there should not be a need to instantiate a default value that is never going to be used!

@minrk should the default value be validated when the traitlet is to be used in a "template" trait like List?

SylvainCorlay:
Fixes  #7760 . @minrk @ssanderson

### DIFF
--- a/IPython/utils/tests/test_traitlets.py
+++ b/IPython/utils/tests/test_traitlets.py
@@ -972,6 +972,19 @@ class TestList(TraitTestBase):
 class Foo(object):
     pass
 
+class NoneInstanceListTrait(HasTraits):
+    
+    value = List(Instance(Foo, allow_none=False))
+
+class TestNoneInstanceList(TraitTestBase):
+
+    obj = NoneInstanceListTrait()
+
+    _default_value = []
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
Besides, I am now overloading the `_resolve_classes` of the element traits for `Tuple` as well.

SylvainCorlay:
Now also resolving element trait names in `Tuple`, which was missing.

minrk:
Nice, thanks. Want to add the test you described in #7760?

minrk:
Excellent, thanks. Will merge once Travis confirms.

SylvainCorlay:
Travis is happy.

PR #7763, opened 2015-02-12
get_body_data: fix for undefined

cgestes:
return undefined instead of "undefined"

useful to fix this line in kernel.js:
    this.ws_url = ws_url || IPython.utils.get_body_data("wsUrl");


Carreau:
Fix seem legit. 
In what case did you had it not working ?

PR #7768, opened 2015-02-13
handle zmq messages that may be processed after websocket is closed

### LINKED ISSUE #7697: Reconnecting error messages with master(-ish)
Version: ee6223ab74eb

My situation is that I set up a remote notebook on a server that I only can reach when in a VPN. Now, my Wifi is shaky for some reason, that's possibly how my VPN get's shaky.
So I needed to close my notebook after a VPN reconnect and the dashboard still reported it as 'running'. When I clicked on the 'running' notebook, I got this error in the console:

```python
RunTimeError: Method not supported by Web Sockets
```

Full error log here:
https://gist.github.com/23ed2e252897d96804a5

Working on OSX 10.9.5 using conda and Python 3.4

minrk:
closes #7697

This actually fixes #7697 twice
1. it prevents an exception from being raised if a zmq message arrives after the websocket has been closed
2. it fixes the handling of any exception raised in websocket handlers, to avoid "Method not supported by Web Sockets" message when tornado tries to send the error. This is done by backporting WebSocketHandler.send_error from tornado 4.1.

### DIFF
--- a/IPython/html/base/zmqhandlers.py
+++ b/IPython/html/base/zmqhandlers.py
@@ -106,12 +106,11 @@ class ZMQStreamHandler(WebSocketHandler):
                 # we can close the connection more gracefully.
                 self.stream.close()
 
-
+    
     def check_origin(self, origin):
         """Check Origin == Host or Access-Control-Allow-Origin.
         
         Tornado >= 4 calls this method automatically, raising 403 if it returns False.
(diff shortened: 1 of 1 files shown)


jdfreder:
I'm not sure of an easy way to test this fix with my single machine, but the code makes sense.  At the time of writing, this is the last PR for 3.0.

rgbkrk:
:tada:

PR #7770, opened 2015-02-13
Do not instantiate traitlet default values if not necessary

### LINKED ISSUE #7769: Trait default values are validated even when not ever used
```Python

class Foo(HasTraits):
    bar = Type(allow_none=False)

foo = Foo(bar=int)
```
yields `The 'bar' trait of a Foo instance must be a subclass of object, but a value of type 'NoneType' (i.e. None) was specified.`

The default value is validated while it is not needed.

SylvainCorlay:
Fixes #7769 . 
This is not very polished but seems to be doing the job. Will not be able to look more into it before the week end.

### DIFF
--- a/IPython/kernel/inprocess/ipkernel.py
+++ b/IPython/kernel/inprocess/ipkernel.py
@@ -52,10 +52,7 @@ class InProcessKernel(IPythonKernel):
     stdin_socket = Instance(DummySocket, ())
 
     def __init__(self, **traits):
-        # When an InteractiveShell is instantiated by our base class, it binds
-        # the current values of sys.stdout and sys.stderr.
-        with self._redirected_io():
-            super(InProcessKernel, self).__init__(**traits)
+        super(InProcessKernel, self).__init__(**traits)
 
         self.iopub_socket.on_trait_change(self._io_dispatch, 'message_sent')
         self.shell.kernel = self
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -356,7 +356,7 @@ class TraitType(object):
         """Create a new instance of the default value."""
         return self.default_value
 
-    def decorate(self, obj):
+    def decorate(self):
         pass
 
     def static_init(self, obj):
@@ -550,7 +550,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                 pass
             else:
                 if isinstance(value, TraitType):
-                    value.decorate(inst)
+                    value.decorate()
                     if key not in kw:
                         value.set_default_value(inst)
 
@@ -845,7 +845,7 @@ class Type(ClassBasedTraitType):
                 value = self._resolve_string(value)
             except ImportError:
                 raise TraitError("The '%s' trait of %s instance must be a type, but "
-                                "%r could not be imported" % (self.name, obj, value))
+                                 "%r could not be imported" % (self.name, obj, value))
         try:
             if issubclass(value, self.klass):
                 return value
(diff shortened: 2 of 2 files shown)


SylvainCorlay:
OK, the failed test checks for something that should not be considered an error if you are fine with this approach.

SylvainCorlay:
Ok, Travis is happy now. This still needs to be polished.

SylvainCorlay:
Rebased to account for the latest changes merged in traitlets, and then simplified a little bit. 
(Btw, the test failure in earlier commit 5d2961a does not look legitimate).

SylvainCorlay:
Just rebased but it seems to make Travis unhappy.

SylvainCorlay:
The failing test seems unrelated.

minrk:
Can you document the new functionality? The methods you replaced were public APIs whose behavior was documented. `decorate` is completely undocumented (and still unclear to me), as is `static_init`. Which is called when, and what do they do? Why is instance_init totally removed, rather than changing what behaviors happen at instance init time?

SylvainCorlay:
Yes, I will add docstrings this afternoon. In short `instance_init` was doing multiple things

 - instantiating (and therefore validating right away) the default values of the `TraitType` instance.
 - decorating the `TraitType` instance
    - with functions such as `_resolve_classes`
    - or by setting the `this_class` of the sub-traits for container types.

`instance_init` has been decomposed into two functions: `static_init` on the one hand which does the first part and `decorate` on the other hand which does the second part. 

The base trait type has an empty `decorate` function which may be overloaded by other trait types, hence I could remove the test for existence of `_resolve_classes` for example, which are specific to certain type of traits. 

Now, the goal is to deffer the execution of `static_init` as much as possible. Initially, the constructor of HasTrait was calling instance_init for all traits. Now it only calls static_init at this point only if a value is provided for the considered attribute. (Actually it calls set_default_value which hooks up potential `_%s_default` functions). In the case where it is not called at this point, no default value has been instantiated. The default value is only created and validated the first time it is accessed line 419

minrk:
Great, thanks for the explanation. I think `decorate` is part of what's tripping me up. `decorate` means something in Python, and it's not what is going on here. It seems like the basics of finishing initialization once there is an instance associated. Is `decorate` always called? If so, I might leave it called `instance_init`, and then separate the *value* initialization (now in static_init) from the Trait initialization, which I think is the actual goal here. If that's the case, I would probably leave `decorate` as `instance_init`, and call `static_init` something like `static_default_value` or `init_default_value`.

SylvainCorlay:
Done, I just added some documentation and renamed the methods accordingly to your recommendation. Indeed decorate is quite overloaded in Python.

minrk:
Terrific, thanks. I think it's a lot easier to follow, now. Will merge when Travis gives the green light.

minrk:
The failing tests are real, but due to a slightly pathological behavior in the in-process kernel init.

The in-process kernel asks for traits passed in kwargs *before* calling `super().__init__`.

A test case:

```python
from IPython.utils.traitlets import HasTraits, Instance

class C(HasTraits):
    t = Instance(int, allow_none=False)

    def _t_default(self):
        return 5

    def __init__(self, **kwargs):
        print(self.t)
        super().__init__(self, **kwargs)

C() # works, t=5
C(t=2) # Fails with t is None, but allow_none=False
```

1. `HasTraits.__new__` sees that `t` is given in kwargs, and thus skips `set_default_value`, preventing `_t_default` from being registered as an 
2. `C.__init__` requests `t`, but it hasn't been assigned yet, and `_t_default` was skipped, by the above, so `t.default_value()` is called instead, returning None.
3. `super().__init__(**kw)` actually assigns c.t from kwargs, as expected.

I don't think it's a problem that you cannot reliably access dynamic trait initialization before calling `HasTraits.__init__`, since nobody should really be doing that. I can remove the behavior in the in-process kernel without issue, I think.

minrk:
I made SylvainCorlay/ipython#3 to remove the weird in-process behavior.

SylvainCorlay:
Thanks, I was just looking again into this. (the tests used to pass before I rebased.)

SylvainCorlay:
Thanks!

PR #7778, opened 2015-02-13
add `%matplotlib notebook`

### LINKED ISSUE #7774: Alias mpl nbagg
Talking to @danielballan today he pointed out that nbagg, while accurate and making lots of sense to mpl developers, is not the clearest our most user friendly name.  

Would it make sense to provide an alias in the mpl related ipython magic to a more suitable name? Would it be possible to get that alias in for 3.0 at this late date?

minrk:
alias for `%matplotlib nbagg`

closes #7774


Carreau:
+1

PR #7779, opened 2015-02-13
Fixed rendering of URL in append_stream function

### LINKED ISSUE #5544: Notebook: convert URL to link fails when delayed
Consider the following snippet in a notebook:
```python
for i in range(10):
    time.sleep(0.2)
    print('http://www.google.com/')
```

Only the first 2 output lines are converted into links, but when the ```sleep``` method is commented out, all 10 output lines are converted into links.

IPython version: 1.1.0
Python version: 2.7.6
OS: Windows 8.1 64bit

abeguerra:
Proposed fix for problem rendering URLs. Addition of call to utils.autoLinkUrls() to ensure correct formatting of HTML. See below for issue thread.

closes #5544


minrk:
:+1:, thanks!

PR #7780, opened 2015-02-13
Fix race condition in javascript kernel message processing

jasongrout:
Because the binary messages are now deserialized using the asynchronous FileReader API, we need to have some way to force the messages to still be processed in the order they are received.  This patch implements a simple processing queue using promises.

### DIFF
--- a/IPython/html/static/services/kernels/kernel.js
+++ b/IPython/html/static/services/kernels/kernel.js
@@ -41,6 +41,7 @@ define([
         this.username = "username";
         this.session_id = utils.uuid();
         this._msg_callbacks = {};
+        this._msg_queue = Promise.resolve();
         this.info_reply = {}; // kernel_info_reply stored here after starting
 
         if (typeof(WebSocket) !== 'undefined') {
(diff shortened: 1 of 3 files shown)


jasongrout:
@minrk, @jdfreder - this is ready for comments.  I'll add a test for it on Monday.

minrk:
This looks about as simple as I could have hoped. Thanks, @jasongrout!

minrk:
A test would be awesome, but this looks great to me.

jasongrout:
@minrk - I think this is ready for review.  When I tried testing it using `iptest js`, it just said `Can't test binary websockets on phantomjs.`, so I'm not sure if the test actually passes.

rgbkrk:
Can slimer test the binary websockets?

takluyver:
In principle, it should be able to. Last time I tried running the tests on
slimer there were other failures, though.

jdfreder:
@jasongrout this looks very similar to the solution you devised for the widget, thanks!  +1 for 3.1 .

minrk:
When I wrote the binary websocket test, I was using slimer. I've since uninstalled it, and it's been a little while since I used it, but I can try running it when I get back tomorrow.

minrk:
@jasongrout I made jasongrout/ipython#7 (if you don't want to add the PR+PR merge commit, you can cherry pick  or rebase after merging), after which the tests pass. Your added tests pass fine, but the changes here and older msgspec changes caused the last segment of the test to fail.

jasongrout:
@minrk: I merged your PR; thanks!

minrk:
@jasongrout thanks!

PR #7791, opened 2015-02-15
Fix empty running table background

parente:
Fixes issue #7789
- Add list_placeholder style
- Change id so list_header class doesn't get added by JS


parente:
Here's what it looks like:

fix-empty-table-placeholder

Carreau:
Seem simple enought, +1 for 3.0 if another dev agree, let's merge.

minrk:
yup, okay

PR #7794, opened 2015-02-16
Minor styling fix for IntText and FloatText widgets

SylvainCorlay:
The class of the top div element for those widget was `widget-text`, instead of `widget-numeric-text`. 
However, the `Text` widget is wider than the `IntText` widget, causing the following mismatch between containing and contained div elements:

wrong_size


SylvainCorlay:
@jdfreder

jdfreder:
Thanks @SylvainCorlay for catching this!

PR #7798, opened 2015-02-16
Extract session buffers as memoryviews

jasongrout:
A bug in pyzmq forces us to have a workaround, which can be deleted when the pyzmq dependency is upgraded to fix zeromq/pyzmq#646.

Thanks to @minrk for helping work on the code and also for the bugfix in pyzmq.

### DIFF
--- a/IPython/html/base/zmqhandlers.py
+++ b/IPython/html/base/zmqhandlers.py
@@ -43,6 +43,8 @@ def serialize_binary_message(msg):
     # don't modify msg or buffer list in-place
     msg = msg.copy()
     buffers = list(msg.pop('buffers'))
+    # for python 2, copy the buffer memoryviews to byte strings
+    buffers = [x.tobytes() for x in buffers]
     bmsg = json.dumps(msg, default=date_default).encode('utf8')
     buffers.insert(0, bmsg)
     nbufs = len(buffers)
--- a/IPython/html/base/zmqhandlers.py
+++ b/IPython/html/base/zmqhandlers.py
@@ -8,6 +8,7 @@ import os
 import json
 import struct
 import warnings
+import sys
 
 try:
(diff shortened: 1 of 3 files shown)


minrk:
:+1: thanks, @jasongrout.

jasongrout:
I feel bad about the python2 only fix in the zmqhandlers file.  It makes a copy on python3 too.  Do we support having branches in our code for python2/python3?

takluyver:
Test failures on Python 3.3 only, but they look real:

```
ERROR: IPython.html.tests.test_serialize.test_serialize_binary

  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/IPython/html/tests/test_serialize.py", line 17, in test_serialize_binary
    bmsg = serialize_binary_message(msg)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/IPython/html/base/zmqhandlers.py", line 56, in serialize_binary_message
    return b''.join(buffers)
TypeError: sequence item 2: expected bytes, memoryview found
```

jasongrout:
Looks like my python 3 optimization was actually added in 3.4: https://bugs.python.org/issue15958

I'll reverse the optimization tomorrow.  Thanks.

takluyver:
It's also fine, as far as I'm concerned, to have an `if sys.version_info < (3, 4):` check to take advantage of the optimisation when it is available.

jasongrout:
@takluyver - done.  let's see if it passes the tests.

takluyver:
Tests are passing, but not merging now as this is milestoned for 4.0.

rgbkrk:
:+1: for getting this in right after the 3.0 release.

PR #7801, opened 2015-02-16
Force display_message to be processed in order

jasongrout:
Otherwise sometimes the display message will be processed before the update message.

### DIFF
--- a/IPython/html/static/widgets/js/widget.js
+++ b/IPython/html/static/widgets/js/widget.js
@@ -155,8 +155,9 @@ define(["widgets/js/manager",
                     this.trigger('msg:custom', msg.content.data.content);
                     break;
                 case 'display':
-                    this.widget_manager.display_view(msg, this)
-                        .catch(utils.reject('Could not process display view msg', true));
+                    this.state_change = this.state_change.then(function() {
+                        that.widget_manager.display_view(msg, that);
+                    }).catch(utils.reject('Could not process display view msg', true));
                     break;
             }
         },


jasongrout:
@jdfreder, @sylvaincorlay - review?

SylvainCorlay:
I confirm that it solves my bug.

jasongrout:
@jdfreder, @minrk - I think this bug fix could be a candidate for 3.0.  The problem Sylvain was seeing was that if you created a widget in the same cell as displaying it, where the widget is dynamically loaded using requirejs, it often would not show up since the display message would beat the update message.

minrk:
Can you show an example that fails?

SylvainCorlay:
Yes, you only need to have a custom model loaded with the new requirejs mechanism. I can send something tomorrow.

SylvainCorlay:
You only need to use a custom model asynchronously loaded. For example, if you do

In /nbextensions/Foo/Bar.js
```JavaScript
define(["widgets/js/widget"], function(Widget) {
    "use strict";
    var Bar= Widget.WidgetModel;
    return {
        Bar: Bar,
    };
});
```

Then we simple declare a Slider that uses this model:
```Python
from IPython.html.widgets import IntSlider
from IPython.utils.traitlets import Unicode
from IPython.display import display

class BarSlider(IntSlider):
    _model_name = Unicode('Bar', sync=True)
    _model_module = Unicode('nbextensions/Foo/Bar', sync=True)

slider = BarSlider()
display(slider)
```

jdfreder:
I haven't seen this problem myself with my ipython-flightwidgets repo.  Is it frequent?  @SylvainCorlay 's example code should work as-is.

SylvainCorlay:
It is because you don't overload the model in your flight widget. This only
happens with overloaded models.

jdfreder:
Oh I see, thanks.

IPython/html/static/widgets/js/widget.js
```diff
@@ -155,8 +155,9 @@ define(["widgets/js/manager",
                     this.trigger('msg:custom', msg.content.data.content);
                     break;
                 case 'display':
-                    this.widget_manager.display_view(msg, this)
-                        .catch(utils.reject('Could not process display view msg', true));
+                    this.state_change = this.state_change.then(function() {
```
jdfreder: state_change is probable a bad name for the promise since we use it more generically, but we can change it later.

PR #7803, opened 2015-02-16
Fix list of union validation

SylvainCorlay:
In the case of e.g. `List(Int() | Bool())` the `name` attribute if `Int()` and `Bool()` was never set. Moving it from `instance_init` to `_resolve_classes` solves the issue. I also added a test. 

Note: this is a simple fix for this specific bug, but the whole instance-init / decoration is reworked in other traitlet PRs to avoid the instantiation of unused default values.


SylvainCorlay:
@minrk 
- This is a bug correction. 
- It only impacts `Union` which is new in 3.0. 

The separation between the default value initialization and the decoration made in the other PR cleans up all this, this is just a quick fix for that bug.

minrk:
Thanks, I think this is okay.

SylvainCorlay:
Thanks

PR #7806, opened 2015-02-17
Remove mention of Ctrl-J & Ctrl-K shortcuts from example notebook

takluyver:
Thanks to @mwaskom for reminding us of this on the mailing list.

I also redid the edit shortcuts screenshot, because the style has changed, so if I hadn't, it would look different from the command mode shortcuts image.


Carreau:
+1 once travis happy.

takluyver:
There should be no way Travis can be unhappy - it doesn't do anything with our examples, as far as I know. But I'm fine with waiting for it to be happy anyway.

Carreau:
In case of random faillure if you merge it send you a email "IPython master broken".
But I doubt it will fail, or at least have any reason to fail.

takluyver:
Yep, but random failure on master is, I believe, no less likely after the tests have passed on the PR. I'm sure I've had failure emails after merging successfully tested PRs.

Anyway, green light!

Issue #7807, opened 2015-02-18
Use consistent terminology describing nearby cells


abalkin:
The keyboard shortcuts screen uses three different pair of terms to describe the two cell adjacent to the current:

1. below/above
1. next/previous
1. before/after

For example, shift-enter is described as "run cell, select **below**", "J" is "select **next** cell", and "B" is "insert cell *after*".  The last choice is truly confusing: "B" means "**a**fter" and "A" means "**b**efore".

takluyver:
Ha, yes. At least for the `a` and `b` shortcuts we should certainly use the terms 'above' and 'below' to help people remember them.

@Carreau this appears to tie into the action names - are we auto generating help strings from action names now? That seems less than ideal, but that's another question. Can you look into this? I think at a minimum we should get rid of 'before' and 'after' in the naming because of the potential confusion with the shortcuts for 'above' and 'below'.

Carreau:
> are we auto generating help strings from action names now

Yes if no help string is given. But action names are even less stable that JS API :-P 

I'm not sure above an below are the good one as notebook presentation can be non-vertical, like in reveal. Consistency would be great, but "A Foolish Consistency is the Hobgoblin of Little Minds".
And where above/below might make sens in some place (UI) before/after make sens (in API)
so I would be careful in choosing.

ping @ellisonbg that will be interested.

abalkin:
@Carreau - "a" and "b" shortcuts don't make sense in reveal because slides are not editable.  In fact, "b" is a shortcut to "blackout" there.  I think when you are editing the the notebook, cells are always arranged vertically.

If some alternative application uses a different arrangement - it should change the shortcuts.

takluyver:
Honestly I don't think it's a big problem which names we use. It's pretty obvious what they all mean, and I doubt anyone could really miss the relationship between spatial organisation (above/below) and logical sequence (next/previous).

But the help text for the 'a' and 'b' shortcuts really shouldn't call them 'before' and 'after'. That's just needlessly confusing. Marked for 3.0 so we can at least fix that.

@abalkin well, you can edit notebooks in reveal mode using https://github.com/damianavila/live_reveal ;-)

Carreau:
> "a" and "b" shortcuts don't make sense in reveal because slides are not editable

Yes they are with `live_reveal`, but in  a slide true, they would be vertically arranged.
I just want to point out that the above/below might not always make sens. 

[...]

PR #7808, opened 2015-02-18
Update quickhelp.js

abalkin:
Use left/right instead of before/after.  See #7807.

### DIFF
--- a/IPython/html/static/notebook/js/quickhelp.js
+++ b/IPython/html/static/notebook/js/quickhelp.js
@@ -38,8 +38,8 @@ define([
             { shortcut: "Cmd-Down",   help:"go to cell end"  },
             { shortcut: "Alt-Left",   help:"go one word left"  },
             { shortcut: "Alt-Right",  help:"go one word right"  },
-            { shortcut: "Alt-Backspace",      help:"del word before"  },
-            { shortcut: "Alt-Delete",         help:"del word after"  },
+            { shortcut: "Alt-Backspace",      help:"delete word to the left"  },
+            { shortcut: "Alt-Delete",         help:"delete word to the right"  },
         ];
     } else {
         // PC specific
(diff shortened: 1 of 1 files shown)


Carreau:
Just to be annoying, how does that behave with arabic, or other RTL dialect :-P

Carreau:
And to respond to myself: 

http://codemirror.net/demo/bidi.html

It is indeed after/before, and not left/right, still the behavior can be configured. 

Though I guess if using RTL lang, we can assume things are customized.

takluyver:
Yeah, in this context, it really is the word before/after the cursor, not left/right. Whereas moving through the text with arrow keys goes in the direction the arrow points, so left/right is appropriate. So I don't think this change makes sense.

+1 to changing 'del' to 'delete', though - do you want to amend this commit to just do that?

abalkin:
Moreover, in RTL locale, you would have to change the pictograms:

screen shot 2015-02-17 at 8 22 21 pm

Carreau:
I'm not sure the pictogram change on RTL, they are still in this direction on keyboard. 
The arrow to the left does always delete Before cursor whatever direction that is.
As in exist sign stickyman can run to the right even if actual dors is on the left.

takluyver:
Well, it [does exist](http://www.fileformat.info/info/unicode/char/2326/index.htm), but we're not about to build logic into the shortcuts dialog to determine whether the cursor is in RTL text. :-)

abalkin:
> The arrow to the left does always delete Before cursor whatever direction that is.

Not really, see http://support.apple.com/en-us/HT201794

screen shot 2015-02-17 at 8 29 52 pm

Carreau:
Hum, I guess I've only seen seen windows-arabic then. 
Joy of internationalisation.

abalkin:
From the Apple Style Guide (https://help.apple.com/asg/mac/2013/ASG_2013.pdf):

screen shot 2015-02-17 at 8 42 05 pm

takluyver:
i.e. even apple is inconsistent between 'Forward' and 'Right'. I would guess that in RTL text it deletes the character after (i.e. to the left of) the cursor.

IPython/html/static/notebook/js/quickhelp.js
```diff
@@ -50,8 +50,8 @@ define([
             { shortcut: "Ctrl-Down",  help:"go to cell end"  },
             { shortcut: "Ctrl-Left",  help:"go one word left"  },
             { shortcut: "Ctrl-Right", help:"go one word right"  },
-            { shortcut: "Ctrl-Backspace", help:"del word before"  },
-            { shortcut: "Ctrl-Delete",    help:"del word after"  },
```
takluyver: and these ones ;-)

abalkin:
Github makes it amazingly easy to edit files.  Is rebasing easy as well?

takluyver:
I don't think so, sadly. But I think for a change this small, it's not worth your time to rebase it.

PR #7809, opened 2015-02-18
Update actions.js

abalkin:
Use left/right instead of before/after.  See #7807.


Carreau:
Seem legit.

abalkin:
I have not tested this.  Give me a few minutes ...

takluyver:
Go ahead. I'm fairly sure it will work ;-)

Carreau:
> Go ahead. I'm fairly sure it will work ;-)

That was more a I'm fine with the change unless someone object. Not a "waiting for travis".

PR #7814, opened 2015-02-18
How to install 1.x using pip

diimdeep:
It's possible, but not stated and obvious because old versions hidden in pypi listing

### DIFF
--- a/docs/source/install/install.rst
+++ b/docs/source/install/install.rst
@@ -2,8 +2,8 @@ IPython requires Python 2.7 or ≥ 3.3.
 
 .. note::
 
-    If you need to use Python 2.6 or 3.2, you can find IPython 1.0
-    `here <http://archive.ipython.org/release/>`__.
+    If you need to use Python 2.6 or 3.2, you can find IPython 1.x
+    `here <http://archive.ipython.org/release/>`__ or specify version in pip: ``pip install ipython==1.2.1``.
 
 Quickstart
 ==========


takluyver:
Meh...I think our install page is already too long, and I don't think it's our job to explain how to use standard installation tools.

Carreau:
Agreed with Thomas, technically I would just drop the note about the `archive.ipython.org` and just state that for python 2.6 you are limitted to 1.x series.

diimdeep:
[Afaik](https://bitbucket.org/pypa/pypi/issue/52/website-should-list-all-versions-of-a#comment-5915082) there is way to individually unhide version, imho 1.2.1 should be visible as latest 1.x, then maybe no need to mention in doc.

diimdeep:
Unhide here https://pypi.python.org/pypi/ipython

Carreau:
The request to unhide 1.2.1 seem reasonable. I'm no owner on PyPi though.

diimdeep:
@fperez @minrk @ellisonbg

minrk:
1.2.1 is unhidden on PyPI

takluyver:
Actually, I'm -1 on that as well. When only one version is visible (as is almost always the case, except when we're doing betas/RCs), https://pypi.python.org/pypi/ipython goes directly to the normal description and overview page for that version. When multiple versions are visible, that URL instead goes to the rather ugly 'Index of Packages' page that it currently shows.

fperez:
Agree with @takluyver

diimdeep:
Isn't ability to discover versions > ugly page ? Seems like next pypi will not have this problem, there is widget on right with versions https://warehouse.python.org/project/ipython/ https://github.com/pypa/warehouse

takluyver:
I think the ugly page is worse until the new version of PyPI lands (anyone know when that's likely?).

We don't really want to encourage people to use 1.x, it's only if you are stuck on an old version of Python for some reason. And in that scenario, I think it's fine to expect you to know things like how to install an older version using pip.

Carreau:
> (anyone know when that's likely?).

Not soon. Have proposed to contribute, but needed PG dump. Asked 2 month ago. Still no response. 

Feel free to revert showing 1.2.1 if you like.

takluyver:
I'll leave it for now, since the 3.0 release candidate means the ugly page is shown anyway. When we release 3.0 final I'll hide 1.2.1 again unless we decide otherwise.

fperez:
I'd suggest putting a link on the README to that Warehouse page, indicating that old versions can be found there. That lets anyone easily find the other ones, without uglifying the main page.

minrk:
The install doc now mentions `pip install 'ipython<2'

PR #7818, opened 2015-02-19
Revert "Prevent image outputs from being wider than their container."

takluyver:
Reverts #7721 (see discussion on that PR)

Reopens #7701

Ping @jdfreder

### DIFF
--- a/IPython/html/static/notebook/less/outputarea.less
+++ b/IPython/html/static/notebook/less/outputarea.less
@@ -131,13 +131,6 @@ div.output_png {
 div.output_jpeg {
 }
 
-// Prevent the image outputs from being wider than the page.
-div.output_png div.ui-wrapper, 
-div.output_svg div.ui-wrapper, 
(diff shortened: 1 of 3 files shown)


minrk:
Given the various options:

1. fix aspect ratio problem in #7721 
2. scroll large images within container (#7718)
3. (this) allow large images to overflow container (2.x behavior, more noticeable now that we have gray background)

I'm slightly inclined to keep the 2.x behavior, even though it makes it easy to produce slightly ugly output. If there were an easy way to express "don't confine to width", then I might be inclined to keep #7721, assuming a simple fix for the aspect ratio is found.

Given that #7721 as-is doesn't quite work, and causes problems when larger images are desirable, I think we should go with this PR for 3.0. We can try to get a better version of #7721 with fixed aspect ratio and that allows explicit disabling of the shrink-to-fit behavior for 3.1.

jdfreder:
Thanks for opening this @takluyver

PR #7822, opened 2015-02-20
Docs: Replace examples/lib and inprocess with examples/Embedding

mspacek:
There might be other instances in the docs that need file path updates, but these are the ones I've found while trying to figure out why matplotlib-related IPython processes hang around after I close my custom qtconsole app.


minrk:
Thanks

PR #7826, opened 2015-02-21
Adding a Color trait type and using it for widget styling

SylvainCorlay:
Adding a color trait type that validates that the provided string is a valid color.
In widgets, defaulting it to `None`. (And when it is None, the css naturally takes over.)

### DIFF
--- a/IPython/html/widgets/trait_types.py
+++ b/IPython/html/widgets/trait_types.py
@@ -1,0 +1,29 @@
+# encoding: utf-8
+"""
+Trait types for html widgets.
+"""
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import re
+from IPython.utils import traitlets
+
+#-----------------------------------------------------------------------------
+# Utilities
+#-----------------------------------------------------------------------------
+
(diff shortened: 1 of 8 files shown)


SylvainCorlay:
This avoids silent failures when setting an invalid color.

SylvainCorlay:
The test failure does not seem legitimate.

minrk:
I'm not sure this belongs in traitlets. Maybe it should be defined in the widget package when it's split out, but it doesn't seem appropriate in base traitlets.

SylvainCorlay:
Yes I thought about it. I kept it there for now since I factored out _coerce_string.

SylvainCorlay:
Moved the Color trait type to the widget package.

minrk:
Thanks, I think that makes the most sense.

Carreau:
Marking as 4.0

IPython/html/widgets/trait_types.py
```diff
@@ -0,0 +1,29 @@
...
+_color_re = re.compile(r'#[a-fA-F0-9]{3}(?:[a-fA-F0-9]{3})?$')
+
+
+class Color(traitlets._CoercedString):
```
minrk: I think this should inherit from Unicode. There's no reason to coerce to str, right?

IPython/utils/traitlets.py
```diff
@@ -1298,17 +1293,27 @@ def coerce_str(self, obj, value):
                     self.error(obj, value)
             return value
 
+
+class ObjectName(_CoercedString):
```
minrk: I don't think we need to separate out the _CoercedString base class.

IPython/html/widgets/trait_types.py
SylvainCorlay: This is what I did initially

IPython/utils/traitlets.py
SylvainCorlay: Indeed I won't need it if I inherit from Unicode.

SylvainCorlay:
Rebased and now inheriting from Unicode.

minrk:
Looks great, thanks!

PR #7832, opened 2015-02-22
remove default_secure, make Session secure by default

minrk:
I think default_secure was an artifact of very early design prototyping when parallel and kernel were separate and had different default behavior, and really isn't necessary at this point.

This removes any use of `default_secure`. Since this is for 3.1, I deprecated instead of removing `default_secure`, even though there's little to no reason that it might be used outside IPython. Session.key now defaults to a random UUID instead of an empty string (no signing).

This was prompted by @parente discovering that all kernels use the same key in their connection files, as a result of default_secure. We were actually relying on this by accident, so there's another small change to get the Session key from a KernelManager in the websocket handlers.

### DIFF
--- a/IPython/kernel/zmq/tests/test_session.py
+++ b/IPython/kernel/zmq/tests/test_session.py
@@ -1,16 +1,9 @@
-"""test building messages with streamsession"""
+"""test building messages with Session"""
 
-#-------------------------------------------------------------------------------
-#  Copyright (C) 2011  The IPython Development Team
-#
-#  Distributed under the terms of the BSD License.  The full license is in
-#  the file COPYING, distributed as part of this software.
-#-------------------------------------------------------------------------------
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
 
-#-------------------------------------------------------------------------------
-# Imports
-#-------------------------------------------------------------------------------
-
+import hmac
 import os
 import uuid
 from datetime import datetime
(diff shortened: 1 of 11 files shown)


IPython/kernel/zmq/tests/test_session.py
```diff
@@ -71,6 +64,9 @@ def test_serialize(self):
         # ensure floats don't come out as Decimal:
         self.assertEqual(type(new_msg['content']['b']),type(new_msg['content']['b']))
 
+    def test_default_secure(self):
+        self.assertIsInstance(self.session.key, bytes)
+        self.assert_(self.session.auth, hmac.HMAC)
```
takluyver: Is this meant to be `assertIsInstance` as well?

takluyver:
Test failures are real - looks like the inprocess kernel machinery assumes that the key is consistent within a process.

parente:
Session gets created and generates a default key for inprocess/ipkernel here:

https://github.com/ipython/ipython/blob/master/IPython/kernel/inprocess/ipkernel.py#L125

I'm not familiar enough with the code to see how the client / ipkernel are supposed to share a key. Is this test setup problem alone?

takluyver:
Thanks. I think it is a real issue, not a test thing. I'd probably fix it by just turning off message signing for the inprocess kernel completely - since the messages aren't sent over the network, I'm pretty sure there's no reason to sign them.

IPython/kernel/zmq/tests/test_session.py
minrk: yes

minrk:
Yup, I had to stop right before committing the inprocess fix.

minrk:
in-process failures are fixed

parente:
Thanks for jumping on this so quickly @minrk. Looks good.

IPython/kernel/zmq/tests/test_session.py
Carreau: Want to fix this test to IsInstance ?

minrk: yup, forgot about this one. Fixed.

Carreau:
thanks !

PR #7838, opened 2015-02-23
Keep kernel messages synchronous

jasongrout:
A problem can happen when two messages come in for different
comms, where the second depends on the first (for example, the
first might be a message setting the state of a widget, and the
second triggering a view creation for the widget).  Since comm
message queues are independent of each other, the second message
could be executed before the first message.

This exposes a more fundamental assumption users are likely to
have that messages from python are processed synchronously.

This includes #7780.

### DIFF
--- a/IPython/html/static/services/kernels/comm.js
+++ b/IPython/html/static/services/kernels/comm.js
@@ -117,6 +117,7 @@ define([
             // don't return a comm, so that further .then() functions
             // get an undefined comm input
         });
+        return this.comms[content.comm_id];
     };
     
     CommManager.prototype.comm_msg = function(msg) {
@@ -134,6 +135,7 @@ define([
             }
             return comm;
         });
+        return this.comms[content.comm_id];
     };
     
     //-----------------------------------------------------------------------
--- a/IPython/html/static/services/kernels/kernel.js
+++ b/IPython/html/static/services/kernels/kernel.js
@@ -855,22 +855,23 @@ define([
     };
     
     Kernel.prototype._handle_ws_message = function (e) {
+        var that = this;
         this._msg_queue = this._msg_queue.then(function() {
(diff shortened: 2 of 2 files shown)


jasongrout:
We probably only need to merge this or #7839, but not both.  This approach makes stronger guarantees about synchronicity of message processing.

jasongrout:
CCing @jdfreder, @minrk for comments

jdfreder:
This may be a better approach for 4.0 (or later), but I think for 3.x the simplicity and minimal impact of #7839 is prefered.  @minrk opinion?

jasongrout:
I think this is better than #7839 for the following reasons:

1. in python, there really is an assumption of synchronous execution.  This is conceptually simpler than the #7839, as it just insists that all message handling be synchronous.  We already have messages deserialized synchronously; this makes the entire execution synchronous.

2. It's simpler to reason about the payload handlers if they are also executing synchronously, and this also guarantees that if they return a promise, they are executed before the next messages are processed.  That's a nice thing to have.

A message handler doesn't have to be synchronous, of course.  If it doesn't return a promise, we won't wait for it.  This just makes the default to wait for handling before trying to handle more messages.

minrk:
After a day of thinking about it, I'm inclined to agree, now. @jdfreder this PR is smaller than it looks, since ~half of it is actually #7780. The actual [changes are here](https://github.com/jasongrout/ipython/compare/jasongrout:message-race...kernel-messages-sync). It's mostly adding returns statements.

jdfreder:
Since 7780 was merged, this is probably the best solution.

minrk:
@jasongrout Can you rebase this one now that #7780 is merged so this accurately represents the diff?

jasongrout:
@minrk: rebased.

minrk:
@jasongrout thanks!

PR #7851, opened 2015-02-24
Fix contentsmanager api error message to give the path name

jasongrout:
@minrk, this is a very simple fix for an obvious omission.


minrk:
Yup, AOK for 3.0.

PR #7855, opened 2015-02-24
don't force instantiation of extra, unused contents manager

### LINKED ISSUE #7849: Contents Manager does not get passed config ?
I can't seem to have my custom contents manager receive a config.

In sessionmanager we seem to have `contents_manager = Instance('IPython.html.services.contents.manager.ContentsManager', args=())` but nowhere is it passed parent or config. Is that normal ?

minrk:
this is never used and never needed, and results in the instantiation of a contents manager that is immediately discarded.

closes #7849

### DIFF
--- a/IPython/html/services/sessions/sessionmanager.py
+++ b/IPython/html/services/sessions/sessionmanager.py
@@ -16,7 +16,7 @@ from IPython.utils.traitlets import Instance
 class SessionManager(LoggingConfigurable):
 
     kernel_manager = Instance('IPython.html.services.kernels.kernelmanager.MappingKernelManager')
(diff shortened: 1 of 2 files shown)


Carreau:
Ahhhh Yess !!! seem to do the trick ! Thanks ! +1.

3.1 ?

takluyver:
I'm happy for this to go into 3.0.

takluyver:
Test failures are real, but a problem with the tests - `test_sessionmanager` instantiates a SessionManager without passing `contents_manager`, so it's not instantiating a default contents manager.

takluyver:
This time the test failure looks unreal. Have restarted the relevant build.

Issue #7860, opened 2015-02-25
Version control of notebooks


Gruek:
I understand that Version control is a large topic and there are numerous issues and feature requests raised for different problems relating to version control.
I am creating this issue to discuss two of the pull requests that I am submitting to combat version control.

1. Visual diffing tool
The best solution to this I found was nbdiff (https://github.com/tarmstrong/nbdiff)
After using nbdiff for a while I found the following issues:
    - not compatible with Python 3
    - not compatible with nbformat 4
    - doesn't compare outputs
    - inaccessible from notebook ui
So using the main diffing algorithm from nbdiff I have developed a new tool that fixes those issues and a few other improvements as part of the ipython code.
It can be used to visually compare a notebook to its previous versions in git from the notebook menu or to compare two different notebook files from the tree view.
http://nbviewer.ipython.org/gist/Gruek/8186a0fb7d6492338f78

2. Git checkpoints
I have added a git checkpoints class which creates a repository in the notebooks dir and does a commit for each checkpoint.
This allows notebooks to have multiple checkpoints and the option to revert to any committed version.
If git isn't available it automatically uses file checkpoints instead.

PR #7861, opened 2015-02-25
Added notebook diff

Gruek:
Part 1 of the issue in #7860 

New feature: Visual diffing tool
It can be used to visually compare a notebook to its previous versions in git from the notebook menu or to compare two different notebook files from the tree view.

http://nbviewer.ipython.org/gist/Gruek/8186a0fb7d6492338f78

### DIFF
--- a/CSIRO BSD MIT License v2.0.txt
+++ b/CSIRO BSD MIT License v2.0.txt
@@ -1,0 +1,43 @@
+CSIRO Open Source Software License Agreement (variation of the BSD / MIT License)
+
+Copyright (c) 2015, Commonwealth Scientific and Industrial Research Organisation (CSIRO) ABN 41 687 119 230.
+
+All rights reserved. CSIRO is willing to grant you a license to this software on the following terms, except where otherwise indicated for third party material.
+
+Redistribution and use of this software in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+* Neither the name of CSIRO nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission of CSIRO.
+
+EXCEPT AS EXPRESSLY STATED IN THIS AGREEMENT AND TO THE FULL EXTENT PERMITTED BY APPLICABLE LAW, THE SOFTWARE IS PROVIDED "AS-IS". CSIRO MAKES NO REPRESENTATIONS, WARRANTIES OR CONDITIONS OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO ANY REPRESENTATIONS, WARRANTIES OR CONDITIONS REGARDING THE CONTENTS OR ACCURACY OF THE SOFTWARE, OR OF TITLE, MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, THE ABSENCE OF LATENT OR OTHER DEFECTS, OR THE PRESENCE OR ABSENCE OF ERRORS, WHETHER OR NOT DISCOVERABLE.
+TO THE FULL EXTENT PERMITTED BY APPLICABLE LAW, IN NO EVENT SHALL CSIRO BE LIABLE ON ANY LEGAL THEORY (INCLUDING, WITHOUT LIMITATION, IN AN ACTION FOR BREACH OF CONTRACT, NEGLIGENCE OR OTHERWISE) FOR ANY CLAIM, LOSS, DAMAGES OR OTHER LIABILITY HOWSOEVER INCURRED.  WITHOUT LIMITING THE SCOPE OF THE PREVIOUS SENTENCE THE EXCLUSION OF LIABILITY SHALL INCLUDE: LOSS OF PRODUCTION OR OPERATION TIME, LOSS, DAMAGE OR CORRUPTION OF DATA OR RECORDS; OR LOSS OF ANTICIPATED SAVINGS, OPPORTUNITY, REVENUE, PROFIT OR GOODWILL, OR OTHER ECONOMIC LOSS; OR ANY SPECIAL, INCIDENTAL, INDIRECT, CONSEQUENTIAL, PUNITIVE OR EXEMPLARY DAMAGES, ARISING OUT OF OR IN CONNECTION WITH THIS AGREEMENT, ACCESS OF THE SOFTWARE OR ANY OTHER DEALINGS WITH THE SOFTWARE, EVEN IF CSIRO HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH CLAIM, LOSS, DAMAGES OR OTHER LIABILITY.
+
+APPLICABLE LEGISLATION SUCH AS THE AUSTRALIAN CONSUMER LAW MAY APPLY REPRESENTATIONS, WARRANTIES, OR CONDITIONS, OR IMPOSES OBLIGATIONS OR LIABILITY ON CSIRO THAT CANNOT BE EXCLUDED, RESTRICTED OR MODIFIED TO THE FULL EXTENT SET OUT IN THE EXPRESS TERMS OF THIS CLAUSE ABOVE "CONSUMER GUARANTEES".  TO THE EXTENT THAT SUCH CONSUMER GUARANTEES CONTINUE TO APPLY, THEN TO THE FULL EXTENT PERMITTED BY THE APPLICABLE LEGISLATION, THE LIABILITY OF CSIRO UNDER THE RELEVANT CONSUMER GUARANTEE IS LIMITED (WHERE PERMITTED AT CSIRO’S OPTION) TO ONE OF FOLLOWING REMEDIES OR SUBSTANTIALLY EQUIVALENT REMEDIES:
+
+(a)               THE REPLACEMENT OF THE SOFTWARE, THE SUPPLY OF EQUIVALENT SOFTWARE, OR SUPPLYING RELEVANT SERVICES AGAIN;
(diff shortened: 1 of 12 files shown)


takluyver:
Thanks for working on both of these pieces, I expect they'll be really useful!

We're just about to break the single big IPython repository up into a number of smaller repositories. So for both of these pieces, I think we should work out ways to develop and install them separately, rather than making them part of the IPython codebase.

ssanderson:
@takluyver from having worked on UI modules like this for Quantopian's codebase, my guess is that it will be easy to pull the python pieces of this into a relatively stable package, but difficult to make the diff UI a separate package.  Diffing feels like a general enough feature that it might belong in the main project.  Any thoughts on making a diff UI a feature of the HTML notebook (and building a protocol for representing diffs of notebooks and/or text files)?

takluyver:
My take on it is that rich diffing is something where we shouldn't bless one solution by integrating it into the notebook UI at present. I'd rather let people build these tools as separate, standalone things, and work out how best to integrate them once the possible approaches have been better explored.

ssanderson:
> My take on it is that rich diffing is something where we shouldn't bless one solution by integrating it into the notebook UI at present. 

@takluyver the issue with this for UI work is that it's so hard to make a stable UI extension that it's hard to justify the time investment necessary to build something polished.  A policy of "wait and see what the community builds" only makes sense if the community has the tools and incentives to explore the space of possible implementations.  (I don't mean this as a knock against IPython's architecture btw; it's just genuinely hard to build a UI that's both pleasant to use and extensible).

Maybe what's needed is support for something like [Tornado's UI Modules](http://tornado.readthedocs.org/en/latest/guide/templates.html?highlight=module), which would allow people to build UI extensions against a set API.

takluyver:
I mean that notebook diffing should for now be a separate, user facing tool, not something integrated into the notebook. i.e. you would run `nbdiff foo.1.ipynb foo.2.ipynb`. This is useful anyway so it can be integrated with git hooks.

Carreau:
Though I understand the point the UI integration is difficult, maintaining UI is a lot of work, and we don't have the bandwidth to work on that without more full time developper.

With more regular contributor we might probably accept, but right not this is completely against what we are doing, which is separating into many repo. 

Though it is a perfect use case for defining the right API to work with different backend. After skimming I think it works only with git for the time beeing. I guess it could be refactor to use different backend postgress ? mercurial ? Google drive ? 


At quick glance I see `Diff` with an lcs method (longuest common subsequence ?) is it a general difflib ?
If so I tried to do a generic pure python [diff package](https://github.com/Carreau/difflib2.py), it would be nice to work on that separately. and reblace the stupid broken difflib in python stdlib. 

Will get a deeper look later though, and really exited to see people working on that.

smoskwa:
One of the reasons I asked Alex to modify IPython was that nbdiff did not meet the requirements of our organisation. It works well if you run it locally however our users connect to notebook servers that are running on remote shared hosts (e.g. login nodes for high performance computing clusters)

That means launching nbdiff requires a ssh or VNC session to the remote host, navigating to the underlying notebook, etc.. which I’ll admit is merely inconvenient. More importantly, nbdiff presents results to at http://localhost:5000 but what we really wanted is a secure per-user port so that it operates when there are many users on the same host.


I’ve proposed that he make some changes:

- all the git-specific code becomes part of his git multiple checkpoint implementation and that is provided as an extension (GitContentsManager)
- the file_diff is added to the default contents manager, a barebones NBDiff class provided, and the tree notebook diff button retained
- the full NBDiff class above can be an extension, and others could provide their own that implement different algorithms
- the diff dropdown menu is removed and instead an option is provided for diffing against checkpoints (similar to the existing revert to checkpoint). this would do very little in the default single-checkpoint environment but if the above GitContentsManager is loaded you basically get the full functionality

That said, I’m not wedded to a particular path forward, it just didn’t seem clear we could achieve what we wanted purely through extensions or as a standalone tool.

I’m keen to support further work on this as it would benefit our users but I’m also aware that Alex’s vacation scholarship ends tomorrow and there are no guarantees on his future availability.

takluyver:
It is now possible to have server extensions - third party code which can add HTTP handlers to the notebook web application. See [an example in cite2c](https://github.com/takluyver/cite2c/blob/master/cite2c_handlers.py). This would make it possible to serve this on the same port as the notebook application.

We're really happy to see people building things like this, but please feel free to get in touch with us at an earlier stage in the project. It takes some time to work out what makes sense in IPython, what should be distributed separately, how to integrate things, and so on. We're not really ready to handle a thousand lines of new code out of the blue, especially a couple of days before the person working on it is due to move on.

Carreau:
> One of the reasons I asked Alex to modify IPython was that nbdiff did not meet the requirements of our organisation. It works well if you run it locally however our users connect to notebook servers that are running on remote shared hosts (e.g. login nodes for high performance computing clusters)
> 
> That means launching nbdiff requires a ssh or VNC session to the remote host, navigating to the underlying notebook, etc.. which I’ll admit is merely inconvenient. More importantly, nbdiff presents results to at http://localhost:5000 but what we really wanted is a secure per-user port so that it operates when there are many users on the same host.

Sure the requirement make sens. I suppose you have seen [JupyterHub](github.com/jupyter/jupyterhub) which is here for multi-user.


> all the git-specific code becomes part of his git multiple checkpoint implementation and that is provided as an extension (GitContentsManager)

Sound great, ping us on the repo once you have it, or even before if you need some help.


> the file_diff is added to the default contents manager, a barebones NBDiff class provided, and the tree notebook diff button retained

I think I would rediscuss that on the PR, but a smaller PR will be easier to handle. I think that "filediff" might make sens on non-file backends like postgress. where you could still do diff betzeen 2 notebook versions, or even 2 notebooks. 

But that would definitively get the discussion started, and/or make us think of the needed API. 

> That said, I’m not wedded to a particular path forward, it just didn’t seem clear we could achieve what we wanted purely through extensions or as a standalone tool.

It might be difficult with extension if we don't have the right API, and we woudl gladely consider adding modifying API if needed. 

This is one of the reason we encourage People to ask first and open PR early, even when not complete, so that we can quickly iterate and review while things are developed. It is relatively hard to review a few thousand lines at once, but easier if you see the PR grow and get/give feedback all along.
Alex did a huge work and probably had to dig a lot into IPython to figure that out. We would have been happy to help you while you were doing it, and we don't bite. 

Even if this does not get merged now, and Alex does not come back, this is a starting point for future work, and we will give you credit for that.

I suppose this is also a good example at how bad we are at communicating how we work and what our future plan are. And we'll do our best to improve that in the future.

Gruek:
> It is now possible to have server extensions - third party code which can add HTTP handlers to the notebook web application. See an example in cite2c. This would make it possible to serve this on the same port as the notebook application.

Looking at this example, I'm not sure how the load_jupyter_server_extension function gets called, where does nbapp come from?

I will try and package this up as a standalone extension with my remaining time.

ssanderson:
@Gruek server extensions are set up from https://github.com/ipython/ipython/blob/master/IPython/html/notebookapp.py#L955 and passed an instance of `NotebookApp`.  From there you can get to the Tornado server via `nbapp.web_app`, which is a subclass of `tornado.web.Application`.  Among other things, you can use `web_app.add_handlers` to register new tornado handlers there.

takluyver:
Thanks @ssanderson 

Server extensions are configured as a list of Python modules - e.g. in the `ipython_notebook_config.py` file, you could have something like this:

```
c.NotebookApp.server_extensions = ['myextensionmodule']
```

The notebook app imports those modules and calls `load_jupyter_server_extension()` from them.

Gruek:
Ok, I've got the handlers working now, thanks for all the help.

My next problem is, I'm not sure how to load in the nbdiff template to the HTMLexporter from the module. Currently, I have the nbdiff.tpl in the same dir of the module as the file that's trying to set the template_file, but it says template file could not be found.

I found these examples here https://github.com/ipython/nbconvert-examples/tree/master/custom_template but not sure how to apply them in the context of a standalone module.

Sorry for all the noob questions.

takluyver:
These aren't noob questions - server extensions are quite new, and we probably haven't worked out all the kinks yet.

I think where you're providing _template_name_default, you also need to provide _template_path_default, something like this:

```python
def _template_path_default(self):
    # The directory containing this module
    return os.path.dirname(os.path.abspath(__file__))
```

Gruek:
Using that code, I get the following error

> IPython.utils.traitlets.TraitError: The 'template_path' trait of an Exporter instance must be a list or None, but a value of class 'str'

I changed it to 

```python
def _template_path_default(self):
    # The directory containing this module
    return [os.path.dirname(os.path.abspath(__file__))]
```

and got template file not found again

takluyver:
The trait error was my mistake, and your fix should be exactly right. I don't know why you're getting template file not found again.

takluyver:
The logic in TemplateExporter to find and load a template is pretty complex. I wouldn't be at all surprised if there was a bug lurking somewhere in there. It's going on my list of things to try to clean up.

Gruek:
Wasn't a bug, just me not including the template in the module.

smoskwa:
In case anyone comes across this - we did pull out most of the functionality into an extension
https://github.com/csiro-scientific-computing/NotebookDiff

(the tree base file diff was not included)

takluyver:
Thanks @Gruek and @smoskwa !

minrk:
Thanks, guys!

Gruek:
Hi again,

So the standalone package now has the full functionality of the pull request, but the process for enabling is currently 5 steps,

First you have to pip install the module, then you have to change the contents manager class in the config then you have to enable the server extension to add the handlers, then you have to install the nbextension to add the git compare menu, and finally there's also a bit that goes into custom.js to add the Diff button to the notebook list

So I'm just wondering if there's a way to repackage this in a way so it is only a 2 step process, where you can install the module and then enable/disable it in the config or through a python script.

Any help would be appreciated.

jasongrout:
If we move the Notebook UI to https://github.com/phosphorjs (which has a nice component architecture), and work on the packaging angle as well, it should be considerably simpler to install and make available a UI extension to the notebook.

PR #7862, opened 2015-02-25
Added git checkpoints

Gruek:
Part 2 of the issue in #7860 

New feature: Git checkpoints
I have added a git checkpoints class which creates a repository in the notebooks dir and does a commit for each checkpoint.
This allows notebooks to have multiple checkpoints and the option to revert to any committed version.
If git isn't available it automatically uses file checkpoints instead.

### DIFF
--- a/CSIRO BSD MIT License v2.0.txt
+++ b/CSIRO BSD MIT License v2.0.txt
@@ -1,0 +1,43 @@
+CSIRO Open Source Software License Agreement (variation of the BSD / MIT License)
+
+Copyright (c) 2015, Commonwealth Scientific and Industrial Research Organisation (CSIRO) ABN 41 687 119 230.
+
+All rights reserved. CSIRO is willing to grant you a license to this software on the following terms, except where otherwise indicated for third party material.
+
+Redistribution and use of this software in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+* Neither the name of CSIRO nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission of CSIRO.
+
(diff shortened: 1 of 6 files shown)


minrk:
This is very nice, and exactly why we have made the checkpoints and file manager classes configurable - so new ideas like this do not need to be in IPython itself to be used. It would be great to develop this as a package outside of IPython.

ssanderson:
@Gruek I'd be happy to help with making this a standalone package.

Carreau:
Same here, I'm working on integration with google drive so I can help with making that an extension. 
This has been on often request feature, and we always said that we would not ship that as part of the core. 

Having it as an extension make iterating on it much faster. 

If needed to be hosted in an organisation we could get it into Jupyter-org, or maybe IPython-contrib.

Gruek:
While I agree this could easily function as an extension, I don't fully understand why it can't be provided as an included configurable option if this is something that is commonly requested and is an improvement over the default file checkpoints. Is it simply a matter of not being able to maintain all the different components? If so, the project is going to grow either way, and this approach of only maintaining the core will leave behind a lot of the functionality. That is just my personal opinion, and I would like to hear yours.

I would certainly appreciate any help with making it standalone if that's the way it has to go, as I don't have any clue on where to begin and what the framework is for ipython standalone extensions. How do people discover what packages are currently out there?

minrk:
We write these APIs and make them configurable specifically to allow people to develop these ideas outside IPython, without the massive inertia associated with being part of IPython itself. When we add functionality to IPython, we are committing to supporting it indefinitely. There is a high cost to that. Making extension points that others can use to provide their own functionality allows experimental ideas such as this to move at their own pace, which is likely very much out of sync from the IPython release cycle. If they are generically useful and popular, it may be merged into IPython itself in the future, but that's not where it should start.

A great example of this is the [ExecutePreProcessor](https://github.com/ipython/ipython/blob/3.x/IPython/nbconvert/preprocessors/execute.py), which began as a simple [demo script](https://gist.github.com/minrk/2620876), was picked up as [runipy](https://github.com/paulgb/runipy), and ultimately landed in IPython.

> I would certainly appreciate any help with making it standalone if that's the way it has to go

Sure, we'd be happy to help. You may want to look at these other examples of custom storage implementations:

- [pgcontents](https://github.com/quantopian/pgcontents) stores notebooks and checkpoints in postgres.
- [nbx](https://github.com/dalejung/nbx) does its own treatment of the filesystem and has some integration with GitHub

The main thing to do is create a Python package that contains your class and upload it to PyPI. This could be a single python module plus a small setup.py for installation. Then people can install your class, and enable it via `ipython_notebook_config.py`:

    c.FileContentsManager.checkpoints_class = 'yourmodule.GitCheckpoints'

Gruek:
This is what I have so far, https://github.com/csiro-scientific-computing/GitCheckpoints

feel free to suggest improvements

minrk:
Thanks, @Gruek, we'll point people to the new repo.

PR #7865, opened 2015-02-25
replace view.apply_with_flags() with view.apply()

drevicko:
For a BalancedView instance, view.apply_with_flags() doesn't seem to exist. 
I presume view.apply() was intended?


minrk:
yup, thanks

PR #7868, opened 2015-02-25
Fix padding issue

### LINKED ISSUE #7867: Default DOMWidget Padding should be 0px, and Button styling issue
@jdfreder

padding

SylvainCorlay:
Fixes #7867

Before

padding

After

padding_fixed

The original change was made in #7139.

### DIFF
--- a/IPython/html/widgets/widget.py
+++ b/IPython/html/widgets/widget.py
@@ -435,7 +435,7 @@ class DOMWidget(Widget):
     width = CUnicode(sync=True)
     height = CUnicode(sync=True)
     # A default padding of 2.5 px makes the widgets look nice when displayed inline.
-    padding = CUnicode("2.5px", sync=True)
+    padding = CUnicode(sync=True)
     margin = CUnicode(sync=True)
 
     color = Unicode(sync=True)


SylvainCorlay:
@jdfreder

jdfreder:
Let me take a look at this in the mega widget alignment testing notebook...

jdfreder:
This PR breaks the alignment of almost every widget.  You can't just remove the padding, see interact:

screen shot 2015-02-25 at 2 14 26 pm

Also a more extensive test:

alignmentbugs

If you want to run the test notebook locally:

https://gist.github.com/jdfreder/9103330

jdfreder:
How about just overriding Button's padding as 0?

SylvainCorlay:
This does not look broken to me. Just that there is not space between the widgets, which is fine.

jdfreder:
No spacing is broken, the padding is intentional.  Alternatively, interact could be modified to add the padding to the widgets so the vanilla widgets themselves could be padding-less.  Although this means people who pass widget instances into interact instead of using the abbreviations will need to know they have to set a padding to get the Widgets to display the same.

SylvainCorlay:
The problem with having a non-zero top-level padding is that it is applied *everywhere*, especially in nested layouts. 
I am not a fan of styling attributes in DOMWidget in general. It would probably be preferable if you want to put some spacing in flexbox layouts to have the flexbox handle it.

jdfreder:
Really what would be ideal is the styling mechanism that either you or @jasongrout proposed (sorry I forget which one of you), which is to have a trickle-down styling class.  We can mimic the same thing in the browser using CSS classes.

SylvainCorlay:
Yes, styling by context.

SylvainCorlay:
I prefer having zero padding. The layout in the top-level screenshot is broken in every possible way. If you want a quick fix for flexbox, it could forcibly add margin to its elements.

minrk:
merged in #7872

PR #7871, opened 2015-02-25
Misc docs cleanup

takluyver:
Fix a few warnings/bad formattings when building the docs. I've also given the autogenerated config docs a polish.


minrk:
Most of these changes make me a little sad, since it's making docstrings slightly worse to appease autodoc, but ok.

PR #7872, opened 2015-02-25
Remove padding without breaking interact style.

### LINKED ISSUE #7867: Default DOMWidget Padding should be 0px, and Button styling issue

jdfreder:
superset of #7868
closes #7867
without re-opening #7139

### DIFF
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -1195,6 +1195,10 @@ h6:hover .anchor-link {
   font-size: 100%;
   font-style: italic;
 }
+.widget-interact div,
+.widget-interact input {
+  padding: 2.5px;
+}
 .widget-area {
   /*
     LESS file that styles IPython notebook widgets and the area they sit in.
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -1195,8 +1195,8 @@ h6:hover .anchor-link {
   font-size: 100%;
   font-style: italic;
 }
-.widget-interact div,
-.widget-interact input {
+.widget-interact > div,
+.widget-interact > input {
   padding: 2.5px;
 }
 .widget-area {
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -9978,6 +9978,10 @@ h6:hover .anchor-link {
   font-size: 100%;
   font-style: italic;
 }
+.widget-interact div,
+.widget-interact input {
+  padding: 2.5px;
+}
 .widget-area {
   /*
     LESS file that styles IPython notebook widgets and the area they sit in.
--- a/IPython/html/static/style/style.min.css
+++ b/IPython/html/static/style/style.min.css
@@ -9978,8 +9978,8 @@ h6:hover .anchor-link {
   font-size: 100%;
   font-style: italic;
 }
-.widget-interact div,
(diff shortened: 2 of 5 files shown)


jdfreder:
@SylvainCorlay

SylvainCorlay:
OK, looking into it.

IPython/html/static/style/ipython.min.css
```diff
@@ -1195,6 +1195,10 @@ h6:hover .anchor-link {
   font-size: 100%;
   font-style: italic;
 }
+.widget-interact div,
```
SylvainCorlay: `.widget-interact>div ?`

IPython/html/static/style/style.min.css
```diff
@@ -9978,6 +9978,10 @@ h6:hover .anchor-link {
   font-size: 100%;
   font-style: italic;
 }
+.widget-interact div,
```
SylvainCorlay: idm

IPython/html/static/widgets/less/widgets.less
```diff
@@ -1,6 +1,13 @@
 @widget-width: 350px;
 @widget-width-short: 150px;
 
+// Pad interact widgets by default.
+.widget-interact {
```
SylvainCorlay: idem and for input as well?

IPython/html/static/style/ipython.min.css
jdfreder: Ah yes, thanks for noticing that.

jdfreder:
@SylvainCorlay should be fixed.

jdfreder:
@minrk @takluyver I don't know if this is still an appropriate for 3.0 fix (it's still mostly styling).  I'll leave it up to you two to decide, marked as 3.1 otherwise.

SylvainCorlay:
I like this solution. We have alternate layout classes (I don't use core IPython's flexbox), and your change does not have side effects for me.

takluyver:
I'd say leave it for 3.1, since CSS fixes seem to have a high probability of introducing new CSS bugs*. But Min understands this stuff better than I do, so I'll let him make the call.

* CSS critical mass: the amount of CSS wherein each bugfix occasions, on average, >1 new problem, leading to an exponential increase in bugs.

250px-castle_romeo

SylvainCorlay:
@takluyver 
This looks *really* broken: 
padding

takluyver:
Perhaps, but the way that the initially proposed fix made all `@interact` widgets look really broken is a good illustration of what I mean about each fix introducing other bugs. I know that one is now taken care of, but we don't have long to find out what else this fix might break.

I consider manually creating HBox containers to be a fairly advanced feature, so I don't think it's a big deal if it looks off when 3.0 goes out. I presume you can work around it by setting attributes on the widgets inside it.

jdfreder:
> This looks really broken:

It's okay, bugs exist in software.  At least we are aware of them! :)  It does have an easy workaround too:

```python
HBox([Button(padding=0), Text(padding=0)])
```

minrk:
Since ~nothing about widgets is going to be anywhere near remotely stable for the duration of 3.x, I'm fine with this for 3.0. 3.1 won't be so long from now, if it introduces something funky.

jasongrout:
@takluyver: so you have a problem with the CSS styling, so you say to yourself, "I know, I'll just add another line of CSS".  Now you have two problems.

PR #7876, opened 2015-02-26
Reproducible docs

Carreau:
With Thomas #7871 should make the docs building mostly reproducible, except for ~20 lines.

### DIFF
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -270,7 +270,11 @@ def _get_type(obj):
     """Return the type of an instance (old and new-style)"""
     return getattr(obj, '__class__', None) or type(obj)
 
-_raise_key_error = object()
+class _RaiseKeyError(object):
+    def __repr__(self):
+        return '<'+self.__module__ + "." + self.__class__.__name__+'>'
+
+_raise_key_error = _RaiseKeyError()
 
 
 class BaseFormatter(Configurable):
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -1322,9 +1322,12 @@ def text_repr(value):
 def eqrepr(value, repr=text_repr):
     return '=%s' % repr(value)
 
+eqrepr.__repr__ = lambda x:'<'+eqrepr.__module__ + "."+'eqrepr>'
 
 def nullrepr(value, repr=text_repr):
     return ''
+
+nullrepr.__repr__ = lambda x:'<'+nullrepr.__module__ + "."+'nullrepr>'
 
 
 #----------------------------------------------------------------------------
@@ -1368,4 +1371,3 @@ if __name__ == "__main__":
     except:
         handler(*sys.exc_info())
     print('')
-
--- a/IPython/html/services/sessions/sessionmanager.py
+++ b/IPython/html/services/sessions/sessionmanager.py
@@ -22,6 +22,7 @@ class SessionManager(LoggingConfigurable):
     _cursor = None
     _connection = None
     _columns = {'session_id', 'path', 'kernel_id'}
+
     
     @property
     def cursor(self):
--- a/IPython/utils/coloransi.py
+++ b/IPython/utils/coloransi.py
@@ -123,6 +123,9 @@ class ColorScheme:
             name = self.name
         return ColorScheme(name, self.colors.dict())
 
+    def __repr__(self):
+        return '<ColorScheme:' + self.name +'>'
+
 class ColorSchemeTable(dict):
     """General class to handle tables of color schemes.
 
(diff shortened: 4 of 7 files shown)


minrk:
-1 on the repr changes. I *really* don't think we should be changing code because the docs do something unpleasant.

takluyver:
Yeah, once again, I don't think it's a problem that docs change on each build. I think it's a problem that we're publishing docs by pushing each build as a git commit - otherwise, the diff wouldn't actually matter.

I would seriously consider your suggestion of replacing commits instead of adding commits in the built docs repo; I don't think that history is useful in any case. It's just a way of getting files to a server, as far as I'm concerned.

jenshnielsen:
I recently set up Travis to publish the matplotlib docs here: https://github.com/matplotlib/matplotlib/blob/master/.travis.yml We deliberately decided to use ```git checkout --orphan gh-pages``` to deploy by replacing commits. So far that seems to work great.

Carreau:
> -1 on the repr changes. I really don't think we should be changing code because the docs do something unpleasant.

This is true also on interactive exploration. I did not change reprs  were it was not making  like stdout that have a random id : 

```
In [4]: ipc.BaseFormatter.pop?
Type: ...
String form: <function BaseFormatter.pop at 0x10227c400>
File:        ~ipython//IPython/core/formatters.py
Definition:  ipc.BaseFormatter.pop(self, typ, default=<object object at 0x101707110>)
```
vs 
```
Definition:  ipc.BaseFormatter.pop(self, typ, default=<IPython.core.formatters._RaiseKeyError>)
```

and thinking about it even more it should not be in this particular case `__classe__.__name__` but also hardcoded as _raise_key_error.

Carreau:
Need rebase.

Carreau:
@ellisonbg I think you meant to comment here. 

It's not only about docs building. The following is soooooo Usefull.

```
In [1]: from jupyter_nbformat import writes

In [2]: writes?
Signature: writes(nb, version=<object object at 0x1024b31e0>, **kwargs)
```

I feel like the following have more sens:
```
In [2]: writes?
Signature: writes(nb, version=<jupyter_nbformat._NO_CONVERT>, **kwargs)
```

ellisonbg:
I find the second to be more confusing as I know what those funny hex
numbers are, but have *never* seen _NO_CONVERT before.

On Sun, Mar 29, 2015 at 12:48 PM, Matthias Bussonnier <

> @ellisonbg <https://github.com/ellisonbg> I think you meant to comment
> here.
>
> It's not only about docs building. The following is soooooo Usefull.
>
> In [1]: from jupyter_nbformat import writes
>
> In [2]: writes?
> Signature: writes(nb, version=<object object at 0x1024b31e0>, **kwargs)
>
> I feel like the following have more sens:
>
> In [2]: writes?
> Signature: writes(nb, version=<jupyter_nbformat._NO_CONVERT>, **kwargs)
>
> —
> <https://github.com/ipython/ipython/pull/7876#issuecomment-87462174>.
>



-- 
Brian E. Granger

Carreau:
But you have a way now to query what NO_CONVERT  is,  while for the object, you have no clue how to get it...

takluyver:
I wonder if we can whip up a `@namerepr` decorator that gives an object/function a repr corresponding to its name.

I might even take the angle brackets away:

    Signature: writes(nb, version=jupyter_nbformat._NO_CONVERT, **kwargs)

The general rule of thumb is that reprs are either roughly a Python expression that will evaluate to the object (give or take relevant imports), or a brief summary of the object, and angle brackets mark the latter. If it's a name in a module, it's roughly eval-able.

Carreau:
woulnd just `class NO_CONVERT(object):pass` do the trick ? (Don't
instanciate)

takluyver:
That gives you a name, though the repr still looks like `<class 'jupyter_nbconvert.NO_CONVERT'>` instead of just `jupyter_nbconvert.NO_CONVERT`. Maybe the answer is to make our inspection machinery smarter about objects in the function signature. It's a pretty good bet that if `type(default_value) is object` you're looking at a sentinel.

Carreau:
```
# no-conversion singleton
class Sentinel(object):
    """Prevent nbformat to convert notebooks to most recent version.
    """

    def __init__(self, name):
        self.name = name

    def __repr__(self):
        return str(self.__module__)+'.'+self.name

NO_CONVERT = Sentinel('NO_CONVERT')
```

??

Is it really worth upgrading inspection ? Once we are python3 only we can rely on Emum right ?

takluyver:
It seems a bit strange to be using Enum for things that only have one possible value, but I suppose it would work. The Sentinel class also works; I wouldn't be against switching to that at least in public APIs like nbconvert.write.

takluyver:
Closing this as superseded by #8192.

PR #7877, opened 2015-02-26
update install doc

minrk:
- mention pip, not setuptools
- update dependency lists
- less detail about most dependencies

### DIFF
--- a/docs/source/install/install.rst
+++ b/docs/source/install/install.rst
@@ -2,29 +2,27 @@ IPython requires Python 2.7 or ≥ 3.3.
 
 .. note::
 
-    If you need to use Python 2.6 or 3.2, you can find IPython 1.0
-    `here <http://archive.ipython.org/release/>`__.
+    If you need to use Python 2.6 or 3.2, you can find IPython 1.x
+    `here <http://archive.ipython.org/release/>`__,
+    or get it with pip::
+    
+        pip install 'ipython<2'
+
 
 Quickstart
 ==========
 
-If you have :mod:`setuptools`,
+If you have :mod:`pip`,
 the quickest way to get up and running with IPython is:
 
 .. code-block:: bash
 
     $ pip install "ipython[all]"
 
-This will download and install IPython and its main optional dependencies:
-
-- jinja2, needed for the notebook
-- sphinx, needed for nbconvert
-- pyzmq, needed for IPython's parallel computing features, qt console and
-  notebook
-- pygments, used by nbconvert and the Qt console for syntax highlighting
-- tornado, needed by the web-based notebook
-- nose, used by the test suite
-- readline (on OS X) or pyreadline (on Windows), needed for the terminal
+This will download and install IPython and its main optional dependencies for the notebook,
+qtconsole, tests, and other functionality.
+Some dependencies (Qt, PyQt for the QtConsole, pandoc for nbconvert) are not pip-installable,
+and will not be pulled in by pip.
 
 To run IPython's test suite, use the :command:`iptest` command:
 
@@ -53,11 +51,6 @@ In many scenarios, this is the simplest method of installing Python packages.
 More information about :mod:`pip` can be found on
 `its PyPI page <http://pypi.python.org/pypi/pip>`__.
 
-.. note::
-
-   On Windows, IPython *requires* :mod:`setuptools`.  We hope to
-   change this in the future, but for now on Windows, you *must* install
-   :mod:`setuptools` to use IPython.
 
 More general information about installing Python packages can be found in
 `Python's documentation <http://docs.python.org>`_.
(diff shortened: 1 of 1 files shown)


Carreau:
Skimmed through, seem fine.

docs/source/install/install.rst
```diff
@@ -71,15 +64,14 @@ with no external dependencies.  However, some Python distributions
...
+editing.  If you install IPython with :mod:`pip`,
+then the appropriate :mod:`readline` for your platform will be installed.
+See below for details of how to make sure you have a working :mod:`readline`.
 
 Installation using pip
 ----------------------
 
-If you have :mod:`setuptools` or :mod:`pip`, the easiest way of getting IPython is
+If you have :mod:`pip`, the easiest way of getting IPython is
 to simply use :command:`pip`:
```
takluyver: This now sounds a bit redundant: "If you have pip, the easiest way of getting IPython is to simply use pip"

docs/source/install/install.rst
```diff
@@ -108,39 +100,20 @@ permissions, you may need to run the last command with :command:`sudo`.
...
-   the command ``pip install pyreadline`` from a terminal, or the binary
-   installer appropriate for your platform from the PyPI page.
-
-3. Install IPython itself, which you can download from `PyPI
-   <http://pypi.python.org/pypi/ipython>`_ or from `our site
-   <http://ipython.org/download.html>`_.  Note that on Windows 7, you *must*
-   right-click and 'Run as administrator' for the Start menu shortcuts to be
-   created.
+1. Install `pip <http://pypi.python.org/pypi/pip>`_, if you don't have it already.
+2. Install IPython: ``pip install ipython``.
```
takluyver: This section seems to now just be repeating what we say above.

docs/source/install/install.rst
```diff
@@ -231,7 +203,7 @@ installed):
     Other Python distributions on OS X (such as Anaconda, fink, MacPorts)
     already have proper readline so you likely don't have to do this step.
 
-When IPython is installed with :mod:`setuptools`, (e.g. using the
+When IPython is installed with :mod:`pip`, (e.g. using the
 ``pip`` command), the correct readline should be installed if you specify the `terminal`
```
takluyver: the 'e.g.' is redundant

docs/source/install/install.rst
```diff
@@ -270,84 +242,67 @@ iptest command:
...
+    pip install "ipython[notebook]"
 
-Jinja
------
+or manual:
 
-The IPython notebook uses the Jinja_ templating tool to render HTML pages.
+.. code-block:: bash
 
+    pip install pyzmq jinja2 tornado mistune jsonschema pygments terminado_
```
takluyver: Trailing underscore here (I guess from copying and pasting the list from above)

docs/source/install/install.rst
```diff
@@ -360,20 +315,19 @@ you can install MathJax locally.
...
+    python -m IPython.external.mathjax /path/to/source/mathjax-MathJax-v2.0-20-g07669ac.zip
 
 For unusual needs, IPython can tell you what directory it wants to find MathJax in::
 
-	python -m IPython.external.mathjax -d /some/other/mathjax
+    python -m IPython.external.mathjax -d /some/other/mathjax
 
-By default Mathjax will be installed in your ipython profile directory, but you
+By default Mathjax will be installed in your ipython directory, but you
 can make system wide install, please refer to the documentation and helper function
```
takluyver: this sentence doesn't read right - "you can make system wide install". Not your fault, but while we're working on this, may as well clean it up.

docs/source/install/install.rst
```diff
@@ -395,6 +349,7 @@ The following browsers are unsupported:
 * Chrome < 13
 * Opera (any): CSS issues, but execution might work
 * Internet Explorer < 10
+* Internet Explorer ≥ 10 (same as Opera)
 
 The following specific combinations are known **NOT** to work:
```
takluyver: I don't think we should bother listing these. They're probably out of date anyway.

docs/source/install/install.rst
minrk: right, this section was supposed to be about installing from source without pip for some reason. I would be inclined to remove the section, then, since Windows is no longer different from other systems in that case. What do you think?

docs/source/install/install.rst
minrk: I disagree. I can remove the 0.12 note, but Safari still rejects websocket connections to self-signed certs.

takluyver: How about a specific note of that, not formatting it as a list? I don't think we should refer to things here unless we're sure they're still correct.

minrk: makes sense, reduced to simple note about Safari + untrusted HTTPS

docs/source/install/install.rst
takluyver: +1, and I see it's already done.

PR #7878, opened 2015-02-26
add downgrade example to whatsnew

minrk:
and discuss `--to notebook` in nbconvert doc

### DIFF
--- a/docs/source/notebook/nbconvert.rst
+++ b/docs/source/notebook/nbconvert.rst
@@ -78,18 +78,44 @@ The currently supported export formats are:
   Basic reStructuredText output. Useful as a starting point for embedding notebooks
   in Sphinx docs.
 
-* ``--to python``
+* ``--to script``
 
-  Convert a notebook to an executable Python script.
-  This is the simplest way to get a Python script out of a notebook.
-  If there were any magics in the notebook, this may only be executable from
+  Convert a notebook to an executable script.
+  This is the simplest way to get a Python (or other language, depending on the kernel) script out of a notebook.
+  If there were any magics in an IPython notebook, this may only be executable from
   an IPython session.
+
+* ``--to notebook``
+
+  .. versionadded:: 3.0
+  
+  This doesn't convert a notebook to a different format *per se*,
+  instead it allows the running of nbconvert preprocessors on a notebook,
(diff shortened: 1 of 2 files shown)


docs/source/notebook/nbconvert.rst
```diff
@@ -78,18 +78,44 @@ The currently supported export formats are:
...
+  
+      ipython nbconvert --to notebook --nbformat 3 mynotebook.ipynb
+  
+  will create a copy of :file:`mynotebook.ipynb` in :file:`mynotebook.v3.ipynb`
+  in version 3 of the :ref:`notebook format <nbformat>`.
+  
+  If you want to convert a notebook in-place (careful!),
+  you can specify the ouptut file to be the same as the input file::
+  
+      ipython nbconvert --to notebook mynb --output mynb ...
```
Carreau: dot dot dot ? I would also be consistent on wether or not you add the .ipynb at the end of the filenames.

docs/source/whatsnew/version3.rst
```diff
@@ -217,6 +217,12 @@ Backwards incompatible changes
     to read and write notebook files
     instead of the deprecated :mod:`IPython.nbformat.current` APIs.
 
+  You can downgrade a notebook to v3 via ``nbconvert``::
+  
+      ipython nbconvert --to notebook --nbformat 3 [notebook]
```
Carreau: `<notebook>` ? isn't `[]` for optional and `<>` for required usually ? That's my feeling, but I'm not sure.

docs/source/notebook/nbconvert.rst
```diff
@@ -78,18 +78,44 @@ The currently supported export formats are:
...
+  :file:`mynotebook.nbconvert.ipynb`.
+  
+  ::
+  
+      ipython nbconvert --to notebook --nbformat 3 mynotebook.ipynb
+  
+  will create a copy of :file:`mynotebook.ipynb` in :file:`mynotebook.v3.ipynb`
+  in version 3 of the :ref:`notebook format <nbformat>`.
+  
+  If you want to convert a notebook in-place (careful!),
```
takluyver: I don't think the 'careful!' here is needed, since you have a warning after the example.

minrk:
comments should be addressed, thanks.

PR #7882, opened 2015-02-26
Don't allow stdin in Execute Preprocessor

### LINKED ISSUE #7880: Making stdin optional when creating new kernels
It would be useful for the `ExecutePreprocessor` (and other things that want to automatically run notebooks on the fly) to be able to easily disable `stdin` -- otherwise, if there is something like a call to `input`, it causes the preprocessor to just hang and eventually time out. I think this would just involve adding a new kwarg to this function: https://github.com/ipython/ipython/blob/master/IPython/kernel/manager.py#L417 and then calling `kc.start_channels(stdin=stdin)`, where `stdin` is whatever value that was passed in (True by default).

Questions:
1. Is there some other way to do this that I missed?
2. If not, does this change sound ok? If so, I'll go ahead and make a PR for it and add an option to disable stdin to the `ExecutePreprocessor` as well.

Carreau:
Execute Preprocessor is most likely not interactive, so
it make sens not to allow `stdin`, or execution will hang.

Should close #7880 if I understand the problem correctly.

Should we make that a configurable ?

### DIFF
--- a/IPython/nbconvert/preprocessors/execute.py
+++ b/IPython/nbconvert/preprocessors/execute.py
@@ -55,7 +55,7 @@ class ExecutePreprocessor(Preprocessor):
         return cell, resources
 
     def run_cell(self, cell):
-        msg_id = self.kc.execute(cell.source)
+        msg_id = self.kc.execute(cell.source, allow_stdin=False)
         self.log.debug("Executing cell:\n%s", cell.source)
         # wait for finish, with timeout
         while True:


takluyver:
I don't think we should make it configurable.

Carreau:
#7833 is an alternative that have test.

jhamrick:
@Carreau You can take the tests from my PR, or I can change my PR to remove configurability; either works for me.

Carreau:
Closing here.

PR #7883, opened 2015-02-26
Disable stdin in ExecutePreprocessor

### LINKED ISSUE #7880: Making stdin optional when creating new kernels

jhamrick:
Fixes #7880 

I made `allow_stdin` be false by default, because it seems to me that for executing a notebook on the command line, that makes the most sense.

### DIFF
--- a/IPython/nbconvert/preprocessors/execute.py
+++ b/IPython/nbconvert/preprocessors/execute.py
@@ -26,11 +26,6 @@ class ExecutePreprocessor(Preprocessor):
         help="The time to wait (in seconds) for output from executions."
     )
 
-    allow_stdin = Bool(
-        False, config=True, 
-        help="Whether stdin should be enabled when executing cells"
-    )
-    
     extra_arguments = List(Unicode)
 
     def preprocess(self, nb, resources):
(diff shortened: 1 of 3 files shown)


Carreau:
Oh, you already made it non configurable :-)  You are fast.

Carreau:
+1, let's wait for travis.

Carreau:
Marking as 3.1, but I'm ok for 3.0

takluyver:
I marked the other PR 3.1, but I'm also happy if we want to put this in 3.0. We should be careful, not paranoid.

PR #7884, opened 2015-02-26
Different handling of cell timeouts in ExecutePreprocessor

jhamrick:
Depending on the use of ExecutePreprocessor, if a cell takes a long time to run and times out, you may either want the whole thing to throw an error and stop (which is what happens now), or you might want it to interrupt the long-running computation and continue executing other cells (e.g., in the case of an autograder). So, this PR make it an option to choose what happens when a cell times out -- either throw an error (the default), or interrupt and continue executing the rest of the cells.

This conflicts with #7883 so will require some reworking after that is merged.

### DIFF
--- a/IPython/nbconvert/preprocessors/execute.py
+++ b/IPython/nbconvert/preprocessors/execute.py
@@ -4,13 +4,14 @@
 # Distributed under the terms of the Modified BSD License.
 
 import os
+from textwrap import dedent
 
 try:
     from queue import Empty  # Py 3
 except ImportError:
     from Queue import Empty  # Py 2
 
-from IPython.utils.traitlets import List, Unicode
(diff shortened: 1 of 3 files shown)


jhamrick:
Rebased now that #7883 has been merged.

minrk:
Looks like this needs a rebase again

jhamrick:
Rebased.

jhamrick:
Travis is passing now, too :)

Carreau:
This does not change API right ? So it could be in 3.1 ?

Carreau:
Optimistically marked as 3.1 to merge "early".

jhamrick:
Well, it does add a new config option. But it doesn't change the default behavior at all.

takluyver:
I think it's probably OK for 3.1

PR #7885, opened 2015-02-26
dynamic traits for HasTraits instance

### LINKED ISSUE #7875: Dynamic Traits for HasTraits Instances and Classes
Currently the installation of dynamic traits for `HasTraits` objects is something that remains absent from the IPython Traitlets module. From what I can see this problem has two prongs: one with installation at a class level (relatively simple) and the other with installation at an instance level (more nuanced). The first prong is for creating traits that are accessed by all the instances of a particular class. The other is for creating traits that are only accessed by a single instance and which remain invisible to all others. I currently have solutions for both problems and will be making a pull request shortly.

---

In either case there will probably need to be a function which takes a `TraitType`, default value, and other arguments to initialize a new `TraitType` instance that will later be installed. It might something look like what's shown below, where `args` and `kw` would act as ways for initializing the default value for a `ClassBasedTraitType`.

```python
def trait_init(type, entry, args=None, kw=None, **metadata):
    if issubclass(type, TraitType):
        if issubclass(type, ClassBasedTraitType):
            if args is None:
                args=()
            if kw is None:
                kw={}
            instc = type(entry,args,kw,**metadata)
        else:
            # type is not class based
            instc = type(entry,**metadata)
            instc.default_value = entry
        return instc
    else:
        raise TraitError('{0} must inherit from TraitType'.format(type))
```

---

For the class level problem, the function that would actually install the trait would be fairly simple (I have a method for removing them as well but describing how they get there should give hints as to how removal works). The only real bit that needs to be done is to set the default value in each of the classes instances. I have come up with a way to accomplish this using two things. The first is a class attribute `_instances` which is a `set` associated with a decorator that uses weakrefs to store `HasTraits` instances. So the `__init__` for `HasTraits` would now include a line of code like the following `self.__class__._instances.add(weakref.ref(self))`. The second would then be the decorator for the attribute  `_instances`. This is shown below and would be used to return all the instances of a particular class:

```python
    @classmethod
    def instances(cls):
        """Get a list of class instances"""
        dead = set()
        for ref in cls._instances:
            obj = ref()
            if (obj is not None):
                if isinstance(obj,cls):
                    yield obj
                else:
                    pass
            else:
                dead.add(ref)
        cls._instances -= dead
```

With this, the next step is to actually implement the function which will do the installing. With a method that will return all the instances of a particular class the function shown below only needs to iterate and set default values. It might look something like what's shown below:

```python
def install_trait(klass, name, trait):
    if issubclass(klass,HasTraits):
        trait.this_class = klass
        trait.name = name
        for obj in klass.instances():
            trait.set_default_value(obj)
        setattr(klass, name, trait)
    else:
        raise TraitError()
```

---

Now moving on to the instance level problem. From what I can see there are two solutions; one involving metaclasses and another using a dictionary attribute in the instance to store `TraitType` objects.

For the metaclass solution, I was talking with @jdfreder who suggested a method where one could use a `HasTraits` object, that was created via a metaclass and stored as an instance level attribute, to hold dynamic traits. At this point it would simply be a matter of using the methods I described above to install `TraitType` objects at a class level.

The non-metaclass solution would override `__getattr__` and `__setattr__` in `HasTraits` to force the instance to utilize `__set__` and `__get__` from a trait in its dynamic trait dictionary (if such a trait existed). With this, the only thing required to install a new dynamic trait would be to add an initialized trait to the dynamic trait dictionary of an instance. I have a rough version of this working at the moment, however it could use some touch ups.

I was told by @ellisonbg and Jonathan that @sylvaincorlay and @jasongrout might have some suggestions on either the class or instance level problems of installing dynamic traits for `HasTraits` objects. This was already getting a bit long to write so let me know if there are questions on what I've written.

rmorshea:
A non-metaclass solution for creating dynamic traits in HasTraits instance. This overrides `__getattr__`, `__setattr__`, and `__hasattr__` in HasTraits to force the instance to utilize `__set__` and `__get__` from a trait in its dynamic trait dictionary (if such a trait exists).

closes #7875

### DIFF
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -242,38 +242,6 @@ def trait_init(type, entry, args=None, kw=None, **metadata):
     else:
         raise TraitError('{0} must inherit from TraitType'.format(type))
 
-def install_trait(klass, name, trait):
-    """Instantiate a trait for klass that is linked to its instances
-    Parameters
-    ----------
-    klass : class
-        The class which will be assigned new traits.
-    name : string
-        The name of the class attribute.
-    trait : TraitType
-        A preinitialized TraitType instance
-    """
-    if issubclass(klass,HasTraits):
-        trait.this_class = klass
-        trait.name = name
-        for obj in klass.instances():
-            trait.set_default_value(obj)
-        setattr(klass, name, trait)
-    else:
-        raise TraitError('{0} must inherit from HasTraits'.format(klass))
-
-def discard_trait(klass, name):
-    """Remove the given trait from klass and all its instances"""
-    delattr(klass,name)
-    for obj in klass.instances():
-        if name in obj.trait_names():
-            delattr(obj, name)
-            del obj._trait_values[name]
-            if name in obj._trait_notifiers.keys():
-                del obj._trait_notifiers[name]
-            if hasattr(obj,'_'+name+'_changed'):
-                delattr(obj,'_'+name+'_changed')
-
 def _validate_link(*tuples):
     """Validate arguments for traitlet link functions"""
     for t in tuples:
@@ -626,8 +594,6 @@ class MetaHasTraits(type):
 
 class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
 
-    _instances = set()
-
     def __new__(cls, *args, **kw):
         # This is needed because object.__new__ only accepts
         # the cls argument.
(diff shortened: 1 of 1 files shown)


takluyver:
I think this is superseded by #8124 - someone reopen if I'm wrong.

rmorshea:
It's just a different way of solving the same problem; @jdfreder uses metaclasses while this overrides __setattr__ and __getattr__ and stores the new traits in a dictionary at an instance level. All depends on what's preferable.

PR #7888, opened 2015-02-27
Don't recompute values

Carreau:
Minor performance improvement.
Was just annoying me to debug the called function.


minrk:
:+1:, will merge after release

PR #7889, opened 2015-02-27
Js completion traceback, erase line on tab, and double complete.

### LINKED ISSUE #4777: Tab completion on `__` deletes line in notebook
In the notebook:

```python
x = object
x.__dict__.__<tab><tab>
```

First tab will create a drop down. the second one will delete the whole line.

Carreau:
The function in charge of actually converting cursor offset to
CodeMirror line number and character number was actually crashing when
the cursor was at the last character (loop until undefined, then access
length of variable, which is undefined).

This was hiding a bug in which when you would completer to a single
completion pressing tab after as-you-type filtering, the completion
would be completed twice.

The logic that was supposed to detect whether or not all completions had
a common prefix was actually faulty as the common prefix used to be a
string but was then changed to an object. Hence the logic to check
whether or not there was actually a common prefix was always true, even
for empty string, leading to the deletion of the line (replace by '') in
some cases.

---

This fixes both an issue which is less than a day old, and one which is more than a year old.

I'm slightly concerned by the 50 ms timeout, that **might** have been a workaround for firefox.

This does fix a bug, so I'm inclined for  3.1, but I'm also fine with 4.0

### DIFF
--- a/IPython/html/static/base/js/utils.js
+++ b/IPython/html/static/base/js/utils.js
@@ -507,11 +507,12 @@ define([
         /**
          * turn absolute cursor postion into CodeMirror col, ch cursor
          */
-        var i, line;
+        var i, line, next_line;
         var offset = 0;
-        for (i = 0, line=cm.getLine(i); line !== undefined; i++, line=cm.getLine(i)) {
-            if (offset + line.length < cursor_pos) {
-                offset += line.length + 1;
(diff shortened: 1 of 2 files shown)


minrk:
We aren't so strict about backporting improvements early in the cycle unless they are API *changes*.

PR #7893, opened 2015-02-27
Add Markdown to the list of downloadable nbconverted formats.

Carreau:
Seem it was just forgotten. The list was not auto generated on purpose
IIRC, as some format, like TeX were deemed not useful in menu.


minrk:
:+1: the list is short and not really changing over time, so generating the list is probably overkill.

PR #7901, opened 2015-02-28
Updated Windows HOME folder

2-5:
Starting from Vista, the Windows HOME folder was changed from
"C:\Documents and Settings" to "C:\Users"

See the bottom of this page:
https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457%28v=vs.85%29.aspx

Since Windows XP is not supported anymore, IPython should update the displayed path.


Carreau:
Seem legit. Thanks !

PR #7902, opened 2015-02-28
fix margin in RadioButtons

gertingold:
This pull request is related to #7740. The change made there fails if the RadioWidget has a label to the left of the radio buttons. The label is set relative to the radio button labels with a padding of 8px but the radio buttons are set with a margin of -20px, thus overwriting the label on the left. This is resolved here by adding a margin of 20px to the radio button labels, thereby leaving space for the radio button and creating sufficient distance to the label on the left.

### DIFF
--- a/IPython/html/static/widgets/less/widgets.less
+++ b/IPython/html/static/widgets/less/widgets.less
@@ -271,9 +271,6 @@
 
     label {
         margin-top: 0px;
+        margin-left: 20px;
     }
-}
-
-.widget-radio {
-    margin-left: 20px;
 }


takluyver:
What happens if the label is on the right of the button? Won't this reintroduce the bug fixed by #7740 in that case?

gertingold:
I don't think so. If there is no description label present on the left, the radio button labels are at 20px from the left margin of the surrounding box. The radio buttons sit at -20px with respect to the radio button labels and are thus perfectly aligned with the surrounding box. So, the behavior in the absence of a description is label is unchanged.

gertingold:
Sorry, I should have included the following pictures right away.

This is the current behavior:
before_change

And this is the behavior after my patch has been applied:
after_change

The first example refers to the situation discussed in #7740, the second example shows the problem with the description label which is partially covered in the first picture but fully visible in the second one.

Carreau:
Seem fine to me. Should we mark 3.1 ?

jdfreder:
Thanks @gertingold , this looks good.

jdfreder:
Whoops, looks like the .css is missing - #7930

Issue #7759, opened 2015-02-12
More useful presentation of inspection info


takluyver:
When you first summon a tooltip on a function, before you hit shift-tab enough times to expand it, the signature is far enough down to be out of view. Instead you see type, repr and path to its module, none of which are especially useful.

This is a reminder to myself to re-evaluate the presentation of this information, re-order it and possibly eliminate some fields. I've been meaning to do this for years.

Carreau:
@stefanv would be pleased as he said tooltip was not heigh enough.

PR #7903, opened 2015-02-28
Rearrange info display to put most useful fields first

### LINKED ISSUE #7817: Method help popovers have become less useful in IPython 3.0.rc1
Forking from #7816 

I've noticed that the help popover (i.e., what you get when you hit shit-TAB in side the text of a function call) now has a lot of information that is not very useful to the task of finding out the function signature for a method.

I've particularly noticed this for Pandas Dataframes, though I assume the issue is more general than that. It's particularly annoying with DataFrames because the textual `__repr__` looks really bad, especially for large frames. (And also because I can never remember how to use their methods).

Here's an example of what we used to get:

screenshot 2015-02-19 10 38 29

And here's what we get now:

screenshot 2015-02-19 10 37 44

(I should note that putting together this issue caused me to note how much more attractive I find the 3.0/Jupyter styling. Well done!)

takluyver:
Closes #7817 and #7759.

This adds specific orderings of fields for basic callables (classes, functions, methods), magics and aliases, to put the most useful information first. Inspecting any other object will still produce the same info as before.

Before and after pics - function:
scatter_before_after

And magic:
timeit_before_after

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -554,23 +554,6 @@ class Inspector:
                 title = header((title+":").ljust(title_width))
             out.append(cast_unicode(title) + cast_unicode(content))
         return "\n".join(out)
-
-    # The fields to be displayed by pinfo: (fancy_name, key_in_info_dict)
-    pinfo_fields1 = [("Type", "type_name"),
-                    ]
-                    
-    pinfo_fields2 = [("String form", "string_form"),
-                    ]
-
-    pinfo_fields3 = [("Length", "length"),
-                    ("File", "file"),
-                    ("Definition", "definition"),
-                    ]
-
-    pinfo_fields_obj = [("Class docstring", "class_docstring"),
-                        ("Init docstring", "init_docstring"),
-                        ("Call def", "call_def"),
-                        ("Call docstring", "call_docstring")]
     
     def _format_info(self, obj, oname='', formatter=None, info=None, detail_level=0):
         """Format an info dict as text"""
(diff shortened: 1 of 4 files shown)


minrk:
Test failures like some missing updates for s/Definition/Signature/.  Otherwise, :+1:

takluyver:
OK, tests fixed.

IPython/core/oinspect.py
```diff
@@ -582,41 +566,62 @@ def add_fields(fields):
...
+            # source found.
+            if detail_level > 0 and info['source'] is not None:
+                displayfields.append(("Source",
+                                      self.format(cast_unicode(info['source']))))
+            elif info['docstring'] is not None:
+                displayfields.append(("Docstring", info["docstring"]))
+
+            add_fields([("Class docstring", "class_docstring"),
+                        ("Init docstring", "init_docstring"),
+                        ("Call def", "call_def"),
```
minrk: other definitions are renamed to Signature, should this one be as well?

minrk:
The only disadvantage I see to this approach is that moving the fields into the body of the method away from attributes makes it harder to modify the structure/order of entries. I'm not sure anyone has ever done that, though, so I'm not concerned about it.

Carreau:
+1 for that as this is one of the things people are the most pissed about. (Or they complain louder)

IPython/core/oinspect.py
takluyver: Done. Let's hope that doesn't make the tests fail again.

takluyver:
Yeah, I don't think we should really be too concerned about people subclassing this and overriding attributes.

IPython/core/oinspect.py
```diff
@@ -582,41 +566,62 @@ def add_fields(fields):
...
+
+            add_fields([("String form", "string_form")])
+
+            # Namespace
+            if info['namespace'] != 'Interactive':
+                displayfields.append(("Namespace", info['namespace'].rstrip()))
+
+            add_fields([("Length", "length"),
+                        ("File", "file"),
+                        ("Definition", "definition"),
```
minrk: one last Definition

takluyver: Done

PR #7905, opened 2015-02-28
Fix read-only file test causing errors on Windows

takluyver:
Will run Windows tests against this.


takluyver:
No, that didn't seem to fix it.

minrk:
Oh, crazy, that's definitely not the commit I meant to push.

minrk:
wait, this isn't my PR, duh. I just skipped the test in #7906.

minrk:
Also for the record, the test passes just fine on my Windows 7 VM, so it might be something related to the Windows Server VM on Jenkins running as admin.

takluyver:
It's like that paradox about whether an omnipotent deity can create a weight that he cannot lift...

Closing in favour of #7906

PR #7908, opened 2015-02-28
actually catch failure to load widget state

minrk:
load callback was called directly, rather than in promise function that catches errors.

cc @jdfreder


jdfreder:
Thanks @minrk , this makes sense since it can't try/catch the function if it has already been called.  @jasongrout this is something to keep our eyes open for, I may have done it elsewhere, or I may forget in the future.

PR #7911, opened 2015-02-28
Make suffixes be optional in nbconvert

jhamrick:
The output suffix (e.g. `.nbconvert`) is currently _always_ applied, which is annoying because it's not always appropriate -- for example, when using a different build directory, you would expect the files to have the same name, and not always have an additional `.nbconvert` added on to them. This PR makes the output suffix be optional, and additionally adds a flag (`--no-output-suffix`) to make it easy to suppress this option on the command line.

I'm also fine not adding the flag and just having it be `--NbConvertApp.use_output_suffix=False` if you'd prefer to keep the number of command line flags minimal.

### DIFF
--- a/IPython/nbconvert/nbconvertapp.py
+++ b/IPython/nbconvert/nbconvertapp.py
@@ -18,7 +18,7 @@ from IPython.core.application import BaseIPythonApplication, base_aliases, base_
 from IPython.core.profiledir import ProfileDir
 from IPython.config import catch_config_error, Configurable
 from IPython.utils.traitlets import (
-    Unicode, List, Instance, DottedObjectName, Type, CaselessStrEnum,
+    Unicode, List, Instance, DottedObjectName, Type, CaselessStrEnum, Bool,
 )
 from IPython.utils.importstring import import_item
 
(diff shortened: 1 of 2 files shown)


minrk:
I've also thought about an `--inplace` flag, which seems like it would be a common desire when using `--execute`. I guess this achieves the same thing, but is more general? How do you express output with the same name, but in a different directory with the current config?

jhamrick:
If you do `--FilesWriter.build_directory=some_directory` then the files will end up in `some_directory`. I like the idea of having an `--inplace` flag, though. Maybe it would actually be better to have the following behavior:

* if the build directory is not specified, add the suffix by default
* if the build directory is not the current directory, always suppress the suffix
* add an `--inplace` flag, which suppresses the suffix (but in effect only really matters when the build directory isn't specified)

minrk:
I like that proposal. You might make the `--inplace` flag also unset FilesWriter.build_directory, so that build_directory can be set in a config file, and overridden via `--inplace` on the command-line.

jhamrick:
Good idea. I'll update the PR so it does these things.

jhamrick:
Ok, updated.

minrk:
Thanks!

PR #7912, opened 2015-03-01
Add relpath option to FilesWriter

jhamrick:
Currently, if you specify files that a notebook is dependent on, the current behavior is to copy the files relative to the directory that you are running nbconvert from. However, you may want them to be relative to a different directory -- for example, in the case where you're converting `source_dir/*.ipynb` (and there are dependent files in `source_dir`) to a build directory of `build_dir/`, then currently you'll get the dependent files as `build_dir/source_dir/some_dependent_file`. This is not always what you want, so you should be able to specify what the relative path is.

This PR adds an option for a relative path, so in the example above, you could specify `FilesWriter.relpath="source_dir"`, and it would properly link the files to `build_dir/`, rather than `build_dir/source_dir/`.

### DIFF
--- a/IPython/nbconvert/exporters/exporter.py
+++ b/IPython/nbconvert/exporters/exporter.py
@@ -142,9 +142,10 @@ class Exporter(LoggingConfigurable):
             resources = ResourcesDict()
         if not 'metadata' in resources or resources['metadata'] == '':
             resources['metadata'] = ResourcesDict()
-        basename = os.path.basename(filename)
+        path, basename = os.path.split(filename)
         notebook_name = basename[:basename.rfind('.')]
         resources['metadata']['name'] = notebook_name
+        resources['metadata']['path'] = path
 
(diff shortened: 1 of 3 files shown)


jhamrick:
Travis is failing on some other test group (kernel), which I'm pretty sure should be unaffected by this PR.

Carreau:
Ah, that was what you were talking about in nbgrader....

I suppose you will stick to 3.x for your class, so marking as 3.1 I guess.

Carreau:
+1

jhamrick:
Yeah, this is one of the things that nbgrader will soon depend on. I think how it works now is kind of unexpected behavior anyways, so is probably appropriate for 3.1.

minrk:
From gitter discussion:

- using relpath makes sense
- further, it should be the default behaviour that relpath is the directory containing the notebook being converted

jhamrick:
Ok, this has been updated so that it uses the directory containing the notebook as the default relpath.

Issue #7909, opened 2015-02-28
nbconvert heading numbering different in 3.0


rlabbe:
I have a very large notebook, but I have reduced the problem to a small demo. I am using notebook to generate a book in PDF format, and use the headings to specify chapters, sections, and so on. I.e. 

```
#Chapter 1 
text
##section 1
text
#Chapter 2
txt
```

and so on. I wrote a custom book.tplx file to generate a PDF from the notebook in the style that I want. The notebook is in IPython 3.0 format.

In IPython 3.0 I get this as my table of contents

0.1 Chapter 1
0.1.1 Section 1
0.2 Chapter 2 

In 2.4 I get (by this I mean I run IPython 2.4 on the same IPython 3.0 notebook)

1 Chapter 1
1.1 Section 1
2 Chapter 2

A second difference is that in IPython 3.0 all of the content for both chapters end up on one page, whereas in 2.4 it correctly starts a new page for each chapter (not suprising since 3.0 doesn't seem to recognize that the chapter has changed).

There are other differences, but all I believe due to the fact that IPython 3.0 does not recognize header 1 (#sometext) cells as chapters.

It probably isn't clear from the comments, but it is the behavior of IPython 2.4 that I want.

My commands are:

```
2.4: ipython nbconvert --to latex --post PDF --template book notebook.ipynb
3.0: ipython nbconvert --to PDF --template book notebook.ipynb
```

notebook.ipynb:

```
{
 "cells": [
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "# Chapter 1"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "hi there"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "## Section 1"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "more text"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "# Chapter 2"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "text"
   ]
  }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
[...]

PR #7914, opened 2015-03-01
Enable chapters for latex report class

jakobgager:
As shown in issue #7909 the latex report template misses chapter headings and thus results in headings always starting with _0_. This adds the required `--chapters` pandoc parameter as proposed by @minrk in the respective issue.


minrk:
Thanks

PR #7915, opened 2015-03-01
pass Session key around in config in ipcontroller

minrk:
ensures Task Scheduler, etc. share the same key

should fix test failures in IPython.parallel caused by #7832

/cc @takluyver

### DIFF
--- a/IPython/parallel/apps/ipcontrollerapp.py
+++ b/IPython/parallel/apps/ipcontrollerapp.py
@@ -357,6 +357,9 @@ class IPControllerApp(BaseParallelApplication):
         self.factory.hub.engine_state_file = os.path.join(self.profile_dir.log_dir, fname)
         if self.restore_engines:
             self.factory.hub._load_engine_state()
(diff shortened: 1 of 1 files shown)


Carreau:
+1

takluyver:
I'm running the Windows tests on it now, but the test failures I've been seeing are in IPython.kernel tests, not IPython.parallel - e.g. https://jenkins.jupyter.org/job/windows-py3.4/149/testReport/

takluyver:
OK, it seems that it has fixed the issue. Not sure how, but: https://jenkins.jupyter.org/job/windows-py3.4-prtest/3/testReport/

takluyver:
Nope, still failing: https://jenkins.jupyter.org/job/windows-py3.4/150/testReport/

Though it does seem to have fixed a hang in the parallel tests on Linux, which is good :+1:

minrk:
Ah, the one in IPython.kernel is because of low resolution timers on Windows. Fix on the way.

minrk:
#7918 should do it

PR #7918, opened 2015-03-02
rev msg_id to avoid signature collisions

minrk:
instead of date, since that can match on low-res timers (Windows)

should fix collisions described in #7915

### DIFF
--- a/IPython/kernel/zmq/tests/test_session.py
+++ b/IPython/kernel/zmq/tests/test_session.py
@@ -91,7 +91,7 @@ class TestSession(SessionTestCase):
         
         content = msg['content']
         header = msg['header']
-        header['date'] = datetime.now()
+        header['msg_id'] = self.session.msg_id
         parent = msg['parent_header']
         metadata = msg['metadata']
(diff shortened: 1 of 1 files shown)


IPython/kernel/zmq/tests/test_session.py
```diff
@@ -100,7 +100,7 @@ def test_send(self):
         ident, msg_list = self.session.feed_identities(B.recv_multipart())
         new_msg = self.session.deserialize(msg_list)
         self.assertEqual(ident[0], b'foo')
-        self.assertEqual(new_msg['msg_id'],msg['msg_id'])
+        self.assertEqual(new_msg['msg_id'],header['msg_id'])
         self.assertEqual(new_msg['msg_type'],msg['msg_type'])
```
takluyver: Should these other asserts be changed to use `header` instead of `msg` as well?

minrk: the entire header is already compared below. This is testing that the msg_id is copied from the header to top-level on recv (a minor convenience copy).

PR #7926, opened 2015-03-02
Fix configurability of codecell

### LINKED ISSUE #7896: Tooltip for class constructors gone in IPython 3.0
The IPython 3.0 upgrade now shows less helpful tooltips. In particular, tooltips for class constructors are now gone.

This is what we get in IPython 2.4:

ipython2 4

And this is what we get now:

ipython3 0

Also worth notice is that auto closing parentheses is also missing in the new version.

Carreau:
code cell where actually not configurable,
this fixes that + add a convenient methods to dump the
content of the configuration in the config file.

this also close #7896 by defaulting closebracket to true,
which is now configurable :-)

It also switches the default magic highlight mode to string so that they
could also be configured.

### DIFF
--- a/IPython/html/static/notebook/js/codecell.js
+++ b/IPython/html/static/notebook/js/codecell.js
@@ -147,17 +147,6 @@ define([
 
     CodeCell.prototype = Object.create(Cell.prototype);
     
-    /**
-     * @private
-     * Update the current state of the config file in the profile with current config. 
-     **/
-    CodeCell.prototype._fill_config_with_current = function(){
-        var keys = ['cm_config','highlight_modes'];
-        for(var i in keys){
-            this.class_config.set(keys[i], this.class_config.get_sync(keys[i]));
-        }
-    }
-
     /** @method create_element */
     CodeCell.prototype.create_element = function () {
         Cell.prototype.create_element.apply(this, arguments);
--- a/IPython/html/static/notebook/js/codecell.js
+++ b/IPython/html/static/notebook/js/codecell.js
@@ -127,25 +127,36 @@ define([
             },
             mode: 'ipython',
             theme: 'ipython',
-            matchBrackets: true
-        }
-    };
-
-    CodeCell.config_defaults = {
+            matchBrackets: true,
+            autoCloseBrackets: true
+        },
         highlight_modes : {
-            'magic_javascript'    :{'reg':[/^%%javascript/]},
-            'magic_perl'          :{'reg':[/^%%perl/]},
-            'magic_ruby'          :{'reg':[/^%%ruby/]},
-            'magic_python'        :{'reg':[/^%%python3?/]},
-            'magic_shell'         :{'reg':[/^%%bash/]},
-            'magic_r'             :{'reg':[/^%%R/]},
-            'magic_text/x-cython' :{'reg':[/^%%cython/]},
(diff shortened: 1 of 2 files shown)


IPython/html/static/notebook/js/codecell.js
```diff
@@ -127,25 +127,36 @@ define([
...
 
+    CodeCell.config_defaults = CodeCell.options_default;
+
     CodeCell.msg_cells = {};
 
     CodeCell.prototype = Object.create(Cell.prototype);
+    
+    /**
+     * @private
+     * Update the current state of the config file in the profile with current config.
```
takluyver: Our general policy is that we don't create config files with the default values, because it makes it harder to change them later. I'm sure there's stuff we need to think more about, but I don't think this is the right approach.

Also, this appears to be a private method, but it's never called.

takluyver:
Does this go into 3.1? Or is it likely to break people's customisation? If we hold it off to 4.0, we should make another PR that just re-enables auto bracket closing.

IPython/html/static/notebook/js/codecell.js
Carreau: The problem is that if you want to change something with current configs, it is **really hard**. With that you can dump easily the current config back to the file.

So by default the file would be empty, but if you call this explicitely, then you "just" have to modify the json on disk.
I can remove it though.

Carreau:
No it wont break people customisation. The thing is with the current state of 3.0 it is **not** configurable with `nbconfig/ntebook.json`

IPython/html/static/notebook/js/codecell.js
takluyver: Perhaps for things like codemirror config, we should merge the user's config with the default, rather than using the user's config *or* the default.

takluyver:
But people may change the options in custom.js - see e.g. #7984.

Carreau:
> But people may change the options in custom.js - see e.g. #7984.

I'm not sure I follow you ? #7984 is already not working.

takluyver:
No, but that's probably just a matter of requiring something or waiting for the relevant event. The CodeCell options can be customised, and people are using that.

takluyver:
Looking at it some more, this shouldn't break the customisation done in #7984. I still think we should think about merging user options for things like cm_config with the default, but that can be a 4.0 thing if you like. In any case, I would get rid of the unused private method before merging this.

Carreau:
> I still think we should think about merging user options for things like cm_config with the default

Not sure what you meant. 

> I would get rid of the unused private method before merging this.

Done.

takluyver:
What I meant about merging: currently we do something like this:

```javascript
cm_config = user_config.cm_config || defaults.cm_config;
```

If we instead do something like this, then writing the defaults to the file would be unnecessary:

```javascript
cm_config = $.extend(defaults.cm_config || {}, userconfig.cm_config);
```

(That code is illustrative, not a real part of the code)

takluyver:
Do you want to look into that in this PR, or leave it for later examination?

Carreau:
> Do you want to look into that in this PR, or leave it for later examination?

Hum, I would do that later. But I can try to make it for 3.1. I'll open a separate issue.

minrk:
Doing it live!

PR #7927, opened 2015-03-02
fix the links to pythonxy

mashenjun:
Hi, 
in the docs/source/links.txt, the links to pythonxy is not valid. I have change it to the right address which is https://code.google.com/p/pythonxy/
Thanks


takluyver:
Thanks

PR #7938, opened 2015-03-03
Fullunicode completion

Carreau:
Alternative to #7937 to play with it see how it feels.

implement completion based on full name of symbol
plus back completion, because it's convenient

```
\gamma -> γ
\γ -> \GREEK SMALL LETTER GAMMA
\GREEK SMALL LETTER GAMMA -> γ

\☁ -> \CLOUD
\CLOUD -> ☁

\black sun with rays -> ☀
\BLACK SUN WITH RAYS -> ☀

------

\pile of poo -> 💩 ## unicode of length 2 break completion completly afterward.
```

(note that font is crap for unicode on github, much nicer on notebook. )

### DIFF
--- a/IPython/core/latex_symbols.py
+++ b/IPython/core/latex_symbols.py
@@ -1295,3 +1295,6 @@ latex_symbols = {
     "\\mtteight" : "𝟾",
     "\\mttnine" : "𝟿",
 }
+
+
+reverse_latex_symbol = { v:k for k,v in latex_symbols.items()}
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -476,6 +476,64 @@ def _safe_isinstance(obj, module, class_name):
 
 
 
+def back_unicode_name_matches(text):
+    u"""Match unicode characters back to unicode name
+    
+    This does  ☃ -> \\snowman
+
+    Note that snowman is not a valid python3 combining character but will be expanded.
+    Though it will not recombine back to the snowman character by the completion machinery.
+
+    This will not either back-complete standard sequences like \n, \b ...
+    
+    Used on Python 3 only.
+    """
+    if len(text)<2:
+        return u'', ()
+    maybe_slash = text[-2]
+    if maybe_slash != '\\':
+        return u'', ()
+
+    char = text[-1]
+    # no expand on quote for completion in strings.
+    # nor backcomplete standard ascii keys
+    if char in string.ascii_letters or char in ['"',"'"]:
+        return u'', ()
+    try :
+        unic = unicodedata.name(char)
+        return '\\'+char,['\\'+unic]
+    except KeyError as e:
+        pass
+    return u'', ()
+
+def back_latex_name_matches(text):
+    u"""Match latex characters back to unicode name
+    
+    This does  ->\\sqrt
+
+    Used on Python 3 only.
+    """
+    if len(text)<2:
+        return u'', ()
+    maybe_slash = text[-2]
+    if maybe_slash != '\\':
+        return u'', ()
+
+
+    char = text[-1]
+    # no expand on quote for completion in strings.
+    # nor backcomplete standard ascii keys
+    if char in string.ascii_letters or char in ['"',"'"]:
+        return u'', ()
+    try :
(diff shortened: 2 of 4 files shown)


IPython/core/latex_symbols.py
```diff
@@ -1294,4 +1294,5 @@
     "\\mttseven" : "𝟽",
     "\\mtteight" : "𝟾",
     "\\mttnine" : "𝟿",
+    "\\snowman" : "☃",
```
minrk: If I understand correctly, the unicode_name_matches means this doesn't need to be added here?

Carreau: Yes, sorry, should have discarded that,

Carreau:
And the rest of the implementation could be rewritten more elegantly.
just wanted to try. Don't want to make a clean implementation, if this is something with a strong push back.

ncoghlan:
To help prevent syntax errors, you may want to filter the result based on its "isidentifier" method. Otherwise folks will be able to enter characters that won't actually work as identifiers (including, alas, the snowman).

Carreau:
You can still enter snowman in comments :-), as well as cats... and we use isidentifier while [precomputing the list](https://github.com/Carreau/ipython/blob/fullunicode/tools/gen_latex_symbols.py#L52), but true, I can do it here also. 

I'm sure you could push a change as what categories of symbols are identifiers in future version of Python :wink:.

minrk:
Why the back unicode completion? That seems really weird for tab completion to undo itself every other time it's triggered.

Carreau:
I dont' see another simple way to know how to know how to type a character you don't know the name. 

You prepend `\`, \☁ -> `\CLOUD` , "Oh, it's cloud", press tab again, and you are back at original code.

ellisonbg:
Does this preserve the existing latex -> unicode completion?

I agree with @minrk that it is a bit odd to have reverse completion. 

I do think this should only work for valid identifiers though.

Carreau:
Yes it preserve latex.

if you see ʘ, how do you type it ? How do you know it's name ?

It already accept only valid identifiers.

On Sun, Mar 29, 2015 at 11:56 AM, Brian E. Granger <notifications@github.com

ellisonbg:
OK, I am more positive if this preserves the latex completion and only
accepts valid identifiers. I think the reverse completion adds complexity
that few people will use, but I dont' feel too strongly about that point.

On Sun, Mar 29, 2015 at 12:31 PM, Matthias Bussonnier <

> Yes it preserve latex.
>
> if you see ʘ, how do you type it ? How do you know it's name ?
>
> It already accept only valid identifiers.
>
> On Sun, Mar 29, 2015 at 11:56 AM, Brian E. Granger <
> notifications@github.com
> > wrote:
>
> > Does this preserve the existing latex -> unicode completion?
> >
> > I agree with @minrk <https://github.com/minrk> that it is a bit odd to
> > have reverse completion.
> >
> > I do think this should only work for valid identifiers though.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ipython/ipython/pull/7938#issuecomment-87455246>.
> >
>
> —
> <https://github.com/ipython/ipython/pull/7938#issuecomment-87461044>.
>



-- 
Brian E. Granger

minrk:
The uncompletion behavior still seems weird to me, but we can keep it if you like it. I do think this needs some more tests before merge, though.

takluyver:
From the dev meeting:

- Tests!
- Including tests that e.g. `\a<TAB>` doesn't expand to `\LATIN SMALL LETTER A`.
- Expansion should offer both latex and long name forms, prioritising the latex when it exists.

Carreau:
Travis seem happy. 

I think i might make `back_unicode_name_matches` and `back_latex_name_matches`more generic with a higher order function, but that might be overkill.

minrk:
@Carreau your last comment suggests you still have more work to do here. Is that true?

Carreau:
The "might be overkill" was more toward a question of what do you think ?

I's a 3 line difference in two 10 line function.

minrk:
I think it's fine as-is.

PR #7951, opened 2015-03-03
make 'download as PDF' say 'via LaTeX'

minrk:
so it's clear that LaTeX is required


takluyver:
+1

jdfreder:
This is great, thanks @minrk !

Issue #7941, opened 2015-03-03
ipython 3.0 broke my readthedocs build


njsmith:
Not sure if this is a bug but it sure was annoying.

patsy.readthedocs.org uses `IPython.sphinxext.ipython_directive` and `IPython.sphinxext.ipython_console_highlighting`. It installs these by providing RTD with a requirements.txt which contains the string `ipython`, so `pip install ipython` happens.

When 3.0 arrived on PyPI, this stopped working: suddenly my RTD builds are failing. Grovelling through the logs reveals that the `pip install ipython` went through fine, but then trying to import the sphinx extensions bailed out saying that I have to install `jsonschema`. Okay, fine, I added that to requirements.txt. Now it bailed out complaining about the lack of `mistune`. Okay, google says that should be pip installable too, so I added that to requirements.txt too. Now it seems to be working? But it seems like you have some broken install_requires or something?

jenshnielsen:
The same think happened for the matplotlib docs build. 

IPython defines a number of additional targets [here](https://github.com/ipython/ipython/blob/master/setup.py#L251) that can be installed with pip as 
```
pip install ipython[...]
```
The intention is that the default one should only install dependencies needed for the ipython terminal.
It seems to me that `jsonschema` and `mistune` should be added to the `doc` target.

takluyver:
Hmm. It shouldn't really require those, but `ipython_console_highlighting` imports `from ..nbconvert.utils.lexers import IPyLexer`. Nbconvert pulls in mistune and nbformat, and the latter pulls in jsonschema.

Probably the best fix would be to move the lexers to somewhere that both nbconvert and sphinxext can import them.

njsmith:
Is the [doc] target for using ipython to build docs, or for building
ipython's docs? Is this target documented anywhere?

Someone might also want to double-check that these are the only packages
missing from that dependency list. The error messages complained about the
"notebook" needing these extra packages, and some of the other
notebook-related packages listed in setup.py are installed on RTD for
unrelated reasons (e.g. jinja2), so the dependencies might have been
satisfied by accident.

takluyver:
I think the doc target is for building IPython's own docs, but I don't think this is written down anywhere.

minrk:
When some nbconvert handlers depended on sphinx, we tried to make sure they weren't imported unless those actual exporters were being used. We should probably do the same with mistune.

minrk:
[...]

PR #7953, opened 2015-03-03
nbconvert: don't require mistune unless it's used

minrk:
- moves mistune-related functionality to filters.markdown_mistune
- raises ImportError on call of markdown2html_mistune, rather than import of nbconvert

This partly, but not entirely, addresses #7941


takluyver:
:+1:

PR #7954, opened 2015-03-03
Move IPython lexers module to lib

### LINKED ISSUE #7941: ipython 3.0 broke my readthedocs build

takluyver:
Closes gh-7941

I've left a backwards compatibility module in IPython.nbconvert.utils.lexers - I don't know if anyone is importing it directly, but since we put it in entry points, we should probably consider it public API.


IPython/nbconvert/utils/tests/lexers.py
```diff
@@ -0,0 +1 @@
+__author__ = 'takluyver'
```
minrk: Did you mean to add this file?

takluyver: Sorry, fat fingering things with my IDE. Should be gone now.

Carreau:
+1.

PR #7955, opened 2015-03-04
don't merge App.extra_extension into App.extensions

minrk:
setting both values could result in extra_extension being omitted from the list

alternately, we could have `_extensions_changed` re-add it, but I don't think we use this list more than once.


takluyver:
+1 - I much prefer this approach to traits changed callbacks.

PR #7960, opened 2015-03-04
Ensure Cache-Control is set once and only once

### LINKED ISSUE #7959: Cache-Control being set twice for no_cache_paths
Whether browsers pick the first or last instance of a repeat header is ill defined. Some will pick the first (Firefox) and wind up caching custom.js practically forever (or until notebook server restart) and others will pick the latter and behave as intended.

screen shot 2015-03-03 at 10 41 59 pm

PR with the fix in a moment.

parente:
Whether browsers should pick the first or last instance of a duped header is ill defined

closes #7959

### DIFF
--- a/IPython/html/base/handlers.py
+++ b/IPython/html/base/handlers.py
@@ -404,7 +404,7 @@ class FileFindHandler(web.StaticFileHandler):
         # disable browser caching, rely on 304 replies for savings
         if "v" not in self.request.arguments or \
                 any(self.request.path.startswith(path) for path in self.no_cache_paths):
-            self.add_header("Cache-Control", "no-cache")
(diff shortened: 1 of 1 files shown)


parente:
Fix for #7959.

parente:
I don't see a true failure in the test run, only "PhantomJS has crashed. Please read the crash reporting guide at ..."

takluyver:
Thanks, this looks sensible. I have prodded that build to restart.

takluyver:
@minrk please merge this unless there's a reason to use add_header instead of set_header here.

minrk:
set_header's 100% the right thing to do.

PR #7961, opened 2015-03-04
General cleanup for python files in docs/

JHoloboski:
Just cleaned up the python files in docs to be a bit more readable. Will continue on with others if this type of thing is wanted.


takluyver:
Thanks for contributing, but we don't go for 'cleanup' PRs - they generate big diffs, lead to merge conflicts, and make *git blame* less useful, for no significant benefit. We sometimes clean up code as we're working on it, but we're don't adhere strongly to any particular style guide.

I'll add a mention of this to the CONTRIBUTING.md

PR #7965, opened 2015-03-04
Mention that values has been renamed to options

patricksnape:
This is a pretty big change that should be documented clearly! Just spent the last couple of hours trying to hunt this one down.


jdfreder:
Thanks @patricksnape !  I missed this while I was documenting the changes, sorry!

PR #7974, opened 2015-03-05
Execute notebooks in their directory, rather than the current working directory

### LINKED ISSUE #7967: ExecutePreprocessor should run from the notebook's directory
If you run `nbconvert` from the command line with the `--execute` flag, but are not in the same directory as the notebook, then the notebook won't run successfully if it depends on external files because the kernel will have been started in the cwd rather than the notebook's directory. So probably `ExecutePreprocessor` should chdir to the notebook's directory first, then chdir back to the original directory.

I'll make a PR for this later today.

jhamrick:
Fixes #7967

Incidentally, this also fixes the `Inline Image.ipynb` test, which was passing but probably shouldn't have been.


jhamrick:
Whoops, this actually won't work in all cases, because sometimes `path` is `''`. Will fix this momentarily...

jhamrick:
Fixed.

netheril96:
Won't this break many existing code?

Carreau:
> Won't this break many existing code?

That's what cell>run all does. So it's weird to expect the execute preprocessor to behave differently.

minrk:
@netheril96 given that this is brand new code released 6 days ago, there probably aren't many codes to break, yet. Plus, the most common case is running nbconvert in the directory with the notebook, where this has no effect. I think we can safely consider the current behavior a bug and this is a fix that makes the behavior consistent with the live notebook, which is the goal of the execute preprocessor.

Issue #7973, opened 2015-03-05
nbconvert: `--to latex --post PDF` deprecated in favour of `--to pdf`?


onesandzeroes:
Am I right in thinking that in v3.0, you no longer convert a notebook to pdf by using `--to latex --post PDF`? Trying to do that (from a script I had lying around that worked with earlier IPython versions) gave me an error:

```
Traceback (most recent call last):
  File "/usr/local/bin/ipython3", line 11, in <module>
    sys.exit(start_ipython())
  File "/usr/local/lib/python3.4/dist-packages/IPython/__init__.py", line 120, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
...
   very long traceback trimmed
...
  File "/usr/local/lib/python3.4/dist-packages/IPython/utils/importstring.py", line 50, in import_item
    return __import__(parts[0])
ImportError: No module named 'PDF'
```

Eventually I decided to just try directly converting using `--to pdf` and it worked fine.

The relevant [docs page](http://ipython.org/ipython-doc/stable/notebook/nbconvert.html) doesn't mention this, so if I'm right about the new behaviour I'm happy to submit a pull to update the docs.

takluyver:
You are right, and a PR would be welcome. :-)

takluyver:
Fixed by #7978

scottsievert:
+1. I had the same issue, Googled it and it looked like I was going to have to reinstall pandoc based on [a SO] question. `--to pdf` worked for me.

[a SO]:http://stackoverflow.com/questions/29156653/ipython-jupyter-problems-saving-notebook-as-pdf

startakovsky:
@takluyver  This is related to the documentation, because yes `--to pdf` works great with `pdflatex` installed, but this is the only place I was able to find the answer and the only thing the documentation states is doing `--to latex --post PDF`.  

Is it in the documentation?

takluyver:
nbconvert is being split out into a separate package, and the docs for it, showing `--to pdf`, are here:
http://nbconvert.readthedocs.org/en/latest/usage.html

The IPython 3.x docs also show `--to pdf`:
http://ipython.org/ipython-doc/3/notebook/nbconvert.html

Note that those docs are versioned (the `/3/` in the URL). If you're looking at docs for an older version, it might well show you the old way of doing it.

PR #7978, opened 2015-03-05
Docs: Update info about nbconvert's latex/pdf conversion

onesandzeroes:
As discussed in #7973. While updating the info about `--to pdf` I noticed that the `book` template seems to have been replaced by `report`, so I've updated that as well. If anyone wants to provide a more detailed description of the report template, feel free.

### DIFF
--- a/docs/source/notebook/nbconvert.rst
+++ b/docs/source/notebook/nbconvert.rst
@@ -59,8 +59,7 @@ The currently supported export formats are:
 
 * ``--to pdf``
 
-  Generates a PDF via latex. Replaces ``--to latex --post PDF``, which is deprecated.
-  Supports the same templates as ``--to latex``.
+  Generates a PDF via latex. Supports the same templates as ``--to latex``.
 
 * ``--to slides``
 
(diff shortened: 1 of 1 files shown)


docs/source/notebook/nbconvert.rst
```diff
@@ -43,21 +43,25 @@ The currently supported export formats are:
...
-    Latex book, derived from Sphinx's manual template.
+    Latex report, providing a table of contents and chapters.
 
   - ``--template basic``
   
     Very basic latex output - mainly meant as a starting point for custom templates.
 
+* ``--to pdf``
+
+  Generates a PDF via latex. Replaces ``--to latex --post PDF``, which is deprecated.
```
takluyver: I think it's just gone, not deprecated. It's probably not worth mentioning what this replaced.

onesandzeroes:
OK, I've removed the mention of the old behaviour. Let me know if there's any other changes to be made.

takluyver:
Thanks, I think this is looking good!

PR #7981, opened 2015-03-05
Refocus notebook after click on save

### LINKED ISSUE #7963: After clicking on the "Save"  in toolbar the focus does not return to notebook
As a result you need to click back on one of the cells in the notebook to continue editting.
This wasnt the case in ipython 2x. Also this does not happen if I click on another button in the toolbar or if I use File->Save and checkpoint from the menu.

Carreau:
closes #7963


takluyver:
Yep, that works

PR #7982, opened 2015-03-06
use $(window).height() instead of window.innerHeight

minrk:
to measure window height when resizing contents

Sometimes these values differ by a pixel, we may never know why.

window.innerHeight seems to be larger when they disagree, causing an extra scrollbar to be drawn at some zoom levels.

### DIFF
--- a/IPython/html/static/base/js/page.js
+++ b/IPython/html/static/base/js/page.js
@@ -56,7 +56,7 @@ define([
 
     Page.prototype._resize_site = function() {
         // Update the site's size.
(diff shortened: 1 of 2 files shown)


minrk:
ping @fperez for testing and @jdfreder for window-size wizardry

fperez:
For the record, didn't fix the problem here.  But thanks for the good intentions! :)

minrk:
I think this did fix at least one instance of an extra scrollbar for me, so one step closer, perhaps.

PR #7983, opened 2015-03-06
contents: double check that os_path is always within root

minrk:
check based on tornado `StaticFileHandler.validate_absolute_path`

I haven't been able to actually escape root, but we should have this check anyway.

/cc @jasongrout


takluyver:
:+1:

Carreau:
+1

PR #7993, opened 2015-03-07
JS: Unfreeze actions.

jdfreder:
You can't monkey patch the actions instance when frozen.

@Carreau

### DIFF
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -6,7 +6,6 @@ define(function(require){
     
     var ActionHandler = function (env) {
         this.env = env || {};
-        Object.seal(this);
     };
 
     /**


Carreau:
Why do you want to monkeypatch ?

jdfreder:
For me, logging, but you could imagine other use cases. I'd be okay with just adding an event if there's a reason you think we should keep the freeze.

Carreau:
The reason of the freezing is to avoid adding attributes due to typo, which  in "use strict" will fail. I catched a few error and weird patching with that. I'm fine with removing it. 

If you need it you ca go ahead and merge.

Carreau:
Seem with @jdfreder you can extend `init.prototype`  so no need to not-seal.

jdfreder:
Yes, thanks @Carreau !

PR #8000, opened 2015-03-09
Fix rename issues in dashboard

mathieu1:
- Do not display the full path of the selected item in the text box (**warning** : this removes the possibility of moving a file/folder using the rename dialog. Maybe someone finds this helpful)
- Adapt dialog for files and notebook.
- Do not allow renaming a running notebook (currently, renaming a running notebook would result in duplication instead of renaming after the open notebook saves/autosaves using the old name)

### DIFF
--- a/IPython/html/static/tree/js/notebooklist.js
+++ b/IPython/html/static/tree/js/notebooklist.js
@@ -438,8 +438,8 @@ define([
         });
         this.selected = selected;
 
-        // Rename is only visible when one item is selected.
-        if (selected.length==1) {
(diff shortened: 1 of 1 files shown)


minrk:
@Carreau you marked this as 4.0. It seems like a fix appropriate for 3.1, no?

Carreau:
Hum, Maybe. Seem we tagged the same day, and my action say modified 4.0,3.1, So I guess I tagged 4.0 by mistake, then switched to 3.1 but was not taken into account.... Seem sensible for 3.x, and code is straitforward.

Carreau:
(rechanging to 3.1)

PR #8005, opened 2015-03-09
Get /files/ unmodified from contents

minrk:
request raw files instead of parsing notebooks

avoids reserializing (inconsistently) on download of notebooks.

### DIFF
--- a/IPython/html/files/handlers.py
+++ b/IPython/html/files/handlers.py
@@ -28,12 +28,12 @@ class FilesHandler(IPythonHandler):
         else:
             name = path
         
-        model = cm.get(path)
+        model = cm.get(path, type='file')
         
         if self.get_argument("download", False):
             self.set_header('Content-Disposition','attachment; filename="%s"' % name)
         
-        if model['type'] == 'notebook':
+        if model['type'] == 'notebook' or name.endswith('.ipynb'):
             self.set_header('Content-Type', 'application/json')
         else:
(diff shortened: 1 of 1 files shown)


IPython/html/files/handlers.py
```diff
@@ -28,12 +28,12 @@ def get(self, path):
...
             name = path
         
-        model = cm.get(path)
+        model = cm.get(path, type='file')
         
         if self.get_argument("download", False):
             self.set_header('Content-Disposition','attachment; filename="%s"' % name)
         
-        if model['type'] == 'notebook':
+        if model['type'] == 'notebook' or name.endswith('.ipynb'):
```
takluyver: I think now that we're specifying `type='file'` above, the returned type can't be 'notebook', so we only need to check the extension.

ssanderson: Should we even be setting an `'application/json'` mimetype here if the data is being morally treated as text?

takluyver: The few lines below will attempt to guess the mime-type from the extension. This is essentially just ensuring that `.ipynb` is known to be a JSON file, since that's going to be a common case. From the recipient's point of view, it's not just text.

Though possibly we should consider using the mimetype we picked for notebooks, `application/x-ipynb+json`.

minrk: removed the unreachable `type='notebook'`. I've left the mime-type as `application/json` for now.

PR #8006, opened 2015-03-09
trigger save before copy if there are unsaved changes.

### LINKED ISSUE #8001: Making a copy of an open notebook only copies (auto)saved changes
`File > Make a copy...` merely makes a copy of the notebook file **in its current (auto)saved state**. This is particularly troublesome when autosave is enabled: from the user standpoint, making a copy results in a copy of the notebook *at some random point in time in the last x minutes* (and this will often go unnoticed, looking like it includes the latest changes).

I would definitely expect `Make a copy...` to include all latest change (otherwise I would go in the dashboard and duplicate the notebook).

A possible way to remove any ambiguity would be a `File > Save as...` dialog (although the exact behavior with respect to Save or Save+Checkpoint would remain a bit unclear). To make it closest to `Make a copy...` one could optionally
- make the default file name being the usual `....-CopyN.ipynb`
- include both a `Save` button and a `Save and Open` button

A minor side issue : the `File` menu does not close in the original notebook upon making a copy.

*Edit:* I'd be happy to open a PR for this if there is a clean and cheap way of reusing current code, but I haven't quite found how to do that yet, and I'd rather leave this to someone else if tweaking the save mechanism is required.

minrk:
just like we do before downloading

closes #8001

### DIFF
--- a/IPython/html/static/notebook/js/menubar.js
+++ b/IPython/html/static/notebook/js/menubar.js
@@ -104,6 +104,9 @@ define([
             window.open(utils.url_join_encode(that.base_url, 'tree', parent), IPython._target);
         });
         this.element.find('#copy_notebook').click(function () {
+            if (that.notebook.dirty) {
+                that.notebook.save_notebook({async : false});
+            }
             that.notebook.copy_notebook();
             return false;
         });


takluyver:
LGTM

mathieu1:
That fix is perfectly fine by me, but might be confusing for people who disable autosave and might end up with saves they did not purposefully trigger (just like they would by downloading). Then again possibly no one will ever complain :)

PR #8017, opened 2015-03-11
don't re-raise the same Exception in NoDB

minrk:
On py3, exceptions keep track of their traceback, resulting in memory growth in the controller because NoDB re-raises the same instance over and over again, which was a weird thing to do, anyway.

Fix is to raise a new Exception() each time.

/cc @diverdude, who reported the issue.


Carreau:
+1, just one question, why not `NODATA = NoData()` ?

minrk:
@Carreau because that's exactly what it did before, and what caused the problem described above.

PR #8020, opened 2015-03-11
Document how to install kernel spec files

### LINKED ISSUE #7988: Document how to make kernels available in notebooks
After seeing the new kernel-switching feature in the release notes for v3.0, it took me quite some time to figure out how to make use of it. As far as I can tell the easiest way to make the Python 2 and 3 kernels available through the notebook's menu is to run:

```bash
ipython2 kernelspec install-self --user
```

for both `ipython2` and `ipython3`. Should this be documented somewhere (e.g. the main notebook page in the docs), or is this still considered a bit of an experimental feature without a set public API?

jankatins:
Closes: #7988

### DIFF
--- a/docs/source/notebook/notebook.rst
+++ b/docs/source/notebook/notebook.rst
@@ -417,6 +417,26 @@ on available options, use::
 
     :ref:`notebook_public_server`
 
+Installing new kernels
+----------------------
+
+Running the notebook makes the current python installation available as a kernel. Other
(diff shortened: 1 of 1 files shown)


Carreau:
Why the hell is coverall commenting on status. It should not. Maybe it is due to me enabling it on jupyter repo yesterday.

Carreau:
This actually passes, don't care about coveralls.

minrk:
Thanks

Issue #8012, opened 2015-03-10
SelectMultiple widget does not always work


jdfreder:
@bobbysan reported this in #5327 .

---

Selecting multiple items using the SelectMultiple widget does not work in some cases for me. If the selected options are highlighted using only the mouse (and are contiguous) it works fine. If selecting only apples and bananas (using Shift and Ctrl) in the example below it does not work. It will highlight/select the last item clicked. Please advise. Thanks.

```python
w = widgets.SelectMultiple(
    description="Fruits",
    options=['Apples', 'Oranges', 'Pears', 'Grapes', 'Peaches', 'Bananas'])
display(w)
```

My setup:
ipython-verson: 3.1.0-dev & 3.0.0 (tried both)
platform: windows 7

PR #8027, opened 2015-03-12
Don't re-implement option.click to trigger select.change

bollwyvl:
This empties the click handler used by SelectMultiple's options, which appears to solve #8012.

I say appears because while I have [tests](https://gist.github.com/bollwyvl/5a6c0dab551f20aba3ac) for the ctrl/shift behavior, I can't make them work reliably. The keyboard stuff (which was working before) is unreliable, but at least seems to understand the event+modifier, but the mouse stuff is nothing. Still investigating casper/phantom APIs to get this right, but in the meantime, this is a pretty self-contained change.

### DIFF
--- a/IPython/html/static/widgets/js/widget_selection.js
+++ b/IPython/html/static/widgets/js/widget_selection.js
@@ -557,6 +557,15 @@ define([
             SelectMultipleView.__super__.update.apply(this, arguments);
             this.$listbox.val(this.model.get('selected_labels'));
         },
+        
+        handle_click: function(){
+            /**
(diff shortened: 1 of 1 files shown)


Carreau:
I would have just override in line 517 SelectWidget. I don't see any reason to override in select multiple and not in select. Am I missing smth ? 

Anyway, thanks !

bollwyvl:
The tests use jquery click on option, which doesn't fire change on the
select... can revisit this, perhaps using the "native"r clicks of casper.

PR #8030, opened 2015-03-12
use literal_eval parsing CLI args

### LINKED ISSUE #8025: commandline parser can't handle --something="all"
It seems that parsing `all` will convert it to the `all()` function and an error is shown:

```python
knitpy_aliases.update({
    'to' : 'KnitpyApp.export_format',
    [...]
})

class KnitpyApp(BaseIPythonApplication):
    # '--to' ends up here
    export_format = CaselessStrEnum(VALID_OUTPUT_FORMATS+["all"],
        default_value=DEFAULT_OUTPUT_FORMAT,
        config=True,
        help="""The export format to be used."""
    )
```

and then when I call it:
```
$ knitpy --to="all" example\hello.pymd # or `--to=all`
[...]
[KnitpyApp] CRITICAL | The 'export_format' trait of a KnitpyApp instance must be
 any of [u'pdf', u'docx', u'html', u'all'] or None, but a value of <built-in fun
ction all> <type 'builtin_function_or_method'> was specified.
```

minrk:
avoids 'all', 'str', etc. turning into builtins instead of strings

Thanks @takluyver for pointing me to literal_eval, which I didn't know about

closes #8025

### DIFF
--- a/IPython/config/loader.py
+++ b/IPython/config/loader.py
@@ -11,6 +11,7 @@ import os
 import re
 import sys
 import json
+from ast import literal_eval
 
 from IPython.utils.path import filefind, get_ipython_dir
 from IPython.utils import py3compat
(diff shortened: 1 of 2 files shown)


takluyver:
Do we want this for 3.1? Technically someone could be using the old behaviour as a feature, though I think it's unlikely.

Issue #8035, opened 2015-03-12
custom.js sample use of require() is wrong, should be define()


parente:
Following the commented sample in custom.js bit me: custom.js can be loaded AFTER the app initialized event in main.js is fired. Here's why:

* In 2.x, custom.js was inlined in the notebook.html template. In this form, require is the correct call: it's the top of the module dependency tree.
* In 3.x, custom is declared as a module dependency in main.js and imported using requirejs. In this form, it must be a define() else requirejs does not guarantee that it will load before the callback in main.js fires. (It basically becomes a separate dependency tree with no guarantee on load order with the one rooted in main.js)

Fix: update the custom.js commented samples to use define() not require().

Pointer to problem: https://github.com/ipython/ipython/blob/master/IPython/html/static/custom/custom.js#L21

and more further down in the file.

parente:
Tnx @jtyberg.

PR #7992, opened 2015-03-06
remove false info in custom.js

Carreau:
## 

Should close #7984 I guess, as we don't really have stable js config.

PR #8036, opened 2015-03-12
Fix custom.js samples to define requirejs module rather than require it.

jtyberg:
Using 'define' guarantees that requirejs loads custom.js before other modules that require it.


jtyberg:
Fixes #8035

Carreau:
Thanks, shoudl probably go in with https://github.com/ipython/ipython/pull/7992 at some point !

PR #8041, opened 2015-03-13
Exclude tokenize from deepreload by default

takluyver:
This should fix recent Windows test failures we've seen on Python 3.4


takluyver:
Apparently this didn't fix the issue: https://jenkins.jupyter.org/job/windows-py3.4-prtest/4/console

This problem should only be possible to reach for people using deepreload on Python 3.4.3 precisely. So I'm going to patch it out of the test server and ignore it for now.

PR #8045, opened 2015-03-13
Merge codemirror config with user config.

### LINKED ISSUE #8038: Merge user default cm_config
See #7926

something like :
`cm_config = $.extend((defaults.cm_config || {}), userconfig.cm_config);`

Optimistically marking 3.1

Carreau:
Simplify customisation of codemirror without duplicating the whole
codecell configuration.

Closes #8038

### DIFF
--- a/IPython/html/static/notebook/js/cell.js
+++ b/IPython/html/static/notebook/js/cell.js
@@ -56,7 +56,6 @@ define([
         // superclass default overwrite our default
         
         this.placeholder = config.placeholder || '';
-        this.read_only = config.cm_config.readOnly;
         this.selected = false;
         this.rendered = false;
         this.mode = 'command';
@@ -76,7 +75,12 @@ define([
 
         // load this from metadata later ?
         this.user_highlight = 'auto';
-        this.cm_config = config.cm_config;
+
+        var class_conf_value = {};
+        if(this.class_config){
+            class_conf_value = this.class_config.get_sync('cm_config');
+        }
+        this.cm_config = utils.mergeopt({}, config.cm_config, class_conf_value);
         this.cell_id = utils.uuid();
         this._options = config;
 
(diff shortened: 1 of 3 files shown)


IPython/html/static/notebook/js/cell.js
```diff
@@ -76,7 +75,12 @@ define([
 
         // load this from metadata later ?
         this.user_highlight = 'auto';
-        this.cm_config = config.cm_config;
+
+        var class_conf_value = {};
+        if(this.class_config){
+            class_conf_value = this.class_config.get_sync('cm_config');
+        }
+        this.cm_config = utils.mergeopt({}, config.cm_config, class_conf_value);
```
takluyver: The naming is getting really confusing here: `class_conf_value` is the value the user has set, if it has arrived, not the value set in the class definition. But if the user config hasn't been loaded yet, both will be the value set on the class, so you'll be merging a dict with itself.

I recommend that we redo this, moving cm_config out of options_default to a separate class-level attribute `default_cm_config`. Then this would look like:

```javascript
utils.mergeopt({}, this.default_cm_config, this.class_config.get_sync('cm_config'))
```

Carreau:
> I recommend that we redo this, moving cm_config out of options_default to a separate class-level attribute default_cm_config. Then this would look like:
> 
> utils.mergeopt({}, this.default_cm_config, this.class_config.get_sync('cm_config'))


Agreed for 4.0, but this would change  "API" in 3.x

IPython/html/static/notebook/js/cell.js
```diff
@@ -29,7 +29,7 @@ define([
         // Reverse `min-height: 18px` scrollbar hack on OS X
         // which causes a dead area, making it impossible to click on the last line
         // when there is horizontal scrolling to do and the "show scrollbar only when scrolling" behavior
-        // is enabled.
+        // is enabled.c
```
takluyver: Stray c

Carreau: Done.

bollwyvl:
great, thanks!

PR #8051, opened 2015-03-14
DOC: Update %matplotlib for bbox_inches

rnelsonchem:
Added a couple of extra lines to the matplotlib magic doc string with instructions on how to shut off the default bbox_inches='tight' behavior.

A little background. Setting bbox_inches to 'tight' for inline figures produces figures that are not identical to ones produced using `savefig` in the same code block. This is problematic when, for example, you are trying to prepare a publication ready figure in the notebook. (You may want to display the figure in the notebook to share the code with collaborators but at the same time produce a very high resolution version for the manuscript.) Adding a couple of lines to the docs to address this problem could help here.

### DIFF
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -57,9 +57,17 @@ class PylabMagics(Magics):
             
             In [2]: set_matplotlib_formats('pdf', 'svg')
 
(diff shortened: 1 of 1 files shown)


Tillsten:
Could you also mention the changes to rcparams? https://github.com/matplotlib/matplotlib/issues/4102#issuecomment-74774717

PR #8060, opened 2015-03-15
Removing extra `client.` in example code

yuvallanger:
I am not 100% sure this is correct in this context.


minrk:
Probably the result of a less-than-careful find-replace. Thanks!

PR #8062, opened 2015-03-15
another stab at confining images to page width

### LINKED ISSUE #7701: Image overflow in IPython Notebook 3 (beta)

minrk:
- double-click toggles max-width confinement for large image output (not markdown)
- uses max-width on both markdown and output images
- `height: auto` fixes aspect ratio problems caused by using max-width alone (#7721)
- img.unconfined disables max-width confinement
- `overflow-x: auto` on markdown (matches output in #7718)

closes #7701

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -637,7 +637,9 @@ class Image(DisplayObject):
     _FMT_PNG = u'png'
     _ACCEPTABLE_EMBEDDINGS = [_FMT_JPEG, _FMT_PNG]
 
-    def __init__(self, data=None, url=None, filename=None, format=u'png', embed=None, width=None, height=None, retina=False):
+    def __init__(self, data=None, url=None, filename=None, format=u'png',
+                 embed=None, width=None, height=None, retina=False,
+                 unconfined=False, metadata=None):
         """Create a PNG/JPEG image object given raw data.
 
         When this object is returned by an input cell or passed to the
@@ -678,6 +680,10 @@ class Image(DisplayObject):
             from image data.
             For non-embedded images, you can just set the desired display width
             and height directly.
+        unconfined: bool
+            Set unconfined=True to disable max-width confinement of the image.
+        metadata: dict
+            Specify extra metadata to attach to the image.
 
         Examples
         --------
@@ -728,6 +734,8 @@ class Image(DisplayObject):
         self.width = width
         self.height = height
         self.retina = retina
+        self.unconfined = unconfined
+        self.metadata = metadata
         super(Image, self).__init__(data=data, url=url, filename=filename)
         
         if retina:
@@ -756,12 +764,19 @@ class Image(DisplayObject):
 
     def _repr_html_(self):
         if not self.embed:
-            width = height = ''
+            width = height = klass = ''
             if self.width:
                 width = ' width="%d"' % self.width
             if self.height:
                 height = ' height="%d"' % self.height
-            return u'<img src="%s"%s%s/>' % (self.url, width, height)
+            if self.unconfined:
+                klass = ' class="unconfined"'
+            return u'<img src="{url}"{width}{height}{klass}/>'.format(
+                url=self.url,
+                width=width,
+                height=height,
+                klass=klass,
+            )
 
     def _data_and_metadata(self):
         """shortcut for returning metadata with shape information, if defined"""
@@ -770,6 +785,10 @@ class Image(DisplayObject):
             md['width'] = self.width
         if self.height:
             md['height'] = self.height
+        if self.unconfined:
+            md['unconfined'] = self.unconfined
+        if self.metadata:
+            md.update(self.metadata)
         if md:
             return self.data, md
         else:
--- a/IPython/core/tests/test_display.py
+++ b/IPython/core/tests/test_display.py
@@ -22,6 +22,8 @@ def test_image_size():
     nt.assert_equal(u'<img src="%s" width="200"/>' % (thisurl), img._repr_html_())
     img = display.Image(url=thisurl)
     nt.assert_equal(u'<img src="%s"/>' % (thisurl), img._repr_html_())
+    img = display.Image(url=thisurl, unconfined=True)
+    nt.assert_equal(u'<img src="%s" class="unconfined"/>' % (thisurl), img._repr_html_())
 
 def test_retina_png():
     here = os.path.dirname(__file__)
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -668,6 +668,23 @@ define([
         return toinsert;
     };
 
+    dblclick_to_reset_size = function (img) {
+        /**
+         * Double-click on an image toggles confinement to notebook width
+         *
+         * img: jQuery element
+         */
+
+        img.dblclick(function () {
+            // dblclick toggles *raw* size, disabling max-width confinement.
+            if (img.hasClass('unconfined')) {
+                img.removeClass('unconfined');
+            } else {
+                img.addClass('unconfined');
+            }
+        });
+    };
+    
     var set_width_height = function (img, md, mime) {
         /**
          * set width and height of an img element from metadata
@@ -676,6 +693,9 @@ define([
         if (height !== undefined) img.attr('height', height);
         var width = _get_metadata_key(md, 'width', mime);
         if (width !== undefined) img.attr('width', width);
+        if (_get_metadata_key(md, 'unconfined', mime)) {
+            img.addClass('unconfined');
+        }
     };
     
     var append_png = function (png, md, element, handle_inserted) {
@@ -689,6 +709,7 @@ define([
         }
         img[0].src = 'data:image/png;base64,'+ png;
         set_width_height(img, md, 'image/png');
+        dblclick_to_reset_size(img);
         toinsert.append(img);
         element.append(toinsert);
         return toinsert;
@@ -706,6 +727,7 @@ define([
         }
         img[0].src = 'data:image/jpeg;base64,'+ jpeg;
         set_width_height(img, md, 'image/jpeg');
+        dblclick_to_reset_size(img);
         toinsert.append(img);
         element.append(toinsert);
         return toinsert;
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -588,6 +588,7 @@ define([
         var toinsert = this.create_output_subarea(md, "output_html rendered_html", type);
         this.keyboard_manager.register_events(toinsert);
         toinsert.append(html);
+        dblclick_to_reset_size(toinsert.find('img'));
         element.append(toinsert);
         return toinsert;
     };
@@ -603,6 +604,7 @@ define([
             html = mathjaxutils.replace_math(html, math);
             toinsert.append(html);
         });
+        dblclick_to_reset_size(toinsert.find('img'));
         element.append(toinsert);
         return toinsert;
     };
@@ -668,7 +670,7 @@ define([
         return toinsert;
     };
 
-    dblclick_to_reset_size = function (img) {
+    function dblclick_to_reset_size (img) {
         /**
          * Double-click on an image toggles confinement to notebook width
          *
--- a/IPython/html/static/notebook/less/outputarea.less
+++ b/IPython/html/static/notebook/less/outputarea.less
@@ -66,6 +66,14 @@ div.output_area {
             margin-right: 0;
         }
     }
+    
+    img, svg {
+        max-width: 100%;
+        height: auto;
+        &.unconfined {
+            max-width: none;
+        }
+    }
 }
 
 /* This is needed to protect the pre formating from global settings such
--- a/IPython/html/static/notebook/less/outputarea.less
+++ b/IPython/html/static/notebook/less/outputarea.less
@@ -107,7 +107,7 @@ div.output_subarea {
     padding: @code_padding;
     .box-flex1();
     // appears to be needed for max-width in children to mean anything on Firefox:
-    max-width: 100%;
+    max-width: calc(~"100% - 14ex");
 }
 
 /* The rest of the output_* classes are for special styling of the different
--- a/IPython/html/static/notebook/less/outputarea.less
+++ b/IPython/html/static/notebook/less/outputarea.less
@@ -106,6 +106,8 @@ div.output_subarea {
     overflow-x: auto;
     padding: @code_padding;
     .box-flex1();
+    // appears to be needed for max-width in children to mean anything on Firefox:
+    max-width: 100%;
 }
 
 /* The rest of the output_* classes are for special styling of the different
--- a/IPython/html/static/notebook/less/renderedhtml.less
+++ b/IPython/html/static/notebook/less/renderedhtml.less
@@ -82,5 +82,12 @@
         margin-right: auto;
     }
     * + img {margin-top: 1em;}
-
+    
+    img, svg {
+        max-width: 100%;
+        height: auto;
+        &.unconfined {
+            max-width: none;
+        }
+    }
 }
--- a/IPython/html/static/notebook/less/textcell.less
+++ b/IPython/html/static/notebook/less/textcell.less
@@ -35,6 +35,10 @@ h1,h2,h3,h4,h5,h6 {
     display: none;
 }
 
+.text_cell.rendered .rendered_html {
+    overflow-x: auto;
+}
+
 .text_cell.unrendered .text_cell_render {
     display:none;
 }
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -890,6 +890,7 @@ div.output_subarea {
   box-flex: 1;
   /* Modern browsers */
   flex: 1;
+  max-width: 100%;
 }
 /* The rest of the output_* classes are for special styling of the different
    output types */
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -1184,6 +1184,9 @@ h6:hover .anchor-link {
 .text_cell.rendered .input_area {
   display: none;
 }
+.text_cell.rendered .rendered_html {
+  overflow-x: auto;
+}
 .text_cell.unrendered .text_cell_render {
   display: none;
 }
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -890,7 +890,7 @@ div.output_subarea {
   box-flex: 1;
   /* Modern browsers */
   flex: 1;
-  max-width: 100%;
+  max-width: calc(100% - 14ex);
 }
 /* The rest of the output_* classes are for special styling of the different
    output types */
--- a/IPython/html/static/style/ipython.min.css
+++ b/IPython/html/static/style/ipython.min.css
@@ -825,6 +825,15 @@ div.output_area .rendered_html img {
   margin-left: 0;
   margin-right: 0;
 }
+div.output_area img,
+div.output_area svg {
+  max-width: 100%;
+  height: auto;
+}
+div.output_area img.unconfined,
+div.output_area svg.unconfined {
+  max-width: none;
+}
 /* This is needed to protect the pre formating from global settings such
    as that of bootstrap */
 .output {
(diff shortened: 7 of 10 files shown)


minrk:
ping @jdfreder, @ellisonbg, @erccarls, @rsmith31415 for testing

Images are confined to width by default; double-click restores the 'native' size of the image. The `unconfined` metadata allows explicit requests for images to ignore confinement.

rsmith31415:
Are you sure it is working in Firefox? In Chrome, works fine using `from IPython.display import Image`.

Firefox:

screenshot from 2015-03-15 16 50 06

Chrome:

screenshot from 2015-03-15 16 52 51

minrk:
@rsmith31415 thanks for testing, I'm never sure if anything works in Firefox. That looks wrong in all kinds of ways.

minrk:
Okay, pushed a small additional max-width, and now it seems to work fine on both FF and Chrome (here, at least). I wasn't able to reproduce the weird overlay you saw, though.

rsmith31415:
Correct. The second issue has been fixed. Unfortunately, the image (1200x1600, by the way) still overflows in Firefox.

minrk:
Interesting! I wasn't able to reproduce the second issue, but the first issue was fixed by the recent change for me. I'm at a loss if it still doesn't work on other machines.

rsmith31415:
You mean the opposite, right? You can't reproduce the first issue (the image overflowing the right side of the notebook) but you fixed the second issue (image overlapping `div.out_prompt_overlay`).

By the way, I'm using Firefox 36.0.1 on Ubuntu 14.10.

minrk:
@rsmith31415 no. I can reproduce the image overflowing to the right, and the latest commit fixed it. I have not seen the image overlapping the prompt overlay.

rsmith31415:
I think the was a happy coincidence. After you fixed the image overflowing to the right, the overlapping was fixed. The problem is that the overflowing still remains in Firefox. However, there is no overflowing if I use markdown instead of `from IPython.display import Image`.

takluyver:
The image appears to be constrained to 100% width in Firefox with the latest commit, but it still overflows to the right, because it's offset by the width of the `Out [n]:` prompt:

screenshot from 2015-03-18 14 06 00

Maybe we can use `calc()` to get the correct size?

takluyver:
Double click to switch between width constrained and 1:1 seems to work, and images in markdown cells are correctly constrained to the notebook width.

rsmith31415:
I noticed that there are different HTML elements (with different classes) being injected depending on the source (e.g. `from IPython.display import HTML` vs `from IPython.display import Image`). Wouldn't it be possible to use a single markup for all of them? I'm saying that because iframes are still overflowing in Chrome and Firefox:

screenshot from 2015-03-18 15 40 01

takluyver:
In that case, you're explicitly setting width/height in a tag in the HTML. I'm not sure there's much we can/should do about that.

minrk:
I applied the max-width behavior to img, svg in particular, but perhaps it should be applied to the outer element, or all children of output. It is not clear to me whether raw HTML output (e.g. iframes) be restricted, or left alone. Perhaps a combination of this and #7718 should be used.

For the moment, I'm inclined to only confine images. The reason being that we already have the double-click to expand behavior on images (I only changed slightly what expand means in this PR), but not on the output area as a whole. That means that right now, there's an easy way to expand larger images to their native size, but not an easy way to expand other HTML elements. Perhaps we should go that way ultimately, but I like to take these things one step at a time.

@takluyver thanks for pointing that out, yes calc should fix it in the latest commit. I thought it was behaving correctly because of my custom css changes the right margin. I should really remember to test CSS PRs without custom CSS.

minrk:
> In that case, you're explicitly setting width/height in a tag in the HTML

max-width takes effect when `width > max-width`. I have it working on every output I produce if I use `*` instead of `img, svg` in this PR. So we definitely *can* do something about it. "Should" is less clear.

rsmith31415:
@takluyver I don't think it should be different to set widths and heights in an iframe than importing an image of arbitrary size. The container should constraint the output's dimensions under any circumstance. Actually, if you change the iframe to have `width:100%;` that fixes it. However, I think it is better to clean that part of the markup to make it more consistent and allow for a more definitive solution. For example, even relatively simple HTML will overflow (in the example below, a table):

screenshot from 2015-03-18 17 29 21


@minrk I think the double click (and single click) behavior is working correctly in other HTML elements (actually, in the case of a table, the double click works but single click has no effect.) The main problem are not the parent elements. The problem is that the HTML tags that are injected need to have proper styling to fit their parents.

minrk:
@rsmith31415 I assume you are referring to the scrolling behavior triggered by clicking on the prompt area, because I was referring to the double-click event we only bind to images.

If we combine this with #7718, then all output will be confined to the area. Arbitrary HTML will overflow in with a scrollbar, and images will shrink to fit. It would also mean that expanding images will not expand them onto the entire page, but rather into a scrolling view that still fits within the width of the notebook. It would make it impossible for images to extend beyond the bounds of the notebook area.

rsmith31415:
@minrk You're right. I was referring to the prompt because in the last commit, I'm seeing again the image overlapping over the prompt, but this time I find it very hard to reproduce consistently.

There is another issue. If you click twice on the prompt (without double clicking), initially the image remains constrained. But if you double click on the image to let it expand and then click twice again on the prompt, the behavior of clicking on the prompt changes between constraining the image and expanding the image.

Last issue: After reducing the browser, all elements are re-positioned accordingly except the image:

screenshot from 2015-03-18 18 37 39

takluyver:
Images are now the correct size for me.

@rsmith31415 , I'm not sure what you mean about the behaviour of clicking the prompt area after double clicking an image. In my brief tests, it's working as expected.

rsmith31415:
@takluyver Yes, sorry. I forgot to describe one crucial step to reproduce this issue. Furthermore, now I see that the issue is not really in the prompt but in resizing the browser. Steps:

1. Start your browser maximized and add a large image.
2. Resize your browser:

screenshot from 2015-03-19 12 14 20

3 . Double click on the image to expand it.
4 . Maximize your browser (now the image is overflowing):

screenshot from 2015-03-19 12 16 42

5 . Click twice on the prompt. The first click works as expected. The second click returns to the image overflowing.

takluyver:
I seem to be seeing the same things as you, but I'm not sure why you think it's wrong.

There are two separate toggles: image constrained/expanded and output open/scrollboxed. When you switch the output to scrollboxed and back, that doesn't change the image size, so if the image was overflowing before, switching the output area state twice will bring it back to overflowing.

I do see another problem, though. When resizing the window with the image constrained to max width, its size doesn't update as the notebook width updates. I can make it resize in width if I set `max-width: 100%` on the div containing the img (`class="ui-wrapper ui-resizable-autohide"`), but then the aspect ratio isn't preserved.

rsmith31415:
Yes, I see what you mean. I think I find confusing the feature of expanding the image outside the notebook. There is also the case in which the output is scrollboxed and you double-click to expand within the scrollbox. I think that is very reasonable. Although the image doesn't seem to preserve its dimensions, it seems like a middle ground between the expanded image and the scrollboxed image. By the way, what is the purpose of allowing the image to increase to its original size?

The problem you described is the same I was mentioning before, right? The image is not responsive to resizing the window.

rsmith31415:
I finally understood the thing that is causing problems. It is the `.resizable()` [method](http://jqueryui.com/resizable/) from jquery-ui. Particularly, the following lines:


			//Create a wrapper element and set the wrapper to the new current internal element
			this.element.wrap(
				$("<div class='ui-wrapper' style='overflow: hidden;'></div>").css({
					position: this.element.css("position"),
					width: this.element.outerWidth(),
					height: this.element.outerHeight(),
					top: this.element.css("top"),
					left: this.element.css("left")
				})
			);

			//Overwrite the original this.element
			this.element = this.element.parent().data(
				"ui-resizable", this.element.data("ui-resizable")
			);

This is setting a width and height in pixels to the element `ui-wrapper`:

screenshot from 2015-03-19 16 28 38

Therefore, when we resize the window, the old width and height remain. This is also adding a class `ui-resizable` to the image element with the same width and height than `ui-wrapper`. By the way, this is the resize handler that  @jdfreder was referring to in the first place.

I was wondering whether we really need that feature. I didn't even know it existed.

ellisonbg:
I am fine without images being resizable in the UI.

On Thu, Mar 19, 2015 at 6:59 PM, rsmith31415 <notifications@github.com>
wrote:

> I finally understood the thing that is causing problems. It is the
> .resizable() method <http://jqueryui.com/resizable/> from jquery-ui.
> Particularly, the following lines:
>
>         //Create a wrapper element and set the wrapper to the new current internal element
>         this.element.wrap(
>             $("<div class='ui-wrapper' style='overflow: hidden;'></div>").css({
>                 position: this.element.css("position"),
>                 width: this.element.outerWidth(),
>                 height: this.element.outerHeight(),
>                 top: this.element.css("top"),
>                 left: this.element.css("left")
>             })
>         );
>
>         //Overwrite the original this.element
>         this.element = this.element.parent().data(
>             "ui-resizable", this.element.data("ui-resizable")
>         );
>
> This is setting a width and height in pixels to the element ui-wrapper:
>
> [image: screenshot from 2015-03-19 16 28 38]
> <https://cloud.githubusercontent.com/assets/599274/6742580/8e487e9a-ce56-11e4-84e9-0cad3c9a9d85.png>
>
> Therefore, when we resize the window, the old width and height remain.
> This is also adding a class ui-resizable to the image element with the
> same width and height than ui-wrapper. By the way, this is the resize
> handler that @jdfreder <https://github.com/jdfreder> was referring to in
> the first place.
>
> I was wondering whether we really need that feature. I didn't even know it
> existed.
>
> —
> <https://github.com/ipython/ipython/pull/8062#issuecomment-83794149>.
>



-- 
Brian E. Granger

rsmith31415:
If removing `.resizable()` is permissible, I think this issue can be solved very easily. I [created](https://github.com/ipython/ipython/compare/master...rsmith31415:css-fix-width-images) a branch with the changes. I tested quickly and works okay:

output4

ellisonbg:
Thanks for investigating this. Let's see what others think and spend time testing it out.

minrk:
I'm fine with removing resize

ellisonbg:
@rsmith31415 do you want to submit a PR to this PR with your proposed changes to get rid of the resizing?

rsmith31415:
@ellisonbg I submitted only the removal of the resizer. After that, I think it only takes a few CSS styles to get the main problem fixed. @minkr's PR also includes the feature of expanding the image to its original size but that can be added easily later.

takluyver:
I don't personally object, but we should be clear that constraining images without being able to easily resize them was what we tried before (#7721), and we got complaints about it from people who wanted to make and look at plots larger than the notebook width, which was part of why we reverted it (#7818). (the other part was that it wasn't consistent across browsers, which I think we've fixed here)

rsmith31415:
@takluyver As you pointed out in #7721, there are two cases: 

1. User wants to add an image larger than the notebook
2. User inadvertently generates image larger than the notebook.

The second case is far more common than the first. The more common case should take precedence. However, an alternative would be to allow scrollbars if the image is wider than the container, but as you stated, it makes an ugly UI. 

If the user wants to see the raw image, it only takes to right-click on the image and select "Copy image URL" and paste it in another tab.

Notice that, in any situation, we don't need to stretch the image. That was one of the concerns in #7721.

minrk:
I can keep the double-click to escape max-width here without the draggable jquery-ui resizable div, which seems to be the source of several of these problems. Then the default is confined to the area, but it can be optionally unconfined after the fact. In that case, it's a simple CSS class toggle, rather than a complex statically sized container.

rsmith31415:
@minrk That's correct. However, if the idea is to let the user see the image in its original size, I think it is better to copy the URL and paste it in another tab because that includes zoom, save as, print and other browser features that the user might want.

minrk:
ok, then we can do away with the entire idea of allowing users to display images larger than the notebook area. That certainly should make things simpler.

minrk:
Okay, this is now simpler than before, and probably ready to go:

- large images, iframes are confined with max-width
- unconfined class overrides this
- double-click toggles the 'unconfined' class on images
- Thanks to #7718, when images are expanded in output, they expand into a scrolled output_subarea, rather than onto the wider page

minrk:
And a test notebook is included, illustrating a few examples of large output.

takluyver:
Is it intentional that unconstrained images in markdown get cut off with no way to scroll them? I don't have a problem with that, I just want to check that I shouldn't be seeing a scrollbar there.

Other than that, everything in the test nb looks good.

rsmith31415:
Looks great. There are a few minor issues, but in general, I think this is really good.

The double click feature is not working in markdown images because after double clicking, the cell changes to show the markdown code, instead of showing the full image.

I believe the image parameters for `width` and `height` in `IPython.display.Image` are not working. They only work to reduce the size of the image. If I want to extend the image to generate scrollbars, that's not possible anymore. However, `iframes` can be extended without limitations. I'm not complaining about the first part, but it is not very consistent.

minrk:
@rsmith31415 the double-click is not applied to markdown images at all. It could be, though. I'm inclined to leave it out for now, though.

We haven't applied the same overflow-x scroll behavior to markdown cells that we have on output-subarea. Should we? It does seem more consistent. I've updated the PR to reflect this.

> I believe the image parameters for width and height in IPython.display.Image are not working

They are working, in that they set the width and height attributes of the img tag. If the image is larger than the display area, max-width takes precedence and shrinks the image. On double-click max-width is unset, and the original width, height determine the display size. The retina image example in the test notebook demonstrates this.

> If I want to extend the image to generate scrollbars, that's not possible anymore.

By default, that's correct, and I think this is the intended behavior. If the `unconfined` metadata is set, then the max-width behavior is disabled, and there will be scrollbars for wide images.

> However, iframes can be extended without limitations

I'm not sure about adding iframes to the max-width treatment. I'm relatively confident that tables, divs, and other generic HTML elements *shouldn't* be added to the shrink-to-fit list. One reason to not add iframes is the `max-width: 100%; height: auto` doesn't preserve the aspect ratio correctly the same way it does on images so extra fixes would be needed to make iframes work. For that reason alone, I'm inclined to leave them out, at least initially.

rsmith31415:
>the double-click is not applied to markdown images at all. It could be, though I'm inclined to leave it out for now.

I think you would expect images using markdown and  `IPython.display.Image` to have the same behavior. However, I think that's not good for markdown cells. How could you add the double click feature without interfering with the double clicking to see the markdown code? 

>They are working, in that they set the width and height attributes of the img tag. If the image is larger than the display area, max-width takes precedence and shrinks the image.

Well, it is working under the hood but not working from the user's perspective. I'm not sure what to do about it.

>One reason to not add iframes is the max-width: 100%; height: auto doesn't preserve the aspect ratio correctly the same way it does on images so extra fixes would be needed to make iframes work. For that reason alone, I'm inclined to leave them out, at least initially.

Actually, you describe there another aspect. Right now, `height` is set to `auto`. That has the unpleasant effect of extending the cell vertically without limits (well, the limit is the image's height).

I think I found something very strange. After double clicking to expand the image, the menu to copy the image URL is disabled. It is also disabled when the image is constrained to 100%. This doesn't happen in a markdown cell.

By the way, I think I asked this before but now that I'm testing this feature, I'm wondering again what is the purpose of allowing to scroll the raw image? I like the part when the image is constrained with `max-width: 100%;` but once I double click, I'm not entirely sure that view is useful.

minrk:
> How could you add the double click feature without interfering with the double clicking to see the markdown code?

I can prevent the event from bubbling up. I just need to move the functionality to a common place shared by text cell rendering and output areas.

> Well, it is working under the hood but not working from the user's perspective.

If the user's perspective is that their width, height should always be respected, regardless of whether it fits in the output area, that's exactly the problematic behavior of 3.0 that we are trying to fix here. Note that the vast majority of explicit width and height are *not* set explicitly by the user, but under the hood by a library.

> That has the unpleasant effect of extending the cell vertically without limits 

I don't understand what you are describing here. It sets the height of the image according to max-width in order to preserve the aspect ratio of the original width, height. Is there any other behavior that's sensible?

minrk:
Okay, I see only one decision left to make here, and it's what to do with the double-click behavior:

1. (current) affects output images, not markdown
2. apply to both markdown, output
3. remove entirely

I don't have a strong preference, but it makes sense that markdown and output should behave the same.

takluyver:
Output also has the scroll option. I think it's OK to have more features for controlling how output is viewed than for markdown images.

rsmith31415:
>If the user's perspective is that their width, height should always be respected, regardless of whether it fits in the output area, that's exactly the problematic behavior of 3.0 that we are trying to fix here. Note that the vast majority of explicit width and height are not set explicitly by the user, but under the hood by a library.


Yes, you're right. I guess that's not a big problem. But it would be better if we could come up with a set of features that are consistent. If it is going to create scrollbars, it should always create scrollbars. If it is going to be constrained to `max-width`, then that should always be the case. Actually, that's not possible because, as you said, some elements (such as dataframes) shouldn't be subject to `max-width`.

>I don't understand what you are describing here. It sets the height of the image according to max-width in order to preserve the aspect ratio of the original width, height. Is there any other behavior that's sensible?

Let me show you:

output

Notice that the image's height is large enough to require an extremely large cell in the vertical direction. In Chrome, it creates a useless vertical scrollbar:

screenshot from 2015-03-27 18 22 07

In Firefox, there is not vertical scrollbar.

If the double click feature is useful, obviously I would like to keep it. However, I can't think of a reason to maintain it right now. We would be sacrificing consistency and simpler code, so in my mind, it needs to be really useful to justify its inclusion.

minrk:
Since the double-click behavior is now reduced to a trivial CSS class toggle (`.unconfined`), I'm currently somewhat inclined to remove it from this PR, and save it for a discrete proposal, once someone decides it is useful.

ellisonbg:
Hmm, I am trying to test this but am unable to read the example notebook:

screen shot 2015-03-29 at 1 47 49 pm

ellisonbg:
More clues, can't create new notebook (reinstalled ipython, submodule update, cache cleared):

screen shot 2015-03-29 at 1 52 39 pm

minrk:
@ellisonbg that was fixed in master. I just rebased here, so should be working now.

ellisonbg:
OK I will give it a shot again...

ellisonbg:
Hmm, still seeing the same issue :(

ellisonbg:
I am seeing these same problem in master as well as other PRs made against master. I was *expecting* master to be broken, but was thinking about PRs would be made against the 3.x branch when master was broken. Or was master supposed to be working?

minrk:
@ellisonbg what ref do you have? This doesn't affect f78345d8e8fed6ddc7faa1609b52adc638122845.

What do you see in `ls ~/github/ipython/ipython/IPython/nbformat/v4`?

minrk:
oh, wait. It's probably stale files. There shouldn't be anything in IPython/nbformat/ at all. Try a `git clean` to remove any untracked files in the repo.

Master's definitely still working, and all tests are continuing to pass.

ellisonbg:
Ahh, got it! It is working now...ok back to this PR...

* I really like all of the behaviors in the example notebook.
* I am fine waiting until someone asks for the unconfined dblclick behavior.

One question though: what is the default setting of images returned by the matplotlib inline backend? My initial thought is that they should have a max-width so they don't escape.

ellisonbg:
I am also fine leaving the unconfined class there and dblclick there. If we don't include the dblclick now, would you keep the unconfined class and kwarg?

minrk:
> what is the default setting of images returned by the matplotlib inline backend?

Like all images, the default behavior is to be confined by max-width.

>  If we don't include the dblclick now, would you keep the unconfined class and kwarg?

Yes, I think being able to request the behavior from the display-side is actually more important than the double-click behavior.

ellisonbg:
OK, I am fine with merging this with or without the dblclick event - everything else looks fine.

takluyver:
Sorry, I should have mentioned somewhere that I diagnosed the missing schema issue in some tests, and it came down to doing `git clean -fxd` to remove old files.

ellisonbg:
Yep, that did it for me...thanks!

On Sun, Mar 29, 2015 at 3:54 PM, Thomas Kluyver <notifications@github.com>
wrote:

> Sorry, I should have mentioned somewhere that I diagnosed the missing
> schema issue in some tests, and it came down to doing git clean -fxd to
> remove old files.
>
> —
> <https://github.com/ipython/ipython/pull/8062#issuecomment-87492430>.
>



-- 
Brian E. Granger

Carreau:
Ok, great let's merge, try and complain :-)

ellisonbg:
Yeah, great work everyone - that one was surprisingly painful...

On Sun, Mar 29, 2015 at 4:33 PM, Matthias Bussonnier <

> Merged #8062 <https://github.com/ipython/ipython/pull/8062>.
>
> —
> <https://github.com/ipython/ipython/pull/8062#event-268050629>.
>



-- 
Brian E. Granger

PR #8065, opened 2015-03-16
fix default options codemirror in codecell.

Carreau:
recent bugfix removed default configuration of codecell.

@bollwyvl does that fix you issue ?


takluyver:
I think this is redundant if #8045 gets merged (which I think it should)

bollwyvl:
yes, this fixes the issue, but so does #8045!

Carreau:
#8045 was merged.

PR #8066, opened 2015-03-17
Let mistune match inline-block-inline math

bollwyvl:
Adds fix for jupyter/nbviewer#420, apparently caused by the pattern of:

``` markdown
some paragraph $some inline math$
$$ some block math $$
some other paragraph $some more inline math$
```

### DIFF
--- a/IPython/nbconvert/filters/markdown_mistune.py
+++ b/IPython/nbconvert/filters/markdown_mistune.py
@@ -23,7 +23,7 @@ from IPython.utils.decorators import undoc
 
 @undoc
 class MathBlockGrammar(mistune.BlockGrammar):
-    block_math = re.compile("^\$\$([^\$]*?)\$\$", re.DOTALL)
+    block_math = re.compile(r"^\$\$(.*?)\$\$", re.DOTALL)
     latex_environment = re.compile(r"^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\1\}",
                                                 re.DOTALL)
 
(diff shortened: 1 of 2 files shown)


IPython/nbconvert/filters/markdown_mistune.py
```diff
@@ -23,7 +23,7 @@
 
 @undoc
 class MathBlockGrammar(mistune.BlockGrammar):
-    block_math = re.compile("^\$\$(.*?)\$\$", re.DOTALL)
+    block_math = re.compile("^\$\$([^\$]*?)\$\$", re.DOTALL)
```
takluyver: I don't understand how these regex changes can make a difference - given that we were using non-greedy patterns before, shouldn't this match the same thing? I trust you that it does fix it, I'd just like to understand better.

bollwyvl: I am as stumped as you. I tried a few things, and this worked. I agree that it seems dark magic, but I don't know how many layers of regexen, parsers, lexers and whatnot are at war inside mistune, mathjax, jinja. There aren't a lot of places to check the internal state of mistune, in specific: very declarative.

Going back to good ol' print statements, it looks like blocks are hitting very infrequently, and a some markdown around `$` is getting ignored. I'll update the tests to fail accordingly!

bollwyvl:
Ah: the example above is not a block according to mistune, presumably because of the whitespace.

```
some text $math$ here

$$
big math
$$

more $math$ here
```
would be.

So perhaps inline math needs to be able to capture `$$` as well...

bollwyvl:
Unless there is a better way to trigger "blockness" this seems like about the best we can do...

bollwyvl:
Is it worth coding up something for casperjs (for example, with [resemble.js](https://github.com/Huddle/Resemble.js#example) that will [capture](http://casperjs.readthedocs.org/en/latest/modules/casper.html#capture) and actually visually compare the running state of the phantom view of a live notebook with the view of a nbconverted page? seems like an expensive nuclear integration testing option, but would begin to address this problem long-term.

takluyver:
I wouldn't spend significant time writing new tests with casper/phantom. We're not very happy with it, and we're considering moving to something else like Selenium.

Also, I think screenshot comparison is likely to be more annoying than useful.

Carreau:
This fixed nbviewer, so +1 to push a fix and see if more complaints.

takluyver:
+1 - I now understand this, though having block math both at the block level and the inline level seems inelegant.

PR #8067, opened 2015-03-17
Keep current input cell at the top or in the middle of the screen

### LINKED ISSUE #8061: Keep current input cell at the top or in the middle of the screen
When you write a new notebook (as opposed to editing an existing one), the cell that you are editing is often the last cell of the notebook.  In the current UI, editing the last cell is inconvenient for several reasons:

1. It stays at the bottom of the screen.   There is no way to scroll it up the window leaving white space below.
1. Once you execute the code, it jumps up to create space for output and keeps jumping as you update your code based on what you see in the output.

I suggest the following to improve the situation:

1. Allow main window scrolling as long as the last input cell is fully visible.
1. Do not re-scroll on `Ctrl-Enter` even if the output shrinks. 
1. Provide keyboard shortcuts to bring the current cell to the middle, top or bottom of the window.

The second proposal is similar to Emacs' `recenter-top-bottom` function that is bound to `Ctrl-L` by default.   `Ctrl-L`  seems to be available in the notebook UI.

abalkin:
Fixes #8061

### DIFF
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -372,14 +372,24 @@ define(function(require){
                 return env.notebook.scroll_manager.scroll(-1);
             },
         },
-        'recenter-top-bottom': {
-            help: "Move the current cell to the center, top or bottom",
+        'recenter': {
+            help: "Move the current cell to the center",
             handler: function (env, event) {
                 if(event){
                     event.preventDefault();
                 }
                 var cell = env.notebook.get_selected_index();
-                return env.notebook.scroll_to_cell(cell);
+                return env.notebook.scroll_middle_to_cell(cell, 0);
+            }
+        },
+        'top': {
+            help: "Move the current cell to the top",
+            handler: function (env, event) {
+                if(event){
+                    event.preventDefault();
+                }
+                var cell = env.notebook.get_selected_index();
+                return env.notebook.scroll_to_cell(cell, 0);
             }
         },
         'save-notebook':{
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -378,7 +378,8 @@ define(function(require){
                 if(event){
                     event.preventDefault();
                 }
-                return env.notebook.scroll_to_bottom();
+                var cell = env.notebook.get_selected_index();
+                return env.notebook.scroll_to_cell(cell);
             }
         },
         'save-notebook':{
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -372,6 +372,15 @@ define(function(require){
                 return env.notebook.scroll_manager.scroll(-1);
             },
         },
+        'recenter-top-bottom': {
+            help: "Move the current cell to the center, top or bottom",
+            handler: function (env, event) {
+                if(event){
+                    event.preventDefault();
+                }
+                return env.notebook.scroll_to_bottom();
+            }
+        },
         'save-notebook':{
             help: "Save and Checkpoint",
             help_index : 'fb',
--- a/IPython/html/static/notebook/js/keyboardmanager.js
+++ b/IPython/html/static/notebook/js/keyboardmanager.js
@@ -79,7 +79,8 @@ define([
             'up'                  : 'ipython.move-cursor-up-or-previous-cell',
             'down'                : 'ipython.move-cursor-down-or-next-cell',
             'ctrl-shift--'        : 'ipython.split-cell-at-cursor',
-            'ctrl-shift-subtract' : 'ipython.split-cell-at-cursor'
+            'ctrl-shift-subtract' : 'ipython.split-cell-at-cursor',
+            'ctrl-l'              : 'ipython.recenter-top-bottom'
         };
     };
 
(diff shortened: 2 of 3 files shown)


Carreau:
Don't add default keybindings to ctrl-anything, they are already taken.

Carreau:
I'm not sure we should play more with the end space. 
We are already not super happy with having to play with it when the pager open/close, I'm not sure of all the implications.

Carreau:
I'll have a look to see how it feel. Having actions to scroll cell to some position seem nice. from the code I guess it's animate, which I woudl like to see.

abalkin:
> from the code I guess it's animate ..

Not really, I hardcoded animation time of 0, but this is easy to change.

abalkin:
> I'm not sure we should play more with the end space. 

Me too.  The feature is usable without the end space tricks as long as you have enough material below your current cell.

The issue that you cannot scroll the last cell to the top of the window is independent of the "recenter" feature.  There must be some standard solution that allows scrolling as long as certain element is visible.  In this case - the last cell of the notebook.

takluyver:
Just to echo what @Carreau said about shortcuts: pretty much all Ctrl+letter shortcuts are unavailable, unfortunately. Every time we've tried to use them, it clashes with something (and I already know what that is for Ctrl+L, because I use it all the time. This is one of the huge downsides of writing UI to run in browsers.

abalkin:
> Don't add default keybindings to ctrl-anything, they are already taken.

Since the feature is modeled after Emacs' Ctrl-L, which itself I think is modeled after terminal ^L code for clear-screen, I would like to keep "Ctrl-L" in the binding.  Note that ultimately, I would like to have only one key combination that will cycle between predefined positions (center - top - bottom by default).

What do you think about Ctrl-Shift-L?

I am not sure this is needed in command mode, but does anyone know what bindings vi uses for similar functions?

Finally, I've implemented "recenter" by moving the top of the cell to the center of the screen.  I don't think this is right.  It will probably be better to move the cursor position to the center instead.

abalkin:
`Ctrl-Shift-L` is toggle-zen-mode in github, which is probably a good sign that it is available for in-browser UI.

abalkin:
Apparently, vi uses `z.`, `z+` and `z-` shortcuts.  `z` is of course way overloaded, but `.` looks like a good candidate.  We can use `.` in command mode and `Ctrl-Shift-.` in edit mode.  Note that `Shift-.` is `>` which brings the right associations with a prompt or a line-tag.

rsmith31415:
Let me see if I understood. This is an attempt to get your third point: 

>Provide keyboard shortcuts to bring the current cell to the middle, top or bottom of the window.

So, the first one: 

>Allow main window scrolling as long as the last input cell is fully visible.

is not implemented yet. Is this correct?

abalkin:
@rsmith31415 - you are correct.

abalkin:
@dbfc1fd - dbfc1fd sort-of gives you "Allow main window scrolling as long as the last input cell is fully visible".  You can bring the last cell to the top by `Ctrl-Shift-L` and then it can be scrolled freely.

rsmith31415:
Oh, okay. I was confused when the first thing I tried was to put the last cell in the middle. 

Yes, I  think you need to readjust a bit the positions. They are too low, at the moment.

abalkin:
> .. you need to readjust a bit the positions.

I think this is due to some space between the need of "end-space" and the end of "notebook_panel".  I need to draw some nested boxes on a piece of paper to get it right.

ellisonbg:
I too am -1 on changing *code* to get docs to behave correctly.

abalkin:
@ellisonbg - I am not sure how to put your comment into the proper context.  What comment are you responding to?

Carreau:
I think Brian reply to the wrong thread. His response was ment for
https://github.com/ipython/ipython/pull/7876

ellisonbg:
Oops, sorry, wrong PR...

ellisonbg:
OK, comments about this one:

* I agree that we need to avoid `ctrl` based keyboard shortcuts.
* I like this type of idea, but I think we want wait to see how we end up fixing some of the other focus related problems. I think it is likely that in fixing those issues, we will get much better scrolling that has some of the benefits of the current PR. Let's get those fixed and revisit this.
* We could implement the logic, but not have any default keyboard shortcuts for now...

ellisonbg:
Decisions from the dev meeting

* Remove adjust_end_space
* Remove keyboard shortcuts
* Rename actions (scroll_cell_center, scroll_cell_top).
* Implement actions as single scroll_cell_precent that takes a parameter controllling the scroll position.

abalkin:
I should be able to get to this at the end of the week, but if someone wants to take over - please let me know.  (My Javascript is better than my French, but not by a wide margin.)

abalkin:
> Implement actions as single scroll_cell_precent ..

I guess that should have been `scroll_cell_percent`.

ellisonbg:
Opps, sorry, yes...

On Tue, Mar 31, 2015 at 11:48 AM, abalkin <notifications@github.com> wrote:

> Implement actions as single scroll_cell_precent ..
>
> I guess that should have been scroll_cell_percent.
>
> —
> <https://github.com/ipython/ipython/pull/8067#issuecomment-88206082>.
>



-- 
Brian E. Granger

minrk:
Since this hasn't been done yet, and it's in the way of the big split, I can finish it up. I want to split out html today.

minrk:
opened #8281, finishing this up.

minrk:
This was finished in #8281

PR #8072, opened 2015-03-17
Make nbconvert a little less chatty.

Carreau:
User don't really care what template is used by default.
and no need to say where the files might be written if they won't be.

### DIFF
--- a/IPython/nbconvert/writers/files.py
+++ b/IPython/nbconvert/writers/files.py
@@ -76,7 +76,11 @@ class FilesWriter(WriterBase):
             # Write all of the extracted resources to the destination directory.
             # NOTE: WE WRITE EVERYTHING AS-IF IT'S BINARY.  THE EXTRACT FIG
             # PREPROCESSOR SHOULD HANDLE UNIX/WINDOWS LINE ENDINGS...
-            for filename, data in resources.get('outputs', {}).items():
+
+            items = resources.get('outputs', {}).items()
+            if items:
+                self.log.info("Support files will be in %s", os.path.join(resources['output_files_dir'], ''))
+            for filename, data in items:
 
                 # Determine where to write the file to
                 dest = os.path.join(self.build_directory, filename)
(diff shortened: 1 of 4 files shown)


Carreau:
Fix @jhamrick complaint also. 

BTW I can `--author`the last commit as you if you like :-)

jhamrick:
:+1: 

> BTW I can --author the last commit as you if you like :-)

No, it's fine to leave it as it is.

This will probably conflict with #8073 which I just submitted but not in a major way.

Can I make one other suggestion, which is to make the "Executing notebook with kernel:" message in the execute preprocessor also be debug?

Carreau:
> This will probably conflict with #8073 which I just submitted but not in a major way.

Yes, just one line. I'll rebase this one once yours is merge. 



> Can I make one other suggestion, which is to make the "Executing notebook with kernel:" message in the execute preprocessor also be debug?

I'm torn. Execution can take quite some time, and without the message, people could think the conversion is stuck. I would like to know what others think.

IPython/nbconvert/writers/files.py
```diff
@@ -76,7 +76,11 @@ def write(self, output, resources, notebook_name=None, **kw):
             # Write all of the extracted resources to the destination directory.
             # NOTE: WE WRITE EVERYTHING AS-IF IT'S BINARY.  THE EXTRACT FIG
             # PREPROCESSOR SHOULD HANDLE UNIX/WINDOWS LINE ENDINGS...
-            for filename, data in resources.get('outputs', {}).items():
+
+            items = resources.get('outputs', {}).items()
+            if items:
+                self.log.info("Support files will be in %s", os.path.join(resources['output_files_dir'], ''))
```
takluyver: Test failure: `KeyError: 'output_files_dir'`

Carreau: Yep. seen travis angry red mail.

jhamrick:
> Execution can take quite some time, and without the message, people could think the conversion is stuck.

Hrm, good point. Ok, maybe it's fine to leave that in, then.

Carreau:
We can try to have a coroutine, that make  a spinning `/-\|/-\|/-\|/-\|/-\|/` or [others](http://stackoverflow.com/questions/2685435/cooler-ascii-spinners).

Carreau:
[Oh gosh](http://jsfiddle.net/CgkQJ/1160/) !

PR #8073, opened 2015-03-18
Break up convert_notebooks so that it is easier to subclass NbConvertApp

jhamrick:
As talked about in the dev meeting, here is a way of breaking up `convert_notebooks` so that `NbConvertApp` is easier to subclass. This adds a separate method to convert a single notebook, and also adds methods for each step of the conversion process. So, apps that subclass `NbConvertApp` can easily override any step in the process, whether that be initializing the resources dict, exporting, writing, or postprocessing.

### DIFF
--- a/IPython/nbconvert/nbconvertapp.py
+++ b/IPython/nbconvert/nbconvertapp.py
@@ -277,61 +277,128 @@ class NbConvertApp(BaseIPythonApplication):
         super(NbConvertApp, self).start()
         self.convert_notebooks()
 
+    def init_single_notebook_resources(self, notebook_filename):
+        """Step 1: Initialize resources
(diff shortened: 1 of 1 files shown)


jhamrick:
Rebased.

Carreau:
As @jhamrick will probably be running on this for her class, it would be nice for that to be one of the first 4.0 PR so that Jess could be considered as running a "3.1+" without major modifications of IPython.

I've tagged it as important just for that.

Carreau:
I'll merge that so you can at least pin a version of master before it breaks. 
Be carefull there have already been some changes on traitlets !

jhamrick:
Thanks!

PR #8078, opened 2015-03-18
Make unicode repr test work on cp1252 Windows

takluyver:
Just limit the unicode sample to characters that can be encoded in cp1252.

Should fix Windows + Python 2 test failures that we've been seeing.


takluyver:
Tests pass on Windows: https://jenkins.jupyter.org/job/windows-py2.7-prtest/7/

PR #8079, opened 2015-03-18
Autocompletion & docstrings for MATLAB interface.

anntzer:
For the Mathworks' official MATLAB engine for Python (which allows calling a MATLAB subprocess from Python), provide autocompletion of function names by manually walking the MATLAB search path; also provide docstrings for functions by calling MATLAB's builtin "help" function.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -133,7 +133,12 @@ def getdoc(obj):
 
     It also attempts to call a getdoc() method on the given object.  This
     allows objects which provide their docstrings via non-standard mechanisms
-    (like Pyro proxies) to still be inspected by ipython's ? system."""
+    (like Pyro proxies) to still be inspected by ipython's ? system.  It also
+    special-cases Mathworks' MatlabEngine."""
+
+    if (type(obj).__module__ == "matlab.engine.matlabengine" and
+        type(obj).__name__ == "MatlabFunc"):
+        return inspect.cleandoc(obj._engine().help(obj._name))
     # Allow objects to offer customized documentation via a getdoc method:
     try:
         ds = obj.getdoc()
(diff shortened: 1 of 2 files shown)


takluyver:
-1 - I want to remove these kinds of special cases from our code, not add more. It should be up to the matlab Python bridge to provide access to docs and introspection by standard Python mechanisms.

Carreau:
Agreed, if you need hooks, or configurability we might be able to add them. But you shouldn't need to modify IPython kernel. We don't do it for rpy2, cython and other bindings.

takluyver:
I think all the relevant hooks are already there:

- For docstrings, you can make `__doc__` a property that looks up the docs you want. I just checked that this works, but if for some reason it doesn't, we also look for and call `obj.getdoc()`, so you could define that.
- For providing a list of attributes, define the `__dir__()` method - [Python docs](https://docs.python.org/3/reference/datamodel.html#object.__dir__).

anntzer:
Given that IPython special-cases Traits and PyCrust objects (via `trait_names` and `_getAttributeNames`), I thought that it would also accept a special case for MATLAB, but I don't feel strongly about it either.  Feel free to close this if you want.

takluyver:
Those special case were put in years ago - I doubt we'd accept them today.

I will close this - if enough people feel it's reasonable to add another special case, we can reopen it.

takluyver:
Opened #8081 to remove the PyCrust special case. I hope we can ultimately remove the special case for traits too.

anntzer:
Actually there is another reason to remove the attempts to call `_getAttributeNames` and `trait_names`, which is that the object could be dynamically providing attributes (especially `trait_names`, which is not even private) and thus calling them could have some unspecified side-effects.  As usual, anything can have side-effects in Python but I now agree that relying on the standard `__dir__` is the least likely to cause problems.

takluyver:
Yep, we've run into a few things, typically object-proxying code, that has problems when we check for attributes like those.

anntzer:
As a side note, relying on trying to call a `getdoc` method for the documentation suffers from the same issue and I would thus suggest getting rid of this too, if possible.

takluyver:
That's true. I suspect it might be harder to get rid of, though. E.g. for something like the `MatlabFunc` class, you might want a standard docstring for the class itself, and dynamically retrieved docstrings for instances - and I don't think it's possible to do both with `__doc__`.

anntzer:
You can do this with a not-so-complicated descriptor:

```
class instanceprop:
    def __init__(self, classval, getinstanceval):
        self._classval = classval
        self._getinstanceval = getinstanceval

    def __get__(self, instance, owner):
        return self._classval if instance is None else self._getinstanceval(instance)


class C:
    __doc__ = instanceprop("class doc", lambda self: "instance doc")
```

Then
```
In [2]: c = C()

In [3]: c?
Type:            C
String form:     <instanceprop.C object at 0x7fd4a5cb8a20>
File:            /home/antony/tests/instanceprop.py
Docstring:       instance doc
Class docstring: class doc
```

Because `__doc__` is set in the locals before the rest of the class definition is executed you could even write the proper decorator so that you can use a normal-looking docstring in the source (left as an exercise to the reader, you need some stack-walking :-)).

Issue #8074, opened 2015-03-18
Update quick reference for History to contain useful information about sessions


nnathan:
It would be convenient for newcomers to include in the quick reference:
1. how to view history for previous ipython sessions
2. edit current and previous history sessions for execution

I think something along the lines of:

```
History:

_i, _ii, _iii    : Previous, next previous, next next previous input
_i4, _ih[2:5]    : Input history line 4, lines 2-4
exec _i81        : Execute input history line #81 again
%rep 81          : Edit input history line #81
_, __, ___       : previous, next previous, next next previous output
_dh              : Directory history
_oh              : Output history
%hist            : Command history of current session. '%hist -g foo' search history for 'foo'
%hist 1/ ~2/     : Command history consisting of session 1 and 2 sessions before current.
%hist 1/1-10     : Command history containing lines 1-10 of session 1.
%hist ~8/1-~6/5  : Command history from first line of 8 sessions ago, to the fifth line of 6 sessions ago.
%edit 0/        : Open up editor to execute code pre-filled with history from current session.
```

The above examples are [shamelessly stolen from IPython.core.magics.history reference docs.](http://ipython.org/ipython-doc/2/api/generated/IPython.core.magics.history.html)

takluyver:
PRs welcome ;-)

PR #8080, opened 2015-03-18
provided more history examples in quick reference

nnathan:
This provides a few more concrete examples of using history and editing with an existing history. I struggled to find examples of editing the full history in an editor and [answers like these](http://stackoverflow.com/questions/16050463/how-to-open-ipython-command-history-on-a-text-editor) didn't really tell the full story. Similarly [trying to get the entire history](http://stackoverflow.com/questions/25124037/ipython-print-complete-history-not-just-current-session) is a tiny comment that is easy to miss.

As I was invited to submit a PR for #8074. Here is what I think the quick reference should include under History.

### DIFF
--- a/IPython/core/usage.py
+++ b/IPython/core/usage.py
@@ -306,7 +306,14 @@ exec _i81        : Execute input history line #81 again
 _, __, ___       : previous, next previous, next next previous output
 _dh              : Directory history
 _oh              : Output history
-%hist            : Command history. '%hist -g foo' search history for 'foo'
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, that looks fine to me - I'll give others a day or so to look at it, and merge if there are no objections.

Carreau:
Looks fine to me, if it can be helpful for someone;

minrk:
:+1:

takluyver:
Great, thanks.

PR #8083, opened 2015-03-18
fix javascript mime-type in HTML exporter

minrk:
It's application/javascript, not text/javascript. The result was javascript output never showing up in nbconvert.


takluyver:
:+1: , but needs a rebase.

minrk:
Rebased.

PR #8091, opened 2015-03-19
TEST: Move test_escape_root to TestFileContentsManager

ssanderson:
The tests for [pgcontents](https://github.com/quantopian/pgcontents) assume that `TestContentsManager` is written against only the public API of the ContentsManager base class. The `root_dir` attribute is specific to the FileContentsManager implementation.


ssanderson:
ping @minrk.  I'm looking at upgrading pgcontents to be properly 3.0 compatible, and test_escape_root is  the only test I have left failing.

Carreau:
Relaunched tests.

Carreau:
+1

minrk:
Yup, sorry about that.

PR #8092, opened 2015-03-19
log error when there is one.

Carreau:
Otherwise it's a pain to debug

---

We already show the end message to the user, but having the full stack in console is helpful to debug...

Well it's JS so not **that** helpful, but still.


takluyver:
:+1:

PR #8098, opened 2015-03-20
Remove image resizer

ghost:
This PR removes the image resizer for images. Removing the resizer helps a lot to solve the image overflow issue in #7701. Related PR: #8062.

Before:
screenshot from 2015-03-20 10 48 02

After:

screenshot from 2015-03-20 10 49 48

### DIFF
--- a/IPython/html/static/notebook/js/outputarea.js
+++ b/IPython/html/static/notebook/js/outputarea.js
@@ -661,12 +661,6 @@ define([
             .width(width)
             .height(height);
 
-        // The jQuery resize handlers don't seem to work on the svg element.
-        // When the svg renders completely, measure it's size and set the parent
-        // div to that size.  Then set the svg to 100% the size of the parent
-        // div and make the parent div resizable.  
-        this._dblclick_to_reset_size(svg_area, true, false);
-
         svg_area.append(svg);
         toinsert.append(svg_area);
         element.append(toinsert);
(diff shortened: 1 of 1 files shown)


jdfreder:
-1 because I like this feature.

rsmith31415:
I don't have anything against this feature but it is causing significant issues and preventing us from implementing solutions that are trivial CSS fixes (e.g. something like [this](https://github.com/ipython/ipython/compare/master...rsmith31415:css-fix-width-images)  as a fix to #7701).

Furthermore, the current jquery-ui `.resizer` is not responsive-friendly, so we can't have a fully responsive (or even fluid) layout. That is by far the more important consideration in terms of user experience and as a foundation to prevent more related issues.

Finally, the resizer (I think) is not widely used or even easily recognized. Its removal is not likely to disrupt the workflow of users. Actually, I would worry about excessively using the resizer to obtain larger images instead of doing it programmatically.

minrk:
We are going to merge all the things - remove resize, add simpler 'expand to original' from #8062, *and* scroll the output_subarea on overflow.

cancan101:
Given that the resize handle is now gone, how do I make the image render larger? I am trying this and it does not seem to work:
```
current = %config InlineBackend.rc
current['figure.figsize'] = (9,6)
%config InlineBackend.rc = current
```

rsmith31415:
@cancan101 

I think this is a question related to matplotlib. You can simply use:

```
import matplotlib
%matplotlib inline

matplotlib.rcParams['figure.figsize'] = (12,9)
```

cancan101:
That seems to work. What then does `InlineBackend.rc['figure.figsize']` even mean?

Carreau:
that tells the default size for the figure when using inline figures with
matplotlib.

cancan101:
I have `%matplotlib inline` set and am plotting using:
```
    import matplotlib.pyplot as plt
    plt.imshow(img)
```
and `InlineBackend.rc['figure.figsize']` does not seem to do anything.

Carreau:
you can try :

   import matplotlib
  matplotlib.rcParams['figure.figsize'] = [sizex, sizey]

but anyway the better way to do it is to learn the OO way of matplotlib.

cancan101:
OO way?

Carreau:
Object oriented. Create figure and axis references and manipute them. 

```
%matplotlib inline
import matplotlib.pyplot as plt
fig, ax = plt.subplots(1,1)
fig.set_figwidth(7)
fig.set_figheight(11)
ax.plot(range(10))
```

rsmith31415:
If I understood @cancan101 's question correctly, I think he wants to set the image size for all figures in the notebook. Otherwise, there is no need to be manipulating global variables.

By the way, it is easier to do this:

```
fig, ax = plt.subplots(figsize=(7,11))
ax.plot(range(10))
```

robcross:
I really really miss the resizer!!!

Changing the figsize is NOT alternative to the resize handle. 

```python
fig, ax = plt.subplots(figsize=(8,10))
ax.plot(range(10))
plt.show()

fig, ax = plt.subplots(figsize=(4,5))
ax.plot(range(10))
plt.show()
```

The font size and line widths are the same, but the plot is twice as big. When I'm trying to make publication ready figures, it is nice so set the figsize that will put the text in proportion with the plot, and use the resize handle to get a closer look.

Is there another solution to quickly and easily scale figures? Like actually scale everything? Or do I have to output to a file and view it from there now?

PR #8100, opened 2015-03-20
fixes for parallel caused by recent PRs

minrk:
- #7798 turned some bytes into memoryviews, causing a few copy calls to fail, since memoryviews are not copyable by default.
- #7708 caught a few unnecessary intiial assignments (that probably predated traitlets) in the launchers that can be removed.


Carreau:
+1

PR #8107, opened 2015-03-21
allow_none=False by default for Instance and Type

SylvainCorlay:
The last trait types for which `allow_none` was `True` by default were `Type` and `Instance`. Since unused default values are not validated anymore, it makes sense to set it to False, so that there is no special case. 

After this, all trait types have `allow_none=False` by default.

having allow_none=False for `Instance` makes sense in  the following cases: 
- container trait types like
  `List(Instance(Foo))`, 
- when we want force the user to specify a value for an attribute

``` python
class Foo(HasTraits):
    # Foo must hold a reference to an instance of Bar provided
    # by the user.
    bar = Instance(Bar)

Foo()   # Trait Error
bar = Bar()
Foo(bar=bar)  # ok
```

This also corrects a bug which I just noticed: setting allow_none=False was not effective on `Tuple`.

### DIFF
--- a/IPython/config/application.py
+++ b/IPython/config/application.py
@@ -174,7 +174,7 @@ class Application(SingletonConfigurable):
         _log_handler.setFormatter(_log_formatter)
     
 
-    log = Instance(logging.Logger, allow_none=True)
+    log = Instance(logging.Logger)
     def _log_default(self):
         """Start logging for this application.
 
--- a/IPython/config/application.py
+++ b/IPython/config/application.py
@@ -174,7 +174,7 @@ class Application(SingletonConfigurable):
         _log_handler.setFormatter(_log_formatter)
     
 
-    log = Instance(logging.Logger)
+    log = Instance(logging.Logger, allow_none=True)
     def _log_default(self):
         """Start logging for this application.
 
--- a/IPython/config/configurable.py
+++ b/IPython/config/configurable.py
@@ -372,7 +372,7 @@ class LoggingConfigurable(Configurable):
     is to get the logger from the currently running Application.
     """
 
-    log = Instance('logging.Logger', allow_none=True)
+    log = Instance('logging.Logger')
     def _log_default(self):
         from IPython.utils import log
         return log.get_logger()
--- a/IPython/config/configurable.py
+++ b/IPython/config/configurable.py
@@ -34,7 +34,7 @@ class MultipleInstanceError(ConfigurableError):
 class Configurable(HasTraits):
 
     config = Instance(Config, (), {})
-    parent = Instance('IPython.config.configurable.Configurable')
+    parent = Instance('IPython.config.configurable.Configurable', allow_none=True)
 
     def __init__(self, **kwargs):
         """Create a configurable given a config config.
@@ -372,7 +372,7 @@ class LoggingConfigurable(Configurable):
     is to get the logger from the currently running Application.
     """
 
-    log = Instance('logging.Logger')
+    log = Instance('logging.Logger', allow_none=True)
     def _log_default(self):
         from IPython.utils import log
         return log.get_logger()
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -295,13 +295,13 @@ class InteractiveShell(SingletonConfigurable):
     disable_failing_post_execute = CBool(False, config=True,
         help="Don't call post-execute functions that have failed in the past."
     )
-    display_formatter = Instance(DisplayFormatter)
-    displayhook_class = Type(DisplayHook)
-    display_pub_class = Type(DisplayPublisher)
+    display_formatter = Instance(DisplayFormatter, allow_none=True)
+    displayhook_class = Type(DisplayHook, allow_none=True)
+    display_pub_class = Type(DisplayPublisher, allow_none=True)
     data_pub_class = None
 
     exit_now = CBool(False)
-    exiter = Instance(ExitAutocall)
+    exiter = Instance(ExitAutocall, allow_none=True)
     def _exiter_default(self):
         return ExitAutocall(self)
     # Monotonically increasing execution counter
@@ -435,16 +435,16 @@ class InteractiveShell(SingletonConfigurable):
                             default_value='Context', config=True)
 
     # Subcomponents of InteractiveShell
-    alias_manager = Instance('IPython.core.alias.AliasManager')
-    prefilter_manager = Instance('IPython.core.prefilter.PrefilterManager')
-    builtin_trap = Instance('IPython.core.builtin_trap.BuiltinTrap')
-    display_trap = Instance('IPython.core.display_trap.DisplayTrap')
-    extension_manager = Instance('IPython.core.extensions.ExtensionManager')
-    payload_manager = Instance('IPython.core.payload.PayloadManager')
-    history_manager = Instance('IPython.core.history.HistoryAccessorBase')
-    magics_manager = Instance('IPython.core.magic.MagicsManager')
+    alias_manager = Instance('IPython.core.alias.AliasManager', allow_none=True)
+    prefilter_manager = Instance('IPython.core.prefilter.PrefilterManager', allow_none=True)
+    builtin_trap = Instance('IPython.core.builtin_trap.BuiltinTrap', allow_none=True)
+    display_trap = Instance('IPython.core.display_trap.DisplayTrap', allow_none=True)
+    extension_manager = Instance('IPython.core.extensions.ExtensionManager', allow_none=True)
+    payload_manager = Instance('IPython.core.payload.PayloadManager', allow_none=True)
+    history_manager = Instance('IPython.core.history.HistoryAccessorBase', allow_none=True)
+    magics_manager = Instance('IPython.core.magic.MagicsManager', allow_none=True)
 
-    profile_dir = Instance('IPython.core.application.ProfileDir')
+    profile_dir = Instance('IPython.core.application.ProfileDir', allow_none=True)
     @property
     def profile(self):
         if self.profile_dir is not None:
@@ -453,7 +453,7 @@ class InteractiveShell(SingletonConfigurable):
 
 
     # Private interface
-    _post_execute = Instance(dict)
+    _post_execute = Instance(dict, allow_none=True)
 
     # Tracks any GUI loop loaded for pylab
     pylab_gui_select = None
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -301,7 +301,7 @@ class InteractiveShell(SingletonConfigurable):
     data_pub_class = None
 
     exit_now = CBool(False)
-    exiter = Instance(ExitAutocall, allow_none=True)
+    exiter = Instance(ExitAutocall)
     def _exiter_default(self):
         return ExitAutocall(self)
     # Monotonically increasing execution counter
@@ -453,7 +453,7 @@ class InteractiveShell(SingletonConfigurable):
 
 
     # Private interface
-    _post_execute = Instance(dict, allow_none=True)
+    _post_execute = Dict()
 
     # Tracks any GUI loop loaded for pylab
     pylab_gui_select = None
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -73,7 +73,7 @@ from IPython.utils.syspathcontext import prepended_to_syspath
 from IPython.utils.text import (format_screen, LSString, SList,
                                 DollarFormatter)
 from IPython.utils.traitlets import (Integer, Bool, CBool, CaselessStrEnum, Enum,
-                                     List, Unicode, Instance, Type)
+                                     List, Dict, Unicode, Instance, Type)
 from IPython.utils.warn import warn, error
 import IPython.core.hooks
 
@@ -296,8 +296,8 @@ class InteractiveShell(SingletonConfigurable):
         help="Don't call post-execute functions that have failed in the past."
     )
     display_formatter = Instance(DisplayFormatter, allow_none=True)
-    displayhook_class = Type(DisplayHook, allow_none=True)
-    display_pub_class = Type(DisplayPublisher, allow_none=True)
+    displayhook_class = Type(DisplayHook)
+    display_pub_class = Type(DisplayPublisher)
     data_pub_class = None
 
     exit_now = CBool(False)
--- a/IPython/html/services/clusters/clustermanager.py
+++ b/IPython/html/services/clusters/clustermanager.py
@@ -20,7 +20,7 @@ class ClusterManager(LoggingConfigurable):
     delay = Float(1., config=True,
         help="delay (in s) between starting the controller and the engines")
 
-    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
+    loop = Instance('zmq.eventloop.ioloop.IOLoop')
     def _loop_default(self):
         from zmq.eventloop.ioloop import IOLoop
         return IOLoop.instance()
--- a/IPython/html/services/clusters/clustermanager.py
+++ b/IPython/html/services/clusters/clustermanager.py
@@ -20,7 +20,7 @@ class ClusterManager(LoggingConfigurable):
     delay = Float(1., config=True,
         help="delay (in s) between starting the controller and the engines")
 
-    loop = Instance('zmq.eventloop.ioloop.IOLoop')
+    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
     def _loop_default(self):
         from zmq.eventloop.ioloop import IOLoop
         return IOLoop.instance()
--- a/IPython/kernel/inprocess/ipkernel.py
+++ b/IPython/kernel/inprocess/ipkernel.py
@@ -46,7 +46,7 @@ class InProcessKernel(IPythonKernel):
     # Kernel interface
     #-------------------------------------------------------------------------
 
-    shell_class = Type()
+    shell_class = Type(allow_none=True)
     shell_streams = List()
     control_stream = Any()
     iopub_socket = Instance(DummySocket, ())
--- a/IPython/kernel/inprocess/ipkernel.py
+++ b/IPython/kernel/inprocess/ipkernel.py
@@ -27,7 +27,8 @@ class InProcessKernel(IPythonKernel):
 
     # The frontends connected to this kernel.
     frontends = List(
-        Instance('IPython.kernel.inprocess.client.InProcessKernelClient')
+        Instance('IPython.kernel.inprocess.client.InProcessKernelClient'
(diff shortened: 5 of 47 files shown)


SylvainCorlay:
This still requires some cleanup.

minrk:
Are there any examples remaining where the default behavior is used? If not, this probably is not the right thing to do.

SylvainCorlay:
- It I type `List(trait=Instance(Foo))`, I would not expect `[None]` to be allowed.
- There are number of cases in IPython codebase were args and kwargs are provided to construct a default value. In those case, I did not need to specify `allow_none=True`
- and currently, having `allow_none=True` by default makes Type and Instance special cases that people have to know about.

SylvainCorlay:
Rebased.

IPython/core/interactiveshell.py
```diff
@@ -295,13 +295,13 @@ class InteractiveShell(SingletonConfigurable):
     disable_failing_post_execute = CBool(False, config=True,
         help="Don't call post-execute functions that have failed in the past."
     )
-    display_formatter = Instance(DisplayFormatter)
-    displayhook_class = Type(DisplayHook)
-    display_pub_class = Type(DisplayPublisher)
+    display_formatter = Instance(DisplayFormatter, allow_none=True)
+    displayhook_class = Type(DisplayHook, allow_none=True)
+    display_pub_class = Type(DisplayPublisher, allow_none=True)
```
minrk: Type is given a default value. Why is allow_none needed here?

IPython/kernel/connect.py
```diff
@@ -432,7 +432,7 @@ def ports(self):
         return [ getattr(self, name) for name in port_names ]
 
     # The Session to use for communication with the kernel.
-    session = Instance('IPython.kernel.zmq.session.Session')
+    session = Instance('IPython.kernel.zmq.session.Session', allow_none=True)
```
minrk: This shouldn't need to allow None, either, since its default is set

IPython/kernel/inprocess/ipkernel.py
```diff
@@ -48,8 +49,8 @@ class InProcessKernel(IPythonKernel):
     shell_class = Type()
     shell_streams = List()
     control_stream = Any()
-    iopub_socket = Instance(DummySocket, ())
-    stdin_socket = Instance(DummySocket, ())
+    iopub_socket = Instance(DummySocket, (), allow_none=True)
+    stdin_socket = Instance(DummySocket, (), allow_none=True)
```
minrk: Again, no need for allow_none

minrk:
I think there's something unexpected in how these traits are initialized, because there are many cases where you have added `allow_none=True` that shouldn't need it. I'm guessing that it's because they are validating a None value that they don't actually use. Can you look into that?

IPython/kernel/connect.py
SylvainCorlay: Hum, is it? There is no argument to construct a Session object.

IPython/kernel/inprocess/ipkernel.py
SylvainCorlay: Agree on this one. I got a bit heavy-handed in adding these `allow_none=True`

IPython/core/interactiveshell.py
SylvainCorlay: Agreed, should not be needed.

minrk:
From your points, I think this change in behavior is the right thing to do. I don't think there should be any need for `allow_none=True` on any of our uses of Type, and I think a few of the Instances with `_foo_default` defined or default constructor args got `allow_none=True` that shouldn't need them. If they do need them, then there's a problem with initialization, where an unused default value is getting validated.

IPython/config/application.py
```diff
@@ -174,7 +174,7 @@ def _log_format_changed(self, name, old, new):
         _log_handler.setFormatter(_log_formatter)
     
 
-    log = Instance(logging.Logger)
+    log = Instance(logging.Logger, allow_none=True)
```
minrk: This has _log_default, so shouldn't need allow_none

IPython/config/configurable.py
```diff
@@ -372,7 +372,7 @@ class LoggingConfigurable(Configurable):
     is to get the logger from the currently running Application.
     """
 
-    log = Instance('logging.Logger')
+    log = Instance('logging.Logger', allow_none=True)
```
minrk: no allow_none

IPython/core/interactiveshell.py
```diff
@@ -295,13 +295,13 @@ class InteractiveShell(SingletonConfigurable):
...
     data_pub_class = None
 
     exit_now = CBool(False)
-    exiter = Instance(ExitAutocall)
+    exiter = Instance(ExitAutocall, allow_none=True)
     def _exiter_default(self):
         return ExitAutocall(self)
```
minrk: no allow_none

IPython/core/interactiveshell.py
```diff
@@ -453,7 +453,7 @@ def profile(self):
 
 
     # Private interface
-    _post_execute = Instance(dict)
+    _post_execute = Instance(dict, allow_none=True)
```
minrk: This should actually be a plain Dict. Not sure how it got Instance. Can you remove the `self._post_execute = {}` in `__init__` as well?

IPython/html/services/clusters/clustermanager.py
```diff
@@ -20,7 +20,7 @@ class ClusterManager(LoggingConfigurable):
     delay = Float(1., config=True,
         help="delay (in s) between starting the controller and the engines")
 
-    loop = Instance('zmq.eventloop.ioloop.IOLoop')
+    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
     def _loop_default(self):
```
minrk: no allow_none

IPython/kernel/client.py
```diff
@@ -49,7 +49,7 @@ class KernelClient(ConnectionFileMixin):
     """
 
     # The PyZMQ Context to use for communication with the kernel.
-    context = Instance(zmq.Context)
+    context = Instance(zmq.Context, allow_none=True)
```
minrk: no allow_none

IPython/kernel/comm/comm.py
```diff
@@ -20,7 +20,8 @@ class Comm(LoggingConfigurable):
     # If this is instantiated by a non-IPython kernel, shell will be None
     shell = Instance('IPython.core.interactiveshell.InteractiveShellABC',
                      allow_none=True)
-    kernel = Instance('IPython.kernel.zmq.kernelbase.Kernel')
+    kernel = Instance('IPython.kernel.zmq.kernelbase.Kernel',
+                      allow_none=True)
```
minrk: no allow_none

IPython/kernel/comm/comm.py
```diff
@@ -28,7 +29,8 @@ def _kernel_default(self):
     iopub_socket = Any()
     def _iopub_socket_default(self):
         return self.kernel.iopub_socket
-    session = Instance('IPython.kernel.zmq.session.Session')
+    session = Instance('IPython.kernel.zmq.session.Session',
+                       allow_none=True)
```
minrk: no allow_none

IPython/kernel/comm/manager.py
```diff
@@ -31,12 +31,13 @@ class CommManager(LoggingConfigurable):
...
                      allow_none=True)
-    kernel = Instance('IPython.kernel.zmq.kernelbase.Kernel')
+    kernel = Instance('IPython.kernel.zmq.kernelbase.Kernel',
+                      allow_none=True)
 
     iopub_socket = Any()
     def _iopub_socket_default(self):
         return self.kernel.iopub_socket
-    session = Instance('IPython.kernel.zmq.session.Session')
+    session = Instance('IPython.kernel.zmq.session.Session', allow_none=True)
```
minrk: no allow_none

minrk:
It seems like most cases where we have `allow_none=True` on Instance it's because the initial assignment happens after the construction of the object, so None is validated even though it is never used. Most of the cases where `allow_none=True` is added, the trait should actually not be allowed to be None.

It seems like the fix should be for the case:

```
class C(HasTraits):
    inst = Instance(SomeObject)

c1 = C(inst=SomeObject()) # correctly skips validation of unused default

c2 = C()
c2.inst # should validate default None

c3 = C()
c3.inst = SomeObject() # shouldn't validate unused default, but does
```

The default value retrieved when setting `c.inst` should *not* be validated, since it is guaranteed to never be used.

Given how traits work, I'm not sure that's feasible to fix right now, but it seems like the right behavior. If you don't think that's doable here, I think we can have the extra allow_none=True as an acceptable wart that we can work on removing later.

SylvainCorlay:
I completely agree with you. I guess that we can live with allowing None in the last case
```
c3 = C()
c3.inst = SomeObject() 
```
since we were allowing None all the time so far.

SylvainCorlay:
Rebased.

IPython/core/interactiveshell.py
```diff
@@ -453,7 +453,7 @@ def profile(self):
 
 
     # Private interface
-    _post_execute = Instance(dict)
+    _post_execute = {}
```
minrk: Sorry, I mean regular `Dict()` traitlet here, not a class attribute dict.

IPython/kernel/ioloop/manager.py
```diff
@@ -36,11 +36,11 @@ def wrapped(self, *args, **kwargs):
 
 class IOLoopKernelManager(KernelManager):
 
-    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=False)
+    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
```
minrk: no allow_none needed here, right?

IPython/kernel/manager.py
```diff
@@ -57,14 +57,14 @@ def _client_class_changed(self, name, old, new):
     # generally a Popen instance
     kernel = Any()
     
-    kernel_spec_manager = Instance(kernelspec.KernelSpecManager)
+    kernel_spec_manager = Instance(kernelspec.KernelSpecManager, allow_none=True)
```
minrk: no allow_none here

IPython/kernel/manager.py
```diff
@@ -57,14 +57,14 @@ def _client_class_changed(self, name, old, new):
...
     
     def _kernel_spec_manager_default(self):
         return kernelspec.KernelSpecManager(ipython_dir=self.ipython_dir)
     
     kernel_name = Unicode(kernelspec.NATIVE_KERNEL_NAME)
     
-    kernel_spec = Instance(kernelspec.KernelSpec)
+    kernel_spec = Instance(kernelspec.KernelSpec, allow_none=True)
```
minrk: no allow_none here

IPython/kernel/zmq/session.py
```diff
@@ -144,13 +144,13 @@ def _logname_changed(self, name, old, new):
...
-    context = Instance('zmq.Context')
+    context = Instance('zmq.Context', allow_none=True)
     def _context_default(self):
         return zmq.Context.instance()
 
-    session = Instance('IPython.kernel.zmq.session.Session')
+    session = Instance('IPython.kernel.zmq.session.Session', allow_none=True)
 
-    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=False)
+    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
```
minrk: no allow_none here

IPython/kernel/zmq/session.py
```diff
@@ -144,13 +144,13 @@ def _logname_changed(self, name, old, new):
         self.log = logging.getLogger(new)
 
     # not configurable:
-    context = Instance('zmq.Context')
+    context = Instance('zmq.Context', allow_none=True)
```
minrk: no allow_none

IPython/kernel/zmq/zmqshell.py
```diff
@@ -367,7 +367,7 @@ def _banner1_default(self):
     # will print a warning in the absence of readline.
     autoindent = CBool(False)
 
-    exiter = Instance(ZMQExitAutocall)
+    exiter = Instance(ZMQExitAutocall, allow_none=True)
```
minrk: no allow_none

ipython_parallel/apps/baseapp.py
```diff
@@ -123,7 +123,7 @@ def _cluster_id_changed(self, name, old, new):
     def _config_files_default(self):
         return ['ipcontroller_config.py', 'ipengine_config.py', 'ipcluster_config.py']
     
-    loop = Instance('zmq.eventloop.ioloop.IOLoop')
+    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
```
minrk: no allow_none

ipython_parallel/apps/logwatcher.py
```diff
@@ -50,13 +50,13 @@ def _url_default(self):
         return 'tcp://%s:20202' % localhost()
     
     # internals
-    stream = Instance('zmq.eventloop.zmqstream.ZMQStream')
+    stream = Instance('zmq.eventloop.zmqstream.ZMQStream', allow_none=True)
     
-    context = Instance(zmq.Context)
+    context = Instance(zmq.Context, allow_none=True)
```
minrk: no allow_none

ipython_parallel/apps/logwatcher.py
```diff
@@ -50,13 +50,13 @@ def _url_default(self):
...
     def _context_default(self):
         return zmq.Context.instance()
     
-    loop = Instance(zmq.eventloop.ioloop.IOLoop)
+    loop = Instance(zmq.eventloop.ioloop.IOLoop, allow_none=True)
```
minrk: no allow_none

ipython_parallel/client/view.py
```diff
@@ -1099,7 +1099,7 @@ def map(self, f, *sequences, **kwargs):
         -------
 
         if block=False
-          An :class:`~ipython_parallel.client.asyncresult.AsyncMapResult` instance.
+          An :class:`~IPython.parallel.client.asyncresult.AsyncMapResult` instance.
           An object like AsyncResult, but which reassembles the sequence of results
           into a single list. AsyncMapResults can be iterated through before all
           results are complete.
```
minrk: Looks like this file reverted some changes in the rebase

ipython_parallel/controller/heartmonitor.py
```diff
@@ -77,9 +77,9 @@ class HeartMonitor(LoggingConfigurable):
         help='Allowed consecutive missed pings from controller Hub to engine before unregistering.',
     )
 
-    pingstream=Instance('zmq.eventloop.zmqstream.ZMQStream')
-    pongstream=Instance('zmq.eventloop.zmqstream.ZMQStream')
-    loop = Instance('zmq.eventloop.ioloop.IOLoop')
+    pingstream=Instance('zmq.eventloop.zmqstream.ZMQStream', allow_none=True)
+    pongstream=Instance('zmq.eventloop.zmqstream.ZMQStream', allow_none=True)
+    loop = Instance('zmq.eventloop.ioloop.IOLoop', allow_none=True)
```
minrk: no allow_none on `loop`

ipython_parallel/controller/scheduler.py
```diff
@@ -180,17 +180,17 @@ def _scheme_name_changed(self, old, new):
...
-    query_stream = Instance(zmqstream.ZMQStream) # hub-facing DEALER stream
+    client_stream = Instance(zmqstream.ZMQStream, allow_none=True) # client-facing stream
+    engine_stream = Instance(zmqstream.ZMQStream, allow_none=True) # engine-facing stream
+    notifier_stream = Instance(zmqstream.ZMQStream, allow_none=True) # hub-facing sub stream
+    mon_stream = Instance(zmqstream.ZMQStream, allow_none=True) # hub-facing pub stream
+    query_stream = Instance(zmqstream.ZMQStream, allow_none=True) # hub-facing DEALER stream
 
     # internals:
-    queue = Instance(deque) # sorted list of Jobs
+    queue = Instance(deque, allow_none=True) # sorted list of Jobs
```
minrk: no allow_none on queue

ipython_parallel/controller/scheduler.py
```diff
@@ -180,17 +180,17 @@ def _scheme_name_changed(self, old, new):
         self.scheme = globals()[new]
 
     # input arguments:
-    scheme = Instance(FunctionType) # function for determining the destination
+    scheme = Instance(FunctionType, allow_none=True) # function for determining the destination
```
minrk: no allow_none on scheme

jupyter_nbformat/sign.py
```diff
@@ -87,7 +87,7 @@ def signature_removed(nb):
 class NotebookNotary(LoggingConfigurable):
     """A class for computing and verifying notebook signatures."""
     
-    profile_dir = Instance("IPython.core.profiledir.ProfileDir")
+    profile_dir = Instance("IPython.core.profiledir.ProfileDir", allow_none=True)
```
minrk: no allow_none

jupyter_nbformat/sign.py
```diff
@@ -388,7 +388,7 @@ class TrustNotebookApp(BaseIPythonApplication):
         """
     )
     
-    notary = Instance(NotebookNotary)
+    notary = Instance(NotebookNotary, allow_none=True)
```
minrk: no allow_none

ipython_parallel/client/view.py
SylvainCorlay: arg :(

IPython/core/interactiveshell.py
```diff
@@ -453,7 +453,7 @@ def profile(self):
 
 
     # Private interface
-    _post_execute = Instance(dict)
+    _post_execute = Dict()
```
minrk: Dict hasn't been imported

minrk:
@SylvainCorlay ping me when you've finished removing the extra allow_none=True, and we can get this one in.

SylvainCorlay:
Sorry for the late reply. I was swamped with meetings and messed up the rebase between two of them. Looking at it now.

IPython/kernel/inprocess/ipkernel.py
```diff
@@ -27,7 +27,8 @@ class InProcessKernel(IPythonKernel):
 
     # The frontends connected to this kernel.
     frontends = List(
-        Instance('IPython.kernel.inprocess.client.InProcessKernelClient')
+        Instance('IPython.kernel.inprocess.client.InProcessKernelClient'
```
minrk: missing comma, here

SylvainCorlay: arg, you beat me to it. I just pushed it.

minrk:
All, green. Thanks for finishing this one!

PR #8111, opened 2015-03-21
Fix error when log_datefmt is changed

### LINKED ISSUE #8110: `_log_format_changed()` missing 3 required positional arguments
In `IPython.config.application`, the `_log_datefmt_changed` handler calls `_log_format_changed` but does not pass it the required arguments. My guess would be that this is the correct implementation:

```
def _log_datefmt_changed(self, name, old, new):
    self._log_format_changed(name, self.log_format, self.log_format)
```

However I am not really sure what the `name` parameter is for, so I might be wrong.

jhamrick:
Fixes #8110

### DIFF
--- a/IPython/config/application.py
+++ b/IPython/config/application.py
@@ -159,7 +159,7 @@ class Application(SingletonConfigurable):
         help="The date format used by logging formatters for %(asctime)s"
     )
     def _log_datefmt_changed(self, name, old, new):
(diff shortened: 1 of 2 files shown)


takluyver:
Is it worth adding a smoketest?

jhamrick:
Ok, test added.

takluyver:
Thanks

PR #8113, opened 2015-03-22
load kwarg traits before loading config

minrk:
in `Configurable.__init__`

avoids problems in the new init of traitlets that skips `_trait_default` if specified in kwargs, which means really nothing should be done before `HasTraits.__init__` has been called.

This does mean that such values are assigned twice. Typically, this will have no effect, except when a value is set by both config and kwarg.

/cc @takluyver because this should fix the failure of the parallel tests to start

I will work on a separate PR to make the traits init more robust to this behavior, because I think there can be other cases of order problems, where the dynamic init should still be called.

### DIFF
--- a/IPython/config/configurable.py
+++ b/IPython/config/configurable.py
@@ -69,6 +69,11 @@ class Configurable(HasTraits):
             self.parent = parent
         
         config = kwargs.pop('config', None)
+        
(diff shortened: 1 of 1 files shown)


minrk:
I think I can do better than this. I'll open another PR, shortly.

PR #8114, opened 2015-03-22
add HasTraits.hold_trait_notifications

minrk:
context manager for holding trait notifications for a batch update of traitlets, used to avoid race conditions when loading multiple values from config or init kwargs.

Used in `HasTraits.__init__` and `Configurable._load_config`

### DIFF
--- a/IPython/config/configurable.py
+++ b/IPython/config/configurable.py
@@ -138,17 +138,20 @@ class Configurable(HasTraits):
             section_names = self.section_names()
         
         my_config = self._find_my_config(cfg)
-        for name, config_value in iteritems(my_config):
-            if name in traits:
-                if isinstance(config_value, LazyConfigValue):
-                    # ConfigValue is a wrapper for using append / update on containers
-                    # without having to copy the 
-                    initial = getattr(self, name)
-                    config_value = config_value.get_value(initial)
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Since the base Widget class has the `hold_sync`, which delays the messages to the js side. How about `hold_trait`?

minrk:
It's not hodling the trait at all, it's holding notification, so maybe `hold_notification`.

minrk:
Since it's a trait method, I think having `trait` in the name is better for trait-related public methods, so probably `hold_trait_notifications`. Since HasTraits is a very deep base class, I think its methods should be quite explicit about being trait-related, so as to avoid indication that they pertain to the terminal class's functionality.

SylvainCorlay:
Yes hold_trait_notifications sounds right. (I edited my initial message right after posting to change it from hold_trait to hold_trait_notifications.)

SylvainCorlay:
What about the counter for the nested contexts?

minrk:
Is there any need for a counter? Nested contexts seem to work properly as-is.

minrk:
Renamed to `hold_trait_notifications`, and nested behavior verified with a test. Thanks for the help reviewing.

SylvainCorlay:
Indeed, this one does not work like the hold_sync, and no need for a counter.

PR #8117, opened 2015-03-23
add utils.getargspec

minrk:
Finish up #7498 by moving the getargspec function to utils.inspect.
The code is verbatim from sphinx, other than imports.


Carreau:
+1

takluyver:
This feels more like a 4.0 thing than a 3.1 thing to me, though I don't feel strongly about it.

takluyver:
Merging milestoned as 4.0 - if anyone feels that it belongs in 3.1, feel free to point out as much.

PR #8120, opened 2015-03-23
Make it possible for msg.buffers in javascript to be either ArrayBuffers or views

jasongrout:
This makes working with message buffers in javascript a bit easier, since having views and having array buffers are both common cases.

### DIFF
--- a/IPython/html/static/services/kernels/serialize.js
+++ b/IPython/html/static/services/kernels/serialize.js
@@ -79,7 +79,8 @@ define([
         for (i = 0; i < msg.buffers.length; i++) {
             // msg.buffers elements could be either views or ArrayBuffers
             // buffers elements are ArrayBuffers
-            buffers.push(msg.buffers[i].buffer || msg.buffers[i])
+            var b = msg.buffers[i];
+            buffers.push(b.buffer instanceof ArrayBuffer ? b.buffer : b);
         }
         delete msg.buffers;
         var json_utf8 = (new TextEncoder('utf8')).encode(JSON.stringify(msg));
--- a/IPython/html/static/services/kernels/serialize.js
+++ b/IPython/html/static/services/kernels/serialize.js
@@ -75,15 +75,17 @@ define([
         msg = _.clone(msg);
         var offsets = [];
         var buffers = [];
-        msg.buffers.map(function (buf) {
-            buffers.push(buf);
-        });
+        var i;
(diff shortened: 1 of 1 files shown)


jasongrout:
@minrk or @jdfreder or @takluyver, if you have time to review this, that would be great.

IPython/html/static/services/kernels/serialize.js
```diff
@@ -75,15 +75,17 @@ define([
         msg = _.clone(msg);
         var offsets = [];
         var buffers = [];
-        msg.buffers.map(function (buf) {
```
minrk: Why not use msg.buffers.map? The logic should be the same, yes?

IPython/html/static/services/kernels/serialize.js
```diff
@@ -75,15 +75,17 @@ define([
...
-            buffers.push(buf);
-        });
+        var i;
+        for (i = 0; i < msg.buffers.length; i++) {
+            // msg.buffers elements could be either views or ArrayBuffers
+            // buffers elements are ArrayBuffers
+            buffers.push(msg.buffers[i].buffer || msg.buffers[i])
```
minrk: Is `.buffer` a good enough check? Or should it be an actual type check?

minrk:
Minor questions in-line, but the principle makes sense.

IPython/html/static/services/kernels/serialize.js
jasongrout: The logic is the same.  according to my trials on jsperf, `.map` is about 1/4 the speed of an explicit for loop, so I went ahead and changed it.

IPython/html/static/services/kernels/serialize.js
jasongrout: All views have a .buffer attribute, and an ArrayBuffer does not.  So it handles those cases nicely.  Not all browser support all views, so I don't want to put a hardcoded list in.  But I could do something like 
```python
b = msg.buffers[i];
buffers.push(b.buffer instanceof ArrayBuffer ? b.buffer : b);
```
if you're more comfortable with that.

jdfreder:
Looks good to me.

IPython/html/static/services/kernels/serialize.js
minrk: Makes sense. Just wanted to know the reason.

PR #8121, opened 2015-03-23
fix compress_user comparison

minrk:
move `.lower()` outside compress_user, otherwise it can prevent the substition, e.g. on OS X, where home is `/Users/[name]`


takluyver:
Debugged this in person with Min, makes complete sense.

PR #7675, opened 2015-02-03
Adding a very simple instance of _Bool widget

SylvainCorlay:
Just a _green check_ / _red cross_ widget typically used for showing whether some input is valid or not.

PR #8123, opened 2015-03-23
Add whatsnew for Valid widget.

jdfreder:
Whoops, looks like this was missing from #7675.  @sylvaincorlay can you check to make sure this is accurate?


SylvainCorlay:
Yes, this looks perfect.

PR #8126, opened 2015-03-23
TEST: Add test for handling of last_modified.

ssanderson:
Saving should increment last_modified.  Renaming should not.  Noticed
because incrementing last_modified on rename results in a scary error
message from the frontend.

### DIFF
--- a/IPython/html/services/contents/tests/test_manager.py
+++ b/IPython/html/services/contents/tests/test_manager.py
@@ -279,7 +279,10 @@ class TestContentsManager(TestCase):
         new_path = 'renamed.ipynb'
         cm.rename(path, new_path)
         renamed = cm.get(new_path)
-        self.assertEqual(renamed['last_modified'], saved['last_modified'])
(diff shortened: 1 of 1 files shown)


ssanderson:
The context for this is that the frontend fires a scary "notebook has changed since last read" warning if last_modified is updated by rename, which is what I was doing in pgcontents.

minrk:
I don't think we need to enforce this behavior on contents services. #8127 adds the missing update of the last_modified field from the `rename` reply. Can you check if that removes the warning for you?

ssanderson:
@minrk #8127 fixes the warning.  A less stringent version of this test might be to assert that saving/renaming leaves the date greater or equal than it was before, which should be sufficient to avoid the warning.  Do you think that would be worth keeping around?

minrk:
Yes, that's a good idea.

ssanderson:
Updated to assert greater or equal.

PR #8127, opened 2015-03-23
update last_modified on rename

minrk:
some storage backends may update the last_modified date on rename

alternative to #8126

/cc @ssanderson


takluyver:
If this is an alternative, and #8126 was merged, should this be closed?

ssanderson:
@takluyver #8126 was modified under the assumption that this would be merged as well.

takluyver:
I see. Merging, then.

Issue #8089, opened 2015-03-19
Run commands should be grouped together in help


abalkin:
In the command mode section, run commands are nicely grouped together:

![screen shot 2015-03-19 at 1 01 25 pm](https://cloud.githubusercontent.com/assets/535197/6735957/3e47f376-ce38-11e4-9f78-2ab41386880d.png)
 
but in the edit mode section they are not grouped and are not presented in the same order.

![screen shot 2015-03-19 at 1 00 42 pm](https://cloud.githubusercontent.com/assets/535197/6735971/5367ede2-ce38-11e4-9488-70de3b9a101b.png)

ChrisDubeau:
I am looking to help with this project and wanted to see if i could fix this issue. With that in mind what section of the project would i find the relevant files that would need to be worked on?

abalkin:
@ChrisDubeau - thanks for your interest in helping the project!

I have recently fixed some other issues in this area and if you look at gh-7809, you will see where the relevant code is.  The order of entries is controlled by the `help_index` attribute in  actions dictionary in IPython/html/static/notebook/js/actions.js.

takluyver:
They are together when I view them, so there must be something varying here.

![screenshot from 2015-03-23 17 10 15](https://cloud.githubusercontent.com/assets/327925/6793308/cf23e0e0-d17f-11e4-8e96-3c39bdebb958.png)

takluyver:
Aha, they're sorted together on Firefox, but not on Chromium.

PR #8128, opened 2015-03-23
Fix for Issue #8089

ChrisDubeau:
Keeps the Command Mode Sorted and Sorts the Edit Mode Hotkeys

### DIFF
--- a/IPython/html/static/base/js/keyboard.js
+++ b/IPython/html/static/base/js/keyboard.js
@@ -250,7 +250,10 @@ define([
             if (a.help_index < b.help_index){
                 return -1;
             }
-            return 0;
+			if (a.help_index = b.help_index){
+				return 0;
+			}
+            return -1;
         });
         return help;
     };


IPython/html/static/base/js/keyboard.js
```diff
@@ -250,7 +250,10 @@ define([
             if (a.help_index < b.help_index){
                 return -1;
             }
-            return 0;
+			if (a.help_index = b.help_index){
```
takluyver: This doesn't look right. We've already checked `a.help_index > b.help_index` and `a.help_index < b.help_index` above, if it's reached this point then they must be equal.

(Also `=` assigns the RHS to the LHS - you'd want `===` to check if they're the same)

ChrisDubeau:
original Here is the Original Codes output it shows the current help layout: 
with This shows the Help menu when i put === instead of = and lastly 
with This is the help menu with this change.

I'm not sure why its like this however.

takluyver:
Thanks for working on this @ChrisDubeau - I've merged #8131 instead because I understand how it resolves the problem.

PR #8131, opened 2015-03-24
handle undefined when sorting quick help

### LINKED ISSUE #8089: Run commands should be grouped together in help

minrk:
Since undefined is neither less than nor greater than anything in Javascript, the sort function was treating it as equal to everything, causing inconsistent behavior depending on the sort algorithm of the browser.

closes #8089

This ensures undefined elements are sorted last in the sequence.

### DIFF
--- a/IPython/html/static/base/js/keyboard.js
+++ b/IPython/html/static/base/js/keyboard.js
@@ -244,13 +244,13 @@ define([
             }
         }
         help.sort(function (a, b) {
-            if (a.help_index > b.help_index){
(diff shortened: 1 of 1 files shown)


takluyver:
This makes sense, and I've confirmed that it fixes the issue for me.

PR #8132, opened 2015-03-24
Only split lines on text output

minrk:
rather than all mime-types, which includes b64 output.

This doesn't change how files are read, so it is still valid for any mime-type data to be split on disk. It just changes which mime-types IPython.nbformat splits on write.

/cc @shoyer who reported the issue on twitter


Carreau:
Is it worth bumbing the fileformat minor ?

+1

takluyver:
As I understand it, the notebooks this will create can be read by 3.0 without missing any information, so I don't think it's necessary to bump the format.

minrk:
Since this doesn't change anything about the format, I don't think there's a reason to bump it. Both behaviors remain within the spec of nbformat 4, we are just changing how we apply some of the optional behavior.

shoyer:
Thanks guys!

shoyer:
Is there a timeframe on IPython 3.1 yet?

minrk:
@shoyer a week-ish

PR #8134, opened 2015-03-24
Try to fix some of ellisonbg focus bugs.

Carreau:
Do not manually focus, but make buttons divs so that by default they
don't get focus.

### DIFF
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -36,7 +36,7 @@ define(function(require){
      *
      **/
     var _actions = {
-        'run-select-next': {
+        'run-select-next-cell-and-focus': {
             icon: 'fa-play',
             help    : 'run cell, select below',
             help_index : 'ba',
@@ -78,7 +78,7 @@ define(function(require){
                 env.notebook.edit_mode();
             }
         },
-        'select-previous-cell' : {
+        'select-previous-cell-and-focus' : {
             help: 'select cell above',
             help_index : 'da',
             handler : function (env) {
@@ -89,7 +89,7 @@ define(function(require){
                 }
             }
         },
-        'select-next-cell' : {
+        'select-next-cell-and-focus' : {
             help: 'select cell below',
             help_index : 'db',
             handler : function (env) {
@@ -115,7 +115,6 @@ define(function(require){
             help_index : 'ef',
             handler : function (env) {
                 env.notebook.copy_cell();
-                env.notebook.focus_cell();
             }
         },
         'paste-cell-before' : {
@@ -268,7 +267,6 @@ define(function(require){
             help_index : 'ha',
             handler : function (env) {
                 env.notebook.kernel.interrupt();
-                env.notebook.focus_cell();
             }
         },
         'restart-kernel':{
@@ -276,7 +274,6 @@ define(function(require){
             help_index : 'hb',
             handler : function (env) {
                 env.notebook.restart_kernel();
-                env.notebook.focus_cell();
             }
         },
         'undo-last-cell-deletion' : {
@@ -381,7 +378,6 @@ define(function(require){
                 if(event){
                     event.preventDefault();
                 }
-                env.notebook.ensure_focused();
                 return false;
             }
         },
--- a/IPython/html/static/notebook/js/keyboardmanager.js
+++ b/IPython/html/static/notebook/js/keyboardmanager.js
@@ -64,7 +64,7 @@ define([
     KeyboardManager.prototype.get_default_common_shortcuts = function() {
         return {
             'shift'       : 'ipython.ignore',
-            'shift-enter' : 'ipython.run-select-next',
+            'shift-enter' : 'ipython.run-select-next-cell-and-focus',
             'ctrl-enter'  : 'ipython.execute-in-place',
             'alt-enter'   : 'ipython.execute-and-insert-after',
             // cmd on mac, ctrl otherwise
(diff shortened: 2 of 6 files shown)


takluyver:
Test failures look real.

Carreau:
Yeah, I'm trying to look at why, but not sure I get it. 

I'll try to focus on that this morning I was doing that on the side yesterday.

Carreau:
Faillure make no sens. It works locally, I rechecked twice and it **is** the right testing code 
Unless phantom does not have click event on div....

Carreau:
Ha ! `$($("div[data-jupyter-action='ipython.run-select-next-cell-and-focus']")[0]).click` is jq object/function but `$("div[data-jupyter-action='ipython.run-select-next-cell-and-focus']")[0].click` is `function(){[native code]}` I bet you that the second one is not a standard on divs.

Carreau:
YEahhh works ! just have to wait for travis to finish "all the tests"

rsmith31415:
> make buttons divs

Uhm, that's probably not the best idea (not very good semantics and definitely bad for accessibility). As I understand it, the problem is not that buttons get the focus after you click on them. Instead, the problem is deciding what element should keep the focus after user interaction (such as clicking a button).

Carreau:
Well, even button "getting" focus in the first place is problematic n the first place. And for accessibility, I doubt people should use button, as each button should have keyboard shortcut. 

we can try `role=button` for accesibility. would that help ?

Carreau:
> we can try role=button for accesibility. would that help ?

`role=button` keep the functionality, should I add it ?

Carreau:
pushed button-role.

rsmith31415:
I think I need to test that idea but this is what I found:

>Keyboard and focus

>Buttons are interactive controls, and therefore must be focusable. If the button role is added to an element that is not focusable by itself (such as `<span>`, `<div>` or `<p>`) the tabindex attribute will have to be used to make the button focusable.

>Buttons need to be operable by mouse users as well as keyboard users. For native HTML buttons, the button's onclick event will fire both for mouse clicks and when the space key is pressed while the button has focus. If role="button" is used to create a custom button, the onclick event will only fire when clicked by the mouse cursor. Because of this, the developer will have to explicitly add a separate key event handler to the element so that the button can be triggered when the space key is pressed.

[Source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role)

So an additional listener is needed to activate the button and a tabindex to navigate properly using the keyboard.

>Well, even button "getting" focus in the first place is problematic n the first place.

Oh.... why is that?

rsmith31415:
Uhm, wait... 

> the tabindex attribute will have to be used to make the button focusable.

Doesn't that defeat the purpose of not giving focus to "buttons"?

Carreau:
> Oh.... why is that?

Because codemirror will lose focus. And it triggers side effect of command mode, scolling, and all these annoying things. 

We won't add tabindex, as keyboard user can use keyboard shortcut. Keyboard user will not "navigate" to button with keyboard. Most of the action do have effect on "selected" cell. So trying to add accessibility to these buttons is a non-issue. Well, with more work, which is on it's way button could have also at title text the shortcut to trigger them.

rsmith31415:
>Because codemirror will lose focus. And it triggers side effect of command mode, scolling, and all these annoying things.

Of course, but you have already fixed that using ' get_selected_cell()`. I mean given the current behavior in master, why do you want to avoid focus on the buttons of the toolbar? Particularly considering that a few milliseconds later, the focus will be moved to the selected cell, as it should.

>We won't add tabindex, as keyboard user can use keyboard shortcut. Keyboard user will not "navigate" to button with keyboard.

Yes, but then you are not really addressing accessibility because screen readers navigate following the tabindex (or using the default browser behavior when no explicit tabindex is set) and obviously, shortcuts are not a solution in that case.

takluyver:
> Of course, but you have already fixed that using get_selected_cell(). I mean given the current behavior in master, why do you want to avoid focus on the buttons of the toolbar? Particularly considering that a few milliseconds later, the focus will be moved to the selected cell, as it should.

This leads to other bugs - I'm not sure if we have a written description of them, but @ellisonbg described them in the last dev meeting. If we have to refocus the current cell, it can lead to scroll jumps.

rsmith31415:
I think the "bug" that was referenced before is the following:

output

Every time the page jumps is because I was pressing tab to focus distinct elements of the page. Another example in LibreOffice similar to what we have (using tab and changing the font size):

output2

I don't think this is a bug. If you scroll down and leave your focus behind and then perform an action, the page will jump to show you the place in which the change took place. Otherwise, it would be confusing. The alternative would be to change focus as you scroll down, but that is not the standard behavior at all and would require a lot of events to keep track of the `innerWidth` and `innerHeight` of the `window` object.

Carreau:
@ellisonbg Ca, you try this one ?

ellisonbg:
Sure, looking at @minrk image confinement right now but will look here soon.

On Sun, Mar 29, 2015 at 1:43 PM, Matthias Bussonnier <

> @ellisonbg <https://github.com/ellisonbg> Ca, you try this one ?
>
> —
> <https://github.com/ipython/ipython/pull/8134#issuecomment-87473562>.
>



-- 
Brian E. Granger

ellisonbg:
The notebook is broken in this branch right now... same issue as the one I was seeing in @minrk s image branch. Sure the problem isn't in master?

Carreau:
can't you check directly the branche and not the paster commit ?

minrk:
As in the other branch, it's not the branch that's broken, it's probably cached files leftover in your checkout. Try `git clean -xfd` to remove everything not in the repo.

ellisonbg:
From a quick test:

* Simply entering command mode from edit mode still can trigger focus jump.
* Clicking on the buttons in the toolbar doesn't cause a jump - it does fix that issue.

But...I think buttons should be buttons, not divs. Being forced on relying on button=div to fix this type of bug is not sustainable in the long run (just the accesibility issues alone, but also for writing semantic HTML). Plus, this doesn't really fix the underlying issues of focus that we have (some of the bugs remain and are unrelated to the buttons).

ellisonbg:
Please see my recent comments on #8054 about these issues...

Carreau:
> But...I think buttons should be buttons, not divs.

Then I give up. You ask things that are opposite, and we don't have 10  people just to work on re-creating a scoll engine.

PR #8135, opened 2015-03-24
Split out the Qt console

takluyver:
The Big Split™  begins with the Qt console.

As we said, this moves it to a jupyter_qtconsole package, and adds a shim so `IPython.qt` imports still work, so long as `jupyter_qtconsole` is on your import path.

Tests will fail because jupyter_qtconsole isn't installed - I think we should just skip the tests for the bits we're in the process of splitting.


minrk:
qt tests still pass after a `pip install -e`. Nicely done.

minrk:
Since `pip install -e` puts the repo root on sys.path, these should continue to work. 'regular' (non-editable) installs won't work while we are doing this, but editable installs should.

minrk:
Speaking of which - do you want to do `pip install -e` in travis?

minrk:
Travis is green. Merging. It begins! :rocket:

PR #8137, opened 2015-03-24
remove profiles from IPython.config

minrk:
in preparation for The Big Split™

profile-aware bits are moved to IPython subclasses.

In addtion, the default profiles in `IPython.config.profile` are removed, and `get_config` is made obsolete (but left as-is) by injecting `c` into the config file namespace.


Carreau:
> Running setup.py install for ipython
>    /home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/setuptools/dist.py:283: UserWarning: >The version specified requires normalization, consider using '4.0.0.dev0' instead of '4.0.0-dev'.

Should we ?

On py2

```
ipython/setupbase.py", line 219, in check_package_data
        assert len(glob(path)) > 0, "No files match pattern %s" % path
    AssertionError: No files match pattern IPython/qt/console/resources/icon/*.svg
```

minrk:
@Carreau the warning is gone in current setuptools

Carreau:
+1

Carreau:
Let's break all the things. Merging.

PR #8139, opened 2015-03-25
Big Split: IPython.parallel -> ipython_parallel

minrk:
added some awareness in iptest of the shims, so that module exclusions are properly recognized. This may not be needed for anything other than parallel (qt doesn't need it, html shouldn't). It's only required when a subset of files under a top-level package should be excluded.


Carreau:
+1

PR #8142, opened 2015-03-25
Using the early validation hook for IntSlider and FloatSlider

SylvainCorlay:
@jdfreder @minrk @jasongrout indeed, during initial setting I must prevent the cross validation.

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -25,7 +25,7 @@ from IPython.core import release, crashhandler
 from IPython.core.profiledir import ProfileDir, ProfileDirError
 from IPython.utils.path import get_ipython_dir, get_ipython_package_dir, ensure_dir_exists
 from IPython.utils import py3compat
-from IPython.utils.traitlets import List, Unicode, Type, Bool, Dict, Set, Instance
+from IPython.utils.traitlets import List, Unicode, Type, Bool, Dict, Set, Instance, Undefined
 
 if os.name == 'nt':
     programdata = os.environ.get('PROGRAMDATA', None)
@@ -215,7 +215,7 @@ class BaseIPythonApplication(Application):
             return crashhandler.crash_handler_lite(etype, evalue, tb)
     
     def _ipython_dir_changed(self, name, old, new):
-        if old is not None:
+        if old is not None and old is not Undefined:
             str_old = py3compat.cast_bytes_py2(os.path.abspath(old),
                 sys.getfilesystemencoding()
             )
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -594,6 +594,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
         """
         if self._cross_validation_lock is True:
             yield
+            return
         else:
             self._cross_validation_lock = True
             cache = {}
@@ -628,8 +629,8 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                     notifications = {}
                     raise e
                 finally:
-                    self._cross_validation_lock = False
                     self._notify_trait = _notify_trait
+                    self._cross_validation_lock = False
                     if isinstance(_notify_trait, types.MethodType):
                         # FIXME: remove when support is bumped to 3.4.
                         # when original method is restored,
--- a/IPython/utils/traitlets.py
+++ b/IPython/utils/traitlets.py
@@ -319,7 +319,6 @@ class TraitType(object):
        accept superclasses for :class:`This` values.
     """
 
-
     metadata = {}
     default_value = Undefined
     allow_none = False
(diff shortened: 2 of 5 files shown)


minrk:
Can you verify that passing conflicting values on init still fails by adding a few tests?

```python
slider = IntSlider(min=1, max=0, value=-1)
```

SylvainCorlay:
Currently, validation does not occur on init.

minrk:
Okay, then I think this one should wait until you add delayed validation to the hold_notifications context (or however it makes the most sense to implement that)

SylvainCorlay:
ok

SylvainCorlay:
@minrk Here is a version of the context manager holding validation. Let me know what you think.

minrk:
I think it looks sensible, but it would appear to not be working just yet. Lots of test failures that look like trait values aren't being set.

SylvainCorlay:
This is ready for further review.

IPython/core/application.py
```diff
@@ -215,7 +215,7 @@ def excepthook(self, etype, evalue, tb):
             return crashhandler.crash_handler_lite(etype, evalue, tb)
     
     def _ipython_dir_changed(self, name, old, new):
-        if old is not None:
+        if old is not None and old is not Undefined:
```
minrk: Why can old be Undefined, now?

IPython/utils/traitlets.py
```diff
@@ -570,43 +572,73 @@ def __new__(cls, *args, **kw):
...
-            if isinstance(_notify_trait, types.MethodType):
-                # FIXME: remove when support is bumped to 3.4.
-                # when original method is restored,
-                # remove the redundant value from __dict__
-                # (only used to preserve pickleability on Python < 3.4)
-                self.__dict__.pop('_notify_trait', None)
-        # trigger delayed notifications
-        for args in notifications:
-            self._notify_trait(*args)
+        else:
```
minrk: Minor style comment (you don't need to change anything). When I have a case of

```python
if special_case:
    dont_do_much()
else:
    the real implementation of the method
```

I like to `return` from the special case block, so I don't have to indent the entirety of the method under an `else` block:

```python
if special_case:
    yield
    return
start
the
rest_of_the_method()
```

IPython/core/application.py
SylvainCorlay: In traitlets.py Line 449, I changed the value of `old` from `None` to `Undefined` in case of key error. 

The reason for that change, is that to roll back in case of bulk rejection, I need to distinguish between unset attributes and attributes set to None.

minrk: ok, makes sense. There may be other cases affected by this. We'll have to keep an eye out.

IPython/utils/traitlets.py
```diff
@@ -570,43 +572,73 @@ def __new__(cls, *args, **kw):
...
+            self._cross_validation_lock = True
+            cache = {}
+            notifications = {}
+            _notify_trait = self._notify_trait
+
+            def cache_values(*a):
+                cache[a[0]] = a
```
minrk: You never use a value other than `cache[name][1]` below, does it make sense to use 

```python
cache_values[a[0]] = a[1]
```
?

minrk: Or better yet, perhaps:

```python
def cache_values(name, old, new):
    cache_values[name] = old
```

IPython/utils/traitlets.py
```diff
@@ -570,43 +572,73 @@ def __new__(cls, *args, **kw):
...
+                        if hasattr(self, '_%s_validate' % name):
+                            cross_validate = getattr(self, '_%s_validate' % name)
+                            setattr(self, name, cross_validate(getattr(self, name), self))
+                except TraitError as e:
+                    self._notify_trait = lambda *x: None
+                    for name in cache:
+                        if cache[name][1] is not Undefined:
+                            setattr(self, name, cache[name][1])
+                        else:
+                            delattr(self, name)
```
minrk: Can you make sure to add a test for the rejected validation? I haven't seen delattr on a traitlet before, and I'd like to be sure it behaves as expected:

```python
c = Test()
with c.hold_trait_notifications():
    c.trait = invalid_value

assert c.trait == default_value
```

It might be that this should be a `self._trait_values.pop(name)`

IPython/utils/traitlets.py
SylvainCorlay: I cannot do that because in the case where 
 - there is no rejection and
 - the custom validation does not change the value,
I need to use the cached notifications:

This is why I eventually do the following:
```
for v in dict(cache, **notifications).values():
    self._notify_trait(*v)
```

IPython/utils/traitlets.py
SylvainCorlay: Ok, I will add a test.

IPython/utils/traitlets.py
minrk: Gotcha, I didn't see that one.

SylvainCorlay:
Added a test.

PR #8151, opened 2015-03-25
Included a way to 'comment out' an %s for a non-arg

brotherjack:
I had problems aliasing the following (because of an '%s' at the end as a mistaken arg): gitld git log --pretty=format:'%%C(yellow)%%h%%Cred%%ad%%Cgreen%%d%%reset%s'

I made this simple alteration to fix that.  I checked the guidelines and I'm not 100% sure if I was supposed to open an issue first(?), but this seems a small change.  This is my first (of hopefully many more) contribution to a project, so I apologize if my etiquette is off.

I tested the result with '%alias test echo %s' and it worked as expected.

### DIFF
--- a/IPython/core/alias.py
+++ b/IPython/core/alias.py
@@ -151,8 +151,8 @@ class Alias(object):
             raise InvalidAliasError("An alias command must be a string, "
                                     "got: %r" % self.cmd)
 
-        nargs = self.cmd.count('%s')
-
+        nargs = self.cmd.count('%s') - self.cmd.count('%%s')
+  
         if (nargs > 0) and (self.cmd.find('%l') >= 0):
             raise InvalidAliasError('The %s and %l specifiers are mutually '
                                     'exclusive in alias definitions.')
--- a/IPython/core/alias.py
+++ b/IPython/core/alias.py
@@ -169,6 +169,10 @@ class Alias(object):
         if cmd.find('%l') >= 0:
             cmd = cmd.replace('%l', rest)
             rest = ''
+         
+        if cmd.find('%%s') >= 1:
+            cmd = cmd.replace('%%s', '%s')
+            
         if nargs==0:
             # Simple, argument-less aliases
             cmd = '%s %s' % (cmd, rest)
--- a/IPython/core/alias.py
+++ b/IPython/core/alias.py
@@ -169,13 +169,12 @@ class Alias(object):
         if cmd.find('%l') >= 0:
             cmd = cmd.replace('%l', rest)
             rest = ''
-         
-        if cmd.find('%%s') >= 1:
-            cmd = cmd.replace('%%s', '%s')
-            
+		
         if nargs==0:
-            # Simple, argument-less aliases
-            cmd = '%s %s' % (cmd, rest)
+			if cmd.find('%%s') >= 1:
+				cmd = cmd.replace('%%s', '%s')
+			# Simple, argument-less aliases
+			cmd = '%s %s' % (cmd, rest)
         else:
             # Handle aliases with positional arguments
             args = rest.split(None, nargs)
--- a/IPython/core/alias.py
+++ b/IPython/core/alias.py
@@ -169,12 +169,12 @@ class Alias(object):
         if cmd.find('%l') >= 0:
             cmd = cmd.replace('%l', rest)
             rest = ''
-		
+        
         if nargs==0:
-			if cmd.find('%%s') >= 1:
-				cmd = cmd.replace('%%s', '%s')
-			# Simple, argument-less aliases
-			cmd = '%s %s' % (cmd, rest)
+            if cmd.find('%%s') >= 1:
+                cmd = cmd.replace('%%s', '%s')
+            # Simple, argument-less aliases
+            cmd = '%s %s' % (cmd, rest)
         else:
             # Handle aliases with positional arguments
             args = rest.split(None, nargs)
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, that looks good to me. Can you add a short test in `IPython.core.tests.test_alias`? Just create an alias with both %s and %%s in the args and check that it has the right nargs.

brotherjack:
Wrote the test you requested, and through said test, found that I had made a design mistake.  The '%%s' should have been converted to an '%s' so that a command like the one mentioned at the top of this thread would alias properly.  It didn't, so it wasn't.  So I made some changes, and wrote an additional test to ensure that the nargs were being calculated correctly.

takluyver:
Well spotted, but now there's a problem when you have a mixture of %s and %%s in an alias, because the %%s get turned into %s:

```
In [11]: alias comargcount echo this is %%s a "commented out" arg and this is not %s

In [12]: comargcount foo
<ipython-input-12-cd130fcdd1d3> in <module>()
----> 1 get_ipython().magic('comargcount foo')

/home/takluyver/.local/lib/python3.4/site-packages/IPython/core/interactiveshell.py in magic(self, arg_s)
   2306         magic_name, _, magic_arg_s = arg_s.partition(' ')
   2307         magic_name = magic_name.lstrip(prefilter.ESC_MAGIC)
-> 2308         return self.run_line_magic(magic_name, magic_arg_s)
   2309 
   2310     #-------------------------------------------------------------------------

/home/takluyver/.local/lib/python3.4/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line)
   2227                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2228             with self.builtin_trap:
-> 2229                 result = fn(*args,**kwargs)
   2230             return result
   2231 

/home/takluyver/.local/lib/python3.4/site-packages/IPython/core/alias.py in __call__(self, rest)
    183                 raise UsageError('Alias <%s> requires %s arguments, %s given.' %
    184                       (self.name, nargs, len(args)))
--> 185             cmd = '%s %s' % (cmd % tuple(args[:nargs]),' '.join(args[nargs:]))
    186 
    187         self.shell.system(cmd)

TypeError: not enough arguments for format string
```

IPython/core/alias.py
```diff
@@ -169,6 +169,10 @@ def __call__(self, rest=''):
         if cmd.find('%l') >= 0:
             cmd = cmd.replace('%l', rest)
             rest = ''
+         
+        if cmd.find('%%s') >= 1:
+            cmd = cmd.replace('%%s', '%s')
```
takluyver: I think you need to move this check inside the `if nargs==0:` block immediately below, so that it only happens if we're not formatting arguments into the command.

IPython/core/tests/test_alias.py
```diff
@@ -38,4 +38,25 @@ def test_alias_args_error():
...
+    nt.assert_equal(cap.stderr.split(':')[0], 'UsageError')
+    
+def test_alias_args_commented():
+    """Check that alias correctly ignores 'commented out' args"""
+    _ip.magic('alias commetarg echo this is %%s a "commented out" arg')
+    
+    with capture_output() as cap:
+        _ip.run_cell('commetarg')
+    
+    nt.assert_equal(cap.stdout, 'this is %s a "commented out" arg')
```
takluyver: This is failing, because the quotes disappear at some point. It's probably best just to leave the quotes out of the test case altogether.

```
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/ipython/ipython/IPython/core/tests/test_alias.py", line 50, in test_alias_args_commented
  nt.assert_equal(cap.stdout, 'this is %s a "commented out" arg')
nose.proxy.AssertionError: 'this is %s a commented out arg' != 'this is %s a "commented out" arg'
- this is %s a commented out arg
+ this is %s a "commented out" arg
```

brotherjack:
I  think things are OK now?  Maybe?

Carreau:
Travis seem happy, will read the code.

IPython/core/tests/test_alias.py
```diff
@@ -38,4 +38,25 @@ def test_alias_args_error():
...
-    nt.assert_equal(cap.stderr.split(':')[0], 'UsageError')
\ No newline at end of file
+    nt.assert_equal(cap.stderr.split(':')[0], 'UsageError')
+    
+def test_alias_args_commented():
+    """Check that alias correctly ignores 'commented out' args"""
+    _ip.magic('alias commetarg echo this is %%s a commented out arg')
+    
+    with capture_output() as cap:
+        _ip.run_cell('commetarg')
```
Carreau: Did you ment `comme**n**targ` here and above ? (actually we don't care as it is a test you could have called it `rfgtrederfgghtui` ...

Carreau:
+1

takluyver:
Looks good, thanks @brotherjack !

IPython/core/tests/test_alias.py
brotherjack: Yeah, it is an abbreviation of "comment met (Dutch for "with") argument".  There is no rational reason why I didn't just call it 'x'.  It would have been a lot easier, and as you pointed out, it's a test so it doesn't matter that it isn't a "meaningful variable name".  I too often have no idea why I do the stupid things I do.

Carreau: No problem, we all do stupid things without reasons :-)

PR #8156, opened 2015-03-26
Remove color alias from Qt console

takluyver:
This was causing test failures on Windows


ccordoba12:
It seems the title is wrong. Shouldn't it be Remove *colors* alias? Sorry for bring picky but I thought this one referred to something else.

Carreau:
Fix title. Relaunch failed test.

takluyver:
It wasn't a mistake, just confusing brevity - I meant that this removes an alias for a kernel config option, which no longer works following #6936.

PR #8158, opened 2015-03-26
update to component (CodeMirror 4.13)

Carreau:


Carreau:
Tested a bit. don't see any bug.

minrk:
4.0?

Carreau:
50/50. 

I don't see any reason to make it 3.1 nor any reason not to make it 3.1

minrk:
Are there known bugs fixed by this? I think updating codemirror in 3.x is fine, but since it's been known to introduce regressions in the past, I'd only backport it if we know it fixes bugs that affect IPython users.

Carreau:
Had a quicklook at changes, and none seem noticible. Maybe a few Scroll issues, but not sure they affect us. Marking 4.0

Issue #7985, opened 2015-03-06
Allow text on toolbar buttons


takluyver:
Toolbar buttons are currently icon-only, which gets confusing as extensions add buttons to the toolbar and try to use the limited set of icons available in Font Awesome.

The add_buttons_group() API should take an additional optional `button_text` key for a very short (usually one word) string to show on the button. icon-only, text-only and icon+text buttons should be allowed.

Carreau:
Closing as this issue is not in IPython itself and if still problematic and relevant should be opened on the right repository.  This will allow to keep the number of opened issue on the IPython repo under control. 

Feel free to keep commenting or reopen if necessary.

Thanks.

PR #8159, opened 2015-03-26
Fix action do not provide icon.

Carreau:
action might be undefined then button will get no icon.
Prevent throwing an uncatched error and insert warning sign on button

Will do better one I allow text instead of icon


minrk:
I don't quite understand the comment. Is this PR ready to merge, or are you planning to add more to it for when there is text instead of an icon? 3.1?

Carreau:
> I don't quite understand the comment.

We can merge. 
I'll refactor that in something cleaner when I try to tackle #7985 
But the fix of #7985 will not go into 3.1. This one should.

minrk:
Gotcha, thanks.

PR #8164, opened 2015-03-26
replace undefined by empty string to get aroung bug in ie

### LINKED ISSUE #8161: undefined windows
in IE, new/duplicate open undefined windows that refresh.

Carreau:
Could close #8161 ...need to test.

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -273,7 +273,7 @@ define([
 
     KernelSelector.prototype.new_notebook = function (kernel_name) {
         
-        var w = window.open(undefined, IPython._target);
+        var w = window.open('', IPython._target);
         // Create a new notebook in the same path as the current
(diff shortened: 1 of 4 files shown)


Carreau:
Works locally.

minrk:
Missing one in tree/js/notebooklist

Carreau:
Done.

minrk:
Verified fix with IE 11, and still fine on Chrome. Will merge once green light confirms.

takluyver:
I thought we heard yesterday that it was fine on IE11 even without this fix, and it was IE 10 that was the problem. I'll see if I have IE10 on my Windows 7 VM.

takluyver:
Nope, I have IE11 as well. I think it's fine to assume that this fix will work and merge away - it doesn't seem to break anything, at least.

minrk:
I could see the 404 page on my IE in master and not in this PR, so it definitely fixes it for at least one version of IE.

PR #8167, opened 2015-03-27
Load schemas from jupyter_nbformat package

takluyver:
Should fix some errors we have been seeing on Python 2 tests.


Carreau:
+1

takluyver:
I couldn't reproduce the errors locally, so I'm not entirely sure if this will fix it, but I suspect so.

PR #8168, opened 2015-03-27
BUG: Don't allow ContentsManager to delete root.

ssanderson:
Doing so leaves the notebook app in a bad state from which it can't recover.


Carreau:
Seem legit. Might need a `no_preserve_root` kwarg though :-D

PR #8172, opened 2015-03-27
Added PageUp and PageDown functionality for the notebook

AtrSykes:
PageUp and PageDown now move between cells rather than stay confined
to their cell

### DIFF
--- a/IPython/html/static/notebook/js/keyboardmanager.js
+++ b/IPython/html/static/notebook/js/keyboardmanager.js
@@ -77,7 +77,9 @@ define([
             'esc'                 : 'ipython.go-to-command-mode',
             'ctrl-m'              : 'ipython.go-to-command-mode',
             'up'                  : 'ipython.move-cursor-up-or-previous-cell',
+            'pageup'              : 'ipython.move-cursor-up-or-previous-cell',
             'down'                : 'ipython.move-cursor-down-or-next-cell',
+            'pagedown'            : 'ipython.move-cursor-down-or-next-cell',
             'ctrl-shift--'        : 'ipython.split-cell-at-cursor',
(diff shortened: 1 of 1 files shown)


Carreau:
Hum, the behavior seem weird to me, I would expect page-up to:
  - bring me to top of cell
  - if already at top of cell move to last line of previous one. 

I'm not sure why it does not do that, and if it's expected behavior.

takluyver:
I think I would have expected it to jump directly to the previous cell (or next, for page down), regardless of where I was in the cell.

Carreau:
Then that's not either what it does. If the cell is more than one screen, it jumps in current CM by one screen unit.

takluyver:
I guess that could be useful if you had particularly long cells.

ellisonbg:
Page up and page down keystrokes should do the same thing as spacebar and shift-spacebar, which use our scroll manager to scroll the notebook div.

Carreau:
No because here we speak of edit mode. PageUp and Down should not insert
space.

On Sun, Mar 29, 2015 at 11:36 AM, Brian E. Granger <notifications@github.com

ellisonbg:
Ahh, sorry I missed that. If this only applies to edit mode, then I would think that the notebook div would:

1. ??? - if the current cell is smaller that one scroll screen.
2. Move one CM scroll unit if that exists in the current cell in that desired scroll direction.

Because of the awkwardness of decided what to do in the 1 case, and the complexity of the logic required for 2 I am not sure this makes sense. I am probably -0.5 on this.

Carreau:
This does not merge cleany anymore, and have lost activity for almost 3 month. 
Feel free to reopen if you plan to work on it again.

PR #8166, opened 2015-03-27
Big Split: IPython.kernel

minrk:
There's still some work to do detangling a few of the files (connect, launcher, kernelspec), which have code destined for both packages. But this gets the bulk of the split done, and smaller reorg can happen as more easily reviewed smaller PRs.

PR #8175, opened 2015-03-27
Big Split: IPython.kernel

minrk:
There's still some work to do detangling a few of the files (connect, launcher, kernelspec), which have code destined for both packages. But this gets the bulk of the split done, and smaller reorg can happen as more easily reviewed smaller PRs.

This is a reissue of #8166

Let's wait until <del>#7757 and</del> #8107 is merged to avoid conflicts.


minrk:
Okay, #8107 is in, and this is rebased. Should be good to go, once Travis gives the green light. Sorry @Carreau for not marking the last one as waiting on another PR.

PR #8176, opened 2015-03-27
Fixes bug that would show "null" as the reason in the don't leave dialog

gilbertw:
IE treats null as a string and displays it.  Instead simply return.  Fix verified in Chrome, FireFox, and IE 10.


minrk:
@gilbertw this fix is fine, but you should probably be aware that IE is officially unsupported by IPython. Nontrivial fixes for IE compatibility will probably not be accepted.

PR #8178, opened 2015-03-28
Fix for CVE-2014-3566 'POODLE' SSL attack

briandrawert:
original commit '22c4922f4796078c5613de9e07e66b8764549cad'

### DIFF
--- a/IPython/html/notebookapp.py
+++ b/IPython/html/notebookapp.py
@@ -19,6 +19,7 @@ import re
 import select
 import signal
 import socket
+import ssl
 import sys
 import threading
 import webbrowser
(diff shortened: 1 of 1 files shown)


IPython/html/notebookapp.py
```diff
@@ -832,6 +833,8 @@ def init_webapp(self):
                 ssl_options['keyfile'] = self.keyfile
         else:
             ssl_options = None
+        # Disable SSLv3, since its use is discouraged.
+        ssl_options['ssl_version']=ssl.PROTOCOL_TLSv1
```
minrk: This will fail when ssl_options is None. It should be moved into the `if` block above.

minrk:
The 'merged with master' commit shouldn't exist. Can you remove it?

briandrawert:
The commit was removed, but now the PR has conflicts.  This is because I was developing off of the 3.0.0 tag, not HEAD.  The 'merged with master' commit was the result of pulling from master and resolving the conflicts.

minrk:
Ah, in that case, can you rebase on master?

briandrawert:
This is deprecated by: https://github.com/ipython/ipython/pull/8179

minrk:
You don't need to open a new PR for a rebase, but that's fine. #8179 has been merged and backported. Thanks!

Issue #8104, opened 2015-03-21
Usability Improvement: Make tool buttons at top right still visible when scrolling down


pybokeh:
In the "Files" view, can we "freeze" the top portion of the view to not only include the jupyter logo, but the tool buttons over at the top right so that when a user scrolls down, those buttons are still visible.  Would be nice after selecting file(s) near the bottom of the directory list, not having to scroll back up to click on the buttons.  Thanks!

PR #8181, opened 2015-03-28
Sticky Menu on notebook list

ardhipoetra:
Based on issue #8104, I made a hack to achieve this solution. Now, when you scroll the notebook list, the menu header (delete, duplicate, rename, etc) will be kept at the top of the list, but still below the jupyter header.


Carreau:
Isn't it what bootstrap `affix` is for ?

ellisonbg:
While I completely agree that we need to address the underlying issue, I don't think we should move forward with this. Right after the big split, we are going to completely refactor the JS code and re-do lots of our UI. I think this change will only complicate that work (which will independently fix this issue).

Carreau:
Which we have been saying for 2 releases now.

If there is a "simple" fix, that would greatly help usability. It's not the
only fix we had on master that was before a full rewrite. I would expect
using bootstrap affix to be much less hackish. Woudl be nice for chrome to
support postion: sticky.

On Sun, Mar 29, 2015 at 11:44 AM, Brian E. Granger <notifications@github.com

ellisonbg:
I should clarify this...I have *long* stopped saying "let's not fix the dashboard because we are about to do a big refactor." We have been doing incremental improvements to the dashboard for a while now. I remain +1 on that (after 4.0 and big split are done).

@jdfreder already looked at fixing the Files tab scroll issue for the 3.0 release. At the time, we were still scrolling the entire page. With the whole page scroll, it was a complete mess to fix this for the Files tab while leaving the Running and Cluster tab the same. Now that we are back to scrolling just the site div, it may be easier - I had forgotten that the page doesn't scroll any longer.

Let's see if some variant of the affix CSS class can get this to work.

rsmith31415:
A general comment: As a hack, this is okay. However, I wouldn't use Javascript to fix things that are supposed to be done in pure CSS.

@Carreau 

When I was writing the PR for a fixed header, I noticed that `position: sticky;` doesn't really work as a replacement of `position: fixed;` at least without a careful consideration of the layout.  `sticky` has the effect of hiding the header when the parent no longer fits the page. Since the browser support is not good at all, I didn't look deep enough to make it work. However, let's say it doesn't work "out of the box".

ardhipoetra:
Yes indeed, it could be solved using bootstrap `affix` function. I already tried that but somewhat it didn't work. So I came up with this hackish solution instead.

Weirdly, when I put `data-spy=affix` in `notebook_toolbar`, it doesn't affect anything. Does anyone have a clue about this?

Carreau:
we might no be loading bootstrap affix.

ellisonbg:
Here are some comments from our discussion at the weekly devmeeting:

* We do want to fix this behavior
* We feel that the implementation in this PR is too hackish
* When we moving to phosphor all of this will be handled differently, but...
* We are open to a clean implementation of this for 4.0 (pre-phosphor).

Our thoughts on a clean implementation:

* The entire dashboard page wouldn't scroll, just the dev that has the list in each of the 3 tabs.
* The affixed content will be outside that scrolling div so you get the desired behavior for free.
* We may need to refactor the main page template to get this behavior without affecting the other pages (terminal, text editor, notebook).

minrk:
The UI tour seems to cause some peculiar scroll behavior that I don't think is normal.

minrk:
This work can resume on the new jupyter/notebook repo, if it should be picked up. I suspect that living with the existing behavior until layout is done with phosphor will be the way to go, though.

PR #8186, opened 2015-03-29
add import hook to shims

minrk:
This ensures that submodule imports return the real target module, not a clone that will confuse `is` and `isinstance` checks.


minrk:
cc: @takluyver 

I also removed the reimplementation of import_item, since it appears that using import_item works properly, now. Not sure what it was that caused that to fail, but it doesn't seem to matter anymore.

takluyver:
Seems sensible, tests passing.

PR #8188, opened 2015-03-29
Big Split: traitlets, config

minrk:
This depends on #8186

There are a few shenanigans in here to avoid circular import problems that shouldn't be needed once genutils has been split out.


minrk:
I think this just leaves html as the remaining conventional split, which we should hold until we've figured out the various UI things we are backporting to 3.1, and then genutils, the gross, tricky one.

takluyver:
When I [surveyed which pieces use which utils](https://ipython.hackpad.com/Untitled-ZUxYtQKUHJj), I found that traitlets is the only user of `IPython.utils.eventful`. Should that be part of traitlets? If we were doing the copy of traitlets for widgets, I might even make it part of widgets, but I think we decided to continue with a single traitlets package for now.

minrk:
There are lots of utils that should move to the split packages. I haven't done any of that, yet. I was going to wait until each package is already separate.

I would be inclined to move the Eventful classes into widgets.trait_types, rather than keeping them in the base traitlets.

takluyver:
Sure, putting them in widgets sounds good, and I think it's fine to leave that for later. In which case, I think this is looking fine.

PR #8190, opened 2015-03-29
update description of sticky tooltip

### LINKED ISSUE #7940: Behavior of feature "tooltip is not dismissed while typing for 10 seconds"
When I press Shift + Tab three times to trigger the tooltip with the timer ("tooltip is not dismissed while typing for 10 seconds"), isn't it supposed to hide after 10 seconds even if I don't type anything? I remember that in previous versions of the notebook, the tooltip was closed automatically after 10 seconds. Now (in Ipython 3.x) it remains open until the user types something, which is a bit counter-intuitive given the description of the feature.

If this is the intended behavior, maybe the description should be simply "tooltip is not dismissed for 10 seconds".

minrk:
> Tooltip will linger for 10 seconds while you type

closes #7940

For 4.0, I think we should remove this state altogether.

### DIFF
--- a/IPython/html/static/notebook/js/tooltip.js
+++ b/IPython/html/static/notebook/js/tooltip.js
@@ -65,7 +65,7 @@ define([
         this._clocklink = $('<a/>').attr('href', "#");
         this._clocklink.attr('role', "button");
         this._clocklink.addClass('ui-button');
-        this._clocklink.attr('title', 'Tooltip is not dismissed while typing for 10 seconds');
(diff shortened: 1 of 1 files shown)


takluyver:
In the state before this, it's dismissed as soon as you start typing, so I actually prefer the wording before this PR. We can probably do better, though. "Tooltip will linger for 10 seconds while you type" ?

minrk:
Message updated

takluyver:
+1, but I'll let someone else merge, since it's now effectively my change ;-)

PR #8191, opened 2015-03-29
small focus/keyboard-related fixes for 3.1

minrk:
- don't disable keyboard shortcuts on main toolbar
- remove some extraneous focus_cell calls from actions.js

This restores most of the toolbar behavior of 2.x.

@ellisonbg can you enumerate some of the remaining focus problems that are still problematic in this PR? I'm trying to keep this reasonable for 3.1.

### DIFF
--- a/IPython/html/static/notebook/js/notebook.js
+++ b/IPython/html/static/notebook/js/notebook.js
@@ -663,10 +663,9 @@ define(function (require) {
     Notebook.prototype.command_mode = function () {
         var cell = this.get_cell(this.get_edit_index());
         if (cell && this.mode !== 'command') {
-            // We don't call cell.command_mode, but rather call cell.focus_cell()
-            // which will blur and CM editor and trigger the call to
-            // handle_command_mode.
-            cell.focus_cell();
+            // We don't call cell.command_mode, but rather blur the CM editor
+            // which will trigger the call to handle_command_mode.
+            cell.code_mirror.getInputField().blur();
         }
     };
 
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -107,7 +107,6 @@ define(function(require){
                 var index = env.notebook.get_selected_index();
                 env.notebook.cut_cell();
                 env.notebook.select(index);
-                env.notebook.focus_cell();
             }
         },
         'copy-selected-cell' : {
@@ -115,7 +114,6 @@ define(function(require){
             help_index : 'ef',
             handler : function (env) {
                 env.notebook.copy_cell();
-                env.notebook.focus_cell();
             }
         },
         'paste-cell-before' : {
@@ -268,7 +266,6 @@ define(function(require){
             help_index : 'ha',
             handler : function (env) {
                 env.notebook.kernel.interrupt();
-                env.notebook.focus_cell();
             }
         },
         'restart-kernel':{
@@ -276,7 +273,6 @@ define(function(require){
             help_index : 'hb',
             handler : function (env) {
                 env.notebook.restart_kernel();
-                env.notebook.focus_cell();
             }
         },
         'undo-last-cell-deletion' : {
--- a/IPython/html/static/notebook/js/actions.js
+++ b/IPython/html/static/notebook/js/actions.js
@@ -377,7 +377,6 @@ define(function(require){
                 if(event){
                     event.preventDefault();
                 }
-                env.notebook.ensure_focused();
                 return false;
             }
         },
--- a/IPython/html/static/notebook/js/maintoolbar.js
+++ b/IPython/html/static/notebook/js/maintoolbar.js
@@ -25,7 +25,6 @@ define([
         this.events = options.events;
         this.notebook = options.notebook;
         this._make();
-        this.notebook.keyboard_manager.register_events(this.element);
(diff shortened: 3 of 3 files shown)


ellisonbg:
This does fix some behaviors, but there are still some others that remain:

* Clicking on save in the toolbar causes a jump because of the `ensure_focus` logic that appears in a few places. Let's try removing that stuff as well (it is super simple and only in a few places).
* Hitting `esc` to enter command mode can cause a jump if the cell is large and not scrolled into the optimal position. The code that does this is carefully commented by me, but I thought that code behaved differently than it does now. It could be a browser change or my memory. @jdfreder do you remember how this behaved in 2.0?
* There are still 2 calls to `focus_cell` in `maintoolbar.js` to manage the cell type and celltoolbar select menues. I am not sure how these will behave after we re-add those to the register_events...

ellisonbg:
Oh and let's do add the cell type and celltoolbar select to register_events rather than the whole toolbar.

minrk:
>  let's do add the cell type and celltoolbar select to register_events 

done

> Clicking on save in the toolbar causes a jump 

fixed

> There are still 2 calls to focus_cell in maintoolbar.js

I think these should stay. The behavior when focus is not sent back to the notebook when a cell is changed seems surprising, because you are operating on an element that isn't focused, and you can't get back to the notebook without clicking somewhere after making the change.

2.x doesn't have register_events and doesn't focus the cell, which means the behavior is mostly the same as register_events + focus, except the select element remains highlighted, even though it won't actually receive keyboard events. I think register_events + focus is an improvement over the 2.x behavior.

> Hitting esc to enter command mode can cause a jump

From testing, it appears that 2.x has the same behavior on this as 3.0, so I don't think it's a regression.

Reading the comment, we are calling `focus_cell()` in order to trigger CM.blur, which in turn triggers command mode. I've updated the PR to blur CM directly, which seems to work, and avoids the focus call. This is the change that makes me most squeamish about backporting.

takluyver:
This resolves my issue.

rsmith31415:
Very nice. I like this approach of selecting which buttons should return focus to the cell.

A minor issue. When I click on "Interrupt kernel", "copy selected cell" and "cut selected cell", the button is maintaining an active state after the click, although the focus doesn't seem to stay in the button:

After clicking:

screenshot from 2015-03-30 15 33 22

ellisonbg:
Great I will test this now...

On Mon, Mar 30, 2015 at 2:38 PM, rsmith31415 <notifications@github.com>
wrote:

> Very nice. I like this approach of selecting which buttons should return
> focus to the cell.
>
> A minor issue. When I click on "Interrupt kernel", "copy selected cell"
> and "cut selected cell", the button is maintaining an active state after
> the click, although the focus doesn't seem to stay in the button:
>
> After clicking:
>
> [image: screenshot from 2015-03-30 15 33 22]
> <https://cloud.githubusercontent.com/assets/599274/6907344/9dc7b2ba-d6f2-11e4-9eaf-f77c7ef7dc34.png>
>
> —
> <https://github.com/ipython/ipython/pull/8191#issuecomment-87839441>.
>



-- 
Brian E. Granger

IPython/html/static/notebook/js/notebook.js
```diff
@@ -663,10 +663,9 @@ define(function (require) {
     Notebook.prototype.command_mode = function () {
         var cell = this.get_cell(this.get_edit_index());
         if (cell && this.mode !== 'command') {
-            // We don't call cell.command_mode, but rather call cell.focus_cell()
-            // which will blur and CM editor and trigger the call to
-            // handle_command_mode.
-            cell.focus_cell();
+            // We don't call cell.command_mode, but rather blur the CM editor
+            // which will trigger the call to handle_command_mode.
+            cell.code_mirror.getInputField().blur();
```
ellisonbg: I am also pretty wary about this change from calling focus_cell to directly bluring the code mirror input field. I seem to remember that there were very subtle bugs that we found when we called blur directly. I probably prefer to revert it this close to 3.1. Longer term I think it would be a good idea to play around with this approach.

ellisonbg:
@minrk is there a reason you didn't remove the implementation of `ensure_focused` methods is `Cell` and `Notebook`? It only existed for this call and I don't want it sitting around waiting to be called.

IPython/html/static/notebook/js/notebook.js
ellisonbg: @jdfreder does your more youthful mind (than mine) remember anything more about what problems were caused by the direct blur approach?

minrk:
> is there a reason you didn't remove the implementation of ensure_focused

No, though I don't feel a strong need to remove public methods in a bugfix release.

IPython/html/static/notebook/js/notebook.js
minrk: Since it is a one line change, we can always do a 3.1.1 reverting it if we notice a problem. I'm also fine reverting it entirely, and saving it for 4.0.

jdfreder: @ellisonbg I don't , but there was a problem at the time.  It may be something that was fixed, I remember CodeMirror had multiple focus related bugs at the time, some of which cause bugs in IPython (like the codemirror blur/focus event firing order).

PR #8194, opened 2015-03-30
Provide the notebook being imported with "get_ipython"

toobaz:
The example code for importing notebooks will fail if the notebook being imported contains some ipython magic. Passing "get_ipython" in the dict to "exec" solves this (at least, it allows "%load_ext" to work).


takluyver:
Seems reasonable, thanks.

PR #8196, opened 2015-03-30
DEV: Allow supplying jinja vars from python config.

ssanderson:
This PR allows the user to specify a dictionary of values to `NotebookApp.jinja_template_vars` whose values are made available in the jinja templating environment.  This is primarily useful in conjunction with the `get_body_data` javascript function to allow programmatic access to Python-specified values from Javascript.  An example of where this is useful is when you want to use different constants in Javascript based on a deployment environment; this allows you to write code in your ipython_notebook_config that looks like:

```
env = os.environ['DEPLOYMENT_ENV']
if env == 'DEV':
    template_vars = dev_template_vars
elif env == 'STAGING':
    template_vars = staging_template_vars
else:
    template_vars = prod_template_vars
c.NotebookApp.jinja_template_vars = template_vars
```

### DIFF
--- a/IPython/html/base/handlers.py
+++ b/IPython/html/base/handlers.py
@@ -119,6 +119,11 @@ class IPythonHandler(AuthenticatedHandler):
             return Application.instance().log
         else:
             return app_log
+
+    @property
+    def jinja_template_vars(self):
+        """User-supplied values to supply to jinja templates."""
+        return self.settings.get('jinja_template_vars', {})
     
     #---------------------------------------------------------------
     # URLs
@@ -250,6 +255,7 @@ class IPythonHandler(AuthenticatedHandler):
             sys_info=sys_info,
             contents_js_source=self.contents_js_source,
             version_hash=self.version_hash,
+            **self.jinja_template_vars
         )
     
     def get_json_body(self):
--- a/IPython/html/notebookapp.py
+++ b/IPython/html/notebookapp.py
@@ -200,6 +200,7 @@ class NotebookWebApplication(web.Application):
             config_manager=config_manager,
 
             # IPython stuff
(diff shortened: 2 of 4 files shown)


Carreau:
if that helps, +1

IPython/html/base/handlers.py
```diff
@@ -250,6 +255,7 @@ def template_namespace(self):
             sys_info=sys_info,
             contents_js_source=self.contents_js_source,
             version_hash=self.version_hash,
+            **self.jinja_template_vars
```
minrk: If you use `ns.update(self.jinja_template_vars)` instead of `**vars`, you will be able to override existing values. Do you think we should allow that?

minrk: @ssanderson did you have an opinion on this? If you don't think there's a use for overrides, this can go in as-is.

ssanderson: Sorry, I missed your first note. Looking at the list here, I could imagine wanting to override the various URLs, contents_js_source, and version_hash, but I'm inclined to say that I'd want to override them explicitly as separate configurables rather than all at once with this mechanism.  Are any of those attributes not yet configurable themselves via alternative means?

ssanderson:
Any chance of a backport to 3.x on this?  We're planning on targeting 3.x until the Big Split™ is done with.  I'm happy to make an extra PR against the 3.x branch if that's all it takes.

minrk:
Since we cut 3.1 on Friday, I'm not sure about 3.x. There's another PR with a similar goal, so it might be a good idea to allow that.

IPython/html/base/handlers.py
minrk: version_hash and sys_info are not configurable.

PR #8203, opened 2015-03-30
autosave & reload on kernelspec switch

### LINKED ISSUE #7783: kernels.js warning.
Not sure if we did, but I don't think we flag the kernels that have a kernel.js and prevent further switching.
We said we should.

Marking as 3.1 but should at least be in docs.

minrk:
when a custom kernel.js was defined, in which case reload is needed to ensure clean js state.

closes #7783


takluyver:
:+1:

PR #8204, opened 2015-03-30
Further split some client/kernel functionality

minrk:
connect.py had some kernel-specific functions and some general Juyter functions

The consoleapp base class is also moved to jupyter_client


takluyver:
There seems to be quite a bit more deleted than added here - did you duplicate some code while splitting client & kernel, and this is paring it back?

minrk:
Yup, there were two unmodified copies of connect.py. I removed client bits from the one in kernel, and kernel bits from the one in client.

PR #8208, opened 2015-03-31
Catch termios.error in pager

vbraun:
This error is raised by termios.tcgetattr on really old Linux versions
(here: Ubuntu 8.04.4) in doctests, that is, without a real stdout.

See http://trac.sagemath.org/ticket/17996#comment:11

### DIFF
--- a/IPython/core/page.py
+++ b/IPython/core/page.py
@@ -100,7 +100,10 @@ def _detect_screen_size(screen_lines_def):
     # flags each time), we just save the initial terminal state and
     # unconditionally reset it every time.  It's cheaper than making
     # the checks.
-    term_flags = termios.tcgetattr(sys.stdout)
+    try:
+        term_flags = termios.tcgetattr(sys.stdout)
+    except termios.error as err:
+        raise TypeError
 
     # Curses modifies the stdout buffer size by default, which messes
     # up Python's normal stdout buffering.  This would manifest itself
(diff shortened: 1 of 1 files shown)


vbraun:
Don't know why the tests fail in some javascript stuff, I didn't touch any of that.

IPython/core/page.py
```diff
@@ -100,7 +100,10 @@ def _detect_screen_size(screen_lines_def):
     # flags each time), we just save the initial terminal state and
     # unconditionally reset it every time.  It's cheaper than making
     # the checks.
-    term_flags = termios.tcgetattr(sys.stdout)
+    try:
+        term_flags = termios.tcgetattr(sys.stdout)
+    except termios.error as err:
+        raise TypeError
```
takluyver: Why turn this into a different exception that the caller will catch, instead of adding `termios.error` to what the caller catches?

takluyver:
I don't mind this simple change, but note that we're not likely to accept more significant changes to support Ubuntu 8.04, which is well out of support now. Even 10.04 will be unsupported in another month.

JS tests fail sporadically for no good reason; I restarted it and it's passed now.

vbraun:
Termios is only conditionally imported in `_detect_screen_size`, if you want to catch `termios.error` elsewhere then you'd also have to provide a stub for the exception etc.

takluyver:
Ah, that makes sense. Could the TypeError have at least a brief error message though - "termios call failed" or something.

takluyver:
Also, a comment mentioning why we're raising a different exception would be good.

vbraun:
Done

takluyver:
Thanks

PR #8212, opened 2015-03-31
Kernel logo to the left

minrk:
## before

screen shot 2015-03-31 at 11 50 59
## after
- logo

screen shot 2015-03-31 at 11 50 12
- no logo

screen shot 2015-03-31 at 11 50 33

The display toggle is changed to visibility, so that changing kernels doesn't move the document name.

cc @ellisonbg

### DIFF
--- a/IPython/html/static/notebook/js/kernelselector.js
+++ b/IPython/html/static/notebook/js/kernelselector.js
@@ -129,9 +129,9 @@ define([
         $("#kernel_indicator").find('.kernel_indicator_name').text(ks.spec.display_name);
         if (ks.resources['logo-64x64']) {
             logo_img.attr("src", ks.resources['logo-64x64']);
-            logo_img.show();
+            logo_img.css('visibility', 'visible');
         } else {
-            logo_img.hide();
+            logo_img.css('visibility', 'hidden');
         }
         
         // load kernel css
@@ -314,10 +314,10 @@ define([
         
         var logo_img = this.element.find("img.current_kernel_logo");
         logo_img.on("load", function() {
-            logo_img.show();
+            logo_img.css('visibility', 'visible');
         });
         logo_img.on("error", function() {
-            logo_img.hide();
+            logo_img.css('visibility', 'hidden');
         });
     };
 
--- a/IPython/html/static/notebook/less/kernelselector.less
+++ b/IPython/html/static/notebook/less/kernelselector.less
@@ -3,7 +3,6 @@
     margin-left: 4px;
     
     .current_kernel_logo {
-        display: none;
         .navbar-vertical-align(32px);
         width: 32px;
         height: 32px;
--- a/IPython/html/static/notebook/less/kernelselector.less
+++ b/IPython/html/static/notebook/less/kernelselector.less
@@ -1,5 +1,6 @@
 #kernel_logo_widget {
-    .pull-right();
+    .pull-left();
+    margin-left: 4px;
     
     .current_kernel_logo {
         display: none;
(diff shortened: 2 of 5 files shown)


Carreau:
gniaa, I'm neutral slightly negative..

takluyver:
Putting it right next to our own logo looks a bit cluttered, to my mind.

jhamrick:
Yeah, I also feel like it looks cluttered. What if it were to the right of the kernel name, before (autosaved)?

takluyver:
That's the notebook name, not the kernel name - I think we want to keep the save indicator close to that.

minrk:
@jhamrick's suggestion:

screen shot 2015-03-31 at 12 23 03

jhamrick:
@takluyver Oh, right -- I got confused since the notebook was called "Julia".

Now that I see it, I like it to the right of the notebook name slightly better than to the left, but not by all that much. And I do agree it is probably better to have the save indicator close to the notebook name.

minrk:
Should we just leave it on the right, then? The root issue this tries to address is adding buttons to the header via the available blocks looks like:

    <button> <logo> <button>

Which I can address easily enough by adding another block for inserting buttons that go to the right of the logo.

ellisonbg:
I mostly agree with others comments. I do like the logo immediately to the R of the Jupyter logo, but agree that the above example things are too cluttered and close. What about using whitespace to separate the kernel logo from the jupyter logo and the notebook name?

Also, when using this branch, the python2 logo no longer shows up at all? Any ideas why that would be the case (cache cleared)?

minrk:
I think more space on the left moves the document title too far to the right. I'm going to propose the new template block to the right of the kernel logo, and give up on this one, since I don't think it will work.

ellisonbg:
Here is an example of more space (not sure how I like it):

screen shot 2015-03-31 at 1 54 21 pm

ellisonbg:
OK, where would you put the buttons then?

```
<logo> <controlpanel> <logout>
```

?

On Tue, Mar 31, 2015 at 1:54 PM, Min RK <notifications@github.com> wrote:

> I think more space on the left moves the document title too far to the
> right. I'm going to propose the new template block to the right of the
> kernel logo, and give up on this one, since I don't think it will work.
>
> —
> <https://github.com/ipython/ipython/pull/8212#issuecomment-88241286>.
>



-- 
Brian E. Granger

takluyver:
Even with a bit more space, I think I prefer it on the right. The kernel name is also on the right, in the menu bar.

minrk:
I opened #8215 with the header block. It looks like this, now:

screen shot 2015-03-31 at 14 25 19

ellisonbg:
This is better than the logo splitting the buttons. Could we not put the
buttons to the L of the logo and keep the logo at the right most position
in the header? Given that the button list could grow further, I think that
makes the most sense.

On Tue, Mar 31, 2015 at 2:26 PM, Min RK <notifications@github.com> wrote:

> I opened #8215 <https://github.com/ipython/ipython/pull/8215> with the
> header block. It looks like this, now:
>
> [image: screen shot 2015-03-31 at 14 25 19]
> <https://cloud.githubusercontent.com/assets/151929/6929707/ce195f6c-d7b1-11e4-8d84-f948356e5868.PNG>
>
> —
> <https://github.com/ipython/ipython/pull/8212#issuecomment-88255259>.
>



-- 
Brian E. Granger

minrk:
It feels a little wrong to me to have anything to the right of the logout button, but if that's preferable, it's easy enough. And the header_buttons block can be eliminated, but I might keep it anyway, for clarity of segments.

minrk:
By accident, #8215 currently puts the kernel logo to the right of the logout button, if you want to check that out.

minrk:
If we do put the kernel logo on the far-right, I would keep the visible toggle instead of the display toggle, so it doesn't move buttons when it comes and goes.

minrk:
Closing in favor of the much simpler #8215.

PR #8213, opened 2015-03-31
disable install from master

minrk:
while it's broken by The Big Split

with informative note about `pip install -e`


takluyver:
+1

rsmith31415:
@minrk Are you sure the message conveys enough information? 

>While we are in the midst of The Big Split,  IPython cannot be installed from master. You can use `pip install -e .` for an editable install, which still works.

Some people might not know what is The Big Split.

PR #8214, opened 2015-03-31
Split IPython specific functions out of utils.path

takluyver:
Functions like `get_ipython_dir()` are moved to a new `IPython.paths` module. This should facilitate moving the remainder of utils.path into genutils. Wrappers are left in utils.path which will still work, but issue a deprecation warning on call.

While doing this, I noticed that the code to check for pre-0.11 config files and warn about them is still in place. I think that has served its purpose and can be retired now.


Carreau:
+1

IPython/core/magics/test_paths.py
```diff
@@ -0,0 +1,246 @@
+from contextlib import contextmanager
```
minrk: I assume this isn't meant to be in magics

takluyver: No, it wasn't. Fixed.

IPython/utils/path.py
```diff
@@ -443,81 +367,19 @@ def target_update(target,deps,cmd):
...
-    """
-    # import here, because profiledir also imports from utils.path
-    from IPython.core.profiledir import ProfileDir
-    try:
-        pd = ProfileDir.find_profile_dir_by_name(get_ipython_dir(), profile)
-    except Exception:
-        # will raise ProfileDirError if no such profile
-        raise IOError("Profile %r not found")
-    return filefind(filename, ['.', pd.security_dir])
+    warn("get_security_file has moved to the IPython.paths module")
```
minrk: get_security_file is going to leave IPython entirely, since it's a Jupyter function.

takluyver: I found we were only using this in one example, and `find_connection_file`, which is already in jupyter_client, works as well for that, so I have removed this function entirely.

PR #8215, opened 2015-03-31
add header_buttons block to page template

minrk:
allows easier access to adding buttons next to the logout button in custom templates.


takluyver:
This is all before the headercontainer, where the save widget and kernel logo go. But I thought the logout button went to the right of that? :confused:

minrk:
It's `float: right`, so the order actually doesn't matter. We can move it below, so the HTML more clearly matches the display, though.

minrk:
Reordered the two blocks, so the order of elements more closely resembles their appearance. Due to the `float` behavior, it has no visual effects.

minrk:
This puts the kernel logo to the right of logout, since the logo was in headercontainer. I guess it makes sense to keep headercontainer after header_buttons, since anything floated right in there should expect to stay to the right of header_buttons.

minrk:
Removed the reordering of blocks, since it causes unexpected behavior for elements floated to the right in headercontaner.

PR #8217, opened 2015-04-01
detangle client/kernel KernelSpec bits

minrk:
- Moves the kernel spec for the IPython kernel to `ipython_kernel.kernelspec`
- install_native_kernel / install-self remain for now,
  but are reduced to simple wrappers around ipython_kernel.
- The 'native' kernel is no longer loaded from memory.
  Instead, IPython is a regular kernel, installed in regular locations.
  The only special treatment it gets is that if it is available and not installed,
  it will be installed at runtime.


jupyter_client/kernelspec.py
```diff
@@ -205,14 +173,10 @@ def install_native_kernel_spec(self, user=False):
...
         """
-        path = self._get_destination_dir(NATIVE_KERNEL_NAME, user=user)
-        os.makedirs(path, mode=0o755)
-        with open(pjoin(path, 'kernel.json'), 'w') as f:
-            json.dump(self._native_kernel_dict, f, indent=1)
-        copy_from = self._native_kernel_resource_dir
-        for file in os.listdir(copy_from):
-            shutil.copy(pjoin(copy_from, file), path)
-        return path
+        from ipython_kernel.kernelspec import write_kernel_spec
```
takluyver: Is this method deprecated?

minrk: I'm not quite sure. I originally thought it would be, but if Jupyter projects are going to have the `ensure_ipython_present`, maybe not? Maybe there should be ipython_kernel.kernspec.install_self(), in which case I can deprecate this method.

minrk: I added `ipython_kernel.kernelspec.install()`, which is called by the now-deprecated `install_native_kernel`

jupyter_client/kernelspec.py
```diff
@@ -195,24 +163,10 @@ def install_kernel_spec(self, source_dir, kernel_name=None, user=False,
         shutil.copytree(source_dir, destination)
 
     def install_native_kernel_spec(self, user=False):
-        """Install the native kernel spec to the filesystem
-
-        This allows a Python 3 frontend to use a Python 2 kernel, or vice versa.
-        The kernelspec will be written pointing to the Python executable on
-        which this is run.
+        """DEPRECATED: Use ipython_kernel.kenelspec.install"""
```
takluyver: Should it warn? I guess it's unlikely that anyone is calling this directly, but maybe it's best for consistency as we move things around.

jupyter_client/kernelspecapp.py
```diff
@@ -92,7 +91,7 @@ def start(self):
             raise
 
 class InstallNativeKernelSpec(BaseIPythonApplication):
-    description = """Install the native kernel spec directory for this Python."""
+    description = """[DEPRECATED] Install the IPython kernel spec directory for this Python."""
```
takluyver: Likewise, should running this application warn?

jupyter_client/launcher.py
```diff
@@ -12,99 +11,6 @@
...
-                    was_flag = name in flags
-            elif len(split) == 1 and any(flag.startswith(name) for flag in flags):
-                # strip flag, but don't swallow next, as flags don't take args
-                stripped.remove(a)
-
-    # return shortened list
-    return stripped
-
-
-def make_ipkernel_cmd(mod='ipython_kernel', executable=None, extra_arguments=[], **kw):
```
takluyver: We could leave a deprecation wrapper for this, though it's probably not worth it.

jupyter_client/kernelspec.py
minrk: Yup, warning added.

jupyter_client/kernelspecapp.py
Carreau: This will already wan if import in a shim right ?

takluyver: But it's importing the new location, not the shim. And in any case, there should be a separate warning for the subcommand being deprecated as opposed to the import.

minrk: I didn't want to put a warning on the application endpoint when there isn't an alternative.

takluyver: That's fair enough. We can always add a warning once we make the alternative.

takluyver:
+1

PR #8224, opened 2015-04-02
Path to glyphicon fonts (used by Bootstrap) was incorrect. Fixed paths t...

tmtabor:
Path to glyphicon font resources incorrect (used by Bootstrap). Fixed paths to coincide with where the appropriate Bootstrap resources are when IPython-components is attached. Useful for those using Bootstrap classes in custom themes.

### DIFF
--- a/IPython/html/static/base/less/variables.less
+++ b/IPython/html/static/base/less/variables.less
@@ -13,8 +13,9 @@
 @logo_height: 28px;
 @border-radius-small: 1px;
 @border-radius-base: 2px;
-@border-radius-large: 3px;;
+@border-radius-large: 3px;
 @grid-gutter-width: 0px;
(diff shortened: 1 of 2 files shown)


minrk:
There shouldn't be any commits to style.min.css only, since this is an autogenerated file. Can you find the right place to put this fix in a LESS file?

Carreau:
Not obvious, but, add a `@icon-font-path` in `IPython/html/static/base/less/variables.less` which value is `"../components/bootstrap/fonts/"` (instead of `"../fonts/"`) should work.

tmtabor:
I've added the change to variables.less and checked it into the branch.

Carreau:
Seem sensible. Backport to 3.1 ? Does anybody use glyphicons ?

minrk:
Yup, makes sense for 3.1

PR #8227, opened 2015-04-02
move serialization utils out of IPython.utils

minrk:
- jsonutil to jupyter_client
- pickleutil to ipython_kernel


jupyter_client/tests/test_jsonutil.py
```diff
@@ -10,9 +10,9 @@
 
 import nose.tools as nt
 
-from IPython.utils import jsonutil, tz
+from jupyter_client import jsonutil, tz
```
takluyver: This will break the tz import, I think - it should be moved to a separate line.

takluyver:
codeutil is moved too, without a backwards compatibility shim. That's fine by me, but I wanted to make it clear.

jupyter_client/tests/test_jsonutil.py
minrk: removed the use of tz by copying the relevant bits into the test file

takluyver:
+1

PR #8228, opened 2015-04-02
remove utils.doctestreload

minrk:
it's unused, and its docstring suggests that it's obsolete on all currently supported Pythons.


Carreau:
+1 once travis greens.

PR #8229, opened 2015-04-02
move IPython.utils.log to traitlets.log

minrk:
since it's a utility for grabbing the logger of the current Application


IPython/utils/log.py
```diff
@@ -1,25 +1,5 @@
...
-    
-    if _logger is None:
-        from IPython.config import Application
-        if Application.initialized():
-            _logger = Application.instance().log
-        else:
-            logging.basicConfig()
-            _logger = logging.getLogger()
-    return _logger
+from traitlets.log import *
```
takluyver: I think this needs a `from __future__ import absolute_import`

minrk: yup, added.

PR #8230, opened 2015-04-02
move sentinel out of IPython.utils.signature

minrk:
utils.signature is a backport of a stdlib file, and not an appropriate place for new code.

Further, to avoid adding a new trivial module to genutils, a copy is added each place we've used a sentinel.

The latter bit is debateable because it's obvious code duplication, but comes from the perspective that genutils is an unfortunate compromise that we have for now, and we will work toward phasing it out, adding no new content to it.


Carreau:
The full sentinel file/class seem a bit overkill now, vs just overwriting object `_repr_`, but we can simplify later.

PR #8233, opened 2015-04-02
Don't use tokenutil in v4 msgspec adapter

minrk:
reimplement javascript token logic from IPython 2, instead.

Avoids jupyter_client dependency on tokenutil, which should stay with IPython.

### DIFF
--- a/jupyter_client/adapter.py
+++ b/jupyter_client/adapter.py
@@ -3,10 +3,10 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import re
 import json
 
 from IPython.core.release import kernel_protocol_version_info
-from IPython.utils.tokenutil import token_at_cursor
 
 
 def code_to_line(code, cursor_pos):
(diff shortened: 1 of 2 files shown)


Carreau:
Travis seemed to be unable to clone ipython, relaunched.

Carreau:
Adding `!` as valid in identifiers would fix a julia bug for inspection worth adding ?

Carreau:
+1 otherwise, regardless of response.

minrk:
@Carreau languages also allow :, ?, etc. ! is an operator in some languages, so it would be wrong to allow it sometimes, and right others. I'm not sure there's a reasonable place to stop, which is why we moved lexing to the kernel. For that reason, I'm somewhat inclined to keep the 2.x behavior verbatim, since it's what v4 kernels have always received, and if they want to improve lexing they should implement v5.

Carreau:
Fair enough. Julia was just one of the only languages widely used for which this is a regression. 


merging,

minrk:
@Carreau it shouldn't be a regression, since prior to this it was using Python lexing, which would also exclude `!`.

Carreau:
Julia was able to patch it, which now they can't . Should have been cleared.

minrk:
Ah, a regression in monkeypatchability between IPython 2 and IPython 3 when using v4, but not a regression introduced here. This PR shouldn't make anything better or worse.

PR #8235, opened 2015-04-02
use testpath.tempdir for utils.tempdir

minrk:
instead of adding tempdir to genutils


takluyver:
Besides tests, we use tempdir in nbextensions, nbconvert (pdf exporter, svg2pdf preprocessor) and console.

PR #8237, opened 2015-04-02
Don't write notebooks to disk when signing them

minrk:
Since the signature isn't stored in the notebook anymore, there's nothing to write.


minrk:
This should go in 3.1

takluyver:
+1

PR #8240, opened 2015-04-03
split utils.io

minrk:
- atomic_writing moved to html.contents.fileio
- unicode_std_stream to nbconvert.utils.io
- remove inappropriate rprint in session


takluyver:
looks good

takluyver:
I think `jupyter_nbconvert.utils.pandoc` needs an `absolute_import`

minrk:
absolute import added. Travis should go green momentarily.

PR #8243, opened 2015-04-03
remove testing.tools.monkeypatch

minrk:
use mock.patch.object instead

part 1 of reducing dependence on testing.tools


takluyver:
Nice, it's a complete drop-in replacement :-)

PR #8245, opened 2015-04-03
move eventful to widgets

minrk:


takluyver:
Min and I had originally thought of moving eventful, and the trait types that use it (EventfulDict, EventfulList), into widgets. They are generic, but they were added for widgets and that remains the only place we use them. I don't have strong feelings about where these belong, but I'd like to make it clear so other people have a chance to consider it.

minrk:
Right. I'm about 50/50. In some sense, this is how I think container traits should always work, which is part of why I lean slightly towards leaving them in traitlets.

jdfreder:
From the meeting, move this into widgets.

minrk:
Moved to widgets

Carreau:
+1

PR #8248, opened 2015-04-03
confine doctests to IPython.core

minrk:
since that's the only place we have any, which removes the need for `@skip_doctest` sprinkled all over things.


takluyver:
+many

PR #8251, opened 2015-04-04
Moving unpack_models to widget.js

SylvainCorlay:
Since every custom widget having child widgets will need the `unpack_model` function, introduced in #7757.
It could probably be in `widget.js` rather than in `widget_box.js`, especially since widget.js is already where ViewList is defined.


SylvainCorlay:
@jasongrout

jasongrout:
Makes sense to me.

SylvainCorlay:
Just rebased and also moved the python side widget serialization to widget.py

jdfreder:
Thanks @SylvainCorlay

PR #8252, opened 2015-04-04
Color trait backport

SylvainCorlay:
This is partial backport of #7826 to 3.x (skipping commit 37e32ddc58ebccc081e0f946465ae0d7d6cecdc6). . It adds the color trait in the widget module, but does not change the widgets themselves. Hence there is no API change.
This will help us support 3.1 and 4.0 with the same codebase.

### DIFF
--- a/IPython/html/widgets/__init__.py
+++ b/IPython/html/widgets/__init__.py
@@ -1,5 +1,7 @@
 from .widget import Widget, DOMWidget, CallbackDispatcher, register
 
+from .trait_types import Color
+
 from .widget_bool import Checkbox, ToggleButton
 from .widget_button import Button
 from .widget_box import Box, FlexBox, HBox, VBox
--- a/IPython/html/widgets/tests/test_traits.py
+++ b/IPython/html/widgets/tests/test_traits.py
@@ -1,0 +1,20 @@
+"""Test trait types of the widget packages."""
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from unittest import TestCase
+from IPython.utils.traitlets import HasTraits
+from IPython.utils.tests.test_traitlets import TraitTestBase
+from IPython.html.widgets import Color
+
+
+class ColorTrait(HasTraits):
+    value = Color("black")
+
+
+class TestColor(TraitTestBase):
+    obj = ColorTrait()
+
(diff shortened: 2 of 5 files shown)


minrk:
SInce this is a new trait, and 3.1 was just released yesterday, I'm not sure we should add it for 3.2. If you ship this trait with whatever downstream code you are maintaining, wouldn't that allow you to support 3.0 as well?

SylvainCorlay:
Ouch, did not see I missed 3.1.

SylvainCorlay:
Well, I use it in multiple independent packages which don't depend on each other otherwise. That is why it was convenient to have it upstream.

SylvainCorlay:
Is there a downside in getting this in? No api change, and little amount of added code. It is like a __future__ import.

minrk:
It seems a bit odd to me to backport something that's totally unused by IPython to a stable release already in bugfix mode, only to provide a minor convenience. We aren't super strict about bugfixes-only, though, so maybe this is appropriate for 3.x.

The downside to backporting new code is typically support - bugfixes, etc. need to be backported as well. It also means freezing an API that will be new in 4.0 that hasn't been released yet. If we do merge this, that means that this API in master needs to enter backward compatibility mode before its first release, and avoid backward-incompatible API changes, and you should make sure to backport any fixes found in 4.0 to 3.x. Can you commit to being responsible for both of those?

SylvainCorlay:
Alright fair enough.

minrk:
Sorry, I meant that as a real question - if you can commit to supporting it, I'm okay with it going in.

SylvainCorlay:
OK, that would be great. It is just that I have a lot a widget code to maintain, and I want users to be able to install it on *a* stable version of IPython, especially if it is going to become more difficult to run from source after the repo split. 
While the API changed a lot between 2.0 and 3.0, I believe that it might be possible to have the same codebase between 3.0 and 4.0 since 4.0 is mostly going to be about the big split. I am fine maintaining this code.

minrk:
@SylvainCorlay great, thanks.

minrk:
Sorry, I know it can be hard to distinguish between a passive aggressive rhetorical question and a real, honest one on the Internet.

SylvainCorlay:
No worries, I really did not take it that way, and the same applies to me, I apologize if I was too blunt. You guys are doing are really super responsive on bug reports and pull-requests and we are really grateful for it.

PR #8255, opened 2015-04-04
reduce use of testing.tools

minrk:
- Move help output test functions to traitlets.tests.utils
  since it's a generic test for any Application object.
- Don't use testing.tools.get_ipython_cmd to run nbconvert in tests.

This should eliminate the need for any testing.tools in genutils.


takluyver:
+1

PR #8256, opened 2015-04-04
remove dependency on utils.process outside IPython

minrk:
- a couple of minor reimplementations of subprocess output
- use shutil.which instead of find_cmd
- copy check_pid to html.utils


takluyver:
That all looks reasonable.

PR #8263, opened 2015-04-07
Pass the arguments to the callback in the on_some_change function

ClaytonPassmore:
Currently on_some_change does not pass the arguments that are passed when the change event gets fired. This change simply passes the arguments to the callback.


jdfreder:
Thanks @ClaytonPassmore .  Next time you contribute a change that changes the API, please write a what's new doc (`\docs\source\whatsnew\pr`) so we can keep track of what has changed since the last release.  I've written one for you in #8275.  You can review it if you'd like.

ClaytonPassmore:
@jdfreder happy to contribute! Thanks for doing the leg work. Now I know for next time!

PR #8268, opened 2015-04-07
Handle comm messages in execute preprocessor

### LINKED ISSUE #8264: Comm messages aren't handled in ExecutePreprocessor
The `ExecutePreprocessor` in nbconvert doesn't properly handle comm messages, resulting in a bunch of errors like this when using widgets or the `nbagg` matplotlib backend:

```
unhandled iopub msg: comm_open
unhandled iopub msg: comm_msg
unhandled iopub msg: comm_msg
...
```

The fix is pretty trivial; it just involves adding another if clause to the if statement that checks the message type:

```
elif msg_type.startswith('comm'):
    continue
```

I feel like this is something that should probably be fixed in 3.x, but since nbconvert has already been moved I'm not sure where I should submit the fix. Should I submit it against master, and whoever backports it will handle making the same changes on 3.x? Should I submit two PRs? Should I just wait until the split is done?

jhamrick:
Fixes #8264

### DIFF
--- a/jupyter_nbconvert/preprocessors/execute.py
+++ b/jupyter_nbconvert/preprocessors/execute.py
@@ -130,6 +130,8 @@ class ExecutePreprocessor(Preprocessor):
             elif msg_type == 'clear_output':
                 outs = []
                 continue
+            elif msg_type.startswith('comm'):
+                continue
 
             try:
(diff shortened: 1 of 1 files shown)


minrk:
Merged and backported. Thanks!

PR #8271, opened 2015-04-07
Allow to set same mode as existing if load fails.

Carreau:
Setting the mode many times in a row, would fail to actually set the
mode if it was failing to load the mode the first time.

The prevent the above behavior by falling back **explicitly** to 'null'
if mode does not exist yet, or can't be loaded.

---

I think it would be nice to backport that for IHaskell. ping @gibiansky.

### DIFF
--- a/IPython/html/static/notebook/js/notebook.js
+++ b/IPython/html/static/notebook/js/notebook.js
@@ -1514,29 +1514,50 @@ define(function (require) {
     Notebook.prototype.cell_toggle_line_numbers = function() {
         this.get_selected_cell().toggle_line_numbers();
     };
+
+
+    //dispatch codemirror mode to all cells.
+    Notebook.prototype._dispatch_mode = function(spec, newmode){
+        this.codemirror_mode = newmode;
+        codecell.CodeCell.options_default.cm_config.mode = newmode;
+        this.get_cells().map(function(cell, i) {
+            if (cell.cell_type === 'code'){
+                cell.code_mirror.setOption('mode', spec);
+                // This is currently redundant, because cm_config ends up as
+                // codemirror's own .options object, but I don't want to
+                // rely on that.
+                cell.cm_config.mode = spec;
+            }
+        });
+
+    };
+
+    // roughly try to check mode equality
+    var _mode_equal = function(mode1, mode2){
+        return ((mode1||{}).name||mode1)===((mode2||{}).name||mode2);
+    };
     
     /**
      * Set the codemirror mode for all code cells, including the default for
      * new code cells.
+     * Set the mode to 'null' (no highlighting) if it can't be found.
      */
     Notebook.prototype.set_codemirror_mode = function(newmode){
-        if (newmode === this.codemirror_mode) {
+        // if mode is the same don't reset,
+        // to avoid n-time re-highlighting.
+        if (_mode_equal(newmode, this.codemirror_mode)) {
             return;
         }
-        this.codemirror_mode = newmode;
-        codecell.CodeCell.options_default.cm_config.mode = newmode;
         
         var that = this;
         utils.requireCodeMirrorMode(newmode, function (spec) {
-            that.get_cells().map(function(cell, i) {
-                if (cell.cell_type === 'code'){
-                    cell.code_mirror.setOption('mode', spec);
-                    // This is currently redundant, because cm_config ends up as
-                    // codemirror's own .options object, but I don't want to
-                    // rely on that.
-                    cell.cm_config.mode = spec;
-                }
-            });
+            that._dispatch_mode(spec, newmode);
+        }, function(){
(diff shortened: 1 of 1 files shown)


minrk:
:+1:

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,42 @@ define(function (require) {
...
-                    // This is currently redundant, because cm_config ends up as
-                    // codemirror's own .options object, but I don't want to
-                    // rely on that.
-                    cell.cm_config.mode = spec;
-                }
-            });
+            that._dispatch_mode(spec, newmode)
+        }, function(){
+            // on error don't dispatch the new mode as re-setting it later will not work.
+            that._dispatch_mode('null','null')
```
jdfreder: Why `'null'` here instead of `null`?

Carreau: `null` don't change cmMode. but `'null'` is a none name for codemirror mode that will just not highlight. (I tried `null`)

Carreau:
Not completly right. has to add extra logic.

jdfreder:
> Not completly right. has to add extra logic.

Okay, I set the `waiting-for-author` tag.  You can remove it when you get it figured out.

Carreau:
Ok, I think that should be it, I locally use kernel.js
```
define(['require','./codemirror-ipython.js','base/js/namespace'], function(require,cmip, IPython){
    return {
        onload: function(){
            console.log('codemirror IPython', cmip)   
            IPython.notebook.set_codemirror_mode({name:'ipython', version:3})
        }
    }
})
```

And it seem to do the trick with this PR.

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,51 @@ define(function (require) {
...
-            that.get_cells().map(function(cell, i) {
-                if (cell.cell_type === 'code'){
-                    cell.code_mirror.setOption('mode', spec);
-                    // This is currently redundant, because cm_config ends up as
-                    // codemirror's own .options object, but I don't want to
-                    // rely on that.
-                    cell.cm_config.mode = spec;
-                }
-            });
+            that._dispatch_mode(spec, newmode)
```
minrk: semicolon

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,51 @@ define(function (require) {
...
+        }, function(){
+            // on error don't dispatch the new mode as re-setting it later will not work.
+            // don't either set to null mode if it has been changed in the meantime
+            if((_mode_equal(newmode, this.codemirror_mode))){
+                that._dispatch_mode('null','null')
```
minrk: semicolon

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,51 @@ define(function (require) {
...
-                    // codemirror's own .options object, but I don't want to
-                    // rely on that.
-                    cell.cm_config.mode = spec;
-                }
-            });
+            that._dispatch_mode(spec, newmode)
+        }, function(){
+            // on error don't dispatch the new mode as re-setting it later will not work.
+            // don't either set to null mode if it has been changed in the meantime
+            if((_mode_equal(newmode, this.codemirror_mode))){
```
minrk: extra parentheses

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,51 @@ define(function (require) {
...
+                // rely on that.
+                cell.cm_config.mode = spec;
+            }
+        });
+
+    }
+
+    // roughly try to check mode equality
+    var _mode_equal = function(mode1, mode2){
+        return ((mode1||{}).name||mode1)===((mode2||{}).name||mode2)
```
minrk: semicolon

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,51 @@ define(function (require) {
...
+
+    }
```
minrk: semicolon

IPython/html/static/notebook/js/notebook.js
```diff
@@ -1495,29 +1495,51 @@ define(function (require) {
...
+            if (cell.cell_type === 'code'){
+                cell.code_mirror.setOption('mode', spec);
+                // This is currently redundant, because cm_config ends up as
+                // codemirror's own .options object, but I don't want to
+                // rely on that.
+                cell.cm_config.mode = spec;
+            }
+        });
+
+    }
```
minrk: semicolon

jdfreder: He must be programming in Typescript too much. ;)  Semicolons throw a compiler error on instance methods.

PR #8272, opened 2015-04-07
remove use of utils.data outside IPython

minrk:
- utils.data.flatten is list(chain.from_iterable())
- copy utils.data.uniq_stable to a couple of places


Carreau:
Want to delete `uniq_stable` from utils.data ? Or used in other places ?

(+1 otherwise)

minrk:
It's still used in IPython core, etc.

Carreau:
Ok.

PR #8273, opened 2015-04-07
Minor fixes on custom serialization, and traitlet hold_traits

SylvainCorlay:
- IPython/core/application
  `old` actually cannot be `None`
- widget.js
  missing `var`
- traitlets.py `delattr` cannot be used. I recall that @minrk mentioned it in an inline comment for the PR where the context manager was added, and this case was actually not triggered by the test.

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -215,7 +215,7 @@ class BaseIPythonApplication(Application):
             return crashhandler.crash_handler_lite(etype, evalue, tb)
     
     def _ipython_dir_changed(self, name, old, new):
-        if old is not None and old is not Undefined:
+        if old is not Undefined:
             str_old = py3compat.cast_bytes_py2(os.path.abspath(old),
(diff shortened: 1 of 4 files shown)


minrk:
Can you add a test or two for the missed cases?

SylvainCorlay:
The only case where this occurs is when, at initialization, one provides a value that is rejected by the cross validation hook. 

Actually the following example raises (`AttributeError: __delete__`) in the older version and the relevant         `TraitError: foobar` in the new version. 

```Python
from IPython.utils.traitlets import HasTraits, Int, TraitError
class Foo(HasTraits):
    bar = Int()
    
    def _bar_validate(self, value, trait):
        if value:
            raise TraitError('foobar')
        return value
foo = Foo(bar=3)
```

I can add this as a test.

PS: I just committed another fix in the widget message buffers.

SylvainCorlay:
Done.

PR #8276, opened 2015-04-08
use dev for maintenance branch version_extra

minrk:
instead of 'maint'

PEP 440 helpfully made anything other than 'dev' unparseable, and evaluate as always less than zero.

cc @jhamrick

### DIFF
--- a/IPython/html/static/base/js/namespace.js
+++ b/IPython/html/static/base/js/namespace.js
@@ -4,7 +4,7 @@
 var IPython = IPython || {};
 define([], function(){
     "use strict";
-    IPython.version = "3.1.0";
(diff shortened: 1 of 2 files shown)


IPython/html/static/base/js/namespace.js
```diff
@@ -4,7 +4,7 @@
 var IPython = IPython || {};
 define([], function(){
     "use strict";
-    IPython.version = "3.1.0";
+    IPython.version = "3.2.0-dev";
```
minrk: this apparently wasn't updated after cutting 3.1, but it should be right, now.

jhamrick:
:+1:

PR #8277, opened 2015-04-08
JQuery dependency in widgets/manager.js and widgets/widget.js

SylvainCorlay:
- Removes the jquery dependency from the widget manager
- Removes jquery calls in widget.js.

(I am looking into backbone.nativeview for the base widget view)

### DIFF
--- a/IPython/html/static/widgets/js/manager.js
+++ b/IPython/html/static/widgets/js/manager.js
@@ -4,11 +4,10 @@
 define([
     "underscore",
     "backbone",
-    "jquery",
     "base/js/utils",
     "base/js/namespace",
     "services/kernels/comm"
-], function (_, Backbone, $, utils, IPython, comm) {
+], function (_, Backbone, utils, IPython, comm) {
     "use strict";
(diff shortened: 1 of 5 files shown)


minrk:
I'm not sure this one is going to get in before the split

SylvainCorlay:
Ok, I can reopen after *The Split*.
I am looking into making the widget views mixins of 
 - their own base backbone view (jquery, d3, native)
 - and the base widget stuff.
This can also be a different PR.

SylvainCorlay:
Actually, it will also make it easier to use phosphor panels as layout widgets.

minrk:
I'm ready to make the HTML split right now. Is this PR all done, or do you still have stuff to do here?

minrk:
Hmm, I think the jQuery removal may need a bit of discussion. Is it okay to wait for this one until the split? We'll do that as quickly as we can.

SylvainCorlay:
No problem, reopening after the split.

minrk:
Great, thanks. I'll get that new repo up for you as quickly as I can.

PR #8280, opened 2015-04-08
add ipython_genutils

minrk:
copies of shared dependencies, that can't feasibly be detangled.

This package will only exist as long as it has to,
and will be killed as soon possible.

From the dev meeting, since IPython Prime will likely preserve the need for the relocated functionality, these are _copies_ of existing files, leaving the originals in IPython. The only one I'm not sure about on that front is py3compat - should py3compat be a copy, or should we use genutils, even in IPython Prime?
- ipstruct, tempdir, py3compat (+encoding) are whole
- path, text are reduced to remove some IPython-only bits
- testing.decorators removes the dependencies on knownfailure,
  since only IPython projects use those at this point.

### DIFF
--- a/ipython_genutils/encoding.py
+++ b/ipython_genutils/encoding.py
@@ -1,0 +1,71 @@
+# coding: utf-8
+"""
+Utilities for dealing with text encodings
+"""
+
+#-----------------------------------------------------------------------------
+#  Copyright (C) 2008-2012  The IPython Development Team
+#
+#  Distributed under the terms of the BSD License.  The full license is in
+#  the file COPYING, distributed as part of this software.
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Imports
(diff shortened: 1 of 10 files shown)


Carreau:
If this is 100% a new package, why not make a new repo **now**, and release it on PyPi ? It's not like we are moving things around and need to keep it working , or is it for nexts PRs ?

minrk:
@Carreau all of the splits are new packages. We are doing the split in this repo because the history of these files will be preserved, even though they are copies.

Carreau:
Ah, true. 

+1 then

minrk:
I realize that since IPython won't use these, genutils will only be used by Jupyter, so maybe jupyter_genutils makes more sense. It is 'genutils from IPython', so I could go either way. Anyone care?

Carreau:
Don't really care as it's meant to disapear, but feel free to rename.

PR #8281, opened 2015-04-08
finish scroll-cell PR

### LINKED ISSUE #8061: Keep current input cell at the top or in the middle of the screen

minrk:
supersedes #8067
closes #8061

### DIFF
--- a/IPython/html/static/notebook/js/keyboardmanager.js
+++ b/IPython/html/static/notebook/js/keyboardmanager.js
@@ -80,8 +80,6 @@ define([
             'down'                : 'ipython.move-cursor-down-or-next-cell',
             'ctrl-shift--'        : 'ipython.split-cell-at-cursor',
             'ctrl-shift-subtract' : 'ipython.split-cell-at-cursor',
-            'ctrl-l'              : 'ipython.recenter',
-            'ctrl-shift-l'        : 'ipython.top'
         };
     };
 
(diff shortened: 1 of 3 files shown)


IPython/html/static/notebook/js/keyboardmanager.js
```diff
@@ -79,7 +79,7 @@ define([
             'up'                  : 'ipython.move-cursor-up-or-previous-cell',
             'down'                : 'ipython.move-cursor-down-or-next-cell',
             'ctrl-shift--'        : 'ipython.split-cell-at-cursor',
-            'ctrl-shift-subtract' : 'ipython.split-cell-at-cursor'
+            'ctrl-shift-subtract' : 'ipython.split-cell-at-cursor',
```
Carreau: Trailing `,` but it's or in JS IIRC.

Carreau:
+1

PR #8283, opened 2015-04-08
move IPython.html to jupyter_notebook

minrk:
This is the last piece of the big split before creating the actual repos.
It needs to wait for #8271, though.


jdfreder:
Everything is awesome! :musical_note:

PR #8288, opened 2015-04-09
Big Split: remove traitlets

minrk:
add [traitlets package](https://github.com/ipython/traitlets) as a dependency.


minrk:
This is based on #8285

minrk:
Since the dependencies are unreleased, there's a requirements.txt for installing them from git, which will become unnecessary once there are released versions.

The alternative approach for this temporary problem is using `dependency_links`, which is deprecated by pip, and disabled by default. The only advantage is that with dependency_links, installation from master can be done with a single call:

    pip install --process-dependency-links -e git+https:...

which isn't true of the requirements.txt, which requires manual clone:

    git clone git+https://github.com/ipython/ipython
    cd ipython
    pip install -r requirements.txt -e .

I'm inclined to use requirements.txt for now, and only switch to dependency_links if the lack of a single call dev-install becomes too annoying.

Carreau:
Min will push the button on my computer...

PR #8307, opened 2015-04-11
makes dictionary key completion use same delimiters as splitter

jeffhussmann:
If a dictionary key contains a `-`, `~`, or `/`, trying to tab-complete from a partially typed key that has made it past the problem character will replace the entire key with just the part after the problem character. This happens because these characters are by default removed from readline's delimiters but not from the relevant regexp pattern in `match_dict_keys`. This fixes this behavior by making `match_dict_keys` take its set of delimiters from the relevant `CompletionSplitter` and adds a test.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -416,7 +416,7 @@ def get__all__entries(obj):
     return [w for w in words if isinstance(w, string_types)]
 
 
-def match_dict_keys(keys, prefix):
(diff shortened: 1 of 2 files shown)


minrk:
Just a note: this is a PR against 3.x due to the Big Split, so if merged for 3.x, it will need to be forward-ported.

takluyver:
I am having trouble understanding the code this modifies, but I think this fix is right. Thanks!

PR #8312, opened 2015-04-13
Base64 video

MarkDaoust:
Embedding video was broken for python3, just giving a traceback.

It was loading the file's binary as bytes, then trying to ".encode" them to base64. Bytes don't have a ".encode" method. The "base64" module does this bytes-->bytes translation.

This should be python2 compatible. 

The decode transforms back to text form bytes for the ".format"

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -877,7 +877,7 @@ class Video(DisplayObject):
         output = """<video controls>
  <source src="data:{0};base64,{1}" type="{0}">
  Your browser does not support the video tag.
- </video>""".format(mimetype, video_encoded.decode('utf-8'))
+ </video>""".format(mimetype, video_encoded.decode('ASCII'))
         return output
 
     def reload(self):
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -6,6 +6,7 @@
 
 from __future__ import print_function
 
+import base64
 import json
 import mimetypes
 import os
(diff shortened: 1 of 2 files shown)


rgbkrk:
Thank you for the PR!

@sukisuki is one of the sprinters at PyCon and we weren't sure where to submit this patch (aim at the backports, forward port to `IPython/lib/display.py`.

Sorry if I advised the wrong direction, I now see that this patch also belongs in https://github.com/ipython/ipython/blob/master/IPython/core/display.py, which has not changed paths.

Should we submit a follow on PR?

minrk:
Typically we prefer to make and review PRs on master, then backport, rather than on the stable version and forward-port. I don't know how strict we want to be about that. For PRs that touch the split repos, they will have to be made as separate PRs against two repos, but since this is truly an IPython PR, our existing backport machinery will work correctly if it's made against master.

Carreau:
I would say that maybe for a couple of weeks (and especially pycon)
installing master is too painful, and we don't (yet) have docs.
But agree that we should relatively soon reask to PR on correct repo.

rgbkrk:
@sukisuki headed out after we lost internet, but I did talk to him in person again. He had also written some tests and would be happy to target master. After we had gotten it to work in the notebook for him, it was clear we could come back to applying the changes again with master relying on tests for validity.

sukisuki:
Thanks for the input.
I'll retarget on master in a second.

PR #8313, opened 2015-04-13
imp module deprecated in 3.4

Carreau:
We still use imp.find_module,(in IPython/utils/module_paths.py)  but
this as no newer drop-in replacement as far as I understand.

Some alternative suggested here:

https://docs.python.org/3/library/importlib.html#importlib.abc.Finder

Deprecated since version 3.3: Use MetaPathFinder or PathEntryFinder
instead.

The various error messages are not clear of the exact version where
this is deprecated.

---

Turned warning into errors to find some extra bad import of IPython.nbformat and found that.

### DIFF
--- a/IPython/core/extensions.py
+++ b/IPython/core/extensions.py
@@ -13,7 +13,11 @@ from IPython.utils.path import ensure_dir_exists
 from traitlets import Instance
 from IPython.utils.py3compat import PY3
 if PY3:
-    from imp import reload
+    try: 
+        from importlib import reload
+    except ImportError :
+        ## deprecated since 3.4
+        from imp import reload
 
 #-----------------------------------------------------------------------------
 # Main class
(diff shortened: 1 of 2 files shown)


IPython/core/extensions.py
```diff
@@ -13,7 +13,11 @@
 from traitlets import Instance
 from IPython.utils.py3compat import PY3
 if PY3:
-    from imp import reload
+    try:
```
minrk: if you are doing try/except, this doesn't need to be inside `if PY3`

Carreau: It would try to import reload from imp in 2.x if I don't wrap in if PY3.

Carreau: scratch that. imp.reload is aliased to the build-in in py2, removed, and drop the PY3 import.

PR #8318, opened 2015-04-13
Remove unneeded Pdb.checkline override

### LINKED ISSUE #6572: Legacy cleanup
IPython/core/debbuger.py contains the following:

```
#######################################################################
# XXX Hack!  Use python-2.5 compatible code for this call, because with
# all of our changes, we've drifted from the pdb api in 2.6.  For now,
# changing:
#
#line = linecache.getline(filename, lineno, self.curframe.f_globals)
# to:
#
line = linecache.getline(filename, lineno)
#
# does the trick.  But in reality, we need to fix this by reconciling
# our updates with the new Pdb APIs in Python 2.6.
#
# End hack.  The rest of this method is copied verbatim from 2.6 pdb.py
#######################################################################
```

Should be easy to clean/test. Marking as Sprint friendly quickfix as we dropped 2.5 and 2.6 support.

drocco007:
Removed the Pdb.checkline override since support for Python 2.5/2.6 has been dropped. /cc @takluyver 

Closes #6572

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -558,36 +558,3 @@ class Pdb(OldPdb):
         namespaces = [('Locals', self.curframe.f_locals),
                       ('Globals', self.curframe.f_globals)]
         self.shell.find_line_magic('psource')(arg, namespaces=namespaces)
-
-    def checkline(self, filename, lineno):
-        """Check whether specified line seems to be executable.
-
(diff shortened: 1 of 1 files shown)


takluyver:
@drocco007 tested this interactively and didn't find any problems, and the tests pass. I'm merging it, but we should keep it in mind if a problem comes up in the debugger in the next few weeks.

PR #8324, opened 2015-04-15
dont' swallow sys.exit

Carreau:
Shoud BaseIPythonApp be a subclass of Jupyter_core App instead of traitlets ?


minrk:
> Shoud BaseIPythonApp be a subclass of Jupyter_core App instead of traitlets ?

No, IPython doesn't depend on anything in Jupyter. I think @ellisonbg discussed at the dev meeting that some of the config file loading should be moved up to the base traitlets Application, though.

PR #8326, opened 2015-04-16
Base64 video master

MarkDaoust:
Embedding video was broken for python3, just giving a traceback.

It was loading the file's binary as bytes, then trying to ".encode" them to base64. Bytes don't have a ".encode" method. The "base64" module does this bytes-->bytes translation.

This is python2 compatible.

The decode transforms back to text form bytes for the ".format"

This is a copy of PR #8312, +1 test, re-based on master.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -877,7 +877,7 @@ class Video(DisplayObject):
         output = """<video controls>
  <source src="data:{0};base64,{1}" type="{0}">
  Your browser does not support the video tag.
- </video>""".format(mimetype, video_encoded.decode('utf-8'))
+ </video>""".format(mimetype, video_encoded.decode('ASCII'))
         return output
 
     def reload(self):
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -6,6 +6,7 @@
 
 from __future__ import print_function
 
+import base64
 import json
 import mimetypes
 import os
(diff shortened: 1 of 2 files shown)


IPython/core/display.py
```diff
@@ -869,14 +870,14 @@ def _repr_html_(self):
...
-            video_encoded = video.encode('base64')
+            video_encoded = base64.b64encode(video)
         else:
             video_encoded = self.data
             mimetype = self.mimetype
         output = """<video controls>
  <source src="data:{0};base64,{1}" type="{0}">
  Your browser does not support the video tag.
- </video>""".format(mimetype, video_encoded)
+ </video>""".format(mimetype, video_encoded.decode('ASCII'))
```
Carreau: I'm a little concern as this changes if `filename` is `None`, `video_encoded = self.data`, which suddenly get decoded as ASCII while it was not before.

takluyver: If you're explicitly passing video data, I think it should always be bytes, in which case this is correct. However, in that case we should be base64-ing `self.data` as well.

Carreau:
Though I dont' see any logic that convert data = string/bytes to the right format, so I'm not sure.

IPython/core/display.py
Carreau: > If you're explicitly passing video data, I think it should always be bytes, in which case this is correct. However, in that case we should be base64-ing self.data as well.

So what do we do ? Change "Api" and fix docstring to say that data can only be "unicode" for filename, or "bytes" in which case it should be base64 encoded ?

takluyver: unicode for filename/url, bytes for actual data. But if you're passing data, it should be the raw data as read from the file, not already base64 encoded. That matches how you pass data to `Image()`, I believe.

sukisuki: Yeah, `Image` expects the raw bytes.

I can fix this, and update this PR.

It still won't be a perfect match for `Image`, which only uses `_repr_html_` for links, but it'll basically work.

Thanks for the feedback.

takluyver: I think that makes sense, thanks.

I don't know enough about video formats to know what `_repr_?_`, what mimetype etc. would make sense for video data - I guess there are several different formats that might be used. So I think it's simplest to just stick with `_repr_html_` for now.

Carreau:
reviving that. We are now on 5.0 branch.

Carreau:
Ok, this still merge cleanly, I'm going to merge soon if no object to clear up the PR queue.

minrk:
Merged, and opened #9265 with a few small fixes

PR #8336, opened 2015-04-20
Revert Axis transparent

Carreau:
```
    Revert "Set axis background to transparent in inlinebackend"

    This reverts commit 902c754fcbd0581ea33912e583cd3ccb8bdb9a5b.

    Making axis transparent seem to confuse and annoy lot of people,

    See #7964 (do not fix it, this will be a 4.0 or 5.0 thing),
    the real fix would be to use matplotlib themes.
```

---

See #7964 
this is a PR agains 3.x


takluyver:
We agreed in the dev meeting that this is the right thing to do. @Carreau , can you make the same change against `ipython_kernel` as well?

Carreau:
done in ipython/ipython_kernel#8

PR #8355, opened 2015-04-21
issue8330, dir2 to not force class members in list

patter001:
For when dir is overridden, don't force the class members back in to the list.


takluyver:
Thanks

PR #8356, opened 2015-04-21
Don't handle traits specially in dir2

takluyver:
I have checked with traitlets, completions still work OK


minrk:
There's a test for the removed behavior that's failing now.

takluyver:
Removed

PR #8365, opened 2015-04-22
Fix some import.

Carreau:
Warn is used ~ line 175,
sys is not used. TryNext and test are not used.

OSError is needed next line

### DIFF
--- a/IPython/__init__.py
+++ b/IPython/__init__.py
@@ -47,9 +47,7 @@ from .core import release
 from .core.application import Application
 from .terminal.embed import embed
 
-from .core.error import TryNext
 from .core.interactiveshell import InteractiveShell
-from .testing import test
 from .utils.sysinfo import sys_info
 from .utils.frame import extract_module_locals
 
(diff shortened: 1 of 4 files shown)


IPython/__init__.py
```diff
@@ -47,9 +47,7 @@
 from .core.application import Application
 from .terminal.embed import embed
 
-from .core.error import TryNext
 from .core.interactiveshell import InteractiveShell
-from .testing import test
```
takluyver: Things imported into `__init__` are usually deliberate in order to expose them in the public API. Static analysis tools should really ignore unused imports in `__init__`.

I'm OK with dropping TryNext from the top level API, but I'd like to keep `test()` - I think the idea was to have a convention that the Scipy stack projects could all be tested by doing `import project; project.test()`.

minrk: Boop, sorry. Yes, test should be put back.

Carreau: I did consider each import given by pyflakes individually, and did remove both on purpose, but if you prefer to keep test, then fine. I'll add it back.

PR #8371, opened 2015-04-24
More removing of import imp on 3.4+

Carreau:


IPython/external/qt_loaders.py
```diff
@@ -110,17 +110,26 @@ def has_binding(api):
                    QT_API_PYQT_DEFAULT: 'PyQt4'}
     module_name = module_name[api]
 
-    import imp
+
+    if sys.version_info.major == 3 and sys.version_info.minor >= 4:
```
takluyver: I think `sys.version_info >= (3, 4)` is neater here. Also future proof for Python 4 ;-)

Carreau: Done.

IPython/external/qt_loaders.py
```diff
@@ -110,17 +110,26 @@ def has_binding(api):
                    QT_API_PYQT_DEFAULT: 'PyQt4'}
     module_name = module_name[api]
 
-    import imp
+
+    if sys.version_info >= (3, 4):
+        def find_module(name, path):
```
takluyver: Since we only need to check if the module exists, I think we can do something simpler based on [find_loader](https://docs.python.org/3/library/importlib.html#importlib.find_loader).

The way we're using it, it also needs to raise if the module is not found, whereas I think this just returns None if there's no match.

Carreau: find_loader is also deprecated, but there is a `find_spec` :-)  that raises a KeyError.

Carreau:
Haha ! It works, now, just to cleanup all the things, and squash and documents.

IPython/utils/module_paths.py
```diff
@@ -26,25 +26,46 @@
...
+    if file is None:
+        return filename
+    else:
+        file.close()
+    if os.path.splitext(filename)[1] in [".py", ".pyc"]:
+        return filename
+    else:
+        return None
+
+if sys.version_info > (3,4):
```
takluyver: Should this be `>=`, by comparison with the rest of the code?

Carreau: Yes, probably, I have a lot to cleanup. `>=` or `>` don't change much as version_info is a 3+ tuple and `(3,4, <anything>)` will compare higher in all cases:

``` 
In [1]: (3,4) < (3,4, None)
Out[1]: True

In [2]: (3,4) < (3,4, -12)
Out[2]: True
```

Carreau:
There is some shuffeling/indenting around, so it's easier to review with [no significant whitespace `?w=1`](https://github.com/ipython/ipython/pull/8371/files?w=1)

Carreau:
Any review on this one ?

IPython/utils/module_paths.py
```diff
@@ -64,20 +56,42 @@ def find_module(name, path=None):
...
-    else:
-        return None
+    return _find_module(name, path)
+
+
+if sys.version_info >= (3,4):
+    import importlib.util
+    def _find_module(name, path):
+        if path is not None:
+            warnings.warn( "`path=` argument deprecated on python 3.5 and above", DeprecationWarning, stacklevel=2)
```
damianavila: souldn't be  "on python 3.4 and above"??

Carreau:
Bump

minrk:
Okay with me, but I'll let @takluyver have the final word.

IPython/utils/tests/test_module_paths.py
```diff
@@ -114,7 +125,15 @@ def test_find_module_1():
 
 def test_find_module_2():
     """Testing sys.path that is empty"""
-    nt.assert_is_none(mp.find_module("xmod", []))
+    # on 3.5 we cannot pass PATH top importlib,
```
takluyver: top -> to, and don't capitalise PATH, it makes it look like an environment variable

IPython/utils/module_paths.py
```diff
@@ -64,20 +56,42 @@ def find_module(name, path=None):
...
-    else:
-        return None
+    return _find_module(name, path)
+
+
+if sys.version_info >= (3,4):
+    import importlib.util
+    def _find_module(name, path):
+        if path is not None:
+            warnings.warn( "`path=` argument deprecated on python 3.4 and above", DeprecationWarning, stacklevel=2)
```
takluyver: I think this should be more explicit - it's not just deprecated, it's actually ignored.

IPython/utils/tests/test_module_paths.py
takluyver: In fact, maybe these tests should just be skipped on Python 3.4 and above.

IPython/utils/module_paths.py
Carreau: done.

IPython/utils/tests/test_module_paths.py
Carreau: done

IPython/external/qt_loaders.py
```diff
@@ -110,24 +110,29 @@ def has_binding(api):
...
-            imp.find_module('QtWidgets', mod.__path__)
+            find_module_or_spec('QtWidgets', mod.__path__)
 
         #we can also safely check PySide version
         if api == QT_API_PYSIDE:
             return check_version(mod.__version__, '1.0.3')
         else:
             return True
-    except ImportError:
+    except ImportError, ValueError:
```
takluyver: You need parentheses around these, otherwise it's invalid syntax on Python 3, and not doing what you expect on Python 2.

IPython/utils/module_paths.py
```diff
@@ -26,25 +26,16 @@
...
-
-# Our own imports
-
+import sys
+import warnings
 
 #-----------------------------------------------------------------------------
-# Globals and constants
+# Classes and functions
 #-----------------------------------------------------------------------------
```
takluyver: We're mostly removing these blocks entirely as we work on the files. Not important, though.

takluyver:
I thought this was still due to get merged in time, it just wasn't very high priority. :confused:

Carreau:
I don't really plan on working on that anymore. I can't get 3.5 to work in conda.

takluyver:
Ah, that's a pity. Well, we can always reopen this once 3.5 is out.

PR #8375, opened 2015-04-24
disable set_default_headers on TermSocket

minrk:
set_default_headers is inherited from IPythonHandler, but doesn't make sense on websockets, and raises an error if there are any headers to set, e.g.:

```
ipython notebook --NotebookApp.allow_origin='*'
```


minrk:
cc @rgbkrk

takluyver:
Cross-link jupyter/jupyter_notebook#50

rgbkrk:
:+1:

PR #8387, opened 2015-04-28
shim widgets before shimming html

minrk:
otherwise weird namespace issues prevent the second shim from being created on Python 2

cc @SylvainCorlay


SylvainCorlay:
Awesome thanks. I confirm it solves the issue for me.

jdfreder:
Awesome, thanks Min.

PR #8388, opened 2015-04-28
run test on 3.5

Carreau:
This should run test on python 3.5, but should not allow builds to be marked as failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,14 @@
 # http://travis-ci.org/#!/ipython/ipython
 language: python
 python:
+    - "nightly"
     - 3.4
     - 3.3
     - 2.7
 sudo: false
+matrix:
+    allow_failures:
+        - python : "nightly"
 before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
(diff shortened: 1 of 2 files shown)


takluyver:
`/home/travis/build.sh: line 41: /home/travis/virtualenv/pythonnightly/bin/activate: No such file or directory` - looks like Travis doesn't support this yet.

Carreau:
weird, it does work [for me on another project](https://travis-ci.org/Carreau/morning/jobs/60455753)

takluyver:
You're not using `sudo: false` on that one, perhaps that's the difference?

Carreau:
Sudo false is not needed anymore, any recently enough created repo run using docker unless asked otherwise. rerun, it "works" in the sens that we seem to have real faillures of tests.

Carreau:
And indeed. oinfo on `sum` in 3.5 returns more informations !

Carreau:
Should I move the DerivedInterrupt class out of locals at module level to fix test faillure ?

takluyver:
that seems like the easiest way to do it, yeah.

Carreau:
> that seems like the easiest way to do it, yeah.

Done.

Carreau:
Travis is not unhappy anymore.

takluyver:
Well, it [still fails](https://travis-ci.org/ipython/ipython/jobs/60463265), but #8390 or something similar would fix that.

takluyver:
Tada! https://travis-ci.org/ipython/ipython/jobs/60782452

PR #8389, opened 2015-04-28
fixes for different ways to import shims

minrk:
- ensure `__all__` and `__dir__` are defined (needed for `import *` from shims)
- make `IPython.html.widget` shim an explicit attribute on the `IPython.html` shim

These ensure all of the following import patterns work:
- `from IPython.html.widgets import *`
- `from IPython.html.widgets import Widget`
- `from IPython.html import widgets`
- `import IPython.html.widgets`

cc @SylvainCorlay for verification

### DIFF
--- a/IPython/html.py
+++ b/IPython/html.py
@@ -13,11 +13,15 @@ warn("The `IPython.html` package has been deprecated. "
 
 from IPython.utils.shimmodule import ShimModule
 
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
I checked out your branch, and tried a few ways to import widgets and things seem to work just fine now.

minrk:
Thanks.

PR #8390, opened 2015-04-28
Skip test that builtin signature can't be inspected

takluyver:
Builtins can now provide a signature, and sum() does so as of Python 3.5. I thought this is related to PEP 436, but that still appears to be in draft status, so I'm not sure.

Should avoid one of two test failures on Python 3.5 - see #8388

### DIFF
--- a/IPython/core/tests/test_oinspect.py
+++ b/IPython/core/tests/test_oinspect.py
@@ -203,6 +203,7 @@ def test_calltip_function2():
     check_calltip(g, 'g', 'g(y, z=3, *a, **kw)', '<no docstring>')
 
 
+@skipif(sys.version_info >= (3, 5))
 def test_calltip_builtin():
     check_calltip(sum, 'sum', None, sum.__doc__)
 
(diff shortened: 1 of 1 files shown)


Carreau:
Sys not defined. 

why not check that we actually get the signature ? 
```
def test_calltip_builtin():
    if sys.version_info >= (3,5):
        check_calltip(sum, 'sum', 'sum(iterable, start)', sum.__doc__)
    else:
        check_calltip(sum, 'sum', None, sum.__doc__)
```

takluyver:
It doesn't seem like the best idea to check for something that could be changed externally (e.g. if another parameter is added, or type hints in the signature). Since we're ultimately getting it by calling [`inspect.signature()`](https://docs.python.org/3/library/inspect.html#inspect.signature), there's not really anything special about builtins any more as far as our code is concerned. They're just callables, which may have a signature. I wonder if there's a more reliable way of getting a callable without a signature.

Carreau:
`[getattr(builtins, x) for x in dir(builtins) if type(getattr(builtins, x)) is type(sum) and not inspect.signature(getattr(builtins, x))]` ?

PR #8400, opened 2015-05-03
iTerm2 experimental backend support.

oselivanov:
Added support of iTerm2 experimental backend.

See:
https://github.com/matplotlib/matplotlib/pull/4397

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -24,7 +24,7 @@ backends = {'tk': 'TkAgg',
             'nbagg': 'nbAgg',
             'notebook': 'nbAgg',
             'inline' : 'module://ipython_kernel.pylab.backend_inline',
-            'iterm2': 'iTerm2'}
+            'iterm2' : 'module://matplotlib_iterm2.backend_iterm2'}
 
 # We also need a reverse backends2guis mapping that will properly choose which
 # GUI support to activate based on the desired matplotlib backend.  For the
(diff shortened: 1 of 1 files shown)


oselivanov:
In general it works, but still shows:
**UsageError: Invalid GUI request 'iterm2', valid ones are: pyglet, osx, qt5, qt, glut, gtk, gtk3, tk, wx**
message after ipython startup.

It would to awesome to get an advice how to do this integration totally right way.

oselivanov:
Just created [matplotlib_iterm2](https://github.com/oselivanov/matplotlib_iterm2).
Right now I use ugly hack (monkypatching ipython in profile config) to activate this backend.

This changeset makes possible to just use iterm2 backend in ipython without any hacks.

takluyver:
This didn't go into matplotlib, so I don't think we should be adding it to a list of known backends in IPython. If it's possible to provide an external backend for matplotlib, we should make sure IPython will work with any available mpl backend, but we shouldn't do that by adding code for each individual one we know about.

oselivanov:
Ok, but right now as far as I understand it's not possible to plug in external backend to ipython since it requires to add it to hardcoded "backends" constant here 
https://github.com/ipython/ipython/blob/master/IPython/core/pylabtools.py

What I do right now (in Installation section in README https://github.com/oselivanov/matplotlib_iterm2) - I do monkey patching of this "backends" constant (and some other one) in ipython user profile to bypass ipython checks. It works, but it's really dirty.

Is there a way to make ipython see external backends?

minrk:
> Is there a way to make ipython see external backends?

I don't think there is right now, but there should be. Most, if not all, alternative backends don't involve a GUI eventloop (e.g. the inline backend that ships as part of IPython), so the mapping IPython uses for that information doesn't need to do anything. It's possible the fix is as simple as changing from `map[key]` to `map.get(key, None)`, but we should think carefully about that.

PR #8402, opened 2015-05-03
Modified markdown rendering to include GFM-style checkboxes

jojoelfe:
Hi,

this is a quick implementation of the task/todo feature of Github flavored markdown as requested in issue #6288. The discussion suggests that his feature should be implemented by marked.js. However, a meaningful implementation should be able to modify the underlying markdown and is therefore out of the scope of marked.js.

For me this would be very helpful as I manage many ongoing experiments with notebooks and having a quick TODO feature is quite nice.

Right now checking a box triggers a complete rerendering of the cell, which might be annoying for long markdown. But I did not want to mess with the set_text function since I don't know if any important events are triggered from there.

Thanks,

Johannes

### DIFF
--- a/IPython/html/static/notebook/js/textcell.js
+++ b/IPython/html/static/notebook/js/textcell.js
@@ -245,6 +245,28 @@ define([
 
     MarkdownCell.prototype = Object.create(TextCell.prototype);
 
+    /** @method bind_events **/
+    MarkdownCell.prototype.bind_events = function () {
+        TextCell.prototype.bind_events.apply(this);
+        var that = this;
+        this.events.on("rendered.MarkdownCell", function(event, data) {
+            var that = data.cell;
+            that.element.find("input").each( function (i, obj) {
+                $(obj).change({'num':i},function (event) {
+                    var text = that.get_text();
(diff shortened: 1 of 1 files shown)


jdfreder:
We've considered maintaining an IPython Markdown flavor in the past and decided the costs outweigh the benefits.  My opinion is that if marked.js implements the markdown to html checkboxes, maybe https://github.com/chjj/marked/pull/587, then we can think about implementing the reverse in the notebook.

However I won't close the PR, giving others a chance to weigh in.

minrk:
The point that @jojoelfe makes about clicking the checkbox modifying the markdown source means it can't feasibly be done *solely* in marked. I'm not sure what that means we should do about it.

jdfreder:
My suggestion was to wait for marked to take the first step, and then implement our part (HTML -> MD).

minrk:
@jdfreder I think that makes sense.

shorty66:
Seems like [chjj/marked](https://github.com/chjj/marked/graphs/contributors) is no longer maintained. We should consider switching to one of the maintained forks.

PR #8405, opened 2015-05-05
FIX : flip atexit registration

tacaswell:
Register the sphinx spcefic atexit function before InteractiveShell
can register it's clean up function.  The functions are executed
last-in, first-out which means in this case the temporary profile
directory is deleted before the final history writing can be done which
results in exceptions/warnings coming out of sphinx.


minrk:
:+1:

PR #8409, opened 2015-05-06
Use mock and testpath for test_paths

takluyver:
This removes our handcoded monkeypatching/restoring code in test_paths, replacing it with context managers from unittest.mock and testpath. `testpath.modified_env` has one advantage over `unittest.mock.patch.dict`: it can temporarily _unset_ specific environment variables as well as setting them.

I also switched a couple of tests to use `assert_isfile` and `assert_isdir` from testpath. These functions should be more informative when the assertion fails than simply asserting on the equivalent os.path functions.

If this is accepted, I'll move testpath to Jupyter maintenance, as we discussed.

### DIFF
--- a/IPython/core/tests/test_paths.py
+++ b/IPython/core/tests/test_paths.py
@@ -1,4 +1,3 @@
-from contextlib import contextmanager
 import errno
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


minrk:
I like this. testpath is pretty slick.

takluyver:
Thanks!

rgbkrk:
Oh my gosh, testpath is really handy. Thanks @takluyver. I'll be using this in other places.

takluyver:
You're welcome :-)

PR #8420, opened 2015-05-08
jupyter_nbformat is now nbformat

minrk:
jupyter/nbformat#6


Carreau:
fails on Py3.5, but not your fault.

PR #8425, opened 2015-05-11
Replace dead link to nose website in install.rst

dwinston:
Hi, I noticed that the existing link (http://somethingaboutorange.com/mrl/projects/nose/) is dead, so I propose changing the link to https://nose.readthedocs.org, the first hit on a Google search for 'python nose'.

After my first commit, I tried a second commit that would reverse the spurious change of adding a newline to the end of file, but it looks like my second commit was a no-op. Sorry. I don't know how to undo the adding of a newline--I only indended to change the hyperlink to nose.

### DIFF
--- a/docs/source/install/install.rst
+++ b/docs/source/install/install.rst
@@ -220,7 +220,7 @@ Another way of getting this is to do:
     $ pip install "ipython[test]"
 
 For more installation options, see the `nose website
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! Your editor is probably configured to ensure the file ends in a newline when you save - that's not a problem.

PR #8427, opened 2015-05-12
fix tests

Carreau:
Not sure about the  `.pyc` vs `.pyw` as things mention `.pyw` below which is a window specific thing.

One test was wrong, and I also change `assert`  for the better `_is_none` or `_is_equal` equivalent.

### DIFF
--- a/IPython/utils/tests/test_module_paths.py
+++ b/IPython/utils/tests/test_module_paths.py
@@ -25,6 +25,8 @@ from IPython.testing.tools import make_tempfile
 
 import IPython.utils.module_paths as mp
 
+import nose.tools as nt
+
 env = os.environ
 TEST_FILE_PATH = split(abspath(__file__))[0]
 TMP_TEST_DIR = tempfile.mkdtemp()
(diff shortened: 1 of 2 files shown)


IPython/utils/tests/test_module_paths.py
```diff
@@ -79,47 +81,48 @@ def test_get_init_2():
...
 
 def test_find_mod_4():
     modpath = join(TMP_TEST_DIR, "pack.py")
-    assert mp.find_mod("pack") == modpath
+    nt.assert_equal(mp.find_mod("pack"), modpath)
 
 def test_find_mod_5():
-    assert mp.find_mod("packpyc") is None
+    modpath = join(TMP_TEST_DIR, "packpyc.pyc")
+    nt.assert_equal(mp.find_mod("packpyc"), modpath)
```
Carreau: This test was wrong after the fix.

damianavila:
:+1:

IPython/utils/tests/test_module_paths.py
rgbkrk: good catch

PR #8428, opened 2015-05-12
Add history_load_length configurable option

### LINKED ISSUE #8426: Load more history into readline buffer at startup
I'm curious why, at startup in core/interactiveshell.py, the readline history is populated with a hard coded value of only 1000 (instead of, say, the configuration parameter history_length) history entries.

I've personally wished I had more history.

aerusso:
This resolves #8426 with a configurable option "history_load_length" determining the number of history entries loaded into the readline buffer at IPython startup, as @takluyver suggested.

I'm currently running this patch on my own machine.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -393,7 +393,7 @@ class InteractiveShell(SingletonConfigurable):
 
     history_load_length = Integer(1000, config=True, help=
         """
-        Determines the number of saved history entries to be loaded
+        The number of saved history entries to be loaded
         into the readline buffer at startup.
         """
     )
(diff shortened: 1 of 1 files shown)


IPython/core/interactiveshell.py
```diff
@@ -391,6 +391,13 @@ def _prompt_trait_changed(self, name, old, new):
 
     history_length = Integer(10000, config=True)
 
+    history_load_length = Integer(1000, config=True, help=
+        """
+        Determines the number of saved history entries to be loaded
```
takluyver: Documentation style point - I'd prefer to leave out the word 'Determines', so this would start 'The number of ...'.

takluyver:
Apart from that, this looks great. :-)

rgbkrk:
Followed up with Thomas' doc style point, merging.

takluyver:
Thanks @aerusso , and congratulations on your first PR being merged! :tada:

PR #8429, opened 2015-05-13
Backport jupyter/notebook#89

Carreau:
fix highlighting of non-python kernels.

mock the cm_config property also when set, unlike in non backported
version of this PR


Carreau:
This backport jupyter/notebook#94 also now, and fix the tests. 
marking as 3.2.

Issue #1394, opened 2012-02-09
Wishlist: Remove hard dependency on ctypes


bfroehle:
`IPython.lib.inputhook` has a hard dependency on `ctypes`.

Recently I've been working with Python on a machine without ctypes.  I was able to get a basic IPython terminal by commenting out much of the file.  Obviously this means I won't have any `%gui` magic abilities, but that's not a problem in my situation.

Perhaps instead of this hard dependency we should consider refactoring the class so that InputHookManager was a pluggable configuration option?  Would this be a realistic project?

Lucretiel:
It looks like this hasn't been 100% resolved. I get the following message when running on a machine without ctypes:

```
WARNING: IPython GUI event loop requires ctypes, %gui will not be available
Traceback (most recent call last):
  File ".../env/bin/ipython", line 4, in <module>
    from IPython import start_ipython
  File ".../env/lib/python3.4/site-packages/IPython/__init__.py", line 49, in <module>
    from .terminal.embed import embed
  File ".../env/lib/python3.4/site-packages/IPython/terminal/embed.py", line 19, in <module>
    from IPython.terminal.ipapp import load_default_config
  File ".../env/lib/python3.4/site-packages/IPython/terminal/ipapp.py", line 46, in <module>
    from IPython.core.shellapp import (
  File ".../env/lib/python3.4/site-packages/IPython/core/shellapp.py", line 27, in <module>
    from IPython.lib.inputhook import guis
  File ".../env/lib/python3.4/site-packages/IPython/lib/inputhook.py", line 283, in <module>
    class NullInputHook(InputHookBase):
  File ".../env/lib/python3.4/site-packages/IPython/lib/inputhook.py", line 202, in decorator
    self.guihooks[toolkitname] = inst
AttributeError: 'InputHookManager' object has no attribute 'guihooks'
```

The issue specifically is that it's trying to apply the `register` decorator to the input hook classes

```
def register(self, toolkitname, *aliases):
    ''' DOCSTRING """
    def decorator(cls):
        inst = cls(self)
        self.guihooks[toolkitname] = inst
        for a in aliases:
            self.aliases[a] = toolkitname
        return cls
    return decorator
```
-------------------------
```
inputhook_manager = InputHookManager()

@inputhook_manager.register('osx')
class NullInputHook(InputHookBase):
    """A null inputhook that doesn't need to do anything"""
    def enable(self, app=None):
        pass
```

[...]

PR #8430, opened 2015-05-13
register now checks for missing ctypes

Lucretiel:
If ctypes is None, then no input hooks may be registered; `InputHookManager.register` skips registration of input hook classes. Also updated `__init__` to no longer skip creating the instance attributes, to prevent AttributeError exceptions at load time.

This fixes an issue discovered in the fix to #1394


takluyver:
Looks reasonable to me, thanks.

PR #8433, opened 2015-05-14
jupyter_notebook renamed to just notebook

takluyver:


IPython/html.py
```diff
@@ -17,12 +17,12 @@
     src='IPython.html.widgets', mirror='ipython_widgets')
```
minrk: do you want to catch ipywidgets here, while you're at it?

takluyver: Done.

Issue #8431, opened 2015-05-13
Decide on shell-style argument parsing system for magics


takluyver:
Our magics are currently split between two different argument parsing systems: getopt, and the argparse+decorator based magic_arguments. The latter is more powerful, and has the advantage that the documentation of individual flags is with their definition, but I think the big stacks of decorators it uses are really ugly, and it's more complex for new developers to understand the flow of the code because there's so much indirection.

A few options:

1. Go back to getopt, deprecate magic_arguments
2. Go all in on magic_arguments
3. Use argparse *inside* magic functions, rather than with decorators. Make `%timeit?` introspection work by invoking `%timeit --help` rather than getting the docstring (because the docstring can't be built by code inside the function).
4. Use something else entirely, e.g. [clize](http://clize.readthedocs.org/en/3.0/) or [docopt](http://docopt.org/).

Ping @fperez , we were just talking about this.

PR #8435, opened 2015-05-14
[NOT FOR MERGING] Timeit output

fperez:
This is just so we can study whether argparse will do the trick for magics in general, or not.

The original intent of this PR, started with @minrk, was to add an output flag to the timeit magics.

I (mistakenly, as it turns out) thought argparse could handle a simple `-o` and `-o NAME` without ambiguity, which getopt definitely can't, so I bit the bullet and rewrote `%timeit` to use argparse.

In the process, it turned out that I couldn't get the argparse-based version to correctly pass the tests, due to some inconsistencies in parsing input with python code in it.  I then ran out of time debugging the problem.

If we can make this work, we can merge this PR, and then we can use the experience to consider whether we want to unify (at least for new/updated code) on argparse in our magics option handling code.

**Important note:** bugs aside, this PR isn't finished in that, after all, we can NOT have `-o` handle both the no-args and the named-arg case, since a case like `-o pass` would be incorrectly interpreted as assigning to a variable named `pass` (unless we resorted to crazy shenanigans in the parsing code).

So in the end, if we can sort the other bugs, we need to handle the named output with a _separate_ flag, say `-O`. I was going to do that but ran out of time with the parsing bugs.

Pinging @Carreau @takluyver from today's conversation.

### DIFF
--- a/IPython/core/magic_arguments.py
+++ b/IPython/core/magic_arguments.py
@@ -132,7 +132,7 @@ class MagicArgumentParser(argparse.ArgumentParser):
     def parse_argstring(self, argstring):
         """ Split a string into an argument list and parse that argument list.
         """
-        argv = arg_split(argstring)
+        argv = arg_split(argstring, strict=False)
         return self.parse_args(argv)
 
 
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -895,50 +895,50 @@ python-profiler package from non-free.""")
     @skip_doctest
     @magic_arguments.magic_arguments()
     @magic_arguments.argument('-n', '--number', type=int, default=0,
-        help="""11
+        help="""execute the given statement NUMBER times in a loop. If this value
+        is not given, a fitting value is chosen.
         """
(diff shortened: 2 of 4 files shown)


takluyver:
Cross link #8431

Carreau:
Might also be of interest http://click.pocoo.org/4/

Carreau:
BTW, i didn't now but optparse is [deprecated since python 3.2](https://docs.python.org/3.5/library/optparse.html)   apparently...

takluyver:
We're not using optparse anywhere, as far as I know. We use a mixture of getopt and argparse, neither of which are marked as deprecated.

Carreau:
AH, sorry getopt.. last discussion we had I remembered speaking about optparse. my bad.

Carreau:
This need rebase.

Carreau:
I'm going to close as no-one is working on that and this clutter the PR list. 
I'm tagging as "Closed PR" to remind us to look at it at some point. Also there is linked issue already.

Issue #8444, opened 2015-05-18
%matplotlib magic: print a list of available backends


cel4:
Is there an easy way to print a list of all available backends? Starting the interpreter with `ipython --matplotlib foo` prints an error message with all available backends. Is there something similar for the `%matplotlib` magic? Wouldn't it be helpful to have something like `%matplotlib --list` to print a list of all available backends?

Carreau:
I don't see any harm in having that. `%matplotlib?` give the list at the bottom already.

Carreau:
Do you want to try a PR ?

cel4:
I just wanted to leave the idea here - I have very little time at the moment. But this is not urgent - so I may put together a PR at some time.

koenvb:
Mind if I give it a try. I am quite a beginner but I assume this is not so hard to do?

takluyver:
It shouldn't be too difficult, and we're quite happy whoever implements it. :-)

The code for the `%matplotlib` and `%pylab` magics is in `IPython.core.magics.pylab`. If you want to know anything else, feel free to ask us, either here on our gitter chat room at https://gitter.im/ipython/ipython

koenvb:
I added something for the %matplotlib magic

```
In [3]: %matplotlib -l
Available matplotlib backends: {'osx': 'MacOSX', 'qt4': 'Qt4Agg', 'qt5': 'Qt5Agg', 'gtk3': 'GTK3Agg', 'notebook': 'nbAgg', 'wx': 'WXAgg', 'qt': 'Qt4Agg', 'nbagg': 'nbAgg', 'gtk': 'GTKAgg', 'tk': 'TkAgg', 'inline': 'module://ipykernel.pylab.backend_inline'}

In [4]: %matplotlib --list
Available matplotlib backends: {'osx': 'MacOSX', 'qt4': 'Qt4Agg', 'qt5': 'Qt5Agg', 'gtk3': 'GTK3Agg', 'notebook': 'nbAgg', 'wx': 'WXAgg', 'qt': 'Qt4Agg', 'nbagg': 'nbAgg', 'gtk': 'GTKAgg', 'tk': 'TkAgg', 'inline': 'module://ipykernel.pylab.backend_inline'}

```

Carreau:
Closed by #8449 and #8506

PR #8449, opened 2015-05-20
added %matplotlib --list so backends are easily visible

koenvb:
linked to #8444 adding --list argument to show the available backends

### DIFF
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -42,6 +42,8 @@ class PylabMagics(Magics):
     @skip_doctest
     @line_magic
     @magic_arguments.magic_arguments()
+    @magic_arguments.argument('-l', '--list', default=None, action='store_true',
+                              help='Show available matplotlib backends')
     @magic_gui_arg
     def matplotlib(self, line=''):
         """Set up matplotlib to work interactively.
@@ -84,9 +86,15 @@ class PylabMagics(Magics):
 
             In [3]: %matplotlib qt
         """
+
         args = magic_arguments.parse_argstring(self.matplotlib, line)
-        gui, backend = self.shell.enable_matplotlib(args.gui)
-        self._show_matplotlib_backend(args.gui, backend)
+
+        if args.list is not None:
+            backends_list = list(backends.keys())
+            print("Available matplotlib backends: %s" % backends_list)
+        else:
+            gui, backend = self.shell.enable_matplotlib(args.gui)
+            self._show_matplotlib_backend(args.gui, backend)
 
     @skip_doctest
     @line_magic
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -42,7 +42,7 @@ class PylabMagics(Magics):
     @skip_doctest
     @line_magic
     @magic_arguments.magic_arguments()
-    @magic_arguments.argument('-l', '--list', default=None, action='store_true',
+    @magic_arguments.argument('-l', '--list', default=False, action='store_true',
                               help='Show available matplotlib backends')
     @magic_gui_arg
     def matplotlib(self, line=''):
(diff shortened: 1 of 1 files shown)


koenvb:
Seems like 

--l 
--li
--lis
--list

are all giving the same result. Dunno exactly why? And if its appropriate

```
In [4]: %matplotlib --l
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']

In [5]: %matplotlib --li
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']

In [6]: %matplotlib --lis
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']

In [7]: %matplotlib --list
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']
```

IPython/core/magics/pylab.py
```diff
@@ -84,9 +86,15 @@ def matplotlib(self, line=''):
 
             In [3]: %matplotlib qt
         """
+
         args = magic_arguments.parse_argstring(self.matplotlib, line)
-        gui, backend = self.shell.enable_matplotlib(args.gui)
-        self._show_matplotlib_backend(args.gui, backend)
+
+        if args.list is not None:
```
Carreau: My guess is as as you do `store_true ` above `args.list` can only be `True̵̵̵|False` so the `... is None` is alway True, and you get into the if.

Carreau:
Do you want to take a stab at fixing the `--list` thing, or should one of us take over ? (There is no hurry)

koenvb:
Will give it another try. can I just update this one then, or should I create a new pull request?

koenvb:
I was looking into it a bit more and I get a bit confused. It feels like when using --l or --li or --lis
the arg.list is always True and I would expect its only True when -l or --list?

So adding these print statements to the code:
(I do not know how to run a debugger in this case? any workflow help is appreciated)
```
print(line)
args = magic_arguments.parse_argstring(self.matplotlib, line)
print(args)
print(args.list)
```

Gives:

```
In [5]: %matplotlib --l
--l
Namespace(gui=None, list=True)
True
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']

In [6]: %matplotlib --li
--li
Namespace(gui=None, list=True)
True
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']

In [7]: %matplotlib --lis
--lis
Namespace(gui=None, list=True)
True
Available matplotlib backends: ['osx', 'qt4', 'qt5', 'gtk3', 'notebook', 'wx', 'qt', 'nbagg', 'gtk', 'tk', 'inline']

In [8]: %matplotlib --lisg
--lisg
UsageError: unrecognized arguments: --lisg
```

I would assume that the 
```args = magic_arguments.parse_argstring(self.matplotlib, line) function```
should reject if it is not -l or --list?

But I am probably wrong here...

**Comment by Carreau** [comment]
we use argparse, and it seem to be [a feature](https://docs.python.org/3.5/library/argparse.html#argument-abbreviations-prefix-matching), we could [disable it](https://docs.python.org/3.5/library/argparse.html#allow-abbrev) though.

And Yes, usually you can just push on github it updates the PR.
If its an argparse problem, we can (decide to) fix  it later.

**Comment by Carreau** [comment]
@takluyver a look, you were mentioned in the issue ? do you think list for for backend is fine, or should we refine as bullet points ?

File: IPython/core/magics/pylab.py (hunk 1 above)
Comment (takluyver) [inline IPython/core/magics/pylab.py]: I think this should just be `if args.list`, then you shouldn't need the default above (with `store_true` the default should automatically be False).

**Comment by takluyver** [comment]
I think a list is fine. I've made one minor suggestion inline.

**Comment by koenvb** [comment]
I updated the line and added something to the docs. 
Let me know if all is fine!

File: IPython/core/magics/pylab.py [hunk 2]
```diff
@@ -42,6 +42,8 @@ class PylabMagics(Magics):
     @skip_doctest
     @line_magic
     @magic_arguments.magic_arguments()
+    @magic_arguments.argument('-l', '--list', default=False, action='store_true',
```
takluyver: You don't actually need the `default=False` here - when you use store_true, the default is automatically False.

Carreau: "Explicit is better than implicit" might be objected :-)

takluyver: I think `'store_true'` makes it clear enough what's going on. All the examples in the argparse docs using store_true don't bother with a default.

HerlonNascimento: 'store_true' makes it clear enough what's going on

takluyver:
Thanks @koenvb

PR #8321, opened 2015-04-14
Test displayhook

ghost:
This is a work in progress and the tests will fail. We will fix in a later commit. This is a test for #8002.

PR #8457, opened 2015-05-22
updated quiet function in the displayhook.py

ghost:
Updated quiet function in the displayhook.py to properly recognize comment and semicolon by using the tokenize module. Thanks to @takluyver for heavy input and assistance in writing this code.


koenvb:
@thethomask I guess something went wrong with this commit. As it is an empty file?

thethomask:
Wat. Empty? I don't know what's going on. I'll look into it.

takluyver:
@thethomask , `test_displayhook.py` is the file you already added in #8321. Do you want to add to that pull request the changes to `displayhook.py` that you put in [this gist](https://gist.github.com/thethomask/fb3a9a5a6dfb385d18ac).

thethomask:
@takluyver Yes. Should be there now.

IPython/core/tests/test_displayhook.py
```diff
@@ -0,0 +1 @@
+code
```
takluyver: We'll need to merge manually to remove this file. (making a note so I remember)

IPython/core/displayhook.py
```diff
@@ -83,12 +84,18 @@ def check_for_underscore(self):
...
         """Should we silence the display hook because of ';'?"""
         # do not print output if input ends in ';'
-        try:
-            cell = self.shell.history_manager.input_hist_parsed[self.prompt_count]
-            return cell.rstrip().endswith(';')
-        except IndexError:
-            # some uses of ipshellembed may fail here
-            return False
+
+	cell = self.shell.history_manager.input_hist_parsed[self.prompt_count]
```
takluyver: The code that was here before catches `IndexError`, in case `input_hist_parsed` doesn't have that many entries. I'm not sure if that can actually still happen, but we should probably still catch that error, just to be on the safe side.

thethomask:
@takluyver updated

takluyver:
Thanks! The try/except actually needs to be around the `...input_hist_parsed[self.prompt_count]` line, though, not the bit that iterates through tokens.

IPython/core/displayhook.py
```diff
@@ -83,9 +84,19 @@ def check_for_underscore(self):
     def quiet(self):
         """Should we silence the display hook because of ';'?"""
         # do not print output if input ends in ';'
+ 
+	cell = self.shell.history_manager.input_hist_parsed[self.prompt_count]
+        sio = io.StringIO(cell)
```
Carreau: Small indent issue here `cell =` seem to miss one leading space.

Carreau:
and the travis error is `TabError: inconsistent use of tabs and spaces in indentation` so you might have a mixing of tabs and space I guess :-)

takluyver:
Oh, I hadn't spotted that. Yes, it should indented using four spaces per indent level.

IPython/core/displayhook.py
```diff
@@ -83,9 +84,19 @@ def check_for_underscore(self):
...
+        tokens = list(tokenize.generate_tokens(sio.readline))
+ 
         try:
-            cell = self.shell.history_manager.input_hist_parsed[self.prompt_count]
-            return cell.rstrip().endswith(';')
+    	    for token in reversed(tokens):
+                if token.type in (tokenize.ENDMARKER, tokenize.COMMENT):
+            	    continue
+                if (token.type == tokenize.OP) and (token.string == ';'):
```
damianavila: I would remove all the parenthesis on this line...

takluyver: I tend to prefer the parentheses: I can never remember exactly which
operators have precedence, so explicit is better than implicit.

damianavila: not a big deal, if you prefer them, I am ok with that...

IPython/core/displayhook.py
```diff
@@ -83,9 +84,19 @@ def check_for_underscore(self):
...
         # do not print output if input ends in ';'
+ 
+	cell = self.shell.history_manager.input_hist_parsed[self.prompt_count]
+        sio = io.StringIO(cell)
+        tokens = list(tokenize.generate_tokens(sio.readline))
+ 
         try:
-            cell = self.shell.history_manager.input_hist_parsed[self.prompt_count]
-            return cell.rstrip().endswith(';')
+    	    for token in reversed(tokens):
```
Carreau: I think there is a tab/space inconsistency here too.

PR #8461, opened 2015-05-22
update platform-specific dependencies

minrk:
- specify with extras_require, so bdists are not sensitive to what platform they are built on.
- add backward-compatible spec for old pip,
  but only use this when not running a bdist.


setup.py
```diff
@@ -222,16 +215,28 @@ def run(self):
...
+# but requires pip >= 6. pip < 6 ignores these.
+extras_require.update({
+    ':sys_platform != "win32"': ['pexpect'],
+    ':sys_platform == "darwin"': ['appnope', 'gnureadline'],
+    'terminal:sys_platform == "win32"': ['pyreadline>=2'],
+    'test:python_version == "2.7"': ['mock'],
+})
+# FIXME: re-specify above platform dependencies for pip < 6
+# These would result in non-portable bdists.
+if any(arg.startswith('bdist') for arg in sys.argv):
```
takluyver: not

setupbase.py
```diff
@@ -485,70 +464,3 @@ def _record_commit(self, base_dir):
...
-
-
-#---------------------------------------------------------------------------
-# bdist related
-#---------------------------------------------------------------------------
-
-def get_bdist_wheel():
-    """Construct bdist_wheel command for building wheels
-    
-    Constructs py2-none-any tag, instead of py2.7-none-any
```
takluyver: Do the current tools get the tag right without our help?

takluyver:
This is a definite improvement.

setupbase.py
minrk: Yup, that part became unnecessary a while ago. This function doesn't even do that part anymore.

setup.py
```diff
@@ -222,16 +214,28 @@ def run(self):
...
-if sys.platform.startswith('win'):
-    extras_require['terminal'].append('pyreadline>=2.0')
-else:
-    install_requires.append('pexpect')
+# Platform-specific dependencies:
+# This is the correct way to specify these,
+# but requires pip >= 6. pip < 6 ignores these.
+extras_require.update({
+    ':sys_platform != "win32"': ['pexpect'],
+    ':sys_platform == "darwin"': ['appnope', 'gnureadline'],
```
rgbkrk: Nice, you're bundling appnope.

Carreau:
+1.

damianavila:
LGTM too...

PR #8462, opened 2015-05-22
re-enable install

minrk:
The Big Split is done enough that installation works again.


takluyver:
:+1:

count down to the first 'where did the notebook go?' bug

rgbkrk:
Haha.

PR #8464, opened 2015-05-23
Add pprint support for containers from collections

RazerM:
`for_type_by_name` used to keep compatiblity with Python 2.4 rather than catching ImportError.

`deque` is already imported in _pretty.py_, but uses same mechanism for consistency.

Tests have been added for each type.

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -816,6 +816,51 @@ _singleton_pprinters = dict.fromkeys(map(id, [None, True, False, Ellipsis,
                                       NotImplemented]), _repr_pprint)
 
 
+def _defaultdict_pprint(obj, p, cycle):
+    name = 'defaultdict'
+    p.begin_group(len(name) + 1, name + '(')
+    if cycle:
+        p.text('...')
+    else:
+        p.pretty(obj.default_factory)
+        p.text(',')
+        p.breakable()
+        p.pretty(dict(obj))
+    p.end_group(len(name) + 1, ')')
+
+def _ordereddict_pprint(obj, p, cycle):
+    name = 'OrderedDict'
+    p.begin_group(len(name) + 1, name + '(')
+    if cycle:
+        p.text('...')
+    elif len(obj):
+            p.pretty(list(obj.items()))
+    p.end_group(len(name) + 1, ')')
+
+def _deque_pprint(obj, p, cycle):
+    name = 'deque'
+    p.begin_group(len(name) + 1, name + '(')
+    if cycle:
+        p.text('...')
+    else:
+        p.pretty(list(obj))
+    p.end_group(len(name) + 1, ')')
+
+
+def _counter_pprint(obj, p, cycle):
+    name = 'Counter'
+    p.begin_group(len(name) + 1, name + '(')
+    if cycle:
+        p.text('...')
+    elif len(obj):
+            p.pretty(dict(obj))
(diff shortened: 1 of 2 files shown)


Carreau:
Hum, this seem to break the test on 2.7.

And did I read 2.4 correctly ? IPython have dropped support for 2.4 a long time ago, we don't even support 2.6 anymore.

RazerM:
Hmm there is an error in collections.py when `super(Counter, self)` is called.

I did see that 2.4 is no longer supported, but I decided not to break compatibility considering the [IPython.lib.pretty][prettydocs] docs still provides example for using on 2.4.

[prettydocs]: https://ipython.org/ipython-doc/dev/api/generated/IPython.lib.pretty.html#module-IPython.lib.pretty

takluyver:
We should change that example, I don't want to suggest that any part of IPython supports 2.4.

takluyver:
The code here looks fine at a quick glance, though. And `for_type_by_name` is probably what we want even without 2.4 support; it probably shouldn't import collections just to define pretty displays for those types.

IPython/lib/pretty.py
```diff
@@ -816,6 +790,47 @@ def for_type_by_name(type_module, type_name, func):
...
+            p.breakable()
+            p.pretty(dict(obj))
+
+def _ordereddict_pprint(obj, p, cycle):
+    name = 'OrderedDict'
+    with p.group(len(name) + 1, name + '(', ')'):
+        if cycle:
+            p.text('...')
+        elif len(obj):
+                p.pretty(list(obj.items()))
```
Carreau: indented too much,

IPython/lib/pretty.py
```diff
@@ -816,6 +790,47 @@ def for_type_by_name(type_module, type_name, func):
...
+            p.pretty(list(obj))
+
+
+def _counter_pprint(obj, p, cycle):
+    name = 'Counter'
+    with p.group(len(name) + 1, name + '(', ')'):
+        if cycle:
+            p.text('...')
+        elif len(obj):
+                p.pretty(dict(obj))
```
Carreau: indented too much.

IPython/lib/tests/test_pretty.py
```diff
@@ -268,3 +270,89 @@ def type_pprint_wrapper(obj, p, cycle):
...
+         '       1018,\n'
+         '       1019])'),
+        (a, 'deque([deque(...)])'),
+    ]
+    for obj, expected in cases:
+        nt.assert_equal(pretty.pretty(obj), expected)
+
+def test_collections_counter():
+    cases = [
+        (Counter(), 'Counter()'),
```
Carreau: this first test case seem to fail on 2.7

Carreau: Actually i don't see why this fail, it is actually `Counter()` that fails, and it has no reason to, unless Stdlib is monkeypatch somewhere else. I kicked the build.

Carreau:
@RazerM don't forget to ping us when you add commit :-) 

I've merged the fix to deepreload that was making test fails. Relaunched the test: All green.

+1

RazerM:
@Carreau Noted, thanks. I wrongly assumed the commit would ping you.

rgbkrk:
Looks good to me too. :+1:

takluyver:
Thanks @RazerM

PR #8476, opened 2015-05-26
Remove documentation of machinery that is now in traitlets

takluyver:
And add link to it with intersphinx (depends on ipython/traitlets#18...oh, @Carreau already merged that. Cool!)

Also removed documentation about security files (no longer part of IPython) and startup files (described elsewhere, and not part of the config system).


minrk:
Why did you remove startup file docs? That's part of IPython, and not going anywhere, right?

takluyver:
"described elsewhere, and not part of the config system"

elsewhere is in the introductory docs for interactive use: http://ipython.org/ipython-doc/3/interactive/tutorial.html#startup-files

minrk:
Thanks.

PR #8477, opened 2015-05-26
Remove ipyparallel docs

takluyver:
These now live in the ipyparallel repo

There are a couple of dangling references from whatsnew entries (to parallel_overview and parallel_db). These will be fixed once the parallel docs are built and we can link to them with intersphinx.


Carreau:
Delete all the things.
preview

PR #8480, opened 2015-05-27
register default deprecation warning filter.

### LINKED ISSUE #8478: DeprecationWarnings should be visible by default in interactive usage
DeprecationWarning and PendingDeprecationWarning are invisible by default. The rationale for this is that they are only useful to people who are writing code, so they should not be shown to end-users who are merely running code.

If someone is typing stuff into the interactive REPL or a notebook, though, and the code they type uses some deprecated functionality, then we should actually show them this warning. We know that the author is sitting right there. And they're probably going to take whatever they tested interactively and move it into a more permanent form and ugh.

This problem is particularly acute for packages that are primarily used through the interactive prompt, like numpy. I'm really tired of getting bugs like this: numpy/numpy#5919

The solution is simple: when entering interactive mode, IPython should do something like:
```
warnings.filterwarnings("default", category=DeprecationWarning, module="__main__")
warnings.filterwarnings("default", category=PendingDeprecationWarning, module="__main__")
```
This means "for any `DeprecationWarning` or `PendingDeprecation` warning which is caused by code running in the module with name `"__main__"`, use the `"default"` handler (which means: display once per line of code that triggered it). (Confusingly, the default handler for these warning messages is not `"default"`, but `"ignore"`. So overriding the default with `"default"` produces a non-default outcome.)

Carreau:
warn user if (s)he use a deprecated feature interactively.

Closes #8478

```
In [1]: %run -i foo
In [2]: foo()
foo.py:4: DeprecationWarning: using a non-integer number instead of an
    integer will result in an error in the future
    print('x'*np.float64(3.5))
xxx

In [3]: from foo import foo
In [4]: foo()
xxx
```

---

Not sure it is the exact right place though.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -665,16 +665,15 @@ class InteractiveShell(SingletonConfigurable):
         elif self.logstart:
             self.magic('logstart')
 
-    @staticmethod
-    def init_deprecation_warnings():
+    def init_deprecation_warnings(self):
         """
         register default filter for (pending) deprecation warning.
 
         This will allow deprecation warning of function used interactively to show
         warning to users, and still hide deprecation warning from libraries import.
         """
-        warnings.filterwarnings("default", category=DeprecationWarning, module="__main__")
-        warnings.filterwarnings("default", category=PendingDeprecationWarning, module="__main__")
+        warnings.filterwarnings("default", category=DeprecationWarning, module=self.user_ns.get("__name__"))
+        warnings.filterwarnings("default", category=PendingDeprecationWarning, module=self.user_ns.get("__name__"))
 
 
     def init_builtins(self):
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -25,6 +25,7 @@ import tempfile
 import traceback
 import types
 import subprocess
+import warnings
 from io import open as io_open
 
 from pickleshare import PickleShareDB
@@ -547,6 +548,7 @@ class InteractiveShell(SingletonConfigurable):
         self.init_pdb()
         self.init_extension_manager()
         self.init_payload()
+        self.init_deprecation_warnings()
         self.hooks.late_startup_hook()
         self.events.trigger('shell_initialized', self)
         atexit.register(self.atexit_operations)
@@ -662,6 +664,18 @@ class InteractiveShell(SingletonConfigurable):
             self.magic('logstart %s' % self.logfile)
         elif self.logstart:
             self.magic('logstart')
+
+    @staticmethod
+    def init_deprecation_warnings():
+        """
+        register default filter for (pending) deprecation warning.
+
+        This will allow deprecation warning of function used interactively to show
+        warning to users, and still hide deprecation warning from libraries import.
+        """
+        warnings.filterwarnings("default", category=DeprecationWarning, module="__main__")
+        warnings.filterwarnings("default", category=PendingDeprecationWarning, module="__main__")
+
 
     def init_builtins(self):
         # A single, static flag that we set to True.  Its presence indicates
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -667,14 +667,12 @@ class InteractiveShell(SingletonConfigurable):
 
     def init_deprecation_warnings(self):
         """
-        register default filter for (pending) deprecation warning.
+        register default filter for deprecation warning.
 
         This will allow deprecation warning of function used interactively to show
         warning to users, and still hide deprecation warning from libraries import.
         """
         warnings.filterwarnings("default", category=DeprecationWarning, module=self.user_ns.get("__name__"))
-        warnings.filterwarnings("default", category=PendingDeprecationWarning, module=self.user_ns.get("__name__"))
-
 
     def init_builtins(self):
         # A single, static flag that we set to True.  Its presence indicates
--- a/IPython/core/tests/test_interactiveshell.py
+++ b/IPython/core/tests/test_interactiveshell.py
@@ -901,3 +901,20 @@ def test_warning_suppression():
             ip.run_cell("warnings.warn('asdf')")
     finally:
         ip.run_cell("del warnings")
+
+
+def test_deprecation_warning():
+    ip.run_cell("""
+import warnings
+def wrn():
+    warnings.warn(
+        "I AM  A WARNING",
+        DeprecationWarning
+    )
+        """)
+    try:
+        with tt.AssertPrints("I AM  A WARNING", channel="stderr"):
+            ip.run_cell("wrn()")
+    finally:
+        ip.run_cell("del warnings")
+        ip.run_cell("del wrn")
--- a/IPython/core/tests/test_interactiveshell.py
+++ b/IPython/core/tests/test_interactiveshell.py
@@ -918,3 +918,27 @@ def wrn():
     finally:
         ip.run_cell("del warnings")
         ip.run_cell("del wrn")
+
+
+class TestImportNoDeprecate(tt.TempFileMixin):
+
+    def setup(self):
+        """Make a valid python temp file."""
+        self.mktmp("""
+import warnings
+def wrn():
+    warnings.warn(
+        "I AM  A WARNING",
+        DeprecationWarning
+    )
+""")
+
+    def test_no_dep(self):
+        """
+        No deprecation warning should be raised from imported functions
+        """
+        ip.run_cell("from {} import wrn".format(self.fname))
+
+        with tt.AssertNotPrints("I AM  A WARNING"):
+            ip.run_cell("wrn()")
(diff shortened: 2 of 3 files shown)


Carreau:
Test added. 

Would @njsmith like to try ? And would that suit him.

minrk:
It is possible for code to be run non-interactively with InteractiveShell, but I think it is appropriate for any use of InteractiveShell to behave as if interactively executed.

:+1: to this

Carreau:
Yes you can run things with execute preprocessor, but then you still want stderr to be captured in the notebook (for the notebook case). So +1 on behaving the same. We can refine the behavior. 

I should add a note to the what's new.

Carreau:
Hum, importing a deprecated module in main does not print the warning either...

Carreau:
Side question, why in #6680 do we go for clearing the registry instead of "always" filters ?

njsmith:
1) we don't have total control over the warning filter; the hack there
makes the existing filter work right.
2)
for i in range(100):
    do_something_deprecated()
^^ with "default" filters and #6680 this will emit one warning each time
the cell is executed. With "always" it will emit 100 warnings every time
the cell is executed.

Carreau:
Ah, make sens !

njsmith:
> Hum, importing a deprecated module in main does not print the warning either...

Ah, this appears to be because the import system rewrite in CPython has totally screwed up module deprecation.

A key feature of warnings is that they are *attributed* -- they are supposed to contain a pointer to the code that needs fixing. This is done via the `stacklevel` argument to `warnings.warn` -- the default `warnings.warn("foo", stacklevel=1)` attributes the warning to the line of code that called `warnings.warn`, which is basically always wrong; or you can use a higher value of `stacklevel=` to attribute the warning to your caller, or your caller's caller, etc. E.g., here's the correct way to warn for a deprecated function:
```
def foo():
    warnings.warn("foo() is deprecated", DeprecationWarning, stacklevel=2)

foo()
```
Here the warning will be attributed to the last line of code, not to the body of `foo`.

This is crucial for our warnings filter -- we're requesting that we only see warnings that are attributed to code running in the `__main__` module.

So if you call `warnings.warn` at the top-level inside a module, then what happens? Who is the 'caller' of the body of a module? Let's find out:
```python
# w.py
import sys
import warnings
import traceback

sys.stdout.write("Nominal call stack:\n")
traceback.print_stack()

sys.stdout.write("\nWarnings:\n")
for i in range(1, 12):
    warnings.warn("At import, stacklevel %s" % (i,), stacklevel=i)
    sys.stdout.write("\n")
```

Output on python 2:
```
>>> import w
Nominal call stack:
  File "w.py", line 6, in <module>
    traceback.print_stack()

Warnings:
w.py:10: UserWarning: At import, stacklevel 1
  warnings.warn("At import, stacklevel %s" % (i,), stacklevel=i)

__main__:1: UserWarning: At import, stacklevel 2

sys:1: UserWarning: At import, stacklevel 3

[...trimmed...]
```
so `stacklevel=2` does the right thing, as you'd expect.

On python 3.3:
```
>>> import w
Nominal call stack:
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1532, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 584, in _check_name_wrapper
  File "<frozen importlib._bootstrap>", line 1022, in load_module
  File "<frozen importlib._bootstrap>", line 1003, in load_module
  File "<frozen importlib._bootstrap>", line 560, in module_for_loader_wrapper
  File "<frozen importlib._bootstrap>", line 868, in _load_module
  File "<frozen importlib._bootstrap>", line 313, in _call_with_frames_removed
  File "./w.py", line 6, in <module>
    traceback.print_stack()

Warnings:
./w.py:10: UserWarning: At import, stacklevel 1
  warnings.warn("At import, stacklevel %s" % (i,), stacklevel=i)

/usr/lib/python3.3/importlib/_bootstrap.py:313: UserWarning: At import, stacklevel 2
  return f(*args, **kwds)

/usr/lib/python3.3/importlib/_bootstrap.py:868: UserWarning: At import, stacklevel 3
  _call_with_frames_removed(exec, code_object, module.__dict__)

/usr/lib/python3.3/importlib/_bootstrap.py:560: UserWarning: At import, stacklevel 4
  return fxn(self, module, *args, **kwargs)

/usr/lib/python3.3/importlib/_bootstrap.py:1003: UserWarning: At import, stacklevel 5
  return self._load_module(fullname)

/usr/lib/python3.3/importlib/_bootstrap.py:1022: UserWarning: At import, stacklevel 6
  return super(FileLoader, self).load_module(fullname)

/usr/lib/python3.3/importlib/_bootstrap.py:584: UserWarning: At import, stacklevel 7
  return method(self, name, *args, **kwargs)

/usr/lib/python3.3/importlib/_bootstrap.py:1532: UserWarning: At import, stacklevel 8
  loader.load_module(name)

/usr/lib/python3.3/importlib/_bootstrap.py:1565: UserWarning: At import, stacklevel 9
  return _find_and_load_unlocked(name, import_)

__main__:1: UserWarning: At import, stacklevel 10

sys:1: UserWarning: At import, stacklevel 11
```
So on python 3.3, if you want to correctly deprecate a module, you have to write `warnings.warn("this module is deprecated", DeprecationWarning, stacklevel=10)`. Duh.

And on python 3.4, it changed again:
```
>>> import w
Nominal call stack:
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1471, in exec_module
  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed
  File "/tmp/w.py", line 6, in <module>
    traceback.print_stack()

Warnings:
/tmp/w.py:10: UserWarning: At import, stacklevel 1
  warnings.warn("At import, stacklevel %s" % (i,), stacklevel=i)

/usr/lib/python3.4/importlib/_bootstrap.py:321: UserWarning: At import, stacklevel 2
  return f(*args, **kwds)

/usr/lib/python3.4/importlib/_bootstrap.py:1471: UserWarning: At import, stacklevel 3
  _call_with_frames_removed(exec, code, module.__dict__)

/usr/lib/python3.4/importlib/_bootstrap.py:1129: UserWarning: At import, stacklevel 4
  self.spec.loader.exec_module(module)

/usr/lib/python3.4/importlib/_bootstrap.py:1200: UserWarning: At import, stacklevel 5
  self._exec(module)

/usr/lib/python3.4/importlib/_bootstrap.py:2226: UserWarning: At import, stacklevel 6
  module = _SpecMethods(spec)._load_unlocked()

/usr/lib/python3.4/importlib/_bootstrap.py:2237: UserWarning: At import, stacklevel 7
  return _find_and_load_unlocked(name, import_)

__main__:1: UserWarning: At import, stacklevel 8

sys:1: UserWarning: At import, stacklevel 9

sys:1: UserWarning: At import, stacklevel 10

sys:1: UserWarning: At import, stacklevel 11
```

So basically the conclusion is that it's not possible to correctly deprecate a module on Python 3. Sweet!

njsmith:
Filed a bug upstream: https://bugs.python.org/issue24305

IPython/core/interactiveshell.py
```diff
@@ -663,6 +665,18 @@ def init_logstart(self):
...
+    @staticmethod
+    def init_deprecation_warnings():
+        """
+        register default filter for (pending) deprecation warning.
+
+        This will allow deprecation warning of function used interactively to show
+        warning to users, and still hide deprecation warning from libraries import.
+        """
+        warnings.filterwarnings("default", category=DeprecationWarning, module="__main__")
+        warnings.filterwarnings("default", category=PendingDeprecationWarning, module="__main__")
```
njsmith: Maybe this should instead say
```
module=self.user_ns.get("__name__")
```
?

njsmith:
Anyway, chatter aside, this LGTM except for the minor comment about using `user_ns["__name__"]`.

Carreau:
> Anyway, chatter aside, this LGTM except for the minor comment about using user_ns["__name__"].

done.

Carreau:
Waiting on CPython to choose a behavior. 
We might deactivate this for PendingDW IIUC. 

See mail thread https://mail.python.org/pipermail/python-ideas/2015-May/033740.html and bug on CPython http://bugs.python.org/issue24294

minrk:
I'm 50/50 on whether we unhide PendingDeprecation. Based on [Nick's comment](https://mail.python.org/pipermail/python-ideas/2015-May/033743.html), I might be slightly inclined to continue ignoring them, and only unhide actual deprecations.

Carreau:
Only show deprecation warnings now.

PR #8483, opened 2015-05-28
respect keepkernel on exit for zmqshell in console

### LINKED ISSUE #8482: quit(keep_kernel=True) doesn't work
I'm using IPython 3.1.0

Carreau:
closes #8482

---

this is against 3x.

### DIFF
--- a/IPython/terminal/console/interactiveshell.py
+++ b/IPython/terminal/console/interactiveshell.py
@@ -212,6 +212,7 @@ class ZMQTerminalInteractiveShell(TerminalInteractiveShell):
                     elif source == 'set_next_input':
                         self.set_next_input(item['text'])
                     elif source == 'ask_exit':
+                        self.keepkernel=item.get('keepkernel', False)
                         self.ask_exit()
                
             elif status == 'error':
@@ -233,7 +234,7 @@ class ZMQTerminalInteractiveShell(TerminalInteractiveShell):
         Only relevant if include_other_output is True.
         """
     )
-    
+
     def from_here(self, msg):
         """Return whether a message is from this session"""
         return msg['parent_header'].get("session", self.session_id) == self.session_id
@@ -409,6 +410,7 @@ class ZMQTerminalInteractiveShell(TerminalInteractiveShell):
                 self.client.input(raw_data)
 
     def mainloop(self, display_banner=False):
+        self.keepkernel = False
         while True:
             try:
                 self.interact(display_banner=display_banner)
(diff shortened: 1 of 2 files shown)


takluyver:
This looks like it will stop the frontend from exiting if `keep_kernel` is True. That's not what should happen - the frontend should exit and leave the kernel running.

Carreau:
Ok, what about that ?

takluyver:
I think this now looks functionally correct, but:

- I wouldn't change anything about the Ctrl-C handling. Ctrl-D is the shortcut to quit.
- I'd like the logic to be simpler overall before we merge it.

Carreau:
> I wouldn't change anything about the Ctrl-C handling. Ctrl-D is the shortcut to quit.

Ctrd-D is not a signal, and does not raise, so its more complicated to handle, Hand here it is really to kill the program when it's stuck. Ctrl-D already exit if you are connected to a kernel. But here it was to escape the infinite loop if you `--existing` and there is no kernels. Still removed though.

I cannot simplify the logic much more, I removed the `ask_exit`. I can make own_kernel not a property if you like.

takluyver:
Ctrl-D at an input prompt should raise EOFError.

IPython/terminal/console/interactiveshell.py
```diff
@@ -115,6 +115,14 @@ def _client_changed(self, name, old, new):
         self.session_id = new.session.session
     session_id = Unicode()
 
+    @property
+    def own_kernel(self):
+        return self._own_kernel
+
+    @own_kernel.setter
+    def own_kernel(self, b):
+        self._own_kernel = bool(b)
```
takluyver: This property does seem redundant. The only two places you're setting it, it's a boolean already.

Carreau:
But you are stuck before.I suppose when you `console --existing` you are actually stuck waiting for the kernel to respond, so EOF does nothing, anyway, that's a separate issue.

IPython/terminal/console/interactiveshell.py
takluyver: s/two places/one place/

takluyver:
With the property gone, I think this looks OK now.

Carreau:
property gone

zer0n:
`Ctrl-D` throws 

```
In [7]:
Do you really want to exit ([y]/n)? y
...
  File ".../site-packages/IPython/terminal/console/interactiveshell.py", line 426, in mainloop
    if self.keepkernel and not self.own_kernel:
AttributeError: 'ZMQTerminalInteractiveShell' object has no attribute 'keepkernel'
```

This is minor, as it doesn't kill the kernel. However, graceful exit would be nicer than throwing exceptions.

takluyver:
@Carreau, can you look into that?

Carreau:
Yep will do .

Carreau:
`IPython/terminal/console/interactiveshell.py` does not exist anymore. Are you on IPython master ? and `jupyter_console` master ?

zer0n:
Nope. Actually, I was using IPython 3.1.0. I updated the changed files manually.

Carreau:
Ah, yes this has been backported. sorry, let me retry on 3.x

PR #8493, opened 2015-05-31
safe_execfile_ipy optionally raises exception in user code

tianhuil:
Implementing requested feature in comment

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2664,7 +2664,7 @@ class InteractiveShell(SingletonConfigurable):
                 # tb offset is 2 because we wrap execfile
                 self.showtraceback(tb_offset=2)
 
-    def safe_execfile_ipy(self, fname, raise_exceptions=False, shell_futures=False):
+    def safe_execfile_ipy(self, fname, shell_futures=False, raise_exceptions=False):
         """Like safe_execfile, but for .ipy or .ipynb files with IPython syntax.
 
         Parameters
@@ -2672,13 +2672,13 @@ class InteractiveShell(SingletonConfigurable):
         fname : str
             The name of the file to execute.  The filename must have a
             .ipy or .ipynb extension.
-        raise_exceptions : bool (False)
-            If True raise exceptions everywhere. Meant for testing.
         shell_futures : bool (False)
             If True, the code will share future statements with the interactive
             shell. It will both be affected by previous __future__ imports, and
             any __future__ imports in the code will affect the shell. If False,
             __future__ imports are not shared in either direction.
+        raise_exceptions : bool (False)
+            If True raise exceptions everywhere.  Meant for testing.
         """
         fname = os.path.abspath(os.path.expanduser(fname))
 
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2664,7 +2664,7 @@ class InteractiveShell(SingletonConfigurable):
                 # tb offset is 2 because we wrap execfile
                 self.showtraceback(tb_offset=2)
 
-    def safe_execfile_ipy(self, fname, shell_futures=False):
+    def safe_execfile_ipy(self, fname, raise_exceptions=False, shell_futures=False):
         """Like safe_execfile, but for .ipy or .ipynb files with IPython syntax.
 
         Parameters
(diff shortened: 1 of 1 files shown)


IPython/core/interactiveshell.py
```diff
@@ -2664,7 +2664,7 @@ def safe_execfile(self, fname, *where, **kw):
                 # tb offset is 2 because we wrap execfile
                 self.showtraceback(tb_offset=2)
 
-    def safe_execfile_ipy(self, fname, shell_futures=False):
+    def safe_execfile_ipy(self, fname, raise_exceptions=False, shell_futures=False):
```
Carreau: Inserting this here will break user code that was previously calling `safe_execfile_ipy(name, True)`, not explicitly setting the kwarg.

IPython/core/interactiveshell.py
```diff
@@ -2711,12 +2713,13 @@ def get_cells():
...
                     result = self.run_cell(cell, silent=True, shell_futures=shell_futures)
-                    if not result.success:
+                    if not result.success and raise_exceptions:
+                        if result.error_before_exec is not None:
+                            raise result.error_before_exec
+                        else:
+                            raise result.error_in_exec
+                    elif not result.success:
                         break
             except:
```
takluyver: This outer except will still catch any errors and display the traceback; for writing programmatic tests, the exception should escape the function call so the caller can catch it.

takluyver:
We should also add a simple test of this in `IPython.core.tests.test_interactiveshell`.

IPython/core/interactiveshell.py
```diff
@@ -2711,12 +2713,13 @@ def get_cells():
...
-                    # raised in user code.  It would be nice if there were
-                    # versions of run_cell that did raise, so
-                    # we could catch the errors.
                     result = self.run_cell(cell, silent=True, shell_futures=shell_futures)
-                    if not result.success:
+                    if not result.success and raise_exceptions:
+                        if result.error_before_exec is not None:
+                            raise result.error_before_exec
+                        else:
+                            raise result.error_in_exec
```
takluyver: It might be worth encapsulating these four lines (roughly) as a method on result, so you could just call e.g. `result.reraise_error()`.

IPython/core/interactiveshell.py
tianhuil: Thanks @Carreau.  I flipped the order of the arguments.

IPython/core/interactiveshell.py
tianhuil: That was silly of me.  I've reraised the exception in the except block.

IPython/core/interactiveshell.py
tianhuil: Good idea.  I've added it in.

IPython/core/interactiveshell.py
```diff
@@ -208,6 +208,16 @@ class ExecutionResult(object):
     def success(self):
         return (self.error_before_exec is None) and (self.error_in_exec is None)
 
+    def raise_error(self):
+        """Reraises error if `success` is `True`, otherwise does nothing"""
```
takluyver: Shouldn't that be if `success` is *False*? Success is True when there isn't an error.

IPython/core/interactiveshell.py
```diff
@@ -2711,14 +2723,14 @@ def get_cells():
...
                 for cell in get_cells():
-                    # self.run_cell currently captures all exceptions
-                    # raised in user code.  It would be nice if there were
-                    # versions of run_cell that did raise, so
-                    # we could catch the errors.
                     result = self.run_cell(cell, silent=True, shell_futures=shell_futures)
-                    if not result.success:
+                    if not result.success and raise_exceptions:
+                        result.raise_error()
+                    elif not result.success:
```
takluyver: I think this is now checking the exceptions up to four times. We can simplify it by doing:

```python
if raise_exceptions:
    result.raise_error()
elif not result.success:
    break
```

IPython/core/interactiveshell.py
```diff
@@ -208,6 +208,16 @@ class ExecutionResult(object):
     def success(self):
         return (self.error_before_exec is None) and (self.error_in_exec is None)
 
+    def raise_error(self):
+        """Reraises error if `success` is `True`, otherwise does nothing"""
+        if not self.success:
+            return
```
takluyver: In fact, this check is redundant - since success just means both errors are None, the two `if` lines below are doing the same check.

IPython/core/interactiveshell.py
tianhuil: Yes, sorry.

IPython/core/interactiveshell.py
tianhuil: Fair enough.

tianhuil:
Cheers.  I fixed these.  Let me know if you have any further comments.

takluyver:
Thanks, this looks good to me now.

Issue #8492, opened 2015-05-31
variable substitution in a %%bash cells first line fails silently if one variable isn't defined


drevicko:
Variable substitution in the first line of `%%bash` cells doesn't happen if one of the variables isn't defined. Instead, the name of the variable with a "$" prepended is sent to bash. This is true even for variables in the list that ARE defined.

To demonstrate, say I have a cell like this:

    %%bash -s $a_variable $another_var
    echo $1 $2

If both variables are defined, `str(a_variable)` and `str(another_var)` are output by bash (since variables are substituted by their string representation). If, on the other hand, at least one of the variables isn't defined, bash outputs this:

    $a_variable $another_var

The bad part about it, is that there is no error, exception or anything to warn you that something's gone wrong.

Carreau:
This is a know limitation of the system, There is no easy way to know wether or not you want to pass a litera; `$something` or expand. The only way is to try expanding, and if it fails fallback on the literal. If you have many `$vars` we cannot distinguish which one fails, so the all raw string get send  to bash. And as passing `$literal` without expansion is something that people want to do, we cannot error either. And this applies to all magics. 

There is a plan for magic to have opt-in to automatic variable expansion, but you can only guess what people want to do up to a certain point.

takluyver:
I remembered that we silently fell back to passing literal `$foo`, though I'm a little surprised that as soon as one expression fails, we abandon trying to format any of them. I guess it's fairly rare that you'd actually want a mix of literal $foo and expanded Python expressions in one command, though.

Carreau:
I remember digging through that code for some close but different reason, where it gives some weird but logical behavior in `${something}` vs `{something}`.. anyway. Keep in mind that we actually eval `$...` and if any raises we continue with the raw string. So that things like that are possible:

```
In [1]: class F:
   ...:     @property
   ...:     def bar(self):
   ...:         return 'yeah'
   ...:

In [2]: !echo {F().bar}
yeah
```

We don't dintinguish a NameError from any other excpetions:

```
    def var_expand(self, cmd, depth=0, formatter=DollarFormatter()):
        """Expand python variables in a string.

        The depth argument indicates how many frames above the caller should
        be walked to look for the local namespace where to expand variables.

        The global namespace for expansion is always the user's interactive
        namespace.
        """
        ns = self.user_ns.copy()
        try:
            frame = sys._getframe(depth+1)
        except ValueError:
            # This is thrown if there aren't that many frames on the stack,
[...]

PR #8494, opened 2015-06-01
DollarFormatter supresses `NameError`s, allowing other expressions to be substituted

drevicko:
Relates to issue #8492 

Suppress exceptoins in DollarFormatter, used for cell magic parsing. This allows one expression to raise a NameError (eg: misspelled variable name) while another expression gets substituted. Currently any exceptions prevent any substitutions.

The code that does the `eval` call is in FullEvalFormatter. I thought it best if that class raises all  exceptions by default, so I added a `supressExceptions` parameter, defaulting to `False`, to it's _vformat method. In the DollarFormatter, ~vformat is overridden (simply calling the super version), with this parameter defaulting to True. If suppressExceptions is True, NameErrors during eval() are caught and the evaluated expression is left unchanged in the formatted text.

It only catches NameErrors - perhaps it'd be better to catch all exceptions? 

It may also be a good idea for cell magics to emit a warning if an exception is raised during variable substitution. For that, we'd have to collect the exceptions here and somehow pass them on to the cell magics...

### DIFF
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -561,7 +561,10 @@ class FullEvalFormatter(Formatter):
 
                 # eval the contents of the field for the object
                 # to be formatted
-                obj = eval(field_name, kwargs)
+                try:
+                    obj = eval(field_name, kwargs)
+                except NameError:
+                    obj = field_name
 
                 # do any conversion on the resulting object
                 obj = self.convert_field(obj, conversion)
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -619,7 +619,7 @@ class DollarFormatter(FullEvalFormatter):
             # Re-yield the {foo} style pattern
             yield (txt + literal_txt[continue_from:], field_name, format_spec, conversion)
     def _vformat(self, format_string, args, kwargs, used_args, recursion_depth, supressExceptions=True):
-        return super(FullEvalFormatter,self)._vformat(format_string, args, kwargs, used_args, recursion_depth, supressExceptions)
+        return super(DollarFormatter,self)._vformat(format_string, args, kwargs, used_args, recursion_depth, supressExceptions)
 
 #-----------------------------------------------------------------------------
 # Utils to columnize a list of string
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -619,7 +619,7 @@ class DollarFormatter(FullEvalFormatter):
             # Re-yield the {foo} style pattern
             yield (txt + literal_txt[continue_from:], field_name, format_spec, conversion)
     def _vformat(self, format_string, args, kwargs, used_args, recursion_depth, supressExceptions=True):
-        return super(FullEvalFormatter,self)._vformat(self, format_string, args, kwargs, used_args, recursion_depth, supressExceptions)
+        return super(FullEvalFormatter,self)._vformat(format_string, args, kwargs, used_args, recursion_depth, supressExceptions)
 
 #-----------------------------------------------------------------------------
 # Utils to columnize a list of string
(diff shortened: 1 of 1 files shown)


IPython/utils/text.py
```diff
@@ -561,7 +561,13 @@ def _vformat(self, format_string, args, kwargs, used_args, recursion_depth):
 
                 # eval the contents of the field for the object
                 # to be formatted
-                obj = eval(field_name, kwargs)
+                try:
+                    obj = eval(field_name, kwargs)
+                except NameError:
+                    if supressExceptions:
+                       obj = field_name
```
drevicko: This doesn't work: we need to recover the __original__ string from which field_name was extracted. If '{foo}' was in the `format_string`, `field_name` at this point is 'foo', so we need to put the curly brackets back. Problem is, we don't know where the field_name came from (could have been '$foo' or '{foo}' or.. others?). 

Only sensible way I can see to do this is to change the `Formatter.parse()` interface, make it return the original text as well as `literal_text, field_name, format_spec, conversion` ... something that perhaps should not be taken lightly! How many other things use that method?

I guess we could make a new `parse_with_original()` method (or a less verbose name ;)

drevicko:
There is another travis fail too: using `$undefined_var` is supposed to raise an exception, but that's the whole point of this PR anyway, so if this apporach is decided apon, we should change the test.

IPython/utils/text.py
takluyver: We definitely shouldn't change `Formatter.parse()`, that's part of the stdlib.

I was in favour of this because I thought it was going to be a very simple change. It now looks like it's rather more complicated, and I don't think it's worth the trouble. I don't think users very often want a literal `$foo` on the same line as interpolating a Python variable `$bar`.

Carreau: Well that where the "then a miracle occurs" in reply to "Easy enough" came from. It is a rather nasty thing we have here.

drevicko: I'd wondered where the miracle was going to come in ;)

Is there an easy way to issue a warning, one of those pink messages like when %plab inline colbbers a variable? If the attempt to format the string fails, a user would have a hint about what happened. Currently it's very mysterious and will likely go unnoticed. I guess that's the main reason I raised this issue..

In looking at this, another thing I noticed is a problem with exclamation marks (e.g.: if someone wants '!=' in their expression) - this also silently fails, as ! has special meaning in format strings. There's also the (tiny) chance a someone uses a regex like `a!k:foo`, which doesn't cause Format.parse to raise anything, but returns 'k' as a `conversion` - . Currently this '!k' would disappear! Perhaps this one is "Easy enough to fix" without a miracle (; Perhaps I'll make another PR on that point...

takluyver: I think it's technically easy to issue a warning, but I don't know if we want to, because people use things like `$HOME` without any expectation that it should interpolate a Python variable. I guess you could guess intent because that's environment variables in bash, and they're conventionally capitalised, but it doesn't seem like a very reliable guide.

drevicko: ~~An interesting point is that it's currently impossible to have a bash-ish $HOME and a python-ish $myvar at the same time...~~ No, actually the bash-ish could be written $$HOME and all's fine (the dollar parser recognises $$ and doesn't eval, replacing it with a single $).

Certainly in my case, I'm much more likely to use python variables than bash variables, though I can imagine cases where it may be necessary? That said, the bash/environment variables could always be put in a %%bash cell that calls the script you want.

I guess I'm naturally biased in this, but imho the bizarreness of warrants a warning more than the inconvenience of warning on use of $HOME.

Carreau:
This need rebase/rework. And we need to decide what behavior we want.

Carreau:

PR #8498, opened 2015-06-01
safe_execfile_ipy now takes raise_exceptions as an arguement

tianhuil:


takluyver:
Thanks, but:

- This is a new feature, I don't think it should be backported to a 3.x bugfix release.
- We have a tool to backport PRs automatically, there's no need to resubmit them.

Crosslink #8493

PR #8499, opened 2015-06-01
add tightlist command

minrk:
provided by pandoc 1.14, causing failure to export to pdf


Juanlu001:
Just came here to report this same issue, you are damn fast! :rocket:

Carreau:
We often suspect that Min borrows Guido's time machine now and then.

PR #8504, opened 2015-06-03
notebook: tree rename dialog - focus filename

hiaselhans:
focus only the relevant part of the filename on rename dialog
this selects all prior to the first dot in the filename


Carreau:
Do you want to issue this aginst jupyter/notebook ? We usually only backport bugfixes.

hiaselhans:
sure, sorry
took me some time to find all the relevant repos :)

PR #8506, opened 2015-06-03
added %matplotlib --list so backends are easily visible (finish #8449)

Carreau:
linked to #8444 adding --list argument to show the available backends

---

@takluyver just remove the `default=False,`


Carreau:
Cross link with  #8449

takluyver:
Thanks both

PR #8517, opened 2015-06-08
blacklist jsonschema 2.5.0

Carreau:
ping @takluyver, @jhamrick


rgbkrk:
Looking good, tests pass.

takluyver:
jsonschema 2.5.1 is out now, so tests should be passing anyway.

takluyver:
Closed on the grounds that this is no longer necessary, and if you have 2.5.0 already installed, we are compatible with it.

rgbkrk:
That's true, since installation would fail anyway.

PR #8519, opened 2015-06-09
Add a more informative error message for IPython directive.

chebee7i:
This PR adds a more informative error message for the IPython directive whenever a bad block is detected. The bad block is when no input prompt was provided and only an output prompt.

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -172,7 +172,7 @@ COMMENT, INPUT, OUTPUT =  range(3)
 def block_parser(part, rgxin, rgxout, fmtin, fmtout):
     """
     part is a string of ipython text, comprised of at most one
-    input, one ouput, comments, and blank lines.  The block parser
+    input, one output, comments, and blank lines.  The block parser
     parses the text into a list of::
 
       blocks = [ (TOKEN0, data0), (TOKEN1, data1), ...]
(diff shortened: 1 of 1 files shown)


IPython/sphinxext/ipython_directive.py
```diff
@@ -657,14 +657,41 @@ def process_block(self, block):
...
+                    e = ('\n\nInvalid block: Block contains an output prompt '
+                         'without an input prompt.\n\n'
+                         'Document source: {0}\n\n'
+                         'Content begins at line {1}: \n\n{2}\n\n'
+                         'Problematic block within content: \n\n{TAB}{3}\n\n')
+                    e = e.format(source, linenumber, content, block, TAB=TAB)
+
+                    # Write, rather than include in exception, since Sphinx
+                    # will truncate tracebacks.
+                    sys.stdout.write(e)
```
takluyver: Is it expected that Sphinx extensions will write error descriptions directly to stdout rather than putting it in the exception being raised? And if so, should it be stderr instead of stdout?

takluyver:
Apart from that note, I think this looks OK. I'm not familiar with the Sphinx APIs involved, but only a couple of lines are actually working with that.

IPython/sphinxext/ipython_directive.py
chebee7i: It's a bit of a mixed a bag I think. I copied what was done in other places in the file (I think this was introduced by the pandas folk). The reason it needs to be printed, I think, is that it simply won't be useful to devs if they have to fetch a tmp file to see the sphinx build errors.

Carreau:
> Apart from that note, I think this looks OK. I'm not familiar with the Sphinx APIs involved, but only a couple of lines are actually working with that.

Rest of the file use stdout. 

+1

Issue #14, opened 2010-05-10
Decide when to use Term and when to print


ipython:
Original Launchpad bug 514613: https://bugs.launchpad.net/ipython/+bug/514613
Reported by: ellisonbg (Brian Granger).

Currently in IPython.core we sometimes use IPython.utils.io.Term to print and sometimes we just do print.  The advantage of using Term is that it will properly handle colors on Windows.  But we don't use it everywhere.  Really, we need to come up with a uniform way of handling printing to stdout/stderr in IPython.  This is particularly relevant as we refactor the core to be used without a terminal.

PR #8521, opened 2015-06-10
Remove old unused class

Carreau:
It seem unused in our codebase. The only place where it is used in
github import from genutils which don't exist in IPython.

Closes ipython/ipython#14  

---

Yes it is our oldest open github issue. Next one is 17 :-)

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -38,7 +38,7 @@ from IPython.testing.skipdoctest import skip_doctest
 from IPython.utils import py3compat
 from IPython.utils.py3compat import builtin_mod, iteritems, PY3
 from IPython.utils.contexts import preserve_keys
-from IPython.utils.io import capture_output
+from IPython.utils.capture import capture_output
 from IPython.utils.ipstruct import Struct
 from IPython.utils.module_paths import find_mod
 from IPython.utils.path import get_py_filename, unquote_filename, shellglob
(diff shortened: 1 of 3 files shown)


IPython/core/magics/execution.py
```diff
@@ -38,7 +38,7 @@
 from IPython.utils import py3compat
 from IPython.utils.py3compat import builtin_mod, iteritems, PY3
 from IPython.utils.contexts import preserve_keys
-from IPython.utils.io import capture_output
+from IPython.utils.capture import capture_output
```
rgbkrk: Good cleanup here.

Carreau: Yeah, I actually made the test fail while cleaning io, and rebase/squashed, but don't tell anyone.

rgbkrk:
[True statement above](https://github.com/search?utf8=%E2%9C%93&q=IOTerm+user%3Aipython&type=Code&ref=searchresults)

[After filtering out duds that uses genutils.IOTerm and some repeated old gunk](https://github.com/search?utf8=%E2%9C%93&q=IOTerm+language%3Apython+-filename%3Aishell+-filename%3Aio+-filename%3Aipython_view+NOT+genutils&type=Code&ref=searchresults), there are a few that use this old class but I don't think they'll be updating.

IPython/utils/io.py
```diff
@@ -10,14 +10,9 @@
...
 
 
-import codecs
-from contextlib import contextmanager
-import io
 import os
-import shutil
 import sys
 import tempfile
-import warnings
```
rgbkrk: Just verified these.

PR #8528, opened 2015-06-10
Don't deep reload collections module on Python 3

takluyver:
@Carreau says this didn't work before, but it works on my machine, so I want to try debugging it on Travis


takluyver:
Closing in favour of #8530. I was trying to make deepreload ignore all stdlib modules by default, but I can't find a reliable way of doing that: they don't all have to be in the same location, and which ones go where is not consistent across Python versions.

PR #8535, opened 2015-06-13
reorder imports in IPython.kernel for backward-compat

minrk:
jupyter_client provides a new implementation of find_connection_file. The old (profile-aware) implementation comes from ipykernel now.

cc @SylvainCorlay


minrk:
Hold on, here - this may not be the right thing to do.

minrk:
I think closing this and relying on jupyter/jupyter_client#30 instead is the way to go.

PR #8545, opened 2015-06-18
Remove old deprecated extensions.

Carreau:
All theses extensions are already in their own repository.

Local Rmagic could in any cases not be imported as %load_ext would raise
if rpy2 not available.

Cython and sympy where still trying to do the right things, which I
removed, and converted to warnings. (cannot use DeprecationWarning as
they would be hidden) and no way to get the right stacklevel as this is
magic that load the ext.


IPython/extensions/cythonmagic.py
```diff
@@ -8,36 +10,17 @@
...
 
 ## still load the magic in IPython 3.x, remove completely in future versions.
 def load_ipython_extension(ip):
     """Load the extension in IPython."""
     
-    print("""The Cython magic has been moved to the Cython package, hence """)
-    print("""`%load_ext cythonmagic` is deprecated; please use `%load_ext Cython` instead.""")
+    warnings.warn("""
+The Cython magic has been moved to the Cython package, hence %load_ext cythonmagic` is deprecated; 
+Please use `%load_ext Cython` instead.""")
```
minrk: It's not deprecated if it doesn't work anymore. I think this is fine if you just remove the "hence..." clause.

IPython/extensions/rmagic.py
```diff
@@ -1,702 +1,13 @@
...
-                R_DOC = dedent(RMagics.R.__doc__),
-                RPUSH_DOC = dedent(RMagics.Rpush.__doc__),
-                RPULL_DOC = dedent(RMagics.Rpull.__doc__),
-                RGET_DOC = dedent(RMagics.Rget.__doc__)
-)
-
-
 def load_ipython_extension(ip):
     """Load the extension in IPython."""
     warnings.warn("The rmagic extension in IPython is deprecated in favour of "
```
minrk: s/is deprecated in favour of/has moved to/

Carreau:
Fixed comments

IPython/extensions/sympyprinting.py
```diff
@@ -17,154 +19,14 @@
...
-        latex_formatter.for_type_by_name(
-            'sympy.matrices.matrices', 'Matrix', print_latex
-        )
-
-        for cls in printable_containers:
-            # Use LaTeX only if every element is printable by latex
-            latex_formatter.for_type(cls, print_latex)
-
-        _loaded = True
+    warnings.warn("The sympyprinting extension in IPython is deprecated, "
```
minrk: missed one more no-longer deprecated.

Carreau: Done.

PR #8560, opened 2015-06-21
fix paths in cleanup_ipc_files

minrk:
self.ports is a dict, not a list

closes #8538


minrk:
Looks like I was wrong about ports always being a dict. I'll make sure that's the case in 4.0, but merge #8538 as the fix for 3.2.

PR #8568, opened 2015-06-24
fix to autoreload for cases when module has no __name__ attribute

jewfro-cuban:


Carreau:
+1,

 I think we should test for that, but I like to know (to write the test) what kind of module might not have a `__name__` as `types.ModuleType` require a `name` argument, so I'm not sure how/why this happend, and if this is not (also) hiding a bug in another place. 

When another core-dev things this is fine, please merge, we can investigate the non existing `__name__` later.

takluyver:
This seems approximately reasonable, but I too would like to understand what's going on rather than just overlooking it.

I wonder if something that's not a module could have wound up in `sys.modules`. @jewfro-cuban , can you show us the result of `set(type(m) for m in sys.modules.values())` in a situation where this comes up.

jewfro-cuban:
I get at stack level:
```python
> c:\users\pabramov\appdata\local\continuum\anaconda3\envs\bpy2.74_py3.4.3\lib\site-packages\ipython\extensions\autoreload.py(524)post_execute_hook()
    523         for modname in newly_loaded_modules:
--> 524             _, pymtime = self._reloader.filename_and_mtime(sys.modules[modname])
    525             if pymtime is not None:
```

```python
value of modname:  '__mp_main__'
```

running 
```python
set(type(m) for m in sys.modules.values())
```
returns: 
```python
{<class 'module'>, <class 'six._MovedItems'>}
```

hope that helps.

jewfro-cuban:
Managed to narrow it down.

It happens when I run a script that includes only the following line:

'''python
from IPython.parallel import Client
'''

Using ver 3.2.0

Carreau:
that's `multiprocessing` ! Digging in.

Carreau:
So this is partially why:

In [1]: %load_ext autoreload
In [2]: %autoreload 2
In [3]: run foo.py
# import multiprocessing 
In [4]: import sys
In [5]: sys.modules['__mp_main__'] == sys.modules['__main__']
Out[5]: False
# which is True in not in `__run__`
```

And we "fake" a `__main__` in `%run`, though this "fake"   module have a name, so I'm not sure.

jewfro-cuban:
So anything stopping from merging the PR ?
I've recently updated to version 4 and it was not merged so I get the same behavior.

jewfro-cuban:
btw, if it will help in debugging, now when using ipython parallel I get this message: 

[autoreload of __mp_main__ failed: Traceback (most recent call last):
  File <env>"\lib\site-packages\IPython\extensions\autoreload.py", line 247, in check
    superreload(m, reload, self.old_objects)
AttributeError: 'module' object has no attribute '__name__'
]

Carreau:
Sorry slipped through. 

I think it would be better to "fix"  the `run` and `reload` magic as a module should always have a name and `sys['__main__']` should equal `sys['_mp_main']`, though I have no issues openning an issue to remember that later and merge this.

bnaul:
Any chance of this getting merged soon? I use `%run` and `autoreload` together all the time and this is rather inconvenient.

minrk:
Opened #9069 for further investigation. Merging now. Sorry for the delay @bnaul and @jewfro-cuban!

PR #8574, opened 2015-06-25
Remove outdated info from README

takluyver:
We no longer use submodules, and a development install is now no different from how you'd do an editable install of any Python package, so there's no need for us to be documenting it explicitly.


rgbkrk:
:shipit:

PR #8588, opened 2015-06-29
handle multiple decorators in oinspect

minrk:
We only check for `__wrapped__` once, which won't correctly find a function wrapped in two or more decorators.

This is theoretically vulnerable to fake objects always returning a `__wrapped__` attribute recursivley. If people prefer to be extra safe, we could put a limit on it by using a for-loop:

``` python
for i in range(100):
    if safe_hasattr(obj, '__wrapped__'):
        obj = obj.__wrapped__
    else:
        break
return obj
```

cc @takluyver, Inspection Overlord.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -201,10 +201,7 @@ def getsource(obj, oname=''):
     else:
         # Get source for non-property objects.
 
-        # '__wrapped__' attribute is used by some decorators (e.g. ones defined
-        # functools) to provide access to the decorated function.
-        if hasattr(obj, "__wrapped__"):
-            obj = obj.__wrapped__
+        obj = _get_wrapped(obj)
 
         try:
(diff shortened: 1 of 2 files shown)


takluyver:
I thought we had already done this... ah, no, we talked about it in #578, but [I found](https://github.com/ipython/ipython/pull/578#issuecomment-1574614) that when `functools.wraps()` is used, `__wrapped__` appears to point directly to the end of the chain, so we didn't bother. Evidently there are other things that set `__wrapped__` that don't do that, so this does make sense.

Back then I was concerned that there might be things that always provided a spurious `__wrapped__` attribute, but searching the codebase now, I see at least two places where we dereference it without worrying about that (`IPython.utils.signatures` and `traitlets.utils.getargspec`), so I think this is OK. Hopefully by now objects doing clever things with `getattr` know to exclude special methods from that.

PR #8592, opened 2015-06-29
Make sure outputs are cleared immediately before execution, rather than immediately after

nitind:
Backport change from https://github.com/jupyter/nbconvert/pull/53 .


takluyver:
Thanks

Issue #8489, opened 2015-05-29
Terminal interface using prompt_toolkit?


takluyver:
[prompt_toolkit](https://github.com/jonathanslenders/python-prompt-toolkit) is a pure-Python terminal interface library aimed at creating interactive consoles. [ptpython](https://github.com/jonathanslenders/ptpython/) is a Python console built with it, including `ptipython` which uses IPython internally. Unlike most terminal interface things, it also works on Windows.

We are considering switching our primary terminal interface from readline to prompt_toolkit, because:

- readline is now rather limiting: we can't do syntax highlighting or proper multiline editing; showing tab completions permanently adds them to the terminal scrollback buffer. We can provide a much better user experience with something like prompt_toolkit.
- Depending on readline is a pain beyond Linux: we use pyreadline on Windows, but that's essentially unmaintained and is about to break with Python 3.5. Macs ship with libedit, which is theoretically equivalent but has bugs, so we rely on the third party gnureadline package.

We'll probably need to maintain a simpler fallback mode for situations where prompt_toolkit may not work (@fperez thought ssh-ing to old servers was a likely example).

@jonathanslenders : how much work do you think it would be to convert our existing readline + input() interface to prompt_toolkit? I know you've already made ptipython, and I've started having a look at that, but I'm not clear how much of that code we'd need and how much duplicates things we already do.

jonathanslenders:
Hi @takluyver,

I'm very happy to hear about this.
``ptipython``, while nice, is not the future. (I don't like the current integration with IPython that I did myself.)

What you need for the repl is not much, probably somewhere about 500 or 1000 lines of code to have a nice front-end for IPython that does all the highlighting and autocompletion.

Basically, we need to have one place in IPython where we read the input. That's a blocking call that runs the input interface, and when the user accepts the input, it returns the Python code as a string.

One issue though is that IPython runs its own event loop. (It has several, I believe.) This is mainly important for rendering the mathplotlib graphs. The challenge is to be able to run prompt_toolkit on these event loops. I think it should be possible. Much of the architecture of prompt_toolkit has improved a lot in the last months: now it's easy to run such an interface on any event loop.

I am very willing to help with the integration. The last few weeks I have been working on a few major refactorings in prompt_toolkit. (many performance improvements + much better I/O abstraction.) I will finish this first. It has to be really stable, and then I can help with IPython.

[...]

PR #8597, opened 2015-07-01
[work in progress] Replacing readline with prompt-toolkit.

jonathanslenders:
Pull request that replaces all the readline functionality with prompt-toolkit. This pull request is meant to remove all readline specific stuff, and replace it with prompt-toolkit. That way we have better cross-platform support. (Especially Windows.) It should still provide a very simple command line. (Nice features, like we have in `ptpython` can be added later on.)

To try:
- Clone the latest prompt-toolkit (not the Pypi one): https://github.com/jonathanslenders/python-prompt-toolkit/
- Install this branch.
- Run `ipython` at the terminal.
- %matplotlib, %pylab, plot(), etc... should all work.

See: #8489.

To be done:
- [ ] TerminalInteractiveShell.interact needs to be cleaned up. I don't think we still need things like `hlen_b4_cell`, `rl_do_indent`, etc...
- [ ] The history integration has to be implemented. Right now we have an in-memory history. (It would be nice if we can make the data format compatible with the one that we already had.
- [ ] Remove or fix `_replace_rlhist_multiline`.
- [ ] `prompt_manager.render(color=False)`. We don't want to have `color=True` anymore. The colors are determined by the Pygments color scheme. We don't want prompt_manager to return VT100 terminal escape codes.
- [ ] Remove the parts of the `InputSplitter` that we don't need anymore.
- [ ] Remove `utils/coloransi.py`. Use Pygments color schemes and prompt-toolkit instead.
- [ ] Fix core/debugger. It used PyColorize. We should replace that with prompt-toolkit highlighting. (We can also highlight everything on Windows.)
- [ ] Unit tests.
- [ ] core/ultratb should become platform independent as well. (Not using coloransi or PyColorize.)
- [ ] core/oinspect.
- [ ] check every reference for pycolorize.
- [ ] Create Pygments color schemes.
- [ ] every place where we expect to write colored output. Use prompt_toolkit.Interface.print_tokens (https://github.com/jonathanslenders/python-prompt-toolkit/blob/f7bfe0ad1fb3cf48d205f517e8963878528fd31d/prompt_toolkit/interface.py#L440)

Further, this needs to be tested, improved, reviewed, etc...


minrk:
Great to see this. Feel free to ping about any specific questions you have. We'll be pretty focuses on shipping 4.0 for the next little while (and SciPy next week), but I'm excited about this.

> Remove utils/coloransi.py

You have no idea how happy that makes me.

jonathanslenders:
Yes, I'll ping you. Probably I'll have many questions when I continue. Saturday I leave for three weeks on holiday, so not sure how much time I'll have. But I hope we should be able to finish it this summer.

takluyver:
I wonder if it might be better to add the prompt_toolkit machinery as a new interface initially, rather than changing TerminalInteractiveShell itself, and then phase out the readline-based TerminalInteractiveShell.

There's some extra complexity that's not visible in this repo, because the terminal frontend for Jupyter kernels (https://github.com/jupyter/jupyter_console ) currently inherits from TerminalInteractiveShell. It probably shouldn't do that, but it was the easiest way to get it up and going. I hope we can get that using prompt_toolkit too, but we should probably break the direct dependency between the two.

takluyver:
Also, thinking a bit more about the event loop integration stuff:

We currently have [two kinds of GUI event loop integration](http://ipython.org/ipython-doc/3/config/eventloops.html). In the kernel, we hand over to the GUI event loop as the main loop, and trust it to call our own event loop at least every `poll_interval` seconds (default 0.05). In the terminal, Python's own event loop is the main one, and it calls the GUI loop in small bursts, via inputhook. But with readline we don't have a choice in this matter. If we could give primary control to the GUI event loop in the terminal, we could consolidate our two kinds of integration into one, which would be really nice.

However, the kernel approach ('call me') may not be quick enough for it our event loop to process every keystroke (in the kernel, it only needs to give us control when there's a message to handle). It's worth investigation, though.

IPython/core/prompts.py
```diff
@@ -350,64 +294,39 @@ def update_prompt(self, name, new_template=None):
             self.templates[name] = multiple_replace(prompt_abbreviations, new_template)
         # We count invisible characters (colour escapes) on the last line of the
         # prompt, to calculate the width for lining up subsequent prompts.
-        invis_chars = _lenlastline(self._render(name, color=True)) - \
-                        _lenlastline(self._render(name, color=False))
+        invis_chars = _lenlastline(self._render(name)) - \
+                        _lenlastline(self._render(name))
```
asmeurer: This looks wrong.

jonathanslenders: Yes, there is actually much more wrong. I hope to find soon some time to continue on this. It is more work than I expected. (Everything related to the terminal and colors is spread all over IPython.)

asmeurer: If you do work on this, you should be aware of https://github.com/ipython/ipython/issues/8724.

minrk:
closing in favor of #9118

PR #8599, opened 2015-07-01
Add usage examples for pre/post-save hooks in docs.

irhs:
Add pre/post-save hooks usage examples from PR description[1](https://github.com/ipython/ipython/pull/6896#issue-48344492) to iPython config doc.

### DIFF
--- a/docs/source/config/intro.rst
+++ b/docs/source/config/intro.rst
@@ -46,6 +46,62 @@ and sets)::
 .. versionadded:: 2.0
    list, dict and set methods for config values
 
+To add a pre/post-save hooks (version 3.0 or later) use::
+
+    c.FileContentsManager.pre_save_hook = pre_save_method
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! The notebook has actually been split out of this repo, though, and it now lives at https://github.com/jupyter/notebook

@jdfreder is co-ordinating the docs for this repo. Jon, I think it's useful to have some examples of save hooks. Can you work out the best place to put them?

jdfreder:
Hmm, doesn't look like there's an exact place for this quite yet, but I think we can make one.  A new section called "configuration" right above https://github.com/jupyter/notebook/blob/master/docs/source/notebook.rst#creating-a-new-notebook-document  could work.  Probably even better would be an entirely new document called configuration.

I think I know enough git foo to do this while retaining @irhs 's commit credit.

takluyver:
I think this could now be added to the `extending/` subfolder of the docs.

PR #8601, opened 2015-07-02
Deprecate and undocument install_ext

Carreau:
It seem to me that Python packaging has come up to a point where
actually telling peoples they can install python extensions not using
package manager is unreasonable.

Recent extension even recommend using `install_ext` as **the way** to
install things, which is unsecure, confusing, and there is no update
mechanisme, nor version conflict resolution.

---

As I think this might be historical I would like @fperez  input.

### DIFF
--- a/IPython/core/magics/extension.py
+++ b/IPython/core/magics/extension.py
@@ -19,6 +19,7 @@ import os
 # Our own packages
 from IPython.core.error import UsageError
 from IPython.core.magic import Magics, magics_class, line_magic
+from  warnings import warn
 
 #-----------------------------------------------------------------------------
 # Magic implementation classes
@@ -42,6 +43,8 @@ class ExtensionMagics(Magics):
           -n filename : Specify a name for the file, rather than taking it from
                         the URL.
         """
+        warn("%install_ext` is deprecated, please distribute your extension(s)"
+             "as a python packages.", UserWarning)
         opts, args = self.parse_options(parameter_s, 'n:')
         try:
             filename = self.shell.extension_manager.install_extension(args,
--- a/docs/source/config/extensions/index.rst
+++ b/docs/source/config/extensions/index.rst
@@ -19,9 +19,6 @@ the `Framework :: IPython tag <https://pypi.python.org/pypi?:action=browse&c=586
 on PyPI.
 
 Extensions on PyPI can be installed using ``pip``, like any other Python package.
-Other simple extensions can be installed with the ``%install_ext`` magic. The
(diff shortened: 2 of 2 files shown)


Carreau:
Any feedback?

minrk:
I'm fine with this. I'd also be fine with de-emphasizing it without the deprecation message. I'd like to hear from @ellisonbg or @fperez.

@Carreau perhaps pinging the list for comment from people who might be using `%load_ext`?

fperez:
Why don't we try to get some input on the list? We might be surprised by what we hear, who knows...

fperez:
Though in principle I agree, the less we invent our own homebrewed, one-of-a-kind installation tools, the better...

Carreau:
> @Carreau perhaps pinging the list for comment from people who might be using %load_ext?

I think you mean `%install_ext` . I'll ping the list.

fperez:
There seems to be good consensus on the list on this, so +1 on moving forward. Thanks for taking the time to communicate it more broadly, always good to get the bigger picture, sometimes we get surprised...

minrk:
@fperez sounds good. Want to do the honors of hitting the green button?

moorepants:
Just wanted to point out this extension:

http://nbviewer.ipython.org/github/jrjohansson/version_information/blob/master/example.ipynb

It is not a package and I have this in probably 50+ notebooks now, all using %install_ext. The repo only has 16 stars so maybe it isn't in wide use though.

Carreau:
@moorepants that exactly the kind of things that we want to avoid. It would take less than 5 minutes to make that a package, and wouldn't change the way to load the extension.

moorepants:
That's fine if you want to avoid it. I just mentioned it because your change will break all of my notebooks. This is just an example of a potential wider use case of `%install_ext` for you all to be aware of.

Carreau:
I'm not sure in what this would break all your notebooks. Install ext should in anyway not be use at the start of all notebooks ? no? Once it's installed. It's installed, the `%load_ext` will still work right ?

moorepants:
Oh, I didn't know that. I just copied the example notebook I shared, so `%install_ext` is in every single notebook. Not sure what other people did. I've never looked at what `%install_ext` actually does.

Carreau:
Well actually install_ext is just downloading a file and putting it on PYTHONPATH at first order. IF version_information.py is a package which is pip-installable, then you can replace `%install_ext <url>` by `pip install <pacakgename>` and the rest works the same.

moorepants:
Ok, I'll submit a PR to that package to make it pip installable. Thanks for explaining.

PR #8602, opened 2015-07-02
add ShimWarning for shimmed imports

minrk:
makes it easier to catch/suppress warnings about shim imports

still shown by default (not a DeprecationWarning)


takluyver:
:+1:

Issue #8543, opened 2015-06-18
Unicode decode error when pressing Tab+Shift for tooltip in ipython notebook 3.1


mmagnuski:
Hey,
I am getting unicode decode error when asking for tooltip (`Tab` + `Shift`).
I am on windows 8 and yes - I have non-ascii characters in my username AND python 2.7.
I understand that you prefer to answer to this kind of reports by urging users to move to python 3, but in my case the tooltip works when I first open the notebook, only then, some time later, stops working.

The error shows in console, not in the notebook, so although I've had the problem for some time only recently have I noticed the error. It says:
```
[IPKernelApp] ERROR | Exception in message handler:
Traceback (most recent call last):
  File "D:\Python\Anaconda\lib\site-packages\IPython\kernel\zmq\kernelbase.py", line 213, in dispatch_shell
    handler(stream, idents, msg)
  File "D:\Python\Anaconda\lib\site-packages\IPython\kernel\zmq\kernelbase.py", line 420, in inspect_request
    content.get('detail_level', 0))
  File "D:\Python\Anaconda\lib\site-packages\IPython\kernel\zmq\ipkernel.py", line 253, in do_inspect
    info = self.shell.object_inspect(name)
  File "D:\Python\Anaconda\lib\site-packages\IPython\core\interactiveshell.py", line 1585, in object_inspect
    detail_level=detail_level
  File "D:\Python\Anaconda\lib\site-packages\IPython\core\oinspect.py", line 749, in info
    out['file'] = compress_user(fname)
  File "D:\Python\Anaconda\lib\site-packages\IPython\utils\path.py", line 88, in compress_user
    if path.startswith(home):
UnicodeDecodeError: 'ascii' codec can't decode byte 0x9c in position 13: ordinal not in range(128)
```

takluyver:
Sorry, but I am going to urge you to move to Python 3 ;-). If it's an intermittent failure, that makes it all the more difficult to debug.

mmagnuski:
Well, ok, probably it's high time I moved to python 3. I've had enough Unicode Adventures. :smile:

takluyver:
Cool, good luck!

PR #8614, opened 2015-07-08
Safely encode paths in IPython.utils.path.compress_user

smathot:
This is a minor bugfix that resolves an issue on Python 2 on Windows systems with non-ascii home folders. On these systems, when IPython tries to show a tooltip for a function from a non-ascii location, a `UnicodeDecodeError` is triggered. This appears to be related to issue #8543. I think this is a safe bugfix, as the old behavior is clearly incorrect on Python 2, and the change does nothing on Python 3. I have given it a few test runs.

The relevant line is 85. I apologize for the other (unnecessary) line changes; Atom stripped the trailing tabs automatically.

### DIFF
--- a/IPython/utils/path.py
+++ b/IPython/utils/path.py
@@ -81,7 +81,8 @@ def unquote_filename(name, win32=(sys.platform=='win32')):
 
 def compress_user(path):
     """Reverse of :func:`os.path.expanduser`
-    """
+    """    
+    path = py3compat.unicode_to_str(path, sys.getfilesystemencoding())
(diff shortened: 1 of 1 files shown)


takluyver:
This looks reasonable.

For future reference, when working on IPython, please turn off the 'strip trailing whitespace' setting in your editor, because it makes the diffs unnecessarily big. This one isn't too bad, though, so you don't need to change it.

minrk:
You can disable the whitespace stripping atom with Settings > Packages > whitespace > disable.

PR #8622, opened 2015-07-10
reshow nbagg figures on display

### LINKED ISSUE #8617: use js version of nbagg figures for display
The display system shows a png of figure objects, even if they are live nbagg figures and can have a live display.

This may require changes on the mpl side as well.

attn @pelson

minrk:
instead of inline images

closes #8617

This isn't quite done, there appears to be something to do about double-display.

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -173,6 +173,16 @@ def mpl_runner(safe_execfile):
     return mpl_execfile
 
 
+def _reshow_nbagg_figure(fig):
+    """reshow an nbagg figure"""
+    try:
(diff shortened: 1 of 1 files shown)


takluyver:
@minrk: tagged as waiting for author because you said there's still something to be done

takluyver:
@minrk This is still tagged waiting-for-author because you said there's something left to do. Can you remember what that was? Does this make sense to do without whatever-it-was?

minrk:
It was a double-display, I think. I'll try to look at it tomorrow.

minrk:
This appears to be working just fine, now.

PR #8650, opened 2015-07-23
Fix config metadata to use .tag()

jasongrout:
This depends on https://github.com/ipython/traitlets/pull/53.


samahangilocano27:
I need to fix my galaxys5

Carreau:
Need rebase. The question is will we force a more recent version of traitlets for IPython 5.0 ?

minrk:
I think it makes sense to do this for 5.0.

PR #8656, opened 2015-07-24
added absolute_import to utils.text module + test

Mrngilles:
Fix the `utils.text` module to use absolute import in python 2.7. Also added the test that the `path` module is properly imported.
Seems also deleted a lot of whitespace trailing.

### DIFF
--- a/IPython/utils/tests/test_text.py
+++ b/IPython/utils/tests/test_text.py
@@ -180,7 +180,7 @@ def test_LSString():
     lss = text.LSString("abc\ndef")
     nt.assert_equal(lss.l, ['abc', 'def'])
     nt.assert_equal(lss.s, 'abc def')
-    nt.assert_in("path", sys.modules)
+    nt.assert_isinstance(lss.p[0], path.path)
 
 def test_SList():
     sl = text.SList(['a 11', 'b 1', 'a 2'])
--- a/IPython/utils/tests/test_text.py
+++ b/IPython/utils/tests/test_text.py
@@ -188,3 +188,6 @@ def test_SList():
     nt.assert_equal(sl.grep(lambda x: x.startswith('a')), text.SList(['a 11', 'a 2']))
     nt.assert_equal(sl.fields(0), text.SList(['a', 'b', 'a']))
     nt.assert_equal(sl.sort(field=1, nums=True), text.SList(['b 1', 'a 2', 'a 11']))
+
+def test_non_local_path_import():
+    nt.assert_in("path", sys.modules)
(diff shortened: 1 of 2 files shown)


IPython/utils/tests/test_text.py
```diff
@@ -80,13 +80,13 @@ def eval_formatter_check(f):
     nt.assert_equal(s, ns['stuff'])
     s = f.format("{stuff!r}", **ns)
     nt.assert_equal(s, repr(ns['stuff']))
-    
+
```
minrk: Looks like you have your editor configured to mess about with whitespace. We would prefer that you avoid committing large whitespace changes.

If you are using atom, this is its extremely unfortunate default behavior, and can be disabled with: Preferences > Packages > whitespace > disable

Mrngilles:
Yes, I usually, trim all whitespace. I will change it right now

takluyver:
Thanks!

About the test: the import is happening inside the method `LSString.get_paths()`, and indeed when I tested this on Python 2, that method failed. There's already a `test_LSString()` function there - if that called `.get_paths()`, we would have noticed this problem earlier.

Mrngilles:
So is this okay? If something else needs tinkering, let me know.

takluyver:
Can you add to `test_LSString` to ensure that `.get_paths()` works, instead of adding a new test?

Mrngilles:
Should be good now

IPython/utils/tests/test_text.py
```diff
@@ -180,6 +180,7 @@ def test_LSString():
     lss = text.LSString("abc\ndef")
     nt.assert_equal(lss.l, ['abc', 'def'])
     nt.assert_equal(lss.s, 'abc def')
+    nt.assert_in("path", sys.modules)
```
takluyver: This should check that `lss.get_paths()` works; there's no real significance to checking that 'path' has been imported here.

Mrngilles: Okay, I think I understand better. But to be sure to have some value back from `lss.get_paths()` I will need to change lss as neither `abc` or `def` are correct path values. I'm thinking about something like : 
```python
lss = text.LSSstring("\n".join(os.getcwd(), os.getcwd())
nt.assert_equal(lss.p, [os.getcwd(), os.getcwd()])
```
Using `os.getcwd()` ensures we have a proper path in any case.

takluyver: Actually, it should be OK with `abc` and `def` - they're valid filenames, and there's no requirement that a path refers to a file that actually exists.

The main purpose of the exercise is a *smoketest*: run some code and check that it doesn't error. But we know that the return values should be `path.path` objects, so you could check like this:

```python
import path
...
def test_LSString():
    ...
    nt.assert_isinstance(lss.p[0], path.path)
```

Mrngilles:
Actually, the file has to exist https://github.com/ipython/ipython/blob/master/IPython/utils/text.py#L70
I am going just going to change the lss to the cwd before making the last test.

takluyver:
Thanks!

PR #8658, opened 2015-07-24
Test on Python 3.5 beta 3

takluyver:


takluyver:
OK, there's some exotic looking failures in deepreload. I'll see if I can debug that.

takluyver:
OK, with a few more exclusions in the deepreload test, everything is passing.

PR #8673, opened 2015-07-29
Support unicode returned by raw_input

### LINKED ISSUE #8670: Support `raw_input` returning unicode
Hello, I'm trying to port my package https://github.com/Drekin/win-unicode-console to Python 2 and I'd like to be compatible with IPython. My code monkeypatches builtin function `raw_input`, which afterwards returns a unicode object rather than str. Unfortunatelly, IPython is not compatible with this under Python 2: `IPython.terminal.interactiveshell.TerminalInteractiveShell.raw_input` calls `IPython.utils.py3compat.str_to_unicode` (which is mapped to `.decode`) on the result of original `raw_input`. But when `raw_input` returns a unicode object, trying to decode it leads to encoding it first, and if it contains a non-ascii character, the extra encoding step leads to `UnicodeEncodeError` since ascii encoding is assumed. The exception leads to `"You or a %run:ed script called sys.stdin.close() or sys.stdout.close()!\nExiting IPython!\n"` warning and IPython REPL exit. It would be nice if a unicode object was left alone rather then tried to be decoded, i.e. if for example `IPython.utils.py3compat.cast_unicode` was used instead of `.str_to_unicode`.

Drekin:
When a unicode string is returned by raw_input, it is now left alone rather than encoded to ascii when tried to be decoded, which lead to UnicodeEncodeError and IPython exit with inaccurate error message. Fixes #8670.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -525,7 +525,7 @@ class TerminalInteractiveShell(InteractiveShell):
         prompt = py3compat.cast_bytes_py2(prompt)
 
         try:
-            line = py3compat.str_to_unicode(self.raw_input_original(prompt))
+            line = py3compat.cast_unicode_py2(self.raw_input_original(prompt))
         except ValueError:
             warn("\n********\nYou or a %run:ed script called sys.stdin.close()"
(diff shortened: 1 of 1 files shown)


minrk:
@Drekin thanks!

PR #8686, opened 2015-08-04
Sphinxext: display shorter error traceback when :okexcept: is specified

jorisvandenbossche:
This is a small patch to push a pandas change upstream (https://github.com/pydata/pandas/pull/10727).

When using `:okexcept:` for code blocks that are not suppressed (e.g. you want to show that a certain line of code raises an error), at this moment the full error traceback is shown. While for documentation/education purposes, it is enough/better to only show the error message itself.

E.g instead of the full traceback you see in the docs here: http://pandas.pydata.org/pandas-docs/version/0.16.2/timeseries.html#ambiguous-times-when-localizing, you would get:

```
In [292]: rng_hourly.tz_localize('US/Eastern')
AmbiguousTimeError: Cannot infer dst time from Timestamp('2011-11-06 01:00:00'), try using the 'ambiguous' argument
```

Open questions:
- I now just changed the behaviour, but this could also be controlled by a new option or a new pseudo-decorator if needed.
- I now only show the error message itself, but it is also possible to e.g. show the starting line (`------...`) of the traceback as well.

cc @chebee7i @tacaswell (as other users of / contributors to the ipython directive)

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -461,6 +461,43 @@ class EmbeddedSphinxShell(object):
         # Fetch the processed output. (This is not the submitted output.)
         self.cout.seek(0)
         processed_output = self.cout.read()
+
+        # context information
+        filename = "Unknown"
+        lineno = 0
+        if self.directive.state:
+            filename = self.directive.state.document.current_source
+            lineno = self.directive.state.document.current_line
+
+        # output any exceptions raised during execution to stdout
+        # unless :okexcept: has been specified.
+        if not is_okexcept and "Traceback" in processed_output:
+            s =  "\nException in %s at block ending on line %s\n" % (filename, lineno)
+            s += "Specify :okexcept: as an option in the ipython:: block to suppress this message\n"
+            sys.stdout.write('\n\n>>>' + ('-' * 73))
+            sys.stdout.write(s)
+            sys.stdout.write(processed_output)
+            sys.stdout.write('<<<' + ('-' * 73) + '\n\n')
+
+        # output any warning raised during execution to stdout
+        # unless :okwarning: has been specified.
+        if not is_okwarning:
+            for w in ws:
+                s =  "\nWarning in %s at block ending on line %s\n" % (filename, lineno)
+                s += "Specify :okwarning: as an option in the ipython:: block to suppress this message\n"
+                sys.stdout.write('\n\n>>>' + ('-' * 73))
+                sys.stdout.write(s)
+                sys.stdout.write(('-' * 76) + '\n')
+                s=warnings.formatwarning(w.message, w.category,
+                                         w.filename, w.lineno, w.line)
+                sys.stdout.write(s)
+                sys.stdout.write('<<<' + ('-' * 73) + '\n')
+
+        # if :okexcept: has been specified, display shorter traceback
+        if is_okexcept and "Traceback" in processed_output:
+            traceback = processed_output.split('\n')
+            processed_output= '\n'.join(traceback[-2:])
+
         if not is_suppress and not is_semicolon:
             #
             # In IPythonDirective.run, the elements of `ret` are eventually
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
cc @tacaswell

takluyver:
Not entirely sure about this - wouldn't you want the output to match what a user would see if they actually run that?

On the other hand, since we don't use the IPython Sphinx directive ourselves, I'm happy for those who do to take the lead.

jorisvandenbossche:
I understand, and that's why I certainly can make it optional if needed.

For our docs, I think that the full traceback only adds clutter to the documentation, that can overshadow what you actually want to show (error tracebacks can get much longer than the one in the example I gave). Furthermore, the full traceback is most of the time not that interesting (apart from where it was initially raised, but for the docs this is something like `<ipython-input-292-8c5fa6a37f5b> in <module>()
----> 1 rng_hourly.tz_localize('US/Eastern')`) and contains some info that is specific to the user anyway (the paths of the files).

takluyver:
That makes sense. I think I'd lean towards making an option for it, but it's probably best to ask the users: would anyone enable 'show the full traceback' if it was an option?

tacaswell:
I would like a truncated traceback that shows the inner most frame or two.  Gives a bit more context to match what the users see when something blows up.

jorisvandenbossche:
@tacaswell What do you mean exactly with the inner most?

Using the example from the docs I linked, I would expect that you mean:

```
In [292]: rng_hourly.tz_localize('US/Eastern')

/home/joris/scipy/pandas/pandas/tseries/index.pyc in tz_localize(self, tz, ambiguous)
   1619 
   1620             new_dates = tslib.tz_localize_to_utc(self.asi8, tz,
-> 1621                                                  ambiguous=ambiguous)
   1622         new_dates = new_dates.view(_NS_DTYPE)
   1623         return self._shallow_copy(new_dates, tz=tz)

/home/joris/scipy/pandas/pandas/tslib.so in pandas.tslib.tz_localize_to_utc (pandas/tslib.c:47148)()

AmbiguousTimeError: Cannot infer dst time from Timestamp('2011-11-06 01:00:00'), try using the 'ambiguous' argument
```

while I would think that for a user the outer most frame (where it is raised in the user his code) is the most useful to see:
```
In [292]: rng_hourly.tz_localize('US/Eastern')
AmbiguousTimeError                        Traceback (most recent call last)
<ipython-input-292-8c5fa6a37f5b> in <module>()
----> 1 rng_hourly.tz_localize('US/Eastern')

AmbiguousTimeError: Cannot infer dst time from Timestamp('2011-11-06 01:00:00'), try using the 'ambiguous' argument
```

But of course, for the docs it is maybe not really about what is useful for the user in real world, but that it matches somewhat the actual output.

To meet the different needs, I could eg make a config option `show_error_traceback_frames` which is by default None (show all), but you can set this to a number (0 -> show no frames apart from the actual message, 1 -> show the 1 inner most, ...). That should be rather easy (the frames are delineated by blank lines)

tacaswell:
My thought was that if the user is sitting in the repl and then get a deep trackback they will only be able to see the inner frame or two so showing them that will help with scanning/pattern matching.

IPython/sphinxext/ipython_directive.py
```diff
@@ -461,6 +461,43 @@ def process_input(self, data, input_prompt, lineno):
...
+                sys.stdout.write(('-' * 76) + '\n')
+                s=warnings.formatwarning(w.message, w.category,
+                                         w.filename, w.lineno, w.line)
+                sys.stdout.write(s)
+                sys.stdout.write('<<<' + ('-' * 73) + '\n')
+
+        # if :okexcept: has been specified, display shorter traceback
+        if is_okexcept and "Traceback" in processed_output:
+            traceback = processed_output.split('\n')
+            processed_output= '\n'.join(traceback[-2:])
```
takluyver: `processed_output` is all the output corresponding to that chunk of code, right? So if anything is printed before the traceback, this discards it?

Carreau:
What the status of this ?

jorisvandenbossche:
The status is that I have to look at this again, but do not have the time right now (trying to finish a phd and such things ...). 
I still think this is potentially useful, but the current implementation should be better tested (the change I did in the pandas version of the ipython directive was reverted because it gave some errors on travis for some reason I was not yet able to understand, locally it runs fine)

jorisvandenbossche:
So you can close it with a closed-pr label for now, unless someone wants to take this up in the short term

Carreau:
No problem. Get your PhD. I'll see if we can find someone to revive this !

Carreau:

-- 


@jorisvandenbossche : See you after your PhD ! Hope things are moving forward !

PR #8706, opened 2015-08-09
Save worst time as well

native-api:
If caching is involved, it's the only number pointing to the fact of how long the operation actually takes.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -69,10 +69,11 @@ class TimeitResult(object):
 
     """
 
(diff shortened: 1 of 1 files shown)


takluyver:
I don't see a problem with that, but if caching is involved, even the worst number is probably going to be thrown way off by the looping that tries to get a more accurate timing.

native-api:
No, if there's caching, the initial attempt will likely be the worst.
If you see the logic @ execution.py:1032-1047,
it guarantees that `worst_tuning' (and then `worst')
will the the time of this attempt.

takluyver:
Ah, got it. Thanks.

PR #8714, opened 2015-08-11
Run autogen scripts on readthedocs

minrk:
in conf.py

includes abspath fixes to remove CWD assumptions from the scripts.


Carreau:
+1

PR #8720, opened 2015-08-12
[WIP] add metadata to completions

willwhitney:
This PR is not ready to be merged, and is intended for discussion and review.

These changes allow for UI like this:

screenshot 2015-08-12 12 22 27

It pulls the names of parameters to a procedure out using `inspect`:

screenshot 2015-08-12 12 19 16

screenshot 2015-08-12 12 18 44

Live values of variables are included as well:

screenshot 2015-08-12 12 19 03

More information about these changes is available in this thread: https://github.com/jupyter/jupyter_client/issues/51

Presently this replaces the `matches` key in `complete_reply`. This is temporary and was only to save time in first getting this visible. We don't yet have a firm decision on where to put this information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ __pycache__
 \#*#
 .#*
 .coverage
+.idea
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -170,7 +170,7 @@ def compress_user(path, tilde_expand, tilde_val):
 
 
 
-def penalize_magics_key(word):
+def penalize_magics_key(match):
     """key for sorting that penalizes magic commands in the ordering
 
     Normal words are left alone.
@@ -195,6 +195,7 @@ def penalize_magics_key(word):
 
     # Move any % signs from start to end of the key 
     # provided there are no others elsewhere in the string
+    word = match['text']
 
     if word[:2] == "%%":
         if not "%" in word[2:]:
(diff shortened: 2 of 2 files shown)


minrk:
Very nice! I suspect we will need to add several of the safety wrappers to some of the inspections that we have accumulated in oinspect for misbehaving objects. In fact, does it make sense to *use* oinspect to get the fields, so we aren't duplicating the inspection machinery?

willwhitney:
Very probably. `completer.py` already used inspect in several places and not oinspect, so I didn't even know that existed.

minrk:
It may not be relevant for the particular calls here, I'm just not sure. Worth peeking at oinspect to see if there any any similar inspection calls wrapped in `_safe_this_or_that`. Since oinspect does the same building of a dict of info about an object, it may be worth reusing that, still.

ssanderson:
@minrk @willwhitney what's the status of this and https://github.com/jupyter/jupyter_client/issues/51 ? I saw in https://github.com/jupyter/notebook/issues/99 Will mentioned he doesn't really have the bandwidth to look at this.  I'd be interested in potentially picking this up depending on how much work we think is still needed.

willwhitney:
@ssanderson this shouldn't be too big a deal to wrap up. Since we're going to go with an Atom-style API the fields need to be renamed/removed and generally shuffled around, but basically all of the code here can be reused.

As a user it would be **amazing** if you picked up the baton.

minrk:
@ssanderson if you wanted to pick it up and @willwhitney doesn't mind, that would be great. I think most of the discussion of the next round of changes is in jupyter/jupyter_client#51.

Carreau:
This al lost momentum, thanks to all for working on this, and to @willwhitney  for starting it  ! I hope we can find more time to work again with you at a later time. 

For now this is cluttering the list of opened PR and push things to the second page, so I'm going to tag as "Closed PR" (that we use to remind us that some work have been done, and need to be updated). And close just to get a clear view of actively worked on PRs. 

Feel free to reopen, or ask to be reopen if needed.

PR #8722, opened 2015-08-12
Remove redundant qtconsole doc page (was moved to jupyter/qtconsole repo)

lebedov:
See [this thread](https://groups.google.com/forum/#!topic/jupyter/nHfgUmoZyBM) for the context.


Carreau:
Wait ... Why do we still have qtconsole docs in IPython dir ? the docs should have been moved [here](https://github.com/jupyter/qtconsole) ?

o_O.

Regardless of that +1 on the change. I just want to check before that we don't have a duplicate file in 2 repos.

takluyver:
Oh yeah, I think that's an oversight. We should probably get rid of the Qt console docs in the IPython repo.

lebedov:
The qtconfig doc pages in the ipython/ipython and jupyter/qtconfig repos do seem mostly duplicated (although the latter does contain a brief mention of the `--colors` option that should be expunged and doesn't mention the `%matplotlib inline` magic).

lebedov:
Should I update the PR to remove the qtconsole page from the ipython/ipython docs altogether?

takluyver:
If it's mainly the same content, yes, please do make a PR to remove this copy. Thanks!

lebedov:
PR updated.

takluyver:
Thanks

PR #8729, opened 2015-08-13
IPython.load_extensions is using a relative path

aiguofer:
I was unable to load some extensions and was able to trace it to this problem:

```
    var load_extensions = function () {
        // load one or more IPython notebook extensions with requirejs
        var extensions = [];
        var extension_names = arguments;
        for (var i = 0; i < extension_names.length; i++) {
            extensions.push("nbextensions/" + arguments[i]);
        }
    ...
```

This creates a relative path which means that whenever you run the function within a notebook, it tries to fetch from <notebook_url>/nbextensions/<extension_path> and, thus, can't find it.

I'm not very familiar with the extensions codebase, but it seems like it should be:
extensions.push("/nbextensions/" + arguments[i]); instead. I'm submitting the PR but feel free to deny if this is correct and I'm just doing something wrong.

### DIFF
--- a/IPython/html/static/base/js/utils.js
+++ b/IPython/html/static/base/js/utils.js
@@ -17,7 +17,7 @@ define([
         var extensions = [];
         var extension_names = arguments;
         for (var i = 0; i < extension_names.length; i++) {
-            extensions.push("nbextensions/" + arguments[i]);
+            extensions.push("/nbextensions/" + arguments[i]);
         }
         
         require(extensions,


takluyver:
That's a requirejs path - it should be transformed to a real URL according to the config here:

https://github.com/jupyter/notebook/blob/master/notebook/templates/page.html#L20

minrk:
An absolute path is definitely incorrect here, because it won't take server prefixes into account. The relative path gets transformed to an absolute path using url prefixes via require.js. config. It's possible that's misbehaving somehow.

aiguofer:
ahhh that makes sense.. so the reason it wasn't working was because I was adding '.js' when trying to load an extension.

Working as expected now, thanks!

PR #8736, opened 2015-08-16
remove mathjax utils

### LINKED ISSUE #8735: Old version of MathJax imported with install_mathjax()
If you run

    from IPython.external.mathjax import install_mathjax
    install_mathjax()

then MathJax 2.4 is downloaded and installed. However, the current version of MathJax is 2.5.

Line 138 of mathjax.py states:

    def install_mathjax(tag='2.4.0', dest=default_dest, replace=False, file=None, extractor=extract_tar):

I suggest this should be updated to 2.5.0

Carreau:
Mathjax is now part of notebook.

closes #8735


IPython/external/mathjax.py
```diff
@@ -1,237 +1,13 @@
...
-        return install_mathjax(file=open(fname, "rb"), replace=replace, extractor=extractor, dest=dest)
-    else:
-        return install_mathjax(replace=replace, dest=dest)
+"""
 
+import sys 
 
 if __name__ == '__main__' :
-    sys.exit(main())
+    sys.exit("IPython.external.mathjax is deprecated")
```
minrk: Do you want to explain that mathjax is already installed, not just that "what you want is gone"?

Carreau: Done.

minrk:
Thanks!

Issue #8724, opened 2015-08-13
IPython (terminal) should be smarter about how it computes invisible characters


asmeurer:
The IPython terminal prompt calculates invisible characters by computing it with and without colors and taking the difference (see https://github.com/ipython/ipython/blob/f2d072e9f357dcee0001f1aa2fe88c6494e5dd1f/IPython/core/prompts.py#L353). 

But there are issues with this. For one thing, custom prompts could have side effects, and this causes it to be rendered multiple times (although I doubt this is a real issue in the wild). More importantly, there can be other invisible characters not accounted for by the built-in color codes, like custom ANSI escape sequences.  If you do this, the out and in2 prompts are indented too far, because it counts characters that should be invisible.

The correct way to test this is to look for `\[` and `\]` (i.e., `\001` and `\002`) blocks, which is what readline does. readline may have a function to check the visible length (I haven't checked), although it shouldn't be hard to implement manually. 

The code also implicitly assumes that the number of invisible characters is static, which also might not be true for a custom prompt with `LazyEvaluate` formatting. 

I'll likely be sending a PR to fix this.

minrk:
> I'll likely be sending a PR to fix this.

Great! Sounds like you've thought it through. Looking forward to it.

asmeurer:
Reading the readline C code, I believe the readline function would be `expand_prompt`, which isn't exposed in the Python API (even in the rl library).

asmeurer:
The basic fix (counting all invisible characters, not just known color codes) is easy. 

The more advanced fix (taking into consideration that a dynamic prompt might have a different number invisible characters each time it is displayed) is awkward with the existing API, which implicitly assumes that the number invisible characters is the same for any given template. PromptManager.invisible_characters is a dictionary mapping prompt type (`in`, `out`, etc.) to number of invisible characters. Really, it should be a function which is called on a prompt string. You can only count the number of invisible characters on a prompt string once it is rendered, and you should only render it once.

My personal use-cases don't have this situation, but it's not hard to imagine some (e.g., say you put a colored git branch in your prompt, but the branch is not there if your current directory is not in a git repository). The current code would do the wrong thing if you don't use the known IPython color codes, and even after the basic fix checking for `\001`, and `\002` it would do the wrong thing if you change directories inside the session, or some other change where the number of color codes changes.

takluyver:
We're planning to move away from readline in favour of prompt_toolkit, in any case, so this may become moot.

[...]

PR #8738, opened 2015-08-16
Be a little smarter about invisible characters in terminal prompts

asmeurer:
This is a partial fix to #8724.

Previously, only known color codes were considered to be invisible. Now, it
looks for any kind of invisible sequence as defined by the \001 \002 delimiters
(which is what readline uses).

The situation could still be improved, as it still assumes that the number of
invisible characters is constant for a given template. Making this work
correctly with the existing API is awkward, so I didn't attempt it, especially
since the readline frontend may be removed at some point in the near future.

### DIFF
--- a/IPython/core/prompts.py
+++ b/IPython/core/prompts.py
@@ -257,6 +257,14 @@ def _lenlastline(s):
     return len(s.splitlines()[-1])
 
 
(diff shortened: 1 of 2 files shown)


Carreau:
+1

asmeurer:
Ping.

Carreau:
Pong. 

Ok let's test.

Issue #8741, opened 2015-08-19
Nicer printing of long lists


naught101:
Currently, a list prints horizontally if the result would be less than 80 columns (not sure if this is settable):

```{python}
In [57]: ['a']*15
Out[57]: ['a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a']
```
and vertically if it would be more than 80 columns
```{python}
In [59]: ['a']*16
Out[59]: 
['a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a',
 'a']
```

Often this is a big waste of space, and for things like `dir()`, it can make things hard to read.

Would it be possible to have a multi-column output, where the column width is `max([len(s) for s in list])`? Something like:

```
In [59]: dir('a')
Out[60]:
[__add__',        '__class__',        '__contains__', '__delattr__',
 '__dir__',       '__doc__',          '__eq__',       '__format__',
 '__ge__',        '__getattribute__', '__getitem__',  '__getnewargs__',
 '__gt__',        '__hash__',         '__init__',     '__iter__',
 '__le__',        '__len__',          '__lt__',       '__mod__',
 '__mul__',       '__ne__',           '__new__',      '__reduce__',
 '__reduce_ex__', '__repr__',         '__rmod__',     '__rmul__',
 '__setattr__',   '__sizeof__',       '__str__',      '__subclasshook__',
 'capitalize',    'casefold',         'center',       'count',
 'encode',        'endswith',         'expandtabs',   'find',
 'format',        'format_map',       'index',        'isalnum',
 'isalpha',       'isdecimal',        'isdigit',      'isidentifier',
 'islower',       'isnumeric',        'isprintable',  'isspace',
 'istitle',       'isupper',          'join',         'ljust',
 'lower',         'lstrip',           'maketrans',    'partition',
 'replace',       'rfind',            'rindex',       'rjust',
 'rpartition',    'rsplit',           'rstrip',       'split',
 'splitlines',    'startswith',       'strip',        'swapcase',
 'title',         'translate',        'upper',        'zfill]
 ```

minrk:
This should be doable, I think. The code where this is done is IPython.lib.pretty. We also have `IPython.utils.text.columnize`, which is a function to turn a list of strings into a columnar list (a la `ls`).

[...]

PR #8748, opened 2015-08-20
Row-first option for columnize

naught101:
This adds a row-first sort to the columnize function, to address part of #8741 .

Original columnise: 

``` python
In [9]: print(columnize(items))
a  c  eeeeeeeeeeeeeeeeeeeeeee  eeeeeeeeeeeeeeeeeeeeeee
b  d  eeeeeeeeeeeeeeeeeeeeeee  eeeeeeeeeeeeeeeeeeeeeee
```

New addition:

``` python
In [10]: print(columnize(items, row_first=True))
a                        b                        c                      
d                        eeeeeeeeeeeeeeeeeeeeeee  eeeeeeeeeeeeeeeeeeeeeee
eeeeeeeeeeeeeeeeeeeeeee  eeeeeeeeeeeeeeeeeeeeeee
```

### DIFF
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -661,7 +661,7 @@ def compute_item_matrix(items, row_first=False, empty=None, *args, **kwargs) :
     items
         list of strings to columize
     row_first : (default False)
-        Whether to to compute columns for a row-first matrix instead ofr
+        Whether to compute columns for a row-first matrix instead of
         column-first (default).
     empty : (default None)
         default value to fill list if needed
@@ -726,7 +726,7 @@ def columnize(items, row_first=False, separator='  ', displaywidth=80):
         The strings to process.
 
     row_first : (default False)
-        Whether to to compute columns for a row-first matrix instead ofr
+        Whether to compute columns for a row-first matrix instead of
         column-first (default).
 
     separator : str, optional [default is two spaces]
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -618,28 +618,28 @@ class DollarFormatter(FullEvalFormatter):
 # Utils to columnize a list of string
 #-----------------------------------------------------------------------------
 
-def _col_chunks(l, nrows, row_first=False):
-    """Yield successive nrows-sized column chunks from l."""
+def _col_chunks(l, max_rows, row_first=False):
+    """Yield successive max_rows-sized column chunks from l."""
     if row_first:
-        ncols = (len(l) // nrows) + (len(l) % nrows > 0)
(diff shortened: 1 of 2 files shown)


IPython/utils/text.py
```diff
@@ -714,6 +725,10 @@ def columnize(items, separator='  ', displaywidth=80):
     items : sequence of strings
         The strings to process.
 
+    row_first : (default False)
+        Whether to to compute columns for a row-first matrix instead ofr
```
Carreau: instead `ofr` :-)

Carreau:
Looks good to me. 

Want to add test for the option in `IPython/utils/tests/test_text.py` ?

IPython/utils/text.py
naught101: also, 'to to'.. gah..

naught101:
Guess that was a good idea :)

I think it's ready to go.

Carreau:
> Guess that was a good idea :)

:-)

naught101:
Hrm.. thoughts please:

currently, the row-first option calculates the optimal layout by starting with the max number of rows, then calculating the minimum number of columns, which may end up being much smaller than the display width. For example, this is the current result:

```{python}
In [3]: [s*15 for s in 'abcdef']
Out[3]: 
['aaaaaaaaaaaaaaa', 'bbbbbbbbbbbbbbb', 'ccccccccccccccc'
 'ddddddddddddddd', 'eeeeeeeeeeeeeee', 'fffffffffffffff']
```

Another option would be to maximize the columns, while remaining under the display limit. which would look like this:

```{python}
In [3]: [s*15 for s in 'abcdef']
Out[3]: 
['aaaaaaaaaaaaaaa', 'bbbbbbbbbbbbbbb', 'ccccccccccccccc', 'ddddddddddddddd'
 'eeeeeeeeeeeeeee', 'fffffffffffffff']
```

which might look nicer in some cases. However, it would never be more efficient in terms of space used, and it would require somewhat more complex code. 

Do you care which option is used?

Carreau:
I think the point of solution 1 was to make completion always closer to the cursor if you wrap around, and as the qtconsole is now it's own package, columnize should  not even used anymore (i grepped quickly but you can re-check)... so feel free to refactor any way you want.

Carreau:
We can also merge that and you can update the algo in another PR.

naught101:
Ok, I just had a though, and added a 'spread' option, which makes columnize spread the columns out to the maximum allowable width (optimal spacing was already calculated, I just left justified the separator to match the padding). 4 Tests included.

As far as I am concerned, this is ready to go - good enough for the pretty print issue anyway.

Carreau:
Ok, let's move forward then

PR #8751, opened 2015-08-20
Fixing unescaped quotes

bollwyvl:
Shows up here:
https://github.com/jupyter/nbviewer/issues/490


Carreau:
Curious to know how it ended up there. 

[..search search search..]

https://github.com/ipython/ipython/pull/8679

bollwyvl:
Yeah, seems like a thing that could be tested for, up to validation :)

Thanks for the quick turn-around!

PR #8758, opened 2015-08-22
Columinsed iterables

### LINKED ISSUE #8741: Nicer printing of long lists

naught101:
fixes #8741 - columnised printing of lists, tuples, dicts, sets, and seqs.

This code doesn't yet have updated tests. pull request is WIP.

I'm currently having trouble running tests locally, so pull-requesting to see which tests need addressing.

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -87,6 +87,7 @@ from collections import deque
 
 from IPython.utils.py3compat import PY3, cast_unicode
 from IPython.utils.encoding import get_stream_enc
+from IPython.utils.text import columnize
 
 from io import StringIO
 
@@ -545,15 +546,18 @@ def _seq_pprinter_factory(start, end, basetype):
             return p.text(start + '...' + end)
         step = len(start)
         p.begin_group(step, start)
-        for idx, x in p._enumerate(obj):
+        rows = columnize([repr(l) for l in obj], row_first=True,
+                         separator=", ", displaywidth=p.max_width).split("\n")
+        rows.remove('')
+        for idx, x in p._enumerate(rows):
             if idx:
-                p.text(',')
                 p.breakable()
-            p.pretty(x)
+            p.text(x)
         if len(obj) == 1 and type(obj) is tuple:
             # Special case for 1-item tuples.
             p.text(',')
         p.end_group(step, end)
+
     return inner
 
 
@@ -583,11 +587,13 @@ def _set_pprinter_factory(start, end, basetype):
                 except Exception:
                     # Sometimes the items don't sort.
                     pass
-            for idx, x in p._enumerate(items):
+            rows = columnize([repr(l) for l in items], row_first=True,
+                             separator=", ", displaywidth=p.max_width).split("\n")
+            rows.remove('')
+            for idx, x in p._enumerate(rows):
                 if idx:
-                    p.text(',')
                     p.breakable()
-                p.pretty(x)
+                p.text(x)
             p.end_group(step, end)
     return inner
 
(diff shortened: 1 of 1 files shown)


Carreau:
I think we might still want to keep some of the "breakable" with `...` behavior, otherwise some long iterable with just make terminal or notebook crash.

naught101:
It does use breakables, but only one per line. For long iterables, this will produce the same number of lines in total, but will usually display more elements. How big is the `...` behaviour by default? I mean, is it pushing the any kind of memory limits? Because if not, I think the current behaviour in this PR is ok (assuming it's working how I think it is).

naught101:
I'm not sure... how should the `max_seq_length` argument should work when the sequence is to be columnised. Do we want to split in the middle of a row, even if

I don't really understand how `max_seq_length` works in general. Why do you send the whole list to the pretty printer if you're going to remove most of it at a breakable later? Doesn't that require a whole lot more unnecessary work in the inner pretty printer loop?

Carreau:
> How big is the ... behaviour by default? I mean, is it pushing the any kind of memory limits? Because if not, I think the current behaviour in this PR is ok (assuming it's working how I think it is).

Beyond a few hundereds of line it will crash the notebook. 
I think the shortness was due to terminal usage where you usually want a pretty repr to fit on your screen. 
We have to refer to older devs than me to know why. 


> I don't really understand how max_seq_length works in general. Why do you send the whole list to the pretty printer if you're going to remove most of it at a breakable later? Doesn't that require a whole lot more unnecessary work in the inner pretty printer loop?

Maybe i'm not sure, I haven't looked at the code.  you can try to git blame to see the age of the code, but I would expect a lot of legacy behavior/compatibility that might not make sens anymore. 

I think we start to enter area where only @fperez know how/when this was design (discreetly pinging him see if he can pitch in to say how much of lib pretty is historical behavior)

fperez:
Actually, the person to ping here is @rkern, who was the one who brought `lib/pretty.py` to IPython.  When I added pretty printing by default, I was only using the stdlib `pprint` module, which `pretty` is meant to replace, and Robert was the one who made those enhancements.

takluyver:
I don't think this is quite right - `IPython.lib.pretty` goes to considerable lengths to do sensible printing of nested data structures, and it looks like this will only do nice formatting on one level of lists.

I think the real fix is to make the 'breakable' functionality smarter so we don't get newlines between every list element. I'm trying to work out how to do that in the code, but it's somewhat complicated.

Carreau:
I'm going to close this as it lost some momentum, and we want to investigate better way of replacing lib. pretty. 

I'm tagging as "Closed PR" to remind us to reopen if someone is ready to start the work again, and will close to keep le list of PR shorter and easier to navigate. Feel free to ask if you need this to be reopen. 

Thanks !

PR #8778, opened 2015-08-28
Use isinstance

SylvainCorlay:
I don't think that `magic.py` should need to import `MetaHasTraits`.


ssanderson:
:+1: this seems straightforwardly more correct to me.

PR #8795, opened 2015-09-08
FIX/API: use inspect.signature for python3

tacaswell:
This allows functions with keyword-only arguments to have their signatures introspected.

Also makes this bit of code 3.6 compatible  (@Carreau I see your 3.5 and raise you a 3.6 :stuck_out_tongue_winking_eye: )

Found this via working on https://github.com/matplotlib/matplotlib/pull/4829 where for python 3.3+ we mark the `data` kwarg as keyword-only.

This is a slight change in behavior as it makes all arguments, not just the optional ones.  This can easily be changed back, but leaving it for now as I would propose that including all arguments is a better behavior.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -843,21 +843,22 @@ class IPCompleter(Completer):
                  getattr(call_obj, '__doc__', ''))
 
         if PY3:
-            _keepers = (inspect.Parameter.KEYWORD_ONLY,
-                        inspect.Parameter.POSITIONAL_OR_KEYWORD)
-            try:
-                sig = inspect.signature(call_obj)
-                ret.extend(k for k, v in sig.parameters.items() if
-                           v.kind in _keepers)
-            except ValueError:
-                pass
+            _keeps = (inspect.Parameter.KEYWORD_ONLY,
+                      inspect.Parameter.POSITIONAL_OR_KEYWORD)
+            signature = inspect.signature
         else:
-            try:
-                args, _, _1, defaults = inspect.getargspec(call_obj)
-                if defaults:
-                    ret += args[-len(defaults):]
-            except TypeError:
-                pass
+            import IPython.utils.signatures
+            _keeps = (IPython.utils.signatures.Parameter.KEYWORD_ONLY,
+                      IPython.utils.signatures.Parameter.POSITIONAL_OR_KEYWORD)
+            signature = IPython.utils.signatures.signature
+
+        try:
+            sig = signature(call_obj)
+            ret.extend(k for k, v in sig.parameters.items() if
+                       v.kind in _keeps)
+        except ValueError:
+            pass
+
         return list(set(ret))
 
     def python_func_kw_matches(self,text):
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -841,24 +841,28 @@ class IPCompleter(Completer):
                 call_obj = obj.__call__
         ret += self._default_arguments_from_docstring(
                  getattr(call_obj, '__doc__', ''))
-        try:
-            if PY3:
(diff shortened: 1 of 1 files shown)


IPython/core/completer.py
```diff
@@ -839,22 +839,30 @@ def _default_arguments(self, obj):
...
+                        inspect.Parameter.POSITIONAL_OR_KEYWORD)
+            try:
+                sig = inspect.signature(call_obj)
+                ret.extend(k for k, v in sig.parameters.items() if
+                           v.kind in _keepers)
+            except ValueError:
+                pass
+        else:
+            try:
+                args, _, _1, defaults = inspect.getargspec(call_obj)
```
JanSchulz: I wonder what would happen if on py2, this could a) unwrap (signature does unwrap until a "siganture" is found, getargspec does not) and b) check for a simple `get_additional_completions(...)` function which we could then set in out decorator. `getdoc()` already does something similar...

so:

```
addons = []
while true:
    if hasattr(call_obj, "get_additional_completions"):
        addons.extend(call_obj.get_additional_completitions())
        break
    if hasattr(call_obj, "__wrapped__"):
         call_obj = call_obj.__wrapped__
    else:
         break
[...]
ret.extend(addons)
```

minrk:
We ship a backport of `inspect.signature` that should work on Python 2 in `IPython.utils.signatures.signature`, if you want to try to just use that instead.

JanSchulz:
Then mpl must ship that as well or needs a dependency on ipython...

minrk:
@JanSchulz I'm not sure I understand. This is a PR to IPython, so why would IPython using its own utility affect MPL?

JanSchulz:
This PR is part of the "unpack labeld data for plotting": make it possible to use a call like this `plot("var1", "var2", data=df)` instead of `plot(df["var1"], df["var2"])`. This is currently implemented in matplotlib/matplotlib#4829 with a decorator, which makes all signatures (= code completitions in python <3.4) look like "plot(*args, **kwargs)". @tacaswell added a `__warpped__` attribute and a `__signature__` attribute (see these two parts of the diff: https://github.com/matplotlib/matplotlib/pull/4829/files#diff-9ffb0d1db51a67ee4f37528e00715b3cR1590  https://github.com/matplotlib/matplotlib/pull/4829/files#diff-9ffb0d1db51a67ee4f37528e00715b3cR1795) in case python understands it so that newer python versions get the right signature and therfore the right code completions.

The problem is that to construct the signature object, mpl needs `inspect.signature` so adding that in py2.x even if IPython understands that, we would need `inspect.signature` included in mpl as well. Using a workaround like `get_addon_completions` would not.

tacaswell:
@JanSchulz Just moving that file over is probably a non-starter for now as mpl (still) supports py2.6 so we don't have `OrderedDict`.

tacaswell:
Probably the best route on the mpl side is to try to import IPython and use it if available.  I think this is ok because the only reason we want to tack the signatures on to legacy python is for nice introspection in IPython.

JanSchulz:
But then your PR here should also try to use the internal ipython copy oft inspect (if python doesn't have it...) in all versions...

tacaswell:
no, mpl will only fall back to the IPython version in versions of python which don't already have `signature`.

tacaswell:
@JanSchulz See https://github.com/JanSchulz/matplotlib/pull/4

tacaswell:
@minrk This look good?  

Can this also be back-ported to what ever branch you are going to release off of next?  This is needed to make a feature in matplotlib 1.5 that @fperez pushed for to work well.

minrk:
@tacaswell the next release will be 4.1, so this will get in. We haven't started making breaking changes on IPython to push master to 5.x and start backporting, yet.

PR #8809, opened 2015-09-13
Correctly count newline in tokenize.

### LINKED ISSUE #8697: Tab display of docstring in function args failing with multiline statement above
My setup:
* IPython 3.2.1
* Clean profile (no local configuration)
* Python 2.7.10
* Mac OS X 10.10.3

I've noticed some strange autocomplete issues since updating to V3 and I think I isolated the root of the problem (or at least one problem...). It seems that tab-display of a function docstring shows the wrong docstring if there is a multi-line statement above the object you are trying to inspect. For example:

```python
a = (1,
     2,
     3)

int()
map()
```

If you put the cursor in the parens in `int()` and Shift+Tab, the docstring for `map` pops up instead.

See video of behavior here:
https://www.youtube.com/watch?v=Zevyejt37Qk&feature=youtu.be

Carreau:
fixes #8697

### DIFF
--- a/IPython/utils/tests/test_tokenutil.py
+++ b/IPython/utils/tests/test_tokenutil.py
@@ -13,7 +13,7 @@ def expect_token(expected, cell, cursor_pos):
         if offset + len(line) >= cursor_pos:
             break
         else:
-            offset += len(line)
+            offset += len(line)+1
     column = cursor_pos - offset
     line_with_cursor = '%s|%s' % (line[:column], line[column:])
     nt.assert_equal(token, expected,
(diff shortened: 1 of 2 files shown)


IPython/utils/tests/test_tokenutil.py
```diff
@@ -13,7 +13,7 @@ def expect_token(expected, cell, cursor_pos):
         if offset + len(line) >= cursor_pos:
             break
         else:
-            offset += len(line)
+            offset += len(line)+1
```
Carreau: Fix error message in case of test faillure, eg if `c==22` it currently show `map(|)` , instead of `|map()`

PR #8810, opened 2015-09-13
Update Python 3.5 testing to RC 4

takluyver:
Remember me? ;-)

I remembered that our 3.5 testing is pinned to a specific version, and thought we should probably update it. Hopefully Travis knows about this version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # http://travis-ci.org/#!/ipython/ipython
 language: python
 python:
-    - "3.5.0b4"
+    - "3.5.0rc4"
     - 3.4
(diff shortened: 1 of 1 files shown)


damianavila:
>Remember me? ;-)

Of course... welcome back! :wink:

Also, LGTM...

Additionally, do you really want to keep testing on `3.3`?

takluyver:
I'm not back properly yet, just dropping in to do this.

We'll probably drop testing on 3.3 at some point once 3.5 is out.

Carreau:
+1, happy to see you not really back. 

3.5 is out :-P

Issue #8822, opened 2015-09-20
Incorrect signature given for method declared using (*args) instead of (self, *args)


anntzer:
```
In [1]: class C:
   ...:     def f(*args): pass
   ...:     def g(*args, **kwargs): pass
   ...: c = C()
   ...: 

In [2]: c.f?
Signature: c.f() # oops
Docstring: <no docstring>
File:      ~/<ipython-input-1-b2acea81d98f>
Type:      method

In [3]: c.g?
Signature: c.g(**kwargs) # oops
Docstring: <no docstring>
File:      ~/<ipython-input-1-b2acea81d98f>
Type:      method
```

One may argue that this is (very) poor style but at least the standard library's `inspect.Signature.bind` is declared like this.

Carreau:
Marking as wishlist. If there is a fix we'll probably accept it.

anntzer:
This was fixed in the stdlib at https://hg.python.org/cpython/rev/3544827d42e6
Given that `IPython.utils.signature` is only used for that purpose, would it be acceptable to try importing the stdlib's `inspect.signature` first (which will succeed for recent enough versions of Python) and then fallback on `IPython.util.signature` on failure?

Carreau:
Seem reasonable. 
Do you want to make a PR ?

anntzer:
Closed by #8823.

PR #8823, opened 2015-09-21
Use stdlib inspect.signature if available.

anntzer:
cf #8822.


Carreau:
hum, I think I would prefer a `signature.py` -> `_signature.py` and `signature.py` being:

```
try:
+    # If stdlib provides this functionality, use it instead as it is more up to
+    # date.
+    from inspect import BoundArguments, Parameter, Signature, signature
+except ImportError:
+    from ._signature import BoundArguments, Parameter, Signature, signature
```

Or it will be easy to miss the conditional import. 

Thoughts ?

anntzer:
I am using the same pattern as for stdlib modules which have both a Python and a C version, e.g. `operator.py`.  Perhaps I can add just a note in the initial docstring?
Or I can switch to your approach, I don't really care but I don't feel like we should overengineer this either.

Carreau:
Let's wait for a third opinion then.

minrk:
I like @Carreau's `_signature` proposal.

Carreau:
Awesome, thanks ! Don't hesitate to write a comment when you push a new commit as we dont' get notified !

Merging !

Issue #8757, opened 2015-08-22
Usage of 'ipython --gui=qt5' do not work as advertized in IPython 4.0


ranvik14:
As discussed in an email thread on 'ipython-user@scipy.org' I'm filling an issue, that I have found when using IPython 4.0 with Python 3.4 in a Windows 10 environment.

<Log>
```
PS C:\Users\Viktor> ipython --help
=========
 IPython
=========

Tools for Interactive Computing in Python
=========================================

    A Python shell with automatic history (input and output), dynamic object
    introspection, easier configuration, command completion, access to the
    system shell and more.  IPython can also be embedded in running programs.

Usage

    ipython [subcommand] [options] [-c cmd | -m mod | file] [--] [arg] ...

[Snip]

--gui=<CaselessStrEnum> (InteractiveShellApp.gui)
    Default: None
    Choices: ('glut', 'gtk', 'gtk3', 'osx', 'pyglet', 'qt', 'qt5', 'tk', 'wx')
    Enable GUI event loop integration with any of ('glut', 'gtk', 'gtk3', 'osx',
    'pyglet', 'qt', 'qt5', 'tk', 'wx').

[Snip]

    ipython nbconvert           # convert notebooks to/from other formats

PS C:\Users\Viktor> ipython --gui=qt5
WARNING: Readline services not available or not loaded.
WARNING: Proper color support under MS Windows requires the pyreadline library.
You can find it at:
http://ipython.org/pyreadline.html

Defaulting color scheme to 'NoColor'
Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 24 2015, 22:44:40) [MSC v.1600 64 bit (AMD64)]
Type "copyright", "credits" or "license" for more information.

IPython 4.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.
[TerminalIPythonApp] WARNING | Eventloop or matplotlib integration failed. Is matplotlib installed?
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
c:\python34\lib\site-packages\IPython\terminal\interactiveshell.py in enable_gui(gui, app)
    305         from IPython.lib.inputhook import enable_gui as real_enable_gui
    306         try:
--> 307             return real_enable_gui(gui, app)
    308         except ValueError as e:
    309             raise UsageError("%s" % e)

[...]

PR #8829, opened 2015-09-23
Qt5 fix

zpincus:
Fix for issue #8757.

The problem was that IPython.external.qt was moved to qtconsole.qt, but IPython.external.qt_for_kernel was still trying to import IPython.external.qt.

Rather than copying the qt.py file back to external, this PR merges the logic back into qt_for_kernel.py. Deferring to qt.py was a bit silly because both qt.py and qt_for_kernel.py do basically the same thing but in slightly different ways. Moreover, the assumption of qt.py was that if the QT_API environment variable is set, it must be in an ETS environment and need to comply by the ETS import logic. This is not valid, as ipython sets the QT_API environment variable to communicate intent as to Qt version, in `IPython/lib/inputhook.py:385`

``` python
@inputhook_manager.register('qt5')
class Qt5InputHook(Qt4InputHook):
    def enable(self, app=None):
        os.environ['QT_API'] = 'pyqt5'
        return Qt4InputHook.enable(self, app)
```

With these changes, the logic for what QT api to import is centralized into qt_for_kernel.

I have also moved qt_for_kernel.py from external to lib, because it doesn't seem to be part of any 3rd party library I can find, and especially not now after this change. Though perhaps a deprecated stub should still be left in external, so that anything that depends on `import Python.external.qt_for_kernel` will still work?


minrk:
Let's leave it in external, since `ipykernel` relies on it. If we move it, it definitely needs a shim.

zpincus:
OK, shim added. Should this also issue a deprecation warning to alert people of the new location?

minrk:
I'd rather leave the file where it was, and just fix the bug.

zpincus:
OK, that should do it. I added comments to the effect that the way that matplotlib is queried for the preferred Qt backend is broken if Qt5 is preferred. I don't know enough about matplotlib to fix this though.

minrk:
@zpincus thanks, I'll look into the matplotlib bit.

zpincus:
Here's what I can glean. From [matplotlib's rcsetup](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/rcsetup.py), here are the valid Qt APIs:
```python
validate_qt4 = ValidateInStrings('backend.qt4', ['PyQt4', 'PySide', 'PyQt4v2'])
validate_qt5 = ValidateInStrings('backend.qt5', ['PyQt5'])
```

Note that matplotlib supports PyQt4v2, but `qt_for_kernel.py` as it stands will barf on that unnecessarily.

Also, the currently valid backends are:
```python
interactive_bk = ['GTK', 'GTKAgg', 'GTKCairo', 'MacOSX',
                  'Qt4Agg', 'Qt5Agg', 'TkAgg', 'WX', 'WXAgg', 'CocoaAgg',
                  'GTK3Cairo', 'GTK3Agg', 'WebAgg', 'nbAgg']
```
So I think that all one needs to do is look at the `rc.backend` variable and if it is `Qt4Agg` then go with the value in `rc.backend.qt4` and if it is `Qt5Agg` then go with the value in `rc.backend.qt5`, and if neither `Qt4Agg` nor `Qt5Agg` is the backend, then maybe refuse the temptation to guess which Qt API to use?

minrk:
@zpincus since we are only getting info from matplotlib if it's already running, it should be correct to avoid asking mpl for anything if the backend is not in `{qt4agg, qt5agg}`. Do you want to take a stab at that, or shall I?

zpincus:
I'll sort it out -- thanks for confirmation that the above seems like the right behavior.

minrk:
Thanks!

PR #8834, opened 2015-09-25
Clarify Qt import logic

zpincus:
Address issue where matplotlib can specify Qt versions that qt_for_kernel.py did not previously recognize.

One outstanding question: if the user requests the "qt" GUI, should that be allowed to import Qt5 if Qt4 is not available? Or should Qt5 only be available if explicitly requested?

### DIFF
--- a/IPython/external/qt_for_kernel.py
+++ b/IPython/external/qt_for_kernel.py
@@ -52,7 +52,7 @@ def matplotlib_options(mpl):
             return [QT_API_PYSIDE]
         elif mpqt.lower() == 'pyqt4':
             return [QT_API_PYQT_DEFAULT]
-        elif mpqt.lower() == 'pyqtv2':
+        elif mpqt.lower() == 'pyqt4v2':
             return [QT_API_PYQT]
         raise ImportError("unhandled value for backend.qt4 from matplotlib: %r" %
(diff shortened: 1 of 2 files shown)


IPython/external/qt_for_kernel.py
```diff
@@ -45,22 +40,30 @@
...
+    backend = mpl.rcParams.get('backend', None)
+    if backend == 'Qt4Agg':
+        mpqt = mpl.rcParams.get('backend.qt4', None)
+        if mpqt is None:
+            return None
+        if mpqt.lower() == 'pyside':
+            return [QT_API_PYSIDE]
+        elif mpqt.lower() == 'pyqt4':
+            return [QT_API_PYQT_DEFAULT]
+        elif mpqt.lower() == 'pyqtv2':
```
minrk: pyqt4v2?

minrk:
> One outstanding question: if the user requests the "qt" GUI, should that be allowed to import Qt5 if Qt4 is not available? 

It seems like picking up Qt5 if it's there makes sense.

IPython/external/qt_for_kernel.py
zpincus: Ah, right. Sorry.

PR #8838, opened 2015-09-25
Allow to rebuild docs with Warning this is an old version.

Carreau:
# PR against 2.x

Update 2.x branch to have a warning when rebuilding the docs that docs are out of date.


Carreau:
Add a docs/requirements.txt file, to allow docs to get the right dependencies.
Trying to move the building to Python 3 now.

Carreau:
Ok, this should build with Python3 too, but it is still worth building with python2 as fabric is not working on py3, and will lead to less pages of the docs being built.

Carreau:
I'm going to self-merge that to be able to switch read the doc back on normal settings.

PR #8846, opened 2015-09-28
fallback to in-memory if creating new history session fails

minrk:
With broken network filesystems, it's probably possible for this to happen at other points, but it seems like the most likely time is creation of the intial session. It could potentially also be in the initialization of the database.


minrk:
Related: #2426

damianavila:
I think is OK as is...

PR #8856, opened 2015-10-02
3.x shrink image

funkyfuture:
i admit, i may freaked out a little.

these commits reduce the image's layer's size from ~575MB to ~502MB and from 25 layers to 9.

i did no research what the removal of software could break in depending images. the `iptest`s are passing.

pick what you want. shall i open a similar pr against the master-branch?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+.git*
+.mailmap
+.travis.yml
+IPython/html/static/components/.git
+IPython/html/static/components/.gitignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,36 +31,29 @@ RUN apt-get update -qq \
         python3-dev \
         sqlite3 \
         zlib1g-dev \
- && rm -rf /var/lib/apt/lists/*
-
-# Install the recent pip release
-RUN curl -O https://bootstrap.pypa.io/get-pip.py \
+ && rm -rf /var/lib/apt/lists/* \
+ \
+ `# Install the recent pip release` \
+ && curl -O https://bootstrap.pypa.io/get-pip.py \
  && python2 get-pip.py \
  && python3 get-pip.py \
- && rm get-pip.py
-
-# In order to build from source, need less
-RUN npm install -g 'less@<3.0' \
+ && rm get-pip.py \
+ \
+ `# In order to build from source, need less` \
+ && npm install -g 'less@<3.0' \
  && npm cache clean
 
-RUN pip install invoke
-
-RUN mkdir -p /srv/
-WORKDIR /srv/
 ADD . /srv/ipython
-WORKDIR /srv/ipython/
-RUN chmod -R +rX /srv/ipython
-
-# .[all] only works with -e, so use file://path#egg
-# Can't use -e because ipython2 and ipython3 will clobber each other
-RUN pip2 install --no-cache-dir file:///srv/ipython#egg=ipython[all] sphinx
-RUN pip3 install --no-cache-dir file:///srv/ipython#egg=ipython[all] sphinx
-
-# install kernels
-RUN python2 -m IPython kernelspec install-self
-RUN python3 -m IPython kernelspec install-self
+RUN chmod -R +rX /srv/ipython \
+\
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,6 @@
+.git
+.git*
+.mailmap
+.travis.yml
+IPython/html/static/components/.git
+IPython/html/static/components/.gitignore
```
rgbkrk: These are certainly a good idea to dockerignore.

Dockerfile
```diff
@@ -1,71 +1,66 @@
 # Installs IPython from the current branch
-# Another Docker container should build from this one to get services like the notebook
+# Other Docker images should build from this one to get services like the notebook
```
rgbkrk: This comment is way out of date now that `jupyter/notebook` provides it.

rgbkrk:
Seems fine with me for the 3.x series.

rgbkrk:
> pick what you want. shall i open a similar pr against the master-branch?

If you look in master, the Dockerfile is just an "echo" telling users to use `jupyter/notebook` instead. I wouldn't go after PRs against master.

minrk:
I'm not sure how much more we should be working on this, since there will not be any more 3.x releases unless we find security problems. @funkyfuture can you try to make your PRs against the [jupyter/notebook](https://github.com/jupyter/notebook) repo or [jupyter/docker-stacks](https://github.com/jupyter/docker-stacks) instead?

funkyfuture:
i'm still confused about that ipython-/jupyter-images.
i am using the `ipython/notebook`-image. am i actually supposed to use that?

what should i use if i wanted to use IPython w/o the notebook?

minrk:
@funkyfuture if you want a docker image without the notebook, it probably doesn't make sense to use IPython as a base image, since installing IPython is just `pip install ipython`, and there are no services to run. Starting with 4.0, the `ipython/ipython`image is deprecated, and we don't plan to maintain a base docker image for IPython anymore. If you want the notebook, you can use `jupyter/notebook` or one of the docker-stacks images.

rgbkrk:
It does seem reasonable that people would want an ipython console docker image, I end up using notebook images for that all the time. As for us maintaining it, we wouldn't want it pinned to the latest whims on master. Seems more like something to contribute to the docker-library image suite (`ipython:3`, `ipython:2`, `notebook`, ... etc.)

minrk:
> It does seem reasonable that people would want an ipython console docker image

Does it, though? What would an IPython base image be useful for? I think base images only make sense if:

1. it provides a complete, working environment or service, or
2. it's hard to install

Since base images are a once-ever choice, and changing the base is a nightmare, I can't see any use for an IPython base image without the notebook.

> we wouldn't want it pinned to the latest whims on master

ipython:3.x is already defined as a notebook server, and since 4.x is current stable, there are no planned releases of 3.x in the future, so for an IPython base image that's not a notebook server, it should start with 4.x at the earliest.

An IPython base image would be three lines:

```dockerfile
FROM debian:jessie # already making a potentially unhelpful choice
apt-get -y update && apt-get -y install python3-pip
pip3 install ipython
```

and if it's *not* a base image and Python is already working, getting IPython is one line, and doesn't tie you down by making a bunch of assumptions. Since Dockerfiles are not composable, base images that aren't complete environments are hardly useful, since you can only pick one. If it were more akin to ansible recipes, I'd be all for IPython

rgbkrk:
To clarify, I'm saying *that* base image would belong in https://github.com/docker-library/official-images as `ipython`.

funkyfuture:
thanks for the feedback, I'll open a similar pr for `jupyter/notebook`.

as a casual IPython user, a [note in the image description](https://hub.docker.com/r/ipython/notebook/) guiding me to `jupyter/notebook` would have been useful.

just throwing some cents here:

> An IPython base image would be three lines:

actually two are possible:

    FROM python
    RUN pip install ipython

> Since Dockerfiles are not composable

i think nested builds are planned for Docker.

minrk:
@funkyfuture thanks, I updated the image description to point to jupyter/notebook. The current Dockerfile in this repo does nothing but point to jupyter/notebook already.

PR #8862, opened 2015-10-04
update coverage html report to coverage 4.0

minrk:
private `_harvest_data()` is now `get_data()`

This fixes recent build failures on Jenkins


minrk:
Merging since the Jenkins failures are getting annoying.

takluyver:
Thanks, I'd just gone to do this, as I was also annoyed by Jenkins failure notices in my inbox.

PR #8876, opened 2015-10-07
set `__doc__` for aliases

### LINKED ISSUE #8874: Output of %magic? is not very helpful
The output of `%magic?` (in the qtconsole) is contains some text about magic functions, followed by
```
%2to3:
    Callable object storing the details of one alias.
    
    Instances are registered as magic functions to allow use of aliases.
%2to3-27:
    Callable object storing the details of one alias.
    
    Instances are registered as magic functions to allow use of aliases.
%2to3-35:
    Callable object storing the details of one alias.
    
    Instances are registered as magic functions to allow use of aliases.
%7z:
    Callable object storing the details of one alias.
    
    Instances are registered as magic functions to allow use of aliases.
%7za:
    Callable object storing the details of one alias.
    
    Instances are registered as magic functions to allow use of aliases.
<elided>
```
which I guess is just the list of all executables in my PATH.
After some scrolling I find the actual magics interspersed in between.

Arch Linux standard packages
```
Jupyter QtConsole 4.0.1
Python 3.5.0 (default, Sep 20 2015, 11:28:25) 

```

minrk:
fixes %magic output summary

closes #8874

### DIFF
--- a/IPython/core/alias.py
+++ b/IPython/core/alias.py
@@ -131,6 +131,7 @@ class Alias(object):
         self.shell = shell
         self.name = name
         self.cmd = cmd
+        self.__doc__ = "Alias for `!{}`".format(cmd)
         self.nargs = self.validate()
 
     def validate(self):


anntzer:
If this prints a bunch of `alias for !foo` in %magic?, I would still argue that the output of %magic? would still be pretty useless.  At least the "actual" magics should be sorted ahead of the aliases for system commands.

minrk:
I think preserving the ordering is more important than grouping magics with similar behavior in the output.

anntzer:
Now I'm confused and can't reproduce this anymore, i.e. system aliases don't appear in %magic.  Hum.

PR #8877, opened 2015-10-07
document kernel installation in multiple environments

### LINKED ISSUE #8873: Documentation unclear about install kernels
The docs say:

> make sure that the desired python installation is active (e.g. activate the environment) and ipython is installed

and 

> run once ipython kernelspec install-self --user (or ipython2 ... or ipython3 ... if you want to install specific python versions)

but when I made sure of the first one, only the ipython version fitting to the version of the active environment is on the path. I think the documentation should make clear that this is only possible using the absolute path to the respective other ipython kernel to install, best with an explicit full command-line example.

On a side note, I am wondering if it's not possible to include this step into the conda install routine? Why would I install the conda ipython package without registering the kernel into an existing (to be checked I guess) jupyter package?

minrk:
closes #8873

### DIFF
--- a/docs/source/install/kernel_install.rst
+++ b/docs/source/install/kernel_install.rst
@@ -1,15 +1,51 @@
 .. _kernel_install:
 
-Kernel Installation
--------------------
+Installing the IPython kernel
+=============================
 
-IPython can be installed (different python versions, virtualenv or conda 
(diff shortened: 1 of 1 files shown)


damianavila:
LGTM

PR #8881, opened 2015-10-07
docs favicon

Carreau:
this enable IPython favicon on RTD.

@cameronoelsen , i think you might be able to create a better icon that reflect the fact that it's docs page.

@willingc @captainsafia you want to do the same for all the Jupyter Projects ?
I think also on JupyterDocs a different icon would be nice too to easily distinguish locally launched notebooks from docs pages. 

@ellisonbg though for an icon ? Exactly the same than RTD, but with a different color ?

screen shot 2015-10-07 at 10 10 24


willingc:
@Carreau I like it :baby_chick: Perhaps drop the jupyter orange into the grey section of RTDs for locally launched.

rgbkrk:
I think I'd like to just merge this since I don't think Cameron will get to it right away - can always make another PR.

PR #8886, opened 2015-10-08
TST: Decode UTF-8 paths before passing them to `os.path` functions

jakirkham:
Fixes an issue with this test under certain environments. This issue appeared in both Python 2 & 3. It's a bit surprising that this issue didn't show up sooner or, for that matter, shows up so infrequently.

### DIFF
--- a/IPython/utils/tests/test_path.py
+++ b/IPython/utils/tests/test_path.py
@@ -430,11 +430,11 @@ def test_unescape_glob():
 
 def test_ensure_dir_exists():
     with TemporaryDirectory() as td:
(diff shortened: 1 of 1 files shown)


rgbkrk:
Well that's exciting.

minrk:
This is now bypassing the intent of the test, which is that ensure_dir_exists works with non-ascii unicode input.

jakirkham:
Good point. I am trying out this improvement ( https://github.com/ipython/ipython/pull/8887 ).

rgbkrk:
Guh. Sorry.

PR #8887, opened 2015-10-09
Revert changes to ensure_dir_exists

jakirkham:
Only decodes inside system functions to make sure that `path.ensure_dir_exists` handles the case of non-unicode supporting systems.

### DIFF
--- a/IPython/utils/path.py
+++ b/IPython/utils/path.py
@@ -437,6 +437,13 @@ def ensure_dir_exists(path, mode=0o755):
 
     The default permissions are 755, which differ from os.makedirs default of 777.
     """
+    path_exists = False
+    try:
+        path_exists = os.path.exists(path)
+    except UnicodeEncodeError:
+        path = path.encode("utf8")
+        path_exists = os.path.exists(path)
+
     if not os.path.exists(path):
         try:
             os.makedirs(path, mode=mode)
--- a/IPython/utils/path.py
+++ b/IPython/utils/path.py
@@ -441,7 +441,7 @@ def ensure_dir_exists(path, mode=0o755):
     try:
         path_exists = os.path.exists(path)
     except UnicodeEncodeError:
-        path = path.encode(sys.getfilesystemencoding() or "utf8")
+        path = path.encode("utf8")
         path_exists = os.path.exists(path)
 
     if not os.path.exists(path):
--- a/IPython/utils/path.py
+++ b/IPython/utils/path.py
@@ -441,8 +441,12 @@ def ensure_dir_exists(path, mode=0o755):
     try:
         path_exists = os.path.exists(path)
     except UnicodeEncodeError:
-        path = path.encode(sys.getfilesystemencoding() or "utf8")
-        path_exists = os.path.exists(path)
+        try:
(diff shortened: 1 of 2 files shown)


jakirkham:
@minrk, this seems to be a working alternative. Relevant CI Docker builds can be seen here ( https://circleci.com/gh/jakirkham/notebook/tree/docker_ci_3 ).

jakirkham:
cc @rgbkrk

IPython/utils/path.py
```diff
@@ -437,6 +437,13 @@ def ensure_dir_exists(path, mode=0o755):
 
     The default permissions are 755, which differ from os.makedirs default of 777.
     """
+    path_exists = False
+    try:
+        path_exists = os.path.exists(path)
+    except UnicodeEncodeError:
+        path = path.encode("utf8")
```
minrk: The encoding should be `sys.getfilesystemencoding() or 'utf-8'`, not unconditionally utf8. This second call can also raise, in which case False should probably be returned.

minrk:
Thanks! This is better. One comment about the fallback, but otherwise AOK.

IPython/utils/path.py
jakirkham: Good point. Sorry, should have thought about that.

jakirkham:
Alright, please let me retest in the docker build and then I will report back here.

jakirkham:
Side note: I'm not sure whether you were aware of this, but repositories can be linked on Docker Hub (under Build Settings). This allows one to link it to the repository they are using as `FROM`. That way, when the upstream repository is pushed to or rebuilt it will trigger a rebuild of yours. This can be very handy in quickly diagnosing issues like the one I ran into last night. As the build errors, there will not be an image available to pull so you don't need to worry about somebody getting a broken image.

jakirkham:
So, that actually failed. I think I know a way around it, but it won't be pretty.

jakirkham:
Ok, that change works. Is this too ugly?

jakirkham:
Let me try one more thing to see if there isn't a better way than this.

jakirkham:
I think this is the best we can hope for at present, unfortunately. If you want it structured differently, I can give that a go.

minrk:
If opening the unicode path doesn't work, then ensure_dir_exists shouldn't claim that it does. Returning False on UnicodeError is probably the right thing for ensure_dir_exists to do. No amount of unicode-forcing in this API is going to fix that. I think instead of coercing to bytes, we should just return False, and skip the unicode test in cases where unicode paths don't work.

jakirkham:
I don't think we need this any more. For that matter, I think we can revert my old changes. Sorry about that.

See this ( https://github.com/jupyter/notebook/pull/547#issuecomment-146908634 ) for my new solution.

jakirkham:
I reverted all my changes and my old commit. The test is now back to the way it was.

minrk:
Thanks, @jakirkham. In the future, you can squash PRs to a single commit with `git rebase -i master` and use the `squash` or `fix` commands.

jakirkham:
Sorry @minrk. I will keep that in mind for the future.

PR #8889, opened 2015-10-09
Better document the %%latex magic latex support.

### LINKED ISSUE #569: Limited Functionality of %%latex cell magic
It looks like when using the %%latex cell magic only math environments are rendered in the notebook. 

in a cell like this: 
```
%% latex 
\[
\hat{H}\Psi = E\Psi
\]
\begin{itemize}
\item First item
\end{itemize}
```

The equation will render correctly in the output cell but not the `itemize` environment. This seems to happen for other latex environments as well, and it is not very consistent. For example the align environment works but the equation environment does not. 

The docs simply say `Render the cell as a block of latex` for `%% latex`. Is this a bug, or is there only a subset of latex that is supported. If the latter is the case it should be added to the docs somewhere!

jdfreder:
closes https://github.com/jupyter/notebook/issues/569

### DIFF
--- a/IPython/core/magics/display.py
+++ b/IPython/core/magics/display.py
@@ -40,7 +40,9 @@ class DisplayMagics(Magics):
     def latex(self, line, cell):
         """Render the cell as a block of latex
         
-        This magic only renders the subset of latex defined by MathJax 
+        The subset of latex which is support depends on the implementation in
+        the client.  In the Jupyter Notebook, this magic only renders the subset 
+        of latex defined by MathJax 
         [here](https://docs.mathjax.org/en/v2.5-latest/tex.html)."""
         display(Latex(cell))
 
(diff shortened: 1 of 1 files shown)


minrk:
Thanks, this makes sense, and is true for most users, who assume it's being called in the notebook. It's a bit technically imprecise, though - the magic isn't actually what's responsible for failing to render, it's the notebook. If put through nbconvert -> tex -> pdf, the whole of latex is available. The QtConsole also doesn't use MathJax, and can call out to 'true' latex to render output (or use matplotlib, which has its own latex subset).

So maybe it's best to say specifically that "The Jupyter notebook application only supports mathjax..." or some such.

jdfreder:
Good point,  I should have been clearer.  See updated text

PR #8907, opened 2015-10-13
ipython[notebook] should depend on ipywidgets

### LINKED ISSUE #8905: For backwards compatibility shouldn't ipywidgets be a dependency of the ipython notebook?
I filed this bug with anaconda:

https://github.com/ContinuumIO/anaconda-issues/issues/470#issuecomment-147775813

Right now, `conda install ipython-notebook` provides a broken install for a widget backwards compatible IPython 4. You must seperately install `ipywidgets` for things to work fully in a IPython 3 backward compatible way.

My suggestion to Continuum was to include `ipywidgets` as a hard dependency of `ipython-notebook` 4. They said this was an upstream issue. It may also affect `pip install ipython[notebook]` but I haven't checked.

minrk:
for 3.x compat. Closes #8905

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -187,7 +187,7 @@ extras_require = dict(
     terminal = [],
     kernel = ['ipykernel'],
     nbformat = ['nbformat'],
-    notebook = ['notebook'],
+    notebook = ['notebook', 'ipywidgets'],
     nbconvert = ['nbconvert'],
 )
 install_requires = [


damianavila:
I agree, LGTM.
Should not be another PR in the jupyter/notebook repo as well? Or we just want to keep the backward compatibility here?

Carreau:
No only here. Jupyter/notebook is a new repo, there is no backward compat reason to make a deps on widget. Especially since there is not even depends on the IPython kernel.

damianavila:
> No only here. Jupyter/notebook is a new repo, there is no backward compat reason to make a deps on widget. 

Someone could argue in the other side, but I agree with you... 

>Especially since there is not even depends on the IPython kernel.

Not true :wink: we bless the ipykernel in the notebook package, but I get what you are pointing at...

PR #8908, opened 2015-10-13
set nonzero exit status when running script fails

### LINKED ISSUE #6912: IPython not setting exit status if script errors
I'm using the subprocess module (Popen) to run a script using IPython, and it doesn't appear to be setting the system status correctly for the process.
```python
    #(script run_test.py)
    import subprocess as proc
    command = "ipython test.py"

    p = proc.Popen( command, shell=True, cwd=None, stdout=proc.PIPE,
                stderr=proc.STDOUT )
    ( stdOut, stdErr ) = p.communicate()

    print p.returncode
    if p.returncode:
        raise Exception( stdOut )
```

In this example, "test.py" is just a simple script that raises (just raise Exception()). When I run this script from the command line (Redhat Enterprise 5.11) I get:

    > python run_test.py
    0

Now, if I change my command in "run_test.py" to use python (instead of IPython), I get:

    (using command = "python test.py")
    > python run_test.py
    1 
      File "run_test.py", line 12, in <module>
        raise Exception( stdOut )
    Exception: Traceback (most recent call last):
      File "test.py", line 2, in <module>
        raise Exception()
    Exception

Any tips on how to get ipython to set the system error code correctly?

Fairly:
closes #6912


minrk:
I don't think there are any changes needed outside `shellapp`. I think we just need to be passing `raise_exceptions=True` when we call safe_execfile. Can you give that a try?

Fairly:
Yes, I think you are right. It didn't occur to me. But I found another problem in safe_execfile. When I passed `raise_exceptions=True` to it, it would raise an `SystemExit` exception even the user script returned normally with a `sys.exit(0)` call. Anyway, I will fix it in the next commit.

IPython/core/interactiveshell.py
```diff
@@ -2674,10 +2674,11 @@ def safe_execfile(self, fname, *where, **kw):
                 # 0
                 # For other exit status, we show the exception unless
                 # explicitly silenced, but only in short form.
-                if kw['raise_exceptions']:
-                    raise
```
minrk: This now ignores `SystemExit(0)`. Is that a deliberate change?

Fairly: Yeah. I set `raise_exceptions` to `True` in `shellapp`. If `safe_execfile` doesn't ignore `SystemExit(0)`, an `SystemError` exception will be raised even user's scripts exit normally. 
Actually, I think this is `safe_execfile` intended to do, as it says in comments.
```python
# If the call was made with 0 or None exit status (sys.exit(0)
# or sys.exit() ), don't bother showing a traceback, as both of
# these are considered normal by the OS:
# > python -c'import sys;sys.exit(0)'; echo $?
# 0
# > python -c'import sys;sys.exit()'; echo $?
# 0
# For other exit status, we show the exception unless
# explicitly silenced, but only in short form.
```

minrk: Terrific, thanks for clarifying.

minrk:
Thanks, merged as a4d98464c801c7475c8b2169e9a372970b72ebfb to avoid the extra merge commits.

fHachenberg:
So how can one interrupt execution of an IPython script and return a statuscode != 0 ?

raise SystemExit(4) 

does not stop execution, because SystemExit is caught unconditionally and never reraised in InteractiveShell.run_code.

PR #8916, opened 2015-10-19
Add \S prompt char for whitespace as wide as the count (similar to \D)

wbolster:
This adds `\S` as a prompt replacement to insert spaces with the same width as the count, similar to what `\D` does with dots. This is mainly useful in continuation prompts, as it allows the continuation prompt to line up nicely with the initial line.

Example configuration snippet:

``` python
c.PromptManager.in_template =  '\\# >>> '
c.PromptManager.in2_template = '\\S ... '
```

Contrived example (continuation lines not indented only to show where the continuation input lines start) using this configuration:

```
2 >>> sum([
  ... 1,
  ... 2,
  ... 3,
  ... ])
```

### DIFF
--- a/IPython/core/prompts.py
+++ b/IPython/core/prompts.py
@@ -160,9 +160,10 @@ prompt_abbreviations = {
     # can get numbers displayed in whatever color they want.
     r'\N': '{count}',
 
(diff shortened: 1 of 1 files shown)


minrk:
I like it.

PR #8925, opened 2015-10-25
For timeit magic, no s at "loops" if only one loop

Naereen:
Just a tiny tiny grammatical correction: no s at "_loops_" if there is only one loop.
It affects the `%timeit` magic, its doc, and the message about "an intermediate result being cached".

A message like "1 loop**s**, best of 3: 307 ms per loop" always bothered me :smile: as there is only one loop.
So ~~loops~~ → **loop** if there is only one.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -79,7 +79,11 @@ class TimeitResult(object):
         self._precision = precision
 
     def _repr_pretty_(self, p , cycle):
-         unic =  u"%d loops, best of %d: %s per loop" % (self.loops, self.repeat,
+         if self.loops == 1:  # No s at "loops" if only one loop
+             unic =  u"%d loop, best of %d: %s per loop" % (self.loops, self.repeat,
+                                            _format_time(self.best, self._precision))
+         else:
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks that's great !
I suppose we don't have to handle the case where there is 0  loops :-P

Naereen:
Well I was wondering about it, but it should never arise right?

Even using `-n0` parameter does not force 0 loop:
```python
In [22]: %timeit -n0 time.sleep(2)
1 loops, best of 3: 2 s per loop
```

Carreau:
Yep, I think the 0 case if fine as is.
I guess we could consider 0 loop if the code was raising an exception, but then we never get to the printing part. 

One other way would have been :
```
u"%d loop%s, best of %d: %s per loop" % ('' if self.loops == 1 else 's' , ...)
```
It should handle correctly 0, but that's bikeshedding.

Naereen:
I know the inline `.. if .. else ..` syntax, but it would have been more confuse here, the `".." % (..)` part of the line was already quite long.

That way, we clearly see that we deal with a special case for priting (I added a comment explaining this if someone works on that line again).

PR #8930, opened 2015-10-27
3.x Close `devnull`at exit to avoid a `ResourceWarning` (Python 3)

jakirkham:
Simply this bugfix ( https://github.com/ipython/ipython/pull/8929 ) ported to the `3.x` line.


minrk:
I'm fine merging this fix, though we should be aware that there is not likely to be another release from the 3.x branch unless we discover new security issues.

jakirkham:
That's fine. I have already started switching over for most of my regular use.

PR #8933, opened 2015-10-28
fix instantiation of shell in test_inline_twice

minrk:
test was failing because get_ipython() returns an instance where inline backend doesn't make sense.


minrk:
Merging since the test failures are annoying me.

PR #8947, opened 2015-11-04
Regression: restore syntax highlighting of source in object_info

jbohannon:
screen shot 2015-11-04 at 11 54 09 am

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -564,7 +564,10 @@ class Inspector:
             for title, key in fields:
                 field = info[key]
                 if field is not None:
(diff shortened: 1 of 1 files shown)


Carreau:
Screenshot with the fix (on my machine). 
screen shot 2015-11-04 at 11 54 31

minrk:
I noticed this recently. We always used to highlight source, but somehow we stopped doing it in IPython 3. Not sure why nobody noticed.

Carreau:
It's was only not highlighting on **some** object. Highlighting on modules and others was still working, and always working in notebook IIRC.

PR #8950, opened 2015-11-05
[WIP] replace PyColorize with Pygments

Carreau:
Just manually porting the color themes.

Just annoying because pygments `red` is "bold red", `darkgreen` is lighter than `green` on light background. `brown` is "yellow".  `turquoise` and `darkteal` are the same. `*foo*` is suppose to be the bold version of `foo` , and logically `*darkgreen*` == `green` ...and `*gray*` == `lightgray`

screen shot 2015-11-05 at 09 40 09

(this missing word is "blink", that as you expect blinks)

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -566,6 +566,7 @@ class Inspector:
                 if field is not None:
                     if key == "source":
                         displayfields.append((title, self.format(cast_unicode(field.rstrip()))))
+                        displayfields.append((title, PyColorize.hl(cast_unicode(field.rstrip()))))
                     else:
                         displayfields.append((title, field.rstrip()))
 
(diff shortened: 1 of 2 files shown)


minrk:
Definitely an improvement. What I *really* want is #289, where we only tokenize these things, and let frontends do the actual coloring.

Carreau:
I'm not completely sure how to do #289 in a pure terminal frontend, 
Seeing your **love** of CSS do you want to comme up with a spec for CSS for terminals ? :-P

minrk:
I'm not either, that's part of why it's one of our longest-standing issues. This is still a good improvement. I just like the opportunity to bring up one of my old pet issues when it's vaguely relevant.

Carreau:
It is an **horror** to automatically convert IPython colorscheme to Pygments ones. 
is it ok if we do not do auto conversion and manually only port the light and dark themes ?

Carreau:
wrong approach. Closing will send a different in progress PR .

alexmojaki:
Was there ever any more progress on this? I agree that having two coloring methods is weird. Plus I want to introduce new functionality in ultratb that uses PyColorize and I don't want to work on deprecated code.

PR #8953, opened 2015-11-07
DeprecationWarning on deprecated colors API

Carreau:
Ok, been deprecated since 2009.
We can probably remove it.

### DIFF
--- a/IPython/core/excolors.py
+++ b/IPython/core/excolors.py
@@ -3,6 +3,8 @@
 Color schemes for exception handling code in IPython.
 """
 
+import warnings
+
 #*****************************************************************************
 #       Copyright (C) 2005-2006 Fernando Perez <fperez@colorado.edu>
 #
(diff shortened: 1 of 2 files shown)


Carreau:
WTF unrelated error reproductible ? Maybe becasue we test on Python 3.5.0.rc4, sent other PR to update that.

IPython/core/excolors.py
```diff
@@ -128,8 +130,17 @@ def exception_colors():
     return ex_colors
 
 
+class Deprec(object):
+
+    def __init__(self, wrapped_obj):
+        self.wrapped=wrapped_obj
+
+    def __getattr__(self, name):
+        warnings.warn("Using ExceptionColors global is deprecated", UserWarning)
```
minrk: Should be a DeprecationWarning, presumably.

Carreau: I was unsure as DeprecationWarning are diable by default.

Carreau: (but Done)

PR #8959, opened 2015-11-08
Set literal

remyleone:


IPython/lib/tests/test_pretty.py
```diff
@@ -115,8 +115,8 @@ def test_sets():
     """
     Test that set and frozenset use Python 3 formatting.
     """
-    objects = [set(), frozenset(), set([1]), frozenset([1]), set([1, 2]),
```
Carreau: so here the non literal seem to be on purpose: `Test that set and frozenset use Python 3 formatting` 
Otherwise, we would have tested the round-tripping of the string representation through eval.

Carreau:
In general we try to avoid huge cleanup PRs, this one is not too big, so it's fine. 
One of the reason is that there is often the possibility to introduce changes to things that are made on purpose (see here example of frozenset), and also because big cleanup that touch many file, make it much harder to git blame, or bisect a specific part of code.

sieben:
@Carreau Fixed

sieben:
@Carreau Are you still interested or should it be closed?

Carreau:
Yes, sorry. Thanks !

PR #8961, opened 2015-11-08
Simplify boolean expression

remyleone:


IPython/core/magics/basic.py
```diff
@@ -454,7 +454,7 @@ def doctest_mode(self, parameter_s=''):
         save_dstore('rc_active_types',disp_formatter.active_types)
         save_dstore('prompt_templates',(pm.in_template, pm.in2_template, pm.out_template))
 
-        if mode == False:
+        if not mode:
```
Carreau: could `mode` be `None` , in which case this is not equivavlent ?

minrk: can mode be None? Doesn't seem like it should be. I don't know what it would mean if it were.

minrk: Upon reading, I don't think it makes sense for mode to be None. Merging.

PR #8963, opened 2015-11-08
Default argument is mutable

remyleone:


Carreau:
Test are failing, it seem like the mutable behavior might have been expected in one case. Likely linereader from a quicklook.

Carreau:
`linereader` seem to be quite an old function, it can likely be replaced by a generator and yield statement now that we only support reasonably recent python versions.

sieben:
@Carreau Do you think it can be merged now?

Carreau:
I'll have a look,  we just don't get notified when there is new commit, so I haven't looked at it. 

I will.

Carreau:
I don't have any objection, though the code is old enough that there might be reason to rely on mutability. I re-read and didn't found any usage of the mutability so +1. 

I'd like @fperez, @minrk and @takluyver input on that too.

IPython/core/magics/code.py
```diff
@@ -500,7 +500,7 @@ def _edit_macro(self,mname,macro):
 
     @skip_doctest
     @line_magic
-    def edit(self, parameter_s='',last_call=['','']):
+    def edit(self, parameter_s='', last_call=None):
```
takluyver: This one is deliberate, and your change will break it. I know it's not considered best practice, but mutable parameters work as a sneaky way to have persistent storage between calls to the same function, so we need to be really careful when 'cleaning up' something like this.

IPython/core/oinspect.py
```diff
@@ -870,8 +870,7 @@ def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
 
         return object_info(**out)
 
-    def psearch(self,pattern,ns_table,ns_search=[],
-                ignore_case=False,show_all=False):
+    def psearch(self, pattern, ns_table, ns_search=None, ignore_case=False, show_all=False):
```
takluyver: As this is just going to be used for iteration, it might be better to make the default the empty tuple `()` rather than defaulting to `None` and then checking for that in code.

IPython/extensions/autoreload.py
```diff
@@ -345,7 +345,7 @@ def __call__(self):
         return self.obj
 
 
-def superreload(module, reload=reload, old_objects={}):
+def superreload(module, reload=reload, old_objects=None):
```
takluyver: I'm not yet sure whether or not this one was deliberate. The autoreload code is rather... interesting.

IPython/utils/wildcard.py
```diff
@@ -20,9 +20,11 @@
 from IPython.utils.dir2 import dir2
 from .py3compat import iteritems
 
-def create_typestr2type_dicts(dont_include_in_type2typestr=["lambda"]):
+def create_typestr2type_dicts(dont_include_in_type2typestr=None):
```
takluyver: Again, this could be made a tuple (or a frozenset, if we wanted to get clever, as it's used for membership checks - but it doesn't matter for one item).

However, since this function is only called once, I'd be inclined just to hardcode the value.

IPython/core/magics/code.py
Carreau: Indeed I missed this one as the default arg is modified by a call to `_find_edit_target`.... Good catch.

PR #8976, opened 2015-11-11
enable codecov

Carreau:
Because coveralls is starting to extremely annoy me, by never fixing any bugs, and regularly asking me to give RW permission to all repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,11 @@ before_install:
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
 install:
     - pip install -f travis-wheels/wheelhouse -r requirements.txt -e file://$PWD#egg=ipython[test] coveralls
+    - pip install codecov
 script:
     - cd /tmp && iptest --coverage xml && cd -
 after_success:
     - cp /tmp/ipy_coverage.xml ./
     - cp /tmp/.coverage ./
     - coveralls
+    - codecov


Carreau:
Also the huge advantage is that CodeCov provide a Chrome and FF extension that show you the coverage **in github file browser** !

.travis.yml
```diff
@@ -11,9 +11,11 @@ before_install:
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
 install:
     - pip install -f travis-wheels/wheelhouse -r requirements.txt -e file://$PWD#egg=ipython[test] coveralls
+    - pip install codecov
```
minrk: should be added to previous line

Carreau: I felt he line was starting to be long, but fair enough.

Carreau: actually on the same like it make pip crash (tried severeal times):

```
$ pip install -f travis-wheels/wheelhouse -r requirements.txt -e file://$PWD#egg=ipython[test] coveralls codecov
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
Fatal Python error: Py_Initialize: Unable to get the locale encoding
ImportError: No module named 'encodings'
Current thread 0x00007f4e9854f700 (most recent call first):
/home/travis/build.sh: line 45:  1845 Aborted                 (core dumped) pip install -f travis-wheels/wheelhouse -r requirements.txt -e file://$PWD#egg=ipython[test] coveralls codecov
```

minrk: makes no sense, but OK.

minrk:
We should probably remove coveralls now.

.travis.yml
Carreau: > makes no sense, but OK.

Yeah, No clue why either. I'll put that on travis weird infrastructure, 
it's failling only only on 3.5, but not for me locally.

minrk: Yeah, I've seen some intermittent failures only on 3.5 recently on other repos in the past few days, as well. I didn't mean to imply that *you* made no sense, only that Travis' behavior makes no sense, as is often the case.

PR #8977, opened 2015-11-12
Coverage highlight

Carreau:
Actually cover more use case of the parser, just because I'm going to ~~break~~ improve it in one of my other PRs.


Carreau:
rebased

PR #8978, opened 2015-11-12
Untest install_ext

Carreau:
install ext is deprecated, let's show the right way of oing things and stop testing it. 

The fake extension is now a real package...


IPython/core/tests/test_magic.py
```diff
@@ -611,8 +611,8 @@ def test_extension():
     try:
         _ip.ipython_dir = tmpdir.name
         nt.assert_raises(ImportError, _ip.magic, "load_ext daft_extension")
-        url = os.path.join(os.path.dirname(__file__), "daft_extension.py")
-        _ip.magic("install_ext %s" % url)
+        url = os.path.join(os.path.dirname(__file__), "daft_extension")
+        _ip.system("pip install %s" % url)
```
minrk: should be `{sys.executable} -m pip` to avoid path weirdness

Carreau: good point. I home that this syntax will not [go away](https://mail.python.org/pipermail/distutils-sig/2015-November/027536.html)

Carreau:
Oops, push the wrong branch 2 weeks ago, fixed. Also cleaned up some imports.

PR #8979, opened 2015-11-12
Actually warn that some function will be removed.

Carreau:
these function are unused by IPython, and are now in genutils.
I think we could still get a Major cycle before removing them.

### DIFF
--- a/IPython/testing/decorators.py
+++ b/IPython/testing/decorators.py
@@ -128,7 +128,7 @@ def make_label_dec(label,ds=None):
     True
     """
 
-    warnings.warn("THe function `make_label_dec` is deprecated and might be removed in next major version of IPython", DeprecationWarning)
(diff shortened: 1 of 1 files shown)


IPython/testing/decorators.py
```diff
@@ -280,6 +284,7 @@ def decorated_dummy(dec, name):
     import IPython.testing.decorators as dec
     setup = dec.decorated_dummy(dec.skip_if_no_x11, __name__)
     """
+    warnings.warn("THe function `make_label_dec` is deprecated and might be removed in next major version of IPython", DeprecationWarning)
```
minrk: The

minrk:
+1

damianavila:
:+1:

Carreau:
Done.

PR #8980, opened 2015-11-12
Move ultra-tb test

Carreau:
Some test where available but not ran on CI.  
Also print to internal buffer instead of stdout needed and extra kwarg.

### DIFF
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -2,9 +2,13 @@
 """Tests for IPython.core.ultratb
 """
 import io
+import sys
 import os.path
 from textwrap import dedent
+import traceback
 import unittest
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,9 +11,11 @@ before_install:
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
 install:
     - pip install -f travis-wheels/wheelhouse -r requirements.txt -e file://$PWD#egg=ipython[test] coveralls
+    - pip install codecov
```
minrk: codecov should be added to previous line, if we are using it (are we?)

Carreau: I removed it, it was there mostly so that I can track the increase in coverage on codecov website. 
On the same line it makes pip crash (cf other PR)

Issue #8983, opened 2015-11-13
cannot `cd ..` 


Carreau:
Seen on @njsmith linux laptop, Debian, IPython 4.0.0, Python 3.4.3

we can `cd ./..`, and any other cd command, but not just `cd ..`

From inside a venv. 

Maybe related, `hist -t` give nothing. 

explicitely calling  get_ipython().magic('cd ..')

apparently the cd magic, get only `.` as argument...

njsmith:
A bit of print debugging reveals the following oddity:
```
(Pdb) p DollarFormatter().vformat("..", args=[], kwargs={})
'.'
```

Carreau:
duplicate of #8919 i think.

Carreau:
closed by #8985

PR #8984, opened 2015-11-13
Workaround debian 3.4.3 bug

### LINKED ISSUE #8919: two characters system command are reduced to one in python3.4.3+ under debian unstable
After installing ipython as the following `pip3 install --user ipython`

I currently use fish shell as a shell so this is the following interaction

```
xcombelle@xcombelle-laptop ~> ipython3
Python 3.4.3+ (default, Oct 10 2015, 09:15:38) 

%quickref -> Quick reference.

In [1]: !ls
fish: Unknown command 'l'
fish: l
      ^
```
the expected result would be the list of files and directory

I debug a little bit and think I found the culprit:
by commenting out [`FullEvalFormatter.v_eval definition`](https://github.com/ipython/ipython/blob/master/IPython/utils/text.py#L543) the problem disapear
I think the problem is related to [this python bug patch](https://hg.python.org/cpython/diff/9eae18e8af66/Lib/string.py) where the use and definition of `_vformat` changed

Carreau:
closes #8919  and #8983 

@njsmith can you test ?

I'll clean up if it works.

### DIFF
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -538,9 +538,21 @@ class FullEvalFormatter(Formatter):
         In [4]: f.format('{3*2}')
         Out[4]: u'6'
     """
+
+    def __init__(self, *args, **kwargs):
+        super(FullEvalFormatter, self).__init__(*args, **kwargs)
+        self._debian_343_workaround = False
+        import sys
+        import string
+        if sys.version_info > (3,4):
+            f = string.Formatter()
+            if type(f.vformat('',[], {})) is not str :
+                self._debian_343_workaround = True
+
+
     # copied from Formatter._vformat with minor changes to allow eval
     # and replace the format_spec code with slicing
-    def _vformat(self, format_string, args, kwargs, used_args, recursion_depth):
+    def _vformat(self, format_string, args, kwargs, used_args, recursion_depth, auto_arg_index=0):
         if recursion_depth < 0:
             raise ValueError('Max string recursion exceeded')
         result = []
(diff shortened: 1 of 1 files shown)


Carreau:
Test fixed. @njsmith can get the patch to apply directly on [here](https://patch-diff.githubusercontent.com/raw/ipython/ipython/pull/8984.patch), @xcombelle  might want to try too.

IPython/utils/text.py
```diff
@@ -538,9 +538,21 @@ class FullEvalFormatter(Formatter):
...
     """
+
+    def __init__(self, *args, **kwargs):
+        super(FullEvalFormatter, self).__init__(*args, **kwargs)
+        self._debian_343_workaround = False
+        import sys
+        import string
+        if sys.version_info > (3,4):
+            f = string.Formatter()
+            if type(f.vformat('',[], {})) is not str :
```
njsmith: I haven't tried the patch yet, but I don't understand how this could work, since AFAICT the patch to 3.4 only changes the return value of `_vformat`, not the return value of `vformat`.

njsmith:
I can confirm that the [this patch](https://hg.python.org/cpython/diff/9eae18e8af66/Lib/string.py) is present in Debian testing's current python3.4 package (they appear to pull from the 3.4 branch on a regular basis, without waiting for releases), and that it's also present in the upstream 3.4 branch that is scheduled to be released as 3.4.4 in a few weeks.

njsmith:
BTW wouldn't it be better for the subclass to just override the public `vformat` method (which is a trivial wrapper around `_vformat`) so that you'd be isolated from this kind of change in internal implementation details?

minrk:
@njsmith yes, we should override vformat instead.

minrk:
#8985 overrides `vformat` instead of `_vformat`

Carreau:
Good point.

PR #8985, opened 2015-11-13
override vformat instead of _vformat

### LINKED ISSUE #8919: two characters system command are reduced to one in python3.4.3+ under debian unstable

minrk:
in our eval formatters

avoids being caught out by changes in private API

this needs to be backported in order to support Python 3.4.4

alternative to #8984
closes #8919 and #8983

cc @Carreau, @njsmith

### DIFF
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -540,9 +540,7 @@ class FullEvalFormatter(Formatter):
     """
     # copied from Formatter._vformat with minor changes to allow eval
     # and replace the format_spec code with slicing
-    def _vformat(self, format_string, args, kwargs, used_args, recursion_depth):
-        if recursion_depth < 0:
-            raise ValueError('Max string recursion exceeded')
+    def vformat(self, format_string, args, kwargs):
         result = []
         for literal_text, field_name, format_spec, conversion in \
                 self.parse(format_string):


Carreau:
lets' get that into the wild.

minrk:
@Carreau I think we should get this out in 4.0.1 and 3.2.3 shortly, so we are ahead of 3.4.4. I've pushed this and some setup.py and qt compatibility fixes to 4.0.x. Is there anything else you think we should include in the backport?

Carreau:
No not really, I'm wondering if 4.0.1 necessary if we cannot 4.1 directly, we have quite a log of bug fixes in 4.1.

minrk:
Since it's a compatibility issue, we should probably do 4.0.1 anyway, even if we do 4.1 right after.

PR #9037, opened 2015-11-24
Fix color schemes in Linux

tayfun:
Object information found by prefixing object with a question mark (like
int?) was printing garbled characters.

git-bisect found 2b1c969628871300a4779963e8776b5a8e966f34 to be the
offending commit. A refactoring in that commit forgot to add one color
and also used the wrong object (it used InputTermColors instead of
TermColors).

ipython-colors

### DIFF
--- a/IPython/core/prompts.py
+++ b/IPython/core/prompts.py
@@ -55,17 +55,17 @@ class LazyEvaluate(object):
         self.func = func
         self.args = args
         self.kwargs = kwargs
-    
+
     def __call__(self, **kwargs):
         self.kwargs.update(kwargs)
         return self.func(*self.args, **self.kwargs)
-    
+
     def __str__(self):
         return str(self())
-    
+
     def __unicode__(self):
         return py3compat.unicode_type(self())
-    
+
     def __format__(self, format_spec):
         return format(self(), format_spec)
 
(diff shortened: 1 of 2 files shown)


minrk:
cc @Carreau who has looked at this stuff most recently.

tayfun:
Test failed, I'll be checking what's wrong with it. Thanks for quick response.

Carreau:
`normal` and `in_normal` are the same. should have been dropped on purpose in a subsequent commit. I think I know what's going on, there are `\001` and `\002` lying around.

Carreau:
Note: \001 and \002 to correctly measure line length (to not count colors codes), they need to be stripped out in some places but not others.

tayfun:
@Carreau I've just put back using in_normal. They are not equal as far as I can see. Thanks for looking into it. Tests pass now.

Carreau:
@minrk didi you ment to merge with all the trailing whitespace removed ?

minrk:
Yeah, didn't feel like complaining about it today, and it's only on empty lines, so blame doesn't suffer.

tayfun:
That was vim removing trailing whitespace automatically. Sorry about that, will take more care next time.

Carreau:
> That was vim removing trailing whitespace automatically. Sorry about that, will take more care next time.

No problem, we tend to avoid when there is too many changes, but it slipps through sometimes. Thanks !

PR #9042, opened 2015-11-29
Actually warn that `ipython <subcommand>` is deprecated.

Carreau:
Remove example, and add a 5 sec wait to have people actually read the
warning message.
## 

I see too many people using it, and I myself will continue doing it if I don't a carrot, or a stick.

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -185,7 +185,7 @@ class BaseIPythonApplication(Application):
         super(BaseIPythonApplication, self).__init__(**kwargs)
         # ensure current working directory exists
         try:
-            directory = py3compat.getcwd()
+            py3compat.getcwd()
         except:
             # exit if cwd doesn't exist
             self.log.error("Current working directory doesn't exist.")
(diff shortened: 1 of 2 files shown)


IPython/core/application.py
```diff
@@ -195,6 +195,16 @@ def __init__(self, **kwargs):
...
     #-------------------------------------------------------------------------
 
+    def initialize_subcommand(self, subc, argv=None):
+        if subc in self.deprecated_subcommands:
+            import time
+            self.log.warning("Subcommand `ipython {sub}` is deprecated and will be removed "
+                             "in future versions.".format(sub=subc))
+            self.log.warning("You likely want to use `jupyter {sub}`... continue "
+                             "in 5 sec".format(sub=subc))
+            time.sleep(5)
```
minrk: probably want to catch KeyboardInterrupt during the sleep to avoid traceback on abort.

Carreau: Good point, done.

PR #9059, opened 2015-12-01
don't install test extension

minrk:
which may require sudo and will pollute the system, causing subsequent runs to always fail.

Instead, make the module importable by putting it on sys.path.


takluyver:
Thanks, I was just setting out to investigate this. I think the fix looks good.

PR #9070, opened 2015-12-14
Move most readline code from InteractiveShell to terminal subclass

takluyver:
This is the first stage of my effort to use prompt_toolkit for the terminal interface.

My plan is to create a new subclass of InteractiveShell which deals with the terminal using prompt_toolkit instead of readline. To facilitate this, I am first trying to separate as much of the readline code as possible out of InteractiveShell to the TerminalInteractiveShell subclass (I don't intend the new subclass to inherit from TIS).

I haven't moved any of the config options relating to readline yet, because moving them will break some people's existing config files. But if we're switching to prompt_toolkit by default, readline related config is going to be broken anyway.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -62,6 +62,7 @@ from IPython.utils import PyColorize
 from IPython.utils import io
 from IPython.utils import py3compat
 from IPython.utils import openpy
+from IPython.utils.contexts import NoOpContext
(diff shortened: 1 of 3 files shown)


takluyver:
Milestoned for 4.1. This should have no functional changes, it's just moving some code to a more appropriate place.

takluyver:
Thanks, Min

PR #9073, opened 2015-12-16
[WIP] Heuristic for better error messages

Carreau:
Proof of concept:

```
In [1]: from numpy import ones

In [2]: def stuff(**kwargs):
   ...:     return ones(**kwargs)
   ...:

In [3]: stuff(size=18)
<ipython-input-3-ee3c5cbdb6f2> in <module>()
----> 1 stuff(size=18)

<ipython-input-2-542e3f8f2f33> in stuff(**kwargs)
      1 def stuff(**kwargs):
----> 2     return ones(**kwargs)
      3

TypeError: ones() got an unexpected keyword argument 'size'. Did you meant one of 'shape' ?
```

See the `did you mean` at the end ? 
English need to be better of course, and here the `one of` is extra as it finds only 1 match. 

Anyway I **think** this would be tremendously useful for beginners to have a framework where before displaying a traceback we could spice it up with potential "fixes".

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -1044,8 +1044,10 @@ class VerboseTB(TBTools):
                     func = last_record[0].f_globals.get(function, None)
                 signature = inspect.signature(func)
                 matches = difflib.get_close_matches(kw, signature.parameters.keys(), 3, 0.3)
-                if matches:
-                    evalue.args = (evalue.args[0]+'. Did you meant one of %s ? ' % ', '.join(map(lambda x : "'"+x+"'",matches)), )
+                if len(matches) == 1:
+                    evalue.args = (evalue.args[0]+". Did you mean '%s' ? " % matches[0]), )
+                elif len(matches) > 1:
+                    evalue.args = (evalue.args[0]+'. Did you mean one of %s ? ' % ', '.join(map(lambda x : "'"+x+"'",matches)), )
         except Exception:
             pass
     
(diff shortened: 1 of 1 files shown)


Carreau:
Inspired by https://github.com/dutc/didyoumean, which is in C, and likely too intusive, and a [recent tweet](https://twitter.com/rachelcolby11/status/676843156957061121).

Carreau:
Does not work (yet) is the function itself is not in the local or global namespace. eg:

```
def foo():
   np.ones()
```

Carreau:
Now work also (a bit) when accessing an (wrong) attribute/methods of an object.
Should do smth similar 
 - with dict KeyErrors, 
 - Out of range access. 
  - SyntaxError unexpected Token, (check for colon on line before)

dsblank:
There is a lot of research on what makes good error messages. I would hate for this to proceed too far without anyone looking at that research.

Carreau:
> There is a lot of research on what makes good error messages. I would hate for this to proceed too far without anyone looking at that research.

I'm looking at what is doable. If you have any insight of regular user issues, and better error message that would be great. 

For now I'm still focusing on technically how we could make these heuristics easy to write. We are definitively far from actually changing how things works in Python itself.

SylvainDe:
Hi,
I stumbled upon this PR via the following Twitter thread : https://twitter.com/cfbolz/status/783313503230844929 . A couple of years ago, I started working on some small project to have such suggestions that might interest you. There is still quite a lot of work to be done (to make it pip installable among other things) but everything can be found here : https://github.com/SylvainDe/DidYouMean-Python .
Any feedback is welcome.

Carreau:
Sweet, that looks great ! I have to test that.

PR #9076, opened 2015-12-17
Write tests for ipdb

Carreau:
Add test for up, down, list and longlist.
## 

I'm refactoring in this area and these things being not tested makes me unconfortable. 

I'm disappointed though that coverage don't work with doctests.

### DIFF
--- a/IPython/core/tests/test_debugger.py
+++ b/IPython/core/tests/test_debugger.py
@@ -100,7 +100,11 @@ def test_ipdb_magics():
     >>> with PdbTestInput([
     ...    'pdef example_function',
     ...    'pdoc ExampleClass',
+    ...    'up',
+    ...    'down',
+    ...    'list',
     ...    'pinfo a',
+    ...    'll',
     ...    'continue',
     ... ]):
     ...     trigger_ipdb()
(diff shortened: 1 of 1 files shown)


minrk:
Is it possible to write this *not* as doctests? We are planning to remove all doctests from IPython, so I'd rather not add more.

takluyver:
Ditto what Min said, I'd like to move away from doctests where possible.

Carreau:
Ok, but I'm not quite sure how. 

CPython test suite of Pdb is using docstrings, and the only other option I can see is subprocess, with re-implementing custom matching for `<blankline>` and `...`. Which seem to me as bad as doctests.

minrk:
Okay. We can merge here, and I'll take it upon my self to rip out the doctests the next time they make me angry.

takluyver:
Tests failing in master on Travis, looks like it's because of this. In particular, it looks like `up` is printing more context than the tests expect.

Carreau:
> Tests failing in master on Travis, looks like it's because of this. In particular, it looks like up is printing more context than the tests expect.

Hum, i'm not sure why they were not failing there, and fail on master. Thanks for fixing.

Carreau:
> Okay. We can merge here, and I'll take it upon my self to rip out the doctests the next time they make me angry.

I would be happy to help you. I just don't see how to test not using strings.

takluyver:
Yeah, I suspect that it may be very tricky for the debugger. Though if we ever add debugger messages to the protocol, I bet the refactoring involved will make testing a lot easier.

takluyver:
Also, good moooooorning California! :sunrise_over_mountains:

PR #9080, opened 2015-12-18
Note removal versions in deprecation warnings

Carreau:
Otherwise we never know when a feature will be removed.

Nothing prevent us to keep longer, but at least giving a version number
instead of "Next Major release" give a hint to users. And also when we
want to cleanup later, we don't have to wonder when the deprecation was
added.
## 

I've put 5.0 as a limit for things already deprecated for a while (like since 1.0). 
And 6.0 for things still potentially used.

### DIFF
--- a/IPython/lib/kernel.py
+++ b/IPython/lib/kernel.py
@@ -5,7 +5,7 @@ Moved to IPython.kernel.connect
 
 import warnings
 warnings.warn("IPython.lib.kernel moved to IPython.kernel.connect in IPython 1.0," 
-        "and will be removed in IPython 6.0.",
+        " and will be removed in IPython 6.0.",
     DeprecationWarning
 )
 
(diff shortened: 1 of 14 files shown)


minrk:
Seems okay.

Carreau:
rebased.

IPython/lib/kernel.py
```diff
@@ -4,7 +4,8 @@
 """
 
 import warnings
-warnings.warn("IPython.lib.kernel moved to IPython.kernel.connect in IPython 1.0",
+warnings.warn("IPython.lib.kernel moved to IPython.kernel.connect in IPython 1.0," 
+        "and will be removed in IPython 6.0.",
```
minrk: missing space

minrk:
Found one typo, otherwise +1.

Carreau:
> Found one typo, otherwise +1.

Here is a space invader which added spaces in the right places: :space_invader:

PR #9086, opened 2015-12-20
ENH added _repr_jpeg_ method for YouTubeVideo object to allow conversion

bkasel:
of the notebook  with Youtube cells to latex and pdf.

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -263,8 +263,14 @@ class YouTubeVideo(IFrame):
         super(YouTubeVideo, self).__init__(src, width, height, **kwargs)
     
     def _repr_jpeg_(self):
-        from urllib import urlopen
-        return urlopen("https://img.youtube.com/vi/{id}/hqdefault.jpg".format(id=self.id)).read()
+        try:
+            from urllib.request import urlopen  # Py3
+        except ImportError:
+            from urllib2 import urlopen
+        try:
+            return urlopen("https://img.youtube.com/vi/{id}/hqdefault.jpg".format(id=self.id)).read()
+        except IOError:
+            return None
 
 class VimeoVideo(IFrame):
     """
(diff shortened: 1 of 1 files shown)


Carreau:
No objection to that. 

Do you want to try adding this information to the docstring ?

It seem also that the canonical way to get the thumbnail is actually `http://img.youtube.com/vi/<insert-youtube-video-id-here>/default.jpg`

And that [other variant are availlable](http://stackoverflow.com/questions/2068344/how-do-i-get-a-youtube-video-thumbnail-from-the-youtube-api) to get bigger sizes.

IPython/lib/display.py
```diff
@@ -255,8 +255,12 @@ class YouTubeVideo(IFrame):
     """
 
     def __init__(self, id, width=400, height=300, **kwargs):
+        self.id=id
         src = "https://www.youtube.com/embed/{0}".format(id)
         super(YouTubeVideo, self).__init__(src, width, height, **kwargs)
+    def _repr_jpeg_(self):
```
Carreau: Also if you are doing modification can you pute a blank line in between the 2 functions ?

bkasel:
OK, I did the changes. I was afraid the HQ version would be to wide for the documents, but it looks good.

Cheers

juhasch:
Would it be possible to return a HTML object, so the Browser loads the image instead of using `urllib` ?
The current implementation requires the notebook server to have a direct internet connection, which might not be the case (e.g. is behind a proxy which requires authentication).

bkasel:
I think this is certainly possible, but then I believe there will not be a jpeg representation in the ipynb file which makes the conversion to latex/pdf more cumbersome. The preferred way the YouTubeVideo is displayed is the Youtube IFrame, which works the way you described.

juhasch:
Yes, I see. Using an `Image` object should work, but it will use `urllib`, too.
Btw. in `IPython/core/display.py` there is a fallback to `urllib2` for Python2.
Might be sensible to do so here, too.

bkasel:
Yes, I will add the fallback.

And I think I will also include the urlopen call in a try/except block to avoid errors when there is no internet connection.

bkasel:
Are there any other changes I have to make?

minrk:
Thanks

PR #9089, opened 2015-12-21
Show some warnings during testing.

Carreau:
It's a good habit, that will push us to cleanup some things, I hope.
Silence a few annoying warnings for a couple of years.

### DIFF
--- a/IPython/testing/iptest.py
+++ b/IPython/testing/iptest.py
@@ -34,6 +34,7 @@ from nose.core import TestProgram
 from nose.plugins import Plugin
 from nose.util import safe_str
 
+from IPython import version_info
 from IPython.utils.py3compat import bytes_to_str
 from IPython.utils.importstring import import_item
 from IPython.testing.plugin.ipdoctest import IPythonDoctest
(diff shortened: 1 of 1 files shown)


IPython/testing/iptest.py
```diff
@@ -58,6 +59,24 @@
 warnings.filterwarnings('ignore', 'wxPython/wxWidgets release number mismatch',
                         UserWarning)
 
+# Enable printing all warnings raise by IPython's modules
+warnings.filterwarnings('default', message='.*', category=Warning, module='IPy.*')
+
+
+if version_info < (6,):
```
minrk: Perhaps 5?

minrk: (or 4.2)

Carreau: Will 4.2 force user to upgrade Traitlets ? I was thinking that for all 4.x required version would not changed, 
that 5.0 would mostly migrate to new API, keeping (some) backward shims + deprecation warnings, and 6.0 actually use new versions. I'm fine changing for earlier versions.

minrk: I think a minor revision bumping a fully backward-compatible minor revision of a dependency is fine. I'm not sure if we will, but I don't think that bumping required traitlets to 4.1 is the sort of change that forces 5.0 (we may well be preparing to release 5.0 by the time we get to it anyway, so it may be moot). There's certainly *no* reason that we would need wait *2* major revisions before bumping a minor revision of a dependency.

Carreau:
Fair enough, changed to 4.2, we can bump to 4.3, or 5.0 by then.

minrk:
Thanks.

PR #9092, opened 2015-12-21
Remove deprecated encodestring in favor of encodebytes

Carreau:
As per the docs:

> [`encodestring` is a deprecated alias](https://docs.python.org/3/library/base64.html#base64.encodestring)

### DIFF
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -16,9 +16,9 @@ from traitlets.config.configurable import SingletonConfigurable
 from traitlets import List, Bool, Unicode
 from IPython.utils.py3compat import cast_unicode, cast_unicode_py2 as u, PY3
 
-if PY3:
+try: # Py3
     from base64 import encodebytes
-else :
+except ImportError: # Py2
     from base64 import encodestring as encodebytes
 
 
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -5,7 +5,6 @@
 # Distributed under the terms of the Modified BSD License.
 
 from io import BytesIO, open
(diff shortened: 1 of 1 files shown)


IPython/lib/latextools.py
```diff
@@ -15,7 +14,12 @@
 from traitlets.config import get_config
 from traitlets.config.configurable import SingletonConfigurable
 from traitlets import List, Bool, Unicode
-from IPython.utils.py3compat import cast_unicode, cast_unicode_py2 as u
+from IPython.utils.py3compat import cast_unicode, cast_unicode_py2 as u, PY3
+
+if PY3:
```
minrk: Rather than checking version, we should check directly for what we are after: `try: import bytes except: import string as bytes`

Carreau: I'm thinking it's easier to cleanup py2 compat code (when we drop it) if it's explicit. I know EAFP. But here I'm leaning more toward "Explicit is better than implicit". I'm alway doubting when I see a `try/except ImportError` for the reason.

minrk: I think it's just a better practice. Checking version numbers directly has often caused bugs because the check ends up being wrong, rather than direct checks for the thing we are using. I know it probably doesn't make a difference in a case like this, but I thin it's a better habit.

minrk: Also, re: "explicit is better than implicit", that's also why I think it should be an *explicit* check for the availability of `encodebytes`, rather than an *implicit* check based on the version flag.

Carreau:
changed to try/except.

minrk:
Thanks!

Issue #9087, opened 2015-12-20
User defined context in print_stack_entry


JamshedVesuna:
When using `ipdb`'s `set_trace()`, `print_stack_entry` sets the [default context to 3 lines](https://github.com/ipython/ipython/blob/master/IPython/core/debugger.py#L335). While this does preserve space, it can be very helpful to have a user defined context size, rather than continuously hitting `l` (list) to display more lines.

It would be great to add a keyword to `ipdb's`'s `set_trace` (such as `context` or even `c`) that passes a context value to [Pdb](https://github.com/ipython/ipython/blob/master/IPython/core/debugger.py#L201). The change to ipython would be minimal: Adding a `context` keyword to `Pdb` and `print_stack_entry`.

Not sure if there is a cleaner way to integrate this (perhaps into [Tracer's set_trace](https://github.com/ipython/ipython/blob/master/IPython/core/debugger.py#L170)?)

Carreau:
We should likely (also) overwrite `do_where` to take as argument the number of line of context to show.

JamshedVesuna:
@Carreau can you elaborate on overwritting `do_where`?

Carreau:
See #9093 for pull request. 

![do_w](https://cloud.githubusercontent.com/assets/335567/11941616/39cf0252-a831-11e5-9ec4-7487e59da705.gif)

JamshedVesuna:
Cool! Can we integrate `do_where` with `print_stack_entry`? 

See [ipdb#35](https://github.com/gotcha/ipdb/issues/35). It seems like there are two ways of adding user defined context values: By editing `bdb.set_trace()`, or by passing a context keyword to `ipdb.set_trace(context=10)`, and pass that to [IPython's Pdb](https://github.com/ipython/ipython/blob/master/IPython/core/debugger.py#L201) as an instance variable, then use that in `print_stack_entry`.  


Unfortunately, there is no easy way to [amend to your PR](https://help.github.com/articles/checking-out-pull-requests-locally/), so I can include your commits in my PR.

Carreau:
> Unfortunately, there is no easy way to amend to your PR, so I can include your commits in my PR.

Oh, you just have the right to push on my PR, but you can send a PR againt my PR or send a new one. I updated mine to add a `context` kwarg and  instance variable.

JamshedVesuna:
[...]

PR #9093, opened 2015-12-21
Add number context argument for do_where/do_w

Carreau:
see #9087 for context (and gif)

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -324,10 +324,10 @@ class Pdb(OldPdb):
     def postloop(self):
         self.shell.set_completer_frame(None)
 
-    def print_stack_trace(self):
+    def print_stack_trace(self, context=5):
         try:
             for frame_lineno in self.stack:
-                self.print_stack_entry(frame_lineno, context = 5)
+                self.print_stack_entry(frame_lineno, context=context)
         except KeyboardInterrupt:
             pass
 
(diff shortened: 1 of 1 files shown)


IPython/core/debugger.py
```diff
@@ -324,15 +324,20 @@ def new_do_restart(self, arg):
...
             for frame_lineno in self.stack:
-                self.print_stack_entry(frame_lineno, context = 5)
+                self.print_stack_entry(frame_lineno, context=context)
         except KeyboardInterrupt:
             pass
 
     def print_stack_entry(self,frame_lineno,prompt_prefix='\n-> ',
-                          context = 3):
+                          context=None):
+        if not context:
```
minrk: is `context=0` valid? If so, all of these `if not context` need to be `if context is not None`

Carreau: I think you mean `if contest is None`, but no 0 does not make sens. Works, as in not crash, but don't make sens, it will print nothing of each frame. And empty string is likely to be a bad input too.
I can also check for purely positive integers too.

Carreau: Side note, here is what 0,1 and 2 do, so that you can judge yourself if it makes sens:
```
ipdb> where 0
  <ipython-input-1-b8ab6ab12ca6>(9)<module>()

> <ipython-input-1-b8ab6ab12ca6>(5)foo()



ipdb> where 1
  <ipython-input-1-b8ab6ab12ca6>(9)<module>()
----> 9 foo()

> <ipython-input-1-b8ab6ab12ca6>(5)foo()
----> 5     raise ValueError('...')



ipdb> where 2
  <ipython-input-1-b8ab6ab12ca6>(9)<module>()
      8     6
----> 9 foo()

> <ipython-input-1-b8ab6ab12ca6>(5)foo()
      4     3
----> 5     raise ValueError('...')
```

JamshedVesuna: I have proper input cleansing in PR #9097

Carreau:
Replaced by #9097

PR #9097, opened 2015-12-22
Added user defined context for Pdb

### LINKED ISSUE #9087: User defined context in print_stack_entry

JamshedVesuna:
Closes #9087

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -29,7 +29,6 @@ from __future__ import print_function
 import bdb
 import functools
 import inspect
-import linecache
 import sys
 
 from IPython import get_ipython
@@ -54,7 +53,6 @@ if '--pydb' in sys.argv:
 
 if has_pydb:
     from pydb import Pdb as OldPdb
-    #print "Using pydb for %run -d and post-mortem" #dbg
     prompt = 'ipydb> '
 else:
     from pdb import Pdb as OldPdb
@@ -94,7 +92,7 @@ class Tracer(object):
     """
 
     @skip_doctest
-    def __init__(self,colors=None):
+    def __init__(self, colors=None):
         """Create a local debugger instance.
 
         Parameters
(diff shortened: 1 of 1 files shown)


Carreau:
You do not need to amend, you can just push more commit, that would have been helpful to see the difference between our PRs. 

The removal of `breaklist = self.get_file_breaks(filename)` was on purpose, it was unused, but agreed it is an unrelated change.

> Just created a new PR #9097 that includes your changes + some style fixes to stay consistent with the rest of the package.

The `if context is None` vs `if not context` is not a style issue, it behave differently for various values like `0` and `''` (empty string), that here make no sens (you at least want 1 line of context). But if we go this route, we might want to check for purely positive non null integers.

JamshedVesuna:
Ah apologies for the PR confusion. I've updated this PR to include your suggested changes. I included the removal of `breaklist = self.get_file_breaks(filename)` as I removed a few miscellaneous print statements as well.

I'd be happy to change it back to `if not context`, as it is cleaner (though that doesn't account for weird inputs like negative ints). 

I noticed that the debugger doesn't raise any exceptions, but rather prints them to stdout. However, when following this format, the error message is printed at the top of the stack trace, making it difficult to find, unless we `break` after printing. Let me know what works best.

Carreau:
I think this is fine let's move forward. We can refine later.

PR #9098, opened 2015-12-22
Support parallel sphinx building

mdboom:
This marks the sphinx extensions that ship with IPython as being safe for parallel reading.

See [here](http://sphinx-doc.org/latest/extdev/index.html#extension-metadata) for the relevant part of the Sphinx spec.

See matplotlib/matplotlib#5597: We'd like to have this in matplotlib, since parallel doc building cuts the build time in half from 9 minutes to 4.5 minutes.


takluyver:
Looks fine to me. We don't actually use the IPython directive in our own docs, so we largely rely on projects like matplotlib and pandas to notice things like this.

Issue #8411, opened 2015-05-07
PyQt5: Could not load requested Qt binding.


dc740:
I'm running python 3 in ubuntu 15.04, and latest ipython release upgraded from pip, and pyqt5 was not detected correctly.
Fixing the missing QtSvg module took me 30 seconds (apt-get install python3-pyqt5.qtsvg), and finding out why iPython would refuse to load PyQt5 took me an hour of reinstalling libraries until I opened the source code.

A descriptive error message would be "nice"

Here is the actual error message:
    Currently-imported Qt library:   'pyqt5'
    PyQt4 installed:                 False
    PyQt5 installed:                 False
    PySide >= 1.0.3 installed:       False
    Tried to load:                   ['pyqt5']

After a lot of debugging, and even installing in virtualenv... I found out that "has_binding" only returns a boolean value inside "IPython/external/qt_loaders.py", so if you have a missing package like QtSvg (which was my case in my regular installation and also in my virtualenv)... then it will return false, even when PyQT5 is actually installed.

jgehrcke:
Same for me:
```
Currently-imported Qt library: 'pyqt5'
PyQt4 installed: False
PyQt5 installed: False
PySide >= 1.0.3 installed: False
Tried to load: ['pyqt5']
```
This is a highly confusing error message, leading to ridiculous debugging work, whereas a proper traceback or error message would have shown immediately that it is just QtSvg which is missing.

Others are also suffering from this confusing error message: http://askubuntu.com/questions/704534/python3-ipython-pylab-on-ubuntu-14-04

Carreau:
could that be considered fixed by #9104 ?

takluyver:
Yep, I think it's significantly clearer now.

PR #9104, opened 2015-12-28
external/qt_loaders: list required Qt modules in import error message

jgehrcke:
This is probably the simplest/easiest way to provide a quick resolution for https://github.com/ipython/ipython/issues/8411.

A more thorough approach would probably be not to shield/hide specific import errors, but to really produce an error message in the moment a certain module cannot be imported. This would, however, require a larger refactoring of the `qt_loaders` module.


takluyver:
:+1:

PR #9109, opened 2016-01-05
Prevent bad `__bool__` from  crashing pinfo

### LINKED ISSUE #9101: Cannot inspect callable that raise in __bool__
The following should not raise.
```python
class NoBoolCall:
    
    def __call__(self):
        pass
    def __bool__(self):
        raise NotImplementedError('Must be implemented')
        
b = NoBoolCall()
b?
```

Likely in `oinspect.py`:

```
835:       if callable_obj:
```

Should be `if callable_obj is not None`, though i'm not sure.

parleur:
Make the change discussed in #9101 and test it (from carreau). 

closes #9101

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -832,7 +832,7 @@ class Inspector:
         else:
             callable_obj = None
 
-        if callable_obj:
+        if callable_obj is not None:
             try:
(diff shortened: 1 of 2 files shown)


Carreau:
+1 ing.

Issue #9091, opened 2015-12-21
update `log.warn` to `log.warning`


Carreau:
As said in the documentation : [`warn` is a deprecated alias](https://docs.python.org/3/library/logging.html?highlight=logging#logging.Logger.warning), we should try not to use it. 

Bonus point to add a warning filter in iptest that warn/fail if we use `log.warn`.

Carreau:
> As said in the documentation : warn is a deprecated alias, we should try not to use it.

This is done. Only the bonus remain.

jamesfe:
What's a good way to tackle this?  I'm interested.

Carreau:
> What's a good way to tackle this? I'm interested.

In `ipython/ipython` this is mostly done, there is a testing question, but its likely to be hard.
I would do the same in the notebook repo:  https://github.com/jupyter/notebook/issues/928

Which is basically just go through files, and rename the methods where necessary. 
(which would almost be automatized)

You're not obliged to tackle all occurrences, only one file at a time is enough.

srinivasreddy:
`srinivas@Srini:~/workspace/github/ipython/IPython$ grep -wr . -e log.warn
`

gave me empty results. so we can close this issue.

Carreau:
I think we can. Thanks !

PR #9110, opened 2016-01-05
[do not merge] Raise if `log.warn` used during tests

Carreau:
cf https://github.com/ipython/ipython/issues/9091,

`log.warn` is deprecated. This **should** make the test fail as it **should** turn usage of `self.warn` into an error.

Though it does not. Any insight as to why ?

[Do not merge]


Carreau:
Ok, nobody knows, closing. Not important.

PR #9111, opened 2016-01-05
Replace log.warn() by log.warning()

parleur:
log.warn() is deprecated so replaced bu log.warning()


Carreau:
cf #9091 for info.

@parleur : want to do the same for notebook and other repos ?

PR #9112, opened 2016-01-05
ensure deprecated_subcommands is defined

minrk:
on the base Application class

cased AttributeError on subclasses other than TerminalIPythonApp


Carreau:
LGTM.

PR #9116, opened 2016-01-07
add deprecated shim for IPython.Config

minrk:
it was accidentally removed altogether in 4.0

This should be included in a 4.0.x release.


Carreau:
+1.

Carreau:
@MinRK, do you take care of backporting ?

minrk:
Yup, I'll do that.

PR #9117, opened 2016-01-08
document existing events

minrk:
so there's a reference for what events fire and when.

### DIFF
--- a/docs/source/config/callbacks.rst
+++ b/docs/source/config/callbacks.rst
@@ -1,6 +1,9 @@
-=====================
-Registering callbacks
-=====================
+.. _events:
+.. _callbacks:
+
+==============
+IPython Events
+==============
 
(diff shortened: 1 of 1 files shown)


takluyver:
:+1:

docs/source/config/callbacks.rst
```diff
@@ -27,9 +30,51 @@ For example::
...
+-----------------
+
+.. code-block:: python
+
+    def shell_initialized(ipython):
+        ...
+
+This event is triggered only once, at the end of setting up IPython.
+Extensions registered to load by default as part of configuration can use this to execute code to finalize setup.
+Callbacks will be passed the IPython instance.
```
takluyver: I think we should say 'the `InteractiveShell` instance'.

minrk: done

PR #8966, opened 2015-11-09
Color everything with pygments

Carreau:
Ok, highly work-in progress, need to cleanup **a lot** of things, 
but it works.

screen shot 2015-11-08 at 20 57 04

PR #9118, opened 2016-01-08
Terminal interface based on prompt_toolkit

takluyver:
This adds a new terminal interface using prompt_toolkit in place of readline. It's on top of my refactoring in #9070.

You can test the new interface by running:

```
python3 -m IPython.terminal.ptshell
```

Multiline editing with syntax highlighting, persistent history and tab completion are all working. So far, I'm really impressed with prompt_toolkit: I've written about 100 lines to make that happen, and while it's taken some research, nothing I've had to do feels like a hack. Kudos to @jonathanslenders :-)

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -170,41 +170,38 @@ def compress_user(path, tilde_expand, tilde_val):
 
 
 
-def penalize_magics_key(word):
-    """key for sorting that penalizes magic commands in the ordering
+def completions_sorting_key(word):
+    """key for sorting completions
 
-    Normal words are left alone.
+    This does several things:
 
-    Magic commands have the initial % moved to the end, e.g.
-    %matplotlib is transformed as follows:
-
-    %matplotlib -> matplotlib%
-
-    [The choice of the final % is arbitrary.]
-
-    Since "matplotlib" < "matplotlib%" as strings, 
-    "timeit" will appear before the magic "%timeit" in the ordering
-
-    For consistency, move "%%" to the end, so cell magics appear *after*
-    line magics with the same name.
-
-    A check is performed that there are no other "%" in the string; 
-    if there are, then the string is not a magic command and is left unchanged.
-
+    - Lowercase all completions, so they are sorted alphabetically with
+      upper and lower case words mingled
+    - Demote any completions starting with underscores to the end
+    - Insert any %magic and %%cellmagic completions in the alphabetical order
+      by their name
     """
+    # Case insensitive sort
+    word = word.lower()
 
-    # Move any % signs from start to end of the key 
-    # provided there are no others elsewhere in the string
+    prio1, prio2 = 0, 0
 
-    if word[:2] == "%%":
+    if word.startswith('__'):
+        prio1 = 2
+    elif word.startswith('_'):
+        prio1 = 1
+
+    if word.startswith('%%'):
+        # If there's another % in there, this is something else, so leave it alone
         if not "%" in word[2:]:
-            return word[2:] + "%%" 
-
-    if word[:1] == "%":
+            word = word[2:]
+            prio2 = 2
+    elif word.startswith('%'):
         if not "%" in word[1:]:
-            return word[1:] + "%"
-    
-    return word
+            word = word[1:]
+            prio2 = 1
+
+    return prio1, word, prio2
 
 
 @undoc
@@ -1206,8 +1203,7 @@ class IPCompleter(Completer):
         # simply collapse the dict into a list for readline, but we'd have
         # richer completion semantics in other evironments.
 
-        # use penalize_magics_key to put magics after variables with same name
-        self.matches = sorted(set(self.matches), key=penalize_magics_key)
+        self.matches = sorted(set(self.matches), key=completions_sorting_key)
 
         #io.rprint('COMP TEXT, MATCHES: %r, %r' % (text, self.matches)) # dbg
         return text, self.matches
--- a/IPython/core/tests/test_shellapp.py
+++ b/IPython/core/tests/test_shellapp.py
@@ -52,8 +52,9 @@ class TestFileToRun(unittest.TestCase, tt.TempFileMixin):
         src = "True\n"
         self.mktmp(src)
 
+        out = 'In [1]: False\n\nIn [2]:'
         err = SQLITE_NOT_AVAILABLE_ERROR if sqlite_err_maybe else None
-        tt.ipexec_validate(self.fname, 'False', err, options=['-i'],
+        tt.ipexec_validate(self.fname, out, err, options=['-i'],
                            commands=['"__file__" in globals()', 'exit()'])
 
     @dec.skip_win32
@@ -63,6 +64,7 @@ class TestFileToRun(unittest.TestCase, tt.TempFileMixin):
         src = "from __future__ import division\n"
         self.mktmp(src)
 
+        out = 'In [1]: float\n\nIn [2]:'
         err = SQLITE_NOT_AVAILABLE_ERROR if sqlite_err_maybe else None
-        tt.ipexec_validate(self.fname, 'float', err, options=['-i'],
+        tt.ipexec_validate(self.fname, out, err, options=['-i'],
                            commands=['type(1/2)', 'exit()'])
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -32,7 +32,7 @@ from IPython.core.shellapp import (
     InteractiveShellApp, shell_flags, shell_aliases
 )
 from IPython.extensions.storemagic import StoreMagics
-from IPython.terminal.interactiveshell import TerminalInteractiveShell
+from .ptshell import PTInteractiveShell as TerminalInteractiveShell
 from IPython.utils import warn
 from IPython.paths import get_ipython_dir
 from traitlets import (
--- a/IPython/terminal/pt_inputhooks/__init__.py
+++ b/IPython/terminal/pt_inputhooks/__init__.py
@@ -1,0 +1,16 @@
+import importlib
+import os
+
+aliases = {
+    'qt4': 'qt'
+}
+
+def get_inputhook_func(gui):
+    if gui in aliases:
+        return get_inputhook_func(aliases[gui])
+
+    if gui == 'qt5':
+        os.environ['QT_API'] = 'pyqt5'
+
+    mod = importlib.import_module('IPython.terminal.pt_inputhooks.'+gui)
+    return mod.inputhook
--- a/IPython/terminal/pt_inputhooks/__init__.py
+++ b/IPython/terminal/pt_inputhooks/__init__.py
@@ -2,7 +2,7 @@ import importlib
 import os
 
 aliases = {
-    'qt4': 'qt'
+    'qt4': 'qt',
 }
 
 def get_inputhook_func(gui):
--- a/IPython/terminal/pt_inputhooks/glut.py
+++ b/IPython/terminal/pt_inputhooks/glut.py
@@ -1,0 +1,141 @@
+"""GLUT Input hook for interactive use with prompt_toolkit
+"""
+from __future__ import print_function
+
+
+# GLUT is quite an old library and it is difficult to ensure proper
+# integration within IPython since original GLUT does not allow to handle
+# events one by one. Instead, it requires for the mainloop to be entered
+# and never returned (there is not even a function to exit he
+# mainloop). Fortunately, there are alternatives such as freeglut
+# (available for linux and windows) and the OSX implementation gives
+# access to a glutCheckLoop() function that blocks itself until a new
+# event is received. This means we have to setup the idle callback to
+# ensure we got at least one event that will unblock the function.
+#
+# Furthermore, it is not possible to install these handlers without a window
+# being first created. We choose to make this window invisible. This means that
+# display mode options are set at this level and user won't be able to change
+# them later without modifying the code. This should probably be made available
+# via IPython options system.
+
+import sys
+import time
+import signal
+import OpenGL.GLUT as glut
+import OpenGL.platform as platform
+from timeit import default_timer as clock
+
+# Frame per second : 60
+# Should probably be an IPython option
+glut_fps = 60
+
+# Display mode : double buffeed + rgba + depth
+# Should probably be an IPython option
+glut_display_mode = (glut.GLUT_DOUBLE |
+                     glut.GLUT_RGBA   |
+                     glut.GLUT_DEPTH)
+
+glutMainLoopEvent = None
+if sys.platform == 'darwin':
+    try:
+        glutCheckLoop = platform.createBaseFunction(
+            'glutCheckLoop', dll=platform.GLUT, resultType=None,
+            argTypes=[],
+            doc='glutCheckLoop(  ) -> None',
+            argNames=(),
+            )
+    except AttributeError:
+        raise RuntimeError(
+            '''Your glut implementation does not allow interactive sessions'''
+            '''Consider installing freeglut.''')
+    glutMainLoopEvent = glutCheckLoop
+elif glut.HAVE_FREEGLUT:
+    glutMainLoopEvent = glut.glutMainLoopEvent
+else:
+    raise RuntimeError(
+        '''Your glut implementation does not allow interactive sessions. '''
+        '''Consider installing freeglut.''')
+
+
+def glut_display():
+    # Dummy display function
+    pass
+
+def glut_idle():
+    # Dummy idle function
+    pass
+
+def glut_close():
+    # Close function only hides the current window
+    glut.glutHideWindow()
+    glutMainLoopEvent()
+
+def glut_int_handler(signum, frame):
+    # Catch sigint and print the defaultipyt   message
+    signal.signal(signal.SIGINT, signal.default_int_handler)
+    print('\nKeyboardInterrupt')
+    # Need to reprint the prompt at this stage
+
+# Initialisation code
+glut.glutInit( sys.argv )
+glut.glutInitDisplayMode( glut_display_mode )
+# This is specific to freeglut
+if bool(glut.glutSetOption):
+    glut.glutSetOption( glut.GLUT_ACTION_ON_WINDOW_CLOSE,
+                        glut.GLUT_ACTION_GLUTMAINLOOP_RETURNS )
+glut.glutCreateWindow( b'ipython' )
+glut.glutReshapeWindow( 1, 1 )
+glut.glutHideWindow( )
+glut.glutWMCloseFunc( glut_close )
+glut.glutDisplayFunc( glut_display )
+glut.glutIdleFunc( glut_idle )
+
+
+def inputhook(context):
+    """Run the pyglet event loop by processing pending events only.
+
+    This keeps processing pending events until stdin is ready.  After
+    processing all pending events, a call to time.sleep is inserted.  This is
+    needed, otherwise, CPU usage is at 100%.  This sleep time should be tuned
+    though for best performance.
+    """
+    # We need to protect against a user pressing Control-C when IPython is
+    # idle and this is running. We trap KeyboardInterrupt and pass.
+
+    signal.signal(signal.SIGINT, glut_int_handler)
+
+    try:
+        t = clock()
+
+        # Make sure the default window is set after a window has been closed
+        if glut.glutGetWindow() == 0:
+            glut.glutSetWindow( 1 )
+            glutMainLoopEvent()
+            return 0
+
+        while not context.input_is_ready():
+            glutMainLoopEvent()
+            # We need to sleep at this point to keep the idle CPU load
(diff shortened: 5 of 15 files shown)


jonathanslenders:
This is really great to see you doing this work @takluyver! I have not had the time to write the shell for IPython. Don't hesitate to let me know if you have any questions about this library.

takluyver:
Thanks! So far I'm working it out OK, and I'm impressed with your docs (though I'm referring to the code quite a bit as well to work out the best way to integrate our stuff). But I'll be in touch if there are any questions I get stuck on!

One thing that gave me a bit of trouble was multiline. Each time the user presses enter, we try to work out whether to prompt for an extra line or execute the code that's already there. I initially tried expressing this by passing a Condition to multiline which ran the 'prompt for more input' test, but it caused some odd behaviour as that changed back and forth. So now I'm passing `multiline=True` and overriding the Enter key event to do the 'prompt for more input' test. I'm perfectly happy with this solution, but I think it's useful for API design to understand where users have gone wrong.

takluyver:
Actually, now that I'm thinking about it, there is something I'd like to ask about:

When the buffer is multiline, is there an easy way to have continuation prompts on each row? Currently there's just blank space to the left, and it's not obvious how much is prompt and how much is indentation. I.e.:

```
# What I currently have:
In [1]: for a in range(5):
            print(a)
            
# What I'd like:
In [1]: for a in range(5):
      :      print(a)
      :      
```

I'm sure this is possible using the deeper API described in 'Building a full screen application', but so far I've been able to stick with the higher level APIs for prompts, and this feels like something that should be simple enough.

jonathanslenders:
Hi @takluyver,

About overriding the Enter key binding, that's fine. That's also what I'm doing. (I think, that behaviour is quite custom for each application.)

About the continuation prompts, that's not yet possible with the ``shortcut`` function, but I will try to implement that asap. (@scopatz from xonsh had the same request: https://github.com/jonathanslenders/python-prompt-toolkit/issues/209 ) Just keep using the higher level API, and assume that it will be possible very soon.

Jonathan

takluyver:
Awesome, thanks! And let me know if I can help with that or anything else in PT - it looks like we're going to be able to ditch a lot of rusty old code, simplify installation and improve the user experience all in one go, so we're very grateful, and accordingly happy to do our bit. :-)

jonathanslenders:
Thanks for the offer! Everything in PT becomes very stable, there's probably not much need to improve it. If there is something that has to be done, it's better documentation. Mostly for all the "internal" stuff that people creating full screen apps would use. Don't worry about that, just focus on the IPython user interface. ;)

takluyver:
Ah, brilliant, and it's good to hear that it's pretty stable as well. I'll let you know of any snags that I run into, and if I meet you at a conference some time, I'll buy you a drink. ;-)

jonathanslenders:
I will remember that ;-)

takluyver:
Another question that's not really specific to prompt_toolkit, but I guess you know quite a bit about terminals. We've had trouble in the past picking default colours, because as far as we can find there's no way to know whether the terminal has a dark or a light background. Do you know of either some secret terminal foo to detect that, or any best practices for picking text colours that will work either way?

jonathanslenders:
Pick colors that work either way. There is no universal way that works everywhere. The default color (black or white) should always be visible. Of course it's possible to specify a background color as well, but that's not respectful regarding the user preferences of the terminal application.

takluyver:
Fair enough. BTW, in the [colours section](http://python-prompt-toolkit.readthedocs.org/en/latest/pages/building_prompts.html#colors) of the docs, it would be useful to see how to override a token's colour back to the plain terminal default, if you want to override a colour set in the default style.

takluyver:
On Windows, pyreadline currently handles translating the ANSI escapes for colours etc. into Windows API calls. Clearly PT can talk to Windows for its own colouring etc., but we'd also like to use colours in other parts of IPython, like for tracebacks.

- Does PT offer an ANSI to Windows-API convertor that we can hook into `sys.stdout` to colour other output? Or should we use a separate package like [colorama](https://pypi.python.org/pypi/colorama) to do this?
- If we use a separate ANSI to Windows terminal package, is there a risk that will interfere with what PT is doing?

takluyver:
Is there a halfway house for the `complete_while_typing` option? In the notebook, we don't offer any completions until the user hits tab to request them, but once the completion popup is visible, it remains visible and updates as the user types more letters. Maybe it's just familiarity, but this feels about right to me.

With `complete_while_typing` on, I find it pops up a bit too eagerly and distracts me, though perhaps I just need to get used to it. With the option off, I find it disappears when I'd expect it to stay.

Carreau:
subscribing to issue.

jasongrout:
(FYI, you can subscribe to an issue by clicking on the "subscribe" button in the upper right, in the "Notifications" heading in the right column.  You don't have to comment to subscribe)

Carreau:
> (FYI, @Carreau, you can subscribe to an issue by clicking on the "subscribe" button in the upper right, in the "Notifications" heading in the right column. You don't have to comment to subscribe)

I know, I did. I wrote the message just for human friendly information.

takluyver:
Now that #9070 has been merged, this accurately shows the changes involved: one new file of 124 lines, and a few changes to completion to improve the ordering of completions. We have syntax highlighting, multiline editing, tab completion and history. The only things I can think of to add are some config options - e.g. to switch to vi input mode.

Did I mention that I'm really impressed with prompt_toolkit? :smile:

Carreau:
> Did I mention that I'm really impressed with prompt_toolkit?

I am too, I need to find more time to contribute to it. 

Is that 4.1 or 5.0 material ?

takluyver:
I think switching to prompt_toolkit properly is a 5.0 task. We could slip it into 4.x labelled 'experimental' and leave the readline interface as the default, but I think I'd rather make it the headline feature of 5.0.

We'll also need to decide: do we leave the readline interface in as an option for a release after switching the default, or take it out straight away? I'm sure there will be some teething troubles, but I don't want to give people the impression that we'll keep maintaining the readline code indefinitely.

Carreau:
> I'm sure there will be some teething troubles, but I don't want to give people the impression that we'll keep maintaining the readline code indefinitel

We can also Make 4.x a LTS and make 5.x no readline, potentially dropping some other legacy code.

rkern:
Rip-rip-rip-out-the-GPL-dependencies!

Though I have looked into making a `prompt_toolkit`-backed `pyreadline`-alike so that it could just be a drop-in replacement without changing the API. It doesn't look like it would be too much work.

takluyver:
Added:

- Config options for the styles used in syntax highlighting, and to enable prompt_toolkit's vi mode
- Support for running in legacy Python, and using appropriate highlighting there (not that there's a massive difference)

If people want to kick the tyres on this, please do. You can start it with:

```
python -m IPython.terminal.ptshell
```

Or to make it run when you type `ipython`, apply this diff:

```diff
diff --git a/IPython/terminal/ipapp.py b/IPython/terminal/ipapp.py
index fe3d008..8268bb9 100755
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -32,7 +32,8 @@
     InteractiveShellApp, shell_flags, shell_aliases
 )
 from IPython.extensions.storemagic import StoreMagics
-from IPython.terminal.interactiveshell import TerminalInteractiveShell
+#from IPython.terminal.interactiveshell import TerminalInteractiveShell
+from .ptshell import PTInteractiveShell as TerminalInteractiveShell
 from IPython.utils import warn
 from IPython.paths import get_ipython_dir
 from traitlets import (
@@ -342,7 +343,7 @@ def start(self):
         # perform any prexec steps:
         if self.interact:
             self.log.debug("Starting IPython's mainloop...")
-            self.shell.mainloop()
+            self.shell.interact()
         else:
             self.log.debug("IPython not interactive...")
```

minrk:
`enable_gui` doesn't apepar to be implemented, so we can't use matplotlib yet.

jonathanslenders:
For the GUI, we probably need this TK inputhook: https://github.com/jonathanslenders/ptpython/blob/master/ptpython/eventloop.py#L19
andpass that to the ``create_eventloop`` function: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/examples/inputhook.py#L73

I will probably move the TK inputhook into prompt_toolkit itself, because it's something that more applications are going to need anyway.

minrk:
We'll probably need to do another implementation of all of our GUI hooks that we've now done for readline and tornado. Since we at least need to support gtk, tk, qt, etc.

jonathanslenders:
@takluyver: I just implemented continuation prompts: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/examples/get-multiline-input.py#L6
Checkout the "master" branch on GitHub. Please let me know if for some reason this doesn't work for you.

takluyver:
Thanks, I've just made of use of that and it seems to be working beautifully. :cake: 

When you've got a moment, I'd appreciate your thoughts on adding colour to other output on Windows. Does PT offer an interface we can use to colour arbitrary other output, outside of the prompt? And if not, is it likely to cause problems if we use another library like colorama alongside PT?

takluyver:
I've borrowed the inputhooks linked above for tk and gtk, and written ones for qt and gtk3.

@jonathanslenders I'm creating the event loop with an inputhook unconditionally, so that we can switch in an inputhook at runtime (as it doesn't look like there's a neat way to change the inputhook in PT's interface). If I've followed the code correctly, that means that it's repeatedly creating and destroying a thread and a pipe in `InputHookContext`, even when the inputhook is a no-op. This seems inefficient, though I don't actually perceive any problem when I run it. Is it worth trying to find a smarter way to switch inputhooks, or are those operations trivial enough that I shouldn't worry?

jonathanslenders:
To answer your first question, regarding colour output. That is possible. There was already a way to do that through ``prompt_toolkit.renderer.print_tokens``. Now, I just made a shortcut to make that easier: ``prompt_toolkit.shortcuts.print_tokens``.

See: http://python-prompt-toolkit.readthedocs.org/en/latest/pages/building_prompts.html#printing-text-output-in-color and
https://github.com/jonathanslenders/python-prompt-toolkit/commit/de1dd9b49d03fcb6a442e314043fbde0949422e8

About the inputhook. You will indeed open a pipe which is maybe not used. But that's nothing to worry about. On my computer, IPython has already 6 fds open when it starts. That is really very lightweight.

takluyver:
We have various bits of code that generate colours using ANSI escape codes directly, because pyreadline on Windows translates the ANSI escapes for colour into Windows API calls. I quite like the idea of ultimately refactoring them to use prompt_toolkit's tokens, but I don't want to try and do that as a part of the initial switch to prompt_toolkit, so I'll probably introduce [colorama](https://pypi.python.org/pypi/colorama) as a dependency on Windows, at least temporarily. That works by wrapping stdout and stderr, but it looks like I can switch to the wrapped versions to run other code, and switch back to unwrapped for the prompt, just to be on the safe side.

IPython definitely creates a lot of things on startup, but that only happens once - and even so, our startup is slower than I would like. It's the creation and desctruction of a pipe+thread several times a second that concerned me. But there doesn't seem to be any noticeable impact, so I'm probably just overestimating the cost of those operations. I still find it hard to think in computer speeds...

Carreau:
Note that https://github.com/ipython/ipython/pull/8966 refactor most of our escape sequence hacks to actually yield tokens all the way down. Right now the last step is to pass through pygments (to generate escape sequences) but that can likely be bypassed.

takluyver:
Nice! I'll try to look at that.

jonathanslenders:
Hi Thomas,

Maybe for ControlC, you want to use ``AbortAction.RETRY``. (Otherwise, the original text is not visible anymore.) If you'd like "KeyboardInterrupt" to be displayed after the original input, I can have a look at that.

```diff
diff --git a/IPython/terminal/ptshell.py b/IPython/terminal/ptshell.py                                                                                                                              
index babaa07..5396f40 100644                                                                                                                                                                       
--- a/IPython/terminal/ptshell.py                                                                                                                                                                   
+++ b/IPython/terminal/ptshell.py                                                                                                                                                                   
@@ -10,7 +10,7 @@                                                                                                                                                                                   
 from prompt_toolkit.filters import HasFocus, HasSelection                                                                                                                                          
 from prompt_toolkit.history import InMemoryHistory                                                                                                                                                 
 from prompt_toolkit.shortcuts import create_prompt_application, create_eventloop                                                                                                                   
-from prompt_toolkit.interface import CommandLineInterface                                                                                                                                          
+from prompt_toolkit.interface import CommandLineInterface, AbortAction                                                                                                                             
 from prompt_toolkit.key_binding.manager import KeyBindingManager                                                                                                                                   
 from prompt_toolkit.key_binding.vi_state import InputMode                                                                                                                                          
 from prompt_toolkit.key_binding.bindings.vi import ViStateFilter                                                                                                                                   
@@ -93,10 +93,6 @@ def _(event):                                                                                                                                                                    
             else:                                                                                                                                                                                  
                 b.insert_text('\n' + (' ' * (indent or 0)))                                                                                                                                        
                                                                                                                                                                                                    
-        @kbmanager.registry.add_binding(Keys.ControlC)                                                                                                                                             
-        def _(event):                                                                                                                                                                              
-            event.current_buffer.reset()                                                                                                                                                           
-                                                                                                                                                                                                   
         # Pre-populate history from IPython's history database                                                                                                                                     
         history = InMemoryHistory()                                                                                                                                                                
         last_cell = u""                                                                                                                                                                            
@@ -133,6 +129,7 @@ def _(event):                                                                                                                                                                   
                             completer=IPythonPTCompleter(self.Completer),                                                                                                                          
                             enable_history_search=True,                                                                                                                                            
                             style=style,                                                                                                                                                           
+                            on_abort=AbortAction.RETRY,                                                                                                                                            
         )                                                                                                                                                                                          
                                                                                                                                                                                                    
         self.pt_cli = CommandLineInterface(app,     
```

takluyver:
Thanks, but I actually like it as it currently is, just clearing the prompt. Ctrl-C is my muscle-memory shortcut for "no, hang on, that's not what I want to do," and I don't see any need to keep the unused code visible. I'm sure when more people come to review this, there will be bikeshedding about it, but for now, that's what feels right to me.

jonathanslenders:
Ok, perfect!

takluyver:
I have now added all the GUI input hooks we support at present - qt, gtk, gtk3, tk, wx, pyglet and glut - and tested them all.

takluyver:
@jonathanslenders another question: with multiline editing, we now have Alt+Enter to force it to execute the code rather than prompting for a new line. :tada:

Can we detect any other key-combos with Enter, like Shift-Enter or Ctrl-Enter? I'd like to use one of these to force a newline. It's not a problem if we can't, but it's a pleasant surprise that we can detect Alt-Enter, so I wonder what else is possible.

Also, out of interest, where is the Alt-Enter keybinding defined? I haven't found it yet in the code.

asmeurer:
I don't think either of those work in most terminals.

takluyver:
Ah, I see Alt-Enter now, it's Escape, ControlJ.

takluyver:
Now that 4.x is a separate branch, I'd like to merge this into master soon so that we have plenty of time to detect problems before releasing 5.0.

At present, it needs a development version of prompt_toolkit (for continuation prompts), so I haven't turned it on by default. Once @jonathanslenders has done a new release, I'll make a new PR that enables it by default and specifies the dependency on prompt_toolkit.

@jonathanslenders I've just added a config option for mouse support, but I left it off by default, following your API. Are there known downsides to enabling it, or would it be reasonable to enable by default? I'm going to turn it on in my own config and see if I run into any problems.

scopatz:
For what it is worth @takluyver, I believe that xonsh has mouse support turned on by default.

takluyver:
Your docs [contradict you](http://xonsh.org/envvars.html#mouse-support) ;-). And in my local testing, the default is also False.

scopatz:
Ahh, you are right.  I have it set True locally.  It might be worth toggling this to True, then.  Sorry for the noise.

takluyver:
No problem, it's good to know that it's working smoothly for you.

Carreau:
> Now that 4.x is a separate branch, I'd like to merge this into master soon so that we have plenty of time to detect problems before releasing 5.0.

I've kept the version number to 4.2, but if a PR changing to 5.0 arrives, I'll merge and then +100

asmeurer:
A common reason to not enable mouse support in terminal applications is that it messes with selection behavior (for copy and paste). Most (all?) terminals have a way to work around this, but not everyone may know about it. If the mouse support here doesn't do that, then it's likely not an issue.

asmeurer:
I guess the xonsh docs also give another argument against mouse support.

jonathanslenders:
@takluyver I hope to do a new prompt_toolkit release, somewhere this month.

Currently, I'm rewriting some code to properly handle arbitrary big inputs. (Right now, it becomes slow around 2000 lines, while my development version handles 100,000 lines smoothly.) This is however irrelevant for most users, so probably I will release first without those changes.

takluyver:
Great! I look forward to the new release. And the large input changes sound amazing, even if I doubt we'd ever use them. I guess that's prompted by things like pyvim?

jonathanslenders:
@takluyver: When do you expect to release a first beta version of IPython 5? Just to make sure that I don't wait too long for pushing a new prompt-toolkit release. If I can wait maybe one or two weeks, that would be good on my side.

Carreau:
> @takluyver: When do you expect to release a first beta version of IPython 5?

Not soon. Is that sufficient for you ?

takluyver:
Yeah, one or two weeks definitely won't be a problem. We don't start doing pre-releases until fairly close to the release, and for IPython 5.0, that's likely to be months away.

That said, I want to get people using this on master fairly soon so that we have plenty of time to discover bugs. If the new prompt_toolkit's coming any time this month, that's fine - if you think it might be longer, I'll temporarily disable the continuation prompts, which should make it work with the released version, and then we can add them back when you have time to do a release.

takluyver:
I decided to disable the continuation prompts so that this can move forwards. It's trivial to re-enable them once there's a new release of prompt_toolkit.

After a bit of fiddling with the tests, this is passing now, so I think it's ready to go.

Carreau:
> After a bit of fiddling with the tests, this is passing now, so I think it's ready to go.

I'll go through that quickly one last time and merge.

Carreau:
Thanks !

takluyver:
:balloon:

asmeurer:
I just tested this. Is it possible for the PTInteractiveShell to inherit from TerminalInteractiveShell so that my existing configuration carries forward? Or is the new configuration too different for this to work?

Carreau:
> so that my existing configuration carries forward?

Which part of your configuration in particular ? it' already inherit form `InteractiveShell` and many methods of `TerminalInteractiveShell` don't make sent for `PTInteractiveShell`.

takluyver:
We could rename `PTInteractiveShell` to `TerminalInteractiveShell` so that it tries to apply all the same config, since config just goes by class names. It doesn't necessarily mean it will all work, though.

asmeurer:
Well I don't have [much](https://github.com/asmeurer/dotfiles/blob/master/.ipython/profile_default/ipython_config.py) configured. Personally, I was jarred my color settings, custom prompt (I need to test if that still works out of the box), and confirm_exit settings were reverted. I guess the problem is that the configuration that was created by IPython ages ago used TerminalInteractiveShell in the skeleton.

takluyver:
Or introduce an empty intermediate class called `TerminalInteractiveShell`. I don't want it to inherit from the old TIS class, though - the aim is to get rid of that.

Prompt configuration will be lost - I'm still thinking about how to make them configurable (and, mildly devil's advocating, whether to even bother - since the two process work, prompt customisation for other frontends has come up occasionally, but never been high enough priority for us to work on it there)

Carreau:
THe color think wil **likely** not work, but I am working on that, and I have some patches in Pygments so that Styles can use `#ansi<color>` to actually match terminals settings. But it will take a few iterations.

takluyver:
The colour config should still work for other bits of colour (tracebacks, `foo?` output, etc.), but it won't affect the input prompts or syntax highlighting as you type. There's a new option `PTInteractiveShell.highlighting_style`, which takes the name of a [Pygments style](http://pygments.org/docs/styles/).

rcarneva:
The new interface is slick, and I really appreciate the incorporation of vi keys. Really nice addition!

takluyver:
Thanks @rcarneva

ndunsworth:
Is there any way to turn off pt in 5.0? 

I really miss the dir like tab completion of readline.  Its the main reason I hit tab. Its a fast dir call.

Please tell me I'm missing some config option in pt or ipy to disable the pulldown autocompletion.

takluyver:
There is no option to switch back to readline. There are discussions going on in other issues about how best to present completions.

ndunsworth:
After a quick search I was not able to find any such discussions, could you provide a few please?

Also where would be a good starting point in the source for completely disabling ipy autocompletion so that readline will work on its own?

Issue #9119, opened 2016-01-09
Infinite while in case of wrong exception


gdoumenc:
In case a code generate an exception with this same exception as cause, the shell never render the exception and stay in loop.

To resolve 'partially' that, we can (al least) just add a simple one step test :
```python
(line 1019)
while evalue:
  formatted_exceptions += self.format_exception_as_a_whole(etype, evalue, etb, lines_of_context,
                                                                  chained_exceptions_tb_offset)
  exception = self.get_parts_of_chained_exception(evalue)

(----v)
  if exception and not exception[1] is evalue:
(----^)
```

takluyver:
Good catch! To go one step better, we could keep a set of ids, and bail out if we reach an id we've already seen:

```python
chained_exc_ids = set()
while evalue:
    chained_exc_ids.add(id(evalue))

    ...
    exception = self.get_parts_of_chained_exception(evalue)
    if exception and id(exception[1]) not in chained_exc_ids:
        ...
```

(something like that, anyway)

Do you want to make a pull request for either of these methods?

gdoumenc:
Your solution is better so I let you do your pull request.. ;)

FYI : I got this issue with pyodbc/sqlalchemy so not unknown package.... :(

Regards

Le 09/01/2016 17:27, Thomas Kluyver a écrit :
>
> Good catch! To go one step better, we could keep a set of ids, and 
> bail out if we reach an id we've already seen:
>
> chained_exc_ids=  set()
> while  evalue:
>      chained_exc_ids.add(id(evalue))
>
>      ...
>      exception=  self.get_parts_of_chained_exception(evalue)
>      if  exceptionand  id(exception[1])not  in  chained_exc_ids:
>          ...
>
> (something like that, anyway)
>
> Do you want to make a pull request for either of these methods?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/ipython/ipython/issues/9119#issuecomment-170257727>.
>

-- 
Guillaume DOUMENC
*FPR / CoWorks*
Your Internet Enabler
http://www.fpr-coworks.com

takluyver:
You're still welcome to do a pull request if you're interested :-)

Have you also filed bugs with those packages? I think there's something wrong if an exception ends up being its own cause.

takluyver:
Closed by #9121

PR #9121, opened 2016-01-10
Check for cycles in chained exceptions

gdoumenc:
Fixing issue #9119

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -1094,12 +1094,14 @@ class VerboseTB(TBTools):
                 etype, evalue, etb = exception
             else:
                 evalue = None
+            chained_exc_ids = set()
             while evalue:
                 formatted_exceptions += self.format_exception_as_a_whole(etype, evalue, etb, lines_of_context,
                                                                          chained_exceptions_tb_offset)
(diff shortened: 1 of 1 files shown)


IPython/core/ultratb.py
```diff
@@ -1094,12 +1094,14 @@ def structured_traceback(self, etype, evalue, etb, tb_offset=None,
                 etype, evalue, etb = exception
             else:
                 evalue = None
+            chained_exc_ids = set()
```
takluyver: I think this misses the first exception value we prepare the traceback for. Of course, if there's a cycle, we'll still pick it up one step later, but I think it would be a bit neater to start the set with evalue in.

gdoumenc: I've let it like that intentionnaly this way to allow seeing there is a cycle.. If you change it your way you should then add a message of loop detection... Helping users to detect a potential issue...

takluyver: Ah, I see what you mean. I think it still won't show you the cycle if it looks like `A-B-C-D-B-...`, but I don't think that's especially important. The main thing is to make sure it doesn't go into an infinite loop. Thanks!

Issue #9122, opened 2016-01-10
mock.call inspection consumes all memory


bartekbrak:
to reproduce:
```
from mock import call
call?
```
Reproduced on python 2.7.10, mock==1.0.1, IPython==4.0.1 and python 3.5, unittest.mock==1.0 so versions shouldn't matter.

The code above will consume all memory.

`call` is actually `call = _Call(from_kall=False)`, `from_kall=True` does not cause the problem, that should be a hint.

While this might be a problem with mock being too magic, this certainly shouldn't lead to a hanged system. I don't use swap, if that matters.

takluyver:
I think mock is being a bit too clever for itself, but IPython definitely shouldn't do that, whatever happens. #9123 should catch it.

PR #9123, opened 2016-01-10
Limit unwrapping __wrapped__ attributes to prevent infinite loops

takluyver:
Closes gh-9122

Some objects lie. Some objects lie again and again and again. Don't trust the objects.


minrk:
Since we already use mock in the tests, want to add a test?

takluyver:
Test added. I didn't test using `mock.call`, because if it does ever fail, I'd like it to produce an error rather than grabbing all the RAM it can and locking up the system.

PR #9124, opened 2016-01-10
Fix max_seq_length input to RepresentationPrinter

wilywampa:
`max_seq_length` is passed to RepresentationPrinter as a positional argument but the order is not correct for `RepresentationPrinter.__init__`. Currently the value passed for `max_seq_len` ends up being used wipes out any `singleton_pprinters` passed in. This fix just passes it as a keyword argument.


takluyver:
Looks good to me; thanks!

PR #9135, opened 2016-01-16
Avoid loading readline in the kernel

takluyver:
This bit of code was loading readline even when there was no need for it. In a Qt console or notebook, try:

```
'readline' in sys.modules
'gnureadline' in sys.modules
```

I plan to rework this code more extensively once we've switched to prompt_toolkit, but for now, this little fix should make the kernel start slightly faster.

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -334,23 +334,25 @@ Currently the magic system has the following functions:""",
         # local shortcut
         shell = self.shell
 
-        import IPython.utils.rlineimpl as readline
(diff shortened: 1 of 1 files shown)


Carreau:
+1

stonebig:
This feature would be nice to have soon for me. Is there a target date for 4.1 release ?

takluyver:
Target date: "soon-ish" ;-) Hopefully within the next month.

stonebig:
ok, thank you for the information.

PR #9138, opened 2016-01-18
log.warn deprecated

Carreau:
Forgotten in  a previous PR.
## 

Do you think a 
`warnings.filterwarnings('default', "The 'warn' method is deprecated, use 'warning' instead", DeprecationWarning, IPython.*)` would be ok, if `'dev' in __version__` ?


takluyver:
I wouldn't object to that.

PR #9142, opened 2016-01-19
Remember list of tempfile for %edit

### LINKED ISSUE #9002: Logging is not comprehensive with regard to %edit
Python 3.4.3, Python 4.0.0.

The IPython session copied to `ipython_session.py` in [this Gist][1] produces the file `my_log.py`, also at that Gist. The first time `%edit` is run, the output line contains the complete contents of the file that was edited. If `%edit` is then called with a function argument, there is no output which contains the contents of the modifications.

If a person in a session is constantly making changes/tweaking a function definition with the `%edit` command, even the most verbose logging session will never capture these modifications, making it difficult to retrace one's steps if they rely on session logs.

[1]: http://gist.github.com/horse-mans/b6691197932935b03fcf

Carreau:
Should fix #9002, still need tests.
There might be a better fix which would be to "mark" the %edit-ed file
code as interactively defined in `In[x]`, though it would likely
conflict with the fact that `In[x]` actually contain `%edit <foo>`
## 

Needs tests

### DIFF
--- a/IPython/core/magics/code.py
+++ b/IPython/core/magics/code.py
@@ -147,6 +147,10 @@ class InteractivelyDefined(Exception):
 class CodeMagics(Magics):
     """Magics related to code management (loading, saving, editing, ...)."""
 
+    def __init__(self, *args, **kwargs):
+        self._knowntemps = set()
+        super(CodeMagics, self).__init__(*args, **kwargs)
+
     @line_magic
     def save(self, parameter_s=''):
         """Save a set of lines or a macro to a given filename.
@@ -660,6 +664,12 @@ class CodeMagics(Magics):
             # nothing was found, warnings have already been issued,
             # just give up.
             return
+
+        if is_temp:
+            self._knowntemps.add(filename)
+        elif (filename in self._knowntemps):
+            is_temp = True
+
 
         # do actual editing here
         print('Editing...', end=' ')


IPython/core/magics/code.py
```diff
@@ -147,6 +147,10 @@ def __init__(self, index):
 class CodeMagics(Magics):
     """Magics related to code management (loading, saving, editing, ...)."""
 
+    def __init__(self, *args, **kwargs):
+        self._knowntemps = set({})
```
takluyver: The `{}` is redundant here - you're making a set from an empty dictionary.

takluyver:
Assigned myself to look into this in more detail

IPython/core/magics/code.py
Carreau: True, fixed.

IPython/core/magics/code.py
```diff
@@ -656,6 +660,7 @@ def edit(self, parameter_s='',last_call=['','']):
             args = str(e.index)
             filename, lineno, is_temp = self._find_edit_target(self.shell, 
                                                        args, opts, last_call)
+            self._knowntemps.add(filename)
```
takluyver: I think this should be outside the `except`, and depend on the `is_temp` variable.

Carreau: Reworked the logic. How does that loks.

IPython/core/magics/code.py
```diff
@@ -694,7 +704,7 @@ def edit(self, parameter_s='',last_call=['','']):
                     self.shell.safe_execfile(filename, self.shell.user_ns,
                                              self.shell.user_ns)
 
-        if is_temp:
+        if is_temp :
```
takluyver: Do you want to remove the space here and squash so that git blame doesn't see you touching this line?

takluyver:
I think that looks fine.

Carreau:
Done.

Issue #8875, opened 2015-10-07
Cannot complete namespace packages (was Where did %rehash go?)


anntzer:
Either %rehash was mistakenly removed, or the doc of %rehashx 

```
In [9]: %rehashx?
Docstring:
Update the alias table with all executable files in $PATH.

This version explicitly checks that every entry in $PATH is a file
with execute access (os.X_OK), so it is much slower than %rehash.

Under Windows, it checks executability as a match against a
'|'-separated string of extensions, stored in the IPython config
variable win_exec_ext.  This defaults to 'exe|com|bat'.

This function also resets the root module cache of module completer,
used on slow filesystems.
File:      /usr/lib/python3.5/site-packages/IPython/core/magics/osm.py

In [10]: %rehash?
Object `%rehash` not found.
```

As a side note I noticed this when I realized that `import mpl_` would not complete `mpl_toolkits` and tried to `%rehash` my cache; I still don't know how to get this to complete properly.

minrk:
Wow, that's an out-of-date docstring. The last version with `%rehash` was 0.10 about five years ago. `%rehashx` is all that we have now. Does it not clear your module completer cache?

anntzer:
At least I still can't complete `import mpl_` into `import mpl_toolkits`.

minrk:
Hm. What's your Python, IPython and OS version? And what do you get for `get_ipython().db.get('rootmodules_cache')`? It only caches imports when it takes a long time to compute them, otherwise it computes them on the fly.

jenshnielsen:
Could it be related to `mpl_toolkits` being a namespace package? Something that we (mpl developers that is)  should perhaps change in the future due to other issues.

minrk:
@jenshnielsen yes, it's possible, however I do find `mpl_toolkits` in my import completions, so I'm trying to figure out what is different between the two environments.

anntzer:
Arch Linux standard packages, on a clean (new) profile.
```
Python 3.5.0 (default, Sep 20 2015, 11:28:25) 
IPython 4.0.0 -- An enhanced Interactive Python.
```
`get_ipython().db.get('rootmodules_cache')` returns None.

jenshnielsen:
@minrk I can reproduce the issue on OSX 10.11 with my pip installed dev ca .1.5.rc2 version of matplotlib but not in a Conda env on the same machine. Not sure what the difference is

minrk:
[...]

PR #9143, opened 2016-01-19
Remove mention to rehash

Carreau:
Fix part of #8875, the import mpl_<tab> not working still need to be
resolved, but can be postponed.

### DIFF
--- a/IPython/core/magics/osm.py
+++ b/IPython/core/magics/osm.py
@@ -98,7 +98,7 @@ class OSMagics(Magics):
           /usr/local/lf9560/bin:/usr/local/intel/compiler70/ia32/bin:...
 
         You can use the alias facility to acess all of $PATH.  See the %rehashx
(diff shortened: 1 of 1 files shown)


IPython/core/magics/osm.py
```diff
@@ -97,9 +97,9 @@ def alias(self, parameter_s=''):
           In [9]: show $$PATH
           /usr/local/lf9560/bin:/usr/local/intel/compiler70/ia32/bin:...
 
-        You can use the alias facility to acess all of $PATH.  See the %rehash
-        and %rehashx functions, which automatically create aliases for the
-        contents of your $PATH.
+        You can use the alias facility to acess all of $PATH.  See the %rehashx
+        function, which automatically create aliases for the contents of your
```
takluyver: create -> creates, now that this is singular

Carreau: Oops, fixed.

Issue #8791, opened 2015-09-04
%%writefile ignores '>>>' at the beginning of lines


petrposik:
If I write the following in the code cell:
```
%%writefile test.txt
>>> fact(5)
120
```
I expect that the text after the %%writefile magic will be present in the text file literally. Instead, only the following is found in the textfile:
```
fact(5)
120
```
IPython probably first filters out the '>>>' string and then passes the result to the cell magic function.It behaves like this in both the notebook, and in the IPython console.  I do not think this is the expected behavior.

Carreau:
o_O thanks for the report. 

Not sure what can be causing this.

Carreau:
@takluyver, this seem to be the input transformers that strips `>>>` before converting %-ish syntax to a `get_ipython().<function call>`.

I'm not sure how to fix as `classic_prompt()` is ment to strip the prompt if `>>>`  appear in one of the first 2 lines (which obviously appear to be the case here). 

And obviously a pasted magics where we start selecting from after the first prompt make sens.

Could `classic_prompt()` yield transformed line, only if all the lines after the `initial_re` match `prompt_re` ?

takluyver:
> Could classic_prompt() yield transformed line, only if all the lines after the initial_re match prompt_re ?

Not really. We have this complex system of transformers handling things line by line so that in a console environment, it can decide after each line whether input is complete, and it should execute, or incomplete, and it should prompt for another line.

We could alleviate this particular case by saying that if the first line begins with `%` or `!`, it must be IPython syntax, so `classic_prompt()` should ignore the rest of the cell. But you'd still have the same problem using writefile if the next line started with `In [1]: `.

Trying to guess what the user means is hard!

Carreau:
I keep seeing this pattern: When we remove users from the equation  90% of all our bugs are solved :-P (10 remaining % being JS, unicode and windows)

takluyver:
See #9144, which should make this a bit smarter. I've actually done it for IPython prompts when inside a cell magic as well.

PR #9144, opened 2016-01-19
Deactivate prompt transformers if the first line matches certain patterns

takluyver:
Deactivate the `>>>` prompt stripper if the first line starts with `%` or `!`, because those are obviously IPython syntax.

Disable the IPython `In [1]:` prompt stripper if the first line starts with `%%`, because we don't want to interfere with the body of a cell magic.

This is all 'what did the user mean' guesswork, but I think this makes our guesses slightly smarter.

We actually had one test that checked that this wasn't happening; I've adjusted it.

Closes gh-8791

### DIFF
--- a/IPython/core/inputtransformer.py
+++ b/IPython/core/inputtransformer.py
@@ -418,12 +418,6 @@ def _strip_prompts(prompt_re, initial_re=None, turnoff_re=None):
     If any prompt is found on the first two lines,
     prompts will be stripped from the rest of the block.
     """
-    def pass_thru(line1):
-        "Pass lines through unaltered until the end of the cell"
-        line = line1
-        while line is not None:
-            line = (yield line)
-
     if initial_re is None:
         initial_re = prompt_re
     line = ''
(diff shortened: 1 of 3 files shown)


IPython/core/inputtransformer.py
```diff
@@ -428,6 +434,14 @@ def _strip_prompts(prompt_re, initial_re=None):
         if line is None:
             continue
         out, n1 = initial_re.subn('', line, count=1)
+        if turnoff_re and not n1:
+            if turnoff_re.match(line):
+                # We're in e.g. a cell magic; disable this transformer for
+                # the rest of the cell.
+                yield from pass_thru(line)
```
Carreau: Yes +1, let's drop python 2.7 !

takluyver:
...I think I just forgot that we still support Python 2 (*Legacy Python* :wink: )

IPython/core/inputtransformer.py
```diff
@@ -446,8 +460,8 @@ def _strip_prompts(prompt_re, initial_re=None):
         
         else:
             # Prompts not in input - wait for reset
-            while line is not None:
-                line = (yield line)
+            yield from pass_thru(line)
```
Carreau: :notes:  2-7 is dying ! :notes:  (with the theme of everything is awesome)

takluyver:
OK, tests passing on Python 2 again

PR #9150, opened 2016-01-20
Add ability to Embed to raise an exception. 

### LINKED ISSUE #9149: Cannot escape IPython.embed() when called repeatedly within function inside a for-loop
I am running `foo.py` from the command line, and it includes a function has a for loop, inside of which is a function that has an IPython.embed() statement. This is to initialize an interpreter at that point within the function to have access to those local variables. 

However, when trying to escape from the embedded interpreter, it only opens a new interpreter (presumably up to k times for the for loop), effectively locking that terminal.

It would be nice to have a command analogous to `ipdb`'s `q`, which immediately throws an exception and exits to the CLI in which `foo.py` was executed.

Carreau:
Allow Embeded instances to raise on exit.

This is use full in case where an IPython embed instance is created in a loop:

```
from IPython import embed
while True:
    embed(header="You will be stuck there")
```

Now using `%raise_on_exist True` in an embeded shell allows the embeded
instannce to raise on exit.

Potential improvement would be flags to pass a custom exception type as
well as a custom message to raise.

Closes #9149

### DIFF
--- a/IPython/terminal/embed.py
+++ b/IPython/terminal/embed.py
@@ -46,37 +46,41 @@ class EmbeddedMagics(Magics):
             print ("This embedded IPython will not reactivate anymore "
                    "once you exit.")
 
-    @line_magic
-    def raise_on_exit(self, parameter_s=''):
-        """%raise_on_exit [True|False]: Make the current embeded kernel to raise an exception on exit.
+    if sys.version_info > (3,):
 
-        You can change that again during current session by calling `%raise_on_exit` with `True`/`False` as parameter 
+        @line_magic
+        def raise_on_exit(self, parameter_s=''):
+            """%raise_on_exit [True|False]: Make the current embeded kernel to raise an exception on exit.
 
-        This function (after asking for confirmation) sets an internal flag so
-        that an embedded IPython will raise a `KillEmbeded` Exception on exit.  This is useful to
-        permanently exit a loop that create IPython embed instance.
-        """
-        parameter_s = parameter_s.strip().lower()
+            You can change that again during current session by calling `%raise_on_exit` with `True`/`False` as parameter
 
-        should_raise = None
-        if parameter_s in {'yes', 'raise', 'true', '1'}:
-            should_raise = True
-        elif parameter_s in {'no', 'false', '0', 'None'}:
-            should_raise = False
-        else:
-            if self.shell.should_raise :
-                print("The current embed instance will raise on exit, use `%raise_on_exit False` to disable")
-                return 
+            This function (after asking for confirmation) sets an internal flag so
+            that an embedded IPython will raise a `KillEmbeded` Exception on exit.  This is useful to
+            permanently exit a loop that create IPython embed instance.
+
+            Available only for Python 3.
+            """
+            parameter_s = parameter_s.strip().lower()
+
+            should_raise = None
+            if parameter_s in {'yes', 'raise', 'true', '1'}:
+                should_raise = True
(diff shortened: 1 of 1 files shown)


takluyver:
Given the similar purpose, maybe this should be an option to the existing `%kill_embedded` magic?

Carreau:
> Given the similar purpose, maybe this should be an option to the existing %kill_embedded magic?

I though about that but : 

1) that makes it more difficult to be Py3 only. 
2) it makes it harder to remember and to see the status of wether it will raise or not. 

Also I want to expand the raise one to take an exception class and optional message, which will make the usage of `kill_embed` extremely complex. 

I was also asked to have the magic just called `raise` and actually exit immediately, so that it would be much more python like:

```
In[12] %raise ValueError('I should raise this on exit')
```

which I found was A bit too magical.

takluyver:
> 1) that makes it more difficult to be Py3 only. 

I don't think it makes much sense to arbitrarily make new features Python 3 only. If there's something where it's an extra burden to support Python 2, it's fine not to, but I don't think we should just put all new features behind `if PY3:` blocks just for the sake of it.

> I was also asked to have the magic just called raise and actually exit immediately

I actually assumed that was what it did, until I looked more carefully. I wouldn't call it `%raise`, but +1 to calling it something like `%exit_raise` and having it exit and raise. That doesn't feel particularly magical, and it's simpler to use.

I don't think we need to support giving it a specific exception to raise, at least for now.

Carreau:
> I don't think it makes much sense to arbitrarily make new features Python 3 only. If there's something where it's an extra burden to support Python 2, it's fine not to, but I don't think we should just put all new features behind if PY3: blocks just for the sake of it.

i don't completely agree, but fine. 



> I actually assumed that was what it did, until I looked more carefully. I wouldn't call it %raise, but +1 to calling it something like %exit_raise and having it exit and raise. That doesn't feel particularly magical, and it's simpler to use.

Done.

IPython/terminal/embed.py
```diff
@@ -46,12 +47,28 @@ def kill_embedded(self, parameter_s=''):
                    "once you exit.")
 
 
+    @line_magic
+    def exit_raise(self, parameter_s=''):
+        """%exit_raise Make the current embedded kernel exit and raise and exception.
+
+        This function (after asking for confirmation) sets an internal flag so
```
takluyver: I don't think it will ask for confirmation. Confusingly, `ask_exit()` means 'ask the shell to exit', not 'ask the user if they want to exit'.

However, if someone has typed out `exit_raise`, I think we can be pretty sure that's what they want, so I'd just change the docstring.

IPython/terminal/embed.py
```diff
@@ -130,6 +147,10 @@ def __call__(self, header='', local_ns=None, module=None, dummy=None,
         if self.exit_msg is not None:
             print(self.exit_msg)
 
+        if self.should_raise:
+            raise KillEmbeded('This instance has been marked as must raise on exit.')
```
takluyver: This error message is not very clear. Let's say something like "Embedded IPython raising error, as user requested"

Carreau:
updated.

takluyver:
I think this could go in a 4.x release if you like - it should be pretty safe. But it's also fine to leave for 5.0 if you want to be strict about it.

Carreau:
I'm fine having it in. I tagged at 5.0 not to delay 4.1

takluyver:
Let's live (slightly) dangerously.

PR #9154, opened 2016-01-20
Clarify multiple ipython installations

jhamrick:
Fixes ipython/ipykernel#27

@takluyver

### DIFF
--- a/docs/source/install/kernel_install.rst
+++ b/docs/source/install/kernel_install.rst
@@ -3,9 +3,13 @@
 Installing the IPython kernel
 =============================
 
-The Jupyter Notebook and other frontends automatically ensure that the IPython kernel is available.
+The Jupyter Notebook and other frontends automatically ensure that the IPython kernel is available
+for whatever version of Python you used to install the Jupyter Notebook.
+
 However, if you want to use a kernel with a different version of Python, or in a virtualenv or conda environment,
-you'll need to install that manually.
+you'll need to install that manually. Note that this will **overwrite** the existing IPython kernel. If you want
+to have multiple installations of the IPython kernel, rather than overwriting the existing installation, see the
+documentation below on `Multiple IPython installations`_.
 
 Using the Python version or environment for which you want to set up the kernel, run::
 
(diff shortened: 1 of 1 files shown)


jhamrick:
Ooops, sorry I didn't mean to make a branch on the actual IPython repo! Should I delete that and recreate this from my own fork?

takluyver:
Don't worry, we'll delete it once the PR is merged, so there's no need to re-do it.

docs/source/install/kernel_install.rst
```diff
@@ -3,9 +3,13 @@
 Installing the IPython kernel
 =============================
 
-The Jupyter Notebook and other frontends automatically ensure that the IPython kernel is available.
+The Jupyter Notebook and other frontends automatically ensure that the IPython kernel is available
+for whatever version of Python you used to install the Jupyter Notebook.
```
takluyver: I'd `s/whatever/the/` here - can't really explain why, it just sounds better in my head.

docs/source/install/kernel_install.rst
```diff
@@ -3,9 +3,13 @@
...
+
 However, if you want to use a kernel with a different version of Python, or in a virtualenv or conda environment,
-you'll need to install that manually.
+you'll need to install that manually. Note that this will **overwrite** the existing IPython kernel. If you want
```
takluyver: This is where it gets complicated. If you're setting it up to make the other one of Python 2/3 available to your notebook server, it *won't* overwrite your existing kernel, because it automatically calls them `python3` and `python2`. We've sort of assumed that that's the most common use case, which is why this section is before the explanation of giving them different names below.

takluyver:
I think we should rework this document a bit more extensively to condense the explanation, rather than sticking more words in there. Do you want to have a go at that? I'm happy to try my hand at it, but it might be tomorrow in UK time.

jhamrick:
Sure, you can give it a try. I do think however we word it, it should clearly answer the question of "how do I install python 2 and python 3 kernels at the same time?"

Carreau:
Need rebase (or closing).

takluyver:
#9155 was an alternative to this, and Jess +1'd it, so I think this can be closed.

PR #9155, opened 2016-01-21
Clarify kernel installation instructions

takluyver:
@jhamrick, how does this look?

Alternative to gh-9154

### DIFF
--- a/docs/source/install/kernel_install.rst
+++ b/docs/source/install/kernel_install.rst
@@ -39,7 +39,9 @@ Kernels for different environments
 ----------------------------------
 
 If you want to have multiple IPython kernels for different virtualenvs or conda environments,
-you will need to specify unique names for the kernelspecs:
+you will need to specify unique names for the kernelspecs.
+
+For example, using conda environments:
 
 .. sourcecode:: bash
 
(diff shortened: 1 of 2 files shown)


docs/source/install/kernel_install.rst
```diff
@@ -3,37 +3,51 @@
...
-If you want to have multiple IPython kernels for different environments,
-you will need to specify unique names for the kernelspecs,
-and you may also want to specify the display name of those kernels,
-so that you can clearly see which is which in the notebook menus:
+If you want to have multiple IPython kernels for different virtualenvs or conda environments,
+you will need to specify unique names for the kernelspecs:
 
 .. sourcecode:: bash
 
     source activate myenv
```
jhamrick: I would add a comment above here just making it clear that this example (i.e. the `source activate` command) is specific to anaconda. Otherwise this comes off to someone who doesn't know about conda envs that they need to run e.g. `source activate myenv` to install the kernel.

jhamrick:
:+1:

docs/source/install/kernel_install.rst
takluyver: Done

jhamrick:
Looks good to me!

PR #9156, opened 2016-01-21
add do_where for PDB, update whatsnew tool.

Carreau:


tools/update_whatsnew.py
```diff
@@ -7,6 +7,7 @@
 
 import io
 import os
+from glob import glob
```
Carreau: I use glob, otherwise listdir picks up my `*.swp~` files.

tools/update_whatsnew.py
```diff
@@ -20,15 +21,15 @@
 
 # 1. Collect the whatsnew snippet files ---------------------------------------
 
-files = set(os.listdir(pr_dir))
+files = set()
+for f in glob(pjoin(pr_dir, '*.rst')):
+    files.add(f)
```
takluyver: Why the for loop? Can't it just be:

```python
files = set(glob(pjoin(pr_dir, '*.rst')))
```

tools/update_whatsnew.py
```diff
@@ -20,15 +21,15 @@
...
 # Ignore explanatory and example files
-files.difference_update({'README.md',
+files.difference_update({pjoin(pr_dir, f) for f in {'README.md',
```
takluyver: No need to subtract `README.md` now it's only finding .rst files.

tools/update_whatsnew.py
```diff
@@ -38,7 +39,11 @@ def getmtime(f):
 features, incompats = [], []
 for path in files:
     with io.open(path, encoding='utf-8') as f:
-        content = f.read().rstrip()
+        try:
+            content = f.read().rstrip()
+        except Exception as e:
+            raise Exception('Error reading "{}"'.format(f))
```
takluyver: This adds one bit of information - what file the error was on - but removes another: what error occurred.

Maybe print the filename and then reraise the original error?

tools/update_whatsnew.py
Carreau: I think I tried and this add the list to the set (got `List is unhashable` error).

I can try to unpack `set(*glob(pjoin(pr_dir, '*.rst')))`

docs/source/whatsnew/pr/do_where.rst
```diff
@@ -0,0 +1,47 @@
...
+    ----> 7     raise ValueError('6 is not acceptable')
+        8     7
+        9     8
+
+    ipdb> where 1
+    <ipython-input-3>(1)<module>()
+    ----> 1 foo()
+
+    > <ipython-input-2>(7)foo()
+    ----> 7     raise ValueError('6 is not acceptable')
```
takluyver: I don't think we're going to want to illustrate this at such length in the whatsnew document. I know we can always trim it down later, but if you could write a shorter description now, that will make life easier when we do come to release time.

takluyver:
Did the code change this is talking about already land? From the title I was expecting to see some diff in the debugger.

tools/update_whatsnew.py
takluyver: Not sure what you tried, but `set(glob(...))` works.

Carreau: done.

tools/update_whatsnew.py
Carreau: done.

tools/update_whatsnew.py
Carreau: God point, that was my intention initially. I likely got distracted.

Carreau:
> Did the code change this is talking about already land? From the title I was expecting to see some diff in the debugger.

Yes.  I added the number of the PR (#9097)

docs/source/whatsnew/pr/do_where.rst
```diff
@@ -0,0 +1,3 @@
+IPython debugger (IPdb) now supports the number of context lines for the
+``where`` (and ``w``) commands. The `context` keyword is also available in various APIs.
+See PR #9097
```
minrk: There's a `:ghpull:` roll that generates links to pull requests.

Carreau: Done.

PR #9161, opened 2016-01-22
Add release instructions for IPython.

Carreau:
Mostly adapt from notebooks one.

### DIFF
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -26,3 +26,51 @@ on the IPython GitHub wiki.
    pycompat
    config
    inputhook_app
+
+Making an IPython release
+=========================
+
+Make sure the repository is clean of any file that could be problematic. 
+You can remove all non-tracked files with:
+
+.. code::
+
+    $ git clean -xfdi
+
+This would ask you for confirmation before removing all untracked files. Make
+sure the ``dist/`` folder is clean and avoid stale build from
+previous attempts.
+
+1. Update version number in ``IPython/core/release.py``.
+
+Make sure the version number match pep440, in particular, `rc` and `beta` are
+not separated by `.` or the `sdist` and `bdist` will appear as different
+releases.
+
+2. Commit and tag the release with the current version number:
+
+.. code::
+
+    git commit -am "release $VERSION"
+    git tag $VERSION
+
+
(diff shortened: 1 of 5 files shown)


docs/source/index.rst
```diff
@@ -21,6 +21,7 @@ Contents
    interactive/index
    config/index
    development/index
+   development/release_process
```
takluyver: I don't think this needs a separate entry in the top-level TOC - let's add it to `development/index` instead.

setup.py
```diff
@@ -287,7 +287,14 @@ def run(self):
...
 setup_args.update(setuptools_extra_args)
 
+
+# loose as `.dev` is suppose to be invalid
+loose_pep440re = re.compile('^(\d+)\.(\d+)\.(\d+((a|b|rc)\d+)?)(\.post\d+)?(\.dev\d*)?$')
+
 def main():
+    import IPython.core.release as r
+    if not loose_pep440re.match(r.version):
+        raise ValueError("Version number '%s' is not valid (should match [N!]N(.N)*[{a|b|rc}N][.postN][.devN])" % r.version)
```
takluyver: Do we want to do this check on every use of setup.py, or only when building packages for release?

docs/source/index.rst
Carreau: I will move it.

setup.py
Carreau: Is it worth the added checking only on build ?

takluyver: This is probably fine.

Carreau:
comment addressed and rebased.

Carreau:
WTF ? Why does not travis install traitlets ?

setup.py
Carreau: Change my mind, couldn't extract dependency from `setup.py` if traitlets not installed when you import `IPython.core.release` in `setup.py`

docs/source/development/index.rst
```diff
@@ -26,3 +26,51 @@ on the IPython GitHub wiki.
    pycompat
    config
    inputhook_app
+
+Making an IPython release
+=========================
```
takluyver: I actually meant just adding the new page to the TOC in this file, not putting the content in here, but it's fine.

PR #9164, opened 2016-01-23
Check setuptools version in setup.py

### LINKED ISSUE #9159: Cannot build ipython 4.0.3 with setuptools 18.2
I was just updating [a bunch of Python packages](https://github.com/NixOS/nixpkgs/pull/12552) for Nix, when I get the following error when building ipython 4.0.3:

```
error in ipython setup command: Invalid environment marker: sys_platform == "darwin" and platform_python_implementation == "CPython"
```
Now I'm not familiar with these markers but I can imagine it is related to the pip 8.0.2 update. 

The error occurs with all Python versions.

Carreau:
Allow better error message when trying to build with old version of
setuptools.

Closes #9159

```
$ python setup.py bdist_wheel
  File "setup.py", line 177, in <module>
    raise ValueError('Setuptools version >18.5 is required')
ValueError: Setuptools version >18.5 is required
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
 install:
+    - pip install "setuptools>=18.5"
     - pip install -f travis-wheels/wheelhouse -e file://$PWD#egg=ipython[test]
     - pip install codecov
 script:
--- a/setup.py
+++ b/setup.py
@@ -172,6 +172,9 @@ needs_setuptools = set(('develop', 'release', 'bdist_egg', 'bdist_rpm',
 
 if len(needs_setuptools.intersection(sys.argv)) > 0:
     import setuptools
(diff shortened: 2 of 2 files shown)


FRidh:
While install_requires is not yet evaluated at that point, I think it would be good to at add an explicit install requirement on "setuptools >=18.5" there
https://github.com/ipython/ipython/blob/master/setup.py#L193

And yes, because you actually cannot evaluate setup.py without that version of setuptools such error message would definitely be helpful. Thanks.

Carreau:
> While install_requires is not yet evaluated at that point, I think it would be good to at add an explicit install requirement on "setuptools >=18.5" there
> https://github.com/ipython/ipython/blob/master/setup.py#L193

Good point. Only drawback is that dev-install works with setuptools < 18.5 so we are a bit artificially restraining versions.

minrk:
:+1:

FRidh:
Thanks for getting this in. I'm sure it will prevent some confusion for other packagers and users who would be doing such update. On Nix we've gone ahead with upgrading to the latest setuptools now. (It did break about 400/6000 Nix-packaged Python packages but the upgrade is inevitable.)

PR #9167, opened 2016-01-25
Truncate tracebacks on recursion error

takluyver:
Identify repeated frames in a recursion error traceback and only show them once.

Closes gh-7833

There's a certain amount of guesswork involved in deciding what to elide. I'm guessing that in most cases, the frames will fall into a cycle (e.g. A calls B calls C calls A), so the code tries to work out how long the cycle is. I've tried to err on the side of showing rather than hiding frames, but in some situations it could still obscure meaningful information.

This is what it looks like in the simplest possible case:

```
In [1]: def f():
   ...:     f()
   ...:     

In [2]: f()
<ipython-input-2-0ec059b9bfe1> in <module>()
----> 1 f()

<ipython-input-1-3ce0ca94edec> in f()
      1 def f():
----> 2     f()
      3 

Last 1 frames repeated, from:

<ipython-input-1-3ce0ca94edec> in f()
      1 def f():
----> 2     f()
      3 

RuntimeError: maximum recursion depth exceeded
```

Showing the repeated frame looks superfluous when the function calls itself directly, but if the loop is 10 functions long I think it will be pretty useful.

### DIFF
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -8,7 +8,12 @@ from textwrap import dedent
 import traceback
 import unittest
 
-from ..ultratb import ColorTB, VerboseTB
+try:
+    from unittest import mock
+except ImportError:
+    import mock    # Python 2
+
+from ..ultratb import ColorTB, VerboseTB, find_recursion
 
 
 from IPython.testing import tools as tt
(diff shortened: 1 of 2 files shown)


takluyver:
Milestoned as 4.1 because #7833 already was, but I don't mind if we want to bump this to a later release.

Carreau:
+1 on principle. 

I would like to iterate on the message which his highly unclear to me, especially the `from`:

```
In [1]: def f(x):return g(x)

In [2]: def g(x): return f(x)

In [3]: f(1)
RecursionError                            Traceback (most recent call last)
<ipython-input-3-90b61b657670> in <module>()
----> 1 f(1)

<ipython-input-1-0433736caa3c> in f(x)
----> 1 def f(x):return g(x)

<ipython-input-2-bae8f3d1cac6> in g(x)
----> 1 def g(x): return f(x)

...
(Last 2 frames repeated 500 times):
...

<ipython-input-1-0433736caa3c> in f(x)
----> 1 def f(x):return g(x)

RecursionError: maximum recursion depth exceeded

```


I'm sad also because it will make rebasing https://github.com/ipython/ipython/pull/8966 a huge pain.

takluyver:
'from' is intended as in "if you look back up the list of stack frames above, the repetition starts from this one:"

minrk:
+1 to this. I don't have a strong opinion on the elision message. I do like the presence of an ellipsis indicating where the elision takes place.

takluyver:
The message is now:

```
... last 3 frames repeated, from the frame below ...
```

Carreau:
As you like. The "from" still look weird to me, but anyway.

+1

takluyver:
@Carreau this is the last thing milestoned for 4.1. But I don't mind if we want to bump this to 4.2 or 5.0, as there is some small risk of it causing problems.

:shipit:

Carreau:
let's live slightly more dangerously.

PR #9170, opened 2016-01-25
Multiproject support for the backport tool

jdfreder:
I've always used the backport tool for PR backporting and I figured this would be useful for everyone in the jupyter and ipython orgs.

### DIFF
--- a/tools/backport_pr.py
+++ b/tools/backport_pr.py
@@ -2,20 +2,27 @@
 """
 Backport pull requests to a particular branch.
 
-Usage: backport_pr.py branch [PR] [PR2]
+Usage: backport_pr.py [org/repository] branch [PR] [PR2]
 
 e.g.:
 
     python tools/backport_pr.py 0.13.1 123 155
 
-to backport PR #123 onto branch 0.13.1
+to backport PRs #123 and #155 onto branch 0.13.1
 
 or
 
     python tools/backport_pr.py 2.1
 
 to see what PRs are marked for backport with milestone=2.1 that have yet to be applied
-to branch 2.x.
+to branch 2.x
+
+or
+
+    python tools/backport_pr.py jupyter/notebook 0.13.1 123 155
+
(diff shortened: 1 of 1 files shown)


takluyver:
:+1:

Carreau:
I'm starting to really miss a bot that would do that fro us, and also deal with issues migrations.

[Homu](https://github.com/barosl/homu) is written in Python and is used by Rust. I'm wondering if we could add functionalities like backport PRs and issues migration.

Carreau:
+1 otherwise.

tools/backport_pr.py
```diff
@@ -2,20 +2,27 @@
 """
 Backport pull requests to a particular branch.
 
-Usage: backport_pr.py branch [PR] [PR2]
+Usage: backport_pr.py [org/repository] branch [PR] [PR2]
```
Carreau: I think adding a `--repo` flag would be better. then we can deprecate the default to ipython/ipython.

jdfreder: Why deprecate the default if this tool lives in ipython/ipython?

Carreau: Because it's easy to missused and backport a PR we don't want to backport, and why would it stay in IPython/IPython ?  Plus the `backport_pr.py org/repo/pr` seem much more natural to me as a syntax if you've [set up your git correctly](https://gist.github.com/piscisaureus/3342247)

jdfreder: > syntax if you've set up your git correctly

heh, that's a matter of opinion, I prefer GitHub's "hub" instead of that setup.




> why would it stay in IPython/IPython ?

I'd be happy if it was moved out to Jupyter/backport-pr or something similar

jdfreder: Or maybe a jupyter/utilities or jupyter/tool repo...

Carreau: Fine in IPython for now I guess.

minrk:
+1 to moving this (and gh_api) to a separate package. I've been meaning to do that for a while.

Issue #9172, opened 2016-01-26
display.display_html has no effect in Jupyter notebook


pwaller:
Versions:
* IPython 4.0.3
* Jupyter 1.0.0

Example code:

```python
import IPython.display
IPython.display.display_html("<b>foo bar</b>")
```

What happens when I run this cell? No visible output.

What is expected? It should show the words "foo bar" in bold.

What works? `IPython.display.display(IPython.display.HTML("<b>foo bar</b>"))`.

Am I doing the right thing here? How do I just show an object formatted as HTML? I want to show several objects in one cell execution, so just putting the object as the last thing in the cell isn't sufficient.

[The documentation](https://ipython.org/ipython-doc/dev/api/generated/IPython.display.html#IPython.display.display_html) seems to imply this should work.

minrk:
`display_html` displays the HTML *representations* of an object. That is, it looks up registered display methods, such as `_repr_html_`, and calls them, displaying the result, if any. `str` objects do not have an HTML representation, so nothing is displayed.

If you want to publish *raw HTML* as a string, then what you want is to pass `raw=True`, which seems to be what you are after:

```python
import IPython.display
IPython.display.display_html("<b>foo bar</b>", raw=True)
```

pwaller:
Thanks @minrk for the quick response. Is there scope for me to send a PR to improve the display documents to mention this caveat? It wasn't clear to me that this is what was happening, even after searching around for a good 30 minutes.

pwaller:
Another kind of fix (even better than the docs) would be to show some kind of indication that there was nothing to show.

minrk:
Definitely feel free to open a PR clarifying the docs, that would be very welcome.

takluyver:
Maybe the `display_x` functions should throw an error if the object passed to them doesn't have an x representation? I would have been confused by that.

pwaller:
I would be in favour of that.

PR #9173, opened 2016-01-26
Clarify display_html docs in case of no HTML repr

pwaller:
If there is no HTML representation for the object, nothing is shown.

See also #9172.


pwaller:
/cc @minrk.

I wasn't totally sure about the language. You also get nothing if you do `raw=True` and pass in an incorrect object, but I couldn't trivially figure out what the rules were. For example, `display_html(1, raw=True` also does nothing, but I can't so easily explain why not.

minrk:
Yeah, it shouldn't fail silently for the wrong type in the raw=True case.

PR #9177, opened 2016-01-26
Fix release instructions.

### LINKED ISSUE #9176: [docs] ad --universal to wheel build instructions.

Carreau:
Closes #9176

### DIFF
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -59,8 +59,8 @@ releases.
 
 .. code::
 
-    $ python setup.py sdist --formats=zip,gztar
-    $ python setup.py bdist_wheel
+    python setup.py sdist --formats=zip,gztar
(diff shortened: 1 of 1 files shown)


docs/source/development/index.rst
```diff
@@ -59,8 +59,8 @@ releases.
 
 .. code::
 
-    $ python setup.py sdist --formats=zip,gztar
-    $ python setup.py bdist_wheel
+    python setup.py sdist --formats=zip,gztar
+    python setup.py bdist_wheel --universal
```
takluyver: I think the universal option can be set in `setup.cfg` if that's preferable to specifying it manually. But for now, this is correct.

minrk: We can't set this in IPython if we want to continue to provide the `ipython2` and `ipython3` executables.

PR #9182, opened 2016-01-27
Fix Image when passing invalid data

### LINKED ISSUE #9181: IPython.display.Image crash if passed invalid bytes
Used in notebook test suite for testing not tested in IPython.

```
Image("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQMAAAAl21bKAAAAA1BMVEUAAACnej3aAAAAAWJLR0QAiAUdSAAAAAlwSFlzAAALEwAACxMBAJqcGAAAAAd0SU1FB94BCRQnOqNu0b4AAAAKSURBVAjXY2AAAAACAAHiIbwzAAAAAElFTkSuQmCC")
```

```
<ipython-input-4-871bd9636d0c> in <module>()
----> 1 Image(b"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQMAAAAl21bKAAAAA1BMVEUAAACnej3aAAAAAWJLR0QAiAUdSAAAAAlwSFlzAAALEwAACxMBAJqcGAAAAAd0SU1FB94BCRQnOqNu0b4AAAAKSURBVAjXY2AAAAACAAHiIbwzAAAAAElFTkSuQmCC".encode())

/Users/bussonniermatthias/anaconda3/lib/python3.5/site-packages/IPython/core/display.py in __init__(self, data, url, filename, format, embed, width, height, retina, unconfined, metadata)
    732                     format = self._FMT_JPEG
    733
--> 734         if format.lower() == 'jpg':
    735             # jpg->jpeg
    736             format = self._FMT_JPEG

AttributeError: 'NoneType' object has no attribute 'lower'
```

Probably a bad logic to detect the format if `data` is `bytes`, we likely also need a Python2 special case

Carreau:
Closes #9181

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -12,7 +12,6 @@ import os
 import struct
 import warnings
 
-from IPython.core.formatters import _safe_get_formatter_method
 from IPython.utils.py3compat import (string_types, cast_bytes_py2, cast_unicode,
                                      unicode_type)
 from IPython.testing.skipdoctest import skip_doctest
@@ -731,13 +730,13 @@ class Image(DisplayObject):
                 if data[:2] == _JPEG:
                     format = self._FMT_JPEG
 
-        if format.lower() == 'jpg':
-            # jpg->jpeg
-            format = self._FMT_JPEG
-
         # failed to detect format, default png
         if format is None:
             format = 'png'
+
+        if format.lower() == 'jpg':
+            # jpg->jpeg
+            format = self._FMT_JPEG
 
         self.format = unicode_type(format).lower()
         self.embed = embed if embed is not None else (url is None)
--- a/IPython/core/tests/test_display.py
+++ b/IPython/core/tests/test_display.py
@@ -43,6 +43,9 @@ def test_retina_jpeg():
     nt.assert_equal(md['width'], 1)
     nt.assert_equal(md['height'], 1)
 
+def test_base64image():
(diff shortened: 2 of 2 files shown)


IPython/core/tests/test_display.py
```diff
@@ -43,6 +43,9 @@ def test_retina_jpeg():
     nt.assert_equal(md['width'], 1)
     nt.assert_equal(md['height'], 1)
 
+def test_base64image():
+    display.Image("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQMAAAAl21bKAAAAA1BMVEUAAACnej3aAAAAAWJLR0QAiAUdSAAAAAlwSFlzAAALEwAACxMBAJqcGAAAAAd0SU1FB94BCRQnOqNu0b4AAAAKSURBVAjXY2AAAAACAAHiIbwzAAAAAElFTkSuQmCC")
```
minrk: This is an invalid image, so it is equivalent to `display.Image("garbage")`

IPython/core/display.py
```diff
@@ -731,6 +730,9 @@ def __init__(self, data=None, url=None, filename=None, format=None,
                 if data[:2] == _JPEG:
                     format = self._FMT_JPEG
 
+        if not format:
+            format=self._FMT_JPEG
```
takluyver: This doesn't feel right. If we can't infer the format, I think we should throw an error rather than assuming it's JPEG. "In the face of ambiguity, refuse the temptation to guess" and all that. ;-)

minrk: We're actually already doing this a few lines below, but setting it to png instead. The problem came when the `.lower()` was added on the wrong side of the `None` check, causing the AttributeError.

minrk: The API as documented allows `Image(b'png_bytes')`, so the fix here is to just move the `None` check that defaults to PNG up a few lines. It shouldn't default to jpeg.

IPython/core/tests/test_display.py
Carreau: No it's not invalid, is a black dot used in `jupyter/notebook` tests.

IPython/core/display.py
Carreau: Done.

PR #9184, opened 2016-01-27
Don't build universal wheels

minrk:
We can't provide ipython2/ipython3 executables with universal wheels.


takluyver:
Good point. The 4.1.0rc1 wheel that's currently up will install `ipython3` and `iptest3` scripts, regardless of which Python is used to install it.

PR #9187, opened 2016-01-27
Update 2fa script a bit more + generate 4.1 changelog.

Carreau:
The number of commits /stats will be off by a few, but not sure it is really that important.

### DIFF
--- a/tools/gh_api.py
+++ b/tools/gh_api.py
@@ -17,7 +17,7 @@ import json
 try:
     import requests_cache
 except ImportError:
-    print("no cache", file=sys.stderr)
+    print("cache not available, install `requests_cache` for caching.", file=sys.stderr)
 else:
     requests_cache.install_cache("gh_api", expire_after=3600)
 
--- a/tools/github_stats.py
+++ b/tools/github_stats.py
@@ -212,8 +212,7 @@ if __name__ == "__main__":
 
     print("We closed %d issues and merged %d pull requests." % (n_issues, n_pulls))
     if milestone:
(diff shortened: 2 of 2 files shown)


tools/gh_api.py
```diff
@@ -17,7 +17,7 @@
 try:
     import requests_cache
 except ImportError:
-    print("no cache", file=sys.stderr)
+    print("cache, not available, install request_cache for caching.", file=sys.stderr)
```
takluyver: requests_cache

takluyver: also, no comma after 'cache'

tools/github_stats.py
```diff
@@ -212,8 +212,7 @@ def report(issues, show_urls=False):
 
     print("We closed %d issues and merged %d pull requests." % (n_issues, n_pulls))
     if milestone:
-        print("The full list can be seen `on GitHub <https://github.com/%s/milestone/%s>`__"
-            % (project, milestone)
+        print("The full list can be seen `on GitHub <https://github.com/{project}/issues?q=milestone%3A{milestone}+>`__".format(project=project,milestone=milestone)
```
takluyver: Getting ready for f strings? ;-)

Carreau: :-P

tools/gh_api.py
Carreau: fixed.

takluyver:
The code changes look fine to me, though I kind of think that since we're doing stats, they should be created at release time so that they're accurate.

Carreau:
> The code changes look fine to me, though I kind of think that since we're doing stats, they should be created at release time so that they're accurate.

fixed, rebased.

PR #9188, opened 2016-01-27
display: Workaround handle case where `format` is `None`.

### LINKED ISSUE #1015: CI: js/notebook test failures
Appears that something on master is causing the `js/notebook` group to fail on tests. Points to this line in iPython ( https://github.com/ipython/ipython/blob/8a1aee298716dd71165eb66f7d297ec4b5b151a3/IPython/core/display.py#L734 ) having a format that is `None`. Should this be happening? Is it something that is being done wrong here or in iPython?

jakirkham:
Fixes https://github.com/jupyter/notebook/issues/1015

If `format` is `None` and `ext` is None and `data` is not of type `bytes`, then `format` could still be `None` at this line. Trying to change `None` to lowercase is not permissible. This should fix that.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -731,7 +731,7 @@ class Image(DisplayObject):
                 if data[:2] == _JPEG:
                     format = self._FMT_JPEG
 
-        if format.lower() == 'jpg':
+        if format and format.lower() == 'jpg':
             # jpg->jpeg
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks for working on this, but PR #9182 was already in the works to fix it.

jakirkham:
Oh, sorry, must have missed it.

PR #9191, opened 2016-01-28
Update release script not to upload insecurly

Carreau:
All in title.

### DIFF
--- a/tools/release
+++ b/tools/release
@@ -50,10 +50,8 @@ if 'upload' not in sys.argv:
     print("`./release upload` to register and release")
     sys.exit(0)
 
-# Register with the Python Package Index (PyPI)
-print( 'Not Registering with PyPI, registering with setup.py is insecure as communication is not encrypted')
+# Not Registering with PyPI, registering with setup.py is insecure as communication is not encrypted
 cd(ipdir)
-# sh('./setup.py register')
 
 # Upload all files
 sh(sdists)
--- a/tools/release
+++ b/tools/release
@@ -51,17 +51,19 @@ if 'upload' not in sys.argv:
     sys.exit(0)
 
 # Register with the Python Package Index (PyPI)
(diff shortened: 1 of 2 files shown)


tools/release
```diff
@@ -51,17 +51,19 @@ if 'upload' not in sys.argv:
...
+# sh('./setup.py register')
 
 # Upload all files
-sh(sdists + ' upload')
+sh(sdists)
 for py in ('2.7', '3.4'):
-    sh('python%s setupegg.py bdist_wheel upload' % py)
+    sh('python%s setupegg.py bdist_wheel' % py)
+
+print('Will not upload with setuptools as upload connection is insecure. Please use `twine upload dist/*` to upload the files to PyPI')
```
takluyver: Why not call twine in the script?

tools/release
```diff
@@ -51,17 +51,19 @@ if 'upload' not in sys.argv:
     sys.exit(0)
 
 # Register with the Python Package Index (PyPI)
-print( 'Registering with PyPI...')
+print( 'Not Registering with PyPI, registering with setup.py is insecure as communication is not encrypted')
```
takluyver: I don't think we need to print a message saying that we're not doing this. A comment is fine. And the call can be removed rather than just commented.

tools/release
Carreau: Sure in the long run, but we have a people visiting today in 1/2 an hour, and I was going to make that in a separate PR when I have time unless So does it in the meantime. (I guess you can use twine as a library). 

Also I'm sure there will be much bikeshed so separate seem better to iterate quickly.

tools/release
Carreau: Done.

tools/release
takluyver: I was thinking of just putting `sh('twine upload dist/*')` in there. No need to work out how to use it as a library. ;-)

PR #9192, opened 2016-01-28
Release tools 2

Carreau:
A few more update to the build process:

Remove not relevant classifier, add (a few)  missing classifier.
Remove download url (pointing to GitHub releases).

Do not try to compile fiels in hidden folders like `.git`

### DIFF
--- a/IPython/core/release.py
+++ b/IPython/core/release.py
@@ -105,12 +105,10 @@ author_email = 'ipython-dev@scipy.org'
 
 url = 'http://ipython.org'
 
-download_url = 'https://github.com/ipython/ipython/downloads'
 
-platforms = ['Linux','Mac OSX','Windows XP/Vista/7/8']
+platforms = ['Linux','Mac OSX','Windows']
 
-keywords = ['Interactive','Interpreter','Shell','Parallel','Distributed',
-            'Web-based computing', 'Qt console', 'Embedding']
+keywords = ['Interactive','Interpreter','Shell', 'Embedding']
 
 classifiers = [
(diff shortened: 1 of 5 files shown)


IPython/core/release.py
```diff
@@ -121,5 +119,8 @@
     'Programming Language :: Python :: 2',
     'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.3',
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
```
takluyver: I think we've discussed this before, but I don't think we should use the specific Python 3.x classifiers. There's no particular reason it wouldn't be compatible with 3.6, or later versions, so this can never be complete. And as far as I know, no tools actually look at the more specific classifiers anyway. I think we should just use the more general `Python :: 3` classifier, and describe in prose that 3.3 or above is required (i.e. as we already do).

IPython/core/release.py
```diff
@@ -105,12 +105,10 @@
 
 url = 'http://ipython.org'
 
-download_url = 'https://github.com/ipython/ipython/downloads'
 
-platforms = ['Linux','Mac OSX','Windows XP/Vista/7/8']
+platforms = ['Linux','Mac OSX','Windows XP/Vista/7/8/10']
```
takluyver: Likewise for this. Let's just say 'Windows'.

minrk:
:+1: after @takluyver's changes

Carreau:
Done, and rebased.

Issue #9180, opened 2016-01-26
investigate tools/build_release


Carreau:
there seem to be a partially automated script to do that. Should we resurect it and document it ?

minrk:
Its use [is documented](https://github.com/ipython/ipython/wiki/Dev:-Releasing-IPython), and it has been used to build every release of IPython, as far as I know. But I don't doubt that it could use an update.

Carreau:
@willingc do you want to take a shot at migrating this from wiki to official docs ?

willingc:
@Carreau Sure no problem :D

Carreau:
release process says:

```
Before generating the GitHub stats, verify that all closed issues and pull requests have appropriate milestones. **This search** should return no results before creating the GitHub stats.
```

The `this search` part can **likely** be automatized.

Carreau:
`python tools/github_stats.py --milestone $MILESTONE > stats.rst` docs:  should add the `--since` or it get confused by `rc`s.

Carreau:
Instructions still have `jsversion`

Carreau:
instruction don't push --tags

PR #9193, opened 2016-01-28
Migrate release creation instructions from wiki to docs

willingc:


willingc:
Addresses #9180 

I've migrated the wiki contents to the docs and split the developer guide into two guides (3rd party development and core IPython development).  cc/@carreau, @minrk, @takluyver

Carreau:
I think `docs/source/coredev/making_release.rst` and `docs/source/coredev/release_process.rst` shoudl be merged and made the same document. 
I'll send you a PR that merge the two together.

Carreau:
various question:

- TAG: do we use `TAG =$VERSION` or `TAG==rel-$VERSION`
- is `6. Get a fresh clone`  necessary if we have `git -xfd`

Carreau:
The wiki is **way** out of dates in many places.

PR #9194, opened 2016-01-29
Improvement on [Migrate release creation instructions from wiki to docs #9193]

Carreau:
I sadly cannot make a PR against @wilingc repo, for whatever reason. 

This is just one extra commit that merge "making a release" and "release process".


willingc:
@carreau Thanks.

Issue #9196, opened 2016-01-29
installation error


anutammewar:
Downloading https://pypi.python.org/packages/source/i/ipython/ipython-4.1.0rc1.zip#md5=a9ff233f176dd99b076b81dc8904ab7a
Processing ipython-4.1.0rc1.zip
Running ipython-4.1.0rc1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-NniNKc/ipython-4.1.0rc1/egg-dist-tmp-pW2Mwb
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/2.7/bin/easy_install", line 8, in <module>
    load_entry_point('setuptools==0.6c11', 'console_scripts', 'easy_install')()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 1712, in main
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 1700, in with_ei_usage
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 1716, in <lambda>
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/core.py", line 152, in setup
    dist.run_commands()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 211, in run
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 446, in easy_install
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 476, in install_item
[...]

PR #9197, opened 2016-01-29
Remove setuptools version check

takluyver:
Fixes gh-9196

Checking version numbers isn't that easy.


setup.py
```diff
@@ -172,9 +172,6 @@ def run(self):
 
 if len(needs_setuptools.intersection(sys.argv)) > 0:
     import setuptools
-    v = tuple(int(x) for x in setuptools.__version__.split('.'))
```
Carreau: I guess `....split('.')[:2]` would have been enough, but ok.

Carreau:
+1

PR #9200, opened 2016-01-29
Factor build process (and gen what's new)

Carreau:
all in title.


Carreau:
I'm going to merge, and release a rc2, to test  instructions and to go home early today. 
We can finalize next month when we release the actual final version.

PR #9202, opened 2016-02-01
compare shell tokens with BashLexer output

### LINKED ISSUE #9201: IPythonLexer test failure with pygments 2.1
Jenkins is showing us a test failure on Windows:

https://jenkins.jupyter.org/job/windows-multi/_pyversion=3.5,label=windows/504/testReport/

It looks like `$HOME` is unexpectedly being split into two tokens. I suspect it's failing since pygments 2.1 was released. I see the same locally on Linux, but it's not failing on Travis (maybe we're not installing pygments so those tests are skipped?).

@abalkin and @chebee7i have done the most signficant work on the lexer machinery. Any ideas?

minrk:
rather than assuming BashLexer's output will never change

closes #9201

### DIFF
--- a/IPython/lib/tests/test_lexers.py
+++ b/IPython/lib/tests/test_lexers.py
@@ -5,6 +5,7 @@
 
 from unittest import TestCase
 from pygments.token import Token
+from pygments.lexers import BashLexer
 
 from .. import lexers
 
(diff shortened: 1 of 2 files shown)


takluyver:
That seems reasonable.

PR #9203, opened 2016-02-02
Latextools: Make latex_to_png_mpl not fail on errors

ccordoba12:
- Also make make both `latex_to_foo` functions to return None on failure.
- This will help to solve the errors qtconsole is having to handle Latex (as reported in jupyter/qtconsole#56)

### DIFF
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -97,9 +97,10 @@ def latex_to_png(s, encode=False, backend=None, wrap=False):
 def latex_to_png_mpl(s, wrap):
     try:
         from matplotlib import mathtext
+        from pyparsing import ParseFatalException
     except ImportError:
         return None
-    
+
     # mpl mathtext doesn't support display math, force inline
     s = s.replace('$$', '$')
     if wrap:
@@ -110,7 +111,7 @@ def latex_to_png_mpl(s, wrap):
         f = BytesIO()
         mt.to_png(f, s, fontsize=12)
         return f.getvalue()
-    except:
+    except (ValueError, RuntimeError, ParseFatalException):
         return None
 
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
Not such a big change, but probably too late to be in 4.1 :-)

minrk:
Seems like it would be preferable for the caller to catch the error, no? I'm not sure silent failure is the right thing to do at this level.

ccordoba12:
Well, I think this helps to simplify things a lot because:

1. These functions already return None when Matplotlib or Latex can't be found. So instead of checking only for that case in qtconsole, we would also need to catch errors there.
2. Possible sources of errors are:
    1. Matplotlib is unable to handle complex Latex. However, that's no easily discernible when reporting that error to users, so I don't see much of a point doing that.
    2. There's a syntax error in Latex. That will generate one of those huge Latex report errors, but again I don't see much of a value reporting that error. I think people would assume that if no representation is generated, that's because there's an error in their Latex.
3. Sympy uses these functions independently to generate latex-png representations for their objects. This was generating a traceback in qtconsole even after jupyter/qtconsole#63 because the error had to be catched in Sympy besides doing it in qtconsole. If other projects also do the same, then all of them would need to catch errors generated by `latex_to_foo` and jupyter/qtconsole#63 is kind of pointless. This PR simply address that possibility, preventing at the same time to show ugly tracebacks in qtconsole.

minrk:
In that case, it should catch *at most* the errors caused by unhandled latex. It certainly shouldn't suppress *all* errors, which is guaranteed to hide real bugs.

ccordoba12:
Ok, I agree :-) I'll do the proper changes.

ccordoba12:
@minrk, I addressed your review, so I think this is ready to me merged.

PR #9211, opened 2016-02-03
Release process tweaks

takluyver:
Things I ran into while releasing 4.1.1.

The only one that might be controversial is removing the `python -m compileall` check before we build packages. Because this is hardcoded to `python`, it runs on Python 2 on my system, and fails on one of our tools scripts that happens to be Python 3 only. We could make this smarter, but I don't think it's worth it. We now have continuous integration and editors with built in static analysis, so if we make a file invalid Python syntax we'll know about it well before release.

I also ran into a problem in that I don't have access to archive.ipython.org (I'm sure I did on one of my other computers, but I forget which). There should probably be a check for that somewhere earlier in the build process, but I haven't included it here.

### DIFF
--- a/docs/source/coredev/release_process.rst
+++ b/docs/source/coredev/release_process.rst
@@ -102,7 +102,7 @@ release is: ``1.3rc1``. Notice that there is no separator between the '3' and
 the 'r'.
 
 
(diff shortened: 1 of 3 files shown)


minrk:
Makes sense to me.

PR #9214, opened 2016-02-04
Nested svg

### LINKED ISSUE #8780: IPython notebook: unable to display nested SVG elements
I am unable to properly display nested SVG images in an IPython/Jupyter notebook.

Sample code:

    from IPython.display import SVG
    nested_svg="""
    <svg xmlns="http://www.w3.org/2000/svg">
      <svg x="0">
        <rect x="10" y="10" height="80" width="80" style="fill: #0000ff"/>
      </svg>
      <svg x="100">
        <rect x="10" y="10" height="80" width="80" style="fill: #00cc00"/>
      </svg>
    </svg>"""
    SVG(nested_svg)

The image gets displayed, but it is also transformed in the process. The nested svg elements are all moved to the top level:

    <div class="output_subarea output_svg output_result"><div>
      <svg xmlns="http://www.w3.org/2000/svg" style="width: 100%; height: 100%;">
      </svg>
      <svg x="0" style="width: 100%; height: 100%;">
        <rect height="80" style="fill: #0000ff" width="80" x="10" y="10"></rect>
      </svg>
      <svg x="100" style="width: 100%; height: 100%;">
        <rect height="80" style="fill: #00cc00" width="80" x="10" y="10"></rect>
      </svg>
    </div></div>

There is also an added attribute `style="width: 100%; height: 100%;"` in all three svg elements.

I have also tried an alternative method:

    from IPython.display import display_svg
    display_svg(nested_svg, raw=True)

And I have tried to specify height and width of the root svg element. Neither helped.

However, I noticed that in the .ipynb files the SVG is untransformed in the "outputs" sections.

The problem occurs in IPython/Jupyter version 4.0, same as in 3.2. Using Python 2.7.10 Anaconda 2.3.0 (32-bit). I was able to reproduce the problem at try.jupyter.org using Python 3.4.3.

parleur:
Solve the issue "IPython notebook: unable to display nested SVG elements #8780" (FIX #8780 )

I know it is in a hacky way... The real problem come from the way SVG is sanitized in ipython, I am sure now after beeing tracking this bug throught notebook, ipykernel and eventally return in ipython :-).

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -492,12 +492,13 @@ class SVG(DisplayObject):
         x = minidom.parseString(svg)
         # get svg tag (should be 1)
         found_svg = x.getElementsByTagName('svg')
-        if found_svg:
-            svg = found_svg[0].toxml()
-        else:
-            # fallback on the input, trust the user
-            # but this is probably an error.
-            pass
+        # if found_svg:
+        #     svg = found_svg[0].toxml()
+        # else:
+        #     # fallback on the input, trust the user
+        #     # but this is probably an error.
+        #     pass
+        svg = found_svg[0].toxml()
         svg = cast_unicode(svg)
         self._data = svg
 
@@ -976,4 +977,3 @@ def set_matplotlib_close(close=True):
     from ipykernel.pylab.config import InlineBackend
     cfg = InlineBackend.instance()
     cfg.close_figures = close
-
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -492,17 +492,20 @@ class SVG(DisplayObject):
         x = minidom.parseString(svg)
         # get svg tag (should be 1)
         found_svg = x.getElementsByTagName('svg')
-        # if found_svg:
-        #     svg = found_svg[0].toxml()
-        # else:
-        #     # fallback on the input, trust the user
-        #     # but this is probably an error.
-        #     pass
+        if found_svg:
+            svg = found_svg[0].toxml()
+        else:
+            # fallback on the input, trust the user
+            # but this is probably an error.
+            pass
         svg = found_svg[0].toxml()
         svg = cast_unicode(svg)
         self._data = svg
 
     def _repr_svg_(self):
+        return self.data
+        
+    def _repr_html_(self):
         return self.data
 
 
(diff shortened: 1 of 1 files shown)


parleur:
from the notebook issue #982
> I am not sure this is a notebook error as the following code produces the desired result ...
```.python
from IPython.display import HTML
nested_svg="""
<svg xmlns="http://www.w3.org/2000/svg">
  <svg x="0">
    <rect x="10" y="10" height="80" width="80" style="fill: #0000ff"/>
  </svg>
  <svg x="100">
    <rect x="10" y="10" height="80" width="80" style="fill: #00cc00"/>
  </svg>
</svg>"""
HTML(nested_svg)
```
But still does not if you:
```.python
SVG(nested_svg)
```

takluyver:
It's interesting to know that that works, but I suspect we want a fix in the notebook code, so that it works for any kernel displaying SVG.

parleur:
Hi Thomas, thanks for reactiveness,
I agree with the principle. 
To be more explicit,  right now I am not sure the kernel is not wrong when he send the svg (not sure of the contrary). 
I have searched a long time in the notebook code where it actually modify the data receive before displaying and I just can confirm it is very transparent relatively with data ( and so svg data ) as it is suppose to be. 
So by elimination, I searched what can be wrong in kernel side and found this fix.

Some questions:
- are we sure this is a notebook pb and why ?
- anybody knows if and where in the notebook the receive data are modified ? (as far as I understand these data are trusted and language agnostic so I don't figure out why the notebook would annoy them ...)

I 'll have a eye on it tomorow.
Thanks.

parleur:
I catch some response, thanks @Carreau. 

I was wrong, the problem was in the notebook ... in notebook/js/outputarea.js line 654 ... The notebook un-nest the svg for presentation purpose assuming there is one svg element. 

So I close the PR.

Carreau:
From discussion, Pierre was confused by [`__html__`](https://github.com/parleur/ipython/blob/nested_svg/IPython/core/display.py#L446-L452), 

```
        """
        This method exists to inform other HTML-using modules (e.g. Markupsafe,
        htmltag, etc) that this object is HTML and does not need things like
        special characters (<>&) escaped.
        """
```

The patch should be a couple of line (that you'll write tomorrow), nice catch and team work !

PR #9225, opened 2016-02-12
only exit on error if `interact` is False

### LINKED ISSUE #9223: interactive -i mode doesn't work after exceptions since 4.1.1
Hi,

I recently noticed that after upgrading from 4.0.3 to 4.1.1 (from anaconda), `-i` flag no longer drops to the interactive shell but quits after an exception. I downgraded to 4.0.3 and it is now working correctly. I searched on google but couldn't find a similar issue. Any ideas?

minrk:
fixes `ipython -i failing-script.py`
closes #9223

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -198,6 +198,8 @@ class InteractiveShellApp(Configurable):
     )
     shell = Instance('IPython.core.interactiveshell.InteractiveShellABC',
                      allow_none=True)
+    # whether interact-loop should start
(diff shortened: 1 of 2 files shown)


minrk:
Since this is a regression in 4.1, we might make a 4.1.2 for this, depending on plans for 4.2.

minrk:
For reference, the regression was introduced by #8908

takluyver:
Milestoned as 4.2 for now, since that's the next release we're currently planning.

PR #9229, opened 2016-02-14
Remove gui_reference and page_guiref because they were moved to qtconsole

ccordoba12:
This is the accompanying pull request to jupyter/qtconsole#98


Carreau:
+1. 

@takluyver do you want to keep & adapt some of that for the prompt-toolkit branch ?

takluyver:
Nah, I suspect there's enough differences that it's easier to describe that independently.

PR #9232, opened 2016-02-16
Get more information if issue setting color scheme

Carreau:
Mainly for debugging purpose, use for rebase/rework of pr 8966

### DIFF
--- a/IPython/utils/coloransi.py
+++ b/IPython/utils/coloransi.py
@@ -12,6 +12,8 @@
 __all__ = ['TermColors','InputTermColors','ColorScheme','ColorSchemeTable']
 
 import os
+import sys
+import traceback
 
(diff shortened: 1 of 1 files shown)


takluyver:
Doesn't exception chaining show you the earlier traceback automatically?

Carreau:
> Doesn't exception chaining show you the earlier traceback automatically?

Is that a question to ask if I'm ok dropping Python 2 ? (you know the answer), Or is there a simple way to re-raise in Legacy Python that I don't remember ?

takluyver:
Generally I'd put debugging code in temporarily and not make a pull request with it. Especially so when it's only needed on Legacy Python. So I'm a bit confused about what this PR is for!

Carreau:
Ok, I'll close that and only use it locally.

PR #9238, opened 2016-02-17
Don't try to pass shell up kwargs from Magics class

takluyver:
See ipython/traitlets#176 - this is causing test failures.

### DIFF
--- a/IPython/core/magic.py
+++ b/IPython/core/magic.py
@@ -520,7 +520,6 @@ class Magics(Configurable):
                 shell.configurables.append(self)
             if hasattr(shell, 'config'):
                 kwargs.setdefault('parent', shell)
-            kwargs['shell'] = shell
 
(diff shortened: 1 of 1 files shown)


minrk:
This was a silent bug, so I think this points to the change upstream being a good idea. I'll make it warn instead of raise for a cycle upstream, though.

IPython/core/magic.py
```diff
@@ -516,11 +516,11 @@ def __init__(self, shell=None, **kwargs):
             raise ValueError('Magics subclass without registration - '
                              'did you forget to apply @magics_class?')
         if shell is not None:
+            self.shell = shell
```
minrk: this is redundant - `self.shell = shell` is already on L525

takluyver: Duh. Fixed in an amended commit.

PR #9246, opened 2016-02-18
Color All the things !

Carreau:
Still al lot of todo in the code and possibly regresions.

In particular I havent' resolved #9244 conflicts, so the corresponding issue will reappear (for now) if we merge that. Though the rebaseing is **extremely** painful.

There are still a few yield from here and there, but the basic is working. 

I'm considering making that a shared branch.

(issuing as another PR as it squashes all the commits from #8966) 

Ping @minrk and @takluyver

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -237,7 +237,7 @@ def _file_lines(fname):
 class Pdb(OldPdb):
     """Modified Pdb class, does not load readline."""
 
-    def __init__(self, color_scheme='NoColor', completekey=None,
+    def __init__(self,color_scheme='NoColor',completekey=None,
                  stdin=None, stdout=None, context=5):
 
         # Parent constructor:
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -33,9 +33,10 @@ import sys
 
 from IPython import get_ipython
 from IPython.utils import PyColorize, ulinecache
-from IPython.utils import coloransi, io, py3compat
-from IPython.core.excolors import exception_colors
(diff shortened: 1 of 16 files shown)


takluyver:
I know you've spent a fair bit of time on this, but I think it's just far too big a change to make and review all in one go. Sorry! I think we need to break it down into a sequence of smaller steps:

* Add pygments terminal-colouring machinery
* One by one, convert the pieces that produce coloured output to the new machinery (oinspect, ultratb, debugger...)
* When there's nothing left using it, remove the old colouring machinery (coloransi, PyColorize, etc.)

Wherever possible, we should always approach significant transitions like this. I'm doing something similar with prompt_toolkit: add the new code, switch over to it, remove the old. PRs like this are just too much to handle in one go.

So even though I feel mean saying this, I think we should close this PR and pull the changes apart into a set of smaller PRs. I'm happy to help with that process if you want.

Carreau:
Technically I tried to decouple the things, but unfortunately it's too much coupled. You end up with some crazy workaround if you want to move only part of the codebase. 

That's what I did in #9232, #9233 , #9234 , and' i'm still trying to reduce that to a minimum, but it's a rabbit hole.

Carreau:
@takluyver, unless we go the route of doing that through many PR in which IPython will be in inconsitant state where some part are either 
1) not used. 
2) not configurable / temporarily with less feature.

Carreau:
Note `IPython/core/excolors.py` is now deprecated and unused so you can add `-150 Lines` to the actual changes of this PR.

PR #9250, opened 2016-02-18
Introduce some classes necessary for Pygments refactor

Carreau:
Might make the further review simpler. 

This basically introduce "Colorable". 
It's basically the class from which all things that can have a **default** theme inherit. 
So that when you start IPython with `--color=foo`, you have consistent theme.

### DIFF
--- a/IPython/utils/PyColorize.py
+++ b/IPython/utils/PyColorize.py
@@ -54,6 +54,8 @@ except AttributeError:
 from IPython.utils.coloransi import TermColors, InputTermColors ,ColorScheme, ColorSchemeTable
 from IPython.utils.py3compat import PY3
 
+from .colorable import Colorable
+
 if PY3:
     from io import StringIO
 else:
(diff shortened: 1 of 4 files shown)


takluyver:
Thank you, this is a much easier change to look at! :cookie:

IPython/utils/PyColorize.py
```diff
@@ -59,6 +59,21 @@
...
+import pygments.styles
+
+
+from traitlets.config import Configurable
+from traitlets import Unicode, Bool
+
+
+available_themes = lambda : [s for s in pygments.styles.get_all_styles()]+['NoColor','LightBG','Linux']
+
+class Colorable(Configurable):
```
takluyver: Let's put this in a new module, rather than reusing PyColorize - PyColorize is specifically about colouring Python source code, whereas this is general machinery for colour output.

Carreau: > Let's put this in a new module, rather than reusing PyColorize - PyColorize is specifically about colouring Python source code, whereas this is general machinery for colour output.

Do you really want a module just for 15 lines of code ?

takluyver: If there's no existing module that makes sense for it (this one doesn't as far as I can see), then yes. But also, I guess there will be some more machinery that can be added to it later.

Carreau:
Moved, and rebased.

takluyver:
Thanks, I think this looks OK.

Carreau:
> Thanks, I think this looks OK.

Ok, I'm going to merge, and rebase/rework my other PR, and submit other smaller patches, is that ok?

takluyver:
Smaller patches sounds like a good plan :smiley:

PR #9255, opened 2016-02-20
4.0.x

tdonovan:


takluyver:
Hi @tdonovan - it seems you're a bit confused about pull requests. Can we help to clarify anything?

PR #9256, opened 2016-02-21
Ensure pager interprets escape sequences

x1o:
If less is specified in $PAGER, but $LESS either doesn't exist or doesn't contain `-r` option, append `-r` to the pager command string.

### DIFF
--- a/IPython/core/page.py
+++ b/IPython/core/page.py
@@ -308,6 +308,13 @@ def get_pager_cmd(pager_cmd=None):
             pager_cmd = os.environ['PAGER']
         except:
             pager_cmd = default_pager_cmd
+    
+    if pager_cmd == 'less':
+    try:
+        os.getenv('LESS').index('-r')
+    except (ValueError, AttributeError):
+        pager_cmd += ' -r'
+    
     return pager_cmd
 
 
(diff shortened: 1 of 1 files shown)


IPython/core/page.py
```diff
@@ -308,6 +308,13 @@ def get_pager_cmd(pager_cmd=None):
             pager_cmd = os.environ['PAGER']
         except:
             pager_cmd = default_pager_cmd
+    
+    if pager_cmd == 'less':
+        try:
+            os.getenv('LESS').index('-r')
```
minrk: LESS can None, and `in` can be used to avoid any errors, so this can be:

```python
if ' -r' not in os.env.get('LESS', ''):
    ...
```

x1o: If it is None, the the AttributeError exception is thrown, which is caught below.

minrk: Catching AttributeError often hides bugs (e.g. if this had a typo `os.geteenv`, it would happily behave as if `-r` is not in LESS`), so it's useful to avoid catching common errors like that which can cause silent hiding of bugs later on.

x1o: Ok, I guess you are right.

takluyver:
Thanks!

PR #9261, opened 2016-02-22
Rename PTInteractiveShell to TerminalInteractiveShell

takluyver:
This means that existing config will get applied, even if not all of it has the desired effect. I haven't thought of any adverse effects of renaming this.

Ping @asmeurer, who requested this.

### DIFF
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -32,7 +32,7 @@ from IPython.core.shellapp import (
     InteractiveShellApp, shell_flags, shell_aliases
 )
 from IPython.extensions.storemagic import StoreMagics
-from .ptshell import PTInteractiveShell as TerminalInteractiveShell
+from .ptshell import TerminalInteractiveShell
 from IPython.utils import warn
 from IPython.paths import get_ipython_dir
 from traitlets import (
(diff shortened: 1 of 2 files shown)


Carreau:
is there some config option of **current** TerminalInteractive shell that would have no effects ?

asmeurer:
I'm guessing the readline stuff.

Carreau:
Ok, let me rephrase:

Is there some config option of current TerminalInteractive shell that would have no effects, 
and for which we should  display a warning that the config option has been deprecated. 
Ie something that "just"  a renaming would break.

takluyver:
Most of the options that are defined on the old TIS class won't work currently, though some of them can be introduced to the new one. Some are also redefined over the parent class (e.g. `confirm_exit`), and these will still work.

But you can also set config options `TerminalInteractiveShell.foo` that are defined as `InteractiveShell.foo`, and there's quite a lot of config options on InteractiveShell.

There are also some options on other classes that are broken by this, e.g. `PromptManager.in_template`.

Carreau:
Ok, fair enough. let see how it will break in the wild :-P

asmeurer:
Perhaps for compatibility purposes any options that are set that don't work with prompt toolkit should raise a user warning, which points to some updated configuration docs.

asmeurer:
I'm not sure if this had any effect. In master, IPython still asks me to confirm if I want to exit.

PR #9262, opened 2016-02-22
Color all the things, partial patch 2

Carreau:
Minimal changes. 

This **does** introduce minimal inconsistency as the color themes are not yet available.

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 import io
 import sys
 from pprint import pformat
+from textwrap import wrap
 
 from IPython.core import magic_arguments, page
 from IPython.core.error import UsageError
@@ -17,6 +18,7 @@ from IPython.utils.path import unquote_filename
 from IPython.utils.py3compat import unicode_type
 from warnings import warn
 from logging import error
+from IPython.utils.colorable import available_themes
 
 
 class MagicsDisplay(object):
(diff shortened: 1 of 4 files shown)


IPython/core/magics/basic.py
```diff
@@ -75,6 +77,9 @@ class BasicMagics(Magics):
     These are various magics that don't fit into specific categories but that
     are all part of the base 'IPython experience'."""
 
+    def __init__(self, *args, **kwargs):
```
minrk: This super does nothing, right?

minrk:
Changes here make sense, I just want to make sure I understand the goal of this partial PR:

I see pygments themes as options for `--colors` and in help output, but when I try to load one, I get:

```python
In [1]: %colors monokai
.../IPython/core/magics/basic.py:334: UserWarning: Error changing exception color schemes.
Unrecognized color scheme: monokai
Valid schemes: ['LightBG', 'NoColor', 'Linux']
```

Is there a missing `available_themes()` call somewhere?

Carreau:
@minrk, no it's not "missing" it's just refactoring pieces by pieces. 

> This does introduce minimal inconsistency as the color themes are not yet available.

It's basically just part of https://github.com/ipython/ipython/pull/9246 , but if I want to make a fully self-contained PR  it will be hundred of lines. To @takluyver request I'm submitting smaller changes that are reviewable, though they might be not totally functional until the end.

minrk:
@Carreau thanks for the clarification. I just wanted to know what things I should expect to work and what I shouldn't, since I can't really review it, otherwise. It would be helpful if you described a bit what you aim to accomplish in these incremental PRs.

One comment inline and a merge conflict, but otherwise :+1:.

PR #9265, opened 2016-02-23
handle raw video data as input to Video

minrk:
Follow-up on #8326

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -824,7 +824,7 @@ class Image(DisplayObject):
 
 class Video(DisplayObject):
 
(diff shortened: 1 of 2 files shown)


takluyver:
Looking at the code already in `Video`, I thought we were going to make it harder than that to embed video data, because that's going to make the notebook unwieldy really fast.

minrk:
This just fixes bugs in the existing implementation. If we want to disable passing data directly, that would be fine, too.

takluyver:
I think the compromise we had agreed was that embedding would require passing `embed=True` explicitly, so that it has to be at least somewhat deliberate.

minrk:
That makes sense. I'll make sure it actually behaves that way (it certainly didn't before).

PR #9269, opened 2016-02-23
Re-enable continuation prompts

takluyver:
Now that prompt_toolkit 0.58 is out, this should work.


Carreau:
Want to make the continuation token `...` again ?

Carreau:
+1 otherwise. And actually merging. I'll make the PR myself to revert to ...

PR #9270, opened 2016-02-24
'Restore `...:` as continuation prompt '

Carreau:
I think it is more visually appealing.


takluyver:
For the record: my version was deliberate, because I find the `...` *less* visually appealing.

Carreau:
> For the record: my version was deliberate, because I find the ... less visually appealing.

Feel free to open an issue where we can discuss that, but I think there is too much historical baggage of IPython having `...:` as continuation prompt.

takluyver:
I don't especially mind, I just wanted to record that it wasn't simply an oversight that the `...` was missing.

Issue #9263, opened 2016-02-22
`clear` alias gone


Carreau:
```
ipython
Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:24:55)
Type "copyright", "credits" or "license" for more information.

IPython 5.0.0.dev -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: clear
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-1-05cac50f47bd> in <module>()
----> 1 clear

NameError: name 'clear' is not defined

In [2]:
```

@takluyver ?

takluyver:
We're probably not loading `TerminalMagics`. If so, paste and cpaste will be missing too.

takluyver:
#9273

PR #9273, opened 2016-02-24
Restore terminal magics and aliases

takluyver:
Closes gh-9263


minrk:
We're seeing intermittent errors in test_nest_embed that I assume are prompt-toolkit related. Does it make sense to do the pexpect-style tests with prompt-toolkit, since it's no longer such a simple input/output loop as it is with readline?

takluyver:
I've added a super simple single line prompt that can be switched in for these kinds of tests - of course they can't fully test the interactive behaviour, but they can test stuff like this. I'll make this test use it.

PR #9276, opened 2016-02-24
Use simplified prompt in test_embed

takluyver:
This should avoid the intermittent failures we saw. Ping @minrk 

I also reworked how the tests were dealing with subprocesses a bit.


Carreau:
+1

PR #9285, opened 2016-03-01
remove setuptools workaround for platform_python_implementation

### LINKED ISSUE #9281: ipython setup darwin/CPython error
Hi on some alternative project, ipython is failing to build on a linux system:

<pre>
Searching for ipython>=3.0.0
Reading https://pypi.python.org/simple/ipython/
Best match: ipython 4.1.1
Downloading https://pypi.python.org/packages/source/i/ipython/ipython-4.1.1.zip#md5=445ff597cccb7818c23c988010f62838
Processing ipython-4.1.1.zip
Writing /tmp/easy_install-gDd17B/ipython-4.1.1/setup.cfg
Running ipython-4.1.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-gDd17B/ipython-4.1.1/egg-dist-tmp-QzsqPK
error: Setup script exited with error in ipython setup command: Invalid environment marker: sys_platform == "darwin" and platform_python_implementation == "CPython"
</pre>

fixing the `requirements.txt` like this solves the problem:
<pre>
diff --git a/requirements.txt b/requirements.txt
index 2b00519..1a20c72 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ decorator
 wheel
 pyyaml
 werkzeug
-ipython>=3.0.0
+ipython==4.0.0
 statistics
 requests
 rlp>=0.4.4
</pre>

In short, ipython 4.1.1 is automatically choosen and fails, while 4.0.0 is fine, and it seems related to the following line in `setup.py`:
`207:  ':sys_platform == "darwin" and platform_python_implementation == "CPython"': ['gnureadline'], `

Cheers

minrk:
This was a workaround for old setuptools, but we've added a dependency on 18.5,
so it shouldn't be required anymore.

Setuptools 20.2 has removed python_implementation (perhaps unintentionally),
so the workaround actually _breaks_ on more recent setuptools.

We should have only made this patch when setuptools needed it, but we can remove it outright, now.

closes #9281

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -232,13 +232,6 @@ if not any(arg.startswith('bdist') for arg in sys.argv):
     else:
         install_requires.append('pexpect')
     
-    # workaround pypa/setuptools#147, where setuptools misspells
-    # platform_python_implementation as python_implementation
-    if 'setuptools' in sys.modules:
-        for key in list(extras_require):
-            if 'platform_python_implementation' in key:
-                new_key = key.replace('platform_python_implementation', 'python_implementation')
-                extras_require[new_key] = extras_require.pop(key)
 
 everything = set()
 for key, deps in extras_require.items():


minrk:
We can cut 4.1.2 immediately after merging this.

takluyver:
Does the requirement on setuptools 18.5 actually work if you already have an older version of setuptools when you're installing IPython? I'd expect it upgrades it after parsing the requirements. In any case, it looks like this is the right thing to do.

minrk:
No, if you install with setuptools < 18, it will still fail. But at least the dependency is expressed *somehow* This only affects people *not* using wheels, though. And if you've ever installed IPython from a wheel, you will have the version of setuptools that's needed to install from source.

minrk:
I'll cut 4.1.2 shortly, thanks.

PR #9287, opened 2016-03-02
FIX: use mpl.get_backend() instead of a global

### LINKED ISSUE #9286: %matplotlib notebook broken with current matplotlib master
With current matplotlib master, executing %matplotlib notebook fails with an `AttributeError: module 'matplotlib.backends' has no attribute 'backend'`

Please find a notebook demonstrating this problem here: https://gist.github.com/ericdill/1bfdc12692aec09047bb

matplotlib/matplotlib#6013 is the PR that broke this.

Cross-posted at matplotlib/matplotlib#6092.

cc @tacaswell

tacaswell:
Use `matplotlib.get_backend()` to look up the current backend rather
than looking at a module-level golabl in `matplotlib.backends`.

Closes #9286

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -211,7 +211,7 @@ def select_figure_formats(shell, formats, **kwargs):
 
     [ f.pop(Figure, None) for f in shell.display_formatter.formatters.values() ]
 
-    if matplotlib.backends.backend.lower() == 'nbagg':
+    if matplotlib.get_backend().lower() == 'nbagg':
         formatter = shell.display_formatter.ipython_display_formatter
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

takluyver:
@tacaswell how close are we to a matplotlib release for which this fix will be needed? If it's pretty soon, we can hurry out another bugfix release of IPython; if not, we'll leave it for 4.2.

tacaswell:
This is only broken on master and we will fix it up there as well.  I think this can wait till 4.2 and should be back-compatible with all viable versions of mpl.

takluyver:
Great, thanks for clarifying.

Issue #8865, opened 2015-10-05
Dict key completion for other dict-like objects (e.g., xray.Dataset)


shoyer:
#5304 added auto-completion for `dict` subclasses and `pandas.DataFrame` objects.

I would like to have dictionary key completion work on [`xarray.Dataset`](http://xarray.pydata.org/en/stable/data-structures.html#dataset) objects, which are conceptually very similar to data frames. `Dataset` is a `collections.Mapping` subclass, but it isn't a `dict` or `pandas.DataFrame`.

What's the right way to add this to IPython? I'm happy to issue a PR adding `xarray.Dataset` explicitly to `IPython.core.completer` [alongside `pandas.DataFrame`](https://github.com/ipython/ipython/blob/2880ecf8868419693ef8afb73761552c541140d0/IPython/core/completer.py#L931), but maybe a more generic approach would be appropriate? I'm sure there are many other projects with their own mapping objects that would also be interested.

takluyver:
I think we should do it by checking for some private attribute like `obj._key_completion_safe_`, which package authors (you, in this case) can set to indicate that we should try to complete keys.

Then we'll probably also need to check `obj._this_shouldnt_exist_` (or similar), to override the above for objects that use `__getattr__` to pretend they have arbitrary attributes.

MatthewTurk:
This is something we'd like to explore for yt as well, where dict-like access is used to generate fields on-demand.  So having something generic would be awesome!

takluyver:
#9289 adds this - can you have a look at the proposed API?

shoyer:
Thanks!

PR #9289, opened 2016-03-02
Allow objects to define their own key completions

takluyver:
Objects can specify their own key completions by defining an `_ipy_key_completions_()` method that returns a list of objects which it expects in a subscript expression. E.g.

```
x._ipy_key_completions_() == ['foo']
x['f<TAB>  # completes foo
```

Closes gh-8865

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -937,8 +937,8 @@ class IPCompleter(Completer):
         def get_keys(obj):
             # Objects can define their own completions by defining an
             # _ipy_key_completions_() method.
-            if _safe_really_hasattr(obj, '_ipy_key_completions_'):
-                return obj._ipy_key_completions_()
+            if _safe_really_hasattr(obj, '_ipython_key_completions_'):
+                return obj._ipython_key_completions_()
 
             # Special case some common in-memory dict-like types
             if isinstance(obj, dict) or\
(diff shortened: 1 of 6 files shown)


shoyer:
This looks great to me! :+1:

minrk:
:+1: nice!

We use `_ipython_` in most places I'm aware of for IPython extension points, so perhaps that instead of `_ipy_`.

takluyver:
Good point, changed it to `_ipython_` as suggested.

IPython/core/completer.py
```diff
@@ -472,6 +472,18 @@ def _safe_isinstance(obj, module, class_name):
     return (module in sys.modules and
             isinstance(obj, getattr(__import__(module), class_name)))
 
+def _safe_really_hasattr(obj, name):
```
minrk: Does it make sense to put this phony-method canary in safe_hasattr itself? I seem to recall that we also have this logic in core.formatters.

takluyver: You're right, we do already have such a canary. I don't think `safe_hasattr` is quite the right function for it, but I'll try to find a way to combine them.

takluyver:
I moved the method-checking code into `IPython.utils.dir2`

minrk:
Perfect, thanks.

tacaswell:
Is there an equivalent method to provide for attribute tab complete?

shoyer:
@tacaswell yes, use `__dir__`.

tacaswell:
Ah, now that you say that I did know that at one point :sheep:

PR #9298, opened 2016-03-05
Use terminal width to determine header length

vincentwoo:
Hi everyone! A thing that has bothered me for a long time is how IPython's exception frames always wrap if your terminal is less than 75 columns wide. As an example, this is what your console looks like if it's only 73 columns wide:

```
-------------------------------------------------------------------------
--
NameError                                 Traceback (most recent call las
t)
/home/coderpad/solution.py in <module>() 
      5     say_hello() 
      6 
----> 7     null.asdf 

NameError: name 'null' is not defined 
```

So I figured I'd go in and replace this magic constant `75` with an adaptive thing based on how wide your terminal was. It appeared that `terminal.py` had a provision to do just that: `get_terminal_size`. Unfortunately, it looked like this method was only implemented on windows, oddly enough.

To get around this, I introduced a dependency on [backports.shutil_get_terminal_size](https://pypi.python.org/pypi/backports.shutil_get_terminal_size), which is a backport of a function that does exactly this that was introduced in Python 3.3. It's pure python and should be pretty portable, but was unsure if this was an okay thing to add to the project, or if I did that correctly. Let me know!

Having done that, I changed `ultratb.py` to allow for terminals narrower than 75. I capped header width at 75, still, because it does look crazy to see hyphens going _all_ the way across a very wide terminal.

So now you get (on a 73 wide tty):

```
-------------------------------------------------------------------------
NameError                               Traceback (most recent call last)
```

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -124,6 +124,7 @@ from IPython.utils import path as util_path
 from IPython.utils import py3compat
 from IPython.utils import ulinecache
 from IPython.utils.data import uniq_stable
+from IPython.utils.terminal import get_terminal_size
 from logging import info, error
 
 import IPython.utils.colorable as colorable
@@ -1029,20 +1030,21 @@ class VerboseTB(TBTools):
         colors = self.Colors  # just a shorthand + quicker name lookup
         colorsnormal = colors.Normal  # used a lot
         exc = '%s%s%s' % (colors.excName, etype, colorsnormal)
+        width = min(75, get_terminal_size()[0])
         if long_version:
             # Header with the exception type, python version, and date
             pyver = 'Python ' + sys.version.split()[0] + ': ' + sys.executable
             date = time.ctime(time.time())
 
-            head = '%s%s%s\n%s%s%s\n%s' % (colors.topline, '-' * 75, colorsnormal,
-                                           exc, ' ' * (75 - len(str(etype)) - len(pyver)),
-                                           pyver, date.rjust(75) )
+            head = '%s%s%s\n%s%s%s\n%s' % (colors.topline, '-' * width, colorsnormal,
+                                           exc, ' ' * (width - len(str(etype)) - len(pyver)),
+                                           pyver, date.rjust(width) )
             head += "\nA problem occurred executing Python code.  Here is the sequence of function" \
                     "\ncalls leading up to the error, with the most recent (innermost) call last."
         else:
             # Simplified header
             head = '%s%s' % (exc, 'Traceback (most recent call last)'. \
-                             rjust(75 - len(str(etype))) )
+                             rjust(width - len(str(etype))) )
 
         return head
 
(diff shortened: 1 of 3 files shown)


minrk:
This seems great to me. Does it make sense for the import and/or dependency to be conditional on the Python versions that require the backport?

vincentwoo:
Since I don't really code much Python and I don't know about what IPython prefers, I opted for the thing that I thought had the smallest surface area. I'll happily conditionally import if that's what you like.

There's also the option of removing ` IPython.utils.terminal.get_terminal_size` entirely and just using the backport directly.

minrk:
We can refine the dependency in future PRs. Thanks!

vincentwoo:
@minrk thanks for merging! Any sense for when IPython 5 is coming out, or if we can get this into 4.2? :)

minrk:
5.0's probably a bit of a way off. I think we can probably put this in 4.2.

vincentwoo:
Awesome!

Achifaifa:
Hi! I found this looking for some issue I'm having. I merged from master, and now I get an import error with `No module named shutil_get_terminal_size`. I suppose this is related somehow? 

Is there a new dependency I wasn't aware of? Or it's some other thing I may be missing?

takluyver:
Yes, a new dep on [backports.shutil_get_terminal_size](https://pypi.python.org/pypi/backports.shutil_get_terminal_size). You can see the change in setup.py.

Achifaifa:
Great, fixed. Thank you.

fperez:
Mmh, I'd suggest catching that import of `backports.shutil_get_terminal_size` and issuing a clearer message. Given that the name of that module carries a dot in it, it's really confusing: I spent some time tracking down the problem down to this PR (much like @Achifaifa), before realizing that I wasn't missing some module called `backports`, but instead that someone literally decided to name a package with a dot in it!!  

I guess there's nothing we can do about that (I think it's unnecessarily confusing to users, imagine if people start making lots of top-level packages with dots in their names, the fun we're going to have tracking down import errors).  But at least we can mitigate it via a message, given this is such a special case.

Just my 2 cents...

vincentwoo:
Are these issues happening because people are developing IPython and just doing a git update without reinstalling packages? What little I understand of pip is that for consumers, this problem shouldn't happy because the package should be installed at pip install or update time.

takluyver:
Yep, it's only an issue for people updating from git - anyone installing a released version, or even installing afresh from git, will get it automatically.

We've been down the road of trying to add better error messages before (with jsonschema), and it later ended up obscuring a problem with jsonschema's dependency specifications, because there was an `ImportError` coming from some other package. I don't think we should try to be clever with this - having to deal with new dependencies occasionally is an occupational hazard of living on master. Maybe we should ping them mailing list to let people know what they need to install, but I don't think it needs changes to the code itself.

fperez:
Yes, I can see the problem with potentially masking other errors.  That could be mitigated by simply adding the info message without blocking the traceback, so the full original traceback is always visible, only with a catch message:

```python
try:
  import foo
except ImportError:
  warn("you seem to be missing new dependency foo, see http://url/to/foo to get it")
  raise
```

But I'm not going to push it :) At least, a little PSA-style announcement on-list when new dependencies arise would be a good idea.  That way, at least the devs who work off master get a heads-up.  In my case, what happened was that I did a git pull but didn't re-run right away, and only noticed the problem on a plane without internet :)  I knew what to do, the git reflog is always my close friend, but a simple heads-up would have saved me a bit of time (esp. later, when I was online and tried to get to the bottom of the problem).

I know the risks of living off master's bleeding edge, but that doesn't mean we can't do a few simple things to soften the edge a little :)

PR #9299, opened 2016-03-06
FIX: account for dpi == 'figure'

tacaswell:
In mpl 1.5.0 'figure' became a valid dpi value, in 2.0 it will
become the default.

xref https://github.com/matplotlib/matplotlib/issues/6113

This should be backported as far back as you are willing to backport.


minrk:
Merged and applied backport to 4.x, 4.0.x, 3.x. If they get another release, this'll be in there.

PR #9304, opened 2016-03-08
WIP Sphinxify docstrings

SylvainCorlay:
Rich html rendering of docstrings in the notebook.

docrepr

The notebook front-end pager already takes both plain text and html. See [here](https://github.com/jupyter/notebook/blob/a1458ce8fc275a7acdbd47a86fd7072e989d38b1/notebook/static/notebook/js/pager.js#L90)

On the other side, the `page` function takes either raw text or a mime bundle. See [here](https://github.com/ipython/ipython/blob/master/IPython/core/page.py#L250).

In this PR, I check for presence of [docrepr](https://github.com/spyder-ide/docrepr). If it is there, I add a `text/html` entry to the mime bundle.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1578,7 +1578,7 @@ class InteractiveShell(SingletonConfigurable):
         with self.builtin_trap:
             info = self._object_find(oname)
             if info.found:
-                return self.inspector._format_info(info.obj, oname, info=info,
+                return self.inspector._get_info(info.obj, oname, info=info,
                             detail_level=detail_level
                 )
             else:
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -75,6 +75,7 @@ from IPython.utils.strdispatch import StrDispatch
 from IPython.utils.syspathcontext import prepended_to_syspath
 from IPython.utils.text import (format_screen, LSString, SList,
                                 DollarFormatter)
+from IPython.utils.tempdir import TemporaryDirectory
 from traitlets import (Integer, Bool, CBool, CaselessStrEnum, Enum,
                                      List, Dict, Unicode, Instance, Type)
 from warnings import warn
@@ -85,11 +86,11 @@ try:
     import docrepr.sphinxify as sphx
 
     def docformat(doc):
-        dirname = tempfile.mkdtemp(prefix='ipython_sphinxify_')
-        return {
-            'text/html': sphx.sphinxify(doc, dirname),
-            'text/plain': doc
-        }
+        with TemporaryDirectory() as dirname:
+            return {
+                'text/html': sphx.sphinxify(doc, dirname),
+                'text/plain': doc
+            }
 except:
     docformat = None
 
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -81,6 +81,18 @@ from warnings import warn
 from logging import error
 import IPython.core.hooks
 
+try:
+    import docrepr.sphinxify as sphx
+
+    def docformat(doc):
+        dirname = tempfile.mkdtemp(prefix='ipython_sphinxify_')
+        return {
+            'text/html': sphx.sphinxify(doc, dirname),
+            'text/plain': doc
+        }
+except:
+    docformat = None
+
 #-----------------------------------------------------------------------------
 # Globals
 #-----------------------------------------------------------------------------
@@ -1539,7 +1551,7 @@ class InteractiveShell(SingletonConfigurable):
         info = self._object_find(oname, namespaces)
         if info.found:
             pmethod = getattr(self.inspector, meth)
-            formatter = format_screen if info.ismagic else None
+            formatter = format_screen if info.ismagic else docformat
             if meth == 'pdoc':
                 pmethod(info.obj, oname, formatter)
             elif meth == 'pinfo':
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -112,7 +112,8 @@ def getdoc(obj):
 
     It also attempts to call a getdoc() method on the given object.  This
     allows objects which provide their docstrings via non-standard mechanisms
-    (like Pyro proxies) to still be inspected by ipython's ? system."""
+    (like Pyro proxies) to still be inspected by ipython's ? system.
+    """
     # Allow objects to offer customized documentation via a getdoc method:
     try:
         ds = obj.getdoc()
@@ -122,14 +123,13 @@ def getdoc(obj):
         # if we get extra info, we add it to the normal docstring.
         if isinstance(ds, string_types):
             return inspect.cleandoc(ds)
-    
     try:
         docstr = inspect.getdoc(obj)
         encoding = get_encoding(obj)
         return py3compat.cast_unicode(docstr, encoding=encoding)
     except Exception:
         # Harden against an inspect failure, which can occur with
-        # SWIG-wrapped extensions.
+        # extensions modules.
         raise
         return None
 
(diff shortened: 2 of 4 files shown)


SylvainCorlay:
So this is still just the quick hack I made yesterday night. Doing this the right way requires some more work refactoring the pdoc and pinfo magics.

ccordoba12:
Thanks Sylvain for working on this!

IPython/core/interactiveshell.py
```diff
@@ -81,6 +81,17 @@
 from logging import error
 import IPython.core.hooks
 
+try:
+    import docrepr.sphinxify as sphx
+
+    def docformat(doc):
+        return {
+            'text/html': sphx.sphinxify(doc, '/Users/scorlay/tmp'),
```
minrk: Presumably this should be a temp dir from `tempfile`.

IPython/core/oinspect.py
```diff
@@ -551,8 +551,12 @@ def _format_fields(self, fields, title_width=0):
     
     def _format_info(self, obj, oname='', formatter=None, info=None, detail_level=0):
         """Format an info dict as text"""
-        info = self.info(obj, oname=oname, formatter=formatter,
+
+        # hack docstring rendering
+        info = self.info(obj, oname=oname, formatter=None,
                             info=info, detail_level=detail_level)
+        return formatter(info["docstring"])
```
minrk: Looks like `formatter` can technically be None. Is that not true?

minrk:
This is awesome. Thanks for getting it started!

takluyver:
:+1: (for the idea - the implementation still needs some polish, but you know that ;-)

Carreau:
Cool. I didn't remember the docstring could be mime-bundle. 

How does a module/function announce the format of its docstrings ?

ssanderson:
:exclamation: :grey_exclamation: :exclamation:

SylvainCorlay:
> How does a module/function announce the format of its docstrings?

It does not. docrepr tries to determine it by looking of occurrences of certain strings in the docstring. https://github.com/spyder-ide/docrepr/blob/master/docrepr/sphinxify.py#L62

SylvainCorlay:
I will update the PR tonight. I have got to complete something unrelated to Jupyter this afternoon...

takluyver:
It looks like the test failure is to do with a change in the output for a magic command:

```
FAIL: IPython.core.tests.test_oinspect.test_pinfo_magic
  File "/home/travis/virtualenv/python3.5.0/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/ipython/ipython/IPython/core/tests/test_oinspect.py", line 424, in test_pinfo_magic
    ip._inspect('pinfo', 'lsmagic', detail_level=0)
  File "/home/travis/build/ipython/ipython/IPython/testing/tools.py", line 390, in __exit__
    assert s in printed, notprinted_msg.format(s, self.channel, printed)
AssertionError: Did not find 'Docstring:' in printed output (on stdout):
-------
List currently available magic functions.
-------
```

SylvainCorlay:
I updated the PR.

I have some questions about the `format` method. Maybe @Carreau will have some insights.

IPython/core/interactiveshell.py
```diff
@@ -81,6 +81,18 @@
 from logging import error
 import IPython.core.hooks
 
+try:
+    import docrepr.sphinxify as sphx
+
+    def docformat(doc):
+        dirname = tempfile.mkdtemp(prefix='ipython_sphinxify_')
```
SylvainCorlay: Should probably be using `tempfile.TemporaryDirectory`.

takluyver: For compatibility with Legacy Python, use our copy from `IPython.utils.tempdir.TemporaryDirectory`.

SylvainCorlay: done.

Carreau:
ipdb test complain because of the following missing  at least. 

```
     Type:           ExampleClass
     String form:    ExampleClass()
     Namespace:      Local...
     Docstring:      Docstring for ExampleClass.
     Init docstring: Docstring for ExampleClass.__init__
```

Carreau:
I'll try to have a deeper look.

IPython/core/oinspect.py
```diff
@@ -633,26 +671,37 @@ def pinfo(self, obj, oname='', formatter=None, info=None, detail_level=0):
...
-        if text:
-            page.page(text)
-    
+        info = self._get_info(obj, oname, formatter, info, detail_level)
+        if info:
+            page.page(info)
+
     def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
+        """DEPRECATED. Compute a dict with detailed information about an object.
+        """
```
Carreau: likely add a `warnign.warn('blah blah deprecated as of IPtyhon 5.0 and will be removed in future versions... blah', deprecationWarning, stack_level=2)` here.

IPython/core/oinspect.py
```diff
@@ -827,7 +874,7 @@ def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
             if safe_hasattr(obj, '__call__') and not is_simple_callable(obj):
                 call_def = self._getdef(obj.__call__, oname)
                 if call_def:
-                    call_def = self.format(call_def)
+                    call_def = call_def
```
Carreau: meh ?

SylvainCorlay: Yeah, I have questions about how format works - which can be done offline. 

In any case, it feels like it should be a formatter and not directly applied in the info dict.

Carreau:
Bug with this branch:

```python
In [1]: class Foo():pass

In [2]: pinfo Foo
Init signature:
[0mFoo[0m[0;34m([0m[0;34m)[0m[0;34m[0m[0m

*** ERROR: EOF in multi-line statement


Docstring:
<no docstring>
Type:
type
```

while master:

```python
In [1]: class Foo():pass

In [2]: pinfo Foo
Init signature: Foo()
Docstring:      <no docstring>
Type:           type
```

IPython/core/oinspect.py
SylvainCorlay: Done.

IPython/core/oinspect.py
```diff
@@ -633,26 +671,37 @@ def pinfo(self, obj, oname='', formatter=None, info=None, detail_level=0):
...
 
         - detail_level: if set to 1, more information is given.
         """
-        text = self._format_info(obj, oname, formatter, info, detail_level)
-        if text:
-            page.page(text)
-    
+        info = self._get_info(obj, oname, formatter, info, detail_level)
+        if info:
+            page.page(info)
```
Carreau: So `pinfo` here is API and my guess is it should still `page` the text  instead of the mime bundle, as pagers are expecting text (in terminal for example). 

Should be add a method `rich_page`, that take a mimebundle, and call it if present ?  But I doubt we can just change what we can pass to page or all hell will break loose, and as it has already frozen with bash on windows I dont' want to have to deal with angry cold deamons.

Carreau:
Thanks for the fixes. Trying to wrap my head around and figure out how to move forward without breaking things.

IPython/core/oinspect.py
SylvainCorlay: The page docstring claims it takes mime bundle and it works in the regular console.

Apparently, the ipdb pager is the only one that only takes text.

Carreau: > The fallback pager is the culprit.

Cool, thanks for digging into that !

ccordoba12:
@SylvainCorlay, I think it would be more efficient to use a single directory and save in there all temp output generated by `docrepr`. Then we could register an `atexit` callback to remove it. At least that's how I designed `docrepr` to be used :-)

Carreau:
mehhhhh when I run this locally it work. My guess is test use a specific pager that does not handle correctly wether it pages strings or mime bundle...

takluyver:
Yes, paging == printing in the tests at the moment:

https://github.com/ipython/ipython/blob/c17dfd79f5c5e8d1b31f8f4ce225f12e7fc0a78e/IPython/testing/globalipapp.py#L142

Carreau:
GRbrbrbrlrblrbrlrb

Carreau:
> GRbrbrbrlrblrbrlrb

That's mean fix in progress.

SylvainCorlay:
> > GRbrbrbrlrblrbrlrb

> That means "fix in progress"

sregnegne

PR #9305, opened 2016-03-08
Adding savef functionality for line-by-line writing of python script

vegabondx:
This functionality enables us to write to a set file line by line instead of block ( with set commands ). In short , it pushes commands to the "stack" which is a file. We can extend this to last N line instead of just the last line. Note that this CANNOT be achieved via alias. We can only do  "save -a -f filename.py" however we would still need to supply the line range.    

I have seen that save is quite inefficient in command line operation since you would need to recheck the saved files for wrong commands ( since all commands do not work ) which becomes cumbersome to go through once you have obtained the result. It is hence much efficient to copy to the file then and there once we realize the set of commands work.  

Possible Enhancements (1) merge with save (2) add option to append Last N lines
Tests conducted: iptest with both python2 and python3. Solution works , kindly suggests improvements / coding changes or compatibility issue. Feel free to suggest alternatives ( names , variables )

### DIFF
--- a/IPython/core/magics/code.py
+++ b/IPython/core/magics/code.py
@@ -53,6 +53,7 @@ range_re = re.compile(r"""
  (?P<end>\d+)?)?
 $""", re.VERBOSE)
 
+filename='temp.py'
 
 def extract_code_ranges(ranges_str):
     """Turn a string of range for %%load into 2-tuples of (start, stop)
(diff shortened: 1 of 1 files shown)


Carreau:
Hi there, 

welcome to GitHub, I see this is your first contribution !

You do not **need** to patch IPython to have a new magic, you can just publish your own package on PyPI. See for example one of [mine](https://github.com/Carreau/cffi_magic/blob/master/cffi_magic/__init__.py). (Which might not have the  best practices though)

Creating your own package is not that hard, and it allows you to publish it at your own rythme, and make the magic available on older version of IPython, without getting blocked by our slow release schedule. 

(I would suggest looking at [flit](http://flit.readthedocs.org/en/latest/) which is not as flexible as normal setup.py but much simpler) 

You can in particular see [our docs](http://ipython.readthedocs.org/en/stable/config/custommagics.html) also on how to do a custom magic, and use `%load_extension` to load it. 

We usually prefer for extension to live (a bit) in their own package for a bit, and then, once they mature, and are stable enough, potentially migrate them in IPython itself (though extension tend to go the other way, because users want updates every month instead of every year). 

Do you feel like creating your own package ? We can help you with that too if you are still uncomfortable. 

I'll still review this code  and put comment inline if necessary.

vegabondx:
Hey @Carreau 

I sincerely thank you for an detailed informative and encouraging response. Indeed I am a novice and realized a practical difficulty while working on a server decided to look it up myself. I ,as you have noticed, am also new to open source software development and dont know things about how these things work although I went through all the possible documents and tested the code before submitting for a review.

I will go though your directives and choose one of those ( pypi + filt ) or magic extensions and see how I can easily download these onto a server with IPython and start using it there. I think your package would definitely act as a handy reference. 

Lastly , what do you suggest I do with this pull request ? ( I have added a some enhancement such that it can push multiple lines ).

Carreau:
> I sincerely thank you for an detailed informative and encouraging response

No problem, we try to do our best, though we dont' always have time to extensively respond to everyone. 

> I went through all the possible documents and tested the code before submitting for a review.

If you find that some document are too hidden, or that some information could be added in some place, feel free to tell us, we can try to improve. 


> I will go though your directives and choose one of those ( pypi + filt ) or magic extensions and see how I can easily download these onto a server with IPython and start using it there. I think your package would definitely act as a handy reference.

Feel free to ask more questions it can be difficult to figure things out. 
And we would be happy to get more contributor, we are low on manpower. 

> Lastly , what do you suggest I do with this pull request ? ( I have added a some enhancement such that it can push multiple lines ).

You can leave it open for now, once you have a repository with that extension in it, you can add the link as a comment, and than we can close the pull requests. 
If you have some enhancement, you can just commit and push on the same branch, the Pull Request will be automatically updated to reflect the latest state.

Carreau:
@vegabondx did you achieved to do an extension ?

vegabondx:
Hey @Carreau . thanks for the information. I did explore the options you suggested but did not try to experiment with those yet. Let me do that by the weekend. Thanks once again.

vegabondx:
Hi Matthias , 
Thanks for your help in advance ! 
I was able to get a couple of things going1. I created a magic functionality as a Class as specified in  custommagics documentation  and using your repository as reference. Here is the link to my Git repo: https://github.com/vegabondx/savef. I have verified that if i run the file as %run savef.py it loads the magic function correctly. 
2. I was also able to make a PyPi package using flit command , again utilizing your repo as reference. I was able to create a package. 
However when I install the package , it does not "install" savef functionality in ipython by default. However when i do the following it works
In [2]: %savefERROR:root:Line magic function `%savef` not found.In [3]: %load_ext savefThe savef module is not an IPython extension.
In [4]: %savef -f
Starting python file: `temp.py`
( i.e. working properly ) 


Could you let me know what I may be missing here ? Also is there any stylistic improvements I should make ? 

Thank you for your guidance !Abhijit

Date: Tue, 22 Mar 2016 11:45:45 -0700

Carreau:
Sorry, my bad, my example was doing some weirds things I should have fixed. I was experimenting with some stuff at import time instead of following best practices. 

i submitted a PR to fix that: 

https://github.com/vegabondx/savef/pull/1

vegabondx:
Thanks @Carreau ! This functionality is now part of savef package in pypi. 
https://pypi.python.org/pypi/savef
https://github.com/vegabondx/savef/

Carreau:
Ok, now that this is working I'm going to close the PR !

Thanks a lot, and looking forward to see more of your PRs !

PR #9306, opened 2016-03-08
Override terminal size in doctests to standardise traceback format

takluyver:
Should fix test failures on Jenkins


takluyver:
[Working](https://jenkins.jupyter.org/job/Linux-PR/1/)

PR #9312, opened 2016-03-10
Add prompt-toolkit eventloop for OS X

minrk:
For `--matplotlib osx`

Uses Native C, Obj-C APIs via ctypes.

### DIFF
--- a/IPython/terminal/pt_inputhooks/osx.py
+++ b/IPython/terminal/pt_inputhooks/osx.py
@@ -6,7 +6,6 @@ import ctypes.util
 
 objc = ctypes.cdll.LoadLibrary(ctypes.util.find_library('objc'))
 CoreFoundation = ctypes.cdll.LoadLibrary(ctypes.util.find_library('CoreFoundation'))
-# Cocoa = ctypes.cdll.LoadLibrary(ctypes.util.find_library('Cocoa'))
 
 void_p = ctypes.c_void_p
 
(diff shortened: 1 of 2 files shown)


Carreau:
Gosh, I'm happy to be writing Python and not C. 

+1

minrk:
This would be much simpler if it were actually in Obj-C. Wrapping it in ctypes is the most not-nice part.

IPython/terminal/pt_inputhooks/osx.py
```diff
@@ -0,0 +1,133 @@
...
+def _NSApp():
+    """Return the global NSApplication instance (NSApp)"""
+    return msg(C('NSApplication'), n('sharedApplication'))
+
+
+def _wake(NSApp):
+    """Wake the Application"""
+    event = msg(C('NSEvent'),
+        n('otherEventWithType:location:modifierFlags:'
+          'timestamp:windowNumber:context:subtype:data1:data2:'),
```
jdfreder: :eyes: whoa

jdfreder:
:+1:

IPython/terminal/pt_inputhooks/osx.py
minrk: Yeah, Calling objC from ctypes is gross. In ObjC this would look like pretty normal kwargs:

```objc
event = [ NSEvent otherEventWithType: ApplicationDefined
                            location: 0
                       modifierFlags: 0
                           timestamp: 0
                        windowNumber: 0
                             context: nil
                             subtype: 0
                               data1: 0
                               data2: 0
]
```

Issue #9309, opened 2016-03-09
[Prompttoolkit] Esc should dismiss reverse search ?


Carreau:
`Crtl-R`  show a reverse search toolbar:

```
In [38]: def foo(object):
    ...:     something
    ...:     else

I-search backward: some|
```

I expect esc to dismiss. It does not. Thoughts ?

Carreau:
Note, Ctrl-C/Ctrl-D, might be good shortcut to dismiss too.

takluyver:
Ping @jonathanslenders - are there any default shortcuts in the Ctrl-R search?

jonathanslenders:
We have ControlG and ControlC to dismiss. Probably I can add Escape as well. (Need to verify this, but it looks like readline accepts escape.) See here for the default shortcuts: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/emacs.py

The reason that ControlC is not working is probably because of this: https://github.com/ipython/ipython/blob/master/IPython/terminal/ptshell.py#L135
That key binding is always active. That means that this will override the default ControlC implementation for the search. Better is to add a filter to make this only handle ControlC when the default buffer is active:
```python
from prompt_toolkit.filters import HasFocus
from prompt_toolkit.enums import DEFAULT_BUFFER
@kbmanager.registry.add_binding(Keys.ControlC, filter=HasFocus(DEFAULT_BUFFER))    
```

I will verify whether the Escape binding can be added next weekend. I also noticed a few minor bugs in the latest prompt_toolkit release. (Thanks to several people for testing/reporting) So, I'll have to push a new bugfix release very soon as well.

Jonathan

takluyver:
Thanks, that should definitely be using a filter. It's taking me a while to get used to PT's filters - I like them, but I'm not used to thinking like that. I'll fix it now.

Carreau:
Closing as per [upstream issue](https://github.com/jonathanslenders/python-prompt-toolkit/issues/265):

> We can't do this by default. Handling escape in reverse search would mean that Alt-Enter doesn't won't work anymore. (Alt-enter in reverse search should execute the current match right away.)
> It's more important to have alt-enter work than to have Esc as an alias for Control-G and Control-C.

PR #9313, opened 2016-03-10
Only handle Ctrl-C when in main input buffer

takluyver:
This allows Ctrl-C to work as normal within Ctrl-R search (see gh-9309)


Carreau:
+1, 

though, now, it dismiss the search immediately. I think in a second iteration it would be nice for Ctrl-C to dismiss search only if buffer is empty, and clear the buffer in the other case (like in other buffers). 

But that can be another PR.

PR #9317, opened 2016-03-11
remove nag screen delay

ivanov:
see https://twitter.com/njgoldbaum/status/708360792677208064

and  my response, which has support of 3 other people.

https://twitter.com/ivanov/status/708409804448993280

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -202,12 +202,8 @@ class BaseIPythonApplication(Application):
             import time
             self.log.warning("Subcommand `ipython {sub}` is deprecated and will be removed "
                              "in future versions.".format(sub=subc))
-            self.log.warning("You likely want to use `jupyter {sub}`... continue "
-                             "in 5 sec. Press Ctrl-C to quit now.".format(sub=subc))
-            try:
-                time.sleep(5)
-            except KeyboardInterrupt:
-                sys.exit(1)
+            self.log.warning("You likely want to use `jupyter {sub}`... in the"
+                             "future".format(sub=subc))
         return super(BaseIPythonApplication, self).initialize_subcommand(subc, argv)
 
     def init_crash_handler(self):


ivanov:
I'd also argue that the old ipython subcommands should continue to work without being deprecated - why would we ever want to isolate the original base of ipython users?

ivanov:
@matthewturk @jiffyclub and Nathan seem to be  supportive of this.

takluyver:
Yeah, I don't care for the 'wait five seconds' thing. I don't mind the warning, and I can see us *eventually* removing the ipython subcommands that are now Jupyter, but I don't think we need to rush to do that. And making it work after a delay feels deliberately obtuse - it should either work properly or not work at all.

jdfreder:
Thanks @ivanov !

ngoldbaum:
Thank guys, I was feeling bummed about this and am glad to see that you're amenable to a patch.

IPython/core/application.py
```diff
@@ -202,12 +202,8 @@ def initialize_subcommand(self, subc, argv=None):
...
             self.log.warning("Subcommand `ipython {sub}` is deprecated and will be removed "
                              "in future versions.".format(sub=subc))
-            self.log.warning("You likely want to use `jupyter {sub}`... continue "
-                             "in 5 sec. Press Ctrl-C to quit now.".format(sub=subc))
-            try:
-                time.sleep(5)
-            except KeyboardInterrupt:
-                sys.exit(1)
+            self.log.warning("You likely want to use `jupyter {sub}`... in the"
+                             "future".format(sub=subc))
```
takluyver: The message will end 'thefuture', because there's no space between the words, I think. Anyone want to do a quick extra PR?

Also, I think the `...` is probably superfluous now.

ivanov: oops, on it

ivanov:
#9318

Issue #9351, opened 2016-03-23
unexisting variable self.log


aongenae:
because of a corrupt sqlite db, I hit the following piece of code in 
`IPython/core/history.py`
https://github.com/ipython/ipython/blob/master/IPython/core/history.py#L520

`self.log` does not seems to exists.

Here is the trace:
```
└─⊳ ipython
Traceback (most recent call last):
  File "/home/aongenae/newenv/bin/ipython", line 9, in <module>
    load_entry_point('ipython', 'console_scripts', 'ipython')()
  File "/home/aongenae/ipython/IPython/__init__.py", line 119, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/home/aongenae/newenv/lib/python2.7/site-packages/traitlets/config/application.py", line 595, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-110>", line 2, in initialize
  File "/home/aongenae/newenv/lib/python2.7/site-packages/traitlets/config/application.py", line 74, in catch_config_error
    return method(app, *args, **kwargs)
  File "/home/aongenae/ipython/IPython/terminal/ipapp.py", line 305, in initialize
    self.init_shell()
  File "/home/aongenae/ipython/IPython/terminal/ipapp.py", line 321, in init_shell
    ipython_dir=self.ipython_dir, user_ns=self.user_ns)
  File "/home/aongenae/newenv/lib/python2.7/site-packages/traitlets/config/configurable.py", line 405, in instance
    inst = cls(*args, **kwargs)
  File "/home/aongenae/ipython/IPython/terminal/ptshell.py", line 238, in __init__
    super(TerminalInteractiveShell, self).__init__(*args, **kwargs)
  File "/home/aongenae/ipython/IPython/core/interactiveshell.py", line 466, in __init__
    self.init_history()
  File "/home/aongenae/ipython/IPython/core/interactiveshell.py", line 1581, in init_history
    self.history_manager = HistoryManager(shell=self, parent=self)
  File "/home/aongenae/ipython/IPython/core/history.py", line 520, in __init__
    self.log.error("Failed to create history session in %s. History will not be saved.",
AttributeError: 'HistoryManager' object has no attribute 'log'
```

takluyver:
Thanks, #9352 should fix that.

makmanalp:
Just pinging to say that I hit something like this today with 4.1.1:

```
Traceback (most recent call last):
[...]

PR #9352, opened 2016-03-23
Inherit history machinery from LoggingConfigurable

takluyver:
HistoryManager uses log (in an error case, which is why we didn't spot it before).

Closes gh-9351

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -27,7 +27,7 @@ except ImportError:
 import threading
 
 # Our own packages
-from traitlets.config.configurable import Configurable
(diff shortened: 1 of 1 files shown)


aongenae:
This was fast... seems indeed to be the right action.
Thank you

minrk:
@takluyver is this a regression in 5.0, or do we want to backport for 4.2?

takluyver:
Backport:
https://github.com/ipython/ipython/blob/rel-4.1.2/IPython/core/history.py#L520

minrk:
Thanks!

Issue #9359, opened 2016-03-27
Confirm-on-exit broken on master


fperez:
This is a regression introduced by the new ptshell code, specifically [here](https://github.com/ipython/ipython/blob/master/IPython/terminal/ptshell.py#L260). It should honor the standard `c.TerminalInteractiveShell.confirm_exit = True` flag in `ipython_config.py`.

Pinging @takluyver...

fperez:
The [equivalent code](https://github.com/ipython/ipython/blob/master/IPython/terminal/interactiveshell.py#L789) in `InteractiveShell` checks the `confirm_exit` flag.

takluyver:
Thanks, #9360 should fix this.

PR #9360, opened 2016-03-27
Add and use confirm_exit flag in prompt_toolkit shell

takluyver:
Closes gh-9359

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -8,7 +8,7 @@ from IPython.core.interactiveshell import InteractiveShell
 from IPython.utils.py3compat import PY3, cast_unicode_py2, input
 from IPython.utils.terminal import toggle_set_term_title, set_term_title
 from IPython.utils.process import abbrev_cwd
-from traitlets import Bool, Unicode, Dict
+from traitlets import Bool, CBool, Unicode, Dict
 
 from prompt_toolkit.completion import Completer, Completion
 from prompt_toolkit.enums import DEFAULT_BUFFER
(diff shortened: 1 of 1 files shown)


fperez:
This fixes the problem, thanks!

But I'm a little worried about the amount of copy/pasted code between `interactiveshell.py` and `ptshell.py`... Shouldn't these two share more code, so we don't have to fix the same issues in two places? I missed the details of the refactoring, so I may be off-base, but it looks to me like a common base class with everything that's common should make more sense than having a `class TerminalInteractiveShell(InteractiveShell)` subclassing in both files...

takluyver:
The old TIS is going to go away before 5.0, so it ultimately won't be
duplicated. However, it couldn't be removed while jupyter_console was still
subclassing it. I just severed that link a few days ago, so removing the
old TIS is in the near future.

This does mean that jupyter_console 4 won't work with IPython 5, but I
think that's an acceptable cost as we decouple those pieces.

fperez:
Got it. But what I am worried about is the breakage for users. It is a
weakness of our config system, but the reality is that we expose to the
public our actual class names. So if the class goes away, every user who
has configured things to, e.g. suppress the exit confirmation, will now get
an error message, and things won't work as expected.

I don't see an easy, clean solution, but I really don't want to keep
breaking the user experience like this... Thoughts?

takluyver:
That's why the new class has the same name as the old, to minimise breakage
:-). Some specific settings are going to stop working where the changes
directly affect them, e.g. prompt customisation, but anything else should
be preserved.

I'll go through the config options on the old class before I get rid of it,
and make sure that as many as possible are copied to the new one.

minrk:
We've deprecated configurable class names in the past for exactly this reason. You can see [InlineBackend](https://github.com/ipython/ipykernel/blob/4.3.0/ipykernel/pylab/config.py#L41) for an example.

PR #9362, opened 2016-03-28
Fix a typo: plot_to_latex -> poly_to_latex

gokhanettin:
It seems  it is `poly_to_latex` function to be registered, not `plot_to_latex`


takluyver:
Well spotted, thanks!

Issue #9361, opened 2016-03-27
Odd, broken buffering of internal messages...


fperez:
Consider the following session (this is on master, right now at `51fa9ee`):

```
In [1]: alias_magic foo

In [2]: 1+2
Out[2]: 3

In [3]: exit
UsageError: the following arguments are required: targetdenali[~]>
```

My call to `alias_magic` was missing one argument, but the error message got completely swallowed (even though it's a print statement inside the `alias_magic` code). It only makes it to stdout at shutdown.  I tried with longer sessions, creating more errors, and they just keep piling up internally until the exit.  Very odd.

Ping @takluyver, I suspect this is also related to the recent ptoolkit work.

takluyver:
Can reproduce, will investigate, but this doesn't look like something that would result from the prompt_toolkit changes.

takluyver:
Should be fixed by #9365

PR #9365, opened 2016-03-28
Show UsageError exception immediately

takluyver:
Closes gh-9361

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1771,7 +1771,7 @@ class InteractiveShell(SingletonConfigurable):
         
         These are special exceptions that shouldn't show a traceback.
         """
-        self.write_err("UsageError: %s" % exc)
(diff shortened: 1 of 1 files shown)


fperez:
Question: does this mean that `self.write_err` is completely broken? If so, is it used anywhere else and should we get rid of it?

takluyver:
`self.write_err` still works if you specify a trailing newline so that some buffer gets flushed. However, I believe there's no longer any reason to use it rather than just writing to `sys.stderr`. I plan to deprecate `InteractiveShell.{write, write_err}` and `IPython.utils.io.{stdout, stderr}`.

fperez:
Sounds good.  Let's make those sloooow deprecations, leaving the API entry points if need be as just redirects to `print(...)`, but I'd like to see us follow as much as possible the "don't break user space" Linus rule...

PR #9367, opened 2016-03-30
Deprecate io.{stdout,stderr} and shell.{write,write_err}

takluyver:
I believe there is no longer any reason to use these


minrk:
Nice.

PR #9369, opened 2016-04-01
ControlZ should suspend IPython to the background.

jonathanslenders:
When ControlZ has been pressed, the shell should suspend on platforms that support it.


Carreau:
Whooo!

PR #9371, opened 2016-04-01
Created IPythonPTLexer (wrapper around BashLexer and PythonLexer)

jonathanslenders:
Created IPythonPTLexer which is a wrapper around BashLexer and PythonLexer.
When the input starts with an exclamation mark, it should use shell highlighting.

(Something similar can be done to get nice highlighting of a %magic.)


Carreau:
Awesome ! At some point we might want to improve the lexer to also detect assignment like `files = !ls`, but it's a good first step !

Thanks !

PR #9373, opened 2016-04-04
Restore option to auto-edit files on syntax error

takluyver:
I'm not entirely sure this was working before the move to prompt_toolkit, but this makes it work.

I have gone through the options on the old TerminalInteractiveShell, and this is the only remaining one that is missing from the new class which it makes sense to copy over; most of the other traitlets that the old one has are unused, as far as I can tell.

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -344,13 +344,10 @@ class TerminalInteractiveShell(InteractiveShell):
                 self.showtraceback()
             else:
                 try:
-                    f = open(err.filename)
-                    try:
+                    with open(err.filename) as f:
                         # This should be inside a display_trap block and I
                         # think it is.
                         sys.displayhook(f.read())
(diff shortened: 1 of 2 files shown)


IPython/terminal/ptshell.py
```diff
@@ -317,5 +324,67 @@ def enable_gui(self, gui=None):
...
+            if not self._should_recompile(err):
+                return
+            try:
+                # may set last_syntax_error again if a SyntaxError is raised
+                self.safe_execfile(err.filename, self.user_ns)
+            except:
+                self.showtraceback()
+            else:
+                try:
+                    f = open(err.filename)
```
Carreau: context manager ?

takluyver: Why not. This is copy-pasted code from the old TIS class, so I don't want to change it too much, but putting this in a `with` is a simple improvement.

Issue #8606, opened 2015-07-03
Look at switching to Jedi for completion machinery


takluyver:
Our current completion machinery is relatively old and brittle. I've been meaning to refactor it, but I've realised that the Python completion library *Jedi* has [support for completing in REPLs](https://jedi.readthedocs.org/en/latest/docs/plugin-api.html#jedi.api.Interpreter). We can probably use that and get rid of our own completion machinery entirely.

One minor concern is performance: I tried using jedi-vim at one stage, and found it slowed the editor down quite a bit. Hopefully it's better now.

I'm interested in looking into this post 4.0.

minrk:
Yeah, I've heard a number of issues about jedi being slow, and requiring anything that interacts with it to do some async handling, which isn't a good fit for readline. It might be something we can handle better with prompt toolkit, though, so something to keep in mind as we try that.

I think if we can make any completions that IPython provides plugins to jedi, that would probably be an improvement for everyone.

takluyver:
@davidhalter: We're looking at using Jedi for completions in IPython. A couple of questions:

1. IPython has some extensions to Python syntax (e.g. `%magics`, `foo?` help). Is there a way to extend Jedi to accommodate these?  

2. We have some concerns about the speed of completions. Our current completion machinery is pretty basic, but it's usually close enough to instant that the user doesn't notice any delay. How quick is Jedi? How much leeway do the configuration options give us to adjust performance?

Thanks!

davidhalter:
@takluyver 
Sorry to delay this for a week. I was on holidays.

1. I guess controlling the exensions won't be a big issue, since you can just put a function in front of the Jedi calls to handle the extensions accordingly. 

2. Giving you a precise specification of how fast Jedi is, tends to be hard. Jedi is definitely optimized for speed and queries that take longer than 200 ms are worth a bug report. However, while most queries are really fast (< 50ms) you will have slower queries, especially after warm-up (generating the caches) and within extremely large codebases. 

Jedi has a ton of optimizations to make completing Python fast, among them:

- A lot of caches (imports, modules, call signatures and a lot them to improve the speed of just one completion)
- A parser that only parses parts of files when they are modified (probably doesn't improve the speed for REPL completions). 
- In REPL completion Jedi falls back to the default Python logic for REPL completion when possible.

[...]

PR #9375, opened 2016-04-04
Initial patch with Jedi completion (no function header description).

liukelly:


Carreau:
Thanks , this is a first part toward #8606 , @liukelly is working with me from BIDS.

IPython/core/completer.py
```diff
@@ -765,9 +775,59 @@ def magic_matches(self, text):
...
+            sys.path.pop(0)
+
+        with open('tmp.txt', 'w') as f:
+            print('before:', before, file=f)
+            for c in completions:
+                print('', file=f)
+                print('full_name:', c.full_name, file=f)
+                print('name:', c.name, file=f)
+                print('name_with_symbols:', c.name_with_symbols, file=f)
+                print('description:', c.description, file=f)
```
Carreau: That's for debug, we need to figure out why `description:` is always `module:`

IPython/core/completer.py
```diff
@@ -76,6 +76,13 @@
 from IPython.utils.py3compat import builtin_mod, string_types, PY3
 from traitlets import CBool, Enum
 
+try:
+    import jedi
+    import jedi.api.helpers
+    import jedi.parser.user_context
+except ImportError:
+    pass
```
Carreau: From in person discussion: handle the case where Jedi is not importable, by skipping the jedi codepath.

IPython/core/completer.py
```diff
@@ -765,9 +775,59 @@ def magic_matches(self, text):
...
+        if self.namespace is None:
+            import __main__
+            namespaces.append(__main__.__dict__)
+        else:
+            namespaces.append(self.namespace)
+        if self.global_namespace is not None:
+            namespaces.append(self.global_namespace)
+
+        # From Jedi docs
+        sys.path.insert(0, os.getcwd())
```
Carreau: we probably want to avoid manipulating path here.

Carreau:
So the description think seem to be a Jedi Bug, I opened an issue:

https://github.com/davidhalter/jedi/issues/713

Carreau:
Travis tests are failing, we should add jedi either as a required (or optional) dependency in `setup.py`. 

@takluyver do we make it optional or required ? If we make it required we can simplify the code in `completer.py` and actually fail on import error.

IPython/core/completer.py
```diff
@@ -765,9 +775,59 @@ def magic_matches(self, text):
...
+                no__name = lambda txt: not txt.startswith('__')
+            else:
+                # true if txt is _not_ a _ name, false otherwise:
+                no__name = lambda txt: not txt.startswith('_')
+            completion_text = filter(no__name, completion_text)
+
+        matches = [before + c_text for c_text in completion_text]
+        return matches
+
     def python_matches(self,text):
```
liukelly: To be deprecated.

IPython/core/completer.py
liukelly: It was from http://jedi.jedidjah.ch/en/latest/_modules/jedi/utils.html. It seems to work without it, so maybe we can take it out, but I don't know why it's there.

takluyver: I'm not sure why that Jedi code does it, but IPython already adds the CWD to `sys.path`, so I'm pretty sure we don't need to do it here.

takluyver:
Once we're satisfied that jedi produces good results, and we're willing to get rid of our own code, I'd say we should make it a regular dependency, not an optional thing.

We should make sure that it doesn't mess up completions with things like magics. We have our own machinery to do completions in such cases, but we should ensure that jedi doesn't do something bizarre with our extra syntax.

IPython/terminal/ptshell.py
```diff
@@ -49,8 +49,7 @@ def get_completions(self, document, complete_event):
         )
         start_pos = -len(used)
         for m in matches:
-            yield Completion(m, start_position=start_pos)
-
+            yield Completion(m, start_position=start_pos, display_meta=m[::-1])
```
liukelly: completion.py in Python Prompt Toolkit [[link](https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/completion.py)]

Carreau: Sweet ! Just for other information, the second column is the completion text reversed for now, as Jedi return a wrong value for now. We planned on adding extra info in this field later.

Carreau:
I've made some improvements for the greedy completion, so that things like `a[0].<tab>` works: https://github.com/liukelly/ipython/pull/1,

This will also need a rebase. 

@liukelly tell us if you are not comfortable with  rebase. 

until jedi is fixed, I think we should just not use the `, display_meta=m[::-1]` part, and open an issue to remind us to check that later.

liukelly:
Sorry for the delay, but I think I resolved all the conflicts.

Carreau:
> Sorry for the delay, but I think I resolved all the conflicts.

No worries. Will have a look.

Carreau:
Looks good to me. is the addition of `path.py` in `setup.py` for tests necessary ?

takluyver:
> is the addition of path.py in setup.py for tests necessary ? 

Probably a mix up in resolving some merge conflicts - that was added in master to make tests pass, and then the need for it was eliminated, so the dep was removed again.

Carreau:
> Probably a mix up in resolving some merge conflicts - that was added in master to make tests pass, and then the need for it was eliminated, so the dep was removed again.

Ok. 

I'll you review that before going to sleep . I know the sun never set on the British empire, but it's not a reason for you not to get some rest.

liukelly:
> is the addition of path.py in setup.py for tests necessary ?

Sorry, I think that came from liukelly/ipython@2f228c0d920cc88cb52b77ba104f1f9c8594f2f1. I don't think we need it.

Carreau:
> Sorry, I think that came from liukelly/ipython@2f228c0. I don't think we need it.
>  Show all checks

No problem, 
let's merge as it, get it in the wild to test. And I'll remove path.py in a subsequent PR. 

Thanks !

Issue #9376, opened 2016-04-06
missing diacritical mark after latex completion.


Carreau:
`F\dot<tab>` display `F` instead of `Ḟ` in cli, the dot is only missing visually. 

![diacritical](https://cloud.githubusercontent.com/assets/335567/14322936/735febd0-fbd3-11e5-99e0-08f633d74300.gif)

Ping @jonathanslenders , @takluyver

takluyver:
From my brief experiments, this appears to display incorrectly in prompt_toolkit (I can see the same issue when I enter it with copy and paste, so it's not specific to our completion machinery). Plain `input()` in Python shows it as expected.

jonathanslenders:
This should be a bug in Python:
```
In [2]: a='F'

In [3]: a
Out[3]: 'Ḟ'

In [4]: for c in a:
   ...:     print(c)
   ...:     
F
```

When iterating through this special character, it removes the dot...

Carreau:
it does not remove it for me, it prints it on the next line:
```
In [7]: for c in a:print(c)
F
 ̇
```

takluyver:
So, just for extra confusion, there are two ways to store the same symbol:

* `Ḟ` : decomposed, two characters
* `Ḟ` : composed, one character

PT appears to be fine with the composed character, but fall over with the decomposed one. This is true for me in ptpython as well as in IPython master.

jonathanslenders:
@Carreau: For me too, it should probably print it on the next line. (I only see whitespace, but I guess my terminal doesn't understand a separate dot.

@takluyver: Yeah, I just noticed. They are not the same F:

[python35] > hexdump wrong-f
0000000 b8e1 009e                              
0000003
jonathan @ arch-jonathan : /tmp
[python35] > hexdump correct-f
0000000 cc46 0087                              
0000003

Let me think how to solve this. Is there a way to convert decomposed characters into one composed character?

jonathanslenders:
I guess the reason that the dot is invisible in the prompt itself, is because prompt-toolkit didn't understood decomposed characters, and handles it as two individual characters. Then wcwidth tells us that the single dot has a width of 0, causing it to be hidden.

takluyver:
Where there is a composed equivalent, yes:

```
unicodedata.normalize('NFC', 'Ḟ')
```

takluyver:
(Github/my browser doesn't seem to display the decomposed form correctly either... unicode is hard!)

jonathanslenders:
[...]

PR #9378, opened 2016-04-06
Bugfix: when doing \dot completion. Compose unicode character from completion.

jonathanslenders:
See: https://github.com/ipython/ipython/issues/9376#issuecomment-206464016

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -4,7 +4,9 @@ from __future__ import print_function
 import os
 import sys
 import signal
+import unicodedata
 from warnings import warn
+from wcwidth import wcwidth
 
 from IPython.core.error import TryNext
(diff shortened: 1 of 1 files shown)


Carreau:
Works with `\dot`, but not with `\ddot` which is invisible...

Carreau:
Ok.. there is no combined form for `F\ddot`. Still improvement. Merging.

Carreau:
Thanks a lot for reactivity !

jonathanslenders:
Apparently, there is no composed form for \\ddot... Currently not sure yet how to handle that.

jonathanslenders:
Actually, many terminals don't even now how to handle that. lxterminal does, xterm doesn't at all, urxvt shows the dots too low.

takluyver:
Gnome terminal can handle the decomposed double-dot OK.

PR #9382, opened 2016-04-08
Switch embedded IPython over to prompt_toolkit shell

takluyver:
Manually tested with the checks in tools/tests/embed/


minrk:
yay!

PR #9385, opened 2016-04-08
Sphinxify docstrings 

Carreau:
Fix test on top of @SylvainCorlay's work.

Should close #9304

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1578,7 +1578,7 @@ class InteractiveShell(SingletonConfigurable):
         with self.builtin_trap:
             info = self._object_find(oname)
             if info.found:
-                return self.inspector._format_info(info.obj, oname, info=info,
+                return self.inspector._get_info(info.obj, oname, info=info,
                             detail_level=detail_level
                 )
             else:
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -91,7 +91,7 @@ try:
                 'text/html': sphx.sphinxify(doc, dirname),
                 'text/plain': doc
             }
-except:
+except ImportError:
     docformat = None
 
 #-----------------------------------------------------------------------------
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -75,6 +75,7 @@ from IPython.utils.strdispatch import StrDispatch
 from IPython.utils.syspathcontext import prepended_to_syspath
 from IPython.utils.text import (format_screen, LSString, SList,
                                 DollarFormatter)
+from IPython.utils.tempdir import TemporaryDirectory
 from traitlets import (Integer, Bool, CBool, CaselessStrEnum, Enum,
                                      List, Dict, Unicode, Instance, Type)
 from warnings import warn
@@ -85,11 +86,11 @@ try:
     import docrepr.sphinxify as sphx
 
     def docformat(doc):
-        dirname = tempfile.mkdtemp(prefix='ipython_sphinxify_')
-        return {
-            'text/html': sphx.sphinxify(doc, dirname),
-            'text/plain': doc
-        }
+        with TemporaryDirectory() as dirname:
+            return {
+                'text/html': sphx.sphinxify(doc, dirname),
+                'text/plain': doc
+            }
 except:
     docformat = None
 
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -81,6 +81,18 @@ from warnings import warn
 from logging import error
 import IPython.core.hooks
 
+try:
+    import docrepr.sphinxify as sphx
+
+    def docformat(doc):
+        dirname = tempfile.mkdtemp(prefix='ipython_sphinxify_')
+        return {
+            'text/html': sphx.sphinxify(doc, dirname),
+            'text/plain': doc
+        }
+except:
+    docformat = None
+
 #-----------------------------------------------------------------------------
 # Globals
 #-----------------------------------------------------------------------------
@@ -1539,7 +1551,7 @@ class InteractiveShell(SingletonConfigurable):
         info = self._object_find(oname, namespaces)
         if info.found:
             pmethod = getattr(self.inspector, meth)
-            formatter = format_screen if info.ismagic else None
+            formatter = format_screen if info.ismagic else docformat
             if meth == 'pdoc':
                 pmethod(info.obj, oname, formatter)
             elif meth == 'pinfo':
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -695,7 +695,8 @@ class Inspector(Colorable):
         """DEPRECATED. Compute a dict with detailed information about an object.
         """
         if formatter is not None:
-            warnings.warn('Inspector.info(formatter) keyword is deprecated as of IPython 5.0 and will have no effects.',
+            warnings.warn('The `formatter` keyword argument to `Inspector.info`'
+                     'is deprecated as of IPython 5.0 and will have no effects.',
                       DeprecationWarning, stacklevel=2)
         return self._info(obj, oname=oname, info=info, detail_level=detail_level)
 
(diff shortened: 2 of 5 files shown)


IPython/core/oinspect.py
```diff
@@ -541,118 +543,168 @@ def _format_fields(self, fields, title_width=0):
...
-            page.page(text)
-    
+        info = self._get_info(obj, oname, formatter, info, detail_level)
+        if info:
+            page.page(info)
+
     def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
+        """DEPRECATED. Compute a dict with detailed information about an object.
+        """
+        warnings.warn('Inspector.info is deprecated as of IPython 5.0',
```
Carreau: Insert here what shoudl be used instead.

Carreau:
Update tests to not use `Inspector.info` if deprecated.

IPython/core/oinspect.py
Carreau: There seem to be no public methods to replace this functionality, correct ?

SylvainCorlay: Actually maybe we should keep the original info. Originally I thought that this should not be public API at all.

Carreau: Ok, what about just issuing the warning if formatter is not `None`, and  decide on new API later ?

IPython/core/interactiveshell.py
```diff
@@ -75,12 +75,25 @@
...
+    import docrepr.sphinxify as sphx
+
+    def docformat(doc):
+        with TemporaryDirectory() as dirname:
+            return {
+                'text/html': sphx.sphinxify(doc, dirname),
+                'text/plain': doc
+            }
+except:
+    docformat = None
```
Carreau:
```
def docformat(doc):
            return {'text/plain': doc}
```

if issues ?

SylvainCorlay: `None` can be passed as a formatter already...

IPython/core/oinspect.py
```diff
@@ -694,7 +694,8 @@ def pinfo(self, obj, oname='', formatter=None, info=None, detail_level=0):
     def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
         """DEPRECATED. Compute a dict with detailed information about an object.
         """
-        warnings.warn('Inspector.info is deprecated as of IPython 5.0',
+        if formatter is not None:
+            warnings.warn('Inspector.info(formatter) keyword is deprecated as of IPython 5.0 and will have no effects.',
```
SylvainCorlay: I did not understand the wording right away. What about
The `formatter` keyword argument to `inspector.info` is deprecated as of IPython 5.0 and will have no effects.

IPython/core/oinspect.py
```diff
@@ -541,118 +543,169 @@ def _format_fields(self, fields, title_width=0):
...
+        info = self._get_info(obj, oname, formatter, info, detail_level)
+        if info:
+            page.page(info)
+
     def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
+        """DEPRECATED. Compute a dict with detailed information about an object.
+        """
+        if formatter is not None:
+            warnings.warn('Inspector.info(formatter) keyword is deprecated as of IPython 5.0 and will have no effects.',
+                      DeprecationWarning, stacklevel=2)
```
SylvainCorlay: I did not understand the wording right away. What about

```
'The `formatter` keyword argument to `Inspector.info` is deprecated as of IPython 5.0 and will have no effects.'
```

IPython/core/tests/test_oinspect.py
```diff
@@ -340,13 +340,13 @@ def test_calldef_none():
 @skipif(not py3compat.PY3)
 def test_definition_kwonlyargs():
     i = inspector.info(f_kwarg, oname='f_kwarg')  # analysis:ignore
-    nt.assert_equal(i['definition'], "f_kwarg(pos, *, kwonly)\n")
+    nt.assert_equal(i['definition'], "f_kwarg(pos, *, kwonly)")
```
SylvainCorlay: So I removed the `\n` in the end of definition because it is part of the formatting to me.

IPython/core/oinspect.py
Carreau: Done.

Carreau:
@SylvainCorlay Should we do documentation as a separate PR ?

Carreau:
@SylvainCorlay you should be able to push on my fork too if you want to push some updates.

SylvainCorlay:
We may want to way for notebook 4.2 to be released before we push this.

Cf bug fix here: https://github.com/jupyter/notebook/pull/1216

Carreau:
Good point.

Carreau:
`way` -> `wait` I suppose.

SylvainCorlay:
wait, @Carreau can release notebook!

Carreau:
Current pager highlight the source. Despite html beeing better in general, the behavior of the PR is worse than master. We should fix that first.

SylvainCorlay:
You mean that we should improve the pager in notebook to highlight python code snippets?

Carreau:
> You mean that we should improve the pager in notebook to highlight python code snippets?

Likely docrepr to highlight things. 
Have to check.

takluyver:
Sphinx is already hooked up to pygments highlighting, it's probably just a matter of telling it to assume that code blocks are Python unless specified otherwise.

Carreau:
> Sphinx is already hooked up to pygments highlighting, it's probably just a matter of telling it to assume that code blocks are Python unless specified otherwise.

docrepr on itself seem to do it (we would still need to embed the css for styling), but this PR seem to not embed the style for whatever reason. I have to figure out why.

SylvainCorlay:
Don't you think we should add this styling on the notebook side?

Carreau:
> this PR seem to not embed the style for whatever reason.

Only does not highlight the source **with pdoc** pinfo does set the right class on html.

Carreau:
> Don't you think we should add this styling on the notebook side?

We could ask docrepr to use the same class name as CodeMirror. But regardless we need to set some styles.

SylvainCorlay:
> We could ask docrepr to use the same class name as CodeMirror. But regardless we need to set some styles.

That could be enought to be in a mergeable state.

Carreau:
> That could be enought to be in a mergeable state.

Probably yes, I'd like `??` to also returned highlighted source as it's main use case. 
`pdoc` magic is broken too, as it does not not seem to handle a mimebundle in some places.

takluyver:
Is it possible to include the style definitions in the HTML we send to the pager?

ccordoba12:
> Is it possible to include the style definitions in the HTML we send to the pager?

As embedded css in the html page?

takluyver:
Yep. I'm thinking that we probably shouldn't add the pygments styles on the frontend, because the fact that we're sending HTML with pygments CSS classes is specific to this kernel. So the styles ought to come from the kernel with the HTML that uses them. That gets into the more general question of getting resources from the kernel that may be used in multiple displayed outputs, which we don't have a great answer for yet.

ccordoba12:
No problem, we can do that in docrepr using jinja2 :-)

@SylvainCorlay, do you have time to work on this? It should be an easy task ;-)

SylvainCorlay:
I can do this on Sunday or Monday. Before that I need a survival kit for a talk :)

ccordoba12:
Jajaja. Please ping me when you have something for review ;-)

Carreau:
Pushed some changes that make `??` be syntax highlighted.

Though I think that we should disable (for now) the html version of the pager as the h1, and the html now force you most of the time to scroll the pager to see what you are looking for. 

So I would **really** like to iterate on how it looks like, and take into account the compactness of the information conveyed.

Ping @ellisonbg to iterate on style.

SylvainCorlay:
Why are you closing this?

Carreau:
Sorry, I won't have the bandwidth to work on that now. Maybe later. But there is significant update that might have to be done, included in the notebook, if we want to  keep the compactness of the  docstring rendering in the pager. 

I worked a day of 2 with that merged in, and the HTML just take a significant amount of extra space, that kept me reaching for my mouse, and scrolling, it was too annoying. 

So while in principle it is an improvement, without careful redesign of the UI it decrease productivity.

PR #9387, opened 2016-04-08
Require path.py for tests

takluyver:
This was previously hidden because pickleshare was pulling in path.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -182,7 +182,7 @@ extras_require = dict(
     parallel = ['ipyparallel'],
     qtconsole = ['qtconsole'],
     doc = ['Sphinx>=1.3'],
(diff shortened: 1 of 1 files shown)


takluyver:
Merging so we get tests passing on master again.

minrk:
Do we require path.py directly, or just in tests?

takluyver:
It's a very soft dep for the main code - it's only used if you do:

```
In [1]: files = !ls
In [2]: files.p
```

However, there's a test that checks this functionality.

minrk:
ok, let's fix that to use pathlib, I suppose? Do we need to backport a setup.py patch adding path.py for 4.0, 4.1?

Issue #9314, opened 2016-03-10
[PromptToolkit] Reverse search Ctrl-C should first clearbuffer before dicmissing


Carreau:
cf comment on https://github.com/ipython/ipython/pull/9313

takluyver:
Closed by #9400

PR #9400, opened 2016-04-14
Have Ctrl-C in search mode clear query, if present

gforsyth:
As suggested in ipython/ipython#9314, when pressing Ctrl-C within the
search-bar, the current search terms will be cleared but the search
buffer will remain focused.  Pressing Ctrl-C when the search buffer is
empty will return focus to the default buffer.


takluyver:
Thanks! This seems to work nicely.

Issue #9379, opened 2016-04-06
TypeError on Shell Completion


brandonwillard:
I get the following error when attempting tab completion within `ipython` console:
```python
Python 2.7.11+ (default, Mar 30 2016, 21:00:42) 
Type "copyright", "credits" or "license" for more information.

IPython 5.0.0.dev -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: dException in thread Thread-7:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/prompt_toolkit/interface.py", line 742, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "/home/bwillar0/projects/python/ipython/IPython/terminal/ptshell.py", line 55, in gets
    m = unicodedata.normalize('NFC', m)
TypeError: normalize() argument 2 must be unicode, not str
```

Changing line 55 of `ptshell.py` to 
```python
m = unicodedata.normalize('NFC', unicode(m))
```
works fine, but I'm not sure if that's really the problem.  Any ideas?

takluyver:
Ping @jonathanslenders - this was a fix that was just made for something else.

I suspect that @brandonwillard 's suggested fix is right, but we'll need to use the `cast_unicode_py2` function from our py3compat module, because the built in `unicode` type is called `str` on Python 3.

jonathanslenders:
That's right. I was expecting that the native string type in IPythen was always unicode in Py2, str in Py3. How is it possible that me got an str it Py2?

takluyver:
We should always be working with unicode, but the completion machinery is pretty old and tricky code, so it's quite possible that some of the results coming back from it are still Python 2 `str`.

minrk:
There's a good chance somewhere the completion is forcing str on py2 for readline reasons. We should be forcing unicode somewhere now, though, before it returns from the completion API.

Carreau:
#9403 have (hopefully) fixed that, leaving open in case we hit more case that produce `str`s.

minrk:
Closing, since this seems to be closed by #9403. Reopen if we see this again.

PR #9403, opened 2016-04-15
Force completion results to be unicode.

Carreau:
Should fix some of #9379


Carreau:
note: mostly using `cast_unicode_py2`, but sometime was just a matter of adding a `u` prefix.

minrk:
It says fixes "some of" #9379. Should it be closed now, or is there more to do?

Carreau:
> or is there more to do

I tested most of it, but I would not be surprised to have missed a few branches that could generate `str`s.
So it "might" be completely fixed, you can still close, and reopen if we see people still complaining.

minrk:
I'll go with close & reopen, since we'll just forget about it if it's fixed and doesn't come up again.

PR #9408, opened 2016-04-19
use pathlib for utils.text.paths

minrk:
instead of path.py

For 4.2

cc @takluyver

### DIFF
--- a/IPython/utils/tests/test_text.py
+++ b/IPython/utils/tests/test_text.py
@@ -19,7 +19,11 @@ import random
 import sys
 
 import nose.tools as nt
-import path
+try:
+    from pathlib import Path
+except ImportError:
+    # Python 2 backport
+    from pathlib2 import Path
 
 from IPython.utils import text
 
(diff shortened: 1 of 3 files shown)


takluyver:
I'm generally in favour of this, but it is a user-facing change, because you will get a different type of object back when you do:

```
files = !ls
files.p
```

So maybe it should be 5.0 rather than 4.2?

minrk:
Either that, or we need to make `path.py` an explicit dependency of IPython.

takluyver:
I don't think it needs to be a dependency - it's only loaded if you do a specific operation that probably most people don't even know about, and even fewer use. We could make an error message that points people to `path.py`, but I'm inclined to just merge this without backporting.

minrk:
Then we're removing a feature in 4.1 from 4.2, aren't we?

takluyver:
No, I'd leave it in exactly as it is. With the change we made to pickleshare, users no longer get `path.py` installed automatically, but that's not changing with IPython versions,  and people who want that feature can still easily install `path.py`.

minrk:
Seems a bit odd to let something break just because we forgot to express a dependency that we always had when we pulled pickleshare out to a standalone package, but I'll go with you.

PR #9410, opened 2016-04-19
changelog for 4.2

minrk:
This should start the release process for 4.2. This is quite a small release, but the `-i` fix is important to get out there. The only thing I see in between us and release is deciding on the path.py/pathlib change in #9408.


minrk:
cc @willingc for release preparations

takluyver:
I think 4.2 is good to go :balloon:

willingc:
@minrk @takluyver This looks good to me. Thanks 🌻

PR #9417, opened 2016-04-22
only use backports.shutil_get_terminal_size on Python 2

minrk:
removes the backport dependency on Python 3, where it's not needed.

candidate for 4.2.1


Carreau:
For info, backported.

PR #9418, opened 2016-04-22
Fallback to :memory: on repeated failure to load history

### LINKED ISSUE #9412: infinite exception handling loop in case of corrupt SQLite3 db
Traceback here: https://gist.github.com/dwf/c659fdd50285b5a4885c10acc34b77e7

Version 4.1.2 installed via conda.

~~If anyone has a workaround it would be much appreciated.~~ I've worked around it by setting `IPYTHONDIR`, but it seems like this could be handled a bit more gracefully somehow.

minrk:
two consecutive failures to load history will result in falling back on :memory:

I also fixed a case where multiple consecutive failures to load would erase the original db with empty ones created after the first failure. The date is now in the corrupt-db filename.

closes #9412

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -71,11 +71,18 @@ else:
     class OperationalError(Exception):
         "Dummy exception when sqlite could not be imported. Should never occur."
 
+# use 16kB as threshold for whether a corrupt history db should be saved
+# that should be at least 100 entries or so
+_SAVE_DB_SIZE = 16384
+
 @decorator
 def catch_corrupt_db(f, self, *a, **kw):
     """A decorator which wraps HistoryAccessor method calls to catch errors from
     a corrupt SQLite database, move the old database out of the way, and create
     a new one.
+
+    We avoid clobbering larger databases because this may be triggered due to filesystem issues,
+    not just a corrupt file.
     """
     try:
         return f(self, *a, **kw)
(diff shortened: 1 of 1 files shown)


IPython/core/history.py
```diff
@@ -81,18 +80,29 @@ def catch_corrupt_db(f, self, *a, **kw):
...
+        self.log.error("Failed to open SQLite history %s (%s).", self.hist_file, e)
+        if self.hist_file != ':memory:':
+            if self._corrupt_db_counter > self._corrupt_db_limit:
+                self.hist_file = ':memory:'
+                self.log.error("Failed to load history too many times, history will not be saved.")
+            elif os.path.isfile(self.hist_file):
+                # Try to move the file out of the way
+                base,ext = os.path.splitext(self.hist_file)
+                now = datetime.datetime.now().isoformat().replace(':', '.')
+                newpath = base + '-corrupt-' + now + ext
```
takluyver: I appreciate the reason for this, but I'm not sure I like the solution - in a case like #9412, if the user doesn't configure a different location, it's presumably going to create two new, empty, corrupt database files for each IPython session, until the profile directory is littered with thousands of them.

Maybe it could just include the corrupt_db_counter - that way you preserve all the files from the last attempt to start IPython, but they're not accumulating.

minrk: That would mean starting IPython twice wipes out your history. In the NFS case prompting this, the db is not actually bad, SQLite is just failing to open it due to locking problems. So we should be making sure that we never overwrite the 'good' db, otherwise the '-corrupt' backup is pretty pointless.

takluyver: Right, but in the NFS case prompting this, it's failing to even create the database. So there's never a useful 'good' database worth preserving.

I'm not too concerned if starting IPython twice with a corrupt database wipes out your history - if we can't use that database, your history is effectively gone anyway, and I think ~no-one is going to take the time to recover it. I'd be happy if it kept it for a bit longer, or was smart enough not to move a database that it had just tried and failed to create. I'm just not happy about the ease with which this could create a great many useless files.

minrk: Updated to limit the no-clobber logic to files >= 16k, which should mean they have *some* content, at least. That means the thrashing where it never works should create at most one file.

takluyver:
Thanks. I'm happy with this now. Leaving it for a day or so in case other people want to look at it.

minrk:
Thanks for the review!

Carreau:
Thanks for that!

PR #9423, opened 2016-04-28
Make some of PTK option live updatable.

Carreau:
Done in a plane, should likely reread it later once awake.


Carreau:
Ok, updated a few functions names.

minrk:
Look okay to me, I'll let @takluyver have the final word.

takluyver:
Looks OK,  let's try it.

PR #9426, opened 2016-04-29
Promote keywords arguments completion.

Carreau:
Mostly when you are completing in a function call, it seem logical that
if keywords are available, they are the ones you are interested in.
## 

Pretty sure there is an issue where we discussed that with @takluyver


takluyver:
I remember discussing that as well, but I can't see the issue now. Maybe it was a mailing list thread or even an IRL discussion.

PR #9429, opened 2016-05-01
Back to explicit styles for the prompt tokens

takluyver:
I didn't notice these had changed in @Carreau's PR. The prompt number in particular was looking horrible with the default settings. I have reverted back to what we had before.

Wrong:

screenshot from 2016-05-01 13-20-59

Right:

screenshot from 2016-05-01 13-21-50


Carreau:
That's weird. I'll re-check the logic. Reverting for now.

PR #9432, opened 2016-05-02
Restore `%kill_embedded` for multiple IPython embed call.

### LINKED ISSUE #9063: Regression : %kill_embedded does not work (broken by #6840 )
The Magic [``%kill_embedded``](http://ipython.readthedocs.org/en/stable/api/generated/IPython.terminal.embed.html?highlight=kill_embedded#IPython.terminal.embed.EmbeddedMagics.kill_embedded) flags the embedded shell instance so that it is not started on subsequent calls of ``IPython.embed()``. However, fo this to work, we need to keep the instance in memory otherwise, we'll be flagging new instances all the time and never read the flags we've set.

In [``IPython/terminal/embed.py``](https://github.com/ipython/ipython/blob/master/IPython/terminal/embed.py#L262), we do several things :

* We save the instance that might be the current singleton instance of the ``InteractiveShell`` class
* We remove it from the singleton
* We get ``InteractiveShellEmbed``'s instance (created if existing) and launch the shell
* As soon as it's ended, we delete the instance
* We replace the original shell instance as the singleton instance.

All these things were done by #6840 in an effort to make embed nestable (but then it broke ``%kill_embeded``)

It's clear that the current behaviour makes it completely impossible for ``%kill_embedded`` to work as expected.

What would be preferred for a PR : to store the flag elsewhere or to improve #6840 to not destroy instances ?

Carreau:
Closes #9063.

Inspect frame and store the current file & line reference of the current
call. Calling kill_embed subsequently store this in a set of deactivated
reference.

The frame inspection is by default to 3, which is the correct number of
frame when using the `Import IPython; IPython.embed()` syntax, it will
be the wrong number of frames when creating an `InteractiveShellEmbed`
manually, leading to kill_embed to likely not work correctly.
Though I expect these case to be rare, and people doing so to be capable
of wrapping their custom call in a context_manager that set
`InteractiveShellEmbed._stack_frames` to the right value depending on
the context.

### DIFF
--- a/IPython/terminal/embed.py
+++ b/IPython/terminal/embed.py
@@ -39,8 +39,7 @@ class EmbeddedMagics(Magics):
         interfering again.
         """
 
-        kill = ask_yes_no("Are you sure you want to kill this embedded instance "
-                         "(y/n)? [y/N] ",'n')
+        kill = ask_yes_no("Are you sure you want to kill this embedded instance? [y/N] ",'n')
         if kill:
             self.shell.embedded_active = False
             print ("This embedded IPython will not reactivate anymore "
@@ -66,13 +65,25 @@ class InteractiveShellEmbed(TerminalInteractiveShell):
     dummy_mode = Bool(False)
     exit_msg = Unicode('')
     embedded = CBool(True)
-    embedded_active = CBool(True)
     should_raise = CBool(False)
     # Like the base class display_banner is not configurable, but here it
     # is True by default.
     display_banner = CBool(True)
     exit_msg = Unicode()
-    
+
+    _inactive_locations = set()
+
+    @property
+    def embedded_active(self):
+        return self._call_location_id not in InteractiveShellEmbed._inactive_locations
+
+    @embedded_active.setter
+    def embedded_active(self, value):
+        if value :
+            if self._call_location_id in InteractiveShellEmbed._inactive_locations:
+                InteractiveShellEmbed._inactive_locations.remove(self._call_location_id)
+        else:
+            InteractiveShellEmbed._inactive_locations.add(self._call_location_id)
 
     def __init__(self, **kw):
         
@@ -81,8 +92,13 @@ class InteractiveShellEmbed(TerminalInteractiveShell):
             warnings.warn("user_global_ns has been replaced by user_module. The\
                            parameter will be ignored, and removed in IPython 5.0", DeprecationWarning)
 
+        self._call_location_id =  kw.pop('_call_location_id', None)
+
         super(InteractiveShellEmbed,self).__init__(**kw)
 
+        if not self._call_location_id:
+            frame = sys._getframe(1)
+            self._call_location_id = '%s:%s' % (frame.f_code.co_filename, frame.f_lineno)
         # don't use the ipython crash handler so that user exceptions aren't
         # trapped
         sys.excepthook = ultratb.FormattedTB(color_scheme=self.colors,
@@ -298,7 +314,8 @@ def embed(**kwargs):
     if saved_shell_instance is not None:
         cls = type(saved_shell_instance)
         cls.clear_instance()
-    shell = InteractiveShellEmbed.instance(**kwargs)
+    frame = sys._getframe(1)
+    shell = InteractiveShellEmbed.instance(_call_location_id='%s:%s' % (frame.f_code.co_filename, frame.f_lineno), **kwargs)
     shell(header=header, stack_depth=2, compile_flags=compile_flags)
     InteractiveShellEmbed.clear_instance()
     #restore previous instance


IPython/terminal/embed.py
```diff
@@ -66,14 +65,27 @@ class InteractiveShellEmbed(TerminalInteractiveShell):
...
     
 
+    @property
+    def embedded_active(self):
+        return self._call_location_id not in InteractiveShellEmbed._inactive_locations
+
+    @embedded_active.setter
+    def embedded_active(self, value):
+        if value :
+            InteractiveShellEmbed._inactive_locations.remove(self._call_location_id)
```
takluyver: This should probably have a check if it's currently in the set - `set.remove()` throws KeyError if you try to remove something that's not there.

IPython/terminal/embed.py
```diff
@@ -66,14 +65,27 @@ class InteractiveShellEmbed(TerminalInteractiveShell):
...
     exit_msg = Unicode('')
     embedded = CBool(True)
-    embedded_active = CBool(True)
     should_raise = CBool(False)
     # Like the base class display_banner is not configurable, but here it
     # is True by default.
     display_banner = CBool(True)
     exit_msg = Unicode()
+
+    _inactive_locations = set({})
```
takluyver: The empty dict is superfluous - just use `set()`.

Carreau:
Done for both comment.

IPython/terminal/embed.py
```diff
@@ -83,6 +96,8 @@ def __init__(self, **kw):
 
         super(InteractiveShellEmbed,self).__init__(**kw)
 
+        frame = sys._getframe(InteractiveShellEmbed._stack_frames)
+        self._call_location_id = '%s:%s' % (frame.f_code.co_filename, frame.f_lineno)
```
takluyver: I don't think this stack depth will be right if the user is instantiating `InteractiveShellEmbed` themselves - I assume you've calibrated it for calling `embed()`? Doing anything with stack depth is super brittle.

Carreau: Yes that's why the PR description says: 

> The frame inspection is by default to 3, which is the correct number of
frame when using the Import IPython; IPython.embed() syntax, it will
be the wrong number of frames when creating an InteractiveShellEmbed
manually, leading to kill_embed to likely not work correctly.
Though I expect these case to be rare, and people doing so to be capable
of wrapping their custom call in a context_manager that set
InteractiveShellEmbed._stack_frames to the right value depending on
the context.

Carreau: Which is actually the message of the commit.

takluyver: Oops, sorry for not reading that.

Why not have `embed()` get its caller from the stack and pass it in to `InteractiveShellEmbed()` as an optional parameter. Then `InteractiveShellEmbed()` can do the right thing if it's not passed in.

Carreau: I can try to take a hash of the stack-trace, like `'-'.join('%s:%s' % (frame.f_code.co_filename, frame.f_lineno) for frame in sys.get_frames()`, which will be more unique depending on how you access the line where embed is called. Might be more robust. But I don't see how to deactivate a specific location permanently, cleanly another way.

takluyver: Also, if you're instantiating ISE directly, I think you can call the same instance in different locations. Should it deactivate the whole instance, or only the location it was called from?

Carreau: I'm in favor of location, and the issues with giving the caller is that the caller can likely change, and still be in the same location. I can add a `call_location` kwarg that take precedence on my logic.

takluyver: I think you may have forgotten to push?

takluyver: I think location makes sense, and I would add a `_call_location` kwarg which embed() passes in.

Carreau:
Like that ?

takluyver:
Yup, like that, but with valid syntax on Python < 3.5 ;-). I think you just need to put the single keyword arg before the `**kwargs`.

Also,  there are a couple of prints that look like they were just for debugging.

Carreau:
> Yup, like that, but with valid syntax on Python < 3.5 ;-)

So picky.

takluyver:
Me and my mate Travis, anal retentive code QA ;-)

Carreau:
> Me and my mate Travis, anal retentive code QA ;-)

Test are passing ( removed testing on py < 3.5 :-P )

takluyver:
> removed testing on py < 3.5 :-P

I am actually checking that you didn't do that, because, you know, QA :-P

IPython/terminal/embed.py
```diff
@@ -81,8 +92,16 @@ def __init__(self, **kw):
...
+        self._call_location_id =  kw.pop('_call_location_id', None)
+
         super(InteractiveShellEmbed,self).__init__(**kw)
 
+        if not self._call_location_id:
+            frame = sys._getframe(1)
+            print('compute caller location')
+            self._call_location_id = '%s:%s' % (frame.f_code.co_filename, frame.f_lineno)
+        else :
+            print('using ', self._call_location_id)
```
takluyver: Leftover debugging?

IPython/terminal/embed.py
```diff
@@ -81,8 +92,16 @@ def __init__(self, **kw):
             warnings.warn("user_global_ns has been replaced by user_module. The\
                            parameter will be ignored, and removed in IPython 5.0", DeprecationWarning)
 
+        self._call_location_id =  kw.pop('_call_location_id', None)
+
         super(InteractiveShellEmbed,self).__init__(**kw)
 
+        if not self._call_location_id:
+            frame = sys._getframe(1)
+            print('compute caller location')
```
takluyver: And here?

IPython/terminal/embed.py
Carreau: Rhah stupid `git add -p`, `git commit -a`

Carreau:
> Leftover debugging?

Amended.

PR #9433, opened 2016-05-03
Jedi completion

Carreau:
Finish up #9375, deactivate the 2 column completion ( with the second column showing types) as ti is broken in Jedi.


Carreau:
Should probably be tested a bit.

PR #9434, opened 2016-05-04
Pick prompt colors from themes, unless using 'default'

Carreau:
Hope that will suit both @takluyver & me. 

if color for prompt defined, use them. 
If not pick already existing color that match. 
If these are undefined, fallback on the current one, 

Unless it's default in which case we still have the current color scheme. 

Thoughts?

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -251,10 +251,23 @@ class TerminalInteractiveShell(InteractiveShell):
         We need that to add style for prompt ... etc. 
         """
         style_cls = get_style_by_name(name)
-        style_overrides = {
-            Token.Prompt: '#009900',
-            Token.PromptNum: '#00ff00 bold',
-        }
+        style_overrides = {}
+
+        tk = style_cls.styles.get( Token.Keyword )
+        if not style_cls.styles.get(Token.Prompt):
+            if tk:
+                style_overrides[Token.Prompt] = tk
+            else:
+                style_overrides[Token.Prompt] = '#009900'
+
+
+        tl = style_cls.styles.get( Token.Literal.Number)
+        if not style_cls.styles.get(Token.PromptNum):
+            if tl:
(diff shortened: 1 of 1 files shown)


takluyver:
I don't see why keyword and literal would make a good colour combination for the prompt in general. I'd be inclined to just fix our default in=green, out=red, and tell people how to configure it if they want something different.

Carreau:
I don't 'really care whether it's keyword/literal/number, I just want something that matches:

screen shot 2016-05-04 at 11 20 33

takluyver:
Pick two tokens at random from the selected theme? :-P

Several of those prompts look pretty ugly to me.

Carreau:
Sure, they're


 not meant for light BG, they don't look good if you dont' change prompt either: 

screen shot 2016-05-04 at 11 26 13

screen shot 2016-05-04 at 11 27 28
screen shot 2016-05-04 at 11 27 39

takluyver:
Could we pick one of the token colours from the theme, and then make bold and non-bold variants of it for the 'In[]:' and the number respectively?

Carreau:
> Pick two tokens at random from the selected theme? :-P

Yes, each time you draw the prompt :-) 

> Could we pick one of the token colours from the theme, and then make bold and non-bold variants of it for the 'In[]:' and the number respectively?

Likely, but `bold` does not always mean `bold`,  technically the ansi code mean emphase, 
so for example here is my terminal mapping, where left is "normal" color, and right is "bold" one. 

screen shot 2016-05-04 at 11 36 44

ALso if the token is already bold, you can't boldify it more.

takluyver:
> ALso if the token is already bold, you can't boldify it more. 

Right, I mean extract the colour (regexes FTW ;-) and make bold and non-bold versions.

Bold may not always mean bold, but it probably gets you at least two similar but slightly distinct styles, which will probably look better together than some random pair of colours from the theme.

PR #9437, opened 2016-05-05
Upgrade to prompt_toolkit 1.0.0.

jonathanslenders:
This merge requests upgrades prompt-toolkit to 1.0.0.

As said, there were a few backwards incompatible changes I had to do in the key bindings. (all backwards incompatible changes were done in a way that it would not crash, but the user would end up using Emacs instead of Vi bindings, for instance.) I think that we can consider prompt-toolkit to be stable at this point. 

It points to `'prompt_toolkit>=1.0.0,<2.0.0'`. I will ensure to do proper testing before pushing new releases.

For IPython, this release means that the decomposed unicode characters are properly displayed.
A full changelog is here:
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/CHANGELOG

Jonathan

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -16,14 +16,12 @@ from IPython.utils.process import abbrev_cwd
 from traitlets import Bool, CBool, Unicode, Dict, Integer
 
 from prompt_toolkit.completion import Completer, Completion
-from prompt_toolkit.enums import DEFAULT_BUFFER, SEARCH_BUFFER
-from prompt_toolkit.filters import HasFocus, HasSelection, Condition
+from prompt_toolkit.enums import DEFAULT_BUFFER, SEARCH_BUFFER, EditingMode
+from prompt_toolkit.filters import HasFocus, HasSelection, Condition, ViInsertMode, EmacsInsertMode
 from prompt_toolkit.history import InMemoryHistory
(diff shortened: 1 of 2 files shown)


IPython/terminal/ptshell.py
```diff
@@ -232,7 +230,10 @@ def _(event):
         self._style = self._make_style_from_name(self.highlighting_style)
         style = DynamicStyle(lambda: self._style)
 
+        editing_mode = EditingMode.VI if self.vi_mode else EditingMode.EMACS
```
takluyver: Are there more modes on the cards? We haven't released this yet, so we could still change from a boolean to a string to allow future extensions.

takluyver:
One question from looking at the diff, but as ever, thanks for the prompt release and helpful changes.

IPython/terminal/ptshell.py
jonathanslenders: In the future, possibly yes, but not in the near future. I wanted to have the flexibility, but I don't have the intention right now to add another set of key bindings. So far, also nobody asked me. Nano/Pico key bindings would be an example.

takluyver: OK. Since it's simple to do it now, I think I'll turn `vi_mode=True` into `edit_mode='vi'` in a separate PR.

PR #9439, opened 2016-05-05
Configure editing_mode instead of vi_mode

takluyver:
This makes life easier if prompt_toolkit gains any extra modes.

Ping @ivanov, who will presumably need to update his config for this. :-)


Carreau:
+1

PR #9440, opened 2016-05-05
Added display_completions_in_columns option.

jonathanslenders:
Added configuration option to get a multi-column completion menu.

This pull request should be fine, however I think there is a bug. For almost every completion, I get the following error. start_index=10 and cursor_pos=1. 
I think there is a bug at this line: https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L807 Not sure, but it could have to do that some completions are case insensitive.

``` python
  File "/usr/lib64/python3.5/threading.py", line 914, in _bootstrap_inner                     
    self.run()                                                                                
  File "/usr/lib64/python3.5/threading.py", line 862, in run                                  
    self._target(*self._args, **self._kwargs)                                                 
  File "/home/jonathan/git/python-prompt-toolkit/prompt_toolkit/interface.py", line 759, in rn
    completions = list(buffer.completer.get_completions(document, complete_event))            
  File "/home/jonathan/git/ipython/IPython/terminal/ptshell.py", line 49, in get_completions  
    cursor_pos=document.cursor_position_col                                                   
  File "/home/jonathan/git/ipython/IPython/core/completer.py", line 1265, in complete         
    self.matches.extend(self.python_jedi_matches(text, line_buffer, cursor_pos))              
  File "/home/jonathan/git/ipython/IPython/core/completer.py", line 827, in python_jedi_matchs
    return [trim_start(before + c_text) for c_text in completion_text]                        
  File "/home/jonathan/git/ipython/IPython/core/completer.py", line 827, in <listcomp>        
    return [trim_start(before + c_text) for c_text in completion_text]                        
  File "/home/jonathan/git/ipython/IPython/core/completer.py", line 809, in trim_start        
    assert start_index <  cursor_pos                                                          
AssertionError
```

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -134,6 +134,11 @@ class TerminalInteractiveShell(InteractiveShell):
     term_title = Bool(True, config=True,
         help="Automatically set the terminal title"
     )
+
+    display_completions_in_columns = Bool(False, config=True,
+        help="Display a multi column completion menu.",
+    )
+
     def _term_title_changed(self, name, new_value):
         self.init_term_title()
     
@@ -289,8 +294,8 @@ class TerminalInteractiveShell(InteractiveShell):
                 'get_prompt_tokens':self.get_prompt_tokens,
                 'get_continuation_tokens':self.get_continuation_tokens,
                 'multiline':True,
+                'display_completions_in_columns': self.display_completions_in_columns,
                 }
-
 
     def _update_layout(self):
         """


takluyver:
Thanks, we just landed a change in the completion machinery, so that might be a new issue (ping @Carreau @liukelly )

Carreau:
@jonathanslenders can you provide a pip freeze, and an example of failing completion, I cannot reproduce :-(

jonathanslenders:
It happens already after typing a lowercase `i` followed by `Tab`. It does not happen after typing an uppercase `I` followed by `Tab`.

Variables in that frame:
```
text='i'
path=''
dot=''
like='i'
completions=[<Completion: id>, <Completion: if>, <Completion: import>, <Completion: ImportError>, <Completion: ImportWarning>, <Completion: In>, <Completion: in>, <Completion: IndentationError>, <Completion: IndexError>, <Completion: input>, <Completion: int>, <Completion: InterruptedError>, <Completion: IOError>, <Completion: is>, <Completion: IsADirectoryError>, <Completion: isinstance>, <Completion: issubclass>, <Completion: iter>]

completions_text=['id', 'if', 'import', 'ImportError', 'ImportWarning', 'In', 'in', 'IndentationError', 'IndexError', 'input', 'int', 'InterruptedError', 'IOError', 'is', 'IsADirectoryError', 'isinstance', 'issubclass', 'iter']

completion='ImportWarning'
```

So I think the line ``start_index = completion.index(text)`` is wrong, because it looks for the lowercase `i` which does appear in the middle of the word.

```
ipython-genutils==0.1.0                                                                                                    
jedi==0.9.0                                                                                                                
pexpect==4.0.1                                                                                                             
pickleshare==0.7.2                                                                                                         
prompt-toolkit==1.0.0                                                                                                      
ptyprocess==0.5.1                                                                                                          
Pygments==2.1.3                                                                                                            
simplegeneric==0.8.1                                                                                                       
six==1.10.0                                                                                                                
traitlets==4.2.1                                                                                                           
wcwidth==0.1.6                                                                                                             
wheel==0.24.0  
```
Python 3.5.1

After some experimenting, it looks completely unrelated to this pull request.

Carreau:
Ok, can reproduce, will track down.

Carreau:
So I have another bug that appear between PTK 01304ec08067787f515f079c82b59819a6d94690 and 1.0.0 , line buffer is never clear on submit and  completer never dismiss also:

bug

Carreau:
This is the first bad commit. 

```
git bisect bad
b5b1a16c8143b3430f7ad6c668c37cab8d919b99 is the first bad commit
commit b5b1a16c8143b3430f7ad6c668c37cab8d919b99
Author: Jonathan Slenders <jonathan@slenders.be>
Date:   Thu May 5 13:36:17 2016 +0200

    Don't reset the current buffer anymore by default in CommandLineInterface.run(). This was very counter-intuitive.

:040000 040000 205a6b79219345035070b4cf0579a993b36180b5 6eb6842972c037d7f769bd84ff980fc55ffd9094 M	prompt_toolkit
```

jonathanslenders:
Yes, I know about that change, but I though that this was fixed in IPython.
We pass `reset_current_buffer=True` here: https://github.com/ipython/ipython/blob/master/IPython/terminal/ptshell.py#L303

Carreau:
> Yes, I know about that change, but I though that this was fixed in IPython.
> We pass reset_current_buffer=True here: https://github.com/ipython/ipython/blob/master/IPython/terminal/ptshell.py#L303

My bad, I did not pull since lunch.

PR #9441, opened 2016-05-05
Fix completions for PTK 1.0

Carreau:
Hotfix for #9440, that we should look at in in more details. 

Congratulation @jonathanslenders on PTK 1.0.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -803,10 +803,13 @@ class IPCompleter(Completer):
 
         def trim_start(completion):
             """completions need to start with `text`, trim the beginning until it does"""
-            if text in completion and not (completion.startswith(text)):
-                start_index = completion.index(text)
+            ltext = text.lower()
+            lcomp  = completion.lower()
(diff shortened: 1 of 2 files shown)


fperez:
I can confirm that this fixes the problem. Thanks for the quick turnaround!

fperez:
I've merged it as-is, since the code looks clean and it does fix a problem that made some completions not work at all. If a more refined fix is required, further PRs welcome.

Thanks @Carreau!

anntzer:
Attribute completions (`obj.<tab>`) seem to be broken for me.

Carreau:
> Attribute completions (obj.<tab>) seem to be broken for me.

Can you give a more detailed example ?

screen shot 2016-05-09 at 12 04 06

Do you have Jedi installed ?

anntzer:
Oh, I didn't realize that jedi was a dependency for object completion now (AFAICT it wasn't one until the update to PTK 1.0).  Sorry for the noise.

takluyver:
The Jedi completion stuff happened to land around the same time as the PTK 1.0 update - they're not particularly related.

PR #9444, opened 2016-05-06
Add %%js alias for %%javascript

jdfreder:
`%%javascript` is a bit long to type, and not the first thing that comes to mind when looking for a JS cell magic.  Here I added a `%%js` magic which is an alias for the long named cousin.

Hey, where's our `%%hypertextmarkuplanguage` magic??? :trollface:

### DIFF
--- a/IPython/core/magics/display.py
+++ b/IPython/core/magics/display.py
@@ -25,24 +25,29 @@ from IPython.core.magic import  (
 @magics_class
 class DisplayMagics(Magics):
     """Magics for displaying various output types with literals
-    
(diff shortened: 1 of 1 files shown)


Carreau:
+1 for signing commit, I need to restart using GPG. Want to update doc to say that it's an alias of `%%javascript` ? 

You can say no.

Carreau:
> +1 for signing commit, I need to restart using GPG. Want to update doc to say that it's an alias of %%javascript ?
> 
> You can say no.

Actually scratch that, I merge and will issue an extra PR.

PR #9452, opened 2016-05-10
MNT: only try to use mpl.use if safe

tacaswell:
If the library you are documenting sets the backend and imports
mpl this `use` will spam the doc build with warnings.  This guards
the call the same way it is done above.


takluyver:
Thanks!

PR #9453, opened 2016-05-11
Make event triggering robust to (un)registration.

### LINKED ISSUE #9447: Event callbacks which register/unregister other callbacks lead to surprising behavior
tl;dr: the loop in [this line](https://github.com/ipython/ipython/blob/1d63c56f6d69d9f4cc47adcb87a7f80744a48260/IPython/core/events.py#L72) isn't robust to `register`/`unregister`. What's the desired behavior?

### The problem

We ran into a funny situation in colaboratory, involving `post_run_cell` callbacks which need to potentially add or remove other `post_run_cell` callbacks. Here's a really simplified version:

```python
ip = get_ipython()
ev = ip.events

def func1(*unused_args):
    print 'invoking func1'
    ev.unregister('post_run_cell', func1)
    ev.register('post_run_cell', func3)

def func2(*unused_args):
    print 'invoking func2'
    ev.unregister('post_run_cell', func2)

def func3(*unused_args):
    print 'invoking func3'

ev.register('post_run_cell', func1)
ev.register('post_run_cell', func2)
```

In principle, this should invoke the three functions in order. In reality, it invokes `func1` and `func3`, but *not* `func2`. Even worse, at the end of this cell's execution, the list of registered callbacks is `[func2, func3]`, which is *not* the list of callbacks we saw execute. So `func2`, the only function that stays in the list of callbacks the whole time, is the only one we don't execute. 😉 

The cause is easy to see after checking out [this line](https://github.com/ipython/ipython/blob/1d63c56f6d69d9f4cc47adcb87a7f80744a48260/IPython/core/events.py#L72): 
* on the first iteration of the loop, `func1` is our callback, and `[func1, func2]` is our list of callbacks. `func1` executes, and our list of callbacks is now `[func2, func3]`.
* the next loop iteration picks up at the second element of the list, namely `func3`. unfortunately, we've now skipped `func2`. sadface.

The lesson is, of course, never mutate a list as you iterate over it.

## Potential solutions

I'm happy to send a PR, but first, I want to make sure we're on the same page about what the **desired** semantics are here. 

I think we want to be flexible: given that the only exposed operations for callbacks are remove and append (and thankfully [**not** reset](https://github.com/ipython/ipython/commit/4043b271fee4f6c36c99c8038018d54cd86b94eb)), we can ensure
1. we execute new callbacks appended to the list, and 
2. we skip removed callbacks if we haven't already executed them.

Other options all involve preventing this behavior. We could freeze state by instead making a copy of the list before we iterate over it, whence any modifications would only be visible on the next triggering of the event. Or we could add code to completely disallow modifications to the currently-executing list of callbacks for the duration of `trigger`, meaning a user would get an error if they tried.

## Our use case

A common pattern we've used is "one-shot callbacks" -- a callback which, when invoked, removes itself from the list of callbacks. We use this for some forms of output manipulation, as well as occasionally to do one-time cleanup that needs to happen after the user's current code completes. 

With the proposed solution, this is easy -- the body of a callback just includes a call to deregister itself. This currently works great, unless the one-shot callback is not the last callback in the list. (The example above was distilled from a variant of this -- a one-shot callback was registering another one-shot callback.) It also allows for forcing a callback to be idempotent -- register as many times as you'd like, and the first one can deregister any copies it finds.

With either of the other solutions, we're stuck basically inventing our own event system, and using it to handle cleanup.

craigcitro:
Event callbacks can register or unregister new callbacks for the same event
while executing, and the previous triggering implementation allowed for event
callbacks to be inadvertently skipped.

The fix is to make a copy of the list of callbacks before executing any of
them. With this change, the resulting semantics are simple: any callbacks
registered before triggering are executed, and any new callbacks registered
are only visible at the next triggering of the event.

Note that this could potentially break existing callers who expected
newly-appended callbacks were immediately executed.

Fixes #9447. Originally based on a patch by @marksandler2.

PTAL @takluyver

### DIFF
--- a/IPython/core/events.py
+++ b/IPython/core/events.py
@@ -69,7 +69,7 @@ class EventManager(object):
         Any additional arguments are passed to all callbacks registered for this
         event. Exceptions raised by callbacks are caught, and a message printed.
         """
-        for func in self.callbacks[event]:
+        for func in self.callbacks[event][:]:
             try:
                 func(*args, **kwargs)
             except Exception:
--- a/IPython/core/tests/test_events.py
+++ b/IPython/core/tests/test_events.py
@@ -30,3 +30,25 @@ class CallbackTests(unittest.TestCase):
         self.em.register('ping_received', cb)
         with tt.AssertPrints("Error in callback"):
             self.em.trigger('ping_received')
+
+    def test_unregister_during_callback(self):
+        invoked = [False] * 3
+        
+        def func1(*_):
+            invoked[0] = True
+            self.em.unregister('ping_received', func1)
+            self.em.register('ping_received', func3)
+
+        def func2(*_):
+            invoked[1] = True
+            self.em.unregister('ping_received', func2)
+
+        def func3(*_):
+            invoked[2] = True
+            
+        self.em.register('ping_received', func1)
+        self.em.register('ping_received', func2)
+
+        self.em.trigger('ping_received')
+        self.assertEqual([True, True, False], invoked)
+        self.assertEqual([func3], self.em.callbacks['ping_received'])
--- a/docs/source/whatsnew/pr/incompat-event-triggering.rst
+++ b/docs/source/whatsnew/pr/incompat-event-triggering.rst
@@ -1,0 +1,7 @@
+Update IPython event triggering to ensure callback registration and
+unregistration only affects the set of callbacks the *next* time that event is
+triggered. See :ghissue:`9447` and :ghpull:`9453`.
+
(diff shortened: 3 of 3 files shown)


craigcitro:
I should say -- I don't know if this merits a note in the `whatsnew/pr` dir, but happy to add one if so.

takluyver:
Thanks Craig. Yes, it would be good to have a whatsnew note. :-)

craigcitro:
Added a `whatsnew` note -- PTAL!

takluyver:
Thanks Cragi

PR #9457, opened 2016-05-12
Do not use `is` for string comparison

ThiefMaster:
Just spotted this; dunno if it's done anywhere else. But using `is` for strings (or numbers) is something that should pretty much never be done.

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -265,7 +265,7 @@ class TerminalInteractiveShell(InteractiveShell):
             Token.Prompt: '#009900',
             Token.PromptNum: '#00ff00 bold',
         }
-        if name is 'default':
+        if name == 'default':
             style_cls = get_style_by_name('default')
             # The default theme needs to be visible on both a dark background
             # and a light background, because we can't tell what the terminal


ssanderson:
On latest master this is the only case I see of this:

```
git --no-pager grep -n -e is\ \[\'\\\"\] -- 
IPython/core/completer.py:939:        # e.g. for "foo (1+bar(x), pa<cursor>,a=1)", the candidate is "foo"
IPython/core/debugger.py:287:        # Set the prompt - the default prompt is '(Pdb)'
IPython/core/magics/execution.py:216:          default sorting key is 'time'.
IPython/core/magics/namespace.py:302:        indicates that the type name for strings is 'str'.
IPython/core/magics/osm.py:421:            raise UsageError("usage is 'set_env var=val'")
IPython/lib/display.py:509:            # Working on a platform where the path separator is "\", so
IPython/sphinxext/ipython_directive.py:38:    The default is 'In [%d]:'. This expects that the line numbers are used
IPython/sphinxext/ipython_directive.py:42:    default is 'Out [%d]:'. This expects that the line numbers are used
IPython/sphinxext/ipython_directive.py:696:                    # added in `process_input`, as it is '' and would introduce
IPython/terminal/ptshell.py:268:        if name is 'default':
docs/source/config/details.rst:90:      strings). This 'scheme' is thus fully safe to use in any terminal.
docs/source/interactive/reference.rst:122:command is 'shadowed' by a variable, you will need the explicit ``%`` prefix to
docs/source/whatsnew/version2.0.rst:325:  ``bbox_inches``, which is 'tight' by default, or set the quality of JPEG figures.
examples/IPython Kernel/Beyond Plain Python.ipynb:1186:    "The default `%xmode` is \"context\", which shows additional context but not all local variables.  Let's restore that one for the rest of our session."
tools/git-mrb:8:If 'onto' branch isn't given, default is 'master'.
tools/git-mrb:10:If 'upstream' repository isn't given, default is 'origin'.
tools/tests/Markdown Pandoc Limitations.ipynb:1379:       "mdcell.set_text('\\nThis is \\\\LaTeX \\\\bf{bold} and \\\\emph{italic} text.\\n');\n",
tools/tests/Markdown Pandoc Limitations.ipynb:1406:       "<div style='display: inline-block; width: 30%; vertical-align: top;'><div style='background: #AAFFAA; width: 100%;'>NBConvert Latex Output</div><pre class='prettyprint lang-tex' style='background: #EEFFEE; border: 1px solid #DDEEDD;'><xmp>This is \\LaTeX \\bf{bold} and \\emph{italic} text.</xmp></pre></div><div style='display: inline-block; width: 2%;'></div><div style='display: inline-block; width: 30%; vertical-align: top;'><div style='background: #FFAAAA; width: 100%;'>NBViewer Output</div><div style='display: inline-block; width: 100%;'><p>This is </p></div></div>"
tools/tests/Markdown Pandoc Limitations.ipynb:1431:    "This is \\LaTeX \\bf{bold} and \\emph{italic} text.\n",
```
All the rest of the hits there are either in comments or docs.

ssanderson:
More background on why this `is` shouldn't be used for equality against literals:

CPython interns all string literals less than some fixed length.  This means that, generally speaking, this will always work:

```
In [6]: 'default' is 'default'
Out[6]: True
```

There is no guarantee, however, that dynamically-constructed strings will be the same object as a literal:

```
In [7]: ''.join(list('default'))
Out[7]: 'default'

In [8]: ''.join(list('default')) is 'default'
Out[8]: False
```

Carreau:
D'oh, stupid me.

takluyver:
Thanks both - yes, this is a bad idea, and it snuck under the radar here.

Carreau:
> There is no guarantee, however, that dynamically-constructed strings will be the same object as a literal:

Watch out, in **IPython REPL** the matching of identity is different than in the normal CPython REPL and in python script as we split the AST, and evaluate each node independently. So things might be different:

```
$ ipython
...
In [1]: a=300; a is 300
Out[1]: False

In [2]: exit()

$ python
...
>>> a=300; a is 300
True
>>>
```

Works for values outside of `[-5:256]`, otherwise the literal will still be the same.

ssanderson:
> Works for values outside of [-5:256], otherwise the literal will still be the same.

Interesting.  I assume this is in vanilla Python, the statements are in the same code object during compilation, which means you're guaranteed to have identity sharing because they're doing  LOAD_CONSTs of the same index, whereas in IPython, compiling the statements separately breaks that guarantee?

takluyver:
Yep, I think that's right. We do it like that so that we can display the result of a naked expression only when it's at the end of a cell - this expression has to be compiled in 'interactive' mode rather than 'exec' mode.

Carreau:
It also depends whether the statement are on the same line, separated by `;` or on different lines IIRC.

PR #9461, opened 2016-05-16
Don't catch failure to import Jedi

takluyver:
Jedi is a dependency now - if it isn't installed, we should fail clearly rather than silently skipping completions.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -77,13 +77,9 @@ from IPython.utils.process import arg_split
 from IPython.utils.py3compat import builtin_mod, string_types, PY3, cast_unicode_py2
 from traitlets import CBool, Enum
 
-try:
-    import jedi
-    import jedi.api.helpers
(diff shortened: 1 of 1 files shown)


takluyver:
I also discarded the `use_jedi_completions` configurable - I think previously this was used to switch between jedi completions and our built in completions. But now it switches between jedi completions and nothing, which doesn't seem very useful.

Carreau:
I'm thinking of reverting the jedi stuff, there are many completions that are wrong now.

takluyver:
I'll try to spend some time figuring out what's going wrong. Working out that I didn't have Jedi installed on this machine was a first step. ;-)

Carreau:
Ok, fair enough, I think the real need is to update IPython's completion API. Which I don't want to do for 5.0. I would prefer to revert jedi for 2 month, get 5.0 out, and work on 6.0.  A change of API is too risky for a LTS.

PR #9465, opened 2016-05-16
Update IPython/core/profileapp.py to use new Traitlets API.

Carreau:


Carreau:
Oops, just realized that `help=` should not be in `tag()`

PR #9466, opened 2016-05-16
Update IPython/core/interactiveshell.py to use new traitlets API.

Carreau:
[need to update this to use `@observe` also]

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -76,7 +76,7 @@ from IPython.utils.syspathcontext import prepended_to_syspath
 from IPython.utils.text import (format_screen, LSString, SList,
                                 DollarFormatter)
 from traitlets import (Integer, Bool, CBool, CaselessStrEnum, Enum,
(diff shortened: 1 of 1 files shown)


Carreau:
> [need to update this to use @observe also]

Done.

IPython/core/interactiveshell.py
```diff
@@ -283,49 +283,55 @@ def _exiter_default(self):
...
-    prompt_out = Unicode('Out[\\#]: ', config=True,
+    prompt_out = Unicode('Out[\\#]: ').tag(config=True,
         help="Deprecated, will be removed in IPython 5.0, use PromptManager.out_template")
-    prompts_pad_left = CBool(True, config=True,
+    prompts_pad_left = CBool(True).tag(config=True,
         help="Deprecated, will be removed in IPython 5.0, use PromptManager.justify")
-    
+
+   
+    @observe('prompt_trait',
```
takluyver: I don't think there's a `prompt_trait` trait, I think that's just what I called the method.

PR #9470, opened 2016-05-17
Update `ptshell.py` to use new traitlets API & rename unnamed callback.

Carreau:
I prefer to have named functions instead of `_` my guess is that if there is a crash one day, we'll have an explicit traceback with a function name. 

I'm ok removing this commit if it does not please someone.

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -13,7 +13,7 @@ from IPython.core.interactiveshell import InteractiveShell
 from IPython.utils.py3compat import PY3, cast_unicode_py2, input
 from IPython.utils.terminal import toggle_set_term_title, set_term_title
 from IPython.utils.process import abbrev_cwd
-from traitlets import Bool, CBool, Unicode, Dict, Integer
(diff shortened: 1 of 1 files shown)


IPython/terminal/ptshell.py
```diff
@@ -223,13 +224,13 @@ def cursor_in_leading_ws(cli):
...
                                    ))
-        def _(event):
+        def _indent_buffer(event):
             event.current_buffer.insert_text(' ' * 4)
 
         # Pre-populate history from IPython's history database
         history = InMemoryHistory()
         last_cell = u""
-        for _, _, cell in self.history_manager.get_tail(self.history_load_length,
+        for __, ___, cell in self.history_manager.get_tail(self.history_load_length,
```
takluyver: ?

Carreau: My linter complain that each of those overwrite underscore defined
previously and which was unused.

takluyver: I'd submit that you need a better linter ;-). But that doesn't really matter.

Issue #9331, opened 2016-03-17
Auto-rewrite when autocall is on miscomputes the number of dashes to print and then messes output prompt positioning


fperez:
Not sure what logic is doing this under the hood, but the problem is pretty obvious:

```
In [1]: autocall 
Automatic calling is: Smart

In [2]: print 1
-> print(1)
1

In [3]: print 1
------------> print(1)
1

In [4]: print 1
-----------------------> print(1)
1
```
On each successive call, more and more dashes get printed (and they aren't the right number even on the first call).  This is certainly a new bug, though I haven't bisected it yet.

Furthermore, once this has happened, the output prompt ends up mis-positioned:

```
In [10]: 3
  Out[10]: 3
```

My `sys_info`:

```
In [9]: print(sys_info())
{'commit_hash': 'aa586fd',
 'commit_source': 'repository',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/fperez/dev/ipython/ipython/IPython',
 'ipython_version': '5.0.0.dev',
 'os_name': 'posix',
 'platform': 'Darwin-15.3.0-x86_64-i386-64bit',
 'sys_executable': '/Users/fperez/usr/conda/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, '
                '11:24:55) \n'
                '[GCC 4.2.1 (Apple Inc. build 5577)]'}
```

takluyver:
Probably related to the prompt_toolkit work, as that has disconnected rewrite prompts from input prompts. See also: #9282.

takluyver:
Surprisingly, it looks like it was not the prompt_toolkit work but #8738 that broke this. Interim fix coming up...

PR #9471, opened 2016-05-17
prompt_toolkit compatible prompt machinery

takluyver:
Fixes gh-9331 (roughly)

The input prompts are now displayed through prompt_toolkit, so this code doesn't know the width it's trying to match.

@jonathanslenders I don't suppose there's a public API for us to get the width of the prompt area from the last run of the prompt_toolkit `CommandLineInterface`?


Carreau:
Failure seem real, but I guess we just need remove the test.

jonathanslenders:
@takluyver Sorry, no there's no public API for that. (Right now, not sure how I would ever expose this.)

takluyver:
OK, thanks - I knew it was a long shot. It should be easier to do this on our side, since we no longer have the ANSI colouring stuff embedded in a string.

takluyver:
This grew a bit: it now removes the `PromptManager` machinery, and adds a new `Prompts` class in the terminal package. Prompts are no longer configurable as template strings (I already broke it for input prompts when switching to prompt_toolkit, but this removes that config system entirely). My idea is that if you want to customise prompts, you should subclass the new `Prompts` class and redefine the prompt generation methods, and then set:

```python
shell.prompts = MyPrompts(shell)
```

This allows more flexibility, while eliminating all the predefined prompt elements we had for string templating.

Carreau:
> This grew a bit: it now removes the PromptManager machinery, and adds a new Prompts class in the terminal package. Prompts are no longer configurable as template strings (I already broke it for input prompts when switching to prompt_toolkit, but this removes that config system entirely). My idea is that if you want to customise prompts, you should subclass the new Prompts class and redefine the prompt generation methods, and then set:
> 
> shell.prompts = MyPrompts(shell)
> This allows more flexibility, while eliminating all the predefined prompt elements we had for string templating.

No objections with this change. @asmeurer  you are a custom-prompt user right ? I think your input might be welcomed.

asmeurer:
Is this related to https://github.com/ipython/ipython/issues/8724?

asmeurer:
I don't have time to look at this in-depth. This is the code I use to customize my prompt (among other things) https://github.com/asmeurer/iterm2-tools/blob/master/iterm2_tools/ipython.py. Can you indicate how that code would change with this? That file also has some comments with some various issues I found with IPython, if that helps any.

asmeurer:
Specifically the issue of #8724:

> The more advanced fix (taking into consideration that a dynamic prompt might have a different number invisible characters each time it is displayed) is awkward with the existing API, which implicitly assumes that the number invisible characters is the same for any given template. 

Does this address that?

takluyver:
This should address that, in that prompts are now generated as a list of token tuples, with a style and a piece of text, rather than as a single string containing ANSI escapes. So we no longer need to deal with 'invisible characters'.

Custom prompt stuff will need to be totally rewritten for IPython 5; I don't see any way round that with the switch to prompt_toolkit. See this section of the diff for the new Prompts class API that you'd have to supply: https://github.com/ipython/ipython/pull/9471/files#diff-9c5b89476acdb54178398f461c809c26R9

asmeurer:
So I should implement the invisible iTerm2 control sequences as tokens, with empty text?

jonathanslenders:
@asmeurer Not sure, but I think if this is an escape-sequence, and passed into prompt-toolkit, then it will probably be escaped and displayed as something like ``^[``. (It expects plain text.)
We are talking about this, right?: https://github.com/asmeurer/iterm2-tools/blob/master/iterm2_tools/shell_integration.py#L101

If so, I think we need a way in IPython to send these sequences before and after calling the prompt function.

takluyver:
There is already a configurable `InteractiveShell.separate_in` which is a string printed before starting the prompt. There's nothing equivalent for after the prompt, but you could use the `pre_run_cell` [event](http://ipython.readthedocs.io/en/stable/config/callbacks.html).

asmeurer:
I specifically need to send a sequence after the prompt but before the user-entered code. I also need to do a separate sequence after the user-entered code but before the output, for which I use [pre_execute](https://github.com/asmeurer/iterm2-tools/blob/06198b46b0ccc6502b2c12e3f08c8310dc6d4ced/iterm2_tools/ipython.py#L117).

takluyver:
I'm not sure how 'after the prompt but before user entered code' works with prompt_toolkit - my understanding is that it can update the multiline prompt area and user input area as you're typing.

jonathanslenders:
Honestly, right now there is no way to do this. Any VT escape sequence will be escaped, there is no way to insert this in the output.

There is this issue: https://github.com/jonathanslenders/python-prompt-toolkit/issues/148
But it didn't make it on my list because of not much interest and only working on iTerm2.

However, one workaround that could possibly work is printing literally everything before the IPython prompt, just like this: ``'\033]133;A\aIn[1]:\033]133;B\a\r'.
So print BEFORE_PROMPT, followed by the actual prompt (but not through prompt-toolkit), then print AFTER_PROMPT, then print carriage return. prompt-toolkit should then overwrite this prompt, but Iterm2 should now  know about these positions.

asmeurer:
If you're talking about "continuation" prompts or any additional prompt that is shown after you start typing, those aren't supported. Basically, it should do

{before_prompt}prompt >{after prompt} \<user code here\>

and if additional "prompts" are generated after the user starts typing code, those will just be considered to be part of the user input.

Carreau:
We should test this branch (that needs rebase) with https://github.com/jonathanslenders/python-prompt-toolkit/pull/299

Carreau:
Figured things out and sent a fix to PTK. 
I'll try to write up documentation for that later and send it to you @takluyver 
Monday is holidays here in the US, so might be a slow day.

Carreau:
Sent a PR to improve/remove & docs :  https://github.com/takluyver/ipython/issues/17

takluyver:
Merged @Carreau's doc changes and rebased. Then merged @Carreau's doc changes again, because I rebased without pulling ;-)

Carreau:
> Merged @Carreau's doc changes and rebased. Then merged @Carreau's doc changes again, because I rebased without pulling ;-)

Nobody's perfect. 

I think I'm +1 one that, with #9515 merged, we might be able to send a fix to @asmeurer iterm2-tools to not break (though it will still need ptk to get a new release for al to work).

Carreau:
Ok, merging this and refining on master if necessary.

PR #9473, opened 2016-05-17
Adopt traitlet 4.2 API.

Carreau:
I actually didn't saw @minrk PR when starting to issue mine, so This is Min's PR, rebased, plus some extra fixes. (I don't expect him to work today as May 17th in Norway is the July 4th of USA. (I knew he had a time machine & a teleporter).


Carreau:
removed WIP.

Issue #8927, opened 2015-10-26
Make it easier to set `suppress_errors=False` in IPython.core.Application


ssanderson:
IPython.core.application swallows exceptions raised during execution of config files unless `suppress_errors` is set to `False`.  When deploying an IPython server, I don't want config errors to be masked, so I'd like to set `suppress_errors=False`, but there's no sane way to set this as far as I can tell.

takluyver:
@minrk maybe an environment variable? Bits affecting how the config loading itself happens probably need some special casing.

minrk:
Yeah, it should be possible to disable it. An environment variable is the easiest way to ensure that it takes effect before any config is loaded.

PR #9478, opened 2016-05-18
Allow a raise on bad config using env variable.

Carreau:
Closes https://github.com/ipython/ipython/issues/8927
## 

Needs docs. 

@ssanderson you were the one requesting that. This PR and this one: https://github.com/ipython/traitlets/pull/228 Both shoudl allow you to do that. would it help ?

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -43,6 +43,16 @@ else:
         "/etc/ipython",
     ]
 
+_envvar = os.environ.get('IPYTHON_SUPPRESS_ERRORS')
+if _envvar in {None, ''}:
+    IPYTHON_SUPPRESS_ERRORS = None
+else:
+    if _envvar.lower() in {'1','true'}:
+        IPYTHON_SUPPRESS_ERRORS = True
+    elif _envvar.lower() in {'0','false'} :
+        IPYTHON_SUPPRESS_ERRORS = False
+    else:
+        sys.exit("Unsupported value for environment variable: 'IPYTHON_SUPPRESS_ERRORS' is set to '%s' which is none of  {'0', '1', 'false', 'true', ''}."% _envvar )
 
 # aliases and flags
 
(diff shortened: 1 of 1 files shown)


IPython/core/application.py
```diff
@@ -43,6 +43,13 @@
         "/etc/ipython",
     ]
 
+_envvar = os.environ.get('IPYTHON_SUPPRESS_ERRORS','').lower()
+if _envvar == 'true':
```
takluyver: The two conventions I'm most used to for boolean environment variables are either 1/0, or anything but the empty string means true.

Carreau: So more like the new commit ?

IPython/core/application.py
```diff
@@ -272,18 +280,30 @@ def _ipython_dir_changed(self, change):
...
         """Load the config file.
 
         By default, errors in loading config are handled, and a warning
         printed on screen. For testing, the suppress_errors option is set
         to False, so errors will make tests fail.
+
+        `supress_errors` default value can be change to `False` by setting the
+        IPYTHON_SUPPRESS_ERRORS environment variable to empty string, or `True`
+        by setting the IPYTHON_SUPPRESS_ERRORS environment variable to anythong
+        else. Leaving IPYTHON_SUPPRESS_ERRORS unset will leave
```
takluyver: Sorry to nitpick again, but there's definitely a convention that an environment variable set to an empty string means the same as an environment variable not set at all.

Carreau: There is conflict here, I cannot have "set to anything but empty string" be `True`, and "unset and empty string" be `False` at the same time, because I also need `None`. And `None` should likely be the default, in which case the value is inherited from traitlets. And the default value of env variable is unset.

takluyver: I know, I hadn't realised there were three possible states when I first commented. I think I'd go for '1' = True, '0' = False, and empty/undefined = default, then. Not sure what it should do if it's set to another value.

Carreau:
What about the following ?

IPython/core/application.py
```diff
@@ -272,18 +282,33 @@ def _ipython_dir_changed(self, change):
...
         printed on screen. For testing, the suppress_errors option is set
         to False, so errors will make tests fail.
+
+        `supress_errors` default value is to be `None` in which case the
+        behavior default to the one of `traitlets.Application`.
+
+        The default value can be changed :
+           - to `False` by setting 'IPYTHON_SUPPRESS_ERRORS' environment variable to '0', or 'false' (case insensitive).
+           - to `True` by setting 'IPYTHON_SUPPRESS_ERRORS' environment variable to '1' or 'true' (case insensitive).
+           - to `None` by setting 'IPYTHON_SUPPRESS_ERRORS' environment variable to '' (empty string) or leaving it unset.
```
takluyver: This isn't changing the default, it is the default!

takluyver:
:+1: , just a nitpick over docstring wording.

Carreau:
/changed/set/ done.

PR #9479, opened 2016-05-18
Draft automatic nightly release.

Carreau:
- Add an (encrypted) pypirc with projectJupyter (projectJupyter[at]gmail.com) login credentials for pypi and testpypi.
- build a `ipython-nightly` package, 
- upload it to test-pypi. 
- I've "registered" the package locally with testpypy, so no need for this step. 
- it tries to upload it only on master;not pr; on ipython/ipython

Todo:
- [ ] set devX value based on number of commits (git describe) 
- [ ] make sure it works
- [ ] switch to plain pypi

Annoying part is that encryption will work only if push on a main branch of the ipython repo itself, so I might create a test branch just for that.


damianavila:
So far, looks good... curious, how are you automating the daily push process? AFAIK, Travis does not provide such possibility (I would love to see that...)

Carreau:
> how are you automating the daily push process? AFAIK, Travis does not provide such possibility (I would love to see that...)

I just push on every merge in master. `nightly` is just an historical term. I could name it `ipython-mergedly`

Carreau:
Ok, it now seem to work, further test need to be done directly on master.

Carreau:
Note, beyond being on travis encrypted, the credentials are in my lastpass, that I would gladly share. 
Plus, the projectJupyter account on PyPI **do not have access** to any of the official repository, so it's kinda "safe" if its credentials leaks.

takluyver:
What's the advantage of doing this with IPython? People can `pip install` the dev version directly from Github. I thought the plan was to do it with `notebook`, where there are extra JS build steps if you install from Github.

damianavila:
>I just push on every merge in master. nightly is just an historical term. I could name it ipython-mergedly

Are you sure you want to push on every merge? If you are too much fine grained you have the risk of people not knowing exactly which dev version is using... even more, one user install a dev1 option and 10 min later another user install a dev2... if some bug was introduced, then both versions have the issue and since you are too grained, you could eventually been looking into the incorrect version, I mean, assuming the issue was because something introduced in dev2 whereas it was introduced in dev1...
The opposite works as well, if you are low grained, then you accumulate a lot of potential bugs in one common version... but I guess if I have to choose a compromise I would run with a daily-based devel release process...
And for the "every commit thing", as @takluyver mentioned, in this case (not the notebook one), a quick pip install dev version should be enough...

Carreau:
> What's the advantage of doing this with IPython?

 - Release are made only if test pasees. 
 - It builds a wheels which is smaller that cloning the repo. 
 - nobody can remember the ssh+git syntax.
    - you likely can't get give the syntax to someone without a piece of paper/keyboard. You can with `pip install ipython-nightly`
 - installing from github requires git (yes some people do not have git)
 - it does not prevent to also do it for the notebook. 
 

> Are you sure you want to push on every merge [...]

 - It's slightly coarser grain than GitHub, as it publish only on success build.
 - We average 1.5 PR merged a day since February 1st. Is there really a difference ?
 - Do you **really** think that someone that report a bug using pip install git+ssh have more chance to install the same version if it issues the command at 10 minutes interval than if we publish nightly ? 
 - Do you really want to set-up the infrastructure to push only once a day ? 


But you are right that it is a choice to be made, though we don't seem to have an agreement on the granularity, and I'm not sure the energy to put into something more complex is worth it.

takluyver:
I think we can also give pip the URL of the Github zip/tar download, in which case it wouldn't need git. I don't have any particular objection to doing this on every commit, I'm just not sure that the added maintenance is worth saving a little bit of typing and a slightly smaller download.

Carreau:
> I'm just not sure that the added maintenance is worth saving a little bit of typing and a slightly smaller download.

I don't think there is such an overhead of maintenance, and we don't commit to infinite nightly. 
I still think it's much easier for users. But it's not really important.

minrk:
>  I'm just not sure that the added maintenance is worth saving a little bit of typing and a slightly smaller download.

I think that's true for this repo, but a nightly might be more valuable on repos like the notebook, where the requirements for building from git are substantially greater than building from master. Maybe start there, and once it's worked out, see if an IPython nightly is worth doing.

> I don't think there is such an overhead of maintenance

I hearby designate @Carreau [the Nightly Manager](http://www.imdb.com/title/tt1399664/).

Carreau:
> hearby designate @Carreau the Nightly Manager.

Yeah, I've tried this show for 2 episode, but It did not catched. 
And sorry, but I have **only** 1 pair of eyes, and my brain cannot directly tap into the mpeg stream and decode it.

PR #9480, opened 2016-05-19
Enable mouse support in PTK by default.

### LINKED ISSUE #9205: Need a shortcut or better way to get quicker edit to my previous code
Is there a shortcut for me to point my mouse to a certain part of my python code and edit it from there instead of holding my arrow keys? Thanks!

Carreau:
Closes #9205 

---

Any reason not to ?

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -116,7 +116,7 @@ class TerminalInteractiveShell(InteractiveShell):
         help="Shortcut style to use at the prompt. 'vi' or 'emacs'.",
     ).tag(config=True)
 
-    mouse_support = Bool(False,
+    mouse_support = Bool(True,
         help="Enable mouse support in the prompt"
     ).tag(config=True)
 


takluyver:
I ran with it enabled for a while, and yes, there are good reasons not to. Enabling mouse support captures mouse events in the terminal, so scrolling with the scroll wheel and selecting text with the mouse stop working while IPython is running. This may be dependent on the terminal emulator, but it happens on at least gnome-terminal and xterm.

I think we should leave it disabled by default, so as not to break people's expectations of terminal interaction.

Carreau:
> I ran with it enabled for a while, and yes, there are good reasons not to. Enabling mouse support captures mouse events in the terminal, so scrolling with the scroll wheel and selecting text with the mouse stop working while IPython is running. This may be dependent on the terminal emulator, but it happens on at least gnome-terminal and xterm.
> 
> I think we should leave it disabled by default, so as not to break people's expectations of terminal interaction.

Fair enough, should this answer be written in response of  #9205 that we can close then ?

takluyver:
Good point, done.

PR #9481, opened 2016-05-19
Use python-prompt-toolkit for ipdb.

Carreau:
Still need to hookup history and completion.


Carreau:
Stupid, that won't work in notebook that highjack raw_input.

takluyver:
I think this is the right approach, we'll just need a fallback based on `input()` to use in the kernel.

PR #9483, opened 2016-05-19
Remove Pydb integration as Pydb is unmaintained and not ion PyPI

Carreau:
Closes # 9482

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -31,7 +31,7 @@ from io import open as io_open
 from pickleshare import PickleShareDB
 
 from traitlets.config.configurable import SingletonConfigurable
-from IPython.core import debugger, oinspect
+from IPython.core import oinspect
 from IPython.core import magic
 from IPython.core import page
 from IPython.core import prefilter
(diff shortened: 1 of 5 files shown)


IPython/core/interactiveshell.py
```diff
@@ -974,7 +974,7 @@ def _set_call_pdb(self,val):
                         'Control auto-activation of pdb at exceptions')
 
     def debugger(self,force=False):
-        """Call the pydb/pdb debugger.
+        """Call the /pdb debugger.
```
takluyver: Leftover /

docs/source/whatsnew/pr/remove-pydb.rst
```diff
@@ -0,0 +1 @@
+Integration with pydb has been removed since pydb development has been stopped since 2012, and pydb is not installable from PyPI
```
takluyver: I think this should be an `incompat-` file, so it winds up in the incompatible changes section of the release notes.

takluyver:
I think this is fine - no need to maintain extraneous complexity.

Carreau:
Done.

PR #9487, opened 2016-05-20
Ipdb with prompt_toolkit

takluyver:
Builds on @Carreau's #9481, with a fallback to the standard input-based machinery when we're in the kernel. I don't much like how I check what we're in, though.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -652,6 +652,9 @@ class IPCompleter(Completer):
                          self.dict_key_matches,
                          ]
 
+        # This is set externally by InteractiveShell
+        self.custom_completers = None
+
     def all_completions(self, text):
         """
         Wrapper around the complete method for the benefit of emacs.
@@ -1072,6 +1075,9 @@ class IPCompleter(Completer):
         return u'', []
 
     def dispatch_custom_completer(self, text):
+        if not self.custom_completers:
+            return
+
         line = self.line_buffer
         if not line.strip():
             return None
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -37,6 +37,8 @@ from IPython.utils import coloransi, py3compat
 from IPython.core.excolors import exception_colors
 from IPython.testing.skipdoctest import skip_doctest
 
+from prompt_toolkit import prompt as ptk_prompt
+
 prompt = 'ipdb> '
 
 #We have to check this directly from sys.argv, config struct not yet available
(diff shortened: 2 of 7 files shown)


dsblank:
This will be great, when done! (Just providing encouraging words of support :)

takluyver:
Thanks Doug :-)

Carreau:
+1. Do you want to add the history manager and Completer in this PR or later ?
(or only for 6.0)

Carreau:
:-(  needs rebase.

takluyver:
Rebased, added history and completion. However, there's now a noticeable pause before it creates each debugger prompt; I want to see if I can reduce this.

Carreau:
@takluyver made a PR against your branch to remove the delay.

It just create a new pt_app, and pt_cli and reuse it across prompt calls.

Carreau:
https://github.com/takluyver/ipython/pull/16

takluyver:
I've cleaned it up with a separate class for the prompt_toolkit based debugging machinery, rather than conditional code to use it inside `IPython.core`.

If we go for this approach, ipdb will probably need some changes to use `IPython.terminal.debugger.TerminalPdb`. Pinging @gotcha.

Carreau:
> I've cleaned it up with a separate class for the prompt_toolkit based debugging machinery, rather than conditional code to use it inside IPython.core.

Yeas, I was thinking about that too. 



> If we go for this approach, ipdb will probably need some changes to use IPython.terminal.debugger.TerminalPdb. Pinging @gotcha.

+1 for merging, should I wait then ?

I also played with adding keybindings for ipdb, but it seem prompt_toolkit does not really like that, 
I'll have to dig into that, but that can wait 6.0.

takluyver:
Keybindings that do things like move between frames? Anything that can produce output on stdout while the prompt is active needs to be run inside [CLI.patch_stdout_context()](http://python-prompt-toolkit.readthedocs.io/en/latest/pages/reference.html#prompt_toolkit.interface.CommandLineInterface.patch_stdout_context). And I don't think it works on stderr, though it should be possible to patch that as well.

Let's give @gotcha a chance to see this before we merge.

Carreau:
> keybindings that do things like move between frames? Anything that can produce output on stdout while the prompt is active needs to be run inside CLI.patch_stdout_context(). And I don't think it works on stderr, though it should be possible to patch that as well.


Sure, or even actually execute the command with the prompt. Technically for 6.0 I'd like to refactor the prompt to get current location and frame, to get something slightly smarter than the current 'just' `ipdb>` 

It would also be pretty neat to have a frame selector, where you can actually press up/down and enter to be dropped at a specific point.

Carreau:
needs rebase BTW.

takluyver:
> It would also be pretty neat to have a frame selector, where you can actually press up/down and enter to be dropped at a specific point. 

It would, but before we get too far into building debugger UIs, we should look at what existing things like [pudb](https://pypi.python.org/pypi/pudb) already provide.

Will rebase.

Carreau:
> It would, but before we get too far into building debugger UIs, we should look at what existing things like pudb already provide.

Fair enough.

Carreau:
Do you want to merge, or leave  @gotcha a few more days ?

takluyver:
I think it's been long enough already. :-)

PR #9489, opened 2016-05-20
Correctly teardown test case, avoid filehandle leaking.

Carreau:


takluyver:
Nice one, those warnings have been bothering me for a while. Just not enough to actually do anything about them. ;-)

Carreau:
> Nice one, those warnings have been bothering me for a while. Just not enough to actually do anything about them. ;-)

Yeah, I'm on fire this week. I tracked a few crazy bugs, where it was literally adding a print statement every 2 lines. 

I'm considering writing a tool that wrap every statement in the AST with a user provided function that stop in a debugger as soon as a specific assumption is false.

takluyver:
Rather than doing it on the AST, I wonder if you could use [sys.settrace](https://docs.python.org/3/library/sys.html#sys.settrace), which is the same underlying API that debuggers plug into.

Carreau:
Hummmmmmmm....
Maybe, though they say that trace function is invoked when a new local
scope is entered, seem to be a bit unclear when which function is called on
which line.

I'll investigate.

PR #9496, opened 2016-05-25
Added option to ptshell for highlighting matching brackets.

jonathanslenders:


takluyver:
Looks good, though I haven't tested it out yet. Are there likely to be any disadvantages? If not, we could turn it on by default.

jonathanslenders:
Not as far as I know. Maybe on some terminal with a weird color configuration. I always turn it on, because it's really helpful.

takluyver:
OK, I'll merge this and make a new PR to turn the default on. Then we can wait to see if anyone complains!

Issue #9388, opened 2016-04-09
prompt_toolkit ignores .inputrc


ivanov:
For as long as IPython used readline, users' `.inputrc` settings were respected at the input prompt.

This does not appear to be the case - as [prompt_toolkit does not currently support .inputrc](https://github.com/jonathanslenders/python-prompt-toolkit/issues/56).

Can we add a note about this? I actually don't see anything in `what's new` for 5.x - only 4.x notes in there.

vi users will want to turn this off, or at least know how to turn on prompt_toolkit's vi mode (I think it has one?).

I get the impression that @takluyver @jonathanslenders seem to be closest to this.

jonathanslenders:
Hi all,

prompt-toolkit does not process ``.inputrc`` yet. Mainly because there was no obvious 1-to-1 mapping of the readline and the prompt-toolkit functionality. Especially because the readline configuration uses vt100 escape codes to define key bindings. This is for historical reasons and not platform independent.

However, detecting whether the editing-mode is vi (``set editing-mode vi``), or whether the user has interest in ``history-search-backward``  (e.g.:``"\e[A":history-search-backward``) is not going to be difficult.

The crash guys are doing this: https://github.com/crate/crash/blob/d8e49c0aec8f1c70c49fe56e7f3e967bde1f44a3/src/crate/crash/command.py#L415
See also: https://github.com/jonathanslenders/python-prompt-toolkit/issues/56

Jonathan

takluyver:
Hi Paul! I wondered when we'd be seeing you after this change ;-). You're right, we should definitely mention it in the release notes.

I don't think we should try to do any interpretation of inputrc: I'd rather be able to say unambiguously 'no, that no longer affects input in IPython, here are the new ways to configure things', rather than dealing with a situation where some entries in there work and others don't.

For reference, the config option to enable vi mode is now `TerminalInteractiveShell.vi_mode` (which is a boolean).

Carreau:
Marking as 5.0 to, at least, document it.

Carreau:
documented in #9497, bumbing to 6.0 to reevaluate.

MarkLodato:
So, how do you configure `"\e[A":history-search-backward`? And how do you get up arrow to be normal up arrow and not history search? In old versions I could do

```python
c.TerminalInteractiveShell.readline_parse_and_bind = ['tab: complete']
```

but that no longer works.

takluyver:
[...]

PR #9497, opened 2016-05-25
Update docs on `.inputrc` and input mode.

Carreau:
Bump #9388 to 6.0

### DIFF
--- a/docs/source/interactive/reference.rst
+++ b/docs/source/interactive/reference.rst
@@ -254,61 +254,18 @@ time you restart it. By default, the history file is named
 Autoindent
 ++++++++++
 
-IPython can recognize lines ending in ':' and indent the next line,
-while also un-indenting automatically after 'raise' or 'return'.
+Starting with 5.0, IPython user `prompt_toolkit` in place of ``readline``,
+it thus can recognize lines ending in ':' and indent the next line,
+while also un-indenting automatically after 'raise' or 'return',
+and support real multi-line editing as well as syntactic coloration
+during edition.
 
-This feature uses the readline library, so it will honor your
-:file:`~/.inputrc` configuration (or whatever file your :envvar:`INPUTRC` environment variable points
-to). Adding the following lines to your :file:`.inputrc` file can make
-indenting/unindenting more convenient (M-i indents, M-u unindents)::
+This feature does not use the ``readline`` library anymore – at least for now
+– , so it will not honor your :file:`~/.inputrc` configuration (or whatever
+file your :envvar:`INPUTRC` environment variable points to).
 
-    # if you don't already have a ~/.inputrc file, you need this include:
-    $include /etc/inputrc
-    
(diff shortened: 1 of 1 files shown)


docs/source/interactive/reference.rst
```diff
@@ -254,61 +254,18 @@ time you restart it. By default, the history file is named
...
+while also un-indenting automatically after 'raise' or 'return',
+and support real multi-line editing as well as syntactic coloration
+during edition.
+
+This feature does not use the ``readline`` library anymore – at least for now
+– , so it will not honor your :file:`~/.inputrc` configuration (or whatever
+file your :envvar:`INPUTRC` environment variable points to).
+
+In particular if you want to change the input mode to ``vi``, you will need to
+set the ``TerminalInteractiveShell.vi_mode`` configuration  option of IPython.
```
jonathanslenders: I believe this was renamed to ``editing_mode``.

Carreau: Good point.

docs/source/interactive/reference.rst
```diff
@@ -254,61 +254,18 @@ time you restart it. By default, the history file is named
...
-You will find the default values in your configuration file.
-
+Starting with 5.0, IPython user `prompt_toolkit` in place of ``readline``,
+it thus can recognize lines ending in ':' and indent the next line,
+while also un-indenting automatically after 'raise' or 'return',
+and support real multi-line editing as well as syntactic coloration
+during edition.
+
+This feature does not use the ``readline`` library anymore – at least for now
+– , so it will not honor your :file:`~/.inputrc` configuration (or whatever
```
takluyver: The 'at least for now' currently implies that it might rely on readline again in the future, whereas I guess you mean that it might respect inputrc again in the future.

As on #9388, I don't think we should try to support `.inputrc` at all - an imperfect emulation of readline's config is going to be more confusing than separating ourselves from it entirely. So I think we should just document that it's not supported, and not hint that maybe one day it will be.

Carreau: Removed.

docs/source/interactive/reference.rst
```diff
@@ -254,61 +254,18 @@ time you restart it. By default, the history file is named
...
-      readline library, as these commands are of the kind which readline
-      accepts in its configuration file.
-    * ``InteractiveShell.readline_remove_delims``: a string of characters to be removed
-      from the default word-delimiters list used by readline, so that
-      completions may be performed on strings which contain them. Do not
-      change the default value unless you know what you're doing.
-
-You will find the default values in your configuration file.
-
+Starting with 5.0, IPython user `prompt_toolkit` in place of ``readline``,
```
takluyver: s/user/uses/

Carreau: `r`  and `s` are so close on french keyboards.

PR #9498, opened 2016-05-25
Add what's new incompat about readline.

Carreau:


docs/source/whatsnew/pr/incompat-readline.rst
```diff
@@ -0,0 +1,4 @@
+IPython 5.0 now use prompt_toolkit, any setting that affect ``readline`` will
```
takluyver: 'uses', 'affects'

'*so* any setting...'

docs/source/whatsnew/pr/incompat-readline.rst
```diff
@@ -0,0 +1,4 @@
...
+have no effect, and has (likely) be relplaced by a configuration option on
```
takluyver: 'has likely been replaced'

Well done on getting *effect* and *affect* the right way round, though - that trips up many native speakers.

Carreau: > Well done on getting effect and affect the right way round, though - that trips up many native speakers.

Maybe it's because I'm french, but the 2 completely different, I never understood how people can confuse the two. 

Still: [I heard you're idea's and their definately good](https://xkcd.com/1238/).

Carreau:
Fixed.

PR #9499, opened 2016-05-25
Document IPYTHON_SUPPRESS_CONFIG_ERRORS.

### LINKED ISSUE #9493: Document IPYTHON_SUPPRESS_ERRORS
See  https://github.com/ipython/ipython/pull/9478/files

Carreau:
Closes #9493

### DIFF
--- a/docs/source/development/config.rst
+++ b/docs/source/development/config.rst
@@ -57,8 +57,10 @@ command-line args.
 
 .. note::
 
-    By default IPython will ignore any invalid configuration files, or any
-    unknown configuration option.
+    IPython configuration options are case sensitive, and IPython cannot
+    catch misnamed keys or invalid values.
+    
+    By default IPython will also ignore any invalid configuration files. 
 
 .. versionadded:: 5.0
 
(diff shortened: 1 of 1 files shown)


docs/source/development/config.rst
```diff
@@ -55,6 +55,17 @@ under :file:`profile_default`. If you want the default config files for the
...
 command-line args.
 
+.. note::
+
+    By default IPython will ignore any invalid configuration files, or any
+    unknown configuration option.
+
+.. versionadded:: 5.0
+
+    IPython can be configured to abort in case of invalid configuration file.
```
takluyver: I think we should mention that this doesn't catch misnamed keys or invalid values: 'invalid' means we couldn't load it at all.

Carreau: That's what I was trying to say, but I got the phrasing wrong.

Carreau:
Is that better ?

takluyver:
Yep, thanks. Not sure why I didn't see this earlier.

PR #9501, opened 2016-05-26
Enable bracket matching by default

takluyver:
See #9496 - @jonathanslenders doesn't know of a reason not to do this.


Carreau:
Neat. I wonder if Pygments support rainbow mode for brackets.

PR #9503, opened 2016-05-26
Make a few test non-optional.

Carreau:
And add a few testing dependencies.

---

@takluyver, we can bikeshed on installing manylinux1 post 5.0.


takluyver:
Looks fine to me, thanks.

PR #9504, opened 2016-05-26
Cleanup deprecated features

Carreau:
Basically go through all the deprecation warning, and if it was mentioned "Will be removed in 5.0". remove it. 

Add more version informations to a few other already existing Deprecation warnings.


IPython/utils/warn.py
```diff
@@ -11,7 +11,7 @@
 import sys
 import warnings
 
-warnings.warn("The module IPython.utils.warn is deprecated, use the standard warnings module instead", DeprecationWarning)
+warnings.warn("The module IPython.utils.warn is deprecated since IPython 4.0 and will be remove in 6.0, use the standard warnings module instead", DeprecationWarning)
```
minrk: Adding "since 4.0" is great, but I don't think we should schedule removal. I don't think deprecated code should be removed until:

1. it's been a really long time since ~everybody actually stopped using it (in years, not just releases), or
2. maintaining its existence causes a real problem

minrk:
👍 to most things here, but in general I don't think we should add a "will be remove in X" message to any deprecation messages until leaving the vestigial code causes more of a problem than clutter.

takluyver:
I'm a bit more gung-ho than Min about removing old code, but I do think it makes sense to consider time, not just release cycles. That said, all the things marked for 6.0 here are things that I don't think are that widely used, so the >= 18 months between those releases feels like long enough.

IPython/utils/warn.py
Carreau: Removed. 

I still think t if you don't give a deadline to people they won't update.

Carreau:
I'm slightly more aggressive for removal this time because 5.0 will be LTS and will be used longer than usual , and I would prefer to not have to deal with people complaining about these deprecated things when they migrate from 5.0 to 8.0. 

Though if you really think some of this should stay longer, I won't fight.

IPython/core/formatters.py
```diff
@@ -37,18 +37,6 @@ class DisplayFormatter(Configurable):
 
     # When set to true only the default plain text formatter will be used.
     plain_text_only = Bool(False).tag(config=True)
-    def _plain_text_only_changed(self, name, old, new):
-        warnings.warn("""DisplayFormatter.plain_text_only is deprecated.
```
takluyver: You removed the deprecation warning but not the deprecated thing here!

takluyver:
I think the things actually removed here are OK, and I understand the desire to clean things up before the LTS, the open question is more about the statements that things will be removed in a future release.

Carreau:
> I think the things actually removed here are OK, and I understand the desire to clean things up before the LTS, the open question is more about the statements that things will be removed in a future release.

How's the new version ?

takluyver:
Great, I think that will be sufficiently non-controversial.

PR #9506, opened 2016-05-27
Add flag for simple_prompt.

Carreau:
As this mode seem to be useful beyond testing, expose it via a flag and
make it official.

### DIFF
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -102,6 +102,11 @@ addflag('autoedit-syntax', 'TerminalInteractiveShell.autoedit_syntax',
         'Turn on auto editing of files with syntax errors.',
         'Turn off auto editing of files with syntax errors.'
 )
+addflag('simple-prompt', 'TerminalInteractiveShell.simple_prompt',
+        "Force simple minimal prompt using `raw_input`",
+        "Use a rich interactive prompt with prompt_toolkit",
+)
+
 addflag('banner', 'TerminalIPythonApp.display_banner',
         "Display a banner upon starting IPython.",
         "Don't display a banner upon starting IPython."
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -26,7 +26,7 @@ from prompt_toolkit.styles import PygmentsStyle, DynamicStyle
 from pygments.styles import get_style_by_name, get_all_styles
 from pygments.token import Token
 
-from .debugger import TerminalPdb
+from .debugger import TerminalPdb, Pdb
 from .pt_inputhooks import get_inputhook_func
 from .interactiveshell import get_default_editor, TerminalMagics
 from .ptutils import IPythonPTCompleter, IPythonPTLexer
@@ -58,6 +58,14 @@ class TerminalInteractiveShell(InteractiveShell):
 
             """
             ).tag(config=True)
+
+    @observe('simple_prompt')
+    def _simple_prompt_changed(self, changes):
+        if changes['new'] == True:
+            self.debugger_cls = Pdb
+        else:
+            self.debugger_cls = TerminalPdb
+
 
     autoedit_syntax = Bool(False,
         help="auto editing of files with syntax errors.",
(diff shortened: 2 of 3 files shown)


Carreau:
@tacaswell would this flag name suits you ? 

@takluyver Should we also check for `os.environ["TERM"] == "dumb"`

tacaswell:
I like `'simple-prompt' `.

IPython/terminal/ptshell.py
```diff
@@ -35,6 +35,7 @@
 from .pt_inputhooks import get_inputhook_func
 from .interactiveshell import get_default_editor, TerminalMagics
 
+_use_simple_prompt = 'IPY_TEST_SIMPLE_PROMPT' in os.environ or not sys.stdin.isatty()
```
takluyver: I think these should probably override config, rather than providing the default value.

IPython/terminal/ipapp.py
```diff
@@ -102,6 +102,11 @@ def make_report(self,traceback):
         'Turn on auto editing of files with syntax errors.',
         'Turn off auto editing of files with syntax errors.'
 )
+addflag('simple-prompt', 'TerminalInteractiveShell.simple_prompt',
+        "Force simple minimal prompt using `raw_input`",
+        "Force simple usage of pompt_toolkit for interface.",
```
takluyver: s/pompt/prompt/

Also, it's not really 'simple'. Maybe 'Use a rich interactive prompt with prompt_toolkit'.

docs/source/whatsnew/pr/simple_prompt.rst
```diff
@@ -0,0 +1,4 @@
+when using IPython as a subprocess, like for emacs inferior-shell, IPython can
+be started with --simple-prompt flag, which will bypass all the prompt_toolkit
```
takluyver: started with the...

Also, 'all the' doesn't scan quite right for a single thing (it needs a plural, as in *all the things*). In general, you can say 'the whole...', but in this case I think it's best just to drop 'all' and say 'bypass the prompt_toolkit input layer'.

docs/source/whatsnew/pr/simple_prompt.rst
```diff
@@ -0,0 +1,4 @@
...
+input layer. In this mode completion, prompt color and many other feature are
```
takluyver: features

takluyver:
As #9487 was merged, we also need to make this switch affect the debugger. It can probably just change `debugger_cls` to the parent class from `IPython.core.debugger`, which is equivalently plain.

IPython/terminal/ptshell.py
Carreau: I think the command line flag should take precedence of the env variable,  especially since it has the `--no-simple-prompt` alternative. IIRC that's how most of our commands works. 

I'm fine changing if you still think it should be.

docs/source/whatsnew/pr/simple_prompt.rst
Carreau: Done.

docs/source/whatsnew/pr/simple_prompt.rst
Carreau: done.

IPython/terminal/ipapp.py
Carreau: done.

Carreau:
> As #9487 was merged, we also need to make this switch affect the debugger. 

Yes. 

> It can probably just change debugger_cls to the parent class from IPython.core.debugger, which is equivalently plain.

Unsure what you mean exactly, I'm confused in the english construct, but gave it a try:

```diff
+    @observe('simple_prompt')
+    def _simple_prompt_changed(self, changes):
+        if changes['new'] == True:
+            self.debugger_cls = Pdb
+        else:
+            self.debugger_cls = TerminalPdb
```

takluyver:
That's basically what I meant, but I'd just change `debugger_cls` inside the if clause - `simple_prompt` can only take effect on startup anyway, so there's no need for traitlets' dynamism.

takluyver:
Since you won't be awake for several hours, I'm going to change it locally and merge.

takluyver:
I realised that `init_prompt_toolkit_cli` would be too late to change `debugger_cls` effectively, but I found another way to do it that I like more, using a property.

PR #9511, opened 2016-05-30
Use default system shell to capture shell output.

pavoljuhas:
Adhere to the user `$SHELL` preference when capturing output from system
commands.  This makes the getoutput() and system() interactive functions
behave in a consistent way.  It also makes user shell customizations, i.e.,
aliases and non-exported shell variables, available within the captured 
command.

Example of an inconsistent behavior when `SHELL=/bin/zsh`:

```
In [1]: !print -l a b
a
b

In [2]: out = !print -l a b

In [3]: out
Out[3]: ['/bin/sh: print: command not found']
```

### DIFF
--- a/IPython/utils/_process_common.py
+++ b/IPython/utils/_process_common.py
@@ -70,8 +70,14 @@ def process_handler(cmd, callback, stderr=subprocess.PIPE):
     sys.stderr.flush()
     # On win32, close_fds can't be true when using pipes for stdin/out/err
     close_fds = sys.platform != 'win32'
-    p = subprocess.Popen(cmd, shell=isinstance(cmd, py3compat.string_types),
-                         executable=os.environ.get('SHELL'),
+    # Determine if cmd should be run with system shell.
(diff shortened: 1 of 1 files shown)


IPython/utils/_process_common.py
```diff
@@ -70,6 +71,7 @@ def process_handler(cmd, callback, stderr=subprocess.PIPE):
     # On win32, close_fds can't be true when using pipes for stdin/out/err
     close_fds = sys.platform != 'win32'
     p = subprocess.Popen(cmd, shell=isinstance(cmd, py3compat.string_types),
+                         executable=os.environ.get('SHELL'),
```
takluyver: This should only be set if we're using the shell - which, from the line above, is the case if `cmd` is a string (...I don't like this code).

I think we should also only be using this on Unix systems.

pavoljuhas:
@takluyver - good points, please see the updated PR.

takluyver:
Thanks, looks OK to me.

PR #9515, opened 2016-05-31
Custom SyntaxError handler.

Carreau:
@asmeurer would this suits, you? You have nothing to change, the `SyntaxError`s should be caughts now.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1604,7 +1604,7 @@ class InteractiveShell(SingletonConfigurable):
         self.InteractiveTB.set_mode(mode=self.xmode)
 
     def set_custom_exc(self, exc_tuple, handler):
-        """set_custom_exc(exc_tuple,handler)
+        """set_custom_exc(exc_tuple, handler)
 
         Set a custom exception handler, which will be called if any of the
         exceptions in exc_tuple occur in the mainloop (specifically, in the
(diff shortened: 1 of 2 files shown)


IPython/core/interactiveshell.py
```diff
@@ -2682,14 +2682,18 @@ def error_before_exec(value):
...
+                    self.CustomTB(etype, value, tb)
+                    return error_before_exec(e)
                 except IndentationError as e:
                     self.showindentationerror()
                     if store_history:
                         self.execution_count += 1
                     return error_before_exec(e)
                 except (OverflowError, SyntaxError, ValueError, TypeError,
                         MemoryError) as e:
-                    self.showsyntaxerror()
```
takluyver: Why's this changing? Won't it just call `showsyntaxerror` anyway?

IPython/core/tests/test_ultratb.py
```diff
@@ -196,7 +196,7 @@ def test_non_syntaxerror(self):
             raise ValueError('QWERTY')
         except ValueError:
             with tt.AssertPrints('QWERTY'):
-                ip.showsyntaxerror()
+                ip.showtraceback()
```
takluyver: From the comment in this test, changing this misses the point of what it's supposed to be testing.

Carreau: Oops.

IPython/core/interactiveshell.py
Carreau: Yes it should, but it seem hard to follow the logic if some places call `syntaxerror()` and others `showtraceback()`, 
so I tried to simplify by onl calling `showtraceback()`, still I think this might be wrong. I'll revert.

Carreau:
Fixed. Thanks for th review.

takluyver:
Looks OK, then.

Issue #9316, opened 2016-03-11
Issue when autocompleting after spaces


Achifaifa:
This sounds similar to #8923, but I don't think they are related enough (Plus this happens on linux). I've seen some issues open regarding autocomplete but I haven't seen one describing this.

When I'm in a folder that contains multiple test files (like `this is a test file 1`), ipython successfully autocompletes the files in instructions like `with open("this<tab>` (It suggest some things that are not files, but whatever, close enough).

However, when pressing tab after a space, it autocompletes literally everything, which is not helpful (I'm not sure if that's the intended behaviour, but it's certainly not what I was expecting).

I can reproduce the issue every time I try to autocomplete a file name after a space.

takluyver:
Readline has a list of 'delimiter' characters, which includes space. When you tab complete, it goes back to the last delimiter character to determine the token it's going to use for completions - so if you do `a = b.<tab>`, it's only completing on `b.`.

We're moving away from readline in master, but for now completion still works like that. It has been that way as long as we've had tab completion, so it's not top priority.

Achifaifa:
Oh nice to know, thank you.

anntzer:
FWIW master (with prompt_toolkit) seems to have the same issues.

takluyver:
Yep, the switch to prompt_toolkit has not directly changed this - the completion machinery needs some reworking.

Carreau:
I doubt I can tackle that for 6.0, bumping to 7.

Issue #8923, opened 2015-10-22
Path autocompletion issues on Windows


anntzer:
Somewhat related to #8364.

Starting in C:\
```
In [3]: cd D<tab>
DeprecationWarning    Documents\ and\ Settings
# DeprecationWarning should not be there, I'd expect to autocmplete D:\ too.
In [3]: cd Program\<tab>
<bunch of LaTeX symbols>
In [3]: cd Program\ <tab> # note the space here
# ===> autocompletes to cd Program\SPACE (literally) and stays stuck on that
In [3]: cd Program\ F<tab>
# autocompletes paths and builtins starting from F
# same for cd Program F<tab> without the backslash
# same for cd "Program F<tab> (opening quote)
```

anntzer:
Actually, there's also
```
In [1]: cd C:/Pro<tab>
# ===> autocompletes to
In [1]: cd C:/Program\ Files<enter>
[WinError 3] The system cannot find the path specified: 'C:/Program\\ Files'
```

anntzer:
wrt the second issue (autocompletion to an invalid windows path): I think I know how to fix it, but that basically comes down to specifying what the quoting rules are for magics that take paths as arguments.  For example, paths with special characters could be just always quoted, although even then it is not clear whether to favor single quotes (the Unix tradition) or double quotes (the Windows approach).

As it is right now, it's also a bit awkward to cd to a path that contains a single quote...

anntzer:
Bumping this issue: what's the status of 5.0?  I'd like to get this fixed asap, because the current behavior is extremely annoying.
I think a reasonable approach is
- on Unix systems, quote metacharacters using backslashes.
- on Windows, quote by surrounding with `"` (which cannot be part of a filename).
If we can agree on the behavior and there's still some reasonable window to get this into 5.0, I can make a PR.

minrk:
@anntzer if you know how to fix it, a PR would be very welcome. We can always talk about the details when the PR is opened. There's time for this to get in to 5.0, but there's always 5.1, so it's not such a worry.

PR #9516, opened 2016-06-01
On Windows, quote paths instead of escaping them.

anntzer:
See the second half of #8923: on Windows, while in C:\,

```
In [1]: cd Prog<tab>
```

now completes to `cd "Program Files"` instead of the incorrect `cd Program\ Files`.  Note that further completion on subfolders remains problematic due to #9316.

On Linux, paths with quotes can now be properly cd'ed into: after `$ mkdir foo\'bar\ \"`,

```
In [1]: cd f<tab>
still completes to `cd foo\'bar\ \"` but the backslashes are now properly kept before passing to the shlex'er.
```

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -121,7 +121,7 @@ def protect_filename(s):
         if sys.platform == "win32":
             return '"' + s + '"'
         else:
-            return "".join("\\" + c if c in PROTECTABLES else c for c in s)
+            return "".join(("\\" + c if c in PROTECTABLES else c) for c in s)
     else:
         return s
 
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -117,9 +117,14 @@ def has_open_quotes(s):
 
 def protect_filename(s):
     """Escape a string to protect certain characters."""
+    if set(s) & set(PROTECTABLES):
+        if sys.platform == "win32":
+            return '"' + s + '"'
+        else:
+            return "".join("\\" + c if c in PROTECTABLES else c for c in s)
+    else:
+        return s
 
-    return "".join([(ch in PROTECTABLES and '\\' + ch or ch)
-                    for ch in s])
 
 def expand_user(path):
     """Expand '~'-style usernames in strings.
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -67,7 +67,7 @@ from IPython.utils.decorators import undoc
 from IPython.utils.io import ask_yes_no
 from IPython.utils.ipstruct import Struct
 from IPython.paths import get_ipython_dir
-from IPython.utils.path import get_home_dir, get_py_filename, unquote_filename, ensure_dir_exists
+from IPython.utils.path import get_home_dir, get_py_filename, ensure_dir_exists
 from IPython.utils.process import system, getoutput
 from IPython.utils.py3compat import (builtin_mod, unicode_type, string_types,
                                      with_metaclass, iteritems)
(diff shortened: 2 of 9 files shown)


IPython/core/completer.py
```diff
@@ -117,9 +117,14 @@ def has_open_quotes(s):
 
 def protect_filename(s):
     """Escape a string to protect certain characters."""
+    if set(s) & set(PROTECTABLES):
+        if sys.platform == "win32":
+            return '"' + s + '"'
+        else:
+            return "".join("\\" + c if c in PROTECTABLES else c for c in s)
```
takluyver: I find this a bit hard to read - can we parenthesise the ternary expression (i.e. so the close-paren is just before `for`).

IPython/core/interactiveshell.py
```diff
@@ -3139,10 +3139,9 @@ def find_user_code(self, target, raw=True, py_only=False, skip_encoding_cookie=T
         code = self.extract_input_lines(target, raw=raw)  # Grab history
         if code:
             return code
-        utarget = unquote_filename(target)
         try:
-            if utarget.startswith(('http://', 'https://')):
-                return openpy.read_py_url(utarget, skip_encoding_cookie=skip_encoding_cookie)
+            if target.startswith(('http://', 'https://')):
```
takluyver: Is this assuming that a URL will never be given in quotes?

takluyver:
You've removed several uses of `unquote_filename` - is this no longer necessary? It's not clear what else is removing quotes from filenames.

anntzer:
As far as I can tell, all magics route their argument (as a single string) through `arg_split` (i.e. `shlex.split(..., posix=True)` on POSIX, and `CommandLineToArgvW` on Windows) first, which should take care of all the unquoting.

takluyver:
OK, fair enough.

IPython/utils/path.py
```diff
@@ -74,11 +74,13 @@ def get_long_path_name(path):
 def unquote_filename(name, win32=(sys.platform=='win32')):
     """ On Windows, remove leading and trailing quotes from filenames.
```
takluyver: The docstring should also say that this is deprecated.

anntzer: Fixed (together with the comment above).

IPython/core/completer.py
```diff
@@ -117,9 +117,16 @@ def has_open_quotes(s):
 
 def protect_filename(s):
     """Escape a string to protect certain characters."""
+    if set(s) & set(PROTECTABLES):
+        if sys.platform == "win32":
+            return '"' + s + '"'
+        else:
+            return "".join(("\\" + c if c in PROTECTABLES else c) for c in s)
+            return s.translate(
```
takluyver: Double return here. I think the first way is probably a bit clearer, because `str.translate` is not very commonly used.

anntzer: I tried it and removed (but forgot to do so completely) it because it behaves differently on Py2 anyways...  Amended and force-pushed a correction.

anntzer:
Note: Actually, the splitting algorithm used on Windows is CommandLineToArgvW (which we may want to just reimplement ourselves instead of relying on ctypes, as it is not that complicated), so quoting on Windows should really use something like https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/
However the solution in this PR (just wrapping with quotes) is at least better than the one before (Windows paths cannot contain quotes, so that would really matter only for e.g. passing an argument that contains quotes to `%run`).

takluyver:
I believe we used to do that splitting ourselves, but there were enough little corner cases that it got wrong that we eventually decided the ctypes call was the lesser of two evils. I don't recall any bugs filed about it since we switched to that, so if it ain't broken, don't fix it ;-).

takluyver:
Thanks

PR #9519, opened 2016-06-01
Keep a reference to last execution result on the shell.

Carreau:
In particular to know if the last command has failed.

Ping @asmeurer . useful as iterm2-tools currently store the last state in a global variable, and reset it to 0 once prompt drawn, but as PTK can redraw the prompt many time, it would require shenanigans to track that correctly. And we can do it (relatively) cheaply in IPython itself.  

Kind of requires https://github.com/ipython/ipython/pull/9471

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -166,6 +166,10 @@ class ExecutionResult(object):
         if self.error_in_exec is not None:
             raise self.error_in_exec
 
+    def __repr__(self):
+        return '<%s object at %x, execution_count=%s error_before_exec=%s error_in_exec=%s result=%s>' %\
+                (self.__class__.__qualname__, id(self), self.execution_count, self.error_before_exec, self.error_in_exec, repr(self.result))
+
 
 class InteractiveShell(SingletonConfigurable):
     """An enhanced, interactive shell for Python."""
@@ -433,6 +437,8 @@ class InteractiveShell(SingletonConfigurable):
 
     # Tracks any GUI loop loaded for pylab
     pylab_gui_select = None
+
+    last_execution_succeeded = Bool(True, help='Did last executed command succeeded')
 
     def __init__(self, ipython_dir=None, profile_dir=None,
                  user_module=None, user_ns=None,
@@ -2607,6 +2613,7 @@ class InteractiveShell(SingletonConfigurable):
         result = ExecutionResult()
 
         if (not raw_cell) or raw_cell.isspace():
+            self.last_execution_succeeded = True
             return result
         
         if silent:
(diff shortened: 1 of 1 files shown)


Carreau:
Change of Plan: 

 Keep only boolean for success,

 Cannot keep a handle on the result object, or it does not get garbage
 collected until end of next run, which break some assumtion about wehn
` __del__ `is called.

IPython/core/interactiveshell.py
```diff
@@ -435,6 +439,8 @@ def profile(self):
     # Tracks any GUI loop loaded for pylab
     pylab_gui_select = None
 
+    last_execution_succeed = Bool(True, help='Did last executed command, succeed')
```
takluyver: No comma after command

takluyver: Also, I'd probably call the variable `..._succeeded` or `..._successful`

IPython/core/interactiveshell.py
```diff
@@ -2615,6 +2621,7 @@ def run_cell(self, raw_cell, store_history=False, silent=False, shell_futures=Tr
         result : :class:`ExecutionResult`
         """
         result = ExecutionResult()
+        self.last_execution_succeed = True
```
takluyver: Resetting it at the start of `run_cell` means that code in the cell cannot use `get_ipython().last_execution_succeed`. Maybe that use case is not important, but it would be pretty surprising.

Carreau: I don't think a cell should be able to get this information from the `shell` once executing a cell. I'll consider it unsupported. 
This is mostly for iterm2-tools that need to change the prompt if the previous command failed. 

I'll move it though.

IPython/core/interactiveshell.py
```diff
@@ -2873,16 +2882,19 @@ def run_code(self, code_obj, result=None):
         except SystemExit as e:
             if result is not None:
                 result.error_in_exec = e
+                self.last_execution_succeeded = False
```
takluyver: `run_code` and `run_ast_nodes` return a failure flag, so I think we can set this in one place in `run_cell` rather than three places in here. As this file is now, that one place is line 2716.

Carreau: Changed.

IPython/core/interactiveshell.py
```diff
@@ -166,6 +166,10 @@ def raise_error(self):
         if self.error_in_exec is not None:
             raise self.error_in_exec
 
+    def __repr__(self):
+        return '<%s object at %x, execution_count=%s error_before_exec=%s error_in_exec=%s result=%s>' %\
+                (self.__class__.__qualname__, id(self), self.execution_count, self.error_before_exec, self.error_in_exec, repr(self.result))
```
takluyver: If you do `%r` (or `{!r}` in new-style formatting), it repr-s the object for you, rather than needing an explicit repr() call. This isn't important, however.

Carreau: Good point. I'll keep that in mind.

PR #9521, opened 2016-06-01
Don't compute length of ZeroWidthTokens.

Carreau:
This simplify subclassing of prompt that inject escape sequences for
prompt toolkit.

### DIFF
--- a/IPython/terminal/prompts.py
+++ b/IPython/terminal/prompts.py
@@ -6,6 +6,8 @@ import sys
 
 from IPython.core.displayhook import DisplayHook
 
+from prompt_toolkit.layout.utils import token_list_width
+
 class Prompts(object):
     def __init__(self, shell):
         self.shell = shell
@@ -18,8 +20,7 @@ class Prompts(object):
         ]
 
     def _width(self):
-        in_tokens = self.in_prompt_tokens()
-        return sum(len(s) for (t, s) in in_tokens)
+        return token_list_width(self.in_prompt_tokens())
 
     def continuation_prompt_tokens(self, cli=None, width=None):
         if width is None:


takluyver:
Is ZeroWidthEscape just a name you picked, or does pygments/ptk give it some official status? A quick search didn't find anything.

Carreau:
> Is ZeroWidthEscape just a name you picked, or does pygments/ptk give it some official status? A quick search didn't find anything.

Sorry should have precised, it is a new PTK thing: 

https://github.com/jonathanslenders/python-prompt-toolkit/pull/299

We can conditional import from PTK to be more clear.

Carreau:
updated.

IPython/terminal/prompts.py
```diff
@@ -6,6 +6,13 @@
 
 from IPython.core.displayhook import DisplayHook
 
+try:
+    from prompt_toolkit.token import ZeroWidthEscape
+except ImportError:
+    # backward compat with PTK 1.x that does not yet define this:
```
takluyver: Is this only going into 2.x? Our requirement is currently for `prompt_toolkit>=1.0.0,<2.0.0`, so this code shouldn't be run with PTK 2. Or is the comment erroneous, and ZWE will be available in a future PTK 1.x release?

Carreau: I just asked Jonathan Slenders. I'm unsure of what will become what.

IPython/terminal/prompts.py
```diff
@@ -19,7 +26,7 @@ def in_prompt_tokens(self, cli=None):
 
     def _width(self):
         in_tokens = self.in_prompt_tokens()
-        return sum(len(s) for (t, s) in in_tokens)
+        return sum(len(s) for (t, s) in in_tokens if t is not ZeroWidthEscape)
```
jonathanslenders: Please, just use ``prompt_toolkit.layout.utils.token_list_width``: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/layout/utils.py#L27

Unless there are objections, ``ZeroWidthEscape`` along with several bug fixes, and a few small improvements, is probably already going in the next release: 1.0.1.  (Backwards-compatibility is very important, but right now, I'm not sure, whether each new feature needs in y-increment in x.y.z. In the future definitely, yes.)

Carreau: no objection, just wanted to know you plan. And thanks for the tip on `token_list_width` !

Carreau:
PR updated.

IPython/terminal/prompts.py
```diff
@@ -6,6 +6,11 @@
 
 from IPython.core.displayhook import DisplayHook
 
+try:
+    from prompt_toolkit.layout.utils import token_list_width
+except ImportError:
+    token_list_width = lambda in_tokens: sum(len(s) for (t, s) in in_tokens)
```
takluyver: The function is there in PTK 1.0.0, so there should be no need for a fallback.

https://github.com/jonathanslenders/python-prompt-toolkit/blob/1.0.0/prompt_toolkit/layout/utils.py#L25

Carreau: Good point. Fixed.

PR #9522, opened 2016-06-01
Test on Python 3.5 development branch

Carreau:


takluyver:
Is it likely to change enough to be worth testing? I don't want to slow Travis down too much.

Carreau:
> Is it likely to change enough to be worth testing? I don't want to slow Travis down too much.

test are run roughly in parallel, it slow down the test by less than a minutes: Elapsed time  3 min 36 sec. 

I would definitively not run that on `notebook`.

takluyver:
Tests are run in parallel except when commits and PRs are coming in fast enough that Travis throttles us. I don't feel very strongly about it, I just think it's worth considering.

Carreau:
> Tests are run in parallel except when commits and PRs are coming in fast enough that Travis throttles us. I don't feel very strongly about it, I just think it's worth considering.

Ok, fair enough. Once we drop 3.3.

PR #9528, opened 2016-06-02
Improve deprecation warning.

### LINKED ISSUE #9526: improve deprecation warnings in IPython/terminal/interactiveshell.py
warn on import, and add version since deprecation.

Carreau:
Raise a warning no import, as it is common to have import that are
unused.

Also add since when this is deprecated.

Close #9526

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -9,10 +9,13 @@ from warnings import warn
 from IPython.utils.decorators import undoc
 from .ptshell import TerminalInteractiveShell as PromptToolkitShell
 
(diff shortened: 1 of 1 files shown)


IPython/terminal/interactiveshell.py
```diff
@@ -9,10 +9,13 @@
 from IPython.utils.decorators import undoc
 from .ptshell import TerminalInteractiveShell as PromptToolkitShell
 
+warn("Since IPython 5.0 `IPython.terminal.interactiveshell` is deprecated in favor of `IPython.terminal.ptshell`.",
```
takluyver: favour :-P

jk

Issue #8364, opened 2015-04-22
tab autocompletion on a different windows drive


anntzer:
`cd <other-drive-letter>:/<tab>` suggests autocompletion using paths of the current drive, not of the other drive.

koenvb:
could you clarify a bit more the problem?
* What python version are you using? ( You can go to the about link in the help menu of the notebook)
* What ipython version are you using?
* Any example notebook you could share explaining your problem?

Then I can try to reproduce it

anntzer:
This is using IPython 3.1.0 on Python 3.4.

Open a QtConsole on a Windows box with multiple drives (I haven't tried the notebook).  I assume you start somewhere in the C: drive.  Type e.g. `cd D:/<tab>`.  The list of suggested completions is wrong, it's a list of files from C:, not from D:.

takluyver:
Does it work if you use `D:\`? Backslashes are the real directory separators on Windows, though forward slashes will work in quite a few places.

anntzer:
`cd  D:/` autocompletes to C:/'s folders.
`cd D:` autocompletes to the union of C:/'s folders and the global namespace (perhaps because IPython thinks this may also be regular Python syntax and complete as if it went `if foo:<tab>`?)
`cd D:\` autocompletes to... some weird list of vaguely LaTeX-like commands:
![complete](https://cloud.githubusercontent.com/assets/1322974/7354280/459cc9b6-eccf-11e4-939b-d9fbbce9bb24.png)

takluyver:
hmm, I bet (py)readline is treating `:` as a delimiter, so we only get the part after that to complete on. This may have to wait for the refactor of the completion machinery that I want to get around to.

minrk:
We do set the delimiters. Should we remove ':' from the default delimiters on Windows? I can imagine that causing all kinds of unpleasantness, though.

takluyver:
Yeah, that sounds like it will cause more problems than it solves, with colons in dictionary literals and slicing syntax.

anntzer:
I have an approximate patch (against master) that looks like this:
```
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -274,6 +274,13 @@ def split_line(self, line, cursor_pos=None):
         l = line if cursor_pos is None else line[:cursor_pos]
         return self._delim_re.split(l)[-1]
 
+    def split_suffixes(self, line, cursor_pos=None):
+        """Split a line of text with a cursor at the given position.
+        """
[...]

PR #9476, opened 2016-05-18
Remove Jedi integration before 5.0

Carreau:
There are a few bugs, and things to iron out that we will likely not
have the chance to do before 5.0, do deactivating (removing) jedi
integration for now, we can always resurrect the current code from
git history.

PR #9533, opened 2016-06-02
Properly complete arguments to %cd.

anntzer:
Fixes #4519, #8364, #9316.

Rely on `arg_split` to re-split the entire line (up to the cursor) so
that directory names that include readline separators are properly
handled.

Because the completed symbol may not be the one that
`IPCompleter.complete` (which relies on naive splitting) believes to be,
it is necessary to tag each candidate completion with the symbol _it_
completes.  (Note that this symbol is not even necessarily a prefix of
the "original" symbol, e.g., on Windows, paths may get a double-quote
prepended upon completion.)  This is done using an attribute on an str
subclass.

Multiple completions may end up completing symbols of different length,
so the `_pad_completions` function is used to "pad" the completions,
pretending that they are all completing the same symbol.

Because completions no longer necessarily start with the completed
symbol, it is up to the individual completers to filter out
non-matching completions.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -537,6 +537,50 @@ def back_latex_name_matches(text):
     return u'', ()
 
 
+def _pad_completions(line, symbol, completions):
+    """Pad completions to the same prefix.
+
+    `symbol` must be a suffix of `line`.  This is a helper to handle
+    completions that complete a suffix of `line` that is longer than just
+    `symbol`.
+
+    Returns a pair `(new_symbol, new_completions)` where `new_symbol` is a
+    suffix of `line` possibly longer than the original symbol, and each
+    completion has been left-padded to start with the new `new_symbol`.
+
+    Completions that provide a `.completed_symbol` attribute use it instead.
+    """
+    # Not a very efficient implementation.
+    completed_symbols = [
+        getattr(compl, "completed_symbol", symbol) for compl in completions]
+    stack = line[:-len(symbol)]
+    new_symbol = symbol
+    while True:
+        new_completed_symbols = []
+        for compl in completed_symbols:
+            for i in range(len(symbol), len(new_symbol) + 1):
+                if compl.startswith(new_symbol[-i:]):
+                    new_completed_symbols.append(new_symbol[:-i] + compl)
+                    break
+            else: # This completion is not paddable to the symbol.
+                break
+        else: # Everything was completed successfully.
+            # Pad the actual completions.
+            new_completions = [
+                ncs[:-len(cs)] + compl
+                for compl, cs, ncs in
+                    zip(completions, completed_symbols, new_completed_symbols)]
+            return new_symbol, new_completions
+        # A completion was not paddable.
+        if not stack:
+            raise ValueError(
+                "Completions {} (completing symbols {}) cannot be padded to "
+                "the same prefix (line: {!r}, symbol: {!r})".format(
+                    completions, completed_symbols, line, symbol))
+        new_symbol = stack[-1] + new_symbol
+        stack = stack[:-1]
+
+
 class IPCompleter(Completer):
     """Extension of the completer class with IPython-specific features"""
     
@@ -1109,13 +1153,7 @@ class IPCompleter(Completer):
             try:
                 res = c(event)
                 if res:
-                    # first, try case sensitive match
-                    withcase = [cast_unicode_py2(r) for r in res if r.startswith(text)]
-                    if withcase:
-                        return withcase
-                    # if none, then case insensitive ones are ok too
-                    text_low = text.lower()
-                    return [cast_unicode_py2(r) for r in res if r.lower().startswith(text_low)]
+                    return [cast_unicode_py2(r) for r in res]
             except TryNext:
                 pass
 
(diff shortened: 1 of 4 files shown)


Carreau:
Hi there !

Thanks a lot !

I'm (a bit) worried of touching completion close to 5.0 beta (I don't expect you to know we are close to beta, that I hope to release in less than a week). We've played with [integrating jedi](https://github.com/ipython/ipython/pull/9375) not so long ago, but had to [roll back](https://github.com/ipython/ipython/pull/9476) recently because of many undesirable side effect. 

I would much better prefer delaying that a bit, and get a complete overhaul of the completion machinery for 6.0 (use Completions object instead of Strings, etc...) and match jedi API than trying to shoehorn a hack. 

Also this refactoring should allow us to do smarter things like attaching metadata to completion (like their type) that would be shown in the UI.

What do you think ?

@takluyver your thoughts too.

anntzer:
I agree that the current solution is quite a hack (especially subclassing str to attach metadata), so it would be better if the completion API was re-thought.
On the other hand this patch only changes a tiny part of the public API (how many custom completers that ship independently from IPython are there?...) and solves a real world issue.  In particular, on Windows, the inability to autocomplete paths in another drive or after `Documents and Settings` is a real pain.

IPython/core/completer.py
```diff
@@ -537,6 +537,50 @@ def back_latex_name_matches(text):
...
+def _pad_completions(line, symbol, completions):
+    """Pad completions to the same prefix.
+
+    `symbol` must be a suffix of `line`.  This is a helper to handle
+    completions that complete a suffix of `line` that is longer than just
+    `symbol`.
+
+    Returns a pair `(new_symbol, new_completions)` where `new_symbol` is a
+    suffix of `line` possibly longer than the original symbol, and each
+    ompletion has been left-padded to start with the new `new_symbol`.
```
Carreau: I think your "Ompletion" is missing a C.

anntzer: fixed

Carreau:
> On the other hand this patch only changes a tiny part of the public API (how many custom completers that ship independently from IPython are there?...) and solves a real world issue

I'm more concern of breaking some others completions than API, especially since 5.0 is supposed to be LTS, so will be supported a bit longer than usual. The patch seem relatively straightforward, so I'm going to re-study-it in more detail tomorrow, and likely give it a go if another core dev agree.

I'm still wondering what the `text.lower()` part could be responsible for.

anntzer:
Apparently, `%CD bla<tab>` would autocomplete directories too... not that it would be that useful, because `%CD blabla` would fail even if blabla is a directory.  So dropping this behavior is prebably warranted.

takluyver:
I'm inclined to say this is too much complexity to add when we're trying to get a release out, especially since it's an LTS release. Refactoring the completion machinery is one of my goals for IPython 6, and I'm loath to add more complexity to the existing system, because it will mean more to wade through when refactoring it. We've lived with this limitation for a long time, so I don't think removing it is urgent.

Carreau:
> I'm inclined to say this is too much complexity to add when we're trying to get a release out, especially since it's an LTS release. Refactoring the completion machinery is one of my goals for IPython 6, and I'm loath to add more complexity to the existing system, because it will mean more to wade through when refactoring it. We've lived with this limitation for a long time, so I don't think removing it is urgent.

Ok, let's wait on that, we can merge it early after a the 5.0 release, and iterate on the completion.

Carreau:
> Ok, let's wait on that, we can merge it early after a the 5.0 release, and iterate on the completion.

@anntzer The advantage is that if this is schedule for 6.0 you can change the API, so as you've already digged through the completer API, and see some things that would need to be changed you can do it  ! :-) We'll be happy to work with you for the new completion API !

anntzer:
OK, please ping me here when this will happen.

anntzer:
Anything new here?  I could drop by BIDS (more or less) whenever you want to start working on this...

Carreau:
I want to push 5.1 out, that has a couple of important bugfix, and then would be happy to have the 5.x and master branch diverge.

Carreau:
We've apparently started to road toward 6.0, if you can just rebase, I
ll be happy to merge (and start working soon on a new completion API)

anntzer:
I think I'd rather work on a new API instead... as you may remember this PR quite a hack.
Does it mean that the entire API of `IPython.core.{completer,completerlib}` is up for grabs?

Carreau:
> I think I'd rather work on a new API instead... as you may remember this PR quite a hack.
> Does it mean that the entire API of IPython.core.{completer,completerlib} is up for grabs?

I would say yes; if we can integrate jedi as well, that would be nice. 
If we can try to keep _some_ compatibility to not break Spyder, IPykernel and other libraries that would be nice. 

We can also go the route of making a  provisional new completer API that is completly opt-in / private. Use it in IPython and make it public later.

Carreau:
Also if you want to close feel free to do it.

anntzer:
I'll leave it open for now and wait until I can come up with something better...

Carreau:
Closing as this has stalled for a long time. Tagging appropriately so that we can receive it when needed.

PR #9549, opened 2016-06-06
Fix test on Python nightly

Carreau:
I'll self merge as soon as test pass, but will still appreciate review.

### DIFF
--- a/IPython/core/tests/test_oinspect.py
+++ b/IPython/core/tests/test_oinspect.py
@@ -270,7 +270,8 @@ def test_info():
     nt.assert_equal(i['type_name'], 'type')
     expted_class = str(type(type))  # <class 'type'> (Python 3) or <type 'type'>
     nt.assert_equal(i['base_class'], expted_class)
-    nt.assert_equal(i['string_form'], "<class 'IPython.core.tests.test_oinspect.Call'>")
+    if sys.version_info > (3,):
+        nt.assert_regex(i['string_form'], "<class 'IPython.core.tests.test_oinspect.Call'( at 0x[0-9a-f]{1,9})?>")
     fname = __file__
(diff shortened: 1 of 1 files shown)


minrk:
I'd like to fix the actual behavior, too, because it seems like the test revealed a real bug - either in IPython or Python, which we are now hiding.

Carreau:
> I'd like to fix the actual behavior, too, because it seems like the test revealed a real bug - either in IPython or Python, which we are now hiding.

No. It's a change of behavior of Python that now shows the object id for types:

```
(rootenv)~/eraseme master$ python3.6
Python 3.6.0a1+ (default, Jun  7 2016, 09:29:50)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
>>> class C:pass
...
>>> C
<class '__main__.C' at 0x7fa2f0d07ad8>
>>>
(rootenv)~/eraseme master$ python3.5
Python 3.5.1 |Anaconda 4.0.0 (x86_64)| (default, Dec  7 2015, 11:24:55)
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
>>> class C:pass
...
>>> C
<class '__main__.C'>
>>>
```

Carreau:
Trying to find the relevant issue/commit.

Carreau:
I'm going to say it's https://hg.python.org/cpython/rev/af29d89083b3

Carreau:
And it's confusing because NEW file says change in the REPL,  but patch and issue says REPR. 
Reported upstream, but the change seem on purpose, so yes, the object adress will show on 3.6 and not on <3.5, so not hiding a bug.

minrk:
Thanks for investigating!

PR #9558, opened 2016-06-07
allow failures on nightly Python

minrk:
while it's useful to run the tests, failures on Python nightly are likely to be not our bugs, and shouldn't generally be considered CI failures.


Carreau:
Fair enough. The issue is that we'll likely never actually see the failure

takluyver:
There are usually a couple of things that we end up fixing in our code with each new Python feature version. But I don't think it's really worth worrying about while it's in alphas. Once it gets to beta, we should start to look at the tests.

PR #9559, opened 2016-06-07
DOC: Fix code/code-block errors

klonuo:
These are fixes for:

```
../ipython/docs/source/interactive/reference.rst:690: ERROR: Error in "code" directive
../ipython/docs/source/interactive/reference.rst:730: ERROR: Error in "code" directive
../ipython/docs/source/interactive/shell.rst:82: ERROR: Unknown directive type "codeblock"
../ipython/docs/source/interactive/shell.rst:95: ERROR: Unknown directive type "codeblock"
```

PS. In diff you may see other lines changed, as my preference is set to clear whitespace on line end

### DIFF
--- a/docs/source/interactive/reference.rst
+++ b/docs/source/interactive/reference.rst
@@ -19,7 +19,7 @@ file and ignore your configuration setup.
 
 Please note that some of the configuration options are not available at
 the command line, simply because they are not practical here. Look into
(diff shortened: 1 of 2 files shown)


minrk:
Thanks. It would be helpful if you could avoid committing those whitespace changes in the future.

PR #9567, opened 2016-06-07
Refactor autogen config to match autogen magic workflow

klonuo:
`autogen_config.py` generated unused `generated.rst`:

```
../ipython/docs/source/config/options/generated.rst:: WARNING: document isn't included in any toctree
```

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -41,8 +41,7 @@ clean_api:
 
 clean: clean_api
 	-rm -rf build/* dist/*
-	-cd $(SRCDIR)/config/options; test -f generated && cat generated | xargs rm -f
-	-rm -rf $(SRCDIR)/config/options/generated
+	-rm -f $(SRCDIR)/config/options/config-generated.txt
 	-rm -f $(SRCDIR)/interactive/magics-generated.txt
 
 pdf: latex
@@ -75,7 +74,7 @@ source/interactive/magics-generated.txt: autogen_magics.py
 	$(PYTHON) autogen_magics.py
 	@echo "Created docs for line & cell magics"
 
-autoconfig: source/config/options/generated
+autoconfig: source/config/options/config-generated.txt
 
 source/config/options/generated:
 	$(PYTHON) autogen_config.py
--- a/docs/autogen_config.py
+++ b/docs/autogen_config.py
@@ -7,7 +7,6 @@ from ipykernel.kernelapp import IPKernelApp
 
 here = abspath(dirname(__file__))
 options = join(here, 'source', 'config', 'options')
-generated = join(options, 'generated.rst')
 
 def write_doc(name, title, app, preamble=None):
     filename = '%s.rst' % name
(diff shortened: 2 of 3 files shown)


Carreau:
+1,  @willingc you are the docmaster. I'll let you decide on this one.

takluyver:
IIRC it's there to act as a target for the Makefile - if `generated.rst` is newer than the source files, make knows it needn't regenerate the API docs (and hence building the docs themselves can be quicker). So I'd rather note that in comments than remove it.

klonuo:
@takluyver you are right, it is supposed to be used in docs Makefile, although it is referenced there as `generated` while created file is `generated.rst`, so it actually doesn't have any impact right now.

It is used to clean generate config files:
```shell
-cd $(SRCDIR)/config/options; test -f generated && cat generated | xargs rm -f
```

and to validate generation:
```shell
autoconfig: source/config/options/generated
```

So probably we should close this PR and rename the file either in Makefile or in autogen_config script, or:

autogen_magic script works similarly, where generated file is used as indicator, this file is also included in rst source and named as `magics-generated.txt`. To make the same workflow for autogen config, we can change `docs/sources/config/options/index.rst` to:

```diff
--- a/docs/source/config/options/index.rst
+++ b/docs/source/config/options/index.rst
@@ -6,7 +6,4 @@ Any of the options listed here can be set in config files, at the
 command line, or from inside IPython. See :ref:`setting_config` for
 details.
 
-.. toctree::
-
-   terminal
-   kernel
+.. include:: config-generated.txt
```

and autogen_config to:

```diff
--- a/docs/autogen_config.py
+++ b/docs/autogen_config.py
@@ -7,19 +7,17 @@
 
 here = abspath(dirname(__file__))
 options = join(here, 'source', 'config', 'options')
-generated = join(options, 'generated.rst')
+generated = join(options, 'config-generated.txt')
+
 
 def write_doc(name, title, app, preamble=None):
-    filename = '%s.rst' % name
-    with open(join(options, filename), 'w') as f:
+    with open(generated, 'a') as f:
         f.write(title + '\n')
         f.write(('=' * len(title)) + '\n')
         f.write('\n')
         if preamble is not None:
             f.write(preamble + '\n\n')
         f.write(app.document_config_options())
-    with open(generated, 'a') as f:
-        f.write(filename + '\n')
 
 
 if __name__ == '__main__':
```

So in Makefile validation can work the same as with autogen magic and clean step will not be necessary.
I can make these changes in new PR if you agree.

willingc:
@klonuo I'm going to defer to @takluyver on the best direction on this PR. Your latest suggestion is reasonable to me.

Carreau:
> I can make these changes in new PR if you agree.

You can also force push on this one and change the title.

takluyver:
I'm at a conference, so I haven't looked in detail at the various autogen
files, but what you're saying sounds sensible from a quick read.

Carreau:
Thanks @klonuo  !

PR #9570, opened 2016-06-07
Swap echo messages

klonuo:
Swap messages for autogen_config and autogen_magics (they were in wrong order) and remove double quotes from `ECHO` which is not needed on Windows and gets printed as such.


Carreau:
Well in seem you are one of our few windows user !

Thanks !

Issue #9556, opened 2016-06-07
PTK completion: Tab, Enter and autofill


klonuo:
- Pressing Tab, causes cursor to jump in completion drop-down box and navigates the same as down-arrow, which is unnecessary duplicate key assignment
 - While inside completion drop-down box - user must accept one of suggested completions, as initial typing is irreversibly changed as navigating through completions
 - Still inside completion drop-down box - pressing Enter, will immediately execute what's on command-line

Suggested solution - simply navigating through completion drop-down box should not automatically change my command-line unless I press Tab on selected item (Enter should do the same too, instead execution). Tab could jump cursor in completion drop-down box, but also do auto-completion for selected item, while letting navigation to down-arrow.

klonuo:
Considering that PTK can provide additional box to each item in completion drop-down box, i.e.:

![gitsome screenshot](https://camo.githubusercontent.com/26c24cf091df9fee2a65b80ae49341a4d9f9782e/687474703a2f2f692e696d6775722e636f6d2f5a5261464770592e706e67)

perhaps pressing <kbd>Tab</kbd> can unroll this sub box with completion signature (one-line docstring) while letting <kbd>Enter</kbd> to do auto complete.

Carreau:
> Pressing Tab, causes cursor to jump in completion drop-down box and navigates the same as ↓, which is unnecessary duplicate key assignment

I think having arrow keys is pretty standard. 


> While inside completion drop-down box - user must accept one of suggested completions, as initial typing is irreversibly changed as navigating through completions

Yes, that's likely a bug. 

> Still inside completion drop-down box - pressing Enter, will immediately execute what's on command-line

That should be fixed on master. I'll do a beta 2 soon. 

> perhaps pressing → can unroll this sub-box with completion signature (one-line docstring). 
> I think that this sub-box can be triggered on a single item (unlike shown on screenshot)

We need to refactor tab completion for that, and that's new feature. 


Thanks for all you propositions really appreciate it. Can you try to separate these in bugs/feature when possible.

Carreau:
Marking as 5.0 as we should try to fix the following:

> While inside completion drop-down box - user must accept one of suggested completions, as initial typing is irreversibly changed as navigating through completions.

klonuo:
> I think having arrow keys is pretty standard.

Sure, but my point here is that <kbd>Tab</kbd> can be used to do actual completion, instead having same function as <kbd>&darr;</kbd> key. It's just a suggestion.

[...]

PR #9572, opened 2016-06-07
Restore previous key bindings.

### LINKED ISSUE #9554: PTK completion: Close completion drop-down box
Completion drop-down box can not be closed with <kbd>Escape</kbd>. User will need to type something to close it, or move cursor back.

Suggestion: <kbd>Escape</kbd> should close completion drop-down box

### LINKED ISSUE #9564: Esc to reset prompt
In 4.x, `Esc` would clear the prompt.  Is there any way to turn this back on in 5.0?

Carreau:
Esc dismiss the completer on next loop tick, or next keypress (technical
limitation) then clear the prompt.

ControlC dismiss the completer without clearing the buffer first.

Closes #9564
Closes #9554
Bump #9556 To feature request.
## 

Thanks to @chris-b1 and @klonuo for pointing that out. 

@jonathanslenders, is there a way to get rid if the delay introduced after pressing escape ?

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -199,9 +199,17 @@ class TerminalInteractiveShell(InteractiveShell):
             else:
                 b.insert_text('\n' + (' ' * (indent or 0)))
 
+        # We have to set eager to True for Escape this will lead to a delay to
+        # dismiss the completer or clear the buffer until next loop, or next input character.
+        # CtrlC will act immediately.
+        @kbmanager.registry.add_binding(Keys.Escape, filter=HasFocus(DEFAULT_BUFFER), eager=True)
         @kbmanager.registry.add_binding(Keys.ControlC, filter=HasFocus(DEFAULT_BUFFER))
         def _reset_buffer(event):
-            event.current_buffer.reset()
+            b = event.current_buffer
+            if b.complete_state:
+                b.cancel_completion()
+            else:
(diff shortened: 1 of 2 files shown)


jonathanslenders:
I didn't test it, but don't merge it like that. ``eager=True`` means that it will replace other key bindings that start with escape. This means that key bindings like ``meta-b``, ``meta-l'', ``meta-f`` etc... won't work anymore. The reason is that when alt is pressed in front of a key binding, a VT terminal sends two keys through the input. Esc + the actual key. So pressing escape, followed by 'b' is the same as typing ``alt-b``.

This is also the reason that there is a delay after escape. If a normal escape has been pressed and nothing follows right after it. (There is a small timer), then we consider it a normal ``escape`` press. This timeout is also noticeable when using Vim and going back to navigation mode.

For Vi users, it would be very confusing to have esc just hide the completion menu. It can do that as well, but it should at least go to navigation mode. For Emacs users, you want to make sure not to shadow all the other meta-style key bindings. It could work by chaining the ``HasCompletions`` filter: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/filters/cli.py#L79

Because of this complexity, I preferred to have ``Ctrl-G`` to hide the completion menu. I think this is what Emacs is doing.

Carreau:
Ok, thanks !

PR #9573, opened 2016-06-07
Add provisional warnings.

Carreau:


IPython/core/interactiveshell.py
```diff
@@ -277,16 +285,29 @@ class InteractiveShell(SingletonConfigurable):
...
+    
     sphinxify_docstring = Bool(False, help=
         """
         Enables rich html representation of docstrings. (This requires the
         docrepr module).
         """).tag(config=True)
+
+    @observe("sphinxify_docstring")
+    def _sphinxify_docstring_changed(self, change):
+        if change['new'] is True:
```
minrk: Generally would use `if change['new']:` here, rather than `is True`.

Carreau: Done

PR #9575, opened 2016-06-08
undeprecate terminal.interactiveshell

minrk:
move terminal.ptshell to terminal.interactiveshell, since it is the new implementation of the class.

The deprecation was there while the readline implementation remained, but there is no need for it now that there is just one implementation of TerminalInteractiveShell.


Carreau:
@takluyver ?

takluyver:
I'm a bit wary because TIS has changed quite a bit, so for anyone importing
it, there's a good chance that something they do with it is broken.
Changing the import doesn't make that any easier, but it does indicate that
it's been changed heavily.

Carreau:
I'm leaning toward @takluyver point of view. We can ask to our BDFL for his opinion.

minrk:
The old class is gone and the deprecated import just returns the new object, so the breakage is exactly the same, whichever module you import it from. The available cases are:

1. They are just instantiating the object, and it works, so the deprecation warning is unhelpful and updating the import requires adding frustrating IPython version compatibility code.
2. It doesn't work, and they need to update how they use TerminalInteractiveShell. They may get a warning that the import changed, but no information about how the object they are trying to use may have changed, so the deprecation warning doesn't improve the experience, it just adds one more thing to change in order to be compatible with the new API.

In both cases, I see no benefit to the (invisible by default) deprecation warning and moved import. In the first case, the no-deprecation is clearly better, while in the second case it seems no worse. Am I missing something? Right now, it reads like changing things just for the sake of changing things. I'm trying to see whose upgrade experience is improved by changing this import.

Carreau:
PTTerminalInteractiveShell had a different name, but it was annoying for users config. So that's why we renamed it back. I can understand both arguments. 

Let's ask @fperez who as access to The Truth to enlighten us with his knowledge.

takluyver:
I agree that there's no practical value in making people change the import as well as the usage. It's like that because of the development process - the new class was introduced before the old one was removed. But it does indicate that we don't plan to make changes this size frequently.

All of this is in that inconvenient 'semi-public API' place - we've invited people to use any of the IPython API in extensions and startup scripts and so forth, but we haven't explicitly designed most of it as public API, and the surface is too big to keep it all compatible as we make changes to the application.

minrk:
I should also mention that I only made this PR because I noticed that we had several imports of the deprecated module still around, causing the deprecation warnings. When I went to fix the warnings, I saw that there didn't seem to be a reason for the deprecation anymore, now that the old implementation is gone. So if we don't take this PR, we still need to fix our imports of the old name in this repo, at least.

fperez:
I don't think there's a particularly clean solution here, and I'm OK with @minrk's approach and rationale, imperfect as the whole thing may be.

One thing though: this needs docs!! A change like this (even if it means that an old class is back) should be noted in the docs, and it's better to get in the habit of updating our docs alongside with code rather than putting all that burden later on the release stage.

Carreau:
I rebased it as #9626, 

> (even if it means that an old class is back)

The old class is back but never disappeared except for people relying on master. There is no reason to have in the docs "hey we temporarily  had this class which is now gone". But yes, I'll precise that explicitly on the mailing list when I publish beta4.

PR #9551, opened 2016-06-06
Sphinxify

SylvainCorlay:
Fixes #9513 : Putting the docrepr functionality behind a flag.

PR #9576, opened 2016-06-08
InteractiveShell.object_inspect_text must return text

minrk:
since that's the whole point of it.

Adds object_inspect_mime for returning the the new mimebundle of formatted outputs, which `_get_info` now returns.

Follow-up of #9551, fixing bugs found in downstream tests in ipykernel.

Includes a minor style change for oinfo text output, moving the colon inside the header block, which is how it was before #9551.


SylvainCorlay:
:+1:

PR #9590, opened 2016-06-09
Restore vi-insert mode `CtrlP` and  `CtrlN` to readline behavior.

### LINKED ISSUE #9584: Behavior of Control-p changed with prompt toolkit and vi bindings
Before the new prompt toolkit changes, in the old readline days, your average vim users would put something like this in their `~/.inputrc`:

```
$include /etc/inputrc
set editing-mode vi
$if mode=vi
set keymap vi-command
$endif
```

With this setup, we startup IPython, and we are in editing mode by default, as seems natural.  In editing mode, Control-p brings up previous lines from the history.   ESC (or its vim equivalents) puts us into navigation mode, which has the same Control-p behavior.

In the new prompt_toolkit terminal emulation, I believe we need to add the following to `~/.ipython/profile_default/ipython_config.py `:

```
# Configuration file for all things ipython
c = get_config()
c.TerminalInteractiveShell.editing_mode = 'vi'
```

In this case, when I start IPython, I am again in editing mode (good) but now Control-p starts autocompletion, until I exit to navigation mode with ESC (etc).   I use Control-p all the time, so it's a significant inconvenience having to prefix all my Control-p's with ESC.   As a vim user, as you can imagine, I hate using the cursor keys, which would be my other option.

It would be good to keep the previous Control-p behavior, and probably by default, but at least available by configuration.

Carreau:
Vi key binding for readline are slightly different than the Vi Editor.
CtrlP and CtrlN in insert mode do recall previous or next history line
instead of calling the completer.

By default Prompt toolkit (1.0.0 at least) implement the Vi-editor
key bindings. Though muscle memory is hard and our vi-mode users[1] have a
hard time adapting. This though overwrite the PTK default and mimic the
readline behavior.

Completer can still be invoked with `<tab>`, and then CtrlP, CtrlN will
select previous / next.

Closes #9584

---

Can @matthew-brett confirm that this is what is desired ?

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -202,6 +202,24 @@ class TerminalInteractiveShell(InteractiveShell):
             else:
                 b.insert_text('\n' + (' ' * (indent or 0)))
 
+        @kbmanager.registry.add_binding(Keys.ControlP, filter=(ViInsertMode() & HasFocus(DEFAULT_BUFFER)))
+        def _previous_history_or_previous_completion(event):
+            """
+            Control-P in vi edit mode on readline is history next, unlike default prompt toolkit.
+
+            If completer is open this still select previous completion.
+            """
+            event.current_buffer.auto_up()
+
+        @kbmanager.registry.add_binding(Keys.ControlN, filter=(ViInsertMode() & HasFocus(DEFAULT_BUFFER)))
+        def _next_history_or_next_completion(event):
+            """
+            Control-N in vi edit mode on readline is history previous, unlike default prompt toolkit.
+
+            If completer is open this still select next completion.
+            """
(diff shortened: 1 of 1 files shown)


matthew-brett:
Yes, excellent - just what the doctor ordered (that's funny because I'm actually a doctor...)

Carreau:
Rebased to fix conflict. 

@fperez if you have a few cycle to review / click the green button as Min and Thomas are busy, that would allow me to likely make a beta3 today.

takluyver:
We'll call you the beta machine ;-)

Not for this PR, but I think we're picking up quite a lot of shortcut definitions nested inside the `init_prompt_toolkit_cli` function - it might make sense to move them out to a separate function and/or module.

PR #9595, opened 2016-06-10
Fix case where completer menu get stuck if nothing is focused.

Carreau:
---

Ping @asmeurer , I think that's the one you've hit.


asmeurer:
Wow, long branch names much? 

It seems to fix the problem, as far as I can tell.

Carreau:
> Wow, long branch names much?

Yeah I got bored of not being able to figure out what was merged and make a recap with `git log --graph --oneline`

PR #9596, opened 2016-06-10
BUG: extra blank line on dedenting keywords

### LINKED ISSUE #9586: Extra blank line needed in multi-line
Thanks for the all the fast fixes so far.   This is really minor, so feel free to punt, but it has been tripping me up a little.  In 5.0, if a function contains a return statement, it takes 1 extra line to close the multi-line block compared to previous behavior.  

Note that if that function (or more realistically, an `if` block) doesn't end in a `return`/`pass`, then the old behavior persists.  Let me know if these examples aren't clear.

#### 4.x
```
In [3]: def f(a):
   ...:     return 5
   ...:
```

#### 5.0b2
```
In [10]: def f(a):
    ...:     return 4
    ...:
    ...:

In [11]: def f(a):
    ...:     a = 5
    ...:
```

chris-b1:
closes #9586 
If I understand correctly, this worked in 4.x because because lines were pushed one at a time to the `IPythonInputSplitter`, where now the the entire block is passed at once, so an empty line gets dropped if there is no whitespace.  eg:

``` python
In [1]: "if True:\n    pass\n".splitlines()
Out[1]: ['if True:', '    pass']

In [2]: "if True:\n    a=4\n    ".splitlines()
Out[2]: ['if True:', '    a=4', '    ']
```

cc: @takluyver

### DIFF
--- a/IPython/core/inputsplitter.py
+++ b/IPython/core/inputsplitter.py
@@ -631,10 +631,6 @@ class IPythonInputSplitter(InputSplitter):
         if not lines_list:
             lines_list = ['']
 
-        # interpet trailing newline as a blank line
-        if lines.endswith('\n'):
-            lines_list += ['']
-
         # Store raw source before applying any transformations to it.  Note
         # that this must be done *after* the reset() call that would otherwise
         # flush the buffer.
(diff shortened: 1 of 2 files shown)


Carreau:
+1

@takluyver quick look ?

takluyver:
Thanks for the detective work. It seems reasonable, so let's merge it and see how things go.

PR #9605, opened 2016-06-10
Add a known issue : with PTK 1 double quote have issues.

Carreau:


takluyver:
It sounds like there are a couple of issues that are fixed in PTK master - it may be best to release IPython 5.0 after a new release of prompt_toolkit so we don't have to field lots more reports of the same issues.

Carreau:
> It sounds like there are a couple of issues that are fixed in PTK master - it may be best to release IPython 5.0 after a new release of prompt_toolkit so we don't have to field lots more reports of the same issues.

Sure. We can just make an RC and wait for PTK release after that.

Issue #9600, opened 2016-06-10
Multi line paste not working anymore in 5.0 beta


jorisvandenbossche:
Copy the below text in the new ipython console:

```
x = 1
y = 2
```

and I only see the first line pasted.

I am using Windows 7, cmd command prompt, and using right clicking to paste. With previous versions of ipython this works (directly executing all but the last line, since there was no multi-line input).

jorisvandenbossche:
Strange thing, if I try to copy multiple lines:

```
x = 1
y = 2
z = 3
```

then the first two (or all but the last) lines are concatenated and of course give a Syntax Error. Result of pasting the above:

```
In [33]: x = 1y = 2
  File "<ipython-input-33-68a99508cab0>", line 1
    x = 1y = 2
         ^
SyntaxError: invalid syntax
```

takluyver:
@jonathanslenders any idea what could be up with pasting on Windows? I wonder if it's a line-ending issue?

klonuo:
I can reproduce this too:

```pytb
In [3]: x = 1y = 2
  File "<ipython-input-3-68a99508cab0>", line 1
    x = 1y = 2
         ^
SyntaxError: invalid syntax
```

pasting same in opened triple quotes works:

```python
In [4]: a = '''x = 1
   ...: y = 2
   ...: '''

In [5]: a
Out[5]: 'x = 1\ny = 2\n'
```

Carreau:
Can both of you try with PTK master ?

jorisvandenbossche:
I get the same on prompt_toolkit master, but, I do get some errors with master when using any arrow key

Carreau:
> I get the same on prompt_toolkit master, but, I do get some errors with master when using any arrow key

If you have errors on IPython/ptk-master can you post them here ?

jorisvandenbossche:
Just opened an issue at ptk: https://github.com/jonathanslenders/python-prompt-toolkit/issues/316

jonathanslenders:
The bug regarding the pasting is probably in this key binding (Ctrl-J). I can reproduce it, but it doesn't happen for ptpython.

https://github.com/ipython/ipython/blob/master/IPython/terminal/ptshell.py#L183

jonathanslenders:
How do you feel if I can make pasting on Windows 99.9% accurate, but not 100%?

[...]

PR #9614, opened 2016-06-14
Ctrl-V paste on Windows

takluyver:
Allow users to paste in Windows with Ctrl-V, avoiding some issues with native paste in cmd, as discussed in gh-9600.

I'm only enabling this in Windows, because native bracketed paste should work well in Unix terminals. And it's disabled if you use vi editing mode, in case it conflicts with other shortcuts. There probably aren't many Windows users who configure vi mode, in any case.

I haven't tested this yet. @jorisvandenbossche , @klonuo , @stonebig : do you have time to test it?

### DIFF
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -288,7 +288,7 @@ class TerminalInteractiveShell(InteractiveShell):
                         return
                 except ClipboardEmpty:
                     return
-                event.current_buffer.insert_text(text)
+                event.current_buffer.insert_text(text.replace('\t', ' ' * 4))
 
         # Pre-populate history from IPython's history database
         history = InMemoryHistory()
--- a/IPython/terminal/ptshell.py
+++ b/IPython/terminal/ptshell.py
@@ -14,7 +14,9 @@ from IPython.utils.process import abbrev_cwd
 from traitlets import Bool, Unicode, Dict, Integer, observe, Instance, Type, default
 
 from prompt_toolkit.enums import DEFAULT_BUFFER, SEARCH_BUFFER, EditingMode
-from prompt_toolkit.filters import HasFocus, HasSelection, Condition, ViInsertMode, EmacsInsertMode, IsDone, HasCompletions
+from prompt_toolkit.filters import (HasFocus, HasSelection, Condition,
+    ViInsertMode, EmacsInsertMode, IsDone, HasCompletions)
+from prompt_toolkit.filters.cli import ViMode
 from prompt_toolkit.history import InMemoryHistory
 from prompt_toolkit.shortcuts import create_prompt_application, create_eventloop, create_prompt_layout
 from prompt_toolkit.interface import CommandLineInterface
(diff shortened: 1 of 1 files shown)


stonebig:
hum, what do you mean by "terminal" ? jupyter-console ?

takluyver:
No, this is nothing to do with Jupyter - it's for pasting code into IPython, in the new prompt_toolkit interface. It's supposed to be another way around the problem described in #9600.

klonuo:
I tested by pasting indented code blocks. Space indented paste feels great, but tab indented does not look good, i.e.:

```python
In [1]: if sys.version_info[0] == 3:  # Python 3
   ...: ^Ikwargs['encoding'] = 'utf8'
   ...:
```
So instead tab there is `^I`

jorisvandenbossche:
Tested it, and seems to work nicely! Thanks @takluyver 

Do you see this as a solution for #9600? Because although it is nice that Ctrl-V works, I think it is important that right-clicking paste also works (as this is how to paste normally in the command line, and people will do this)

takluyver:
Thanks all.

I see this as a partial solution to #9600. Making native paste work is up to @jonathanslenders, and he's doing so, but it involves some amount of heuristics, so it can never be 100% accurate. By providing another way to paste, I hope that fewer people will run into edge cases. Being able to paste with the common keyboard shortcut is also very convenient.

@klonuo can you try tweaking the code to add `.replace('\t', '    ')`, and see if that makes it work with code containing tabs?

takluyver:
That code sample is meant to be `.replace('\t', '<4 spaces>')`. For some reason Github seems to be collapsing multiple spaces today. :angry:

jorisvandenbossche:
> Being able to paste with the common keyboard shortcut is also very convenient.

That certainly is! 
Would making Ctrl-C work for paste also be possible?

takluyver:
> Would making Ctrl-C work for paste also be possible?

You mean copy, right? ;-)

It probably is, but we don't have equivalent code to push to the clipboard, so it's a bit trickier. We could depend on pyperclip for it, though. It might also be counterintuitive, because I don't think we can easily get the text you've selected with the mouse, so it would only work with prompt_toolkit's selection (by default, Ctrl-space to enter selection mode). Overall, I think that's best left to work out in a future release.

jorisvandenbossche:
> You mean copy, right? ;-)

Indeed :-) But it is certainly not critical as a working multi-line paste

IPython/terminal/ptshell.py
```diff
@@ -271,6 +273,23 @@ def cursor_in_leading_ws(cli):
...
+                try:
+                    text = win32_clipboard_get()
+                except TryNext:
+                    try:
+                        text = tkinter_clipboard_get()
+                    except (TryNext, ClipboardEmpty):
+                        return
+                except ClipboardEmpty:
+                    return
+                event.current_buffer.insert_text(text)
```
klonuo: @takluyver `event.current_buffer.insert_text(text.replace('\t', ' ' * 4))` works fine here

takluyver: Thanks, updated the PR.

Carreau:
thanks @takluyver !

randy3k:
How to enable Mac terminal bracketed paste mode? It doesn't seem enabled.

takluyver:
It should be in the 5.0 beta releases as well. You need a terminal emulator that supports it, and then it should just work.

Issue #9589, opened 2016-06-09
double quotes do not appear immediately


anntzer:
Typing `"` at the prompt does not print anything -- I need to type another character and both the quote and the following character will appear at the same time.  This is a bit annoying as I am used to typing `""` <left arrow> and fill in the string after (which is still possible... just weird).
This does not happen with single quotes (`'`).

takluyver:
That sounds like something is trying to do combining characters, e.g. " + a -> ä.

This isn't happening for me - can you give more details about your system that might be relevant? Can you try with other terminal emulators?

anntzer:
Arch Linux, tested both with Konsole and xterm.
Typically using a French keyboard layout but switching to US (via setxkbmap) did not help.
Note that it doesn't actually print out a combining character even if I type e.g. `"a`.

Carreau:
That's super weird... let me see if I know someone with a French keyboard local and a linux computer.

Carreau:
@parleur and/or @NelleV if you have a Linux Computer + French keyboard can you try the beta2 of IPython and try to type a double quote?

```
$ pip install ipython --pre
```

Thanks !

NelleV:
I have no problem with the double quote on linux, using bépo.

NelleV:
It works fine using an azerty as well.

Carreau:
Thanks @NelleV 

Humm... @anntzer what locale do you have ? IIRC you are close to Berkeley we might be able to debug that in person. 

I'm going to mark that as "need-info" and not block releases on it for now.

anntzer:
I can drop by D-lab tomorrow if you're around.

Carreau:
Hum Friday might be a bit busy, especially afternoon. You can try to ping me and I can tell you whether I'm available, but it will likely be hard.

anntzer:
How's late morning today like?

anntzer:
11am in front of Barrows?

Carreau:
> 11am in front of Barrows?

I'm there  (north entrance) on the wood bench.

Carreau:
So it's triggered by `--TerminalInteractiveShell.editing_mode='vi'`, @matthew-brett I'm curious why you haven't reported this one ? Or do you never use double quotes ? (can you reproduce it ?)

Carreau:
(that seem to be fixed on ptk master)

matthew-brett:
I usually use single quotes, but I can reproduce it.

anntzer:
[...]

Issue #9540, opened 2016-06-05
Completion usability: don't pop up dialog for single completion


fperez:
Tab completion currently has a serious usability problem compared to the old (albeit more limited) readline one: when there's a single completion it still pops up the dialog, which is very distracting as it stops your flow to read irrelevant information.

We should fix this before 5.0 goes out...

takluyver:
Ping @jonathanslenders

Carreau:
Is that critical ? the fix is 2 lines in prompt toolkit, and needs us to subclass Buffer and hack quite a bit around in IPython itself ?

Carreau:
There will though be another annoying side effect, likely due to an IPython "bug". 
As PTK, is complete "as you type", typing `im<tab>` would autocomplete to `import`, and try to keep completing. But at this stage, as `import|` try to complete, return **only** `importlib`, so in the end `imp<tab>` complete to `importlib`. 

So the fix is likely a bit more involved.

jonathanslenders:
See: https://github.com/jonathanslenders/python-prompt-toolkit/pull/311
I made a fix, but with some remarks.

Carreau:
Marking as 5.1, as non critical.

jorisvandenbossche:
A related remark about this: due to the completion dialog popping up, I cannot just 'enter' to execute the command.

An example workflow (to check the version of a library, but of course is legit for all cases where there is only one possibility but you just want to complete to save typing) which worked previously: 
type `pd.__v` 
-> press 'tab' (completed to `pd.__version__`) 
-> press 'enter' 
-> you get the result

This workflow is now broken, because you get a completion dialog after pressing tab, and then Enter does not work anymore to execute. I either have to press 'Escape' to remove the dialog, or to press 'down button' to select the the first and only completion suggestion, before 'Enter' gives you the result.

Personally I find this quite disturbing for this (I think common) workflow (and thus rather critical for 5.0)

Tested with 5.0.0b3 on Windows

Carreau:
> A related remark about this: due to the completion dialog popping up, I cannot just 'enter' to execute the command.
> 
> An example workflow (to check the version of a library, but of course is legit for all cases where there is only one possibility but you just want to complete to save typing) which worked previously: 
> type pd.__v 
> -> press 'tab' (completed to pd.__version__) 
> -> press 'enter' 
> -> you get the result
> 
[...]

PR #9624, opened 2016-06-15
Pin prompt_toolkit to >=1.0.1

Carreau:


Carreau:
Closes #9605 #9589 #9540

fperez:
Great, I just tested and PT 1.0.1, and it's already much nicer! My old muscle memory now works again much more smoothly. Let's go ahead, merging this.

PR #9626, opened 2016-06-15
Undeprecate interactiveshell

Carreau:
Rebase of #9575

Issue #9597, opened 2016-06-10
PTK tab completion with many alternative hard to review and navigate


matthew-brett:
With the new PTK autocompletion, even with:

```
ipython --TerminalInteractiveShell.display_completions_in_columns=True
```

and after Control-L, the result is a fairly large array of potential completions but:

a) It's not nearly large enough for big namespaces like numpy, and there is no way that I can see the entire list blurted out to the terminal, as was possible with readline and (related);
b) It is hard to navigate the list.   Cursor keys and Control-N/P work to move single elements, but in large panes of options, this is - er - painful.

minrk:
Should completions-in-columns be the default?

klonuo:
related: #9555

matthew-brett:
I guess completion in columns would be a good default, in that it's more similar to the readline version.  I think it's usually what you want, as well.

Carreau:
> Should completions-in-columns be the default?

I would not be against.

takluyver:
One potential reason not to do completions in columns is that we're looking at providing extra info next to each completion. That presumably only works with a single column of completions. So if people get used to multi-column completions now, we may later have to either change the default or hide the new feature behind a setting no-one's likely to find.

It's possibly not a good enough reason to stop us doing it, but it's worth considering.

matthew-brett:
Related - and y'all probably know this - but you can't use the right / left cursor keys to navigate in the pane.  So if I tab complete on `np.`, even after Control-L, I see down to `cos`, but to get to `cosh` I need to do in the order of 100 Control-Ns.

jonathanslenders:
```
but you can't use the right / left cursor keys to navigate in the pane. 
```

This is on my list of improvements for a next version of prompt-toolkit. It's not easy to do right now, because of the decoupling between the UI and the key bindings + completion code. In a next version, this should be easier to implement.
However, adding a key binding that goes forward/backward 10 completions at the time should already be possible.

hainm:
> Should completions-in-columns be the default?

Please please. I find the default completion is very annoying (vs my old habit seeing all the contents at once).

Carreau:
Closing now that we have the multicolumn by default and the 'readlinelike' version.

PR #9627, opened 2016-06-15
 Extend the completion layout to use a third readlinelike option

Carreau:
```
The `display_completions_in_columns` thus becomes deprecated, and the
Boolean is replaced by an Enum: `column`, `multicolumn`, `readlinelike`

which give us more freedom to implement extra-layout later.
```

---

On top of #9626, should help with #9597

I can see 4th layout for 6.0 that would be like readline, but where the cursor would move through the completions without redrawing the prompts. Much like QtConsole.

I also make multicolumn the default as seem to be popular.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -148,8 +148,9 @@ class TerminalInteractiveShell(InteractiveShell):
         help="Automatically set the terminal title"
     ).tag(config=True)
 
+    # Leaving that for beta/rc tester, shoudl remove for 5.0.0 final. 
     display_completions_in_columns = Bool(None,
-        help="Display a multi column completion menu.", allow_none=True
+        help="DEPRECATED", allow_none=True
     ).tag(config=True)
 
     @observe('display_completions_in_columns')


IPython/terminal/interactiveshell.py
```diff
@@ -1,21 +1,589 @@
...
+    term_title = Bool(True,
+        help="Automatically set the terminal title"
+    ).tag(config=True)
+
+    display_completions_in_columns = Bool(None,
+        help="Display a multi column completion menu.", allow_none=True
+    ).tag(config=True)
+
+    @observe('display_completions_in_columns')
+    def _display_completions_in_columns_changed(self, new):
```
minrk: Do we need a deprecation for an unreleased feature that's only been in master briefly?

Carreau: No, i'll remove it post beta. I know of a few person that have set up this
option. It's (relatively) cheap to keep.

Carreau:
Oops, one commits of this sneaked into #9626 .. in particular 0668587b74996a0c1bb043687902c5d0738e177a. Apologies,

PR #9631, opened 2016-06-16
Fix interact() backward compat.

Carreau:
Ping @asmeurer

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -35,6 +35,8 @@ from .pt_inputhooks import get_inputhook_func
 from .prompts import Prompts, ClassicPrompts, RichPromptDisplayHook
 from .ptutils import IPythonPTCompleter, IPythonPTLexer
 
+DISPLAY_BANNER_DEPRECATED = object()
+
 
 def get_default_editor():
     try:
@@ -455,7 +457,11 @@ class TerminalInteractiveShell(InteractiveShell):
             self.pt_cli.application.buffer.text = cast_unicode_py2(self.rl_next_input)
             self.rl_next_input = None
 
-    def interact(self):
+    def interact(self, display_banner=DISPLAY_BANNER_DEPRECATED):
+
+        if display_banner is not DISPLAY_BANNER_DEPRECATED:
+            warn('interact `display_banner` argument is deprecated since IPython 5.0. Call `show_banner()` if needed.', DeprecationWarning, stacklevel=2)
+
         while self.keep_running:
             print(self.separate_in, end='')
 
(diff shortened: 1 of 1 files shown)


asmeurer:
Maybe I'm testing it wrong, but it doesn't seem to fix the issue for me.

Carreau:
What about now ?

asmeurer:
Yep, seems to fix it. Do we need to make any changes on the SymPy side for future versions? No deprecation are shown when I start `isympy` with this (which I am :+1: to).

Carreau:
We might drop the kw at some point, but I don't know when.

I'm usually the kind of people that cleanup deprecation pretty aggressively, but many things often miss the cleanup deadline  and and up staying in the codebase for a few years.

asmeurer:
Well, as I said, I didn't see any warnings. I'm all for deprecating and removing things, but you have to make your deprecations actually seen. Otherwise, it's a waste of time.

For SymPy, we have a special `SymPyDeprecationWarning` subclass of `DeprecationWarning`, which we enable always.

Carreau:
> Well, as I said, I didn't see any warnings. I'm all for deprecating and removing things, but you have to make your deprecations actually seen.

We did that, it had the opposite effect, as:
- If your isympy users update IPython, they get warnings until you release a new version
- which can take a long time.
- Thus depends project/users would add a "ignore all warnings" not to get complaints:
- leading to actually nothing being ever shown.

We go a long way to enable any DeprecationWarning in IPython repl **by default** but only when emitted from user code. 

Plus you seem you seem to have a few filterwarnings, why not the following:

```python
import warnings
from sympy import __version__
if 'dev' in __version__:
    action ='default'
    travis = (os.environ.get('TRAVIS','') == 'true')
    if travis:
        action = 'error'
    warnings.filterwarnings(action, message='.*', category=DeprecationWarning, module='sympy.*')
    if [int(n) for n in __version__.split('.')[0:2]] < (2,0):
        # muffle  inspect.getargspec() deprecated for now, might want to have a look later.
        warnings.filterwarnings('ignore', category=DeprecationWarning, module='.*',
                        message='.*inspect.getargspec.*')
```

Such that : 
  1) Only devs see these warnigns, 
  2) You only see the warnigns that are relevant to sympy ?

Plus you would catch deprecation warning from other things than IPython.

Carreau:
For reference: #8478

asmeurer:
I suppose if someone wants to silence a warning, that's their prerogative/fault, but we want everyone to at least have a chance to see them.  It's also *extremely important* that all deprecation warnings be silencable by replacing code with something non-deprecated (otherwise use `PendingDeprecationWarning`, for which it totally makes sense to be off by default).

> If your isympy users update IPython, they get warnings until you release a new version

But the other option (assuming we hadn't caught this now) would be no one notices that anything is wrong, until IPython releases a version that removes the deprecated feature, which completely breaks SymPy. The warnings way, folks bug us, and if the bug us enough, we'll be motivated to do a release.  But the breaking way, we have to do a bugfix release right away.

Carreau:
@asmeurer I hear you, but that's unreasonable to ask from all the users that are not developpers, and assuming you want to enable it for sympy then `filterwarnings('default', '.*', DeprecationWarning, module='sympy.*')` will enable that for all your user, and only at Sympy level, so you will be warned if sympy uses deprecated feature, regardless of whether it's from IPython or another package.

Carreau:
Opened https://github.com/sympy/sympy/pull/11249  as a proof of concept. I **think** this is a good compromise, let's see if it show more deprecation in the test-suite. 

If that would really help you we **could** make a subclass of `DeprecationWarning` in IPython but we would likely not enable it by default.

Issue #9632, opened 2016-06-17
Test failure on Windows


takluyver:
@anntzer , can you take a look? I suspect this was introduced by #9516.

```
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\10113\lib\unittest\case.py", line 58, in testPartExecutor
    yield
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\10113\lib\unittest\case.py", line 600, in run
    testMethod()
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\10113\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "c:\jenkins\workspace\windows-multi\_pyversion\3.5\label\windows\IPython\core\tests\test_completer.py", line 74, in test_protect_filename
    nt.assert_equal(s1p, s2)
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\10113\lib\unittest\case.py", line 820, in assertEqual
    assertion_func(first, second, msg=msg)
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\10113\lib\unittest\case.py", line 1193, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\10113\lib\unittest\case.py", line 665, in fail
    raise self.failureException(msg)
'" abc"' != "' abc'"
- " abc"
+ ' abc'

    """Fail immediately, with the given message."""
>>  raise self.failureException('\'" abc"\' != "\' abc\'"\n- " abc"\n+ \' abc\'\n')
```

From Jenkins test report here: https://jenkins.jupyter.org/job/windows-multi/645/testReport/

takluyver:
Fixed by #9634. There are several more failures now, but that's a new issue.

PR #9634, opened 2016-06-17
Fix quoting tests on Windows.

anntzer:
Quoting should be done with double quotes, not single quotes.

See #9632.


takluyver:
Thanks :-)

PR #9647, opened 2016-06-21
Remove the "autoedit_syntax" feature.

### LINKED ISSUE #9603: "return to editor to correct syntax error" is annoying
When I type a syntax error in the IPython console, like `def test()` without the colon, I get a prompt 

```
In [1]: def test
  File "<ipython-input-1-50eadd3e41eb>", line 1
    def test
            ^
SyntaxError: invalid syntax

Return to editor to correct syntax error? [Y/n]
```

This is annoying, and seems completely unnecessary, since with prompt_toolkit you now can reasonable edit the input in the console.

Carreau:
IPython used to (a long time ago, in a galaxy far far away) have the
ability to automatically open an editor in case a wild syntax error
appears. The configuration option to enable that was not working for a
few years, and apparently we by mistake re enabled it, to discover that
the feature is actually broken.

So this plainly remove the code to support this feature, at the
exception of the `fix_error_editor` hook. Indeed it is public API, so
for now  as it seem to be used only for this feature, we'll just raise a
UserWarning (there is roughly 0 chance of this being tested as it's used
mostly interactively, so DeprecationWarnings would be unseen).

We'll remove later if no complaints

Closes #9603

### DIFF
--- a/IPython/core/hooks.py
+++ b/IPython/core/hooks.py
@@ -37,6 +37,7 @@ example, you could use a startup file like this::
 
 import os
 import subprocess
+import warnings
 import sys
 
 from IPython.core.error import TryNext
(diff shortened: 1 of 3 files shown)


IPython/core/hooks.py
```diff
@@ -90,6 +91,14 @@ def fix_error_editor(self,filename,linenum,column,msg):
 
     Call ip.set_hook('fix_error_editor',youfunc) to use your own function,
     """
+
+    warnings.warn("""
+`fix_error_editor` is pending deprecation as of IPython 5.0 and will be removed
+in future versions. It appear to be used only for automatically fixing syntax
+error that has been broken for a few years and has thus been removed. If you
+append to use this function and still need it please make your voice heard.
```
takluyver: append -> happened

Also, let's be more specific about 'make your voice heard', e.g. point them to the issue tracker.

IPython/core/hooks.py
```diff
@@ -90,6 +91,14 @@ def fix_error_editor(self,filename,linenum,column,msg):
 
     Call ip.set_hook('fix_error_editor',youfunc) to use your own function,
     """
+
+    warnings.warn("""
+`fix_error_editor` is pending deprecation as of IPython 5.0 and will be removed
+in future versions. It appear to be used only for automatically fixing syntax
```
takluyver: appears

IPython/core/hooks.py
```diff
@@ -90,6 +91,14 @@ def fix_error_editor(self,filename,linenum,column,msg):
 
     Call ip.set_hook('fix_error_editor',youfunc) to use your own function,
     """
```
takluyver: Let's also note the deprecation in the docstring.

takluyver:
+1 from me to the removal

IPython/core/hooks.py
Carreau: Rhooo this one I should not get wrong.

Carreau: Opened issues and added link to this specific issue and to ML.

IPython/core/hooks.py
Carreau: DOne.

PR #9655, opened 2016-06-23
Restore the ability to use %colors to switch terminal theme.

### LINKED ISSUE #9648: `--colors` flag seems to be ignored, `%colors` doesn't work
See:

image

I can't seem to get the no-color or LightBG schemes to work. I know there's a new color system in place, but

a) the old ones should be grandfathered in and still work
b) nocolor should still be available
c) the `%colors` magic still only lists the old ones and doesn't actually work.

Carreau:
And restore previous coloring of prompts on classical color themes.
Unlike 4.x this will **not** change with your terminal background.

If the hightligh_style of TerminalInteractiveSHell is set to `legacy`
then it uses the value of TermianlInteractiveShell.colors to select the
theme:

monokai for darkbg/linux (by decision of BDFL), and old prompt values.
default for lightbg

Closes #9648

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -358,6 +358,8 @@ Defaulting color scheme to 'NoColor'"""
         # Set exception colors
         try:
             shell.InteractiveTB.set_colors(scheme = new_scheme)
+            shell.colors = new_scheme
+            shell.refresh_style()
             shell.SyntaxTB.set_colors(scheme = new_scheme)
         except:
             color_switch_err('exception')
--- a/docs/source/whatsnew/version5.rst
+++ b/docs/source/whatsnew/version5.rst
@@ -8,7 +8,7 @@ IPython 5.0
 Released June, 2016
 
 IPython 5.0 now uses `prompt-toolkit` for the command line interface, thus
-allowing real multi-line editing and syntactic coloration as you type. 
+allowing real multi-line editing and syntactic coloration as you type.
 
 
 When using IPython as a subprocess, like for emacs inferior-shell, IPython can
@@ -20,7 +20,7 @@ Backwards incompatible changes
 ------------------------------
 
 
-The `install_ext magic` function which was deprecated since 4.0 have now been deleted. 
+The `install_ext magic` function which was deprecated since 4.0 have now been deleted.
 You can still distribute and install extension as packages on PyPI.
 
 Update IPython event triggering to ensure callback registration and
@@ -63,6 +63,13 @@ See `TerminalInteractiveShell.prompts` configurable for how to setup your prompt
 Most of the above remarks also affect `IPython.core.debugger.Pdb`, the `%debug`
 and `%pdb` magic which do not use readline anymore either.
 
+The color handling has been slightly changed and is now exposed
+through, in particular the colors of prompts and as you type
+highlighting can be affected by :
+``TerminalInteractiveShell.highlight_style``. With default
+configuration the ``--colors`` flag and ``%colors`` magic behavior
+should be mostly unchanged. See the `colors <termcolour>`_ section of
+our documentation
 
 Provisional Changes
 -------------------
--- a/docs/source/whatsnew/version5.rst
+++ b/docs/source/whatsnew/version5.rst
@@ -7,13 +7,13 @@ IPython 5.0
 
 Released June, 2016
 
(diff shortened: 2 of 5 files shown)


Carreau:
@fperez would you like to try/review this ? If this suits you I'll document it.

IPython/core/magics/basic.py
```diff
@@ -324,6 +324,7 @@ def colors(self, parameter_s=''):
           %colors nocolor
         """
         def color_switch_err(name):
+            raise ValueError()
```
fperez: Why the change to an exception?

Carreau: Oops, debug thing.

fperez:
Mmh, there's still a bug, though it looks like one we've had since 4.x (but not in 3.x): if I have colors assigned in my config file, then using `--colors` doesn't work (the prompt color should change below):

image


In 3.x it worked fine (no need to bisect when the bug appeared, but I see it in 4.x too):

image

fperez:
BTW, I even think you it would be cool to leave the new default prompt colors unchanged for the "linux" case: the highlight of the prompt numbers is very nice. So we can modernize them a bit, all I want to get is the ability to control the three 'basic' schemes (nothing, dark, light) from the command line including prompts...

In summary:

- for 'linux', we can use the current green with brighter number default (green/red)
- for 'lightbg', I'd do something similar, with a brighter/bolder number (blue/red)

And otherwise apply the fixes for `--colors` to work.

fperez:
Confirming the problem about not honoring the command-line override is a traitlets bug, as suspected by @Carreau... Rolling traitlets back from master to 4.0 did the trick. So let's forget about that part as an ipython issue.

Carreau:
> Confirming the problem about not honoring the command-line override is a traitlets bug, as suspected by @Carreau... Rolling traitlets back from master to 4.0 did the trick. So let's forget about that part as an ipython issue.

The question is was this bug release in any of the stable versions of traitlets ?

fperez:
yes, it appeared between 4.0 and 4.1, it's present in 4.1.0.

Carreau:
Opened as https://github.com/ipython/traitlets/issues/248

fperez:
Great, so the last question is: do you have the patience to make the last modifications I suggested (using the current default colors for linux and a blue/red version of the same --with brighter prompt numbers-- for lightbg)?  

We could merge as-is, but I think that will be both cleaner (as it only introduces *one* new set of prompt colors, not two) and nicer (as it does modernize things a bit and diminishes the mismatch between using `--colors` and manually selecting highlight styles).

Carreau:
Only problem is "monokai" is unreadable on lighbg:

colors

Pushed your required changes and docs. 

@willingc do you want to get a looks at the docs additions I made ?

fperez:
I think it's OK: the default one is a good compromise for light/dark, where as the explicit request for light/dark themes will optimize specifically for those.

jbasalone:
👍

Carreau:
Thanks @jbasalone  and @fperez. Rebase for conflict and pushed. 
@willingc assigned to have a look at the docs.

docs/source/whatsnew/version5.rst
```diff
@@ -20,7 +20,7 @@ Backwards incompatible changes
 ------------------------------
 
 
-The `install_ext magic` function which was deprecated since 4.0 have now been deleted. 
+The `install_ext magic` function which was deprecated since 4.0 have now been deleted.
```
willingc: have -> has

Carreau: Done.

Carreau: I give you access to my fork so if needed you can make changes directly.

willingc:
@Carreau I'm making the rest of the edits and will push to your branch. Seems simpler :-)

Carreau:
> @Carreau I'm making the rest of the edits and will push to your branch. Seems simpler :-)

Thanks ! Seems great ! Appreciated !

Carreau:
Thanks @willingc !

Carreau:
Ok, self merging. I'll try to do an RC this afternoon.

fperez:
Thanks everyone, kudos to @willingc for your prompt (no pun intended :) doc work. Love seeing good code landing with proper docs.  Next thing you know, our project-wide practices will actually have improved!! ;)

PR #9660, opened 2016-06-24
Don't rely upon traitlets copying self.config

minrk:
when saving CLI config to reload later for highest priority.

traitlets 4.1-4.2.1 don't recreate self.config in update_config (rightly so), so when saving CLI config for later re-loading, make sure it's a copy so it doesn't get changed during config file loading, which defeats the purpose.

To me, this is the real fix for ipython/traitlets#248

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -13,6 +13,7 @@ object and then create the configurable objects, passing the config to them.
 # Distributed under the terms of the Modified BSD License.
 
 import atexit
+from copy import deepcopy
 import glob
 import logging
 import os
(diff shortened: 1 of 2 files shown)


minrk:
This could be considered for 4.x backport, if it's considered a high priority.

Carreau:
Good point.

fperez:
the actual code amount is fairly small, and it is a regression in 4.x, so +0.5 on backporting. But given our thinly spread resources, I won't cry if you decide not to...

Carreau:
FOr info, backported.

PR #9662, opened 2016-06-24
colors in interactive shell

adamgreenhall:
I'd like color in my embedded shell. Maybe there's a good reason why this is the default, but it seems strange that this is the only override.


Carreau:
Hey, 
Sorry for the delay, this PR slipped through the crack, thanks !

Indeed from the git history the deactivation was a temporary workaround. It seem reasonable to reenable colors !

PR #9664, opened 2016-06-25
Added keyboard shortcuts docs

### LINKED ISSUE #9623: Documenting keyboard shurtcuts
IMHO it would be sensible if we can provide description to newly introduced keyboard shortcuts.

Personally I hunted Google to learn about Ctrl+O, but there seem to be many other functions exposed with default shortcuts.

Ctrl+S / Ctrl+R =  `I-search` - seems to search some parts of history
Ctrl+Space = `Selection mode` - already discussed as feature from modern emacs bindings
...
Many other Ctrl+? bindings at least. 

So my suggestion is that we should provide reference to used bindings in documentation. At least just those exposed from PTK

klonuo:
Closes #9623


Carreau:
Great thanks ! I'll have a look at that !

@willingc might want to have a look as well.

Carreau:
@takluyver you had a pass at documentation recently, and are a bit more aware that I on the autogenerated part.

Carreau:
I've looked at whether we can fake a CLI instance  and triage based on that, but that seem a bit complicated. Doable, but complicated. 

I'm considering pre-importing `prompt_toolkit`, and monkey-patching to replace the filters and have finer-grained control of when they return True/False. to triage more easily. still turning that in my head. 

The diff appears to be wrong for whatever reason on GitHub, I'm going to mer that as it is an improvement, and we can iterate on subsequent PR. 

Thanks !

Issue #9639, opened 2016-06-18
Several test failures on Windows relating to win_unicode_console


takluyver:
#9621 introduced several failures ending with the same error from `win_unicode_console`. It looks like it's assuming that `sys.std*` objects will always have a fileno. @Drekin , I'm not sure if we can easily fix this from our side - any chance that `win_unicode_console` could catch the AttributeError and do something sensible?

Ping @klonuo as well.

```
  File "c:\jenkins\workspace\windows-multi\_pyversion\2.7\label\windows\IPython\utils\py3compat.py", line 219, in input
    return builtin_mod.raw_input(prompt)
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\21469\lib\site-packages\win_unicode_console\raw_input.py", line 81, in raw_input
    tty = check_stream(sys.stdin, STDIN_FILENO) and check_stream(sys.stdout, STDOUT_FILENO)
  File "C:\Users\ADMINI~1\AppData\Local\Temp\2\21469\lib\site-packages\win_unicode_console\streams.py", line 273, in check_stream
    _fileno = stream.fileno()
AttributeError: StringIO instance has no attribute 'fileno'
```

[Jenkins test report](https://jenkins.jupyter.org/job/windows-multi/647/testReport/)

Drekin:
In Python 3 `StringIO` has `fileno` raising `io.UnsupportedOperation`, which is catched. I can easily add a check for `AttributeError` as well. There were some other changes in `win_unicode_console` since the last release, but I'll try to put together the new release soon.

Drekin:
Can you use ˋwin_unicode_console` from the `Developement` branch to run your tests?

takluyver:
That seems to have [done it](https://jenkins.jupyter.org/job/windows-multi/649/), thanks! Can you let us know when you release, because we'll need to either wait for that or pull the win_unicode_console stuff back out for our 5.0 release.

Drekin:
Yes, I'll let you know. I'll try to put it together soon.

Drekin:
win_unicode_console 0.5 is out.

takluyver:
Thanks, tests pass with the new version.

PR #9665, opened 2016-06-26
Add WUC version requirement in setup.py

klonuo:
By adding WUC #9621 several test were failing #9639

@Drekin released WUC 0.5 which addresses these errors.

I can confirm WUC 0.5 resolves test failures on Windows with Python 2.7


takluyver:
Thanks!

Issue #9670, opened 2016-06-28
Default syntax higlighting colours are unreadable on light background


takluyver:
![screenshot from 2016-06-28 15-33-23](https://cloud.githubusercontent.com/assets/327925/16419469/c5c44d7c-3d45-11e6-8e1d-f3b7b2ed27b2.png)

Ping @Carreau @fperez 

I had tried to pick a set of colours that are moderately passable on both light and dark background terminals, so I'm kind of annoyed to start IPython today and find that it's assuming a dark background again.

Tracing the code paths through, it looks like you've moved my colour picks under the 'lightbg' option, which is only the default on macs.

takluyver:
Following comment on #1272

takluyver:
See #9672 which leaves the prompt colours as they were but uses the same set of highlighting colours for the default on all platforms.

Carreau:
> I had tried to pick a set of colours that are moderately passable on both light and dark background terminals, so I'm kind of annoyed to start IPython today and find that it's assuming a dark background again.

Sorry about that, the colors seemed pretty bad on @fperez screen, so we decided to use stronger colors.
The reasoning to switch prompt colors is to give feedback when you use `%colors`. 
I honestly forgot the default was different on various platforms. I think you are right that dark theme should be readable on white bg if it can be the default.

takluyver:
I'm fine with having two sets of prompt colours, but I think the default colours really need to be readable both ways round. There's no way of working out which one we're on, and there's a lot of people using each.

I'm sure the colours I picked aren't great, and I'm happy if we can find some more pleasing ones. But in the tests I've done, they are basically legible either way, and I think that has to be a constraint for the defaults.

fperez:
I mostly wanted the explicit 'Linux' color scheme (which now is *not* the default as it used to) to be  really optimized for dark bg, but I fully agree that the default should be readable on both, sorry if in the last updates we didn't cross-check enough.

Carreau:
That the problem, Linux is **still** default on windows and Linux...

takluyver:
The default didn't change, your OS changed ;-)

fperez:
Ah, got it! Well, in any case, the 'default' being different from Linux/LightBG as in #9674, looks like a good solution to me.

Carreau:
@takluyver if you are happy can we close that ?

takluyver:
Yup, thanks.

PR #9672, opened 2016-06-28
Ensure default highlighting is visible in light & dark backgrounds

takluyver:
Closes gh-9670

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -371,28 +371,27 @@ class TerminalInteractiveShell(InteractiveShell):
         """
         if name == 'legacy':
             legacy = self.colors.lower()
-            if legacy == 'linux':
-                style_cls = get_style_by_name('monokai')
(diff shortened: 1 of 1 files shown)


Carreau:
Ping @fperez who really wanted a stronger coloring on dark background, as on his screen the dark theme + default is close to unreadable

Carreau:
the 3rd option is to add a default 'color' which is "neutral" and make that the default, so that `color=linux` and `color=lightBg` both can **increase** the contrast.

fperez:
Yes, my last comment on #9670 was pretty much what @Carreau said above, I agree with that idea.

fperez:
`linux` and `ligthbg` are really user-optimized "forced dark" or "forced light" background choices and there for backwards compatibility. 

Given the impossibility of doing background detection, I agree with having the default be a reasonable compromise for both.

Carreau:
See #9674 that would likely make everyone happy.

takluyver:
Closing as we merged #9674 instead.

PR #9673, opened 2016-06-28
Remove colors_force attribute

takluyver:
We originally checked for readline to enable colour. When we created the kernel, the `colors_force` attribute was added to skip this check and always enable colour in the kernel.

Now both the kernel and the terminal interface are readline-free with colour enabled, so this flag is no longer needed.


Carreau:
+1

PR #9677, opened 2016-06-29
Some docs work to prepare for 5.0

takluyver:
- Improve release notes
- Simplify docs on terminal colours
- Add info on customising prompts
- Update docs on integrating event loops

This also adds back an interface to add input hooks for other event loops beyond those we define.


minrk:
Very nice. cc @willingc for docs/release review, but 👍 from me.

takluyver:
I've also reverted part of #9567 - I hadn't had time to look closely at it before, but I prefer having separate pages for the kernel and terminal options rather than one big page.

docs/source/whatsnew/version5.rst
```diff
@@ -5,71 +5,69 @@
...
+  event. See :ghissue:`9447` and :ghpull:`9453`.
+- Integration with pydb has been removed since pydb development has been stopped
+  since 2012, and pydb is not installable from PyPI.
+- The ``autoedit_syntax`` option has apparently been broken for many years.
+  It has been removed.
+
+New terminal interface
+~~~~~~~~~~~~~~~~~~~~~~
+
+The overhaul of the terminal interface will probably cause a range of minor issues;
```
willingc: Really like this section for the user 👍 

Suggested text, but fine as is too.

The terminal interface's overhaul (or update) may cause a range of minor issues for existing users. This is inevitable for such a significant change, and we've done our best to minimize these issues. Some issues that we are aware of as well as some suggestions for the user:

docs/source/config/details.rst
```diff
@@ -2,89 +2,85 @@
...
+Custom Prompts
+==============
+
+.. versionchanged:: 5.0
+
+From IPython 5, prompts are produced as a list of Pygments tokens, which are
+tuples of (token_type, text). You can customise prompts by writing a method
+which generates a list of tokens.
+
+There are four kinds of prompt:
```
willingc: Not sure if this should be prompt or prompts. Likely could be either. No need to revise if you are fine with prompt.

willingc:
@takluyver Nice!

A couple of minor comments. +1 to merging. Someone with permissions on the repo will need to do the merge.

docs/source/whatsnew/version5.rst
takluyver: Reworded somewhere between my text and your suggestion.

docs/source/config/details.rst
takluyver: I think it could be either - it sounds slightly better in my head as is.

PR #9680, opened 2016-06-30
Improve docs, in particular shortcuts. 

Carreau:
Mostly cleanup and reorder documentation.

### DIFF
--- a/IPython/utils/warn.py
+++ b/IPython/utils/warn.py
@@ -10,11 +10,15 @@ from __future__ import print_function
 
 import sys
 import warnings
+from IPython.utils.decorators import undoc
 
 warnings.warn("The module IPython.utils.warn is deprecated since IPython 4.0, use the standard warnings module instead", DeprecationWarning)
 
+@undoc
 def warn(msg,level=2,exit_val=1):
-    """Standard warning printer. Gives formatting consistency.
+    """Deprecated
+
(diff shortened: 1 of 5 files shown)


IPython/utils/warn.py
```diff
@@ -10,11 +10,15 @@
 
 import sys
 import warnings
+from IPython.utils.decorators import undoc
 
 warnings.warn("The module IPython.utils.warn is deprecated since IPython 4.0, use the standard warnings module instead", DeprecationWarning)
 
+@undoc
```
takluyver: If the whole of this module should be undocumented, it's easier to add it as a skip pattern in `autogen_api.py` than to mark every function as undoc.

docs/autogen_shortcuts.py
```diff
@@ -5,6 +5,10 @@
```
takluyver: I haven't had a chance to look into this script yet, so I won't try to review changes to it - at some point I'll look through it as a whole.

IPython/utils/warn.py
Carreau: > If the whole of this module should be undocumented, it's easier to add it as a skip pattern in autogen_api.py than to mark every function as undoc.

Done.

Carreau:
Thanks !

PR #9681, opened 2016-06-30
Cleanup IPython bash completion from Jupyter commands.

Carreau:
Requested by @fperez.

### DIFF
--- a/examples/IPython Kernel/ipython-completion.bash
+++ b/examples/IPython Kernel/ipython-completion.bash
@@ -24,7 +24,7 @@ _ipython()
 {
     local cur=${COMP_WORDS[COMP_CWORD]}
     local prev=${COMP_WORDS[COMP_CWORD - 1]}
-    local subcommands="notebook qtconsole console kernel profile locate history nbconvert kernelspec install-nbextension trust "
+    local subcommands="kernel profile locate history kernelspec"
     local opts="help"
     if [ -z "$__ipython_complete_baseopts" ]; then
         _ipython_get_flags baseopts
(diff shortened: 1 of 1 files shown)


fperez:
Mmh, testing this on a local merge, I'm getting some weird behavior on bash/OSX... If I type `ipython <TAB>`, I get this:

image

Can you replicate this?  Any idea why those spurious 't' letters are appearing?

Carreau:
yep:

```
$ echo '- -' | sed -e 's/ / \t/g'
```

Oh...noooo OS X sed does not respect the gnu standard and does print a `t` instead of a tab but it work on linux.

Carreau:
Sidenote this is present in 4 as well.

fperez:
Yup, it's probably old, I just tested it now.  Remember, OSX userland is BSD, not GNU :)  Lots of little annoying differences like this...

fperez:
BTW, that means we don't necessarily have to fix this as a separate bug. We can merge this and leave the issue open if need be, at least the main ipython/jupyter cleanup got done, and we're no worse than before on this specific issue.

examples/IPython Kernel/ipython-completion.bash
```diff
@@ -24,7 +24,7 @@ _ipython()
 {
     local cur=${COMP_WORDS[COMP_CWORD]}
     local prev=${COMP_WORDS[COMP_CWORD - 1]}
-    local subcommands="notebook qtconsole console kernel profile locate history nbconvert kernelspec install-nbextension trust "
+    local subcommands="kernel profile locate history kernelspec"
```
minrk: kernelspec is also a jupyter command, not an IPython one

Carreau: > kernelspec is also a jupyter command, not an IPython one

Good call, done.

Issue #9658, opened 2016-06-24
Tab completion quirk


afshin:
The new `prompt_toolkit` completion is pretty sweet. I noticed an issue where it completes a word and then still opens a completion menu.

The process to get there was to type: `from ipywidgets im<TAB>`, and here is what happens:

![ipython](https://cloud.githubusercontent.com/assets/159529/16347143/0f9af2b2-3a43-11e6-8b7a-071da1347956.gif)

cc: @minrk @fperez

Carreau:
I think that kind-of make some sens as after import you **want** to import something, but I see the issue. 
cc @jonathanslenders

Carreau:
See jonathanslenders/python-prompt-toolkit#328 that IIUC should fix it. Will try to come up with a test.

fperez:
We discussed it today during JLab meeting and I agree with @afshin here (plus, that's how IPython 4.x used to work): getting immediately a list of *all possible* imports is nearly useless, as it's likely to be huge.  Most users will at least type a few characters to narrow the search to something more manageable, and then hit TAB. I think that pattern makes sense...

Carreau:
> We discussed it today during JLab meeting and I agree with @afshin here (plus, that's how IPython 4.x used to work): getting immediately a list of all possible imports is nearly useless, as it's likely to be huge. Most users will at least type a few characters to narrow the search to something more manageable, and then hit TAB. I think that pattern makes sense...

Yes I know, that why I said `I see the issue`  and send the PR `jonathanslenders/python-prompt-toolkit#328` . Note that it touches only the `from somthing i<tab>` as IPython completer return `import ` which has the 2 following specificities: 
 1) is the only completions
 2) has a trailing space.

Every case that do not match the 2 above conditions behave correctly and do not re-query the completer.

Carreau:
Note, edited my first comment to add explicit `<sarcasm>` tags.

fperez:
No worries :) And thanks for the follow-up on PT! I think we're getting very close to 5.0 being a *huge* improvement on 4.x, without anything to complain about for even the most nit-picky obsessives like myself :)

afshin:
@Carreau, thanks!

takluyver:
This was fixed in #9682, but not closed for some reason.

PR #9682, opened 2016-06-30
Do not emit single completions trailing with space.

Carreau:
Workaround #9658, that can thus be delayed post 5.0


minrk:
Works! I don't fully understand the comment about bumping #9658 post 5.0, though. Doesn't this fix it, and we can clean it up a bit once we require a later prompt_toolkit?

Carreau:
> Doesn't this fix it, and we can clean it up a bit once we require a later prompt_toolkit?

Yes it fixes it, but I consider stripping the whitespace part of a completion an undesirable feature.
So this is more a workaround for now, but we should not forget to undo that with a new PTK version.

I can rephrase.

Carreau:
Rephrased, and made the decorator private.

PR #9684, opened 2016-06-30
Remove readline related code. Pass 1

Carreau:
@takluyver


Carreau:
Note that this is quite a bit of removal of code that **should** be unused, though it's late in Beta cycle, so I'll understand if you think that much removal is time sensitive and should be delayed.

PR #9685, opened 2016-06-30
Remove readline , Pass II

Carreau:
On top of #9684 to make things easier to read.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -5,9 +5,6 @@ This module started as fork of the rlcompleter module in the Python standard
 library.  The original enhancements made to rlcompleter have been sent
 upstream and were accepted as of Python 2.3,
 
-
-This is now mostly Completer implementation made to function with
-prompt_toolkit, but that should still work with readline.
 """
 
 # Copyright (c) IPython Development Team.
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1,7 +1,7 @@
 # encoding: utf-8
 """Word completion for IPython.
 
-This module stared as fork of the rlcompleter module in the Python standard
+This module started as fork of the rlcompleter module in the Python standard
 library.  The original enhancements made to rlcompleter have been sent
 upstream and were accepted as of Python 2.3,
 
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1,50 +1,13 @@
 # encoding: utf-8
 """Word completion for IPython.
 
-This module is a fork of the rlcompleter module in the Python standard
+This module stared as fork of the rlcompleter module in the Python standard
 library.  The original enhancements made to rlcompleter have been sent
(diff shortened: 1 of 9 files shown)


Carreau:
Can't do that, IPykernel rely on NoOpContext.

IPython/core/completer.py
```diff
@@ -1,50 +1,13 @@
 # encoding: utf-8
 """Word completion for IPython.
 
-This module is a fork of the rlcompleter module in the Python standard
+This module stared as fork of the rlcompleter module in the Python standard
```
minrk: started

IPython/utils/contexts.py
```diff
@@ -59,6 +61,13 @@ def __exit__(self, *exc_info):
...
 
 class NoOpContext(object):
-    """Context manager that does nothing."""
+    """
+    Deprecated
+    
+    Context manager that does nothing."""
+
+    def __init__(self):
+        warnings.warn("""NoOpContext is deprecated since IPython 5.0 """)
```
minrk: DeprecationWarning

minrk:
Minor comments inline, but 👍

Carreau:
Fixed the 2 comments. Thanks for the review.

Carreau:
@takluyver a look ?

IPython/core/completer.py
```diff
@@ -1,50 +1,13 @@
...
-- GNU readline is also used by the built-in functions input() and
-  raw_input(), and thus these also benefit/suffer from the completer
-  features.  Clearly an interactive application can benefit by
-  specifying its own completer function and using raw_input() for all
-  its input.
-
-- When the original stdin is not a tty device, GNU readline is never
-  used, and this module (and the readline module) are silently inactive.
+This is now mostly Completer implementation made to function with
+prompt_toolkit, but that should still work with readline.
```
takluyver: Should it still work with readline? How would one use it? I'd probably just remove this mention of readline; it's not a use case we're trying to support, as far as I'm concerned.

IPython/core/interactiveshell.py
```diff
@@ -1922,10 +1880,7 @@ def showindentationerror(self):
 
     def init_readline(self):
         """Moved to terminal subclass, here only to simplify the init logic."""
```
takluyver: This docstring should probably say 'deprecated' now. It's no longer in the terminal subclass, now that the terminal subclass uses PTK.

takluyver:
Other than a couple of minor notes, :+1: !

IPython/core/completer.py
Carreau: Ok, so there are more code we can strip out. I'll remove the sentence and strip out more code later.

Carreau:
> Other than a couple of minor notes, 👍 !

Done fo both.

Carreau:
Yeah !

PR #9663, opened 2016-06-24
Docs overhall.

Carreau:
Remove many unnecessary section or inaccurate, build locally using RTD
theme to simplify testing locally.

Decrease the depth of some section. And do not duplicate table of
content which anyway is always on the sidebar.

---

Ping @willingc

PR #9693, opened 2016-07-01
fix docs requirement that prevent RTD to build

Carreau:
Going to self merge as soon as test passes.


Carreau:
introduced by a bad rebase of https://github.com/ipython/ipython/pull/9663/files

PR #9696, opened 2016-07-01
Update readme instructions.

Carreau:
from @michaelpacer feedback during 4.2.1 release.


Carreau:
Ok, I'll go for self-merge.

Issue #9652, opened 2016-06-23
IPython beta: interactive matplotlib magic breaks terminal (Windows)


jorisvandenbossche:
When activating interactive figures with `%matplotlib`, and after plotting the first figure, the terminal does not capture any keyboard input anymore

```
In [6]: %matplotlib
Using matplotlib backend: Qt4Agg

In [7]: plt.plot([1,2,3])
Out[7]: [<matplotlib.lines.Line2D at 0xad116a0>]
```

After that, I cannot type or paste anything anymore, also not when the figure is closed. Something with the event loop?

I am working on Windows 7, and should be on master for both ipython and prompt toolkit.

jorisvandenbossche:
This does not happen when I use `%matplotlib tk`, so only for the qt backend (and the others are not available on my Windows)

Carreau:
> After that, I cannot type or paste anything anymore, also not when the figure is closed. Something with the event loop?

Hum likely. I can't reproduce though on Mac OS. we will need someone on windows to help figure that out.

takluyver:
Thanks @jorisvandenbossche . The code for the Qt event loop is here:
https://github.com/ipython/ipython/blob/master/IPython/terminal/pt_inputhooks/qt.py

I'd guess that something to do with waiting for events on a file descriptor doesn't work on Windows; I know that `select()` on Windows only accepts sockets, not fds.

I forget which input hooks I copied from ptpython, but if this is one, @jonathanslenders might be interested.

jonathanslenders:
This one is not coming from ptpython.

What we could do on Windows is using a separate thread to monitor the fd and quit the qt loop from there.

jorisvandenbossche:
@Carreau I noticed that you changed the milestone :-) 
I know at a certain point you have to tag a release, but IMO this one is rather critical (at least for me this is the reason I am currently not using the beta version anymore, plotting is rather critical in my interactive work). 
Is there anything I can do to help? (I have a windows, but not much knowledge about qt and event loops)

Carreau:
Sorry about than, Min switched back the issue to 5.0.

takluyver:
@jorisvandenbossche when someone has attempted to fix this, it would be very useful if you could test the fix. I'll try to find time tomorrow if no-one beats me to it.

Carreau:
> @jorisvandenbossche when someone has attempted to fix this, it would be very useful if you could test the fix. I'll try to find time tomorrow if no-one beats me to it.

[...]

PR #9702, opened 2016-07-06
Try to get QT To work on windows.

Carreau:
@jorisvandenbossche  can you try that ?

I honestly have close to no clue what I am doing, as eventloop and threading is far from where I am comfortable. 

I guess the other alternative would be to mimic the other inputhook are doing on the IPykernel side and  timeshare. I'll see if I can borrow a windows box tomorrow from one of my officemates.

I'm also unsure if I can use things like `msvcrt.kbhit()` with `prompt_toolkit`.

CC @fperez , @minrk @takluyver .

### DIFF
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -1,11 +1,39 @@
 from IPython.external.qt_for_kernel import QtCore, QtGui
 
+import threading
+import time
+
+import select
+watched_fd = set({})
+
+def _stop_qt_ev_on_redable(fd, ev):
+    """QtNotifier seem to not work on windows, we manually watch the file"""
+    while True:
+        ready, _, _ = select.select((fd,), (), (), .1)
+        time.sleep(0.05)
+        if ready:
+            watched_fd.remove(fd)
+            ev.exit()
+            break
+
+
 def inputhook(context):
     app = QtCore.QCoreApplication.instance()
     if not app:
         return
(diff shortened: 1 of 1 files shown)


fperez:
I haven't worked with this code recently and I don't know why we have a regression, but as a general rule, threads at the heart of our event loop system have been historically nothing but a source of pain. A long time ago, we were able to move away from threads (the original solution for GUI integration dating back to ~2003/4), and that was a *huge* gain in stability.

We should really think long and hard about how to figure out a non-thread solution.

For reference, this is an attempt to fix #9652.

Carreau:
Agreed I would prefer to not use thread, but it was [suggested](https://github.com/ipython/ipython/issues/9652#issuecomment-228453026) so I tried.

Something [like that](https://github.com/ipython/ipython/blob/master/IPython/lib/inputhookqt4.py#L118-L124) would be the alternative I guess.

fperez:
That sounds a bit more palatable than a threads solution (plus, it's what we've used for years, so it has the benefit of being battle-tested).

Carreau:
> That sounds a bit more palatable than a threads solution (plus, it's what we've used for years, so it has the benefit of being battle-tested).

I'll try.

minrk:
I like #9702 a lot more, and it's working for me. We've had headaches with eventloops in threads for years, from the old `-wthread` days, so I'm happy that we have a solution that doesn't require reintroducing a thread for this.

PR #9705, opened 2016-07-06
Try to fix QT event loop take II

Carreau:
Second try at fixing #9652 

@jorisvandenbossche can you try that as well ? It's late so it's not a clean implementation. In particular you likely cannot deactivate the eventloop.

CC @fperez , @minrk 

@takluyver, should we try to merge the lib/inputhook and the pt_inputhook together ? It looks like the former could sometime just take an (optional) `context` argument.


minrk:
Thanks, @Carreau! I've tested this and it's working on my Windows VM.

jorisvandenbossche:
Yes, I can confirm it is working for me as well. Thanks a lot!

minrk:
Isn't lib/inputhook pretty much deprecated, now that we have the pt_inputhooks?

I'm okay with merging this as-is for 5.0, and then targeting a proper merge of pt_inputhooks and lib/inputhook for 5.1.

I've also confirmed that the behavior with this one is also fine on OS X.

takluyver:
Yes, lib/inputhook is deprecated. I'm going to try to make a variant of this which doesn't involve lib/inputhook.

takluyver:
#9706 is my try

Carreau:
If they are deprecated we should raise deprecation warnings.

Issue #9712, opened 2016-07-06
Deprecate lib/inputhook


Carreau:
As far as I understand it is unused. RIght ? 

@takluyver should we sprinkle deprecation warnings for 5.0 ?

takluyver:
Yep, it should be unused with the switch to prompt_toolkit.

I might even risk removing it directly - I'm not aware of anyone using that code for anything else. But the sensible thing to do is to have a deprecation cycle to see if I'm wrong about that.

Carreau:
closed by #9713

PR #9713, opened 2016-07-06
Cleanup and add more deprecation warnings.

Carreau:
cc @takluyver

### DIFF
--- a/IPython/lib/inputhook.py
+++ b/IPython/lib/inputhook.py
@@ -1,5 +1,7 @@
 # coding: utf-8
 """
+Deprecated since IPython 5.0
+
 Inputhook management for GUI event loop integration.
 """
 
@@ -98,7 +100,9 @@ else:
 
 
 class InputHookManager(object):
-    """Manage PyOS_InputHook for different GUI toolkits.
+    """DEPRECATED since IPython 5.0
+
+    Manage PyOS_InputHook for different GUI toolkits.
 
     This class installs various hooks under ``PyOSInputHook`` to handle
     GUI event loop integration.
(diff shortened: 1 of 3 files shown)


IPython/lib/inputhook.py
```diff
@@ -177,13 +197,17 @@ def clear_app_refs(self, gui=None):
             the app for that toolkit.  References are not held for gtk or tk
             as those toolkits don't have the notion of an app.
         """
+        warn("`clear_app_refs` is deprecated since IPython 5.0 and will be removed in future versions.",
```
takluyver: Rather than warning on individual methods, it may be more informative and less noisy to warn if the class is instantiated.

takluyver: I forgot that we instantiate it inside this module. Maybe we should just warn when the module is imported. Or is that annoying for tests?

Carreau: Yep that's annoying for tests. I did add the warning on import/instantiation, and removed it.

Carreau: Actually it should not be imported on master. Let me retry.

Carreau: Readded the warning at import time. And removed an old filtered warning.

Carreau:
If @fperez (or @ellisonbg) has a few minutes to review / and or merge. I can try to do a RC2 that fix matplotlib plus have this is tonight.

minrk:
👍

Carreau:
I'm wondering if I should release a rc2 after that, or just go through the final process and make a relese.

Carreau:
Will close #9712

fperez:
I just reviewed it, it all looks good. Thanks, solid work.

PR #9717, opened 2016-07-07
Update release instruction

### LINKED ISSUE #9695: [docs] Release instruction
Explain why we set variable **before** setting them.

describe what `_versions_extra` in release instruction does. 

Say that build_release check pep440 version earlier. 
Make it check env-variable.

Describe how to build docs (cd docs, make html)

Get a fresh clone and `cd ipython

`./release` is wrong, it's `tools/release` and it's not a quote it's bash. 

Add to the list to unhide PyPI releases.

Carreau:
closes #9695 

---

That should simplify a bit @fperez task of releasing 5.0.

### DIFF
--- a/docs/source/coredev/release_process.rst
+++ b/docs/source/coredev/release_process.rst
@@ -14,26 +14,33 @@ handy reminder and checklist for the release manager.
 ----------------------------
 
 Set environment variables to document previous release tag, current
-release milestone, current release version, and git tag::
-
-    PREV_RELEASE=4.0.0
-    MILESTONE=4.1
-    VERSION=4.1.0
-    BRANCH=master
+release milestone, current release version, and git tag.
 
 These variables may be used later to copy/paste as answers to the script
 questions instead of typing the appropriate command when the time comes. These
 variables are not used by the scripts directly; therefore, there is no need to
-`export` the variables.
+`export` the variables. Use the following in bash:
+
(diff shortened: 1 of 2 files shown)


docs/source/coredev/release_process.rst
```diff
@@ -42,20 +49,22 @@ If a major release:
 
           python tools/update_whatsnew.py
 
-    - update `docs/source/whatsnew/development.rst`, to ensure it covers
+    - update ``docs/source/whatsnew/development.rst``, to ensure it covers
       the major release features
-    - move the contents of `development.rst` to `versionX.rst` where `X` is
+
+    - move the contents of ``development.rst`` to ``versionX.rst`` where `X` is
```
fperez: Should this be ` \`\`X\`\` `?

fperez: Argh, don't know how to escape all those backticks in md... Anyway, what I meant is, should `X` have double-backticks instead of single?

fperez:
Very minor comment, otherwise good to merge and we'll proceed!

docs/source/coredev/release_process.rst
Carreau: Does not really matter as our default directive is `pre` anyway. And I understood you ment to type ``` `` `X` `` ```

PR #9731, opened 2016-07-10
Simplify debugger classes.

anntzer:
Directly get the color scheme from the interactive shell (which is
created if not already present, anyways).  The goal is ultimately to
allow the IPython debugger to be a drop-in replacement of the stdlib pdb
with the same signature (in particular, this would obviate the need for
plugins such as nose-ipdb and pytest-ipdb; nose and pytest could instead
let their `--pdb` command line arg take an argument specifying the Pdb
(sub)class to use, e.g. `py.test --pdb=IPython.core.debugger.Pdb`).

Also deprecate the `Tracer` class (and related utilities), which can be
fairly trivially replaced by `IPython.core.debugger.Pdb().set_trace()`
(we can let `Pdb` catch `BdbQuit` itself).
# 

It may be also possible to make `color_scheme` a kw-only arg to `Pdb`, but it would be awkward to do so while maintaining back-compatibility (because it is used positionally and as the _first_ argument in a few places).  So another option would be to break back-compatibility without a deprecation period, and directly make it a kw-only arg in IPython 6 (this is trivial to do in a Py3-only codebase).

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -30,6 +30,7 @@ import bdb
 import functools
 import inspect
 import sys
+import warnings
 
 from IPython import get_ipython
 from IPython.utils import PyColorize, ulinecache
@@ -62,6 +63,8 @@ def BdbQuit_excepthook(et, ev, tb, excepthook=None):
     All other exceptions are processed using the `excepthook`
     parameter.
     """
+    warnings.warn("`BdbQuit_excepthook` is deprecated since version 5.1",
+                  DeprecationWarning)
     if et==bdb.BdbQuit:
         print('Exiting Debugger.')
     elif excepthook is not None:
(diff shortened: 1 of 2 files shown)


Carreau:
Great ! Thanks !

I don't think we'll do that in the 5.x branch but as soon as we release 5.1 and for master toward a 6.0 I'll be happy to get things like that in !

Unless @takluyver thinks the deprecation messages in 5.1 are worth it. 

Thoughts ?

IPython/core/debugger.py
```diff
@@ -62,6 +63,7 @@ def BdbQuit_excepthook(et, ev, tb, excepthook=None):
     All other exceptions are processed using the `excepthook`
     parameter.
     """
+    warnings.warn("`BdbQuit_excepthook` is deprecated", DeprecationWarning)
```
Carreau: Nitpicking, but once we agree on  when to deprecate (5.1 or 6.0), I like to add `...deprecated since version X.y` otherwise it's painful for user to figure out when they can remove potential shims , and for us to figure out when we can delete the code.

anntzer:
PS: https://github.com/pytest-dev/pytest/pull/1712 seems on its way in.  nose2 seems to directly call the `pdb` API (instead of instantiating a `Pdb` instance) so it would require a bit deeper refactoring.

takluyver:
I kind of think that the colour stuff should be in the terminal subclass. But that would remove coloured output from debugging in Jupyter frontends, so probably not...

takluyver:
I think I'd be OK with this going in for 5.1, but I don't feel strongly either way.

IPython/core/debugger.py
anntzer: Added deprecation as of 5.1 following @takluyver's comment.

Carreau:
5.1 it will be then.

Carreau:
Ok, let's merge this one.

PR #9736, opened 2016-07-11
Add option to use 24bit color in prompt toolkit

wilywampa:
This exposes the true (24-bit) color option in prompt_toolkit to IPython's configuration. Same change as this: https://github.com/jupyter/jupyter_console/pull/90

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -14,7 +14,7 @@ from traitlets import Bool, Unicode, Dict, Integer, observe, Instance, Type, def
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
 from prompt_toolkit.filters import (HasFocus, Condition, IsDone)
 from prompt_toolkit.history import InMemoryHistory
-from prompt_toolkit.shortcuts import create_prompt_application, create_eventloop, create_prompt_layout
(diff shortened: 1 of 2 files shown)


Carreau:
Great thanks ! Do you think it is worth mentioning in the what's new ?

https://github.com/ipython/ipython/blob/master/docs/source/whatsnew/pr/README.md


I'll let it settle a few hours to leave others a chance to have a look.

IPython/terminal/interactiveshell.py
```diff
@@ -142,6 +142,13 @@ def refresh_style(self):
         help="Override highlighting format for specific tokens"
     ).tag(config=True)
 
+    true_color = Bool(False,
+        help=("Use 24bit colors instead of 256 colors in prompt highlighting. "
+              "If your terminal supports true color, the following command "
+              "should print 'TRUECOLOR' in orange: "
+              "printf \"\\x1b[38;2;255;100;0mTRUECOLOR\\x1b[0m\\n\"")
```
dsblank: Should that be print(...) ?

Carreau: No, that's a bash command :-)

dsblank: Oh, I see. Thanks!

Issue #9728, opened 2016-07-09
TerminalInteractiveShell.display_completions_in_columns can be removed.


Carreau:
It was added only for the beta/rc period and was not in IPython 4.x
```
--TerminalInteractiveShell.display_completions_in_columns=<Bool>
    Default: None
    DEPRECATED
```

PR #9738, opened 2016-07-12
Remove display_completions_in_columns

takluyver:
Closes gh-9728


willingc:
Looks good @takluyver

PR #9754, opened 2016-07-13
Unify docs on custom prompts

takluyver:
We had unwittingly documented this in two different places; this condenses them into one, leaving a reference from the other location.

Ping @willingc (again ;-) )


willingc:
@takluyver Missed this (did have this repo in my watch list - it is now), sorry. Thanks for completing the PR. And the changes you merged look good :-)

Issue #9588, opened 2016-06-09
In the PTK terminal, control-o should indent the newly created line


njsmith:
Suppose I have:

```
In [5]: def f():
   ...:     [CURSOR]a = 1           
```

If I press control-O to open up a new line, I get:

```
In [5]: def f():
   ...:     [CURSOR]
   ...: a = 1
```

I would prefer to get:

```
In [5]: def f():
   ...:     [CURSOR]
   ...:     a = 1
```

Carreau:
I'm going to tag that as "new feature", that would be one more carrot for Python 3 once implemented :-P. 
I might have a got at that before but not promises !

gforsyth:
@njsmith -- what should happen if you have 
```console
In [5]: def f():
     ...:     a = 1[CURSOR]
```

and hit Ctrl-O.  Do you expect a dedent or a preserved indentation level?

njsmith:
I guess it doesn't matter that much? But an indent seems slightly more convenient?

I guess the general proposal would be: hitting control-O should be equivalent to hitting Enter, except that: (a) it should never cause the current block to be executed, (b) it should leave the cursor where it started.

Carreau:
THis has been partially addressed by #9770 (well, all things listed in here seem to work). 
The one which is weird is : 

```
def foo():[CURSOR]
    a = 1$
````
`Ctrl-O` end up with:

```
def foo():[CURSOR]
$
    a = 1$
```
Which is still copying indent (if it's a method you are still at class level at least).
But we appear to have the logic for dedenting one level after `pass`, `return` and indenting one more after `:` so we should be able to improve.

PR #9770, opened 2016-07-16
preserve margins when using Ctrl-O for newline

gforsyth:
This should address ipython/ipython#9588.  Control-O will insert a new
line, preserve the current cursor position and preserve the indentation
level on the newline

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -58,7 +58,7 @@ def register_ipython_shortcuts(registry, shell):
 
     registry.add_binding(Keys.ControlO,
                          filter=(HasFocus(DEFAULT_BUFFER)
-                                & insert_mode))(newline_with_copy_margin)
+                                & EmacsInsertMode()))(newline_with_copy_margin)
 
     if shell.display_completions == 'readlinelike':
         registry.add_binding(Keys.ControlI,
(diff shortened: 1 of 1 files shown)


gforsyth:
@njsmith -- do you want to try this out and see if it meets your needs?

IPython/terminal/shortcuts.py
```diff
@@ -56,6 +56,10 @@ def register_ipython_shortcuts(registry, shell):
                                  & cursor_in_leading_ws
                         ))(indent_buffer)
 
+    registry.add_binding(Keys.ControlO,
+                         filter=(HasFocus(DEFAULT_BUFFER)
+                                & insert_mode))(newline_with_copy_margin)
```
takluyver: I guess that the filter probably needs to check for Emacs mode too.

gforsyth:
@takluyver -- I think this is ready for another look.  I removed the `insert_mode` filter since the emacs mode filter (I think) always overlaps with that.

takluyver:
Thanks. Yes, I think that works - EmacsInsertMode should be effectively emacs & insert.

takluyver:
@njsmith I have given this a quick test, but I'd like you to test that it actually behaves as you want.

Carreau:
Marking as 5.1

Carreau:
Ok,  looks good.

PR #9773, opened 2016-07-18
temp workaround for missing libraries

wwoesten:
a better permanent way is needed


takluyver:
This was a commit based on an old version of IPython. IPython no longer includes a bundled copy of Pexpect.

Issue #9768, opened 2016-07-16
 invalid continuation byte with ipython5.0


zhiyue:
```bash
C:\Users\zhiyue>ipython
Python 2.7.11 |Anaconda 2.3.0 (64-bit)| (default, Feb 16 2016, 09:58:36) [MSC v.1500 64 bit (AMD64)]
Type "copyright", "credits" or "license" for more information.

IPython 5.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.
```

When IPython upgrade to 5 later, so that the use of theano has been a problem, when I import the `theano` package, it can not display the `cudnn` information.

```
In [1]: import theano
DEBUG: nvcc STDOUT Using gpu device 0: GeForce GT 750M (CNMeM is enabled with initial size: 75.0% of memory, cuDNN None)
```
And then led to the `theano` can not be used under the IPython
I found that the output debug information may be the cause of this problem, as follows

```bash
In [1]: import theano
DEBUG: nvcc STDOUT Using gpu device 0: GeForce GT 750M (CNMeM is enabled with initial size: 75.0% of memory, cuDNN None)

In [2]: theano.sandbox.cuda.dnn.version()
INFO (theano.gof.compilelock): Refreshing lock C:\Users\zhiyue\AppData\Local\Theano\compiledir_Windows-10-10.0.10586-Intel64_Family_6_Mo
del_58_Stepping_9_GenuineIntel-2.7.11-64\lock_dir\lock
DEBUG: nvcc STDOUT ---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-2-f186563a86b7> in <module>()
----> 1 theano.sandbox.cuda.dnn.version()

D:\develop\Anaconda\lib\site-packages\theano\sandbox\cuda\__init__.pyc in dnn_version()
    410         f = theano.function([], DnnVersion()(),
    411                             theano.Mode(optimizer=None),
--> 412                             profile=False)
    413         dnn_version.v = f()
    414     return dnn_version.v

D:\develop\Anaconda\lib\site-packages\theano\compile\function.pyc in function(inputs, outputs, mode, updates, givens, no_default_updates
, accept_inplace, name, rebuild_strict, allow_input_downcast, profile, on_unused_input)
    318                    on_unused_input=on_unused_input,
[...]

PR #9778, opened 2016-07-19
Be more lenient when bytes are sent to stdout/stderr on Py2+Windows

takluyver:
Closes gh-9768

@Drekin @zhiyue : can you test this?

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -320,14 +320,36 @@ class TerminalInteractiveShell(InteractiveShell):
             pre_run=self.pre_prompt, reset_current_buffer=True)
         return document.text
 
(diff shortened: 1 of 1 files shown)


Drekin:
I'm quite busy right now, I'll take a look in a few days.

Carreau:
@Drekin did you had a chance to test that ? If not, @takluyver do you want to still get that in ?

takluyver:
I would still like to get it in, but it needs testing on Windows. I'll reboot my laptop and try it.

takluyver:
In my testing, this does print a replacement `?` instead of giving a big traceback.

Carreau:
> In my testing, this does print a replacement ? instead of giving a big traceback.

Ok, let's merge then.

Issue #9775, opened 2016-07-18
%load's automatic comment line causes IndentationError


joonro:
I started using `%load` a lot for interactive work. An issue is that it automatically adds a comment line on the top of the loaded script, which can cause `IndentationError` when the loaded script is indented. Hence I have to scroll all the way to the top and delete the comment line for it to work. For example, with the following `test_load.py` file:

```python
import os

if __name__ == '__main__':
    print(1)
```

```python
In [12]: %load -r 4: ./test_load.py

In [13]: # %load -r 4: ./test_load.py
    ...:     print(1)
    ...:     
    ...: 
  File "<ipython-input-13-126c9c2d676e>", line 2
    print(1)
    ^
IndentationError: unexpected indent
```

It would be great if there is a way either to exclude this line, automatically indent the comment line, or automatically unindent the actual code so it will run without any user input.

takluyver:
Can you try with #9779?

joonro:
I just cloned your branch and tried it in a terminal IPython, and it worked perfectly:

```python
In [3]: %load -r 4: ./test_load.py

In [4]: # %load -r 4: ./test_load.py
   ...: print(1)
   ...:
   ...:
```

Thank you so much for such a quick response!

takluyver:
Great, thanks :-)

hainm:
thanks for fixed PR too. I was about to "complain" but found this closed issue.

PR #9779, opened 2016-07-19
Dedent leading lines on %load -r

takluyver:
Closes gh-9775


takluyver:
@joonro confirmed that this works for his use case - see #9775.

Carreau:
Ok, let's merge this one ! Thanks both !

PR #9785, opened 2016-07-20
Restore completion in debugger.

### LINKED ISSUE #9759: no tab completion in terminal when using debugger
@Carreau I chatted with Brian Granger, who said I should bring this to your attention.

OS: Mac OS 10.10.5
python -V: Python 3.5.2 : Anaconda custom (x86_64)
ipython --version : 5.0.0
source of ipython: conda-forge build

When I am in the ipython terminal and using the debugger, tab completion does not work (i.e., I just get a tab).  This happened in the version of IPython I had before installing this a few minutes ago, so I wonder if it's something wrong with the conda-forge build?

I'm at Scipy, so if someone wants a demo of the problem, I'm happy to do that.

Carreau:
Close #9759

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -28,7 +28,7 @@ except ImportError:
     except ImportError:
         profile = pstats = None
 
-from IPython.core import debugger, oinspect
+from IPython.core import oinspect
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks

Issue #9784, opened 2016-07-20
IPython 5 + Qt breakage: no QApplication gets created, so inputhook does nothing.


zpincus:
### Problem
In IPython 4,  `%gui qt` starts up a `QApplication` automatically if one isn't already running, and returns it to the user. This does not happen in IPython 5.

This is a problem because unless a `QApplication` instance has been created, the Qt inputhook doesn't do anything. The relevant code for IPython 5, in [terminal/pt_inputhooks/qt.py](https://github.com/ipython/ipython/blob/master/IPython/terminal/pt_inputhooks/qt.py#L4), looks like this:
```python
def inputhook(context):
    app = QtCore.QCoreApplication.instance()
    if not app:
        return
```

In IPython 4, `%gui qt` winds up (through a complex chain of events) executing [lib. inputhookqt4.create_inputhook_qt4()](https://github.com/ipython/ipython/blob/master/IPython/lib/inputhookqt4.py#L66), which does the following:
```python
        app = QtCore.QCoreApplication.instance()
        if app is None:
            app = QtGui.QApplication([" "])
```
Moreover, in IPython 4, `%gui qt` returns this `QApplication` instance to the user.

None of this happens in IPython 5. Presumably this change was not intentional, and enabling the Qt GUI in IPython 5 should still create a `QApplication`? (There is (ancient?) code in [lib/guisupport.py](https://github.com/ipython/ipython/blob/master/IPython/lib/guisupport.py#L109) that returns or creates a `QApplication`, so someone could use that if they want.)

### Example
The following works fine in IPython 4:
```python
%gui qt
from PyQt5.QtWidgets import QWidget # or PyQt if you use Qt4
w = QWidget()
w.setGeometry(300, 300, 250, 150)
w.setWindowTitle("Test")
w.show()
```
but breaks horribly in IPython 5. This fixes things:
```python
%gui qt
from PyQt5.QtWidgets import QApplication, QWidget
app = QApplication([''])
w = QWidget()
w.setGeometry(300, 300, 250, 150)
w.setWindowTitle("Test")
w.show()
```

Using a virtual environment, one can test this pretty simply by using `pip` to install IPython 4 vs. 5, and pyqt5 via a wheel (statically linked to the Qt libs) that has recently appeared. This is much easier than having fuss with hand-building PyQt etc.

### Resolution 
I would submit a PR, but first I'd like to know where to put the relevant initialization code. 

[...]

PR #9789, opened 2016-07-21
Create a QApplication for inputhook if one doesn't already exist

takluyver:
Closes gh-9784

### DIFF
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -1,10 +1,15 @@
 import sys
 from IPython.external.qt_for_kernel import QtCore, QtGui
 
+# If we create a QApplication, keep a reference to it so that it doesn't get
+# garbage collected.
+_appref = None
+
 def inputhook(context):
+    global _appref
     app = QtCore.QCoreApplication.instance()
     if not app:
-        return
+        _appref = app = QtGui.QApplication([" "])
     event_loop = QtCore.QEventLoop(app)
 
     if sys.platform == 'win32':


takluyver:
See discussion on #9784 before merging this.

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,10 +1,15 @@
...
 
+# If we create a QApplication, keep a reference to it so that it doesn't get
+# garbage collected.
+_appref = None
+
 def inputhook(context):
+    global _appref
     app = QtCore.QCoreApplication.instance()
     if not app:
-        return
```
Carreau: Want to drop a DeprecationWarning/(UserWarning maybe) here if App does not exists ? so that we can remove it at some point later?

takluyver: If we do decide to keep it in place, I wouldn't really consider it deprecated.

Carreau: I think we broke the use case in 5.0, and I'm fine restoring the behavior in next minor/patch version {especially since LTS]
I don't think in the long run we should create the App if it does not exists. 

So I would reinstate it to not break user-code/habits. But warn about it, and remove it in the Py3 - Only branch.

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,10 +1,15 @@
 import sys
 from IPython.external.qt_for_kernel import QtCore, QtGui
 
+# If we create a QApplication, keep a reference to it so that it doesn't get
+# garbage collected.
+_appref = None
```
tacaswell: 90% sure that pyqt/pyside already do this.

tacaswell:
@Carreau I disagree about not creating this the App.  If the user has asked to use Qt event loop integration, then clearly they want to use Qt :wink: .  

The `qApp` is a enforced-by-segfault singleton at the c++ level (the official line is creating more than one `QApplication` in the same process is undefined behavior) so all libraries that need a qApp need to go through this 'get the global if None create' dance ([mpl does it here](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/backends/backend_qt5.py#L121) ) 

Having IPython do this as part of the integration is a big convenience for users.

IPython/terminal/pt_inputhooks/qt.py
takluyver: At least one of the options we support seemingly doesn't - I tried without it and things went wrong. I can't remember which one I was testing with, though.

PR #9813, opened 2016-07-26
Support running directories with __main__.py files. Closes #9812

### LINKED ISSUE #9812: __main__.py files not used in directory execution
Take the following directory structure:
```
test_dir
    __main__.py
```
where `__main__.py` reads
```python
print("Hi")
```

Using the standard `python` command, I can:
```
$ python3 test_dir
Hi
```

Under `ipython`, this produces:
```
$ ipython3 test_dir
[TerminalIPythonApp] WARNING | File not found: 'test_dir'
```

This feature was added in Python 2.5. I suspect this error is simply because `ipython` hasn't implemented this functionality yet, but I don't see any open issues.

This feature is documented at https://docs.python.org/3/using/cmdline.html#interface-options.

Is there a specific reason why `ipython` doesn't use [runpy.run_path](https://docs.python.org/3/library/runpy.html#runpy.run_path) for running scripts passed as arguments?

If somebody could point me to a file in which I could patch this, I'd be happy to. I found [these lines](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L2475-L2477) in `safe_execfile`, but `runpy.run_file` isn't entirely equivalent to `py3compat.execfile`, since it doesn't support the `locals` or `compiler` arguments.

controversial:
Note that this does not add support running zip files via the same mechanism, which the `python` executable does support.

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -100,7 +100,7 @@ shell_aliases['cache-size'] = 'InteractiveShell.cache_size'
 
 class InteractiveShellApp(Configurable):
     """A Mixin for applications that start InteractiveShell instances.
-    
+
     Provides configurables for loading extensions and executing files
     as part of configuring a Shell environment.
 
@@ -126,7 +126,7 @@ class InteractiveShellApp(Configurable):
 
     # Extensions that are always loaded (not configurable)
     default_extensions = List(Unicode(), [u'storemagic']).tag(config=False)
-    
+
     hide_initial_ns = Bool(True,
         help="""Should variables loaded at startup (by startup files, exec_lines, etc.)
         be hidden from tools like %who?"""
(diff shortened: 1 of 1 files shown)


IPython/core/shellapp.py
```diff
@@ -100,7 +100,7 @@
 
 class InteractiveShellApp(Configurable):
     """A Mixin for applications that start InteractiveShell instances.
-    
+
```
takluyver: Not a big deal, but if you're contributing to IPython/Jupyter more in the future, we will ask you to configure your editor not to strip whitespace in lines you haven't touched - it can create needless merge conflicts, and it makes `git blame` less useful.

controversial: Sorry! I didn’t even notice Atom did that on save. If I contribute more in
the future I’ll fix that.

controversial: @takluyver Should I fix this now or just in the future?

takluyver: No worries - Atom's behaviour is a sensible default for new files, but a bit annoying for big codebases like this with trailing whitespace already there. We'll live with it for this PR.

takluyver:
And I think the actual fix looks good; thanks!

controversial:
Awesome! Glad I helped. Maybe I'll tackle zip archives later, I'll see. I don't actually need that behavior, though, so I probably won't.

PR #9820, opened 2016-07-28
Quit IPython on Ctrl-\ (SIGQUIT)

### LINKED ISSUE #9810: IPython 5 shell does not react to SIGQUIT (CTRL + \)
In previous IPython versions it was possible to terminate an IPython session quickly by sending an `SIGQUIT`, i.e., by pressing <kbd>CTRL</kbd>+<kbd> \ </kbd>. This is useful when having an `embed()` in a loop with a large number of iterations. Pressing <kbd>CTRL</kbd>+<kbd>d</kbd> followed by <kbd>y</kbd> is not practical. Since `%kill_embedded` is currently also broken there is no convenient way to terminate the process.

Carreau:
Closes #9810

---

Sending the signal externally already works, just the `ctrl-\` binding was not working.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -26,6 +26,8 @@ def register_ipython_shortcuts(registry, shell):
                                  & insert_mode
                         ))(newline_or_execute_outer(shell))
 
+    registry.add_binding(Keys.ControlBackslash)(force_exit)
+
     registry.add_binding(Keys.ControlP,
                          filter=(ViInsertMode() & HasFocus(DEFAULT_BUFFER)
                         ))(previous_history_or_previous_completion)
@@ -144,6 +146,12 @@ def reset_search_buffer(event):
 
 def suspend_to_bg(event):
     event.cli.suspend_to_background()
+
+def force_exit(event):
+    """
+    Force exit (with a non-zero return value)
+    """
+    sys.exit("Quit")
 
 def indent_buffer(event):
     event.current_buffer.insert_text(' ' * 4)


eudoxos:
The problem here is that with prompt-toolkit, terminal settings are not restored. Alternatively, there should be a way to prevent ipython from trapping the Ctrl-\ shortcut, normally used to send SIGQUIT, of which handler I wanted to use to restore terminal settings.

takluyver:
I can't replicate that with Ctrl-\ - my terminal seems to work fine after it's quit. I do see that some terminal settings don't get restored if you suspend IPython (Ctrl-Z).

Do you see the text 'Quit' when you press Ctrl-\, or something else like `[1]+  Stopped                 ipython`?

eudoxos:
I see `Exit message: Quit` of which second part comes I guess from https://github.com/ipython/ipython/commit/80515cc251c35cbde4a2bf331e4470fd213ed4fc#diff-e9a59ec7c6360d84aafa289edca1c50bR154 . I updated IPython and prompt-toolkit to lastest from github now, still the same happens. `Exit message: Quit` but echo is disabled in bash. That means Ctrl-\ is handled by IPython as it should, but the cleanup is not done.

Strange that you don't reproduce that. FWIW, I am on Ubuntu 16.04 with fairly standard env, running in libvte-based terminal emulator (terminator).

takluyver:
The `Exit message:` confuses me; I don't see that in my terminal, and that text is nowhere in the IPython source. I was testing with Gnome terminal in Ubuntu 16.04 before, I've just tried it with both Gnome terminal and terminator in Fedora, and it's working for me in both (and not showing `Exit message:`).

eudoxos:
Solved. The problem was here. I forgot my code overrides sys.exit to avoid shutdown issues in external lib - it calls a c++ routine doing some cleanup and then calling exit(1), which explains why the python cleanup restoring terminal was never called. The "Exit message:" is also from my code. Apologies and thanks for helping. Keep up the great work on IPython!

takluyver:
OK, thanks for letting us know.

PR #9823, opened 2016-07-29
fix some deprecations

### LINKED ISSUE #9817: IPython 5.0 deprecates but still uses utils.io.IOStream
Resulting in unavoidable self-referencing deprecation warnings. We should either remove/limit the warnings, or actually stop using them.

io.IOStream is deprecated in 5.0, but is still used in a few places:

- [Terminal]InteractiveShell.init_io
- testing.globalipapp

minrk:
various deprecations, especially our own usage of deprecated APIs within this package
- remove few remaining references to, uses of io.stdout
- suppress deprecation warnings when initializing deprecated `utils.io.std*`
- globalipapp.StreamProxy is now totally unused
- one missing traitlets 4.2 API update in core.formatters
- get gui keys from pt_inputhooks instead of deprecated lib.inputhook
- stop passing deprecated color_scheme to Pdb
- nt.assert_equals in test_latextools

closes #9817

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -294,14 +294,13 @@ class EmbeddedSphinxShell(object):
         IP = InteractiveShell.instance(config=config, profile_dir=profile)
         atexit.register(self.cleanup)
 
-        # io.stdout redirect must be done after instantiating InteractiveShell
-        io.stdout = self.cout
-        io.stderr = self.cout
+        sys.stdout = self.cout
+        sys.stderr = self.cout
 
         # For debugging, so we can see normal output, use this:
         #from IPython.utils.io import Tee
-        #io.stdout = Tee(self.cout, channel='stdout') # dbg
-        #io.stderr = Tee(self.cout, channel='stderr') # dbg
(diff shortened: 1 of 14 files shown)


IPython/sphinxext/ipython_directive.py
```diff
@@ -294,14 +294,15 @@ def __init__(self, exec_lines=None):
         IP = InteractiveShell.instance(config=config, profile_dir=profile)
         atexit.register(self.cleanup)
 
-        # io.stdout redirect must be done after instantiating InteractiveShell
-        io.stdout = self.cout
-        io.stderr = self.cout
+        # stdout redirect must be done after instantiating InteractiveShell
+        # FIXME: is that still true in IPython 5?
```
takluyver: I don't think so; for `io.stdout` we needed to because instantiating `InteractiveShell` set `io.stdout` to something else. I don't think initialising the shell changes `sys.std*`.

takluyver:
:+1:

IPython/sphinxext/ipython_directive.py
minrk: Thanks, I'll just remove the comment, then.

PR #9825, opened 2016-07-29
Do not update the unders if they are user defined

Carreau:
Closes jupyter/notebook#1628 (once there are tests).

Still shift the internal reference to self._,__,___ but do not assign it
in user ns.

I would expect this to be a change of behavior and have the (slight)
side effect that if any of the _, __, or __ are defined, the none of
these are updated.

We could update the logic to do the right think on leapfrog _, __ or __
if user-set, but is it worth it.

---

Tagging as 6.0, but submitting now as it was lingering on my head and taking mental space.

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -76,6 +76,9 @@ class DisplayHook(Configurable):
         # particular uses _, so we need to stay away from it.
         if '_' in builtin_mod.__dict__:
             try:
+                user_value = self.shell.user_ns['_']
+                if user_value is not self._:
+                    return
                 del self.shell.user_ns['_']
             except KeyError:
                 pass
@@ -195,13 +198,23 @@ class DisplayHook(Configurable):
         if result is not self.shell.user_ns['_oh']:
             if len(self.shell.user_ns['_oh']) >= self.cache_size and self.do_full_cache:
                 self.cull_cache()
-            # Don't overwrite '_' and friends if '_' is in __builtin__ (otherwise
-            # we cause buggy behavior for things like gettext).
 
-            if '_' not in builtin_mod.__dict__:
-                self.___ = self.__
-                self.__ = self._
-                self._ = result
+            # Don't overwrite '_' and friends if '_' is in __builtin__
+            # (otherwise we cause buggy behavior for things like gettext). and
+            # do not overwrite _, __ or ___ if one of these has been assigned
+            # by the user.
+            update_unders = True
+            for unders in ['_'*i for i in range(1,4)]:
+                if not unders in self.shell.user_ns:
+                    continue
+                if getattr(self, unders) is not self.shell.user_ns.get(unders):
+                    update_unders = False
+
+            self.___ = self.__
+            self.__ = self._
+            self._ = result
+
+            if ('_' not in builtin_mod.__dict__) and (update_unders):
                 self.shell.push({'_':self._,
                                  '__':self.__,
                                 '___':self.___}, interactive=False)
@@ -209,7 +222,7 @@ class DisplayHook(Configurable):
             # hackish access to top-level  namespace to create _1,_2... dynamically
             to_main = {}
             if self.do_full_cache:
-                new_result = '_'+repr(self.prompt_count)
+                new_result = '_%s' % self.prompt_count
                 to_main[new_result] = result
                 self.shell.push(to_main, interactive=False)
                 self.shell.user_ns['_oh'][self.prompt_count] = result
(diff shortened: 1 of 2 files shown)


takluyver:
The way the regular Python shell does it is actually to put its automatic `_` into builtins, so if you define `_` in the main namespace, it shadows the automatic one. Maybe we should emulate that.

Carreau:
> The way the regular Python shell does it is actually to put its automatic _ into builtins, so if you define _ in the main namespace, it shadows the automatic one. Maybe we should emulate that.

Good point, I'd like @fperez advice on that, if he would be ok with this change.

Carreau:
> The way the regular Python shell does it is actually to put its automatic _ into builtins, so if you define _ in the main namespace, it shadows the automatic one. Maybe we should emulate that.

Well no we have a thing that check we don't shove things into builtins:

```
   def check_for_underscore(self):
        """Check if the user has set the '_' variable by hand."""
        # If something injected a '_' variable in __builtin__, delete
        # ipython's automatic one so we don't clobber that.  gettext() in
        # particular uses _, so we need to stay away from it.
        if '_' in builtin_mod.__dict__:
            try:
                del self.shell.user_ns['_']
            except KeyError:
                pass
```

fperez:
The question is: does `gettext` work in the default shell ok? It puts `_` in `__builtins__`, so if we follow that strategy we'd have to either change our strategy in `check_for_underscore` or break gettext...

Carreau:
> The question is: does gettext work in the default shell ok?

Yes it does in the Python one, though apparently not in IPython it does weird things:

```
>>> import gettext
>>> gettext.install('foo')
>>> _
<bound method NullTranslations.gettext of <gettext.NullTranslations object at 0x1019288d0>>
>>> _
<bound method NullTranslations.gettext of <gettext.NullTranslations object at 0x1019288d0>>
>>> _ = 3
>>> _
3
>>> _
3
>>> _
3
```

In IPython (master) : 


```
In [1]: import gettext; gettext.install('foo')

In [2]: _
Out[2]: <bound method NullTranslations.gettext of <gettext.NullTranslations object at 0x105288b70>>

In [3]: _
Out[3]: <bound method NullTranslations.gettext of <gettext.NullTranslations object at 0x105288b70>>

In [4]: _ = 3

In [5]: _
Out[5]: 3

In [6]: _
Out[6]: <bound method NullTranslations.gettext of <gettext.NullTranslations object at 0x105288b70>>

In [7]: _
Out[7]: <bound method NullTranslations.gettext of <gettext.NullTranslations object at 0x105288b70>>
```

So my guess is that we (likely) don't want to shove things in builtins (because we don't want to affect other modules that might make use of `_` as gettext) and that 2) we don't want to revert `_` to gettext  if use have set it to something else (but that a separate bug).

Carreau:
Refined the logic:

```
Summary
    if user have set _
       _ == value set by the user
    elif _ in builtins:
        _ == value in builtins._
    else:
        _ == previous result.

Note that this logic may fall down if the use set _ to a specific value,
and have this same value returned while _ is also in builtin:

    In [1]: import gettext ; gettext.install('foo')

    In [2]: _ = 'Example'

    In [3]: _
    Out[3]: 'Example'

    In [4]: _
    Out[4]: 'Example'

    In [5]: _
    Out[5]: <bound method NullTranslations.gettext of <...>>
```

That's because the object in history and set by the user have the same ID so there is no easy/simple way to "detect" that the value of `_` is not from history.

I would though consider this an edge case.

takluyver:
I think it's fine to consider that an edge case.

IPython/core/displayhook.py
```diff
@@ -195,21 +198,31 @@ def update_user_ns(self, result):
...
-                self.___ = self.__
-                self.__ = self._
-                self._ = result
+            # Don't overwrite '_' and friends if '_' is in __builtin__
+            # (otherwise we cause buggy behavior for things like gettext). and
+            # do not overwrite _, __ or ___ if one of these has been assigned
+            # by the user.
+            update_unders = True
+            for unders in ['_'*i for i in range(1,4)]:
+                if not unders in self.shell.user_ns:
```
takluyver: This check seems unneccessary as you're using `.get()` below anyway.

takluyver: nvm, I see the point of it now.

takluyver:
This looks OK to me; I'll give it a while for others to look.

IPython/core/displayhook.py
Carreau: > nvm, I see the point of it now.

Does the mean the code is too complicated and should be simplified ?

takluyver: It's not the clearest, but I don't currently see an obvious way to simplify it.

Issue #9751, opened 2016-07-13
Tab completion for paths does not support windows-style paths


tmr232:
When using tab-completion on a path, it seems to only work with forward-slashes and not back-slashes. While Python makes the use of forward-slashes possible for Windows, I can't get path completion for any path I copy from other apps.

Carreau:
Can you double the backslashes ? backslash means escaping in python (unless r"" strings are used) so this might be why.

tmr232:
Doesn't matter. `'c:\\'` does not autocomplete, `r'c:\'` does not autocomplete, `'c:/'` autocompletes correctly.

Liam3851:
Relatedly-- the drive letter is being completely ignored on Windows under ipython 5.0. For example, if I type `pd.read_csv("Z:/` and press `<tab>` it gives me completions from the C: drive, not the Z: drive. The completions start with a /, as a UNIX path would (e.g. `'/Anaconda/'). If I type `pd.read_csv("Z:/folder_in_z` where folder_in_z is a folder that exists on the Z: drive but not the C: drive, I get no completions at all. 

In short, combined with tmr's report, it appears that the path is being parsed as a UNIX path, rather than a Windows path. This is a new issue in IPython 5; tab completion was working fine under IPython 4.x.

tmr232:
I finally had time to start looking at the cause of the bug. Here are my findings:
1. The text passed to the completer is split at the colon (`:`) sign, removing the drive letter. See [here](https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L267) and also the delimiter list [here](https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L219)
2. Backslashes (`\`) are removed. Probably because they are used on Linux to escape spaces. See [here](https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L732)

There is also the issue with `(`, `[`, and `!` being possible file- or directory-names on Windows.

PR #9826, opened 2016-07-29
Improved Windows path completion.

tmr232:
Removed `\` and `:` from the delimiter list for Windows, and
stopped deleting `\` chars from paths on Windows.

This relates to #9751, but I don't think it solves it completely.


Carreau:
+1, let's try it.

Carreau:
Thanks a lot !

PR #9843, opened 2016-08-05
Allow starting TerminalInteractiveShell more than once

### LINKED ISSUE #193: Problems with IPython 5
Trying to use `pudb` with `IPython 5` has been causing me problems. 

Pressing `!` to go to `ipdb` works fine for the first time, but getting out of the shell and trying in again doesn't do anything. Have not thoroughly tested but everything seems fine on the first try, and quitting `pudb` and restarting it makes it fine again (but again only for one run). 

I guess it has something to do with the massive changes introduced in IPython 5?

minrk:
1. always start with `keep_running=True` in `interact()`, which avoids `interact()` being a no-op on all runs after the first.
2. don't close eventloop at end of mainloop (not sure this is the right thing to do)
3. pass `self`, not `self.Completer` to PTCompleter, to ensure that it's always hooked up to shell.Completer, which can be replaced over time (pudb).

cc @Carreau, @takluyver for the right thing to do about `_eventloop.close()`, since I doubt this is the right thing. Maybe register close with `atexit.register`?

cc @inducer this fixes https://github.com/inducer/pudb/issues/193

### DIFF
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -3,6 +3,7 @@ from wcwidth import wcwidth
 
 from IPython.utils.py3compat import PY3
 
+from IPython.core.completer import IPCompleter
 from prompt_toolkit.completion import Completer, Completion
 from prompt_toolkit.layout.lexers import Lexer
 from prompt_toolkit.layout.lexers import PygmentsLexer
@@ -13,8 +14,11 @@ import pygments.lexers as pygments_lexers
 class IPythonPTCompleter(Completer):
     """Adaptor to provide IPython completions to prompt_toolkit"""
     def __init__(self, shell):
+        if isinstance(shell, IPCompleter):
+            raise TypeError("IPythonPTCompleter expects an InteractiveShell"
+                            " instance in IPython 5.1, not a Completer")
         self.shell = shell
-    
+
     @property
     def ipy_completer(self):
         return self.shell.Completer
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -12,8 +12,12 @@ import pygments.lexers as pygments_lexers
 
 class IPythonPTCompleter(Completer):
     """Adaptor to provide IPython completions to prompt_toolkit"""
-    def __init__(self, ipy_completer):
-        self.ipy_completer = ipy_completer
+    def __init__(self, shell):
+        self.shell = shell
+    
+    @property
+    def ipy_completer(self):
+        return self.shell.Completer
 
     def get_completions(self, document, complete_event):
         if not document.current_line.strip():
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -1,3 +1,12 @@
+"""prompt-toolkit utilities
+
+Everything in this module is a private API,
+not to be used outside IPython.
+"""
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
 import unicodedata
 from wcwidth import wcwidth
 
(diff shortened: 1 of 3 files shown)


IPython/terminal/ptutils.py
```diff
@@ -12,8 +12,12 @@
 
 class IPythonPTCompleter(Completer):
     """Adaptor to provide IPython completions to prompt_toolkit"""
-    def __init__(self, ipy_completer):
-        self.ipy_completer = ipy_completer
+    def __init__(self, shell):
```
Carreau: **gruml gluml** change of API.. **gruml, gruml** but I see the reason. 
Can we check the type of shell and print a deprecation warning if it's a Completer  ?

Carreau:
> cc @Carreau, @takluyver for the right thing to do about _eventloop.close(), since I doubt this is the right thing. Maybe register close with atexit.register?

Probably, or should we have an event loop "reference counting" and a  `with eventloop` context manager that hold the eventloop ?

IPython/terminal/ptutils.py
minrk: Not every internal utility class is a public API, so I have no qualms changing this one. I can make it backward-compatible, though.

minrk: Added now in a fully backward-compatible way, and explicitly declared the module as private APIs.

IPython/terminal/ptutils.py
```diff
@@ -1,8 +1,18 @@
+"""prompt-toolkit utilities
+
+Everything in this module is a private API,
+not to be used outside IPython.
```
takluyver: Let's add it to the packages skipped for the API docs, in the list [here](https://github.com/ipython/ipython/blob/345bf7d28a831088209162eb855385b94187a362/docs/autogen_api.py#L33).

takluyver: In fact, [here](https://github.com/ipython/ipython/blob/345bf7d28a831088209162eb855385b94187a362/docs/autogen_api.py#L59).

minrk: Done.

minrk:
> should we have an event loop "reference counting" and a with eventloop context manager that hold the eventloop ?

I don't think that quite works because it seems like the refcount would never be anything other than one after the class is instantiated, since the loop is setup in `__init__`, and shouldn't be closed until the object is garbage collected (atexit or `__del__`). It's not recursive, so the count doesn't go up, it's starting and stopping the same IPython instance once in a while.

takluyver:
I've just had a look at the prompt_toolkit code, and it doesn't look like there should be any ill effects of not calling `eventloop.close()`. Some fds will remain open, but you probably only have one InteractiveShell object, so a few fds lingering shouldn't be a problem.

If we do call it anywhere, I think `__del__` would be the best place. That assumes that no-one is taking a reference to the event loop and using it outside of the shell object, but I think that's probably safe. Or maybe prompt_toolkit should have a `__del__` method on the event loop classes to clear up their fds. (@jonathanslenders)

minrk:
I tried putting it in `__del__`, but it doesn't seem to ever be called, perhaps due to circular references/global links created between pt and/or IPython.

takluyver:
There certainly are going to be circular references through `IPythonPTCompleter` and `TerminalInteractiveShell.inputhook`. But recent versions of Python should be able to break the cycle and call `__del__`.

minrk:
Python 3.5 isn't recent enough, at least.

takluyver:
I've just tested with a `__del__` method on TerminalInteractiveShell, and it does get called when the interpreter (3.5) exits. I was doing this to check that I hadn't misunderstood something. I don't think it's terribly important to do cleanup in `__del__`, and it would be better done in `Eventloop.__del__` anyway. So I'm merging this as is.

minrk:
> 've just tested with a __del__ method on TerminalInteractiveShell, and it does get called when the interpreter (3.5) exits

I did the same and it didn't (and again just now, with conda CPython 3.5.1 on OS X), so it's at least not a trustworthy event.

Agreed on the other points, though. Thanks!

PR #9851, opened 2016-08-09
Also capture execution results using sys.displayhook

takluyver:
From discussion on mailing list. It's not clear if this is what we want to do, or if it should be controlled by a separate option, but it's submitted for consideration.


minrk:
I think it makes more sense to more people this way. Relatively few people really understand or care about the difference between displayhook and display.

Carreau:
+1; @takluyver is there a need for a line in the what's new ?

takluyver:
I've added a whatsnew entry.

PR #9856, opened 2016-08-10
Document colours and escapes in prompts

takluyver:
Closes gh-8724


willingc:
@takluyver Looks good. Thanks.

PR #9861, opened 2016-08-11
Handle all OTP authentication methods.

### LINKED ISSUE #9860: tools/github_stats.py still fails on two factor auth with app
See #9179. I think #9186 fixed it only for SMS codes, not for OTPs generated from a smartphone app.

Carreau:
Closes #9860

@takluyver ?

### DIFF
--- a/tools/gh_api.py
+++ b/tools/gh_api.py
@@ -66,13 +66,13 @@ def get_auth_token():
     response = requests.post('https://api.github.com/authorizations',
                             auth=(user, pw), data=json.dumps(auth_request))
     if response.status_code == 401 and \
-       response.headers.get('X-GitHub-OTP') == 'required; sms':
-        print("Your login API resquest a SMS one time password", file=sys.stderr)
-        sms_pw = getpass.getpass("SMS password: ", stream=sys.stderr)
+            'required;' in response.headers.get('X-GitHub-OTP', ''):
+        print("Your login API requested a one time password", file=sys.stderr)
+        otp = getpass.getpass("One Time Password: ", stream=sys.stderr)
         response = requests.post('https://api.github.com/authorizations',
                             auth=(user, pw), 
                             data=json.dumps(auth_request),
-                            headers={'X-GitHub-OTP':sms_pw})
+                            headers={'X-GitHub-OTP':otp})
     response.raise_for_status()
     token = json.loads(response.text)['token']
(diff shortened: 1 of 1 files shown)


tools/gh_api.py
```diff
@@ -66,13 +66,13 @@ def get_auth_token():
     response = requests.post('https://api.github.com/authorizations',
                             auth=(user, pw), data=json.dumps(auth_request))
     if response.status_code == 401 and \
-       response.headers.get('X-GitHub-OTP') == 'required; sms':
-        print("Your login API resquest a SMS one time password", file=sys.stderr)
-        sms_pw = getpass.getpass("SMS password: ", stream=sys.stderr)
+            'required;' in response.headers.get('X-GitHub-OTP'):
```
takluyver: What does `response.headers.get()` give if it's not there? A normal dict would give `None`, which will fail an `in` test.

takluyver: That wasn't very clear: I mean that trying to check `x in None` will error rather than returning True/False.

takluyver:
This worked for me. Other than my note above, I think it looks fine.

tools/gh_api.py
Carreau: > That wasn't very clear: I mean that trying to check x in None will error rather than returning True/False.

Good point; fixed. Arguably it should **always** require OTP right ?

tools/gh_api.py
```diff
@@ -66,13 +66,13 @@ def get_auth_token():
     response = requests.post('https://api.github.com/authorizations',
                             auth=(user, pw), data=json.dumps(auth_request))
     if response.status_code == 401 and \
-       response.headers.get('X-GitHub-OTP') == 'required; sms':
-        print("Your login API resquest a SMS one time password", file=sys.stderr)
-        sms_pw = getpass.getpass("SMS password: ", stream=sys.stderr)
+            'required;' in response.headers.get('X-GitHub-OTP', ''):
+        print("Your login API resquest a one time password", file=sys.stderr)
```
takluyver: resquest -> requested

takluyver:
Probably anyone doing a release should have  a one-time-password, but I don't think the login mechanism should enforce that.

Carreau:
untypoed. 

Is there anything I can do to help you ? 

I'm also considering bumping a few things from 5.2 to 6.0 as we have already a few PR for 6.0.

PR #9862, opened 2016-08-11
Clean documentation process

Carreau:
We don't push on ghpages anymore.


willingc:
Looks good. Thanks @Carreau

PR #9863, opened 2016-08-11
Remove most of the ipyparallel info in the documentation. 

Carreau:
We already point to ipyparallel on RTD.

### DIFF
--- a/docs/source/config/integrating.rst
+++ b/docs/source/config/integrating.rst
@@ -43,10 +43,12 @@ For example::
 Custom exception tracebacks
 ===========================
 
-Rarely, you might want to display a different traceback with an exception -
-IPython's own parallel computing framework does this to display errors from the
-engines. To do this, define a ``_render_traceback_(self)`` method which returns
-a list of strings, each containing one line of the traceback.
+Rarely, you might want to display a custom traceback when reporting an
+exception. To do this, define the custom traceback using
+`_render_traceback_(self)` method which returns a list of strings, one string
+for each line of the traceback. For example, the `ipyparallel
+<http://ipyparallel.readthedocs.io/>`__ a parallel computing framework for
+IPython, does this to display errors from multiple engines.
 
 Please be conservative in using this feature; by replacing the default traceback
 you may hide important information from the user.
--- a/docs/source/development/config.rst
+++ b/docs/source/development/config.rst
@@ -41,20 +41,17 @@ The next thing you need to know is what to call your configuration file. The
 basic idea is that each application has its own default configuration filename.
 The default named used by the :command:`ipython` command line program is
 :file:`ipython_config.py`, and *all* IPython applications will use this file.
-Other applications, such as the parallel :command:`ipcluster` scripts or the
-QtConsole will load their own config files *after* :file:`ipython_config.py`. To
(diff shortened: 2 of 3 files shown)


docs/source/config/integrating.rst
```diff
@@ -44,9 +44,10 @@ Custom exception tracebacks
 ===========================
 
 Rarely, you might want to display a different traceback with an exception -
```
willingc: Rarely, you might want to display a custom traceback when reporting an exception. To do this, define the custom traceback using ``_render_traceback_(self)`` method which returns a list of strings, one string for each line of the traceback. For example, the `ipyparallel <http://ipyparallel.readthedocs.io/>`__ computing framework uses this to display errors from multiple engines.

docs/source/development/config.rst
```diff
@@ -41,8 +41,8 @@ The next thing you need to know is what to call your configuration file. The
 basic idea is that each application has its own default configuration filename.
 The default named used by the :command:`ipython` command line program is
 :file:`ipython_config.py`, and *all* IPython applications will use this file.
-Other applications, such as the parallel :command:`ipcluster` scripts or the
-QtConsole will load their own config files *after* :file:`ipython_config.py`. To
+Other applications,
```
willingc: Remove comma after applications

willingc:
@Carreau A couple of suggestions. Looks good to merge after these edits 👍

docs/source/config/integrating.rst
```diff
@@ -44,9 +44,10 @@ Custom exception tracebacks
...
-IPython's own parallel computing framework does this to display errors from the
-engines. To do this, define a ``_render_traceback_(self)`` method which returns
-a list of strings, each containing one line of the traceback.
+`ipyparallel <http://ipyparallel.readthedocs.io/>`__ computing framework does
```
takluyver: This doesn't scan quite right. Maybe

> *ipyparallel`, a parallel computing framework for IPython, does this...

docs/source/development/config.rst
takluyver: This generality is a bit confusing now, because there are only two applications we maintain that use this, and the kernel isn't really what most people would think of as an application.

Maybe be more specific:

> The IPython kernel will load its own config file *after* `ipython_config.py`.

docs/source/development/config.rst
```diff
@@ -93,7 +90,7 @@ Profiles
 ========
 
 A profile is a directory containing configuration and runtime files, such as
-logs, connection info for the parallel apps, and your IPython command history.
+logs, connection info, and your IPython command history.
```
takluyver: It no longer contains connection info normally, because that's in a Jupyter location. Maybe it does for ipyparallel (@minrk?), but that's a bit confusing, so I'd probably not mention it here.

minrk: Yes, it does still contain IPython parallel connection info, and not any other. So while this statement wasn't true for several releases, it is again now.

Carreau:
Thanks, comments addressed !

willingc:
Thanks @Carreau

Carreau:
Note: backported despite what our tool says apparently.

PR #9864, opened 2016-08-12
cleanup unused code from utils/process.py

anantkaushik89:
Removing dead code from utils/process.py which was not used anywhere in the codebase.


IPython/utils/process.py
```diff
@@ -41,6 +41,9 @@ def find_cmd(cmd):
         from IPython.utils.process import pycmd2argv
         argv = pycmd2argv(get_ipython_module_path('IPython.terminal.ipapp'))
 
+    Note, The code for pycmd2argv has been removed now as it was not used
+    anywhere. get_ipython_module_path should give us that same result.
```
takluyver: If we're removing this function,  we should fix the example not to use it, rather than adding a note that the example is outdated.

The modern advice would be to launch IPython using Python's ability to launch a module:

```
argv = [sys.executable, '-m', 'IPython']
```

takluyver:
Milestoned for 6.0, because it removes a function, and it's possible other packages imported and loaded that.

Can you also add a brief note describing this in the [whatsnew/pr directory](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr) of the docs? Give it a filename starting with `incompat-`, so that it will be listed under 'Incompatible changes'.

IPython/utils/process.py
anantkaushik89: Hi @takluyver , thank you for reviewing the pull request. I do have one doubt.

What I understood from pycmd2argv and find_cmd functions, the need is to give the actual path where a module resides. And the previous author had mentioned that the pycmd2argv was a function used for IPython modules.

Can I replace the above set of lines with this example ?

        import sys 
        argv = [sys.executable, get_ipython_module_path('IPython.terminal.ipapp')]

Please let me know if you are suggesting something on other lines.

takluyver: That would be an accurate replacement of what the example currently does.

However, since we wrote this example, Python has added the `-m` flag, which provides an easier way to do what the example is trying to provide. And because of changes to our code, I don't think the original example actually works as intended any more. So my suggestion changes what the example code actually does, but fixes it for what it's intended to do.

anantkaushik89: Thanks @takluyver . Could you please help in one more doubt? I am a newbie to this project and thus have these doubts. Please pardon me for this.

So what I need to do is mention the array as 

   import sys
   argv = [sys.executable, '-m', get_ipython_module_path('IPython.terminal.ipapp')]

Reason I think we need get_ipython_modul_path is that the given module might not be in the same hierarchy mentioned.

Could you please correct me here? Again, thanks a ton for your patience.

takluyver: No problem. No, you don't need to use `get_ipython_module_path` there. The `-m` flag takes the importable name of a module (like `IPython.terminal.ipapp`), not the path (like `.../IPython/terminal/ipapp.py`).

Using `sys.executable` means it will run with the same Python interpreter, which should find the same copy of IPython you're calling it from. There are ways to break that, but we needn't worry about them.

anantkaushik89:
@takluyver pardon me for the delay. Could you please let me know if things look fine to you now?

IPython/utils/process.py
```diff
@@ -37,9 +37,8 @@ def find_cmd(cmd):
     is a risk you will find the wrong one.  Instead find those using the
     following code and looking for the application itself::
 
-        from IPython.utils.path import get_ipython_module_path
-        from IPython.utils.process import pycmd2argv
-        argv = pycmd2argv(get_ipython_module_path('IPython.terminal.ipapp'))
+        import sys
+        argv = [sys.executable, '-m', 'IPython.terminal']
```
takluyver: This doesn't actually work:

```
$ python -m IPython.terminal
/usr/bin/python: No module named IPython.terminal.__main__; 'IPython.terminal' is a package and cannot be directly executed
```

The module name needs to be either `IPython` or `IPython.terminal.ipapp`.

anantkaushik89:
@takluyver Corrected the example based on your feedback.

willingc:
@takluyver LGTM. Thanks @anantkaushik89

takluyver:
Thanks!

PR #9867, opened 2016-08-12
Better support for KeyboardInterrupt during prompt

jdemeyer:
The `prompt_toolkit` library catches the CTRL-C key and handles it specifically. However, when an interrupt happens by other means or at another time, the internal state is no longer consistent. To reproduce:

```
In [1]: from signal import *

In [2]: def handler(num, frame): raise KeyboardInterrupt

In [3]: signal(SIGALRM, handler); alarm(1)
Out[3]: 0

In [4]:                                                                                                                                                                 

KeyboardInterrupt escaped interact()


In [4]:                                                                                                                                                                 
  File "/usr/local/src/sage-config/local/bin/ipython", line 5, in <module>
    start_ipython()
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/IPython/__init__.py", line 119, in start_ipython
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/traitlets/config/application.py", line 596, in launch_instance
    app.start()
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/IPython/terminal/ipapp.py", line 348, in start
    self.shell.mainloop()
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/IPython/terminal/interactiveshell.py", line 402, in mainloop
    self.interact()
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/IPython/terminal/interactiveshell.py", line 385, in interact
    code = self.prompt_for_code()
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/IPython/terminal/interactiveshell.py", line 321, in prompt_for_code
    pre_run=self.pre_prompt, reset_current_buffer=True)
  File "build/bdist.linux-x86_64/egg/prompt_toolkit/interface.py", line 389, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
  File "build/bdist.linux-x86_64/egg/prompt_toolkit/eventloop/posix.py", line 55, in run
    assert not self._running
AssertionError
```

This patch fixes this by explicitly stopping the prompt.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -439,8 +439,14 @@ class TerminalInteractiveShell(InteractiveShell):
             try:
                 self.interact()
                 break
-            except KeyboardInterrupt:
-                print("\nKeyboardInterrupt escaped interact()\n")
+            except KeyboardInterrupt as e:
+                print("\n%s escaped interact()\n" % type(e).__name__)
+                # An interrupt during the eventloop will mess up the
+                # internal state of the prompt_toolkit library.
+                # Stopping the eventloop fixes this, see
+                # https://github.com/ipython/ipython/pull/9867
+                if hasattr(self, '_eventloop'):
+                    self._eventloop.stop()
 
     _inputhook = None
     def inputhook(self, context):


IPython/terminal/interactiveshell.py
```diff
@@ -401,8 +401,10 @@ def mainloop(self, display_banner=DISPLAY_BANNER_DEPRECATED):
             try:
                 self.interact()
                 break
-            except KeyboardInterrupt:
-                print("\nKeyboardInterrupt escaped interact()\n")
+            except KeyboardInterrupt as E:
```
takluyver: Following convention, let's use a lowercase e for the exception.

takluyver:
This looks reasonable, but do you have a way to reproduce the issue, so I can test it interactively?

jdemeyer:
> This looks reasonable, but do you have a way to reproduce the issue, so I can test it interactively?

The easiest is using `alarm()`, see updated description above. Alternative: send a `SIGINT` directly to the process.

takluyver:
Thanks.

Pinging @jonathanslenders to be aware of this. I'll mark it for 5.2 in case he says it's the wrong way to do it.

jonathanslenders:
This is okay to handle it this way, however, I think it can be (should be?) fixed in prompt_toolkit as well. (A try/finally somewhere should do it.)

Just one remark, we are handling one specific case here. The one where a ``KeyboardInterrupt`` exception is raised. The signal handler could raise another exception as well, possibly breaking other code.

If I recall correctly, signal handlers should actually *never* raise exceptions, because this breaks the API that functions expose. As a caller of a function, you assume a certain contract. This is defined by the inputs, the outputs and the exceptions that the function raises. As a caller of a function, you have then the responsibility to either handle these exceptions or let them propagate. If it you decide to let an exception propagate, then it becomes part of the callers API. Now, if a signal handler would throw in an exception, it will break the contract that these functions have with each other.

Similar, signal handlers should probably not acquire locks or call code that acquires locks, because the lock could be hold by the main thread, which is unable to release it as long as the signal handler runs. Signal handlers should probably also not touch global variables, if it's not able to do it atomically. So what you can do in a signal handler without breaking applications is essentially limited.

As far as I understand, a few things that are safe from a signal handler are: appending something to a list; writing something to a unix pipe, or setting a boolean flag.

jonathanslenders:
FYI: http://jvns.ca/blog/2016/06/13/should-you-be-scared-of-signals/

> If you register a signal handler, it’s called in the middle of whatever code you happen to be running. This sets up some very onerous restrictions on what a signal handler can do: it can’t assume that any locks are unlocked, any complex data structures are in a reliable state, etc. The restrictions are stronger than the restrictions on thread-safe code, since the signal handler interrupts and stops the original code from running. So, for instance, it can’t even wait on a lock, because the code that’s holding the lock is paused until the signal handler completes. This means that a lot of convenient functions, including the stdio functions, malloc, etc., are unusable from a signal handler, because they take locks internally.

jdemeyer:
> Just one remark, we are handling one specific case here. The one where a KeyboardInterrupt exception is raised. The signal handler could raise another exception as well, possibly breaking other code.

Or there could be other exceptions (unrelated to signals) due to bugs.

> If I recall correctly, signal handlers should actually never raise exceptions, because this breaks the API that functions expose. As a caller of a function, you assume a certain contract. This is defined by the inputs, the outputs and the exceptions that the function raises. As a caller of a function, you have then the responsibility to either handle these exceptions or let them propagate. If it you decide to let an exception propagate, then it becomes part of the callers API. Now, if a signal handler would throw in an exception, it will break the contract that these functions have with each other.

I agree in general, but I do think that `KeyboardInterrupt` is a special case. For non-interactive code, it's fine to not care about `KeyboardInterrupt`. However, for interactive programs like IPython, you should be prepared to get a `KeyboardInterrupt`.

jdemeyer:
> FYI: http://jvns.ca/blog/2016/06/13/should-you-be-scared-of-signals/

That's about the low-level Unix signals, not about how Python deals with signals. So I don't think it is very relevant.

jonathanslenders:
Hi @jdemeyer ,

Unfortunately, no. Thinking about it, the pull request is wrong and it can still break stuff. It only hides whatever will break.

> Or there could be other exceptions (unrelated to signals) due to bugs.

In case of a bug, these exception should propagate and probably crash the application. (See PEP20.)

> That's about the low-level Unix signals, not about how Python deals with signals. So I don't think it is very relevant.

These are still signals. The only difference is that in Python, the signal handler is only called after the GIL is released. Everything else is exactly the same as a signal handler in C and it has the same restrictions.

I think that you confuse what is the cause and the handler of ``KeyboardInterrupt`` in a normal situation.
When using GNU readline, it's the readline code that raises KeyboardInterrupt (when Ctrl_c is pressed) and the application that should handle it. Not the other way around. What you simulated was the application that raised ``KeyboardInterrupt`` and prompt_toolkit (a readline replacement) that has to catch it.) In the case of readline, the ``KeyboardInterrupt`` exception is clearly part of the signature of the ``(raw_)input`` function. Every piece of code that calls this function is supposed to either handle that or let it propagate. Code that does not call ``input`` is *not* supposed to handle ``KeyboardInterrupt``. Really, there is no way that arbitrary code should be able to handle ``KeyboardInterrupt`` if it doesn't call ``input``. This would require putting a ``try/finally`` around almost every line of code in the whole program.

By the way, by default, when calling ``prompt_toolkit.shortcuts.prompt``, it will raise ``KeyboardInterrupt`` when Control-C has been pressed. It raises it because it is coded that way, and it's part of the ``prompt`` signature that it can raise that exception.

I hope this makes sense.

jonathanslenders:
So, if we would merge this pull request, and somebody else would raise an exception from a signal handler like in your example. It won't give this exception, but it can turn any other data structure in an inconsistent state, causing indirectly arbitrary errors in all places of the code.

Take for instance these two lines of code:
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/buffer.py#L427
These are supposed to be atomic. There is no lock around it, because these lines of code are not accessed from any other thread. But if a signal handler interferes it suddenly isn't atomic anymore. The data structure becomes corrupt, and any other exception could occur. In this case we could get an ``IndexError``.

Maybe you have ever seen a ``@timeout`` decorator that uses exceptions and signals to set time outs around functions. People try this, but it has exactly the same issues and is *not* reliable.

It's the responsibility of the author of a signal handler to write code that is not just thread-safe, but even stronger, safe to put in a signal handler.

jonathanslenders:
Just curious. What is your use case?
Maybe we could still merge it, while it's say 90% of the time probably going to be safe.

I realize that the default signal handler for SIGINT actually does raise a KeyboardInterrupt, which is probably a case of practicality beats purity. But if that exception is caught at some point, the application is not guaranteed to remain in a consistent state.

If it is important, I guess we could replace some signal handlers right before calling prompt_toolkit, and execute them in a thread safe way.

jdemeyer:
> The only difference is that in Python, the signal handler is only called after the GIL is released.

This "only" difference is really a very major difference.

> Everything else

Which is not much, except maybe the thing about locks.

jdemeyer:
> Just curious. What is your use case?

My main use case is `alarm()`. Typical example:
```
alarm(5); some_computation();
```
This is meant to interrupt `some_computation()` after 5 seconds. However, if `some_computation()` happens to take less than 5 seconds, it will crash IPython. This used to work correctly in IPython 4 and earlier.

There is another issue with IPython 5 that it breaks GDB debugging because of CTRL-C not working anymore. But I have not investigated this further.

jdemeyer:
> It's the responsibility of the author of a signal handler to write code that is not just thread-safe, but even stronger, safe to put in a signal handler.

`<rant>`
Python has decided that it's OK to `raise KeyboardInterrupt` in a signal handler, since they do that by default. I would argue that it's the job of the language designer (in this case Python) to design a language which offers proper control over signals. I agree that it is impossible to write Python code which is robust under signals. I see this as a shortcoming in Python. While reading https://github.com/jonathanslenders/python-prompt-toolkit, I saw "Pure Python" listed as first feature and immediately thought: "that's a bug, not a feature". I am a big fan of [Cython](http://cython.org/), where it is possible to have full control over signal handling.
`</rant>`

jdemeyer:
> So, if we would merge this pull request, and somebody else would raise an exception from a signal handler like in your example. It won't give this exception, but it can turn any other data structure in an inconsistent state, causing indirectly arbitrary errors in all places of the code.
> 
> Maybe you have ever seen a @timeout decorator that uses exceptions and signals to set time outs around functions. People try this, but it has exactly the same issues and is not reliable.

Of course you are right. Given that `KeyboardInterrupt` can already introduce corruption, I don't see the problem with this pull request, which seems to work fine in most cases. You can either try to fix most cases, or say "we cannot fix everything, so we should just fix nothing instead".

Remember that this is about IPython, where (IMHO) user-friendlyness should be more important than complete robustness.

jonathanslenders:
> I don't see the problem with this pull request, which seems to work fine in most cases. You can either try to fix most cases, or say "we cannot fix everything, so we should just fix nothing instead".

Agreed. Definitely about user-friendlyness. But I'll try later today if there is something that I can do in prompt_toolkit to handle it in a safer way. (I have some ideas.) If I find a safer solution, I'd prefer that. Otherwise, we'll merge this as a best effort of what is technically possible.

Regarding Python's decision on ``KeyboardInterrupt``. I think that's mostly for use in the REPL, where you don't care if interrupted code ends up in a broken state, and where the practicality goes above the correctness. In a normal application, the only sensible actions for ``KeyboardInterrupt`` are to ignore it or to quit the application anyway.
Looking at how rust (which is by design free of race conditions and thread safe) probably won't allow any custom code to be executed in a signal handler and enforces people to use signalfd should say enough.
https://github.com/rust-lang/rfcs/issues/1368

Sorry if I sounded too negative in the first place. I really want to avoid that we are concealing/suppressing bugs, rather than solving them. But you're right that we should do the best effort to make it user-friendly anyway.

jdemeyer:
> I think that's mostly for use in the REPL, where you don't care if interrupted code ends up in a broken state, and where the practicality goes above the correctness.

This is **exactly** the use case of IPython.

jonathanslenders:
Yes, but this is meant for when the REPL executes code. (Because somebody presses Ctrl-c) It's not while reading the input, that's different.

jonathanslenders:
Hi all, I made a pull request to prompt_toolkit that should handle exceptions coming from signals in a safe way: https://github.com/jonathanslenders/python-prompt-toolkit/pull/382

@jdemeyer: Could you test this pull request? (check out that prompt_toolkit branch and install it.)

I'm not sure whether this is something I want to do by default for all prompt_toolkit applications or make it opt-in (for IPython). I guess it depends on the cost of replacing signal handlers.

IPython/terminal/interactiveshell.py
```diff
@@ -439,8 +439,14 @@ def mainloop(self, display_banner=DISPLAY_BANNER_DEPRECATED):
...
                 self.interact()
                 break
-            except KeyboardInterrupt:
-                print("\nKeyboardInterrupt escaped interact()\n")
+            except KeyboardInterrupt as e:
+                print("\n%s escaped interact()\n" % type(e).__name__)
+                # An interrupt during the eventloop will mess up the
+                # internal state of the prompt_toolkit library.
+                # Stopping the eventloop fixes this, see
+                # https://github.com/ipython/ipython/pull/9867
```
jonathanslenders: I think if we put the following two lines in a ``finally``, it will be more safe against other exceptions.

takluyver: Just to check, it's safe to call `.stop()` on an already-stopped eventloop, right?

jonathanslenders: Yes, just like closing a file twice should not hurt.

takluyver:
@jdemeyer are you happy to move the extra lines into a `finally:` block as @jonathanslenders  suggested?

Carreau:
Should we bump this to 5.3 ?

takluyver:
Continued in #10215 as I think we were just waiting on a simple change to merge this. Closing here.

jdemeyer:
Sorry for that... I had a million other things on my mind.

PR #9870, opened 2016-08-13
[WIP] Add support for MathML mime type

gnestor:
Related to https://github.com/jupyter/notebook/pull/1672

I'd like to support the following usage: 

``` py
from IPython.core.display import MathML
MathML(r"""<math>
    <mrow>
        <mi>a</mi> <mo>&InvisibleTimes;</mo> <msup><mi>x</mi><mn>2</mn></msup>
        <mo>+</mo><mi>b</mi><mo>&InvisibleTimes;</mo><mi>x</mi>
        <mo>+</mo><mi>c</mi>
    </mrow>
<math>""")
```

MathML is rendered by MathJax, so the following currently works in the notebook:

``` py
from IPython.core.display import HTML
HTML(r"""<math>
    <mrow>
        <mi>a</mi> <mo>&InvisibleTimes;</mo> <msup><mi>x</mi><mn>2</mn></msup>
        <mo>+</mo><mi>b</mi><mo>&InvisibleTimes;</mo><mi>x</mi>
        <mo>+</mo><mi>c</mi>
    </mrow>
<math>""")
```

Therefore, the display method for `MathML` can just use the `_repr_html_`.

I would also like to support the MathML mime type which is `application/mathml+xml`.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -473,7 +473,7 @@ class Math(TextDisplayObject):
         
     def _repr_html_(self):
         if self.format == u'mathml':
-            return self.data
+            return "<math>%s</math>" % self.data
 
 
 class Latex(TextDisplayObject):
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -25,10 +25,9 @@ from IPython.testing.skipdoctest import skip_doctest
 __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
 'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json',
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
-'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'JSON', 'Javascript',
-'Image', 'clear_output', 'set_matplotlib_formats', 'set_matplotlib_close',
-'publish_display_data']
-
+'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'MathML', 'SVG',
+'JSON', 'Javascript', 'Image', 'clear_output', 'set_matplotlib_formats',
+'set_matplotlib_close', 'publish_display_data'] 
 #-----------------------------------------------------------------------------
 # utility functions
 #-----------------------------------------------------------------------------
@@ -91,6 +90,7 @@ def publish_display_data(data, metadata=None, source=None):
     * text/html
     * text/markdown
     * text/latex
+    * application/mathml+xml
     * application/json
     * application/javascript
     * image/png
@@ -475,8 +475,13 @@ class Latex(TextDisplayObject):
 
     def _repr_latex_(self):
         return self.data
+        
+        
+class MathML(TextDisplayObject):
 
-
+    def _repr_html_(self):
+        return self.data
+        
 class SVG(DisplayObject):
 
     # wrap data in a property, which extracts the <svg> tag, discarding
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -467,8 +467,13 @@ class Markdown(TextDisplayObject):
 class Math(TextDisplayObject):
 
     def _repr_latex_(self):
-        s = self.data.strip('$')
-        return "$$%s$$" % s
+        if self.format == u'latex' or not self.format:
+            s = self.data.strip('$')
+            return "$$%s$$" % s
+        
+    def _repr_html_(self):
+        if self.format == u'mathml':
+            return self.data
 
 
 class Latex(TextDisplayObject):
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -775,6 +775,23 @@ class LatexFormatter(BaseFormatter):
     print_method = ObjectName('_repr_latex_')
 
 
+class MathMLFormatter(BaseFormatter):
+    """A MathML formatter.
+
+    To define the callables that compute the MathML representation of your
+    objects, define a :meth:`_repr_mathml_` method or use the :meth:`for_type`
+    or :meth:`for_type_by_name` methods to register functions that handle
+    this.
+
+    The return value of this formatter should be a valid HTML snippet that
+    could be injected into an existing DOM. It should *not* include the
+    ```<html>`` or ```<body>`` tags.
+    """
+    format_type = Unicode('application/mathml+xml')
+
+    print_method = ObjectName('_repr_html_')
+
+
 class JSONFormatter(BaseFormatter):
     """A JSON string formatter.
 
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -775,24 +775,6 @@ class LatexFormatter(BaseFormatter):
     print_method = ObjectName('_repr_latex_')
 
 
-class MathMLFormatter(BaseFormatter):
-    """A MathML formatter.
-
-    To define the callables that compute the MathML representation of your
-    objects, define a :meth:`_repr_mathml_` method or use the :meth:`for_type`
-    or :meth:`for_type_by_name` methods to register functions that handle
-    this.
-
-    The return value of this formatter should be a valid HTML snippet that
-    could be injected into an existing DOM. It should *not* include the
-    ```<html>`` or ```<body>`` tags.
-    """
-    # format_type = Unicode('application/mathml+xml')
-    format_type = Unicode('text/html')
-
-    print_method = ObjectName('_repr_html_')
-
-
 class JSONFormatter(BaseFormatter):
     """A JSON string formatter.
 
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -787,7 +787,8 @@ class MathMLFormatter(BaseFormatter):
     could be injected into an existing DOM. It should *not* include the
     ```<html>`` or ```<body>`` tags.
     """
-    format_type = Unicode('application/mathml+xml')
+    # format_type = Unicode('application/mathml+xml')
+    format_type = Unicode('text/html')
 
     print_method = ObjectName('_repr_html_')
 


gnestor:
I'm not able to test this against my dev version of ipython. is there anything that I need to do after I make changes to the source?

IPython/core/formatters.py
```diff
@@ -775,6 +775,23 @@ class LatexFormatter(BaseFormatter):
...
+    or :meth:`for_type_by_name` methods to register functions that handle
+    this.
+
+    The return value of this formatter should be a valid HTML snippet that
+    could be injected into an existing DOM. It should *not* include the
+    ```<html>`` or ```<body>`` tags.
+    """
+    format_type = Unicode('application/mathml+xml')
+
+    print_method = ObjectName('_repr_html_')
```
takluyver: Shouldn't this be `_repr_mathml_`?

IPython/core/display.py
```diff
@@ -475,8 +475,13 @@ class Latex(TextDisplayObject):
 
     def _repr_latex_(self):
         return self.data
+        
+        
+class MathML(TextDisplayObject):
```
takluyver: @minrk do you think it's better to have a separate `MathML` display class, or combine the functionality into the `Math` class with a switch to distinguish between formats?

We do the former for SVG, but the latter for JPG/PNG in the `Image` class. I could go either way on this.

takluyver:
A development install should be possible with `pip install -e .` (or `python setup.py symlink`, which is specific to IPython). Then you should have an `iptest` command to run.

IPython/core/display.py
gnestor: I was thinking of the latter since `Math` currently only renders LaTeX and if we now support other mathematical notation formats like MathML, it would imply that `Math` also render them.

IPython/core/formatters.py
gnestor: I was wondering about this... How does `_repr_mathml_` map to the `MathML` display method? In the case of MathML, it should be displayed as HTML, so it should essentially use the `HTML` formatters/display methods. The only difference is that we want to support the MathML mime type vs. just `text/html`.

minrk: > How does _repr_mathml_ map to the MathML display method?

That's what `print_method` does. It looks for methods with this name, and calls it. Then `format_type` is used to put that in the mathml+xml mime-type field in the format dict.

If it should actually be 'displayed as HTML', then it shouldn't be a mathml mime-type, but the appropriate HTML elements, and actually use `_repr_html_`, and never the mathml mime-type.

I could have sworn that a while ago we added a `_repr_mime_` that would return a mimebundle, so we wouldn't need to declare and register classes and methods for every supported mime-type, but apparently we haven't managed to do that yet. Given the spec, and the DisplayFormatter class, we really don't need dedicated methods and classes for every mime-type, as all arbitrary mime-types are already supported, even those we don't have classes or methods for.

IPython/core/display.py
takluyver: I don't see an obvious reason why `Math` shouldn't also support MathML. It would  have to have an extra switch, e.g. `Math('...', format='mathml')`. But I'm also not especially attached to doing it that way, I just think it's worth considering.

IPython/core/formatters.py
gnestor: I updated this to `format_type = Unicode('text/hmtl')`

IPython/core/display.py
```diff
@@ -467,16 +467,26 @@ def _repr_markdown_(self):
...
     def _repr_latex_(self):
-        s = self.data.strip('$')
-        return "$$%s$$" % s
+        if self.format == u'latex' or not self.format:
+            s = self.data.strip('$')
+            return "$$%s$$" % s
+        
+    def _repr_html_(self):
+        if self.format == u'mathml':
+            return self.data
```
gnestor: The intention here is to render LaTeX by default, but if a format switch is used `Math("<math>...</math>", format="mathml")` is used, then render as HTML.

gnestor: Idea: We could also use regex and if `self.data` matches an xml pattern, render as HTML?

gnestor:
> A development install should be possible with pip install -e . (or python setup.py symlink, which is specific to IPython). Then you should have an iptest command to run.

@takluyver I did mean running unit tests or test suites but actually testing that changes that I made to ipython in notebook. I tried:

```py
from IPython.core.display import MathML
MathML(r"""<math>
    <mrow>
        <mi>a</mi> <mo>&InvisibleTimes;</mo> <msup><mi>x</mi><mn>2</mn></msup>
        <mo>+</mo><mi>b</mi><mo>&InvisibleTimes;</mo><mi>x</mi>
        <mo>+</mo><mi>c</mi>
    </mrow>
<math>""")
```

and it returned:

```py
ImportError                               Traceback (most recent call last)
<ipython-input-9-a886233ec497> in <module>()
----> 1 from IPython.core.display import HTML, MathML
      2 MathML(r"""<math>
      3     <apply>
      4         <plus/>
      5         <apply>

ImportError: cannot import name 'MathML'
```

So either the ipython used by notebook is not my dev version or I didn't expose `IPython.display.MathML` method correctly...

gnestor:
> A development install should be possible with pip install -e . (or python setup.py symlink, which is specific to IPython). Then you should have an iptest command to run.

@takluyver I didn't mean running unit tests but actually testing changes that I make to ipython in notebook. I tried:

```py
from IPython.core.display import MathML
MathML(r"""<math>
    <mrow>
        <mi>a</mi> <mo>&InvisibleTimes;</mo> <msup><mi>x</mi><mn>2</mn></msup>
        <mo>+</mo><mi>b</mi><mo>&InvisibleTimes;</mo><mi>x</mi>
        <mo>+</mo><mi>c</mi>
    </mrow>
<math>""")
```

and it returned:

```py
ImportError                               Traceback (most recent call last)
<ipython-input-9-a886233ec497> in <module>()
----> 1 from IPython.core.display import HTML, MathML
      2 MathML(r"""<math>
      3     <apply>
      4         <plus/>
      5         <apply>

ImportError: cannot import name 'MathML'
```

So either the ipython being used by notebook is not my dev version or I didn't expose the `IPython.display.MathML` method correctly...

IPython/core/formatters.py
```diff
@@ -775,6 +775,24 @@ class LatexFormatter(BaseFormatter):
...
+    objects, define a :meth:`_repr_mathml_` method or use the :meth:`for_type`
+    or :meth:`for_type_by_name` methods to register functions that handle
+    this.
+
+    The return value of this formatter should be a valid HTML snippet that
+    could be injected into an existing DOM. It should *not* include the
+    ```<html>`` or ```<body>`` tags.
+    """
+    # format_type = Unicode('application/mathml+xml')
+    format_type = Unicode('text/html')
```
minrk: If this is the format, then MathMLFormatter shouldn't exist.

IPython/core/display.py
minrk: I like `Math` with a switch.

IPython/core/display.py
minrk: When given math for latex, we don't expect it to include the `$$` wrapper, that is `Math('...')` replaces the `$$` as the wrapper for the math inside. To match that for mathml might be to *add* the `<math>` tag, which might make it a bit harder to identify math by regex.

Right now, you can do:

```python
Math('a=5')
```

so I would expect

```python
Math('a=5', format='mathml')
```

to probably work as well.

takluyver:
I think it should still have `MathMLFormatter` and the `_repr_mathml_` special method. MathML can be embedded in HTML, but it's still a separate format (like SVG).

minrk:
@takluyver to me, the question is whether it should be distinguished in the file format, nbconvert, etc. as something distinct from HTML, rather than a supported HTML element that should be transparent to Jupyter. I'm not sure that mathml should be distinguished in that way because, unlike SVG, it is not generally distinct file (I've never seen a mathml file) that needs special treatment. Where are we likely to want to take MathML out of the context of HTML and handle it separately from normal HTML?

takluyver:
Fair enough. In principle, I can imagine there being tools that can handle MathML but not arbitrary HTML, and it would be nice if they could just check mimetypes rather than trying to figure out if HTML contains MathML. But in practice, I don't know of any such tools either.

If you think MathML should be embedded into HTML output, do you also want us to revert jupyter/notebook#1672, which added support for MathML as a separate mimetype?

minrk:
@takluyver good point about jupyter/notebook#1672, we should probably do the same thing on both sides, whichever we decide. I could go either way, but currently lean slightly toward treating MathML as 'just HTML'  unless someone has an existing practical use case for not doing so. I'd like to hear from @poke1024 and @rgbkrk on whether they think MathML should be treated with a dedicated mime-type or as HTML.

gnestor:
@minrk Once this is merged, I will update notebook to reflect our decision...

gnestor:
@takluyver @minrk To step back: If we're not adding a MathML mime type or formatter, then we are simply adding a switch to the `Math` display method to allow it to render MathML without `<math>` tags. Is this worth pursuing or should we just leave it as is (because users can currently render MathML using the `HTML` display method).

minrk:
@gnestor good question! I don't know what the current state of demand is for MathML support. I haven't noticed demand coming from the IPython side for producing MathML more conveniently. The general support in Jupyter is motivated primarily by @poke1024's mathics kernel, I believe, which is why I would like to know how significant the distinction is between MathML as a snippet mimetype vs an HTML element.

gnestor:
I think that we should just close this until we get some requests for MathML support in ipython. If you agree, I will revert https://github.com/jupyter/notebook/pull/1672.

minrk:
That makes sense to me.

gnestor:
@minrk Ok. How do you suggest I go about reverting https://github.com/jupyter/notebook/pull/1672?

minrk:
@gnestor the easiest way is with the revert button on the PR:

screen shot 2016-08-23 at 10 34 46

If you want to revert less than all of the PR, you might want to compose the commit by hand, or start with that auto-revert and pull it locally, modifying it to your liking.

poke1024:
@minrk @gnestor Sorry for not answering sooner. For Mathics alone, a dedicated MathML type is not an advantage in terms of clarity, since Mathics (speaking from its current design) will sometimes output MathML that contains SVGs that (using foreignObjects) contains HTML that contains, again, MathML that contains SVGs and so on. We do this to mix formula and graphics content, so graphics (SVG) might contain some MathML formula text than in turn might contain some graphics again (for example an exponent that is not a letter but an image); the output might end up being a wild mix of formats embedded into each other that stresses the content standards involved to its limits. The whole thing is, on the outer level, just a block of HTML.

This is a very special case, I believe. Most other kernels using math output will probably stick to very clean MathML-only content. Yet, as you said, it's hard to tell, as long as there are no other users.

gnestor:
Ok, given that I think it makes sense to remove MathML mime types from notebook. 

@minrk See https://github.com/jupyter/notebook/pull/1710 which reverts https://github.com/jupyter/notebook/pull/1672.

minrk:
@poke1024 thanks for the info, your perspective is useful on these things.

@gnestor thanks for working it out!

PR #9871, opened 2016-08-14
Switch master to development of 6.0

takluyver:
I'm thinking about this because we've just released 5.1, but I'm happy for it to wait a bit longer if we prefer. As planned, 6.0 will require Python 3, while we will continue to support 5.x for some time for Python 2 users.

Before merging this, we should send a message to the mailing list for any Python 2 users following master, and make a 5.x branch off master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,30 +5,12 @@ python:
     - 3.5
     - 3.4
     - 3.3
-    - 2.7
-    - pypy
 sudo: false
 before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
 install:
     - pip install "setuptools>=18.5"
-    # Installs PyPy (+ its Numpy). Based on @frol comment at:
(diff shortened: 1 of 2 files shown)


willingc:
@takluyver Excellent! 

In light of Mozilla's recent support of PyPy (https://morepypy.blogspot.com/2016/08/pypy-gets-funding-from-mozilla-for.html) for Python 3, we should consider testing the dev version of PyPy. I wouldn't hold up merging this PR, but there's likely benefit to keep testing PyPy in some way.

takluyver:
Yes, at some point we'll probably add pypy back, but I think pypy3 is still on Python 3.2, which we don't support. I assume it will be a while before the Mozilla funding produces results.

willingc:
@takluyver Agree :-) Should we open a reminder issue to add back pypy to Travis when support for 3.4/3.5 is further along? So we don't forget to do so later.

minrk:
It begins! @willing I opened #9874 for testing on PyPy once there is a PyPy that can run IPython 6.

takluyver:
I have updated 5.x to just after I released 5.1, and pushed it here. I'll send a message to the mailing list.

PR #9872, opened 2016-08-14
Deprecate -e flag for %notebook magic

takluyver:
Maybe the magic did something else in the past? Now the `-e` or `--export` flag is just confusing.

I've left it in place so it will still work, but have no effect.

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -561,7 +561,6 @@ Currently the magic system has the following functions:""",
 
         This function can export the current IPython history to a notebook file.
         For example, to export the history to "foo.ipynb" do "%notebook foo.ipynb".
-        To export the history to "foo.py" do "%notebook foo.py".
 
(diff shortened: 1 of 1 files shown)


IPython/core/magics/basic.py
```diff
@@ -563,22 +560,25 @@ def notebook(self, s):
         """Export and convert IPython notebooks.
 
         This function can export the current IPython history to a notebook file.
-        For example, to export the history to "foo.ipynb" do "%notebook -e foo.ipynb".
-        To export the history to "foo.py" do "%notebook -e foo.py".
+        For example, to export the history to "foo.ipynb" do "%notebook foo.ipynb".
+        To export the history to "foo.py" do "%notebook foo.py".
```
minrk: I think `%notebook foo.py` won't work, only notebooks can be written.

takluyver: Good point, removed

PR #9873, opened 2016-08-14
Warn about universal wheels in the docs.

Carreau:
cc @takluyver


minrk:
We should also consider stopping the 2/3 suffix for ipython 6. Not necessarily do it, but at least consider it.

Carreau:
Well... for sure the `2` suffix :-) I'm mainly concern about the 5.x LTS, and this is why I send a doc change.

takluyver:
I've sent you a  small PR to tweak the wording a bit: Carreau/ipython#13 . With that, I think it's fine.

Carreau:
> I've sent you a small PR to tweak the wording a bit: Carreau/ipython#13 . With that, I think it's fine.

Merged in.

takluyver:
Likewise :)

PR #9876, opened 2016-08-15
More changes to indicate Python 3 requirement

takluyver:


setup.py
```diff
@@ -239,6 +239,7 @@ def run(self):
 extras_require['all'] = everything
 
 if 'setuptools' in sys.modules:
+    setuptools_extra_args['python_requires'] = '>=3.3'
```
takluyver: This requires setuptools >= 24.2.0:
https://setuptools.readthedocs.io/en/latest/history.html#v24-2-0

damianavila: Don't you need to require this on TravisCI? https://github.com/ipython/ipython/blob/master/.travis.yml#L13

Btw, should not be something to detect/request this requirement at the setup.py level?

Carreau:
🍰  !

I think this would likely need an announce on the mailing list.

setup.py
damianavila: @takluyver ^^^, in case you miss the previous notification :wink:

takluyver: Thanks, yes, we should bump the requirement on Travis.

PR #9885, opened 2016-08-18
Add support for ipympl backend to magic

jenshnielsen:
This enables you to do `%matplotlib ipympl` to use https://github.com/matplotlib/jupyter-matplotlib backend. In the long term I expect that the 

Needs to go together with the matching change in ipykernel. Which I will open a pull request for in a moment. 

I am not sure how to best handle the interdependency between this and ipykernel
With this change the backend will be added to the list of available backends via
`%matplotlib --list` but will not actually be functional without the change in ipykernel

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -23,7 +23,8 @@ backends = {'tk': 'TkAgg',
             'osx': 'MacOSX',
             'nbagg': 'nbAgg',
             'notebook': 'nbAgg',
-            'inline' : 'module://ipykernel.pylab.backend_inline'}
+            'inline': 'module://ipykernel.pylab.backend_inline',
+            'ipympl': 'module://ipympl.backend_nbagg'}
 
 # We also need a reverse backends2guis mapping that will properly choose which
 # GUI support to activate based on the desired matplotlib backend.  For the
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -212,8 +212,8 @@ def select_figure_formats(shell, formats, **kwargs):
     formats = set(formats)
 
     [ f.pop(Figure, None) for f in shell.display_formatter.formatters.values() ]
-
-    if matplotlib.get_backend().lower() == 'nbagg':
+    mplbackend = matplotlib.get_backend().lower()
+    if mplbackend == 'nbagg' or mplbackend == 'module://ipympl.backend_nbagg':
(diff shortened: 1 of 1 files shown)


jenshnielsen:
We could also consider having `%matplotlib notebook` and `%matplotlib nbagg` conditionally activating ipympl and falling back on the old non widget backend

SylvainCorlay:
Should I remove the enabling in `ipympl.__init__` then?

jenshnielsen:
@SylvainCorlay I think this is probably the way we want to go forward but I am not sure we should remove the init in `ipympl` just yet. 

This should also enable interactive usage but there are some issues that I need to investigate and fix before this is merged. Specifically having a cell with just a single plot command i.e. 

```
plt.plot(range(10))
```
 will show a figure the first time that it is executed but not if it's re executed. That currently works with `%matplotlib notebook` and will work with `ipympl.__init__` if a `plt.show()` is added. I think that `ipympl.__init__` does not currently activate interactive mode in matplotlib so the explicit show is needed

SylvainCorlay:
This looks good to me. @takluyver are all the 5.2-milestoned issues eventually going to be backported or is the backport done as we go?

takluyver:
There's a script in the tools folder that we use to do backports whenever we're thinking about it.

Carreau:
should figure out a way to use entrypoint instead of having all backend hardcoded ?

SylvainCorlay:
> should figure out a way to use entrypoint instead of having all backend hardcoded ?

The part of the mapping that does
```
'tk': 'TkAgg',
'gtk': 'GTKAgg',	
'gtk3': 'GTK3Agg',
'wx': 'WXAgg',
'qt': 'Qt4Agg', 
'qt4': 'Qt4Agg',	
'qt5': 'Qt5Agg',
'osx': 'MacOSX'
'nbagg': 'nbAgg',
```
could be removed if matplotlib backend names were more concise :) The other ones are about support for jupyter.

jenshnielsen:
The problem with backend names is that matplotlib backends are bacically 2 different things lumped together in one. 

* Which rendering engine to use. I.e. Agg, Cairo, Pdf ...
* Which guiframework, GTK2/3, QT4/5, notebook ...

On the matplotlib side we therefore have. GTKAgg and GTKCairo and so on.
IPython only really cares about the second point since thats what determines which eventloop integration to use.

takluyver:
Closing in favour of #10059, which is an updated copy of this.

PR #9893, opened 2016-08-21
Update prompts.py

pietvo:
Get rid of spurious newline after the output prompt (Out[n]: ) in cases where prompt_toolkit is not used (e.g. running ipython inside Emacs)

### DIFF
--- a/IPython/terminal/prompts.py
+++ b/IPython/terminal/prompts.py
@@ -71,4 +71,4 @@ class RichPromptDisplayHook(DisplayHook):
             if self.shell.pt_cli:
                 self.shell.pt_cli.print_tokens(tokens)
             else:
-                print(*(s for t, s in tokens), sep='')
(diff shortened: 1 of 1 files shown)


takluyver:
I think your solution on the mailing list of using `end=''` was neater.

pietvo:
You mean 'neater' as in 'more esthetically pleasing?
I choose the write for a couple of reasons:

1. the only other output statement in that method also uses sys.stdout.write
2. why go through print if all that it eventually will do is the same sys.stdout.write?
3. If I remember correctly, the code in prompt_toolkit also uses sys.stdout.write
4. the print statement is not that natural as it needs non-default parameters for sep and end and if sys.stdout.write had originally been used the bug probably would not have occurred

But feel free to replace it with the print call.

takluyver:
I did mean more aesthetic or readable. But you're the one fixing it, so if you prefer this way, that's fair enough.

Issue #9395, opened 2016-04-13
AttributeError: 'IPCompleter' object has no attribute 'custom_completers'


Carreau:
We seem to set it in `interactiveshell.py`, but it shoudl likely handle the case when it does not exists, 
and should be wrapped in setters/getters.

```
In [1]: from IPython.core.completer import IPCompleter

In [2]: c = IPCompleter(get_ipython())

In [3]: c.complete('foo')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-1faa54daaabd> in <module>()
----> 1 c.complete('foo')

/Users/bussonniermatthias/dev/ipython/IPython/core/completer.py in complete(self, text, line_buffer, cursor_pos)
   1182         # Start with a clean slate of completions
   1183         self.matches[:] = []
-> 1184         custom_res = self.dispatch_custom_completer(text)
   1185         if custom_res is not None:
   1186             # did custom completers produce something?

/Users/bussonniermatthias/dev/ipython/IPython/core/completer.py in dispatch_custom_completer(self, text)
   1094         # for foo etc, try also to find completer for %foo
   1095         if not cmd.startswith(self.magic_escape):
-> 1096             try_magic = self.custom_completers.s_matches(
   1097                 self.magic_escape + cmd)
   1098         else:

AttributeError: 'IPCompleter' object has no attribute 'custom_completers'
```

minrk:
It should probably be declared on the class so it's always defined

Carreau:
> It should probably be declared on the class so it's always defined

Yes, I just leave an easy open bug that does not bother anyone for a new contributor !

thesageinpilani:
@Carreau  sir, i tried something, is this valid, i commented the line in __init__ of IPCompleter object i.e.,
`self.custom_completers = None `
and it gave the following result.  

```
In [1]: from IPython.core.completer import IPCompleter

In [2]: c = IPCompleter(get_ipython())

In [3]: c.complete('foo')

In [4]:
```

the error disappeared, i thought that custome_completers object is being instantiated in the parent class of Completer, However i am a beginner , please review and help me with this, thank you.

Carreau:
Note, this has (partially been already fixed by https://github.com/ipython/ipython/pull/9487) but it now raises the following:

```
In [1]: from IPython.core.completer import IPCompleter

In [2]: c = IPCompleter(get_ipython())

In [3]: c.complete('foo')
---------------------------------------------------------------------------
[...]

PR #9897, opened 2016-08-23
commented self.custom_completers = None

ghost:
regarding AttributeError: 'IPCompleter' object has no attribute 'custom_completers' #9395

I have commented the intialization of null value in the `__init__` of IPCompleter object, since it is subclass of  Completer.py where custom_completers is being created.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -662,7 +662,7 @@ class IPCompleter(Completer):
                          ]
 
         # This is set externally by InteractiveShell
-        self.custom_completers = None
+        # self.custom_completers = None
 
     def all_completions(self, text):
         """


Carreau:
Hi ! 

Thanks a lot for that; that does not seem to do it for me I still get a different error; 

```
In [3]: from IPython.core.completer import IPCompleter

In [4]: c = IPCompleter(get_ipython())

In [5]: c.complete('foo')
<ipython-input-5-1faa54daaabd> in <module>()
----> 1 c.complete('foo')

/Users/bussonniermatthias/dev/ipython/IPython/core/completer.py in comp(*args, **kwargs)
     82     @wraps(complete)
     83     def comp(*args, **kwargs):
---> 84         text, matches =  complete(*args, **kwargs)
     85         if len(matches) == 1:
     86             return text, [matches[0].rstrip()]

/Users/bussonniermatthias/dev/ipython/IPython/core/completer.py in complete(self, text, line_buffer, cursor_pos)
   1191         # Start with a clean slate of completions
   1192         self.matches[:] = []
-> 1193         custom_res = self.dispatch_custom_completer(text)
   1194         if custom_res is not None:
   1195             # did custom completers produce something?

/Users/bussonniermatthias/dev/ipython/IPython/core/completer.py in dispatch_custom_completer(self, text)
   1088
   1089     def dispatch_custom_completer(self, text):
-> 1090         if not self.custom_completers:
   1091             return
   1092

AttributeError: 'IPCompleter' object has no attribute 'custom_completers'
```

Did you committed all your changes ?

thesageinpilani:
i am very sorry sir, i think i may have done something wrong. I will try to fix it. Thank you

Carreau:
> i am very sorry sir, i think i may have done something wrong. I will try to fix it. Thank you

Don't worry; it might be me that have a wrong setup. 

Thanks a lot for wanting to contribute !

thesageinpilani:
sir, i have removed any previous installs of ipython and python and installed python 3.5.2 and running windows 10 and using git i cloned the repo and ran python -m IPython it gave the following messages.

`C:\Users\rkchi\Desktop\repos\ipython>python -m IPython
Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:01:18) [MSC v.1900 32 bit (Intel)]

IPython 6.0.0.dev -- An enhanced Interactive Python.
%quickref -> Quick reference.

In [1]: from IPython.core.completer import IPCompleter

In [2]: c = IPCompleter(get_ipython())

In [3]: c.complete('foo')
C:\Users\rkchi\Desktop\repos\ipython\IPython\core\completer.py in complete(self, text, line_buffer, cursor_pos)
   1203                 for matcher in self.matchers:
   1204                     try:
-> 1205                         self.matches.extend(matcher(text))
   1206                     except:
   1207                         # Show the ugly traceback if the matcher causes an

C:\Users\rkchi\Desktop\repos\ipython\IPython\core\completer.py in python_matches(self, text)
    800                 matches = []
    801         else:
--> 802             matches = self.global_matches(text)
    803         return matches
    804

C:\Users\rkchi\Desktop\repos\ipython\IPython\core\completer.py in global_matches(self, text)
    356         for lst in [keyword.kwlist,
    357                     builtin_mod.__dict__.keys(),
--> 358                     self.namespace.keys(),
    359                     self.global_namespace.keys()]:
    360             for word in lst:

AttributeError: 'IPCompleter' object has no attribute 'namespace'
Out[3]: ('foo', [])`

Carreau:
Thanks, sorry I was away for a few day I can get back on that and look at what's happening.

Carreau:
Ok, so let's have look at what's happening, you indeed get the error message `AttributeError: 'IPCompleter' object has no attribute 'namespace'`, so this piece of code assume there is a `namespace` attribute to `self` which definitively is not there. 

We have 2 choices here: 

  1. Handle the case where `namespace` does not exists if it make sens to not have `namespace`.
     - Likely with `hasattr(self, 'namespace')` or `getattr(self, 'namespace')`. I'll assume you know what these 2 are doing, or know how to find what they are doing feel free to ask if you need more info. 
  2. Make sure `namespace` is set on the class (figure out why it is not) 

Let's go with 2. 
After some digging which take som practice: 
  - `namespace` is [set (or not) here](https://github.com/ipython/ipython/blob/54cc9d25fdde97698cac9ef5659d2f4d24ea7dba/IPython/core/completer.py#L312-L316) on the IPcompleter superclass. 
  - You can see that it's accessed/updated [here](https://github.com/ipython/ipython/blob/54cc9d25fdde97698cac9ef5659d2f4d24ea7dba/IPython/core/completer.py#L333-L344) when calling `Completer.complete` and inparticular if it was not passed to `__init__` it uses `__main__`
  -  Though if you look at the superclass IPCompleter, the [overridden method `complete`](https://github.com/ipython/ipython/blob/54cc9d25fdde97698cac9ef5659d2f4d24ea7dba/IPython/core/completer.py#L1132-L1221) does not handle this case. 
  - `IPCompleter__init__` does not handle [this case either](https://github.com/ipython/ipython/blob/54cc9d25fdde97698cac9ef5659d2f4d24ea7dba/IPython/core/completer.py#L627-L628) (it could have), also in particular it calls `Completer.__init__` instead of `super(IPcompleter, self).__init__` because this is reallllly old code. 
 
My take on that (it's not the only solution) would be to add around [this position](https://github.com/ipython/ipython/blob/54cc9d25fdde97698cac9ef5659d2f4d24ea7dba/IPython/core/completer.py#L1166) a snippet like:

```
if self.use_main_ns:
            self.namespace = __main__.__dict__
```

A cleaner fix would be to change in `__init__` the `self.namespace` to `self._namespace` and make `self.namespace` a `@property` that return the right thing (maybe add a setter as well). 

Does that make some sens ?

Carreau:
Closing this as you fixed it in #9936. Thanks !

PR #9899, opened 2016-08-23
Backslash-terminated lines support in multiline strings

### LINKED ISSUE #9841: End of multiline string that contains a backslash-terminated line is not recognized
IPython 5.0/PTK1.0.5 does not recognize that

```
"""
foo\
bar
"""
```

is a full string and that it should execute the typed-in string after that point.

denistex:
This PR fixes #9841. Changed regexps a bit.

Comments are welcome.

### DIFF
--- a/IPython/utils/_tokenize_py2.py
+++ b/IPython/utils/_tokenize_py2.py
@@ -80,9 +80,9 @@ Single = r"[^'\\]*(?:\\.[^'\\]*)*'"
 # Tail end of " string.
 Double = r'[^"\\]*(?:\\.[^"\\]*)*"'
 # Tail end of ''' string.
-Single3 = r"[^'\\]*(?:(?:\\.|'(?!''))[^'\\]*)*'''"
+Single3 = r"[^'\\]*(?:(?:\\.|\\\n|'(?!''))[^'\\]*)*'''"
 # Tail end of """ string.
-Double3 = r'[^"\\]*(?:(?:\\.|"(?!""))[^"\\]*)*"""'
+Double3 = r'[^"\\]*(?:(?:\\.|\\\n|"(?!""))[^"\\]*)*"""'
 Triple = group("[uUbB]?[rR]?'''", '[uUbB]?[rR]?"""')
 # Single-line ' or " string.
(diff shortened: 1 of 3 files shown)


minrk:
Assigned @takluyver as the one most familiar with tokenizing stuff, but looks OK to me.

IPython/utils/_tokenize_py2.py
```diff
@@ -80,9 +80,9 @@ def maybe(*choices): return group(*choices) + '?'
 # Tail end of " string.
 Double = r'[^"\\]*(?:\\.[^"\\]*)*"'
 # Tail end of ''' string.
-Single3 = r"[^'\\]*(?:(?:\\.|'(?!''))[^'\\]*)*'''"
```
takluyver: These are copies of the standard library `tokenize` module, with a couple of changes we needed. I'd like to eventually get rid of them and rely on the regular `tokenize` module, so I'm very reluctant to accept any changes in them. If you really think that tokenize is doing something wrong (which is quite possible - it is a bit buggy), please file an issue on the [Python bug tracker](http://bugs.python.org/) and link to it here.

denis-it:
Hi @takluyver and thanks for your comment! Indeed I jumped the gun with this PR, sorry.

I made a bit more research into standard `tokenize` module and found that it is used in a strange way in IPython `TokenInputTransformer` class.

Standard `tokenize` module expects that `readline` callable passed into `tokenize` or `generate_tokens` returns data line by line, just like `readline` method of standard `file` object. In this case `\n` symbol can not exist in returned line, so it is not added to the regexps.

But `TokenInputTransformer.get_line` method (that is passed to `generate_tokens`) returns all data in one line (lines separated with `\n`) on first call and throws `TokenError` exception on further calls. I don't completely understand reasons of this tricky logic yet, but I think it should be changed and `tokenize` module should be used in right way.

So how do you think? Am I in a right way? I could continue research, add a couple of tests and rewrite `TokenInputTransformer`, if this is ok with you. :)

Thanks and sorry for delay.

takluyver:
The issue was fixed by #10157.

Issue #9000, opened 2015-11-18
Introspection of HasTraits classes


akhmerov:
Right now many IPython-related classes have traits and have very limited possibilities of introspection of the traits. `ipywidgets` is one notorious example, where initializing a widget is extremely hard without going through the tutorial, or calling `HasTraits` methods. I assume it is also related to matplotlib plans to use traits.

As one option to deal with the problem from the IPython side, `?` could also handle the traits and trait descriptions.

akhmerov:
pinging @minrk who suggested this as the most straightforward way for simplifying interactive work with `ipywidgets` for those without perfect recall.

Carreau:
I doubt we can do that for 5.0, so bumping to 6.0

PR #9905, opened 2016-08-24
Introspection of HasTraits classes

denistex:
Hi guys!

This is PR for #9000. I'm not sure if introspection output looks good enough (it's not colored and hard to parse visually), so check it please. How do you think?

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -372,6 +372,29 @@ def find_source_lines(obj):
 
     return lineno
 
+
+def get_class_traits_info(cls):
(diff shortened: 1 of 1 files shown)


takluyver:
I'm not sure that we should special case a specific object system (traitlets) in our introspection machinery. I'll try to think about better protocols for complex dynamic objects to give us more info about themselves.

denis-it:
Great idea!) So should I close this PR?

takluyver:
This is now quite old, and I don't think special casing traitlets is the right way forwards. So I'm going to close this. We can use #9000 for further discussion of how to tackle it.

PR #9916, opened 2016-08-27
Added '%matplotlib agg' option.

### LINKED ISSUE #9389: Turning off %matplotlib inline (feature request)
Apologies if I'm just missing it in the docs.

Currently, (IPython v4.2) it doesn't seem possible to undo calling `%matplotlib inline` if one doesn't want to revert to an interactive backend (one can only switch between different interactive backends). The reason why I believe it is useful to sometimes switch to a non-interactive backend is a combination of:
* Proliferation of using `pyplot` as the plotting interface used by libraries
* Lack of end-user control over which figures get plotted if the figures are created using `pyplot.figure`

If it was possible to have an IPython analog of `matplotlib.use('agg')`, the end user would always have control over which figures are plotted.

teoguso:
Behaves quite like inline, but it does not
display figures automatically.
Closes  #9389. 
cc @akhmerov @fperez 

Done at the jupyter sprint at EuroScipy 2016.

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -246,7 +246,7 @@ def find_gui_and_backend(gui=None, gui_select=None):
     Parameters
     ----------
     gui : str
-        Can be one of ('tk','gtk','wx','qt','qt4','inline').
+        Can be one of ('tk','gtk','wx','qt','qt4','inline','agg').
     gui_select : str
         Can be one of ('tk','gtk','wx','qt','qt4','inline').
         This is any gui already selected by the shell.
@@ -254,7 +254,7 @@ def find_gui_and_backend(gui=None, gui_select=None):
     Returns
     -------
     A tuple of (gui, backend) where backend is one of ('TkAgg','GTKAgg',
-    'WXAgg','Qt4Agg','module://ipykernel.pylab.backend_inline').
+    'WXAgg','Qt4Agg','module://ipykernel.pylab.backend_inline','agg').
     """
 
     import matplotlib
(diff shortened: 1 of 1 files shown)


IPython/core/pylabtools.py
```diff
@@ -261,6 +262,8 @@ def find_gui_and_backend(gui=None, gui_select=None):
     if gui and gui != 'auto':
         # select backend based on requested gui
         backend = backends[gui]
+        if gui == 'agg':
+            gui = None
```
minrk: This is needed because of an error in ipykernel, where it doesn't recognize `'agg'` as a GUI and it errors if it doesn't know what GUI to use. None indicates "no GUI", which is correct in principle, but there may be a better place to put that information.

We could modify something either here or in ipykernel to make it easier to specify "this backend needs no GUI integration", which would be true for nbagg, inline, agg.

akhmerov:
Thanks! Do you mind also updating the docstring of the magic so that it's know that this backend exists?

IPython/core/pylabtools.py
```diff
@@ -261,6 +262,8 @@ def find_gui_and_backend(gui=None, gui_select=None):
     if gui and gui != 'auto':
         # select backend based on requested gui
         backend = backends[gui]
+        if gui == 'agg':
```
willingc: Since `'agg'` is an addition to the function, it would be helpful to add it to the function's docstring parameters for `gui`.

> We could modify something either here or in ipykernel to make it easier to specify "this backend needs no GUI integration", which would be true for nbagg, inline, agg.

Good suggestion.

minrk:
Thanks, @teoguso!

PR #9925, opened 2016-09-01
Only make .tar.gz sdists when releasing

takluyver:
PEP 527, which was just accepted, says that there can only be one sdist per release. Tarballs are already the more common sdist format on PyPI, so let's go with that.


minrk:
👍 I think we have some "making a release" docs in one or two places that should be updated to match (@willingc @Carreau)

willingc:
Timely update @takluyver :smile:

PR #9929, opened 2016-09-02
Add shortcut to edit cell

takluyver:
I've integrated some code @carlsmith posted on the mailing list to open the current cell in a text editor to edit a larger chunk of code. This reduces the need for `%edit`, and has the benefit that edited code goes into the IPython history.

What should the shortcut be? Carl used **Ctrl-N**, but nothing about that says 'edit' to me. I've gone for **F2**, which is used in Excel to edit the selected cell, and I've enabled it for both emacs and vi keybindings. If we don't use an F-key (or something funky like Insert), we probably need separate emacs and vi shortcuts.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -62,6 +62,10 @@ def register_ipython_shortcuts(registry, shell):
                          filter=(HasFocus(DEFAULT_BUFFER)
                                 & EmacsInsertMode()))(newline_with_copy_margin)
 
+    registry.add_binding(Keys.F2,
+                         filter=HasFocus(DEFAULT_BUFFER)
+                        )(open_input_in_editor)
+
     if shell.display_completions == 'readlinelike':
         registry.add_binding(Keys.ControlI,
                              filter=(HasFocus(DEFAULT_BUFFER)
@@ -170,7 +174,9 @@ def newline_with_copy_margin(event):
         pos_diff = cursor_start_pos - cursor_end_pos
         b.cursor_right(count=pos_diff)
 
-
+def open_input_in_editor(event):
+    event.cli.current_buffer.tempfile_suffix = ".py"
+    event.cli.current_buffer.open_in_editor(event.cli)
 
 
 if sys.platform == 'win32':


Carreau:
No objection for F2. Is this 5.2 material ?

takluyver:
It's a simple enough addition that I'd be inclined to backport it to 5.x.

Carreau:
> It's a simple enough addition that I'd be inclined to backport it to 5.x.

Fair enough, let's add F2 for Py2.

melda7711:
hmmmmm

[ayam petarung](http://m8m8id.com/my/sabung-ayam)

anntzer:
Can vimmers also get "v" (in command mode) for this purpose?  It is the equivalent shortcut in vi-mode for bash & friends.

takluyver:
Sounds reasonable. Do you want to make a PR adding it? See `IPython.terminal.shortcuts`.

carlsmith:
Thanks for adding this @takluyver.

P.S. I only went with `n` because I normally use `ne` (the Nice Editor).

plumps:
Funny, in my case, it doesn't seem to respect ```$EDITOR```, while ```%edit``` does.

Carreau:
> Funny, in my case, it doesn't seem to respect $EDITOR, while %edit does.

Strange. Try updating Prompttoolkit. 
What version of IPython and PTK do you have ?

plumps:
IPython 5.2.2
prompt-toolkit 1.0.13

plumps:
Ah my ```EDITOR``` value of ```subl -w``` (which means sublime shall wait to return until files are closed) seams to be the culprit.

```EDITOR=vim``` works though

**Comment by Carreau** [comment]
> Ah my EDITOR value of subl -w (which means sublime shall wait to return until files are closed) seams to be the culprit.
> 
> EDITOR=vim works though

Ha ! Need to be fixed in prompt_toolkit. As things get passed manually to `subprocess.call` it's trying to find the executable `subl -w`

As a workaround create a  `sublw` file in your $PATH,make it executable and use it as your $EDITOR:

```
~/bin $ cat sublw
#!/bin/bash
subl -w $1
 ~/bin $ chmod u+x sublw
```

That should work.

Carreau:
I've submitted a patch to prompt toolkit.

plumps:
I'll wait for the patched prompt-toolkit

PR #9934, opened 2016-09-07
Show (Call) Signature of object first.

haobibo:
When user press Shift+Tab to call up a Tooltip to show the information of a function/method/callable object, I suggest IPython display (Call) Signature fist, so that users can know how to call the function quickly. Sometimes, when object with much information, the (Call) Signature will be hidden in Jupyter Notebook Tooltip.


minrk:
👍 @takluyver?

takluyver:
Seems reasonable.

PR #9939, opened 2016-09-09
Add more information about adding kernels to jupyter

pprzetacznik:
Add:
- information about current solution with virtualenv,
- add deprecated solution.

### DIFF
--- a/docs/source/install/kernel_install.rst
+++ b/docs/source/install/kernel_install.rst
@@ -53,3 +53,33 @@ For example, using conda environments:
 The ``--name`` value is used by Jupyter internally. These commands will overwrite
 any existing kernel with the same name. ``--display-name`` is what you see in
 the notebook menus.
+
+Using virtualenv you can add your new environments by linking to specific instance of your Python:
+
+.. sourcecode:: bash
+
+    python -m ipykernel install --prefix=/path/to/new_env --name python_new_env
+  
+Note that this command will create new configuration for kernel in one of it's prefered location (see ``jupyter --paths`` command for more details):
+
+* system-wide (e.g. /usr/local/share),
+* in Jupyter's env (sys.prefix/share),
+* per-user (~/.local/share or ~/Library/share)
+
+You can also create configuration in temprary location and install it in Jupyter (copy configuration files) with:
+
+.. sourcecode:: bash
+
+    ipython kernel install --prefix /tmp
+    jupyter kernelspec install /tmp/share/jupyter/kernels/python3
+
+**Deprecated:** You can also create configuration for new environment by new configuration's file in `~/.ipython/kernels/my_env/kernel.json` with following content::
+
+   {
+      "argv": ["~/path/to/env/bin/python", "-m", "IPython.kernel",
+         "-f", "{connection_file}"],
+      "display_name": "Python 3 (my new env)",
+      "language": "python"
+   }
+
+Please note that kernel detection in standard ``~/.ipython/`` location is supported as part of backward-compatibility and may not be provided in further versions.
--- a/docs/source/install/kernel_install.rst
+++ b/docs/source/install/kernel_install.rst
@@ -60,7 +60,7 @@ Using virtualenv or conda envs, you can make your IPython kernel in one env avai
 
     /path/to/kernel/env/bin/python -m ipykernel install --prefix=/path/to/jupyter/env --name 'python-my-env'
 
-Note that this command will create new configuration for kernel in one of it's prefered location (see ``jupyter --paths`` command for more details):
+Note that this command will create a new configuration for the kernel in one of the prefered location (see ``jupyter --paths`` command for more details):
 
 * system-wide (e.g. /usr/local/share),
 * in Jupyter's env (sys.prefix/share),
--- a/docs/source/install/kernel_install.rst
+++ b/docs/source/install/kernel_install.rst
@@ -66,7 +66,7 @@ Note that this command will create new configuration for kernel in one of it's p
 * in Jupyter's env (sys.prefix/share),
 * per-user (~/.local/share or ~/Library/share)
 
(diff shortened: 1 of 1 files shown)


docs/source/install/kernel_install.rst
```diff
@@ -53,3 +53,33 @@ For example, using conda environments:
...
+* per-user (~/.local/share or ~/Library/share)
+
+You can also create configuration in temporary location and install it in Jupyter (copy configuration files) with:
+
+.. sourcecode:: bash
+
+    ipython kernel install --prefix /tmp
+    jupyter kernelspec install /tmp/share/jupyter/kernels/python3
+
+**Deprecated:** You can also create configuration for new environment by new configuration's file in `~/.ipython/kernels/my_env/kernel.json` with following content::
```
minrk: Let's remove the discussion of the deprecated case. We don't want anyone to be putting kernels here, and adding it to the docs only makes it more likely.

docs/source/install/kernel_install.rst
```diff
@@ -53,3 +53,33 @@ For example, using conda environments:
...
+
+    python -m ipykernel install --prefix=/path/to/new_env --name python_new_env
+  
+Note that this command will create new configuration for kernel in one of it's prefered location (see ``jupyter --paths`` command for more details):
+
+* system-wide (e.g. /usr/local/share),
+* in Jupyter's env (sys.prefix/share),
+* per-user (~/.local/share or ~/Library/share)
+
+You can also create configuration in temporary location and install it in Jupyter (copy configuration files) with:
```
minrk: For this case, I think it only makes sense to do this if you want to edit the kernelspec by hand before installing it. I would add that to the discussion of this one, or remove the more complicated two-step case.

docs/source/install/kernel_install.rst
```diff
@@ -53,3 +53,33 @@ For example, using conda environments:
 The ``--name`` value is used by Jupyter internally. These commands will overwrite
 any existing kernel with the same name. ``--display-name`` is what you see in
 the notebook menus.
+
+Using virtualenv you can add your new environments by linking to specific instance of your Python:
+
+.. sourcecode:: bash
+
+    python -m ipykernel install --prefix=/path/to/new_env --name python_new_env
```
minrk: This is a complicated case and could use some explicit notes about which things point to which Python. Some clarification:

> Using virtualenv or conda envs, you can make your IPython kernel in one env available to Jupyter in a different env. To do so, run `ipykernel install` from the kernel's env, with `--prefix` pointing to the Jupyter env:

    /path/to/kernel/env/bin/python -m ipykernel install --prefix=/path/to/jupyter/env --name 'python-my-env'

pprzetacznik:
@minrk would you mind to take a look at this after updates?

docs/source/install/kernel_install.rst
```diff
@@ -53,3 +53,23 @@ For example, using conda environments:
...
+
+    /path/to/kernel/env/bin/python -m ipykernel install --prefix=/path/to/jupyter/env --name 'python-my-env'
+
+Note that this command will create new configuration for kernel in one of it's prefered location (see ``jupyter --paths`` command for more details):
+
+* system-wide (e.g. /usr/local/share),
+* in Jupyter's env (sys.prefix/share),
+* per-user (~/.local/share or ~/Library/share)
+
+In case where you want to edit generated kernelspec configuration before installing it you can do the same with two steps approach. You can create configuration in temporary location and install it in Jupyter (copy configuration files) with:
```
minrk: > If you want to edit the kernelspec before installing it, you can do so in two steps.
> First, ask IPython to write its spec to a temporary location:
>
>     ipython kernel install --prefix /tmp
>
> edit the files in /tmp/share/jupyter/kernels/python3 to your liking, then when you are ready, tell Jupyter to install it (this will copy the files into a place Jupyter will look):
>
>     jupyter kernelspec install /tmp/share/jupyter/kernels/python3

docs/source/install/kernel_install.rst
```diff
@@ -53,3 +53,23 @@ For example, using conda environments:
 The ``--name`` value is used by Jupyter internally. These commands will overwrite
 any existing kernel with the same name. ``--display-name`` is what you see in
 the notebook menus.
+
+Using virtualenv or conda envs, you can make your IPython kernel in one env available to Jupyter in a different env. To do so, run ipykernel install from the kernel's env, with --prefix pointing to the Jupyter env:
```
willingc: It may be helpful to add a sentence on "why" you would want to use a kernel from another environment.

minrk: @willingc any good idea here? For me, I might just link to the documentation for the env projects, rather than give our own case for why one would use envs.

docs/source/install/kernel_install.rst
```diff
@@ -53,3 +53,28 @@ For example, using conda environments:
...
+
+.. sourcecode:: bash
+
+    /path/to/kernel/env/bin/python -m ipykernel install --prefix=/path/to/jupyter/env --name 'python-my-env'
+
+Note that this command will create new configuration for kernel in one of it's prefered location (see ``jupyter --paths`` command for more details):
```
willingc: Small grammar corrections :-)

...will create a new configuration fo the kernel in one of the preferred locations...

docs/source/install/kernel_install.rst
willingc: @minrk Rereading this, I'm inclined to leave as proposed in the PR. I'm guessing that the "why" is because conda populates the conda env name to the notebook interface, but I'm not seeing a real clean way to illustrate that concisely and without more confusion. Let's leave as proposed in the PR. One small grammar change below and this is good to merge or merge and iterate.

@pprzetacznik Thanks for the PR and improving the documentation! :sunny:

minrk:
@pprzetacznik thanks!

Issue #9683, opened 2016-06-30
Navigate through history, ignore consecutive duplicates.


Carreau:
I think that would be nice,  not sure if it should be moved in PTK.

segevfiner:
I vaguely remember having had to press UP multiple times to skip duplicate/identical history entries in some IPython version, but I'm not seeing this behavior after testing recently. Searching around the code I found this: [jonathanslenders/python-prompt-toolkit/prompt_toolkit/buffer.py#L1127-L1130](https://github.com/jonathanslenders/python-prompt-toolkit/blob/4a27a142de525bb5a84edfb7df0bfa673ef9847e/prompt_toolkit/buffer.py#L1127-L1130) and there is obviously also the referenced PR #9943 which suggests that the version I saw this at were `IPython>=5.0,<5.2`.

Or did you mean something else entirely in this issue? 😃

Ritiek:
I use 5.1.0 but it works fine for me. Using same command consecutively shows up the command only once in the history when pressing up..

segevfiner:
@Ritiek I'm not entirely sure, but I think you will get duplicates only after re-opening IPython since the code in PR #9943 is only used during init. During an IPython session PTK prevents duplicates in the in-memory history but before PR #9943 IPython added duplicates to the history on history load at startup.

Ritiek:
@segevfiner Yeah, you're right. I can see the duplicates now.

srinivasreddy:
safely closed?

takluyver:
Yup, thanks.

segevfiner:
It seems thst IPython might ignore saving and loading consecutive duplicates. But when there are distant duplicates that are still the same they are not skipped over when doing an up/down arrows history search (When you type a partial command and then use the up/down arrows to search though history) or from ctrl-r. That means that you can do:
```python
In [1]: 123
Out [1]: 123

In [2]: 456
Out [2]: 456

In [3]: 123
Out [3]: 123
```

And you will find `123` twice in the history search.

Maybe that's what @Carreau meant or is that a different issue?

Carreau:
> Maybe that's what @Carreau meant or is that a different issue?

I honestly don't remember. I think having no consecutive duplicate is already good. If the question re-present itself, we can think about non-consecutive duplicate. I can see case where non-consecutive are useful.

takluyver:
Yeah, I think non-consecutive duplicates is fine, and this is now working as expected.

carlsmith:
[...]

PR #9943, opened 2016-09-11
Only track unique history of commands

JamshedVesuna:
Right now `last_cell` will always be an empty string. The intended use is to track the last command to only track a unique history.

Also removed some trailing whitespace 😄 

Related to #9683.


minrk:
Thanks!

takluyver:
Switched milestone to 5.2 because this is a simple bugfix.

PR #9946, opened 2016-09-13
Don't require win_unicode_console on Python 3.6

takluyver:
It looks like the changes @Drekin's win_unicode_console makes have essentially been integrated into Python 3.6, so it will use the Unicode APIs for Windows console access by default. See [PEP 528](https://www.python.org/dev/peps/pep-0528/).

So we shouldn't need to load WUC on Python 3.6, and it may even have some strange interactions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -218,7 +218,8 @@ extras_require.update({
     ':python_version == "2.7" or python_version == "3.3"': ['pathlib2'],
     ':sys_platform != "win32"': ['pexpect'],
     ':sys_platform == "darwin"': ['appnope'],
-    ':sys_platform == "win32"': ['colorama', 'win_unicode_console>=0.5'],
+    ':sys_platform == "win32"': ['colorama'],
+    ':sys_platform == "win32" and python_version < "3.6"': ['win_unicode_console>=0.5'],
     'test:python_version == "2.7"': ['mock'],
 })
 # FIXME: re-specify above platform dependencies for pip < 6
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -218,7 +218,8 @@ def run(self):
     ':python_version == "2.7" or python_version == "3.3"': ['pathlib2'],
     ':sys_platform != "win32"': ['pexpect'],
     ':sys_platform == "darwin"': ['appnope'],
-    ':sys_platform == "win32"': ['colorama', 'win_unicode_console>=0.5'],
+    ':sys_platform == "win32"': ['colorama'],
+    ':sys_platform == "win32" and python_version < "3.6"': ['win_unicode_console>=0.5'],
```
takluyver: This syntax should be valid as of setuptools 17.1, released June 2015. I'm OK with depending on that.

minrk:
Cool!

minrk:
Backport to 5.2?

takluyver:
I think it probably makes sense to, yeah. Though do we want to be a bit more conservative with the setuptools version required for 5.x?

setup.py
minrk: We already have similar ones above, I suspect we already require something similar.

takluyver: Specifically the `<` comparison - I believe `==` and `!=` were supported before other comparsions.

minrk: I'm guessing that since setuptools 17 is more than a year ago, I suspect that most people with setuptools < 17 have setuptools 0.6 + pip 1.5.6, and need to upgrade anyway. I doubt many folks are stuck on setuptools 16.

Issue #9940, opened 2016-09-11
Improve documentation for IPython.core.debugger and IPython.terminal.debugger


tillahoffmann:
The documentation for `IPython.core.debugger` states that it is deprecated but it doesn't provide any information about what should be used instead. A few questions:

* Was `IPython.core.debugger` deprecated in favour of `IPython.terminal.debugger`? 
* If so, what should be used in jupyter notebooks (`IPython.terminal.debugger.set_trace()` doesn't appear to work when used in a notebook)?

willingc:
@tillahoffmann Sorry this isn't clearer in the documentation. Pinging @takluyver and @carreau.

takluyver:
I don't think the module is deprecated - where do you see that? The `Tracer` class is; we should probably add a `set_trace` function to core.debugger as well as terminal.debugger.

tillahoffmann:
Yes, sorry, my description was not specific enough. Is this the correct way to set a breakpoint in the `add` function using the `core` debugger?

```python
from IPython.core.debugger import Pdb
debugger = Pdb()


def add(a, b):
    debugger.set_trace()
    return a + b
```

takluyver:
Not sure, I don't actually use a debugger much ;-). It looks correctish, though, so if it works, go for it!

willingc:
@tillahoffmann And if it works, please let us know and we can open an issue to improve debugger documentation. Thanks.

Carreau:
You should be able to do

```python
from IPython.terminal.debugger import set_trace
set_trace()
```

The difference between `IPython.terminal.debugger` and `IPython.core.debugger` are minimal, 
the former being slightly more aware of the terminal frontend.

takluyver:
I think this arose because `IPython.terminal.debugger` doesn't work in a notebook (or in IPython running as a Jupyter kernel generally). We should probably add `IPython.core.debugger.set_trace()` for such cases.

Carreau:
> I think this arose because IPython.terminal.debugger doesn't work in a notebook (or in IPython running as a Jupyter kernel generally). We should probably add IPython.core.debugger.set_trace() for such cases.

Yep, and we should likely figure out how to nest things a bit less.

tillahoffmann:
> I think this arose because IPython.terminal.debugger doesn't work in a notebook (or in IPython running as a Jupyter kernel generally). We should probably add IPython.core.debugger.set_trace() for such cases.

[...]

PR #9947, opened 2016-09-14
Add set_trace to core debugger.

tillahoffmann:
This PR adds a `set_trace` method to `IPython.core.debugger` to mirror the functionality of `IPython.terminal.debugger`. See #9940 for a discussion.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -629,5 +629,4 @@ def set_trace(frame=None):
 
     If frame is not specified, debugging starts from caller's frame.
     """
-    pdb = Pdb()
-    pdb.set_trace()
+    Pdb().set_trace(frame or sys._getframe().f_back)
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -623,12 +623,11 @@ class Pdb(OldPdb, object):
         do_w = do_where
 
 
-def set_trace(frame=None, condition=True):
-	"""
-	Start debugging from `frame` if `condition` is satisfied.
+def set_trace(frame=None):
+    """
+    Start debugging from `frame`.
 
-	If frame is not specified, debugging starts from caller's frame.
-	"""
-	if condition:
-		pdb = Pdb()
-		pdb.set_trace()
+    If frame is not specified, debugging starts from caller's frame.
+    """
+    pdb = Pdb()
+    pdb.set_trace()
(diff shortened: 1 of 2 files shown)


IPython/core/debugger.py
```diff
@@ -621,3 +621,14 @@ def do_where(self, arg):
                 self.print_stack_trace()
 
         do_w = do_where
+
+
+def set_trace(frame=None, condition=True):
```
takluyver: Let's stick to the same API as `pdb` and `IPython.terminal.debugger`, with `set_trace()` taking no parameters. We can discuss adding parameters as a separate step, and if we do we'll want to keep `core` and `terminal` in sync.

IPython/core/debugger.py
```diff
@@ -621,3 +621,14 @@ def do_where(self, arg):
...
+	"""
+	Start debugging from `frame` if `condition` is satisfied.
+
+	If frame is not specified, debugging starts from caller's frame.
+	"""
+	if condition:
+		pdb = Pdb()
```
tillahoffmann: These shouldn't be tabs.

IPython/core/debugger.py
```diff
@@ -621,3 +621,13 @@ def do_where(self, arg):
                 self.print_stack_trace()
 
         do_w = do_where
+
+
+def set_trace(frame=None):
```
takluyver: The other `set_trace()` functions also don't have a `frame` parameter, though the `set_trace()` *method* does. I'd either remove it from here, or also add it to the version in `IPython.terminal`. @Carreau, opinions?

Also, if we decide to keep the parameter, it should probably actually do something ;-)

tillahoffmann: Yes, it should indeed do something. Let me know if you'd prefer to remove the `frame` parameter or add it to `IPython.terminal` as well.

takluyver: I could go either way, but I'd like to know what @Carreau thinks when he's online in a couple more hours. :-)

Carreau: Happy to add it to the function, it should likely be passed to `Pdb.set_trace()` by adding 1 to it, and we should even likely start the debugging 1 frame up from the current one to be consistent with normal Pdb.

takluyver: Makes sense, though it takes a frame object, not an offset number. Here's the `set_trace` definition from pdb:

```python
def set_trace():
    Pdb().set_trace(sys._getframe().f_back)
```

Carreau (approved):
Seem good to me ! (trying new GitHub reviews things)

takluyver:
Looks good, thanks @tillahoffmann

PR #9950, opened 2016-09-15
enable quitting for -m IPython.terminal.debugger

mbyt:
When debugging with `python -m IPython.terminal.debugger myscript.py` and an exception occurs after hitting `c`, you cannot quit the debugger any more (see lower part of the description of #9942 for more detail).

The actual problem is that quit does not quit the debugging of the script but rather continues (affected [code line](https://github.com/ipython/ipython/blob/master/IPython/core/debugger.py#L285)). I guessed this is necessary for the notebook interface. However, it is the root cause for an never ending debugging session for this scenario.

This pull request re-enables quitting if debugging with `python -m IPython.terminal.debugger`.

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -86,5 +86,11 @@ def set_trace(frame=None):
 
 if __name__ == '__main__':
     import pdb
+    # IPython.core.debugger.Pdb.trace_dispatch shall not catch
+    # bdb.BdbQuit. When started through __main__ and an exeption
+    # happend after hitting "c", this is needed in order to
+    # be able to quit the debugging session (see #9950).
+    old_trace_dispatch = pdb.Pdb.trace_dispatch
     pdb.Pdb = TerminalPdb
+    pdb.Pdb.trace_dispatch = old_trace_dispatch
     pdb.main()
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -87,7 +87,7 @@ def set_trace(frame=None):
 if __name__ == '__main__':
     import pdb
     # IPython.core.debugger.Pdb.trace_dispatch shall not catch
-    # bdb.BdbQuit. When started through __main__ and an exeption
+    # bdb.BdbQuit. When started through __main__ and an exception
     # happend after hitting "c", this is needed in order to
     # be able to quit the debugging session (see #9950).
     old_trace_dispatch = pdb.Pdb.trace_dispatch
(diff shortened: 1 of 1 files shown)


mbyt:
Update: I looked into this again and found a less intrusive way to achieve the same thing. It just also monkey patched `TerminalPdb.trace_dispatch`.

This is much nicer / less destructive as my previous attempt. Therefore, I rebased and force pushed the branch (vanished the previous attempt...).

takluyver:
Thanks for looking into this. I suspect your solution makes sense, but I'm not really sure why we override `trace_dispatch` to catch BdbQuit in the first place. Any ideas?

Also pinging @fperez since the debugger code goes back before I joined the project.

mbyt:
> but I'm not really sure why we override `trace_dispatch` to catch `BdbQuit` in the first place. Any ideas?

No. Actually I first locally removed that one and then played a little bit with the debugger in the terminal. For me it did not change any other debugging experience / behaviour.

IPython/terminal/debugger.py
```diff
@@ -86,5 +86,11 @@ def set_trace(frame=None):
 
 if __name__ == '__main__':
     import pdb
+    # IPython.core.debugger.Pdb.trace_dispatch shall not catch
+    # bdb.BdbQuit. When started through __main__ and an exeption
```
fperez: Tiny typo: exeption -> exception.

fperez:
I very vaguely remember having a reason for catching that exception, but honestly I don't remember the specifics.  Obviously would have been good to document that more explicitly, but that ship sailed a good 10+ years ago.

The changes proposed here look like a minimally intrusive solution, +1. Thanks!

IPython/terminal/debugger.py
mbyt: thanks for pointing that out. Corrected in https://github.com/ipython/ipython/pull/9950/commits/3e812afa6e1f917695b517fc077ae54e5dab6c79

IPython/terminal/debugger.py
```diff
@@ -87,7 +87,7 @@ def set_trace(frame=None):
 if __name__ == '__main__':
     import pdb
     # IPython.core.debugger.Pdb.trace_dispatch shall not catch
-    # bdb.BdbQuit. When started through __main__ and an exeption
+    # bdb.BdbQuit. When started through __main__ and an exception
     # happend after hitting "c", this is needed in order to
```
willingc: Tiny typo in `happened` (missing second e)

mbyt: corrected in b817f5e

willingc: Thanks for fixing the typo :cookie: I will let @takluyver or @fperez merge. Thanks for the contribution.

Issue #9932, opened 2016-09-06
Debug magic fails to parse breakpoint files with colon in path


jlumpe:
The `%debug` magic has a `breakpoint` option formatted like `--breakpoint FILE:LINE`, where the file name and line number are separated by a colon. This command fails when a windows-style path containing a colon is used, e.g. beginning with `c:\`.

Example:

```
>>> mymodule.__file__
'c:\\users\\me\\git\\project\\mymodule.py'
>>> %debug --breakpoint {mymodule.__file__}:123 mymodule.foo()
ValueError...
```

The error occurs [here:](https://github.com/ipython/ipython/blob/5c9c918bc0a11057184ec143da13b68994f59666/IPython/core/magics/execution.py#L441)

```
    if breakpoint:
        (filename, bp_line) = breakpoint.split(':', 1)
        bp_line = int(bp_line)
```

where the breakpoint argument is incorrectly split at the first colon resulting in `bp_line` containing part of the file path. I think changing line 440 to

```
 (filename, bp_line) = breakpoint.rsplit(':', 1)
```

to split at the last colon should work just fine.

Carreau:
Thanks ! We'll have a look. 

@thesageinpilani you wanted something a bit more complicated; do you want to have a look at this one ?

It should not be that hard to fix, but figuring out where to do it might scratch your head a bit. 
I don't give you any hints on purpose, but feel free to ask here when you have any questions; I will be happy to describe things in more details.

thesageinpilani:
sure sir, i will try my best.

Carreau:
@thesageinpilani you can use the following to try:

```
In [1]: with open('mymodule.py', 'w') as f:
   ...:     f.write("""
   ...: def foo():
   ...:     1+1
   ...:     """)

In [2]: import mymodule

In [3]: %debug --breakpoint {mymodule.__file__}:3 mymodule.foo()
```

@jlumpe gave also a lot of indication on where and how to fix that. 

If you have any questions on how this things works, feel free to ask.

Carreau:
Closed by #9953 and backported.

PR #9953, opened 2016-09-16
corrected split with rsplit

ghost:
sir, @Carreau i changed the split with rsplit, i did it because with split only the first that is splitted from the string is taken, but with rsplit the last string is taken, which is what we want. please verify sir.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -437,7 +437,7 @@ python-profiler package from non-free.""")
 
     def _debug_exec(self, code, breakpoint):
         if breakpoint:
-            (filename, bp_line) = breakpoint.split(':', 1)
+            (filename, bp_line) = breakpoint.rsplit(':', 1)
             bp_line = int(bp_line)
         else:
             (filename, bp_line) = (None, None)


Carreau:
Great !
That seem to be a good fix !

I'll let some time to other dev to have a look. 

A few comments on how to make your contribution even better for subsequent fixes.

It's not easy here, but sometime we try to come up with a unit-test to make sure it does not break later.
One way we could do that here, is to refactor the following:

```
          if breakpoint:
             (filename, bp_line) = breakpoint.rsplit(':', 1)
              bp_line = int(bp_line)
          else:
              (filename, bp_line) = (None, None)
```

into a function, for example `_get_filename_and_line`, use it, and test it. 
Up to you if you like to try do that. 

Here it's a pretty some simple fix so there is no need to but you can also add an `rst` file into `docs/source/whatsnew/pr/` folder that will end-up in the what's new notes. 

 That was some tips to make your code even better and how we could have done it. 

A GitHub tip as well. 

When you make your message commit you can use the phrase `Fixes #9932` for example. 
That will automatically link your pull request with the target issue, and when your fix is merged the issue will be automatically closed !

And last improvement you can try next time when writing your commit message, 

Try to describe _why_ you do some changes instead of which changes you are doing. This is extremely useful when trying to understand code or why a bug appear in some line a few month or years later. In particular here with the diff it is pretty easy to see you changed `split` to `rsplit`, a better commit message might be:

> Fix breakpoint linenumber detection on Windows.
>  
> Change `split` to `rsplit` as windows paths can contain `:` when including drive letters. 

This should close  #9932, @jlumpe you reported the issue would that suit you ?

Thanks !

jlumpe:
Looks like this should fix it!

thesageinpilani:
@Carreau , sir i will remember this points and apply them when contributing.

Carreau:
> i will remember this points and apply them when contributing.

Thanks;  those are advice; it often takes years to completely internalize these, and apply them. 
You can see it mostly as guidelines. 

Thanks a lot for your contributions, it is really important for you to like contributing and enjoy participating.

thesageinpilani:
sir, i really do enjoy participating and contributing and i am learning a lot contributing, I am very grateful for your support and for your guidance all this time.

Carreau:
Thanks ! Merged.

PR #9965, opened 2016-09-22
Allow the debugger to work in code which calls __import__("__main__")

### LINKED ISSUE #9941: Using the IPython debugger in a unittest fails to continue
The new tab completion with IPython 5 is great! My congratulations, I really do not want to miss that again. Not even in the debugger --- where I discovered the following behaviour:

To my understanding IPython 5.1.0 allows debugging with tab completion by inserting the following line `from IPython.terminal.debugger import TerminalPdb ;TerminalPdb().set_trace()` (see also #9940). However, when doing so in the script below and simply hitting `c` for continue, it will exit with an attribute error:
```
  File "../snippets/test_ipython_debugger_in_unittest.py", line 21, in <module>
    if __name__ == '__main__':
  File "/usr/lib/python3.4/unittest/main.py", line 92, in __init__
    self.parseArgs(argv)
  File "/usr/lib/python3.4/unittest/main.py", line 139, in parseArgs
    self.createTests()
  File "/usr/lib/python3.4/unittest/main.py", line 146, in createTests
    self.module)
  File "/usr/lib/python3.4/unittest/loader.py", line 146, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.4/unittest/loader.py", line 146, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.4/unittest/loader.py", line 114, in loadTestsFromName
    parent, obj = obj, getattr(obj, part)
AttributeError: 'module' object has no attribute 'suite'
```

Using `import pdb; pdb.set_trace()` (commented line in code below) in contrast continues as expected (without exception).

----

```python
import unittest

class ATest(unittest.TestCase):
    def test_true(self):
        self.assertTrue(True)

def suite():
    return unittest.makeSuite(ATest, 'test_')

from IPython.terminal.debugger import TerminalPdb; TerminalPdb().set_trace()
#import pdb; pdb.set_trace()

if __name__ == '__main__':
    unittest.main(defaultTest='suite')
```

mbyt:
this basically resolves #9941 as suggested by @takluyver in https://github.com/ipython/ipython/issues/9941#issuecomment-248071079.

In summary, it is needed to run e.g. standard unittests within the debugger. For more details see #9941.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -227,10 +227,14 @@ class Pdb(OldPdb, object):
         self.shell = get_ipython()
 
         if self.shell is None:
+            save_main = sys.modules['__main__']
             # No IPython instance running, we must create one
             from IPython.terminal.interactiveshell import \
                 TerminalInteractiveShell
             self.shell = TerminalInteractiveShell.instance()
+            # needed by any code which calls __import__("__main__") after
+            # the debugger was entered. See also #9941.
+            sys.modules['__main__'] = save_main 
 
         if color_scheme is not None:
             warnings.warn(


takluyver:
Thanks, this looks good to me. I don't think it can easily be tested, because the test framework creates a shell instance, so this branch wouldn't be taken in the tests.

PR #9967, opened 2016-09-26
Doc update. Add reasoning behind PopupWidget removal.

FermiParadox:
Short explanation of [why it was removed](https://gitter.im/jupyter/jupyter?at=57e7075c7270539a6d843a4e).


willingc:
Thanks @user5061 for the update on the widget.

cc @SylvainCorlay for review

Carreau:
Inline Sylvain Response on gitter (which might not be permanent)

> Sylvain Corlay `@SylvainCorlay` Sep 24 16:08
> Popup widget was removed indeed. We are moving away from bootstrapjs widgets. For the good cause
bootstrap is a bad citizen of the browser. creates global variable etc..
With jupyterlab richer layouts should be enabled.


Thanks !

willingc:
Thanks @Carreau

PR #9970, opened 2016-09-30
Start refactoring handling of color.

Carreau:
This is one more step into getting rid of the old way of handling
colors in IPython, in particular here we move the scheme/style as a
configuration option of the parser instead of passing it around into the
formats functions.

Also remove our own usage of deprecated arguments.

That removes as well the IPython/utils/PyColorize.py standalone usage, drop a few "skip is not Py3" and a couple of Deprecation warning that were annoying me when testing locally.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -386,28 +386,18 @@ def _fixed_getinnerframes(etb, context=1, tb_offset=0):
 # can be recognized properly by ipython.el's py-traceback-line-re
 # (SyntaxErrors have to be treated specially because they have no traceback)
 
-_parser = PyColorize.Parser()
-
 
 def _format_traceback_lines(lnum, index, lines, Colors, lvals=None, scheme=None):
     numbers_width = INDENT_SIZE - 1
     res = []
     i = lnum - index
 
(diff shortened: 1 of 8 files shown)


IPython/core/ultratb.py
```diff
@@ -394,20 +394,16 @@ def _format_traceback_lines(lnum, index, lines, Colors, lvals=None, scheme=None)
...
-    if scheme is None:
-        ipinst = get_ipython()
-        if ipinst is not None:
-            scheme = ipinst.colors
-        else:
-            scheme = DEFAULT_SCHEME
-
+    global _parser
+    if scheme and (_parser.style != scheme):
+        _parser = _parser = PyColorize.Parser(style=scheme)
```
takluyver: Does this actually need to be a global? Instantiating a `Parser` looks like it's cheap, so we could just do it each time we need it.

Also, you're assigning to the same name twice here.

takluyver:
:+1:

IPython/core/ultratb.py
Carreau: Fixed.

takluyver:
Looks good to me, thanks.

PR #9971, opened 2016-09-30
Add the option expand the toc in the sidebar.

Carreau:
This add small `[+]` before each of the entry in the toctree in the
sidebar that can be expanded.

---

cc @willingc, i suppose we might want that on all the repos ? 

rtd

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -39,6 +39,10 @@ else:
     html_theme = "sphinx_rtd_theme"
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
+html_theme_options = {
+    'collapse_navigation': False,
+}
+
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
 # absolute, like shown here.


Carreau:
(might not work on current RTD, need to check first), and depending on the version of the theme deployed there, might need an extra option.

willingc:
@Carreau The deployed RTD theme uses this as one of its project-wide defaults. See https://github.com/snide/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/theme.conf#L10 for all theme defaults used on RTD.

Unless overridden in a jupyter project's `conf.py` for RTD., this should be the default on our docs. See https://jupyterhub.readthedocs.io.

It should be the default behavior when building locally or on RTD when its not listed in the options block.

Carreau:
Well it's strange because it definitively does not work on current IPython.

screen shot 2016-09-30 at 19 10 37

the jupyterhub does not behave like that either. Note that in above gif, when I expand the sections, the page does not change.

willingc:
Hmm... we may have found a bug in the theme. What browser are you using?

The initial activation of the + is a bit wonky. You need to click twice on the far left as in this GIF where the + should be.

hubplus





JupyterHub on Safari static image

screen shot 2016-09-30 at 19 44 42

Carreau:
I'm going to close that for now.

PR #9976, opened 2016-10-03
Let IPython.lib.guisupport detect terminal-integrated event loops

### LINKED ISSUE #9974: is_event_loop_running_qt4 seems broken in IPython 5
In IPython 4 I can test if the Qt event loop has already been started by IPython (if this is the case, IPython remains responsive while my Qt application runs at the same time). However, in IPython 5 this doesn't work anymore. The `is_event_loop_running_qt4` function returns False, even when I use the `%gui qt` magic or equivalent command line option.

The logic I use for detecting this is:

```
import sys
from PyQt4 import QtGui

def in_ipython():
    "Returns True if IPython is running, False for the regular Python."
    try:
        from IPython.core.getipython import get_ipython
    except ImportError:
        return False
    else:
        return get_ipython() is not None

def get_qapp(*args, **kwargs):
    "Gets the global Qt application object. Creates one if it doesn't exist."
    qApp = QtGui.QApplication.instance()
    if qApp:
        print("Returning existing QApplication")
        return qApp
    else:
        print("Creating new QApplication")
        return QtGui.QApplication(*args, **kwargs)

def start_qt_event_loop(qApp):
    """ Starts the eventloop if it's not yet running."""
    if in_ipython():
        try:
            from IPython.lib.guisupport import is_event_loop_running_qt4, start_event_loop_qt4
            if is_event_loop_running_qt4(qApp):
                print("IPython event loop already running. GUI integration possible.")
            else:
                print("Starting (non-interactive) IPython event loop")
                start_event_loop_qt4(qApp)
            return
        except Exception as ex:
            print("Falling back on non-interactive event loop: {}".format(ex))

    print("Starting (non-interactive) event loop")
    return qApp.exec_()

if __name__ == "__main__":
    qApp = get_qapp(sys.argv)
    start_qt_event_loop(qApp)
```

I run this program with the `--gui=qt` and `-i` options so that IPython starts the Qt event loop, starts my GUI, and remains interactive afterwards.

When running this program in IPython 4.2.1 it prints _"IPython event loop already running. GUI integration possible."_ as expected. In IPython 5.1.0 it prints _"Starting (non-interactive) IPython event loop"_ and the IPython prompt becomes unresponsive.

So, how can I detect if the event loop is already running? And is the `is_event_loop_running_qt4` function obsolete? Or will there be a `is_event_loop_running_qt5` for Qt5?

Thanks in advance and please let me know if you need more information.

takluyver:
Closes gh-9974

This is a bit more invasive than most backported changes, but it fixes a regression in IPython 5. My thinking:
- The `guisupport` APIs that worked before should continue working until/unless we deprecate them.
- There should be a common way to check if an event loop is already running in both the terminal and an IPython kernel.
- It should be possible to check for any event loop, not just Qt and Wx (which `guisupport` has checks for).

My plan is to make a public attribute `shell.active_eventloop`, which is either None or a string naming the event loop which IPython will run when waiting for input. E.g. `qt` or `gtk3`. (Todo: should we also expose the event loop object in cases where there is one? Not sure if anything useful can be done with it). This PR adds that attribute for terminal IPython; if we agree on it I'll make a separate PR for ipykernel.

The functions in guisupport then become a convenient shortcut for checking this, and we can decide whether to deprecate them in favour or something more uniform, or add similar convenience functions for other common event loops.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2891,6 +2891,8 @@ class InteractiveShell(SingletonConfigurable):
     # Things related to GUI support and pylab
     #-------------------------------------------------------------------------
 
+    active_eventloop = None
+
     def enable_gui(self, gui=None):
         raise NotImplementedError('Implement enable_gui in a subclass')
     
--- a/IPython/lib/guisupport.py
+++ b/IPython/lib/guisupport.py
@@ -57,16 +57,10 @@ so you don't have to depend on IPython.
 
 """
 
-#-----------------------------------------------------------------------------
-#  Copyright (C) 2008-2011  The IPython Development Team
-#
-#  Distributed under the terms of the BSD License.  The full license is in
-#  the file COPYING, distributed as part of this software.
-#-----------------------------------------------------------------------------
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
 
-#-----------------------------------------------------------------------------
-# Imports
-#-----------------------------------------------------------------------------
+from IPython.core.getipython import get_ipython
 
 #-----------------------------------------------------------------------------
 # wx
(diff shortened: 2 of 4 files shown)


Carreau:
> My plan is to make a public attribute shell.active_eventloop, which is either None or a string naming the event loop which IPython will run when waiting for input. E.g. qt or gtk3. (Todo: should we also expose the event loop object in cases where there is one? Not sure if anything useful can be done with it)

That's seem like an acceptable plan. DO you wan tto make it part of this PR ?

takluyver:
The bit about exposing the event loop object? I'd like a clear use case for adding it to public API, rather than just exposing things by default.

minrk (approved):

minrk:
Sounds good to me. I'm okay holding the instance-publishing bit until someone can come up with a good reason to use it (and saving it for 6.0).

PR #9984, opened 2016-10-05
Improve %timeit magic using average + standard deviation instead of best/worst

pablogsal:
As Victor Stinner noted in its [perf module](http://perf.readthedocs.io/en/latest/) returning the minimum value (best run) and the maximum value (worst run) is a bad strategy because it represents a bad statistical representation (see "Statistically Rigorous
Java Performance Evaluation" by Georges et.al. 2007).

 Far from being perfect this PR implement the following changes:
- Report average + standard deviation instead of min and max.
- Increase default runs from 3 to 7 (actually it checks if the default runs in the timeit module is less than 7 and if so, it uses 7 instead).

This change has been inspired in a similar one in the [Pypy project](https://bitbucket.org/pypy/pypy) as can be checked in [this commit](https://bitbucket.org/pypy/pypy/commits/fb6bb835369e).

The tests suite is corrected as the proper docstrings but no new tests regarding this change are currently implemented.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -79,21 +79,15 @@ class TimeitResult(object):
         self.compile_time = compile_time
         self._precision = precision
         self.timings = [ dt / self.loops for dt in all_runs]
-        self._average = None
-        self._stdev = None
 
     @property
     def average(self):
-        if self._average is None:
-            self._average = math.fsum(self.timings) / len(self.timings)
-        return self._average
+        return math.fsum(self.timings) / len(self.timings)
 
     @property
     def stdev(self):
-        if self._stdev is None:
-            mean = self.average
-            self._stdev = (math.fsum([(x - mean) ** 2 for x in self.timings]) / len(self.timings)) ** 0.5
-        return self._stdev
+        mean = self.average
+        return self._stdev = (math.fsum([(x - mean) ** 2 for x in self.timings]) / len(self.timings)) ** 0.5
 
     def __str__(self):
         if self.loops == 1:  # No s at "loops" if only one loop
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -72,27 +72,45 @@ class TimeitResult(object):
 
     """
 
-    def __init__(self, loops, repeat, average, stdev, all_runs, compile_time, precision):
+    def __init__(self, loops, repeat, all_runs, compile_time, precision):
         self.loops = loops
         self.repeat = repeat
-        self.average = average
-        self.stdev = stdev
         self.all_runs = all_runs
         self.compile_time = compile_time
         self._precision = precision
+        self.timings = [ dt / self.loops for dt in all_runs]
+        self._average = None
+        self._stdev = None
+
+    @property
+    def average(self):
+        if self._average is None:
+            self._average = math.fsum(self.timings) / len(self.timings)
+        return self._average
+
+    @property
+    def stdev(self):
+        if self._stdev is None:
+            mean = self.average
+            self._stdev = (math.fsum([(x - mean) ** 2 for x in self.timings]) / len(self.timings)) ** 0.5
+        return self._stdev
+
+    def __str__(self):
+        if self.loops == 1:  # No s at "loops" if only one loop
+            return (u"%s loop, average of %d: %s +- %s per loop (using standard deviation)"
+                       % (self.loops, self.repeat,
+                          _format_time(self.average, self._precision),
+                          _format_time(self.stdev, self._precision)))
+        else:
+            return (u"%s loops, average of %d: %s +- %s per loop (using standard deviation)"
+                       % (self.loops, self.repeat,
+                          _format_time(self.average, self._precision),
+                          _format_time(self.stdev, self._precision)))
 
     def _repr_pretty_(self, p , cycle):
-        if self.loops == 1:  # No s at "loops" if only one loop
-             unic = (u"%s loop, average of %d: %s +- %s per loop (using standard deviation)"
-                        % (self.loops, self.repeat,
-                           _format_time(self.average, self._precision),
-                           _format_time(self.stdev, self._precision)))
-        else:
-            unic = (u"%s loops, average of %d: %s +- %s per loop (using standard deviation)"
-                        % (self.loops, self.repeat,
-                           _format_time(self.average, self._precision),
-                           _format_time(self.stdev, self._precision)))
-        p.text(u'<TimeitResult : '+unic+u'>')
+       unic = self.__str__()
+       p.text(u'<TimeitResult : '+unic+u'>')
+
 
 
 class TimeitTemplateFiller(ast.NodeTransformer):
@@ -1051,17 +1069,7 @@ python-profiler package from non-free.""")
                     break
 
         all_runs = timer.repeat(repeat, number)
-        timings = [ dt / number for dt in all_runs]
-
-        def _avg(numbers):
-            return math.fsum(numbers) / len(numbers)
-
-        def _stdev(numbers):
-            mean = _avg(numbers)
-            return (math.fsum([(x - mean) ** 2 for x in numbers]) / len(numbers)) ** 0.5
-
-        average = _avg(timings)
-        stdev   = _stdev(timings)
+        timeit_result = TimeitResult(number, repeat, all_runs, tc, precision)
 
         if not quiet :
             # Check best timing is greater than zero to avoid a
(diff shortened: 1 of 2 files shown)


pablogsal:
By the way, I have prepared a C extension module using cython for the average and standard deviation functions to be more performant. If you think is a good idea I can add it to the mix but maybe it is not worth it.

takluyver:
Thanks, I know little about performance testing, but I suspect this makes sense.

What's the significance of the number 7?

IPython is pure Python, and we wouldn't add C extensions to the mix without a good reason. I think the performance of the Python average and stdev functions will be fine.

IPython/core/tests/test_interactiveshell.py
```diff
@@ -171,9 +171,9 @@ def test_can_pickle(self):
                      "    def __init__(self,x=[]):\n"
                      "        list.__init__(self,x)"))
         ip.run_cell("w=Mylist([1,2,3])")
-
```
takluyver: Could we ask you to redo this without removing all the trailing whitespace from the files? Sorry, but it messes up git blame and can cause unnecessary merge conflicts.

If you're using Atom, which strips trailing whitespace by default, you can disable that in the settings of the 'whitespace' package.

IPython/core/magics/execution.py
```diff
@@ -71,23 +72,27 @@ class TimeitResult(object):
...
     """
 
-    def __init__(self, loops, repeat, best, worst, all_runs, compile_time, precision):
+    def __init__(self, loops, repeat, average, stdev, all_runs, compile_time, precision):
         self.loops = loops
         self.repeat = repeat
-        self.best = best
-        self.worst = worst
+        self.average = average
+        self.stdev = stdev
```
takluyver: I wonder if it would make more sense for the `TimeitResult` object to hold all of the recorded timings, and the calculation of summary statistics like average and stdev to be part of this object.

takluyver (changes requested):

pablogsal:
The problem is that the standard deviation is a measure on how each number deviates from the mean value and in order to have proper relevance it needs more than three measures. The number 7 fits well when testing how much measures you need to start having a standard deviation with proper relevance ( Although is the same number in the pypy commit I have tested this myself).

IPython/core/tests/test_interactiveshell.py
pablogsal: Yeah, sorry for the whitespace. I'm using vim and vim gonna do vim stuff.I will clean the PR.

IPython/core/magics/execution.py
pablogsal: I was thinking about that approach but I think that because those numbers are not going to change and because the printed message to stdout is much more informative with the proper statistics I think that the current way is good enough. in cases you think something would suit better I more than happy to adapt the code.

pablogsal: On the other hand because the printed message to stdout ia the same that coils be used to refactor some code and remove redundancy which I think is more than enough reason to move the calculation to the __init__ method.

takluyver: Yeah, I wasn't suggesting that the printed representation change. But you can also get the `TimeitResult` object back, and I can imagine wanting to calculate other statistics from the repeated measurements, so it would be useful if it stored all of them.

pablogsal: The "problem"? with that is that if the statistic functions are methods, then we will be calculating the standard deviation and the average **unless** we cached those at least. This is because we need those values to print its __str__ or __repr__ representation. Should I use this approach or the calculation of this two values should be done in the __init__ method?

pablogsal: Check if the current status in 395b333 is ok.

Carreau:
I'm fine with 7, computers get faster each days (in average).

IPython/core/magics/execution.py
```diff
@@ -71,23 +72,45 @@ class TimeitResult(object):
 
     """
 
-    def __init__(self, loops, repeat, best, worst, all_runs, compile_time, precision):
+    def __init__(self, loops, repeat, all_runs, compile_time, precision):
```
Carreau: The timeit magic can return the result, removing `best` and `worst` attributes will break some code that rely on this. I doubt there are much code using this though.

IPython/core/magics/execution.py
takluyver: I would do them as simple properties or methods. Caching the values feels like very premature optimisation. This code isn't performance critical, and it's only dealing with a handful of numbers, so even with a Python loop, it will be totally instantaneous from a user perspective.

pablogsal:
By the way, I think that with the print message refactored into the TimeitResult object it would be easier to add more fancy stuff to the mix in the future and in adition it simplifies the codebase a lot too!

IPython/core/tests/test_interactiveshell.py
pablogsal: Done

IPython/core/magics/execution.py
```diff
@@ -71,23 +72,45 @@ class TimeitResult(object):
...
+
+    @property
+    def stdev(self):
+        if self._stdev is None:
+            mean = self.average
+            self._stdev = (math.fsum([(x - mean) ** 2 for x in self.timings]) / len(self.timings)) ** 0.5
+        return self._stdev
+
+    def __str__(self):
+        if self.loops == 1:  # No s at "loops" if only one loop
```
Carreau: use a `plural` variable which is `s` if `self.loops  != 1` and one more string formatting ?

IPython/core/magics/execution.py
```diff
@@ -71,23 +72,45 @@ class TimeitResult(object):
...
 
     def _repr_pretty_(self, p , cycle):
-         if self.loops == 1:  # No s at "loops" if only one loop
-             unic =  u"%d loop, best of %d: %s per loop" % (self.loops, self.repeat,
-                                            _format_time(self.best, self._precision))
-         else:
-             unic =  u"%d loops, best of %d: %s per loop" % (self.loops, self.repeat,
-                                            _format_time(self.best, self._precision))
-         p.text(u'<TimeitResult : '+unic+u'>')
+       unic = self.__str__()
```
Carreau: Isn't `self.__str__()` supposed to be `str(self)` ?

IPython/core/magics/execution.py
```diff
@@ -71,23 +72,45 @@ class TimeitResult(object):
...
+       p.text(u'<TimeitResult : '+unic+u'>')
```
Carreau: Note that if we decide to not backport this you can avoid the u prefix as master is already Python 3 only. But you can leave it for now.

IPython/core/magics/execution.py
```diff
@@ -1036,43 +1060,29 @@ def timeit(self, line='', cell=None):
...
+        timeit_result = TimeitResult(number, repeat, all_runs, tc, precision)
 
         if not quiet :
             # Check best timing is greater than zero to avoid a
             # ZeroDivisionError.
             # In cases where the slowest timing is lesser than a micosecond
             # we assume that it does not really matter if the fastest
             # timing is 4 times faster than the slowest timing or not.
-            if worst > 4 * best and best > 0 and worst > 1e-6:
-                print("The slowest run took %0.2f times longer than the "
```
Carreau: I think I would leave this warning message. It is relatively recent, and was requested because users were regularly getting cached results.

IPython/core/magics/execution.py
pablogsal: In e7913fa is done with properties and caching. Should I remove the caching?

Carreau:
Thanks ! That will be great !

IPython/core/magics/execution.py
pablogsal: In 8504424 the caching is removed leaving those as properties.

IPython/core/magics/execution.py
pablogsal: Restored in 95ae1d8

IPython/core/magics/execution.py
pablogsal: By the way (and forgive me for the comment) but I have bootstrapped this performance measure. :)

Carreau (approved):

Carreau:
I'm +1. 

Any preference if it should be backported to 5.x (which supports python 2) of if it should be a IPython 6 / Python 3 only feature ?

willingc:
One small question: Does changing timeit have any impact on folks that may have been using it previously?

Carreau:
> One small question: Does changing timeit have any impact on folks that may have been using it previously?

It does not change the API, (or at least not in a breaking manner). It will just change what's printed to be more realistic. 

Any advance usage of timeit should likely use the timeit module directly anyway.

pablogsal:
The retuned object is a superset of the last one so nothing would be breaking for users that were relaying on that. What is different is the returned message that now has average and standard deviation instead of min value and the fact that samples 7 times instead of 3 so users will notice a longer time in the measure process with the defaults.

pablogsal:
> Any advance usage of timeit should likely use the timeit module directly anyway.

Or as is discussed on [here](http://perf.readthedocs.io/en/latest/) and [here](https://haypo.github.io/journey-to-stable-benchmark-average.html) not use it at all for performance measurement.

takluyver:
I'd say don't backport it, but I don't feel strongly.

pablogsal:
There is something else I should change (I ask this because the PR says "changes requested")?

takluyver:
I think it looks good - sorry, I requested changes before and forgot that I'd done that. I'm not 100% enamoured of Github's new review system.

PR #9985, opened 2016-10-05
Only suggest a named argument once

tmr232:
Added a check to the `python_func_kw_matches` completer to not show named arguments after they've been used in the function call.

So, for example, when writing:

``` python
def my_func(x=1, y=2, z=3):
    pass

my_func(x=1, y=2
```

The completer would suggest `z=`, but not `x=` or `y=`.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -886,8 +886,7 @@ class IPCompleter(Completer):
         # parenthesis before the cursor
         # e.g. for "foo (1+bar(x), pa<cursor>,a=1)", the candidate is "foo"
         tokens = regexp.findall(self.text_until_cursor)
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks. This doesn't seem to be working quite right, however - this test is failing:

```python
ip.ex('def myfunc(a=1,b=2): return a+b')
s, matches = c.complete(None, 'myfunc(1,b')
nt.assert_in('b=', matches)
```

tmr232:
I still work mostly in 2.7. Seems like the issue was the 2.7 `itertools.izip` vs the 3.x `zip`.

takluyver:
Thanks, seems to be working now :-)

PR #9987, opened 2016-10-06
Pass parent to child for configuration to propagate

Carreau:
Some cleanup and API preparation for the color refactoring. 

---

This thing is a mess and is going to be tough to refactor...

### DIFF
--- a/IPython/utils/PyColorize.py
+++ b/IPython/utils/PyColorize.py
@@ -44,12 +44,7 @@ import sys
 import token
 import tokenize
 
-try:
-    generate_tokens = tokenize.generate_tokens
-except AttributeError:
-    # Python 3. Note that we use the undocumented _tokenize because it expects
-    # strings, not bytes. See also Python issue #9969.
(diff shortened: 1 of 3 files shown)


IPython/utils/PyColorize.py
```diff
@@ -44,12 +44,7 @@
 import token
 import tokenize
 
-try:
-    generate_tokens = tokenize.generate_tokens
-except AttributeError:
-    # Python 3. Note that we use the undocumented _tokenize because it expects
-    # strings, not bytes. See also Python issue #9969.
-    generate_tokens = tokenize._tokenize
+generate_tokens = tokenize.tokenize
```
minrk: This was `_tokenize`, not `tokenize` for str/bytes distinctions. If we want to use `.tokenize`, we need to pass bytes, or wrap it in an encoder.

Carreau:
Travis is happy. `tokenize.generate_tokens` is available on Python 3 and does not emit a deprecation warning.

PR #9992, opened 2016-10-10
Update IPython.lib.demo

lejenome:
This is a set of three patches to fix demo.py and add more useful features so it can be used as a slide tool:
- patch1: fix broken code
- patch2: use pygments to highlight python code and optionally rst comments. and make demo.py usable outside ipython
- patch3: add __main__ entry to demo.py so it can be used directly from the command line.

Support pygments is very useful because it supports more colorschema styles and offers more lexers (including the used rst lexer). Highlighting code was moved into it's own `highlight()` function so it can be easly customized on subcalsses (e.g: support markdown comments, add `>>>` before python code)

Screenshot of the same demo file:
- original IPython.lib.demo inside qtconsole

2016-10-10-131159_1366x768_scrot
- the new IPython.lib.demo inside qtconsole with rst enabled, pygments formatter16m and colorful style.

2016-10-10-131402_1366x768_scrot
- the new IPython.lib.demo run as command line directly on the shell with rst enabled, and with the default formatter (terminal) and style (default)

``` sh
python -m IPython.lib.demo --rst demo.py
```

2016-10-10-131853_1366x768_scrot
- the new IPython.lib.demo run as command line directly on the shell with rst enabled, pygments formatter16m and colorful style.

``` sh
python -m IPython.lib.demo --rst demo.py --formatter terminal16m --style colorful
```

2016-10-10-131915_1366x768_scrot

### DIFF
--- a/IPython/lib/demo.py
+++ b/IPython/lib/demo.py
@@ -627,3 +627,43 @@ class ClearDemo(ClearMixin,Demo):
 
 class ClearIPDemo(ClearMixin,IPythonDemo):
     pass
+
+
+def slide(file_path, noclear=False, format_rst=True, formatter="terminal",
+          style="native", auto_all=False, delimiter='...'):
+    if noclear:
+        demo_class = Demo
+    else:
+        demo_class = ClearDemo
+    demo = demo_class(file_path, format_rst=format_rst, formatter=formatter,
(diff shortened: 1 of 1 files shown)


minrk:
Great, thanks!

PR #9994, opened 2016-10-11
Document automagic behaviour on unknown argument

hellerve:
Hey,

the `%automagic` command negates the previous setting if it is called with an argument that it does not understand. This behaviour, however, was previously undocumented. This PR addresses that.

Cheers

### DIFF
--- a/IPython/core/magics/auto.py
+++ b/IPython/core/magics/auto.py
@@ -44,6 +44,8 @@ class AutoMagics(Magics):
 
          - off, 0, False: to deactivate.
 
+         - anything else: negate previous setting.
+
         Note that magic functions have lowest priority, so if there's a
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, but it does say just above this "Without argumentsl toggles on/off". The behaviour with unrecognised arguments is coincidental, so I don't think it makes sense to document it.

hellerve:
Understandable. Still, does it make sense to leave any behavior just lying there undocumented?

Anyway, as this is relatively minor, I don't want to go around bikeshedding on this. Thanks for reviewing this PR!

takluyver:
I'd look at this as: documentation is there for people wondering 'how do I...' and 'what things can I do?' So it only needs to describe what's useful. If you're wondering 'what happens if I...', then you can either try it, or look at the code :-)

PR #9998, opened 2016-10-12
Mnt remove unneeded pylab imports

tacaswell:


minrk:
Thanks!

tacaswell:
How far back can we talk you into backporting this?

takluyver:
It's slated for inclusion in the 5.x series. We're not intending to do any more 4.x releases, so that's as far back as makes sense.

Issue #9999, opened 2016-10-12
Add a metadata value called "expanded" to the JSON display class


aleury:
I would like to extend the JSON display object class to accept a metadata parameter called `expanded`. The purpose is to be able to control the expansion of a JSON display component used in the [nteract](https://github.com/nteract/nteract) project. Please see https://github.com/nteract/nteract/issues/962 for an example.

rgbkrk:
Example:

![nteract JSON support](https://cloud.githubusercontent.com/assets/1060558/19253579/2198b7b6-8f1a-11e6-89b0-85de1390ddea.png)

takluyver:
Closed by #10000 !

PR #10000, opened 2016-10-12
feat(JSON): Add expanded metadata to JSON display class

aleury:
Addresses https://github.com/ipython/ipython/issues/9999

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -190,7 +190,7 @@ def display_pretty(*objs, **kwargs):
 
 def display_html(*objs, **kwargs):
     """Display the HTML representation of an object.
-    
+
     Note: If raw=False and the object does not have a HTML
     representation, no HTML will be shown.
 
@@ -513,13 +513,35 @@ class SVG(DisplayObject):
 
 class JSON(DisplayObject):
     """JSON expects a JSON-able dict or list
-    
+
     not an already-serialized JSON string.
-    
+
     Scalar types (None, number, string) are not allowed, only dict or list containers.
     """
     # wrap data in a property, which warns about passing already-serialized JSON
     _data = None
+    def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=None):
+        """Create a JSON display object given raw data.
+
+        Parameters
+        ----------
+        data : dict or list
+            JSON data to display. Not an already-serialized JSON string.
+            Scalar types (None, number, string) are not allowed, only dict
+            or list containers.
+        url : unicode
+            A URL to download the data from.
+        filename : unicode
+            Path to a local file to load the data from.
+        expanded : boolean
+            Metadata to control whether a JSON display component is expanded.
+        metadata: dict
+            Specify extra metadata to attach to the json display object.
+        """
+        self.expanded = expanded
+        self.metadata = metadata
+        super(JSON, self).__init__(data=data, url=url, filename=filename)
+
     def _check_data(self):
         if self.data is not None and not isinstance(self.data, (dict, list)):
             raise TypeError("%s expects JSONable dict or list, not %r" % (self.__class__.__name__, self.data))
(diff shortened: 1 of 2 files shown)


IPython/core/display.py
```diff
@@ -512,31 +512,54 @@ def _repr_svg_(self):
...
-    """JSON expects a JSON-able dict or list
-    
-    not an already-serialized JSON string.
-    
-    Scalar types (None, number, string) are not allowed, only dict or list containers.
-    """
+
     # wrap data in a property, which warns about passing already-serialized JSON
     _data = None
+    def __init__(self, data=None, url=None, filename=None, expanded=True, metadata=None):
```
rgbkrk: I'm tempted to say that we should default `expanded` to `False` so the amount of data isn't overwhelming. I could go either way (especially since I was the one asking about it).

IPython/core/display.py
```diff
@@ -622,7 +645,7 @@ def _pngxy(data):
 def _jpegxy(data):
     """read the (width, height) from a JPEG header"""
     # adapted from http://www.64lines.com/jpeg-width-height
-    
+
```
rgbkrk: Some of the IPython devs won't be happy about the whitespace changes here (not me), you may want to strip these out.

rgbkrk:
🎉 🎈 Congratulations 🎈 🎉 

   You made the 10000th PR!

🎈🎉 🎈🎉 🎈🎉 🎈🎉 🎈

rgbkrk:
If this ends up being accepted, we'll want to document in the section about mime bundle metadata in `jupyter_client` as well.

/cc @willingc

rgbkrk (approved):

IPython/core/display.py
aleury: Yeah I'm sorry if the whitespace diffs are a bit annoying. I believe it was done automatically by my editor on save (Atom). I can try to correct it if desired.

IPython/core/display.py
aleury: Either way sounds good to me! Just let me know. 👍

IPython/core/display.py
```diff
@@ -512,31 +512,54 @@ def _repr_svg_(self):
 
 
 class JSON(DisplayObject):
-    """JSON expects a JSON-able dict or list
-    
-    not an already-serialized JSON string.
-    
-    Scalar types (None, number, string) are not allowed, only dict or list containers.
-    """
+
```
minrk: Why was the docstring removed? Can we get it back?

IPython/core/display.py
minrk: You can fix atom by going to: packages > whitespace > disable.

I'm mostly done objecting to PRs that do this, but I would always prefer that they not happen. I'm really disappointed in Atom's user-hostile aggressive defaults.

minrk:
👍 with a minor docstring change.

IPython/core/display.py
willingc: @minrk Likely an oversight when __init__ was added and the docstring contents were moved to its docstring. I agree it would be good to have the class docstring as well.

:tada: for writing docstrings @aleury

willingc: @minrk your clone was 1 minute faster than my typing ;-)

IPython/core/display.py
rgbkrk: Let's go with `expanded=False`, unless anyone has objections.

IPython/core/display.py
aleury: Ah! Thanks so much. I agree it's super annoying. I attempted to find a trick to undo it with git but couldn't find anything that worked.

IPython/core/display.py
aleury: Sure, I can do that! My thinking was to make it similar to the `Image` class, which is why I removed the class doc string. I'll add it back though. Thanks!

minrk:
Thanks! 🎉 10000 🎉

aleury:
So happy to have the opportunity to contribute the 10,000th PR to IPython! Thanks for building such a great project! 😄

willingc:
🎈 🍰 🎉 A party for @aleury and the entire IPython team. 🎂 🎆 🍾

PR #10009, opened 2016-10-15
Do not capture BdbQuit.

### LINKED ISSUE #10006: impossible to exit debugger traces
I like to set traces, and from what I understand from the docs, The Right Way is to use `TerminalPdb`.
However, I can not find out how to quit the `TerminalPdb` debugger---maybe I'm just daft.
Say I set a trace in a long loop:

``` python
from IPython.terminal.debugger import TerminalPdb
pdb = TerminalPdb()

while True:
    pdb.set_trace()
```

The only way to exit this program that I found is to kill it.

My questions are:
- Is TerminalPdb the best way to set traces?
- Is this expected behavior?
- How do I quit?

Thanks!

(I'm on IPython 5.1.0)

Carreau:
If one want to exit Pdb cleanly, it is possible to just use `continue`
(or `c`) that will keep up until a potential next invocation.

Quit is _meant_ to raise an error.

Closes #10006

---

Thanks @amueller for the report. Would that suit you ?

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -283,12 +283,6 @@ class Pdb(OldPdb, object):
         self.color_scheme_table.set_active_scheme(scheme)
         self.parser.style = scheme
 
-    def trace_dispatch(self, frame, event, arg):
-        try:
-            return super(Pdb, self).trace_dispatch(frame, event, arg)
-        except bdb.BdbQuit:
-            pass
-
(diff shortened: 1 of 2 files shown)


amueller:
That's raised by calling ``quit()``? And/or by pressing "Ctrl+D"? Then yes.

amueller:
thanks :)

PR #10012, opened 2016-10-17
use `import_module` instead of `__import__` 

### LINKED ISSUE #10008: use import_module instead of __import__
Since Python 3.3 it is [_requested_ to use importlib.import_module instead of `__import__`](https://docs.python.org/3/library/importlib.html#importlib.__import__).

The reason is not really given, but I guess there are reasons.

drgarcia1986:
FIX #10008

I think it's a pythonic way to dynamic import a module.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -26,6 +26,7 @@ import sys
 import unicodedata
 import string
 import warnings
+from importlib import import_module
 
 from traitlets.config.configurable import Configurable
 from IPython.core.error import TryNext
@@ -481,7 +482,7 @@ def _safe_isinstance(obj, module, class_name):
     """Checks if obj is an instance of module.class_name if loaded
     """
     return (module in sys.modules and
-            isinstance(obj, getattr(__import__(module), class_name)))
+            isinstance(obj, getattr(import_module(module), class_name)))
 
 
 def back_unicode_name_matches(text):
--- a/IPython/core/completerlib.py
+++ b/IPython/core/completerlib.py
@@ -22,6 +22,7 @@ import inspect
 import os
 import re
 import sys
+from importlib import import_module
 
 try:
     # Python >= 3.3
@@ -155,7 +156,7 @@ def is_importable(module, attr, only_modules):
 
 def try_import(mod, only_modules=False):
     try:
-        m = __import__(mod)
+        m = import_module(mod)
     except:
         return []
     mods = mod.split('.')
--- a/IPython/core/completerlib.py
+++ b/IPython/core/completerlib.py
@@ -159,9 +159,6 @@ def try_import(mod, only_modules=False):
         m = import_module(mod)
     except:
         return []
-    mods = mod.split('.')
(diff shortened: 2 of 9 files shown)


IPython/utils/importstring.py
```diff
@@ -23,17 +24,17 @@ def import_item(name):
...
        The module that was imported.
     """
-    
+
     parts = name.rsplit('.', 1)
     if len(parts) == 2:
         # called with 'foo.bar....'
         package, obj = parts
-        module = __import__(package, fromlist=[obj])
+        module = import_module(package, fromlist=[obj])
```
minrk: I think this is the only one that isn't working. `fromlist` isn't supported in `import_module`, so we need to add some logic for importing objects. I'm not quite sure what the most obvious new-syntax for "import a.b.c, which might be a module or an object", since it will depend on what it is.

My guess is the best we can do is approximately:

```python
package, obj = parts
try:
    return import_module(name) # full name is a module
except ImportError:
    # not a module, maybe an attr on the parent module
    module = import_module(package)
    return getattr(module, obj)
```

drgarcia1986: you're right I missing that, give me a minute

drgarcia1986: done

IPython/core/completer.py
```diff
@@ -544,45 +545,45 @@ def back_latex_name_matches(text):
...
         help="""Whether to merge completion results into a single list
-        
+
         If False, only the completion results from the first non-empty
         completer will be returned.
         """
     ).tag(config=True)
     omit__names = Enum((0,1,2), default_value=2,
         help="""Instruct the completer to omit private method names
-
```
takluyver: Could we ask you to redo this without stripping all the whitespace from these files? Sorry, but it messes up git blame. If you're using atom, you can stop it doing this in Preferences > Packages > Whitespace.

IPython/core/completerlib.py
```diff
@@ -155,7 +156,7 @@ def is_importable(module, attr, only_modules):
 
 def try_import(mod, only_modules=False):
     try:
-        m = __import__(mod)
+        m = import_module(mod)
     except:
         return []
     mods = mod.split('.')
```
takluyver: There's some code here that can be cleaned up now. When you `__import__('foo.bar')`, it actually returns the `foo` module, and you have to get `bar` manually. `import_module` does this for us.

```
 1│ __import__('IPython.display')
 1> <module 'IPython' from '/home/takluyver/.local/lib/python3.5/site-packages/IPython/__init__.py'>

 2│ import importlib

 3│ importlib.import_module('IPython.display')
 3> <module 'IPython.display' from '/home/takluyver/.local/lib/python3.5/site-packages/IPython/display.py'>
```

takluyver:
Thanks. Other than the one more comment I just made, this looks good.

takluyver:
Thanks, I think this looks good.

drgarcia1986:
You're welcome! :smile:

Issue #10013, opened 2016-10-18
(Quasi-)incorrect keyword argument completion


anntzer:
The signature e.g. of `np.frombuffer` is
```
frombuffer(buffer, dtype=float, count=-1, offset=0)
```
but `frombuffer(<tab>` proposes completions with `count=`, `dtype=` and `offset=`.  Note that it is unlikely that one will ever want to write a call that starts with `np.frombuffer(count=...)`, because you'd typically want to pass the `buffer` argument first (yes, I know, technically you can pass it as keyword after another keyword argument.  I don't think that's a common case :-))

ivanov:
Would including `buffer=` in the completions be the behavior you expect?

anntzer:
I'm not really sure what the best solution is.  It may just be not completing any kwarg at that position, or only `buffer=`.

ivanov:
so the code in #10019 adds `buffer=` to the kwargs completion, but those still get sorted alphabetically. It'd be nice if the kwargs could complete in order they were written, but that's difficult to do with the code we have right now.

anntzer:
Would it be possible to just not complete kwargs at all if there are positional arguments that have not been filled in yet?  This would basically only miss the case of people wanting to specify positional arguments using keyword arguments, which, while legal, is pretty rare in my experience.

takluyver:
Doing that requires the completion code to understand all of the partial function call up to that point. Doing this with our current hacky pile of regex matching is going to be fiddly and error prone. We need to either switch to using jedi for completions, or create some better completion machinery of our own, before we can do anything like that.

anntzer:
On a related note, `np.load("<TAB>` (note the presence of the quote) completes kwargs (and globals) rather than filenames, which is incorrect.

Ark-kun:
>On a related note, np.load("<TAB> (note the presence of the quote) completes kwargs (and globals) rather than filenames, which is incorrect.

Same with the KEYWORD_ONLY functions. Suggesting arbitrary objects is incorrect for such function calls.

```
def my_func(*, aaa: str, bbb: int = 4) -> bool:
    """my_func docsrting"""
    return True
```

`my_func(<TAB>` should only have two suggestions: `aaa=` and `bbb=`. `my_func(xxx)` is always invalid in this case.

krassowski:
[...]

PR #10019, opened 2016-10-21
include positional arguments in completions

ivanov:
Sounds like this is the sort of thing that would address #10013.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -642,9 +642,7 @@ class IPCompleter(Completer):
 
         #regexp to parse docstring for function signature
         self.docstring_sig_re = re.compile(r'^[\w|\s.]+\(([^)]*)\).*')
-        self.docstring_kwd_re = re.compile(r'[\s|\[]*(\w+)(?:\s*=\s*.*)')
-        #use this if positional argument name is also needed
-        #= re.compile(r'[\s|\[]*(\w+)(?:\s*=?\s*.*)')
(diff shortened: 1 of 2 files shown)


takluyver:
Test failure:

```
FAIL: IPython.core.tests.test_completer.test_default_arguments_from_docstring
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/ipython/ipython/IPython/core/tests/test_completer.py", line 377, in test_default_arguments_from_docstring
    nt.assert_equal(kwd, ['ncall', 'resume', 'nsplit'])
AssertionError: Lists differ: ['self', 'int', 'resume', 'int'] != ['ncall', 'resume', 'nsplit']
```

ivanov:
Before I go and start trying to fix the remaining test, do we even want to do this? I can see it either way. I think the bigger issue is that our completion needs an overhaul to be more useful, as Thomas pointed out in [this comment](https://github.com/ipython/ipython/issues/10013#issuecomment-255536301)

How do others feel?

takluyver:
@Carreau I forget the details of what was going wrong with jedi. Should we be re-enabling that as the first step in overhauling completion?

Carreau:
Yes we can. Jedi had unicode-side-effects and the wrappers had bugs on
splittokens.

Carreau:
Jedi does not do that. We'll have to provide that ourselves or submit a patch jedi.

Carreau:

PR #10021, opened 2016-10-22
Remove sys version checks

ivanov:
This all started out innocently enough with me trying to remove sys_version checks for Python 3
since we're now committed to using Python 3.  I started off going  through and
remove code which depend on sys_version checks, and then removed some decorators, and the relevant `from __future__` imports.

the biggest part of the diff here is that `IPython/utils/tokenize2.py` which used to load either `_tokenize_py3.py` or `_tokenize_py2.py` has been replaced with the contents of `_tokenize_py3.py` (and  `_tokenize_py2.py` has been removed).


IPython/core/tests/test_interactiveshell.py
```diff
@@ -143,26 +143,13 @@ def __repr__(self):
 
     def test_future_flags(self):
         """Check that future flags are used for parsing code (gh-777)"""
```
takluyver: This test no longer actually tests the thing it was meant to test.

There is a hidden future import that we can use for testing, though it may be CPython-only:

```python
from __future__ import barry_as_FLUFL
1 <>  2   # This is now valid syntax
```

tools/tests/embed/embed_division.py
```diff
@@ -1,5 +1,4 @@
 """This tests that future compiler flags are passed to the embedded IPython."""
-from __future__ import division
```
takluyver: This file is also now not useful; it could also be adapted to use `barry_as_FLUFL`.

takluyver:
Thanks Paul; this mostly looks good, other than a few specific things I've commented on.

IPython/core/tests/test_interactiveshell.py
Carreau: or we can drop the test ? No ?

IPython/utils/text.py
```diff
@@ -517,7 +516,7 @@ def get_field(self, name, args, kwargs):
 # inside [], so EvalFormatter can handle slicing. Once we only support 3.4 and
 # above, it should be possible to remove FullEvalFormatter.
 
-@skip_doctest_py3
+@skip_doctest
```
Carreau: let's not skip doctest anymore and make the docstring python3 , even better could be a way to have that not a doctest.

IPython/utils/text.py
```diff
@@ -574,7 +573,7 @@ def vformat(self, format_string, args, kwargs):
         return u''.join(py3compat.cast_unicode(s) for s in result)
 
 
-@skip_doctest_py3
+@skip_doctest
```
Carreau: same.

Carreau (dismissed):

Carreau:
Thanks ! lot @ivanov. It's always a good feeling to remove a lot of code.

IPython/core/tests/test_interactiveshell.py
takluyver: Future flags are still a thing, and in Python 3.5 & 3.6 there's a real future flag that makes a difference again (`from __future__ import generator_stop`). So the test does make sense, but there isn't a great way to test it across all supported Pythons. The `barry_as_FLUFL` feature works at least for CPython (I haven't checked if other implementations support it).

Carreau:
🍰

ivanov:
+ ☕️  !

Issue #10003, opened 2016-10-13
Circular dependency matplotlib <- ipython <- matplotlib


daym:
Hi,

when using matplotlib-1.5.3 and ipython 3.2.1 there's a circular dependency.

matplotlib requires ipython in order to build the docs.

ipython requires matplotlib.sphinxext.mathmpl .

Also, there's an extra cycle: python-ipython requires python-numpy requires python-matplotlib requires python-ipython.

Can this be disentangled?

tacaswell:
I suggest we keep this discussion in the mpl issue.

ivanov:
closing this, the discussion continued in matplotlib/matplotlib/issues/7264.

PR #10022, opened 2016-10-22
Remove matplotlib requirement for building docs

takluyver:
See gh-10003

We weren't actually using most of these extensions, and the one we were using can be replaced by a standard Sphinx directive.


Carreau:
Thanks !

willingc:
Nice! Thanks @takluyver.

PR #10023, opened 2016-10-22
correct the  argument to warn function

srinivasreddy:


takluyver:
Thanks. The argument should actually be there, but i think it needs to be called `stacklevel` rather than `level`. It affects which line of code the warning shows up for.

srinivasreddy:
Done. Thanks.

takluyver:
Thanks :-)

PR #10024, opened 2016-10-22
use assertEqual instead of assertFalse or assertTrue

srinivasreddy:


takluyver:
I don't think this is an improvement.

If the code looks like `self.assertTrue(a == b)`, then using `self.assertEqual(a, b)` means it can give a better error message. But there's no clear advantage of doing `self.assertEqual(a, True)` instead of `self.assertTrue(a)`.

IPython/core/tests/test_inputsplitter.py
```diff
@@ -142,7 +142,7 @@ def test_source(self):
         self.isp._store('1')
         self.isp._store('2')
         self.assertEqual(self.isp.source, '1\n2\n')
-        self.assertTrue(len(self.isp._buffer)>0)
+        self.assertEqual(len(self.isp._buffer)>0, True)
```
takluyver: This could use `assertGreater`.

srinivasreddy:
Thanks.

Only one advantage, i can come up here
```
        self.assertEqual(None, False)
        self.assertFalse(None)  #assuming function defaulted on return type which supposed to return boolean
```

takluyver:
That's a valid point :-)

Carreau:
> Thanks.
> 
> Only one advantage, i can come up here
> 
>         self.assertEqual(None, False)
>         self.assertFalse(None)  #assuming function defaulted on return type which supposed to return boolean



> That's a valid point :-)

Enough for me. We can make the greater than in a separate PR. 

Thanks !

PR #10027, opened 2016-10-23
fun: whether this compile?

srinivasreddy:


takluyver:
It may be possible to speed up some of these with `scandir` once we stop supporting Python < 3.5.

Carreau:
> It may be possible to speed up some of these with scandir once we stop supporting Python < 3.5.

OR we can use the scandir module which is available on PyPI !

PR #10033, opened 2016-10-25
load multiple modules simultaneously

srinivasreddy:


IPython/extensions/autoreload.py
```diff
@@ -500,6 +500,11 @@ def aimport(self, parameter_s='', stream=None):
             else:
                 stream.write("Modules to reload:\n%s\n" % ' '.join(to_reload))
             stream.write("\nModules to skip:\n%s\n" % ' '.join(to_skip))
+        elif isinstance(modname.split(" "), list):
```
takluyver: `str.split()` will always return a list, even if you call it with an empty string. So this check is redundant. It does need some checking, however, because as it stands, you'll never end up in the branch below for `-foo`.

Carreau:
Thanks for the Pull Request !

I'd lige @fperez input on this IIRC the point of `aimport` was to be similar to `import ` so it should likely be with a comma as separation like import does.

Carreau:
Mering, I'll switch the separator to commas later.

PR #9449, opened 2016-05-09
"smart command mode" for ipdb (like pdb++).

anntzer:
If e.g. "c" is a variable currently visible, then "c" should print it
instead of running the "continue" Pdb command.  This idea was first
implemented by pdb++.

It is possible to force the standard pdb behavior (bypass variable
checking) by prefixing the command with "!!", e.g. "!!c" to continue.

PR #10035, opened 2016-10-25
override the emptyline method

tcztzy:
# What's the problem

When the command name is conflicted with variable name, the behavior of `emptyline` is print the value of the variable even the variable is not `lastcmd`. 

```
> /home/v-zit/test.py(1)<module>()
----> 1 n = 1
      2 next = 11
      3 print('end')

ipdb> n
> /home/v-zit/test.py(2)<module>()
      1 n = 1
----> 2 next = 11
      3 print('end')

ipdb>
1
ipdb>
```
# What did I do

I override the default `emptyline` method, which come from `cmd` module and not be called neither `pdb` nor `ipython`. As #9449, always add the "!!" prefix to force the standard behavior.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -307,6 +307,14 @@ class Pdb(OldPdb, object):
             return super(Pdb, self).parseline("!" + line)
         return super(Pdb, self).parseline(line)
 
+    def emptyline(self):
+        # Fix for #9449: If the commands are conflicted with the variable,
+        # such as `n` command and `n` variable, then the emptyline behavior as
+        # `!n` which print the value of variable, it performs not as we usually
+        # expect. Just add the "!!" prefix force standard behavior.
+        if self.lastcmd:
+            return self.onecmd("!!" + self.lastcmd)
+
     def new_do_up(self, arg):
         OldPdb.do_up(self, arg)
     do_u = do_up = decorate_fn_with_doc(new_do_up, OldPdb.do_up)


tcztzy:
**Test results for commit dbcb485 merged into master (27b6ba5)**
Platform: linux


Not available for testing: python2.7, python3.3

takluyver:
I think this would be more confusing than just repeating the last command entered, as it currently does.

The 'smart command mode' introduced in #9449 seems to be causing problems for some users. If you want to argue for it to be reverted back to how it worked before, go for it. But I don't think adding more 'guess what the user meant' like this makes it easier to use.

takluyver:
(I've just spotted that #10037 has been opened to propose switching this back)

tcztzy:
I do not guess what the user meant, just make the action entering a blank line works as it did before.

> Entering a blank line repeats the last command entered. Exception: if the last command was a list command, the next 11 lines are listed.
https://docs.python.org/3/library/pdb.html

takluyver:
It doesn't repeat the last command entered if the last command was inspecting a variable, though, because it always prepends `!!`, whether or not the last command had that. E.g.

```
ipdb> n
1
ipdb> n
1
ipdb> 
ipdb>   # This just stepped forward
```

tcztzy:
Because `n` variable is not a command, actually, in source code of pdb, `!n` which print the value of a variable `n` don't record in the `lastcmd` attribute.

cmd.py
```python
    def parseline(self, line):
        """Parse the line into a command name and a string containing
        the arguments.  Returns a tuple containing (command, args, line).
        'command' and 'args' may be None if the line couldn't be parsed.
        """
        line = line.strip()
        if not line:
            return None, None, line
        elif line[0] == '?':
            line = 'help ' + line[1:]
        elif line[0] == '!':
            if hasattr(self, 'do_shell'):
                line = 'shell ' + line[1:]
            else:
                return None, None, line
        i, n = 0, len(line)
        while i < n and line[i] in self.identchars: i = i+1
        cmd, arg = line[:i], line[i:].strip()
        return cmd, arg, line
```

A line is start with `!` which mean you want to `exec` it, and pdb consider it is not a command.

takluyver:
As the 'smart command mode' has been reverted, I'm closing this as no longer needed.

PR #10041, opened 2016-11-02
remove test_pr.py

ivanov:
This was the last item to address in resolving #7127, where it was
deemed that there isn't much value in keeping this around.

pinging @takluyver who brought test_pr.py into this world <3


takluyver:
Farewell, `test_pr.py`!

minrk:
Thanks, @ivanov!

Issue #9381, opened 2016-04-07
object.__sub<tab> does not propose __subclasses__ as tab-completion


anntzer:
All's in the title.

Probably something to do with being defined on the metaclass, and not that important I guess...

takluyver:
It's possible that the work @liukelly and @Carreau are doing on using Jedi completions will improve this, but I'm not sure.

srinivasreddy:
I tested this on master,

`In [1]: object.__subclasshook__`

that is what i received. I think we can close this issue. @takluyver

Carreau:
> I tested this on master,
> 
> In [1]: object.__subclasshook__
> 
> that is what i received. I think we can close this issue.

Indeed, thanks. Closing.

Issue #9606, opened 2016-06-10
Missing dunder completions


anntzer:
It seems that completion for the `__name__` and `__qualname__` attributes are missing for some functions:
```
np.roll.__n<TAB> -> __name__
sys.excepthook.__n<TAB> -> no completion
```
(note that within IPython, `sys.excepthook` is `IPython.terminal.ipapp.TerminalIPythonApp.excepthook`), so a "normal" function implemented in Python).

ivanov:
Thank you for your report. I've put up two possible workaround which at least address the `__dict__` not being ignored... Can't find a workaround  `__name__` and `__qualname__` at the moment, not sure what else has changed in the meantime.  #8355 is the pull request which created at least a part of this regression.

Carreau:
#10046 have fixed part of that. 

Bumping to a later IPython version to see if we can do better.

Issue #8330, opened 2015-04-17
Class attributes are shown regardless of __dir__ override 


patter001:
For some user interactions I'm trying to limit what the tab completion options are. I have implemented a `__dir__`, and dir(object) has the expected results. However, the tab completion still shows functions from the class -- inherited from a parent class.

I tried adding a `@classmethod` `__dir__` and I tried adding `__dir__` to the parent class, but neither one seems to help -- though..the `@classmethod` feels like it should have...and that might either be a python bug or "not a supported feature" :)

minrk:
This is the result of `dir2`, which explicitly looks for class attributes in addition to the output of `__dir__`, so it's not a bug per se, but behaving as intended. Whether it *should* behave that way is a different question.

@takluyver thoughts on what we should do about giving objects control over their own completions?

takluyver:
The default `__dir__()` method does return class attributes (including attributes of parent classes) as well as instance attributes, which makes me think that it should be up to code implementing `__dir__()` to show those if they want to - i.e. most implementations should do something like `return super().__dir__() + [...]`.

On the other hand, we presumably add the class attributes explicitly because at some point something didn't provide those from `__dir__()`. I don't know if that was an obsolete version of Python or some library, or if it would still be an issue if we changed it now.

mortada:
Interesting to see this thread because I was about to ask the same question. We are trying to limit the tab complete options dynamically for `pandas` `Series` (details in this PR https://github.com/pydata/pandas/pull/9910) 

Basically some of the attributes are only valid if the `Series` contains the appropriate data type (`Series.str` only works if it has strings, `Series.dt` only works if it has datetimes), so we'd like to get the tab completion to reflect that. However to our surprise removing the items from `__dir__` does not remove them from the tab completion.

Carreau:
OTOH, it should not be to the library to mess with completion, the
completer should be smart enough to know things like that. (I guess that
what types are for) and I doubt __dir__ is only use for completions.

takluyver:
I'd lean slightly towards saying that `__dir__` is informative, not definitive, so it's OK to use it to hide methods that you're sure are not relevant, like `s.str` for a non-string series. Off the top of my head, I can't think of anything other than completions that uses `dir(obj)`.

If we don't want to recommend that, though, then I think we should support a separate hook that objects can override to provide an informative list of attributes e.g. `_completions_dir_`. But my preference is to use `__dir__` for that rather than making a new mechanism.

[...]

PR #10046, opened 2016-11-07
restore tab completion behavior

### LINKED ISSUE #10044: Autocompletion with Init-Signatures of C extension classes
I don't quite get the autocompletion to work with C extension classes. These classes have a Signature (at least since python 3.4 by utilizing `__text_signature__`) but it seems IPython cannot autocomplete the parameters in `__text_signature__`.

For example:

```
>>> import pickle
>>> import inspect
>>> inspect.Signature.from_callable(pickle.Pickler)
<Signature (file, protocol=None, fix_imports=True)>
```

Note that the signature is saved as `__text_signature__`-attribute on ``pickle.Pickle``.

I would expect that after printing`pickle.Pickler(` and then hitting TAB it would show:
```
file=
protocol=
fix_imports=
# other stuff
```
but it doesn't. 

Classes seem to be special in this regard - IPython is definitly able to get the parameters for functions and methods! It just doesn't work for classes.

ivanov:
this is a minimal patch to restore proper tab completion behavior, a
partial revert of #8355. I think avoiding the explicit poking into the
__bases__ classes address the concerns #8330.

This is a minimal patch that closes #10044, #9381, and partially addresses #9606.

It is a bug that IPython tab-completion does not provide a superset of the
completions provided by the built-in `rlcompleter` module in Python.

I'll put up another alternative version that reverts #8355 as an alternative to this patch.

### DIFF
--- a/IPython/utils/dir2.py
+++ b/IPython/utils/dir2.py
@@ -41,6 +41,9 @@ def dir2(obj):
         # TypeError: dir(obj) does not return a list
         words = set()
 
+    if safe_hasattr(obj, '__class__'):
+        words |= set(dir(obj.__class__))
+
     # filter out non-string attributes which may be stuffed by dir() calls
     # and poor coding in third-party modules
 


ivanov:
for context - in `master`- if you `import os` and try to tab-complete the dunder methods, you'll get 
```
os.__all__
os.__builtins__
os.__cached__
os.__doc__
os.__file__
os.__loader__
os.__name__
os.__package__
os.__spec__
```

whereas if you `import rlcompleter` in plain python, you'll see 
```
os.__all__
os.__class__(
os.__delattr__(
os.__dict__
os.__doc__
os.__file__
os.__format__(
os.__getattribute__(
os.__hash__(
os.__init__(
os.__name__
os.__new__(
os.__package__
os.__reduce__(
os.__reduce_ex__(
os.__repr__(
os.__setattr__(
os.__sizeof__(
os.__str__(
os.__subclasshook__(
```

with this PR, tab completing on `os.__<tab>` will yield 

```
os.__all__
os.__builtins__
os.__cached__
os.__class__
os.__delattr__
os.__dict__
os.__dir__
os.__doc__
os.__eq__
os.__file__
os.__format__
os.__ge__
os.__getattribute__
os.__gt__
os.__hash__
os.__init__
os.__le__
os.__loader__
os.__lt__
os.__name__
os.__ne__
os.__new__
os.__package__
os.__reduce__
os.__reduce_ex__
os.__repr__
os.__setattr__
os.__sizeof__
os.__spec__
os.__str__
os.__subclasshook__
```


- this partially solves #9606
, specifically that `import nump as np` followed by `np.__d<tab>` does provide both `__doc__` AND ` __dict__` as options.

It does not, however,  provide `__name__` nor `__qualname__` completions to `sys.excepthook.__n<TAB>`, though it does provide `__new__` and `__ne__`, neither of which is provided in master.

ivanov:
also closes #9381

minrk:
Should this only be for dunder methods, any underscore-private methods, or everything? It seems like setting a custom `__dir__` can be meant to exclude inherited methods that you don't want to expose.

e.g.

```python
class A(object):
    def foo(self): pass

class B(A):

    def bar(self): pass
    
    def __dir__(self):
        return ['bar']
```

Seems like it's meant to *exclude* foo from completion, but this would restore foo, wouldn't it?

takluyver:
I've been digging into the different cases a bit:

- `os.__class__` (etc.) - the module type has a [specific `__dir__` implementation](https://hg.python.org/cpython/file/92f4b35d84d5/Objects/moduleobject.c#l709) which only lists the contents of the module, not attributes from the type.
- `object.__subclasses__` (#9381) - type objects also have a specific `__dir__` which excludes values from metaclasses, and a [comment there](https://hg.python.org/cpython/file/92f4b35d84d5/Objects/typeobject.c#l3180) indicates that this is deliberate:
> `__dir__` for type objects: returns `__dict__` and `__bases__`. We deliberately don't suck up its `__class__`, as methods belonging to the metaclass would probably be more confusing than helpful.
- #10044 is a separate issue which is not fixed by this
- `instance.method.__name__` (in #9606) - currently something of a mystery to me, but I don't think either this PR or #10047 makes it show up. `func.__name__` works, just not methods.

I think adding class attributes in total defiance of `__dir__` is probably too heavy-handed a response to this; class authors should be able to customise attribute completions if they like. A few possible options:

1. Add some special cases around basic types like modules and types, to lean more towards showing attributes.
2. Go into show-it-all mode (i.e. respect `__dir__` less) when the attribute name we're completing starts with double-underscore (or even single underscore).
3. Add another special method we check for (`_attribute_completions_`?) to customise attribute completions, and default to this more aggressive dir implementation if it's not defined. People who already use `__dir__` to customise completions might be annoyed about this.

Carreau:
I'll lean toward 2. We have/had `limit_to__all__` option on the completer, which default to False, and is marked as deprecated. So consistency lean toward ignoring custom `__dir__`.

Carreau:
Ok, 

I'm going to merge to restore previous behavior and open an issue to bikeshed as to which behavior is actually the one we want in the long run.

PR #10047, opened 2016-11-07
restore more tab completion behavior

### LINKED ISSUE #8330: Class attributes are shown regardless of __dir__ override

### LINKED ISSUE #10044: Autocompletion with Init-Signatures of C extension classes

ivanov:
An alternative to #10046, which closes #10044, #9381, and partially adresses #9606.

This is a complete revert of #8355, which resolved #8330, but instead
lead to completions not matching the set provided by standard libarary's
`rlcompleter`, which is in conflict with the behavior reported in #8330.

### DIFF
--- a/IPython/utils/dir2.py
+++ b/IPython/utils/dir2.py
@@ -19,6 +19,18 @@ def safe_hasattr(obj, attr):
     except:
         return False
 
+def get_class_members(cls):
+    ret = dir(cls)
+    if safe_hasattr(cls, '__bases__'):
+        try:
+            bases = cls.__bases__
+        except AttributeError:
+            # `obj` lied claiming to have __bases__ attr, ignore
+            pass
+        else:
+            for base in bases:
+                ret.extend(get_class_members(base))
+    return ret
 
 def dir2(obj):
     """dir2(obj) -> list of strings
@@ -40,6 +52,9 @@ def dir2(obj):
     except Exception:
         # TypeError: dir(obj) does not return a list
         words = set()
+
+    if safe_hasattr(obj, '__class__'):
+        words |= set(get_class_members(obj.__class__))
 
     # filter out non-string attributes which may be stuffed by dir() calls
     # and poor coding in third-party modules


ivanov:
for context - in `master`- if you `import os` and try to tab-complete the dunder methods, you'll get 
```
os.__all__
os.__builtins__
os.__cached__
os.__doc__
os.__file__
os.__loader__
os.__name__
os.__package__
os.__spec__
```

whereas if you `import rlcompleter` in plain python, you'll see 
```
os.__all__
os.__class__(
os.__delattr__(
os.__dict__
os.__doc__
os.__file__
os.__format__(
os.__getattribute__(
os.__hash__(
os.__init__(
os.__name__
os.__new__(
os.__package__
os.__reduce__(
os.__reduce_ex__(
os.__repr__(
os.__setattr__(
os.__sizeof__(
os.__str__(
os.__subclasshook__(
```

with this PR, tab completing on `os.__<tab>` will yield 

```
os.__all__
os.__builtins__
os.__cached__
os.__class__
os.__delattr__
os.__dict__
os.__dir__
os.__doc__
os.__eq__
os.__file__
os.__format__
os.__ge__
os.__getattribute__
os.__gt__
os.__hash__
os.__init__
os.__le__
os.__loader__
os.__lt__
os.__name__
os.__ne__
os.__new__
os.__package__
os.__reduce__
os.__reduce_ex__
os.__repr__
os.__setattr__
os.__sizeof__
os.__spec__
os.__str__
os.__subclasshook__
```


This partially solves #9606, specifically that `import nump as np` followed by `np.__d<tab>` does provide both `__doc__` AND ` __dict__` as options.

It does not, however,  provide `__name__` nor `__qualname__` completions to `sys.excepthook.__n<TAB>`, though it does provide `__new__` and `__ne__`, neither of which is provided in master.

I thought this behavior differed from #10046 , but it seems they are the same.

ivanov:
this also closes #9381.

ivanov:
It seems #10046 is simpler,  and I can't come up with a case where the behavior of this PR differs from that one, so perhaps the right thing to do is jus to close this one in favor of the other one.

takluyver:
From looking into the dir machinery in Python, I agree that #10046 should be equivalent, so I'll close this for now and focus discussion on that one.

PR #10048, opened 2016-11-08
Implement update_display

minrk:
- Sibling PR: ipython/ipykernel#204
- The IPython piece of: jupyter/jupyter_client#209

```python
display(obj, display_id='xxx')
```

sets a display id.

```python
update_display(obj, display_id='xxx')
```

updates the display in-place.

This is implemented through adding the `transient` dict to display-data, where `display_id` resides.

Still todo:

- [x] handle ipykernel not supporting the extra arguments (current stable)
- [x] handle-API for setting a display-id automatically and updating it
- [x] test exercise
- [x] example notebook

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -79,6 +79,7 @@ def _display_mimetype(mimetype, objs, raw=False, metadata=None):
 # Main functions
 #-----------------------------------------------------------------------------
 
+# use * to indicate transient is keyword-only
 def publish_display_data(data, metadata=None, source=None, *, transient=None, **kwargs):
     """Publish data and metadata to all frontends.
 
@@ -114,7 +115,7 @@ def publish_display_data(data, metadata=None, source=None, *, transient=None, **
         to specify metadata about particular representations.
     source : str, deprecated
         Unused.
-    transient : dict
+    transient : dict, keyword-only
         A dictionary of transient data, such as display_id.
         """
     from IPython.core.interactiveshell import InteractiveShell
@@ -214,6 +215,7 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
         return DisplayHandle(display_id)
 
 
+# use * for keyword-only display_id arg
 def update_display(obj, *, display_id, **kwargs):
     """Update an existing display by id
 
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -26,7 +26,7 @@ __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
 'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'JSON', 'Javascript',
 'Image', 'clear_output', 'set_matplotlib_formats', 'set_matplotlib_close',
-'publish_display_data']
+'publish_display_data', 'update_display']
 
 #-----------------------------------------------------------------------------
 # utility functions
@@ -78,7 +78,7 @@ def _display_mimetype(mimetype, objs, raw=False, metadata=None):
 # Main functions
 #-----------------------------------------------------------------------------
 
-def publish_display_data(data, metadata=None, source=None):
+def publish_display_data(data, metadata=None, source=None, transient=None, **kwargs):
     """Publish data and metadata to all frontends.
 
     See the ``display_data`` message in the messaging documentation for
@@ -113,11 +113,15 @@ def publish_display_data(data, metadata=None, source=None):
         to specify metadata about particular representations.
     source : str, deprecated
         Unused.
+    traisient : dict
+        A dictionary of transient data.
         """
     from IPython.core.interactiveshell import InteractiveShell
     InteractiveShell.instance().display_pub.publish(
         data=data,
         metadata=metadata,
+        transient=transient,
+        **kwargs
     )
 
 def display(*objs, **kwargs):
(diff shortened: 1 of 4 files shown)


IPython/core/display.py
```diff
@@ -113,11 +113,15 @@ def publish_display_data(data, metadata=None, source=None):
...
         Unused.
+    traisient : dict
+        A dictionary of transient data.
         """
     from IPython.core.interactiveshell import InteractiveShell
     InteractiveShell.instance().display_pub.publish(
         data=data,
         metadata=metadata,
+        transient=transient,
+        **kwargs
```
rgbkrk: Leaving room for later I see. 😄

IPython/core/display.py
```diff
@@ -145,11 +149,22 @@ def display(*objs, **kwargs):
...
-    include = kwargs.get('include')
-    exclude = kwargs.get('exclude')
-    metadata = kwargs.get('metadata')
+    raw = kwargs.pop('raw', False)
+    include = kwargs.pop('include', None)
+    exclude = kwargs.pop('exclude', None)
+    metadata = kwargs.pop('metadata', None)
+    transient = kwargs.setdefault('transient', {})
+    if 'display_id' in kwargs:
+        transient['display_id'] = kwargs.pop('display_id')
```
rgbkrk: 👍

rgbkrk (approved):

IPython/core/display.py
```diff
@@ -167,7 +182,14 @@ def display(*objs, **kwargs):
...
                 _merge(md_dict, metadata)
-            publish_display_data(data=format_dict, metadata=md_dict)
+            publish_display_data(data=format_dict, metadata=md_dict,
+            **kwargs)
+
+
+def update_display(*objs, **kwargs):
+    """Update an existing display"""
+    kwargs['update'] = True
+    return display(*objs, **kwargs)
```
rgbkrk: 👍 I like that I can use `display` with `update` = True and not have to use `update_display` directly as a function. :smile: /cc @ivanov

rgbkrk (approved):

minrk:
Here's [an example](https://nbviewer.jupyter.org/gist/minrk/300aaa15de9890b3af145cb993e9c964) demonstrating this with current master of ipykernel, notebook now that those PRs have been merged.

This is my interpretation of proposals from @fperez and @jasongrout about creating a handle and auto-generating a display_id with a boolean flag (in this case, I used `display_id=True` rather than a separate `track=True` flag):

- `display(obj, display_id=True)` generates a new display_id and returns a `DisplayHandle`
- `display(obj, display_id='given-id')` returns a `DisplayHandle` with given display_id
- `display(obj)` returns None
- `update_display(obj, display_id='id')` `display_id` is required
- `DisplayHandle.display(obj)` is equivalent to `display(obj, display_id=handle.display_id)`
- `DisplayHandle.update(obj)` is equivalent to `update_display(obj, display_id=handle.display_id)`

So the quickest way to create and update a display without needing to generate an id yourself:

```python
handle = display(x, display_id=True)
handle.update(y)
handle.update(z)
```

Possible functionality that is currently excluded:

1. the ability for items to define and record their own display_id, such as:

  - `display(x)` being equivalent to `display(x, display_id=x._ip_display_id_())`
  - `update_display(x)` discovering display_id from x itself (same as above, really)
  - allowing ExecuteResults to have display_id (again, ultimately boils down to getting display_id from the object)

IPython/core/display.py
```diff
@@ -78,7 +79,7 @@ def _display_mimetype(mimetype, objs, raw=False, metadata=None):
 # Main functions
 #-----------------------------------------------------------------------------
 
-def publish_display_data(data, metadata=None, source=None, transient=None, **kwargs):
+def publish_display_data(data, metadata=None, source=None, *, transient=None, **kwargs):
```
willingc: @minrk What does the `*` represent? Or typo?

IPython/core/display.py
```diff
@@ -182,14 +209,68 @@ def display(*objs, **kwargs):
...
                 # kwarg-specified metadata gets precedence
                 _merge(md_dict, metadata)
-            publish_display_data(data=format_dict, metadata=md_dict,
-            **kwargs)
+            publish_display_data(data=format_dict, metadata=md_dict, **kwargs)
+    if display_id:
+        return DisplayHandle(display_id)
+
 
+def update_display(obj, *, display_id, **kwargs):
```
willingc: Would be good to explain `*` in docstring here and above. Thanks!

minrk: It is in the docstring, in that `*` is Python's syntax to denote the end of positional args and `display_id` is labeled as keyword-only in the docstring. What would you recommend for comments regarding adopting new Python syntax?

rgbkrk: Wow, didn't realize that was available as a feature in Python. 😎

minrk: I added some comments on the methods to clarify these. Thanks!

minrk:
For @Carreau, @takluyver: I made use of the fact that we are Python 3-only now by making some new args explicitly keyword-only but still in the signature. Hooray for progress!

We can refine/revise the APIs in subsequent PRs, but this one is complete enough for people to play with and build opinions about, I think. Thanks, @rgbkrk!

willingc:
@minrk Thanks for the clarification re: '*' and arguments. I hadn't seen it used in the wild yet ;-)

Carreau:
Sweet !

@willingc , you will also see `/` in docstrings:

```
In [2]: input?
Signature: input(prompt=None, /)
Docstring:
Read a string from standard input.  The trailing newline is stripped.

The prompt string, if given, is printed to standard output without a
trailing newline before reading input.

If the user hits EOF (*nix: Ctrl-D, Windows: Ctrl-Z+Return), raise EOFError.
On *nix systems, readline is used if available.
```

It means "end of positional only arguments". Above it means that `prompt` default to None, but that you can't say `input(prompt='foo')`, you have to use `input('foo')`. It's an even more rare beast than `*` :-)

rgbkrk:
Oh does this mean no update display in Python 2 for now? I misconstrued earlier sentiment that we'd keep doing a python 2 kernel for a bit and enforcing python 3 for services (notebook, hub).

takluyver:
We're planning to keep supporting IPython 5 for Python 2 users for some time, but make IPython 6 require Python 3. As this is a new feature in IPython, Min has targeted it for IPython 6. We could maybe make a distinction between user-facing features and infrastructure features, and allow backporting some of the latter. I'd be -0.25 on doing that.

Separately, at some point the applications will require Python 3. Hub already does. For those, we don't plan to make an LTS version with Python 2 support as we have for IPython.

minrk:
I'm okay targeting this for 5.x as well, if people have that preference. I just wanted to note that my experience was slightly improved by the fact that IPython 6 has dropped py2 support. The differences are small, so it wouldn't be a big deal to backport.

It is only `ipython/ipython` that's dropped py2 so far, so the ipykernel, notebook changes will still propagate to py2 users. A library that wants to support this in py2 could do `ip.display_pub.publish(data=,metadata=,transient=)`.

rgbkrk:
Ok, great. I'll use `ip.display_pub.publish` for now within the lib I'm working with that supports both.

Carreau:
I would be ok backporting.  My Backporting bot that you can just `@-mention` is not ready yet though.

Carreau:
TIL @mention is a reserved name on github so you don't need to take extra-care when writing it.

rgbkrk:
@carreau would you be willing to backport this one? It's the feature I want to use and support across PySpark installations for updating job statuses.

Carreau:
I think that is reasonable enough to make API used by many package like TQDM consistant between stable version. Let's see if it applies cleanly to 5.x

@meeseeksdev backport

Carreau:
Ok, may need manual backport.

Issue #10037, opened 2016-10-26
Python variables obscur ipdb commands


tkossak:
This PR: #https://github.com/ipython/ipython/pull/9449 introduced annoying problem: I couldn't step through the following code (execute line 3):

```
n = 1
next = 1
print('end')
```

Because n/next commands weren't working:

```
$ ipdb test.py
> /tmp/test.py(1)<module>()
----> 1 n = 1
      2 next = 1
      3 print('end')

ipdb> next
> /tmp/test.py(2)<module>()
      1 n = 1
----> 2 next = 1
      3 print('end')

ipdb> next
> /tmp/test.py(3)<module>()
      1 n = 1
      2 next = 1
----> 3 print('end')

ipdb> next
1
ipdb> n
1
ipdb> !n
1
ipdb> !next
1
```

I discovered that in this case I can use standard pdb or I can get it to work in ipdb by using `!!n` (it still has a bug that @tcztzy tried to solve in https://github.com/ipython/ipython/pull/10035) but what do you think about changing this behavior: all ipdb builtin commands should work as they meant to and if we want to see python variables we use i.e.: `!n` instead of `n`. Variables' names that are not ipdb commands could still be displayed by just writing their name.

[StackOverflow thread](http://stackoverflow.com/questions/40224066/ipdb-commands-obscured-by-variables)

takluyver:
@anntzer, you wrote PR #9449, do you want to make the case for that behaviour?

Everyone: let's try using the Github 'reactions' to vote on this. Add a 'thumbs up' on this issue to vote for the original behaviour (`n` always being the debugger 'next' command), or 'thumbs down' to vote for the 'smart command mode' from #9449 (`n` shows the variable n if it exists, otherwise acts as the 'next' command).

We'll take votes as an indicator, not a binding decision.

anntzer:
My personal experience with debugging (which may be different from other people's...) is that in the vast majority of cases, I call `pdb.set_trace()` at the right place or `pdb.pm() / %debug` after an exception bubbles out, check where I am, inspect a few variables, and leave.  I very rarely step the debugger.  Especially in the context of IPython, I'd guess that most users also enter the debugger via `%debug` and use it for post-mortem'ing.

Thus, the choice of favoring printing variables rather than executing commands (which I did not invent: I got the idea from https://pypi.python.org/pypi/pdbpp/#smart-command-parsing) simply reflects my usage pattern: most of the times, printing the variable value is what I want to do.

[...]

PR #10050, opened 2016-11-09
Revert debugger 'smart command mode'

takluyver:
This is a manual revert of PR #9449

Closes gh-10037

I don't think there's currently enough support for smart command mode to be worth adding a config option for it now. We can always add this later.


Carreau:
Want to add something to the docs and what's new ?

Carreau:
Actually we can do it in a separate PR.

PR #10052, opened 2016-11-09
Whatsnew entry for reverting smart debugger

takluyver:


ivanov:
thanks @takluyver ! 👍

ivanov:
oops, wrong milestone, changed to 6.0 now.

takluyver:
#10050 is milestoned for 5.2 - I think there's enough people who dislike the new mode that it makes sense to revert it in a bugfix. That's open for discussion, but the milestone here should match that of #10050.

PR #10056, opened 2016-11-15
fix typo

fuho:


willingc:
Thanks for the corrections @fuho :cookie:

PR #10061, opened 2016-11-22
remove IPython.core.prompts

twmr:
since it is no longer used by ipykernel (removed in

commit 91c61ac16559576fa817d1712db7cf3df1142fdf
Author: Dave Willmer <dave.willmer@gmail.com>
Date:   Mon Apr 25 18:25:33 2016 +0100

    Cleanup comm and comm manager classes
)

Also see the last ipython commit, where prompts.py was changed

commit 566cfa8349748a97e8dfa590fe7a7c6d981b5b2b
Author: Thomas Kluyver <thomas@kluyver.me.uk>
Date:   Wed May 18 15:14:20 2016 +0100

    Remove PromptManager class

    LazyEvaluate is still used in ipykernel

What surprises me is that the LazyEvaluator was removed from ipykernel before @takluyver removed the PromptManager class from ipython. Do we support that ipython6 can be used together with ipykernel < 4.4 ?


minrk:
> What surprises me is that the LazyEvaluator was removed from ipykernel before @takluyver removed the PromptManager class from ipython.

In master, yes, but not in a release. ipykernel 4.4 was released in August. In general, master of every individual IPython or Jupyter package should work with the *stable* version of everything else unless something has gone wrong. Beyond that, every breaking change should really solve a problem. I don't think cleaning up is a good enough reason to break compatibility in general, but we've done two releases of ipykernel that will work, and IPython 6 is still some time off, so I think this is probably okay.

Carreau:
I'm bumping that to 7.0 as I think we are closing in to 6.0 and it's cheap to keep one more release.

Carreau:

Issue #10090, opened 2016-12-14
Add repr method for arbitrary mime-types


minrk:
We have `_repr_foo_` methods for a few common mime-types, but this mapping doesn't extend well to custom mime-types. I propose we add a new magic method that returns the full, formatted mime data, such as:

```python
def _repr_mime_(self):
    """Return mimetype-bundled representation(s) of me"""
    return {
        'data': {
            'application/vnd.tableschema.v1+json': {...},
            'text/html': '<div>...',
        'metadata': {},
    }
```

details: should this override or be in addition to existing methods if found?

cc @rgbkrk

rgbkrk:
Thanks for raising this up! It should likely override since it's a full manifest.

minrk:
Oops, sorry for the dupe of #10058. Copying comment from there: The `_ipython_display_` escape hatch makes this less urgent (and is perhaps why we still haven't done it), but this should still be done nonetheless.

Carreau:
Do any of you want to tackle that for 6.0, or should we bump to wishlist ?

rgbkrk:
Wishlist is fine, I'm working on other items for the next couple weeks.

Issue #9456, opened 2016-05-12
''BMP' format not supported?


zhengwx11:
from ipython.display import Image
Image("a_image_file.bmp")

Output: Cannot embed the 'bmp' image format

 'BMP' is a VARY simple image format with nearly no compression techniques. Why it is not supported?

Or do I miss something?

takluyver:
I think it's just never come up. BMP is not widely used on the web: PNG is widely supported and ~always preferable. For what we're doing, the complexity of the file format doesn't matter - we just pass the data through to the browser - so it was written with the most common formats, png and jpeg.

zhengwx11:
well, that is a really a bad news to me.
thank you for your replying anyway.


在 2016/5/12 22:52, Thomas Kluyver 写道:

srinivasreddy:
PR raised here - https://github.com/ipython/ipython/pull/10094

takluyver:
Note to people thinking of implementing this: I'd like to see two things before an implementation.

1. A persuasive case as to why this is necessary. BMP format images are not common nowadays, they are larger than other file formats and can be losslessly converted to PNG images (which are already supported). Given that every extra piece in the code is added complexity to maintain, is this piece valuable enough to add?
2. If you do think this is worth supporting, should we add the machinery to send bmp data to the browser for use in a data URL? Or should we add an automatic conversion for bmp data to png data, using a tool like Pillow where it's available?

PR #10094, opened 2016-12-16
Add support for bmp image format

srinivasreddy:


srinivasreddy:
@takluyver  i don't know how to test this. Could you please guide?

takluyver:
I'm not sure what to test other than checking that `Image` allows `format='bmp', embed=True`.

I don't think this will actually make `Image` do what you want - you add support for putting bmp data in, but nothing ever takes it out of `Image`. Maybe the best option is to convert to png on demand, and only support embedding bmp if the tools we need to do the conversion are found. To be honest, I'm still not sure why anyone wants to embed bmp data anyway...

IPython/core/display.py
```diff
@@ -858,8 +876,10 @@ def __init__(self, data=None, url=None, filename=None, format=None,
             if ext is not None:
                 if ext == u'jpg' or ext == u'jpeg':
                     format = self._FMT_JPEG
-                if ext == u'png':
+                elif ext == u'png':
                     format = self._FMT_PNG
+                elif ext == u'bmp' or ext == u'bm':
```
takluyver: Is `.bm` really an extension for bitmaps? I've never seen it, Wikipedia [doesn't mention it](https://en.wikipedia.org/wiki/BMP_file_format), and my system does not recognise it. Let's stick to `.bmp` unless there's a clear need for another extension.

srinivasreddy: Done.

IPython/core/display.py
```diff
@@ -952,6 +972,10 @@ def _repr_jpeg_(self):
         if self.embed and (self.format == u'jpeg' or self.format == u'jpg'):
             return self._data_and_metadata()
 
+    def _repr_bmp_(self):
```
takluyver: We don't have anything that checks for `_repr_bmp_` at the moment, and I'm not convinced it's common enough to be worth adding code for. @minrk ?

minrk: I would prefer to avoid adding any new `_repr_<single-type>_` methods, instead opting for any new representations being added via a single arbitrary mime-type method (#10090).

srinivasreddy: @takluyver  Done. @minrk

takluyver:
I think this PR is once again in a state that won't achieve anything; it lets you put bmp format data into `Image`, but nothing will actually get it out of there.

Nobody's ever made a persuasive case for embedding bmp images: they're not common nowadays, and can be losslessly converted to png (which we do support embedding). I'm going to close this for now. If anyone wants to make the case, please do so on #9456.

PR #10097, opened 2016-12-20
Add `pexpect` to  install_requires list

srinivasreddy:
`pexpect` package has been used here - https://github.com/ipython/ipython/blob/master/IPython/utils/_process_posix.py#L23


takluyver:
Thanks; it's already actually in the file a few lines below. It's only required on non-Windows platforms, and platform-dependent requirements work a bit differently.

srinivasreddy:
Thanks. 👍

PR #10099, opened 2016-12-21
Drop requirements which were only needed on Python 2

takluyver:


setup.py
```diff
@@ -212,20 +212,15 @@ def run(self):
 # but requires pip >= 6. pip < 6 ignores these.
 
 extras_require.update({
-    ':python_version == "2.7"': ['backports.shutil_get_terminal_size'],
     ':python_version == "2.7" or python_version == "3.3"': ['pathlib2'],
```
minrk: want to update this one to only be 3.3?

takluyver: Done

Carreau:
Thanks !

PR #10100, opened 2016-12-21
Add stacklevel=2 to warnings function

srinivasreddy:


IPython/core/interactiveshell.py
```diff
@@ -354,9 +354,8 @@ def _exiter_default(self):
     @observe('prompt_in1', 'prompt_in2', 'prompt_out', 'prompt_pad_left')
     def _prompt_trait_changed(self, change):
         name = change['name']
-        warn("InteractiveShell.{name} is deprecated since IPython 4.0 and ignored since 5.0, set TerminalInteractiveShell.prompts object directly.".format(
-                name=name)
-        )
+        warn("InteractiveShell.{name} is deprecated since IPython 4.0 and ignored since 5.0, set TerminalInteractiveShell.prompts object directly.".format(name=name), stacklevel=2)
```
takluyver: I suspect that `stacklevel=2` is insufficient for traitlet changed methods, because there's likely to be several more Python layers between them and the user. @Carreau, have you fixed warnings on any of these?

Carreau:
No, not to any of theses, and IIRC, depending on how/when you call it the
number of stacklevel may differ.

IPython/core/interactiveshell.py
takluyver: @srinivasreddy could you back out this change for now? I'd rather it was wrong in the obvious simple way than wrong in a new and confusing way. Thanks!

srinivasreddy: @takluyver I presume rest are OK.

takluyver: Yep, I think so.

srinivasreddy: @takluyver Done. Let's merge it.

takluyver:
Thanks

PR #10102, opened 2016-12-22
Insert a new line after catching KeyboardException in ask_yes_no

sachet-mittal:
Issue: 10098

### DIFF
--- a/IPython/utils/io.py
+++ b/IPython/utils/io.py
@@ -173,6 +173,7 @@ def ask_yes_no(prompt, default=None, interrupt=None):
         except KeyboardInterrupt:
             if interrupt:
                 ans = interrupt
+            print("\r")
         except EOFError:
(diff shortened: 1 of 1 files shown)


IPython/utils/io.py
```diff
@@ -173,6 +173,7 @@ def ask_yes_no(prompt, default=None, interrupt=None):
         except KeyboardInterrupt:
             if interrupt:
                 ans = interrupt
+                print("\r")
```
takluyver: I think this should be outside the if clause, i.e. indented one level less, so that it also does the right thing if the caller hasn't supplied an answer to use on interrupt.

sachet-mittal: Done.

takluyver:
Thanks, looks good to me. :-)

Issue #9102, opened 2015-12-26
remove `call_tip` in oinspect.


Carreau:
Is unused in IPython codebase, and in ipykernel too. 

it also seem like the `if callable_obj` ~ line 835 in oinspect does not have many effect if removed , as the `IPython.utils.signature` utils use for the `definition` of the callable above, is already capable of extracting the callable information. 

We might be able to simplify this logic.

Marking as 5.0 as API breaking, might want a deprecation warning before if we decide it's worth removing.

minrk:
Yup, I think oinspect could stand a refactoring. There's also been discussion of making oinspect itself a standalone package, so it could be used outside IPython.

srinivasreddy:
Raised a PR for this - https://github.com/ipython/ipython/issues/9102

jzf2101:
@Carreau are we ready to close this?

jzf2101:
@minrk ?

minrk:
Yes, thanks!

PR #10104, opened 2016-12-23
remove call_tip function and test case for that

srinivasreddy:
https://github.com/ipython/ipython/issues/9102


Carreau:
Thanks ! Though there seem to be some use of `check_calltip` that would need to be removed as well. And I think we want to go through a Deprecation period (ie undocument with `@undoc` and raise DeprecationWarnings) before removing.

srinivasreddy:
@Carreau pls let me know what do you think!.

Carreau:
Hi @srinivasreddy, 

Sorry I mis-expressed myself ! 

The `@undoc` and `DEPRECATED ` docstring are great ! 

I would not have removed the all docstring, and left the test for the later actual removal.
Also the next version of IPython will likely be 6.0 and not 5.2 but you had no way of doing that. 

So I'll merge and we can fix things if necessary later !

Thanks ! you contribution are highly appreciated.

PR #10105, opened 2016-12-24
Convert iteritems() to items()

srinivasreddy:


IPython/utils/py3compat.py
```diff
@@ -172,9 +172,6 @@ def isidentifier(s, dotted=False):
             return all(isidentifier(a) for a in s.split("."))
         return s.isidentifier()
 
-    xrange = range
-    def iteritems(d): return iter(d.items())
-    def itervalues(d): return iter(d.values())
```
takluyver: Be careful removing stuff, even if it's not used anywhere in IPython; we've had a couple of times before when we've removed things from IPython and then found that we're still using them in another module, like ipykernel.

srinivasreddy: Thanks a ton. Then i will keep this file as is. Will gradually change all projects references of module py3compat.py to py3 function calls,  then i think about remove it.

takluyver:
Thanks! This is now quite a bit to review, so please hold any more changes for future PRs.

srinivasreddy:
Okay

srinivasreddy:
@takluyver  If you want me to divide this PR into smaller ones, please let me know.

IPython/terminal/magics.py
```diff
@@ -26,7 +26,7 @@ def get_pasted_lines(sentinel, l_input=py3compat.input, quiet=False):
         prompt = ""
     while True:
         try:
-            l = py3compat.str_to_unicode(l_input(prompt))
```
minrk: str_to_unicode is a no-op on Python 3, and should be replaced with no call at all, rather than casting `str` to `str`.

srinivasreddy: Done.

IPython/core/logger.py
```diff
@@ -193,8 +192,8 @@ def log_write(self, data, kind='input'):
             write = self.logfile.write
             if kind=='input':
                 if self.timestamp:
-                    write(str_to_unicode(time.strftime('# %a, %d %b %Y %H:%M:%S\n',
-                                        time.localtime())))
+                    write(str(time.strftime('# %a, %d %b %Y %H:%M:%S\n',
```
takluyver: The `str` here shouldn't be necessary, I think. `str_to_unicode` is a no-op on Python 3, because `str` is unicode.

IPython/core/magics/logging.py
```diff
@@ -138,7 +137,7 @@ def logstart(self, parameter_s=''):
                 for n in range(1,len(input_hist)-1):
                     log_write(input_hist[n].rstrip() + u'\n')
                     if n in output_hist:
-                        log_write(str_to_unicode(repr(output_hist[n])),'output')
+                        log_write(str(repr(output_hist[n])),'output')
```
takluyver: Again, we shouldn't need the call to `str`.

IPython/core/magics/namespace.py
```diff
@@ -460,8 +459,8 @@ def type_name(v):
                 try:
                     vstr = str(var)
                 except UnicodeEncodeError:
-                    vstr = unicode_type(var).encode(DEFAULT_ENCODING,
-                                               'backslashreplace')
+                    vstr = var.encode(DEFAULT_ENCODING,
+                                      'backslashreplace')
```
takluyver: I'm pretty sure this except block can't happen any more under Python 3, but leave it for now; I'll double check that and work out if anything else is needed instead.

IPython/core/prompts.py
```diff
@@ -20,7 +20,7 @@ def __str__(self):
         return str(self())
 
     def __unicode__(self):
-        return py3compat.unicode_type(self())
+        return self.__str__()
```
takluyver: The `__unicode__` magic method is no longer used on Python 3, so we can get rid of it.

IPython/core/tests/test_profile.py
```diff
@@ -121,7 +121,7 @@ def test_list_profiles_in():
     # No need to remove these directories and files, as they will get nuked in
     # the module-level teardown.
     td = tempfile.mkdtemp(dir=TMP_TEST_DIR)
-    td = py3compat.str_to_unicode(td)
+    td = str(td)
```
takluyver: Again,  we don't need to call `str()`. Here, we don't need the line at all.

IPython/utils/path.py
```diff
@@ -87,7 +87,7 @@ def unquote_filename(name, win32=(sys.platform=='win32')):
 def compress_user(path):
     """Reverse of :func:`os.path.expanduser`
     """
-    path = py3compat.unicode_to_str(path, sys.getfilesystemencoding())
+    path = str(path)
```
takluyver: Same again.

takluyver:
I've looked through all the changes now; once the specific bits I pointed out are fixed, I think this will be ready. Thanks!

srinivasreddy:
@takluyver  Done. Please feel free to squash.

takluyver:
Thanks. I don't generally bother squashing unless someone has added/deleted a whole load of stuff and then undone it in a later commit.

PR #10108, opened 2016-12-25
Show traceback when an exception occurs for `%time`

srinivasreddy:


srinivasreddy:
here is the output
```
In [3]: %%time 
   ...: foo
   ...: 
<timed eval> in <module>()

NameError: name 'foo' is not defined
```

takluyver:
Thanks

PR #10124, opened 2017-01-05
Add appveyor.yml

srinivasreddy:


Carreau:
I can't seem to see the IPython org on AppVeyor. @rgbkrk for whatever reason my brain tell me you can help me with that.

srinivasreddy:
I don't think we have created an `ipython/ipython` account yet on appveyor.

Carreau:
> I don't think we have created an ipython/ipython account yet on appveyor.

the problem is not that it's not created, it is that app-veyor does not list me has having access. So I can't create it.

rgbkrk:
I don't think we have an IPython org on appveyor (at least none that I can access). We do have a jupyter org account, we could create an IPython one.

Carreau:
> I don't think we have an IPython org on appveyor (at least none that I can access). We do have a jupyter org account, we could create an IPython one.

Well if you can figure out how to enable the IPython org or the IPython/IPython repo on Appveyor, that would be great.  It just plainly refuse to admit I have access to the IPython org, though it does show me all of Jupyter/Xonsh/BIDS/... and the bunch or orgs I'm in. My guess is that IPython use to be a User before an org so it's confused.

rgbkrk:
Set up an IPython org and added you as a maintainer @Carreau. https://ci.appveyor.com/project/IPython/ipython

I think you'll have to make another commit or rebase to trigger AppVeyor for this PR. I'd prefer to merge this sooner rather than later so it doesn't make other builds fail by them not having the AppVeyor config setup.

Carreau:
Thanks a lot @rgbkrk ! 

I've rebased and pushed so Appveyor should have been triggered. I'll merged once test passes, thanks !

rgbkrk:
Thanks @srinivasreddy and @Carreau!

Carreau:
Thanks !

PR #10135, opened 2017-01-07
remove python2 specific code

srinivasreddy:


IPython/lib/pretty.py
```diff
@@ -108,15 +108,7 @@ def _safe_getattr(obj, attr, default=None):
...
-if PY3:
-    CUnicodeIO = StringIO
-else:
-    class CUnicodeIO(StringIO):
-        """StringIO that casts str to unicode on Python 2"""
-        def write(self, text):
-            return super(CUnicodeIO, self).write(
-                cast_unicode(text, encoding=get_stream_enc(sys.stdout)))
-
+CUnicodeIO = StringIO
```
Carreau: We may want to search for CUnicodeIO and replace instance by StringIO directly then deprecate CUnicodeIO

Carreau:
Thanks ! That's great ! I've left one comment that i'll turn into an issue.

IPython/lib/pretty.py
srinivasreddy: I am not sure of how to go about deprecating variables. Point in case please?

Carreau: Hum, that's a bit tricky especially at import, but as here it is a class, so what you can do is make a dummy wrapper that warns on instantiation. That's the non-hackish way of doing it : 

```
class CUnicodeIO:

    def __init__(self, *args, **kwargs):
         warnings.warn("CUnicodeIO is deprecated since IPython 6.0 in favor of io.StringIO", DeprecationWarnings, stacklevel=2)
        super().__init__(*args, **kwargs)
```

The hackish way would be (for example) to dynamically create a module using [ModuleType](https://docs.python.org/3.5/library/types.html#types.ModuleType), and replace itself in `sys.modules`. The module you create _may_ be able to implement a `__getattribute__` and raise a deprecation warning on access. 

But that's ugly. Grep for ShimModule to see how we've done similar things in other places.

PR #10139, opened 2017-01-08
Add 3.6 version target

srinivasreddy:
Now that python 3.6  is released, I want to execute the test suite as on 3.6 as well and also nightly build points 3.7.0a  now.


srinivasreddy:
@Carreau  What do you think of this?

Carreau:
Yep, sounds about right.

PR #10144, opened 2017-01-09
Remove uses of compatibility builtin_mod and builtin_mod_name

takluyver:
@srinivasreddy here's one you can review :-)

### DIFF
--- a/IPython/lib/deepreload.py
+++ b/IPython/lib/deepreload.py
@@ -32,14 +32,13 @@ re-implementation of hierarchical module import.
 #  the file COPYING, distributed as part of this software.
 #*****************************************************************************
 
+import builtins as builtin_mod
 from contextlib import contextmanager
 import imp
 import sys
 
 from types import ModuleType
 from warnings import warn
-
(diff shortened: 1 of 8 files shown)


IPython/lib/deepreload.py
```diff
@@ -320,13 +319,10 @@ def deep_reload_hook(m):
     return newm
 
 # Save the original hooks
-try:
-    original_reload = builtin_mod.reload
-except AttributeError:
-    original_reload = imp.reload    # Python 3
+original_reload = imp.reload    # Python 3
```
srinivasreddy: Looks like imp.reload is deprecated. May be we should use `importlib.reload` ?

srinivasreddy (changes requested):

IPython/lib/deepreload.py
takluyver: Well spotted, thanks!

takluyver: Ah, rats, we still need `imp.reload` while we're supporting 3.3. I'll drop the last commit.

srinivasreddy (approved):

Issue #10140, opened 2017-01-08
Deprecate CUnicodeIO


Carreau:
#10135 got rid f our CUnicodeIO wrapper, which is now an identity alias to StringIO.
We should check for usage, and remove/deprecate accordingly.

PR #10145, opened 2017-01-09
Deprecate CUnicodeIO

takluyver:
I don't see any usage of it, but it's easy enough to deprecate it for a cycle just in case.

Closes gh-10140


srinivasreddy (approved):
Awesome. :+1:

Carreau:
`@undoc` ?

Carreau:
+1

takluyver:
undoc-ed

Issue #10138, opened 2017-01-08
Initial blank line of multiline output


kwankyu:
The following is normal.
```
In [1]: class Foo(object):
   ...: 
   ...:     def __repr__(self):
   ...:         return "foo"
   ...: 
   ...: 
   ...: class Bar(object):
   ...: 
   ...:     def __repr__(self):
   ...:         return "bar\nbar"
   ...:     

In [2]: Foo()
Out[2]: foo

In [3]: Bar()
Out[3]: 
bar
bar
```
But in Sage that uses IPython, the initial blank line for multiline output is just redundant.
```
sage: class Foo(object):
....: 
....:     def __repr__(self):
....:         return "foo"
....: 
....: 
....: class Bar(object):
....: 
....:     def __repr__(self):
....:         return "bar\nbar"
....:     
sage: 
sage: Foo()
foo
sage: Bar()

bar
bar
```
I am not sure if this is a bug of IPython or of Sage. I would appreciate comments by anyone who knows better.

kwankyu:
I investigated this issue a bit more. The method `write_output_prompt` in the class `RichPromptDisplayHook` sets the attribute `prompt_end_newline` to true when the output prompt ends explicitly with `\n`. In Sage, the output prompt is just empty string. I think that in this case too, the attribute `prompt_end_newline` should be set true. Thus we can consider that the current erroneous behavior in Sage is caused by this bug in IPython.

takluyver:
Sounds reasonable. Do you want to make a PR?

kwankyu:
Actually I am not familiar with IPython code and collaborative development in Github. So it would take some time for me to make a pull request of a patch. If you or any one can do this easily, then I would be happy to concede.

takluyver:
OK, see #10146

PR #10146, opened 2017-01-09
Empty prompt effectively ends with newline

takluyver:
Closes gh-10138

### DIFF
--- a/IPython/terminal/prompts.py
+++ b/IPython/terminal/prompts.py
@@ -62,11 +62,15 @@ class RichPromptDisplayHook(DisplayHook):
     """Subclass of base display hook using coloured prompt"""
     def write_output_prompt(self):
         sys.stdout.write(self.shell.separate_out)
-        self.prompt_end_newline = False
+        # If we're not displaying a prompt, it effectively ends with a newline,
+        # because the output will be left-aligned.
+        self.prompt_end_newline = True
+
         if self.do_full_cache:
             tokens = self.shell.prompts.out_prompt_tokens()
-            if tokens and tokens[-1][1].endswith('\n'):
-                self.prompt_end_newline = True
+            if tokens and not tokens[-1][1].endswith('\n'):
+                # Ask for a newline before multiline output
+                self.prompt_end_newline = False
             if self.shell.pt_cli:
                 self.shell.pt_cli.print_tokens(tokens)
             else:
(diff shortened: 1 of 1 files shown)


kwankyu:
Thank you!

IPython/terminal/prompts.py
```diff
@@ -62,11 +62,15 @@ class RichPromptDisplayHook(DisplayHook):
...
-        self.prompt_end_newline = False
+        # If we're not displaying a prompt, it effectively ends with a newline,
+        # because the output will be left-aligned.
+        self.prompt_end_newline = True
+
         if self.do_full_cache:
             tokens = self.shell.prompts.out_prompt_tokens()
-            if tokens and tokens[-1][1].endswith('\n'):
-                self.prompt_end_newline = True
+            if tokens and not tokens[-1][1].endswith('\n'):
```
kwankyu: The test should be
 `tokens and len(tokens[-1][1]) > 0 and not tokens[-1][1].endswith('\n')`

kwankyu:
Unfortunately, the patch does not fix the issue yet, as an empty prompt still passes the "not ending with newline" test. See the comment attached to the code.

kwankyu:
For your informtion, Sage uses
```
    def out_prompt_tokens(self):
        return [
            (Token.OutPrompt, ''),
        ]   
```

kwankyu:
If you think that a token string should not be empty and Sage should use `[]` for empty output prompt, then your original code will work.

takluyver:
It's not an error for a token string to be empty, but as far as I know there's no need: Sage could simply use an empty list.

Your version might still do the wrong thing if there's an empty final token preceded by one or more non-empty tokens. I've pushed a different fix.

kwankyu:
As you said, empty token string would be of no use, possibly except in the case when the whole prompt is an empty string just as in Sage. So I think we can ignore the impractical case that non-empty tokens followed by an empty final token.

Your fix will work, but it seems to me a bit overdone as the tokens may be rendered twice, once for the test and once for `self.shell.pt_cli.print_tokens`(thought I don't know what this is exactly :-)

takluyver:
'Rendering' them for the check is just concatenating strings together, so I'm not too worried about the performance.

kwankyu:
Ok. I confirm that the patch works well. 

Do I have anything to do for this pull request to be merged and closed, like giving a positive review?

takluyver:
Thanks! There's nothing specific you need to do. We don't normally merge our own PRs, so I'll give other people a few days to merge it. If nobody does and there are no objections, I'll merge it myself.

willingc (commented):
@takluyver LGTM

Issue #9908, opened 2016-08-24
Startup performance. 


Carreau:
I had to use IPython the other days in a bootcamp, on machine that do not have ssd, or 146Gb of Ram, and 36Ghz CPU. IPython took up to **10sec** to start in the command line. I think we should try to make the startup time smaller and see if there are a number of operations that could be delayed to later.

gforsyth:
@Carreau have you played around with @scopatz's amalgamation package yet?   https://github.com/xonsh/amalgamate

https://github.com/xonsh/lazyasd might help also

takluyver:
I've done a bit of work on this years ago - mainly trying to eliminate imports of modules we don't actually need. It's hard to make much difference that way, but we're probably due for another pass. @scopatz's tools look neat, though I'd be a bit worried about the clever things they do with imports breaking other things we haven't thought of.

I wonder how much of that startup time is due to the `pkg_resources` problem where it scans all installed packages on import?

gforsyth:
On our end, is was almost entirely the `pkg_resources` problem

scopatz:
Hi all, my guess is that lazyasd would get IPython most of the way there.  As @takluyver said, I think you probably end up actually using most of your imports in IPython, so really the only thing amalgamate would help with is filesystem thrashing (which can be a big deal on older machines).

Implementing lazily loaded data structures, lazily compiled regexes, etc can have a huge effect on startup time.

> @scopatz's tools look neat, though I'd be a bit worried about the clever things they do with imports breaking other things we haven't thought of.

We have yet to encounter anything that wasn't a couple of line change in xonsh.  There are some annoyances with debugging and amalgamate. We have yet to figure out how to get the traceback to point to the original file.

> I wonder how much of that startup time is due to the pkg_resources problem where it scans all installed packages on import?

> On our end, it was almost entirely the pkg_resources problem

Once the code base is lazy, `pkg_resources` does become the dominating effect.  However, lazyasd has a 'fix' for this by loading it in a background thread. This happens is such a way that if someone actually tries to use it, the calling thread will block until the module is loaded.  See https://github.com/xonsh/lazyasd#background-imports

This is why I think you might be able to mostly get away with lazyasd.  In any event, it is just a single file that you should feel free to drop into IPython if you don't want to have another dependency.  As someone who isn't doing the work, I think it is worth trying out :)

willingc:
@Carreau Do you remember which operating system took 10s at the bootcamp?

[...]

PR #10147, opened 2017-01-10
Prevent pygments loading pkg_resources for faster startup

takluyver:
Following discussion on #9908, this avoids loading pkg_resources when IPython starts, making it faster to start.

How much faster? The timings vary quite a bit, but at a conservative estimate I see a drop from 7-8s cold start time to 5-6s, so ~25% saved. Here's how to test cold-start time on Linux:

```
sync ; echo 3 | sudo tee /proc/sys/vm/drop_caches; time ipython -c pass
```

This is an alternative to doing it with lazyasd (discussed on #9908). I looked into the code for that, but I was reminded of the maxim that debugging is harder than writing code, so if you write the cleverest code you can, you're not smart enough to debug it. I don't want to end up debugging lazy-loading module issues.

The downside of this is that I'm copying part of the API of pygments; if that API changes, things using it might be broken until we copy the changes. However, pygments development seems to be [slow](https://bitbucket.org/birkenfeld/pygments-main/commits/all), so for the moment this does not seem like a significant risk. I would submit this change to pygments itself, but as even my tiny bugfix PR has had no attention in months, so it seems unlikely that it will go anywhere.

### DIFF
--- a/IPython/__init__.py
+++ b/IPython/__init__.py
@@ -40,6 +40,10 @@ See IPython `README.rst` file for more information:
 
 """)
 
+# Monkeypatch pygments.plugin to avoid loading pkg_resources, which is slow.
+from . import pygments_plugin
+sys.modules['pygments.plugin'] = pygments_plugin
+
 # Make it easy to import extensions - they are always directly on pythonpath.
 # Therefore, non-IPython modules can be added to extensions directory.
 # This should probably be in ipapp.py.
--- a/IPython/pygments_plugin.py
+++ b/IPython/pygments_plugin.py
@@ -1,0 +1,70 @@
+"""
+Modified copy of pygments.plugin; doesn't load pkg_resources until its functions
+are called.
+
+    pygments.plugin
+    ~~~~~~~~~~~~~~~
+
+    Pygments setuptools plugin interface. The methods defined
+    here also work if setuptools isn't installed but they just
+    return nothing.
+
+    lexer plugins::
+
+        [pygments.lexers]
+        yourlexer = yourmodule:YourLexer
+
+    formatter plugins::
+
+        [pygments.formatters]
+        yourformatter = yourformatter:YourFormatter
+        /.ext = yourformatter:YourFormatter
+
+    As you can see, you can define extensions for the formatter
+    with a leading slash.
+
+    syntax plugins::
+
+        [pygments.styles]
+        yourstyle = yourstyle:YourStyle
+
+    filter plugin::
+
+        [pygments.filter]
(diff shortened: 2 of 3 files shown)


scopatz:
I really wonder if there is anything that we can do to speed up pygments development...

takluyver:
The idea of doing a friendly fork and trying to get permission to upload new releases as 'pygments' has occurred to me. But I'm not eager to be maintaining yet another package, and most of the pull requests open are tweaks to the lexers for a wide variety of languages, which I imagine are hard for one person to review. Splitting them out as separate pieces would make some things easier, but then it needs integration work; those included batteries are nice from a user perspective.

Maybe we could identify a collection of key fixes (like this one) and try to encourage a one-time new release? Feels a bit rude to be pleading that *my bug's important*, though.

Or maybe identify other projects using pygments and try to assemble a maintenance team? I believe there are quite a few projects that depend on it.

scopatz:
Yeah, I am not super fond of maintaining a new project either.  However, a maintenance team is not a bad idea. Either forking or moving development over would be fine with me.

Carreau:
+1 for gathering a team.

I know there are a few changes that `@jonathanslenders` could use for PTK, [like color naming, he asked me to do and are still on my TODO list]

If we do a "friendly fork", I would suggest the name `dyes`.

Changing names could also be a reason to drop some legacy compat/behavior.

scopatz:
Well, we should probably start by just asking if the maintainer would be open to having more maintainers. `dyes` sounds good to me :)

Carreau:
> Well, we should probably start by just asking if the maintainer would be open to having more maintainers. dyes sounds good to me :)

Well, my guess is that we should just pitch in and start replying to bug reports and send patches. They will likely not say no to help.

scopatz:
yeah that is even better, assuming they respond.

takluyver:
I have made a pull request on pygments with this change:

https://bitbucket.org/birkenfeld/pygments-main/pull-requests/690/delay-loading-pkg_resources-until-it-is/diff

minrk:
Cool, thanks @takluyver! Let's give the upstream PR some time and then decide. I'd be happy if we can rely on "upgrade pygments" vs shipping a monkeypatch.

takluyver:
The PR has been merged into pygments. I'll leave this open for now, but hopefully there will be a new release soon.

minrk:
@takluyver thanks! We can merge this one if/when it turns out a pygments release isn't going to happen in a reasonable time frame.

takluyver:
Happily, pygments 2.2 has been released with this fix. :-)

Issue #10137, opened 2017-01-08
Importing pyplot causes bouncing dock icon on mac


ngoldbaum:
This is illustrated by the following gif:

![image](http://g.recordit.co/ARO7Maic0K.gif)

Importing pyplot in the normal python repl causes the python rocket icon to appear but not bounce distractingly. Importing it inside IPython causes the icon to bounce annoyingly for almost a minute.

This is using a framework build of Python from Homebrew.

tacaswell:
which backend are you using?  Does running `plt.ion()` or `%matplotlib` help?

My knee-jerk / making this up as I go guess is: When IPython de-coupled from readline, it also decoupled from the `PyOS_InputHook` machinery.   In the plain python case, the GUI is starting up (starting the bouncing), the GUI code installs the hook, and when it finishes starting up some how phones home to let the rocket know to stop bouncing because the gui main loop runs intercolated with the repl.  In IPython, that hook is still installed, but does not matter because IPython is running through prompt toolkit.  If you start up the IPython event loop integration the GUI will phone home and stop the bouncing.

Or I could be wildly off base.

ngoldbaum:
This is using the `MacOSX` backend. Using `plt.ion() or `%matplotlib` after the `pyplot` import doesn't stop the bouncing and using `%matplotlib` before doing the pyplot import just causes the icon to start bouncing.

tacaswell:
Does it happen with the tk backend?

ngoldbaum:
Nope, no rocket icon at all with `TkAgg`.

Carreau:
I'm going to stay that's linked to #9312 
I'm going to try to understand it. Ping  @minrk.

Carreau:
I'm going to say that [this](https://github.com/ipython/ipython/blob/efc9e7678bdb10b47f60289f47c355d6b8be4048/IPython/terminal/pt_inputhooks/osx.py#L134-L135):

```python
    if not window_count:
        return
```

Need to be removed as the eventloop integration need to run even if there is no windows... it appears to be working for me. 
Alternative would be to make sure  `inputhook(context)` does at least one loop even if no windows.. so along:

```python
    global at_least_once
    if not window_count and not at_least_once:
        at_least_once = True
        return
```
@ngoldbaum would you mind trying ? 
Also I suppose it stop bouncing as soon as you have your first window ?

Carreau:
[...]

PR #10150, opened 2017-01-12
Always run the OS X loop even if no windows.

Carreau:
Otherwise this can trigger infinite Python-Icon-In-Dock bouncing.
See #10137. I will guess that this is because application on OS X may
not have windows and still need to process events.

It may be that an alternative is to run the loop only once the first
time, but I'm unsure.

---- 

Ping @minrk : I have no clue what I am doing.

### DIFF
--- a/IPython/terminal/pt_inputhooks/osx.py
+++ b/IPython/terminal/pt_inputhooks/osx.py
@@ -127,12 +127,6 @@ def _stop_on_read(fd):
 def inputhook(context):
     """Inputhook for Cocoa (NSApp)"""
     NSApp = _NSApp()
-    window_count = msg(
-        msg(NSApp, n('windows')),
-        n('count')
-    )
-    if not window_count:
-        return
     _stop_on_read(context.fileno())
     msg(NSApp, n('run'))
     if not _triggered.is_set():


minrk:
Thanks for working on this one! I'll have a look tomorrow. I'll have to refresh my memory on the last few bugs we worked out on this stuff, in case this is trading a new bug for an old one.

minrk:
Appears to work, and isn't related to the previous bugfixes on this tricky one. Thanks!

Carreau:
Thanks. Let's try to autobackport.

@meeseeksdev backport to 5.x

minrk:
Cool! Can the auto-backport copy the milestone, as well?

Carreau:
> Cool! Can the auto-backport copy the milestone, as well?

It could, I added an [issue](https://github.com/MeeseeksBox/MeeseeksDev/issues/16).

asmeurer:
I was using this inputhook in my own prompt_toolkit application. Before this patch, it worked just fine. After this patch, it causes my application to be completely unresponsive to all keyboard input.

asmeurer:
Your suggested code from the commit message works for me. I can't reproduce the original bouncing Dock issue with the Anaconda Python, so I can't say how it affects that.

Carreau:
Hum, sorry about that. We would need to better understand all the interactions and when to do what. Event loop are hard.

asmeurer:
Oh, scratch that your commit message fix doesn't work (didn't save the file correctly 😳).

Carreau:
> Oh, scratch that your commit message fix doesn't work (didn't save the file correctly 😳).

it's ok, GitHub have a delete comment button if needed.

asmeurer:
Well I'm not ashamed of making a mistake. And it already sent the email, so it would just be confusing to you.

minrk:
@asmeurer do you have an example that doesn't behave right?

asmeurer:
Even a dead simple prompt-toolkit shell reproduces the issue:

```py
#!/usr/bin/env pythonw

# Basic REPL
from traceback import print_exc

from prompt_toolkit.shortcuts import prompt, create_eventloop

from IPython.terminal.pt_inputhooks.osx import inputhook

_globals = _locals = globals().copy()

while True:
    command = prompt(">>> ", eventloop=create_eventloop(inputhook))
    try:
        res = eval(command, _globals, _locals)
        print(repr(res))
    except SyntaxError:
        try:
            exec(command, _globals, _locals)
        except BaseException as e:
            print_exc()
    except BaseException as e:
        print_exc()
```

With the latest IPython, the shell hangs at startup. It doesn't accept any input, and you have to kill it with `kill`. 

If you use an IPython before this change, it works perfectly. Here's how I've been testing "perfectly", running this at the command line:

```py
>>> import matplotlib
>>> matplotlib.interactive(True)
None
>>> import matplotlib.pylab as plt
>>> plt.plot([1, 2])
[<matplotlib.lines.Line2D object at 0x10e09db70>]
>>>
```

Things it should do:

- The plot should popup immediately, with a line.
- The prompt should not hang (it should show the next `>>>` immediately, before closing the plot). 
- The plot should be interactive (try clicking and using one of the toolbar buttons).
- Closing the plot should work. 

Ideally, I could just import this code directly from IPython (for now, I have copied the code without this change), although I understand that IPython only needs to support IPython. I realise that it works fine in IPython itself, but I'd like at least to understand why. Plus it sounds like you weren't really sure about this change to begin with, so maybe there is something wrong with it, even from the IPython side.

Issue #10113, opened 2017-01-02
Print during import screws up input prompt


vidartf:
Steps to reproduce:

1. Create a module which prints something during import, e.g. a file `hello_world.py`:

```python
print('Hello world!')
```

2. Try to use tab-completion for sub-modules in an import statement

```python
In [1]: import hello_world.<hit tab>
```

For this minimal example there wouldn't be any submodules, but it will autocomplete `py` to match the filename. What happens is that the print statement during import interferes with the input prompt.

Example output:
```
>ipython
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)]
Type "copyright", "credits" or "license" for more information.

IPython 5.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import hello_world.Hello world!
py
```

vidartf:
Note: After the tab completion, the cursor will be on the wrong line (at least for Windows' `cmd` terminal).

takluyver:
See #10153

PR #10153, opened 2017-01-13
Use prompt_toolkit's patch_stdout_context when finding completions

takluyver:
Closes gh-10113


Carreau:
@meeseeksdev backport to 5.x

Issue #9829, opened 2016-07-31
importdenier meta path hook has no effect


anntzer:
IPython installs a meta path hook at https://github.com/ipython/ipython/blob/master/IPython/external/qt_loaders.py#L50 to prevent multiple incompatible Qt bindings to be loaded at the same time; however this has no effect as can be seen by importing first PyQt4.QtCore and then PyQt5.QtCore (the incompatible imports are caught by PyQt itself).

This is because the hook is at the _end_ of `sys.meta_path` (`sys.meta_path.append(ID)`); moving it to the beginning makes it work properly.  However, given that PyQt itself raises a `RuntimeError` when trying to load incompatible versions, I'm not even sure this hook is really needed at all.

takluyver:
I think the import denier may be necessary if you try to load PySide and PyQt, as I'm not sure that either of those checks for the other.

I guess this code was originally developed with Python 2, where `sys.meta_path` is empty by default, and anything added to it takes precedence over the normal import mechanisms. On Python 3, `sys.meta_path` includes the normal import mechanisms, so you're right; we should be prepending to it.

anntzer:
Indeed they don't (I just checked)... hum that must lead to interesting bugs :-)

Carreau:
Is that really critical ? Or can I mark it for 6.0 only ?

anntzer:
Probably not that important, if no one had noticed it so far...

minrk:
Since it's a "better error message, but still an error" situation, it shouldn't be critical.

takluyver:
I've set this as 5.2 - I'd say it's worth backporting once we fix it, but it's not important enough to hold up any release over.

takluyver:
#10154 fixes this

PR #10154, opened 2017-01-13
Move ImportDenier to the front of sys.meta_path

takluyver:
Closes gh-9829


Carreau:
Should that be a 5.x thing ?

PR #10156, opened 2017-01-13
relax help_all output check

minrk:
traitlets may change small details of section headings. Don't worry about them too much


takluyver:
@meeseeksdev backport

PR #10157, opened 2017-01-13
Fix detection of complete input with line continuations inside a multiline string

### LINKED ISSUE #9841: End of multiline string that contains a backslash-terminated line is not recognized

takluyver:
I thought that tokenize accepted multiple lines of input as a single 'line' from the readline callback, and it does in most cases, but in this case the behaviour makes a difference.

Closes #9841.

Alternative to #9899.

### DIFF
--- a/IPython/core/inputtransformer.py
+++ b/IPython/core/inputtransformer.py
@@ -120,25 +120,18 @@ class TokenInputTransformer(InputTransformer):
     """
     def __init__(self, func):
         self.func = func
-        self.current_line = ""
-        self.line_used = False
+        self.buf = []
         self.reset_tokenizer()
-    
+
     def reset_tokenizer(self):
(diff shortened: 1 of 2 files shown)


Carreau:
can you rebase/repush. Travis is unhappy because GitHub was showing unicorns, and refuse to get restared.

takluyver:
Rebased, Travis is happy now. Appveyor's still going, but as it doesn't run the tests, we're basically just heating up someone's datacentre there.

Carreau:
Do you want to attempt a backport ?

takluyver:
Let's have a go!

@meeseeksdev backport to 5.x

PR #10164, opened 2017-01-17
Run appveyor tests

srinivasreddy:
Do not review yet

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,7 +41,9 @@ install:
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
   - "%CMD_IN_ENV%  pip install setuptools>=18.5 --upgrade"
   - "%CMD_IN_ENV%  pip install nose coverage"
-  - "%CMD_IN_ENV%  pip install -r requires.txt"
+  - "%CMD_IN_ENV%  pip install .[test]"
+  - "%CMD_IN_ENV%  mkdir results"
+  - "%CMD_IN_ENV%  cd results"
 test_script:
-  - "%CMD_IN_ENV%  nosetests ."
+  - "%CMD_IN_ENV%  iptest --coverage xml"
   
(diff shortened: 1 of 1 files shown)


takluyver:
IPython has a specific `iptest` command to run its tests - it uses nose, but does a bit of extra setup beforehand.

appveyor.yml
```diff
@@ -40,7 +40,8 @@ init:
 install:
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
   - "%CMD_IN_ENV%  pip install setuptools>=18.5 --upgrade"
-  - "%CMD_IN_ENV%  pip install nose coverage ipython[test] --upgrade"
+  - "%CMD_IN_ENV%  pip install nose coverage"
+  - "%CMD_IN_ENV%  pip install -r requires.txt"
```
takluyver: This file doesn't exist. I think what you want is `pip install .[test]`

appveyor.yml
```diff
@@ -40,7 +40,8 @@ init:
 install:
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
   - "%CMD_IN_ENV%  pip install setuptools>=18.5 --upgrade"
-  - "%CMD_IN_ENV%  pip install nose coverage ipython[test] --upgrade"
+  - "%CMD_IN_ENV%  pip install nose coverage"
+  - "%CMD_IN_ENV%  pip install .[test]"
 test_script:
-  - "%CMD_IN_ENV%  python -c 'True'"
+  - "%CMD_IN_ENV%  nosetests ."
```
takluyver: This should use our `iptest` command rather than `nosetests`. You may need to cd to a different directory before running it. Have a look at `.travis.yml` for comparison.

takluyver (changes requested):

takluyver:
The new failure is 'Don't run iptest from the IPython source directory' - you need to create a directory somewhere (e.g. `mkdir results` and then `cd results`). This makes sure that it's running with the installed copy of IPython.

takluyver (approved):

takluyver:
Thanks, that appears to be working :-)

PR #10179, opened 2017-01-18
remove unicode specifier

srinivasreddy:


Carreau:
Thanks !

Don't go cleaning the all codebase of u prefix though. They have no overhead (unlike if statement) and that would screw up blame. 

Do you want to try to make a more involved bugfix of feature ?

Issue #10170, opened 2017-01-18
TAB-completion exception with modules


jdemeyer:
In IPython (command-line or notebook), type
```
from ipywidgets. import interactive
```
and now press [TAB] when the cursor is after the dot. This gives
```
[IPKernelApp] ERROR | Exception in message handler:
Traceback (most recent call last):
  File "/.../python2.7/site-packages/ipykernel/kernelbase.py", line 228, in dispatch_shell
    handler(stream, idents, msg)
  File "/.../python2.7/site-packages/ipykernel/kernelbase.py", line 425, in complete_request
    matches = self.do_complete(code, cursor_pos)
  File "/.../python2.7/site-packages/ipykernel/ipkernel.py", line 257, in do_complete
    txt, matches = self.shell.complete('', line, line_cursor)
  File "/.../python2.7/site-packages/IPython/core/interactiveshell.py", line 1986, in complete
    return self.Completer.complete(text, line, cursor_pos)
  File "/.../python2.7/site-packages/IPython/core/completer.py", line 84, in comp
    text, matches =  complete(*args, **kwargs)
  File "/.../python2.7/site-packages/IPython/core/completer.py", line 1193, in complete
    custom_res = self.dispatch_custom_completer(text)
  File "/.../python2.7/site-packages/IPython/core/completer.py", line 1117, in dispatch_custom_completer
    res = c(event)
  File "/.../python2.7/site-packages/IPython/core/completerlib.py", line 253, in module_completer
    return module_completion(event.line)
  File "/.../python2.7/site-packages/IPython/core/completerlib.py", line 236, in module_completion
    return try_import(mod)
  File "/.../python2.7/site-packages/IPython/core/completerlib.py", line 163, in try_import
    m = getattr(m, module)
AttributeError: 'module' object has no attribute ''
```

takluyver:
I can't reproduce this at present on master; I'll keep trying later on.

Carreau:
> I can't reproduce this at present on master; I'll keep trying later on.

Bug report is on 2.7 and I can reproduce.

Carreau:
Ouch, that one is nasty. 

```
def complete(self, text=None, line_buffer=None, cursor_pos=None):
        """Find completions for the given text and line context.

        Note that both the text and the line_buffer are optional, but at least
        one of them must be given.

        Parameters
        ----------
          text : string, optional
            Text to perform the completion on.  If not given, the line buffer
            is split using the instance's CompletionSplitter object.

[...]

PR #10181, opened 2017-01-18
Don't fail when completing modules ending with dot.

Carreau:
Mitigate #10170 don't considered as a fix as it does not populate the
completion, just prevent crash.

 Consider as a Backport of #10180


Carreau:
@meeseeksdev backport to 5.x

ian-r-rose:
The robot always turns on it's maker.

Carreau:
I know right? I need to teach him how to behave.

Issue #8733, opened 2015-08-14
`ipython_directive` @savefig: should detect missing static html dir and not place output in source dir


stefanv:
1. The `ipython_directive`'s `@savefig` currently has its default output set relative to the static html directory.  If that directory is missing, it places the output _outside of the source tree_.  The build succeeds with a warning from Sphinx that it cannot find the image—which does not help to find the source of the problem.  Solution: raise an error/warning if the directive's output path is invalid.
2. `@savefig` stores its output in the source directory.  A better default would be to store it in `build/_static`.

minrk:
@stefanv can you test #8888? I think it fixes these things, but I'm not super experienced with this stuff.

stefanv:
That does not seem to work for me, but it may very well be that I am not testing correctly.  The repository I am testing against is

https://github.com/statlab/permute

Where I see:

```
internal padding after /home/stefan/akad/projekte/permute/doc/source/user/npc/kenya.rst:30: WARNING: image file not readable: user/npc/../../../build/html/_static/test.png
```

(after commenting out ``ipython_savefig_dir`` from the config)

minrk:
Thanks for testing, I'll use that repo to figure out what I'm not quite doing correctly.

Carreau:
@stefanv I do not seem to be able to reproduce using `permute` and coomenting the `ipython_savefig_dir` could sphinx-build have picked up a ipython savefig directive from another env ?

```
 git diff -U0
diff --git a/doc/source/conf.py b/doc/source/conf.py
index 4fb1ba1..a6bff13 100644
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -147 +147 @@ html_static_path = ['_static']
-ipython_savefig_dir = '../build/html/_static'
+# ipython_savefig_dir = '../build/html/_static'
```

```
$ make html
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v1.3.3
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 29 source files that are out of date
updating environment: 29 added, 0 changed, 0 removed
checking for /Users/bussonniermatthias/dev/permute/doc/source/permute.bib in bibtex cache... not found
parsing bibtex file /Users/bussonniermatthias/dev/permute/doc/source/permute.bib... parsed 10 entries
reading sources... [100%] user/two-sample
[...]

PR #10183, opened 2017-01-19
Create savefig_dir in Sphinx IPython directive

takluyver:
Closes gh-8733 (hopefully)


takluyver:
I currently prefer #10186 as a fix instead of this, as it doesn't produce warnings during the docs build.

Issue #9944, opened 2016-09-13
Graceful forced exit?


jdemeyer:
Is there a way for an **application** to signal to `ipython` that it should exit gracefully? My use case is installing a signal handler to force IPython to exit.

I would like something like this to work:
```
In [1]: import sys, signal

In [2]: def exit_now(*args): sys.exit()

In [3]: signal.signal(signal.SIGALRM, exit_now)
Out[3]: 0

In [4]: signal.alarm(2)
Out[4]: 0

In [5]:
```

It used to be that forcing an EOF on stdin would work, but `prompt_toolkit` does not handle that correctly anymore. See also https://github.com/jonathanslenders/python-prompt-toolkit/issues/392

takluyver:
`get_ipython().ask_exit()` is the official way, I think.

We've always caught SystemExit, and I presume that's because people sometimes `%run` code which uses `sys.exit()` to finish, while the user doesn't want to be booted out of the interactive prompt.

minrk:
Shouldn't we be doing a graceful exit on SIGTERM? I know we aren't, but it seems like we should.

jdemeyer:
> Shouldn't we be doing a graceful exit on SIGTERM?

That is what I am trying to do in https://github.com/sagemath/cysignals. However, [the trick](https://github.com/sagemath/cysignals/blob/master/src/cysignals/signals.pyx#L92) which used to work with IPython 4 no longer works now.

jdemeyer:
> We've always caught SystemExit, and I presume that's because people sometimes %run code which uses sys.exit() to finish, while the user doesn't want to be booted out of the interactive prompt.

I understand that and I am not asking that to change.

jdemeyer:
> get_ipython().ask_exit() is the official way, I think.

That works on the IPython part, but not on the `prompt_toolkit` part. Executing `ask_exit()` while `prompt_toolkit` is waiting for a prompt will not cause IPython to exit. So this looks more like a `prompt_toolkit` problem and not an IPython problem.

takluyver:
That's expected, I think - `ask_exit` sets a flag which we only check in our interaction loop,  not while PT is running the event loop for the prompt. However, I'm a bit confused about why you want to exit programmatically while there's a prompt displayed.

Maybe we should just be installing a signal handler for SIGTERM. How would that interact with what you're trying to do?

jdemeyer:
> Maybe we should just be installing a signal handler for SIGTERM

[...]

PR #10185, opened 2017-01-19
Exit terminal interface gracefully on signals

takluyver:
Closes gh-9944

This sets signal handlers for SIGTERM and SIGHUP while the prompt_toolkit prompt is active, letting it clean up the terminal before the process exits. When the prompt is not active, we restore the previous signals.

This can't work (as far as I know) if other code has set signal handlers at the C level. I don't think this is likely to come up often, but there's a new config option catch_signals=False which users can set to disable our manipulation of signal handlers.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -405,7 +405,7 @@ class TerminalInteractiveShell(InteractiveShell):
         self.ask_exit()
         self._eventloop.call_from_executor(lambda: self.pt_cli.set_return_value(None))
 
-    _end_signals = (signal.SIGTERM, signal.SIGHUP)
+    _end_signals = ()
 
     @contextmanager
     def _real_handling_signals(self):
@@ -434,6 +434,7 @@ class TerminalInteractiveShell(InteractiveShell):
     def init_signals(self):
         if self.catch_signals and os.name == 'posix':
             self._maybe_handling_signals = self._real_handling_signals
+            self._end_signals = (signal.SIGTERM, signal.SIGHUP)
 
     rl_next_input = None
 
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -1,6 +1,8 @@
 """IPython terminal interface using prompt_toolkit"""
 
+from contextlib import contextmanager
 import os
+import signal
 import sys
 import warnings
 from warnings import warn
@@ -190,6 +192,10 @@ class TerminalInteractiveShell(InteractiveShell):
         help="Highlight matching brackets .",
     ).tag(config=True)
 
+    catch_signals = Bool(True,
+         help="Catch SIGTERM and SIGHUP to exit gracefully",
+    ).tag(config=True)
+
     @observe('term_title')
     def init_term_title(self, change=None):
         # Enable or disable the terminal title.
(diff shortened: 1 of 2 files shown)


jdemeyer:
It would be nice if this would also close IPython gracefully:
```
In [1]: import os; f = open(os.devnull); os.dup2(f.fileno(), 0)
```
With this branch, this code goes into an infinite loop.

jdemeyer:
Instead of trying to save and restore signals (which is guaranteed to fail for C-level signal handlers), I would much prefer if the signal handler would be installed just once at startup.

jdemeyer:
Apart from the manipulation of the signal handlers, this code works fine for me and I would be happy to support this in `cysignals`.

takluyver:
I restore the default signal handler so that the default behaviour applies while user code is running. I tried installing it at the start and leaving it in place, but then SIGTERM during user code only exits once it would display the next prompt.

jdemeyer:
> I restore the default signal handler so that the default behaviour applies while user code is running.

As you say yourself, this doesn't quite work when signal handlers have been installed at the C level.

> I tried installing it at the start and leaving it in place, but then SIGTERM during user code only exits once it would display the next prompt.

I'd rather try to fix this instead...

takluyver:
It's easy enough to stop Python code by raising an exception, but if the main thread is inside some long running C code, I don't know of a good way to stop it on a signal from Python. I think setting it back to the default is the best option here, even if that means that people using C signal handlers have to disable our signal handling functionality.

I've brought this up on a Python issue here: http://bugs.python.org/issue13285

Carreau:
Added a change in the what's new.

Carreau:
I suggest we move forward with this even if it's not perfect. We can refine in 5.3 if needed.

jdemeyer:
> I suggest we move forward with this even if it's not perfect. We can refine in 5.3 if needed.

Please no! This is going to break `cysignals` badly.

jdemeyer:
Or at least make `catch_signal=False` the default so nothing changes for existing use cases.

jdemeyer:
Another alternative: make it easy to change `change_signal=False` at runtime such that other packages which need to play with signals can do that.

Carreau:
Ok, so then should we bump to 5.3 ? This is one of the 2 remaining issues for 5.2.

takluyver:
Bumping to 5.3 so we can think more about this.

Carreau:
This PR have conflicts now. :-(

Carreau:
@takluyver, thoughts about this one ?

takluyver:
I don't think anyone wants it. I wrote it attempting to address a need @jdemeyer brought up, but it apparently makes things worse for him, not better. I don't think we can just set  a Python signal handler and leave it, because a Python signal handler won't break out of compiled code which the user may have called.

Jeroen has implemented the machinery in Cysignals to restore OS-level signal handlers from Python, but we don't want a compiled dependency in IPython. And there doesn't seem to be much interest in getting that functionality into the Python `signal` module - though there isn't antipathy either, so maybe if someone made a pull request it might go somewhere.

Closing this for now, anyway.

jdemeyer:
In any case, I feel like this is mainly a `prompt_toolkit` issue (https://github.com/jonathanslenders/python-prompt-toolkit/issues/392)

PR #10186, opened 2017-01-19
Put savefig images back in source directory

takluyver:
Simplifying the logic a bit.

I think there was an issue with the way we were finding the source directory: when it included a docstring from code, the source directory was that of the code. This always uses the main source directory of the docs.

Closes gh-8733

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -126,6 +126,7 @@ Authors
 
 # Stdlib
 import atexit
+import errno
 import os
 import re
 import sys
@@ -358,9 +359,9 @@ class EmbeddedSphinxShell(object):
         source_dir = self.source_dir
         saveargs = decorator.split(' ')
         filename = saveargs[1]
-        # insert relative path to image file in source
-        outfile = os.path.relpath(os.path.join(savefig_dir,filename),
-                    source_dir)
+        # insert relative path to image file in source (as absolute path for Sphinx)
+        outfile = '/' + os.path.relpath(os.path.join(savefig_dir,filename),
+                                        source_dir)
 
         imagerows = ['.. image:: %s'%outfile]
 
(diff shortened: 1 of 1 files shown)


IPython/sphinxext/ipython_directive.py
```diff
@@ -358,9 +359,9 @@ def process_image(self, decorator):
         source_dir = self.source_dir
         saveargs = decorator.split(' ')
         filename = saveargs[1]
-        # insert relative path to image file in source
-        outfile = os.path.relpath(os.path.join(savefig_dir,filename),
-                    source_dir)
+        # insert relative path to image file in source (as absolute path for Sphinx)
+        outfile = '/' + os.path.relpath(os.path.join(savefig_dir,filename),
```
Carreau: `/` or `os.path.sep` ?

Carreau:
+1

minrk:
I think this should work.

fmaussion:
Yes, this solved our problem on read the docs. Thanks!

IPython/sphinxext/ipython_directive.py
takluyver: I believe `/` is right, because it's effectively a URL rather than a platform path. We write all the paths in rst files with forward slashes. I suspect this means it shouldn't be doing `os.path.join()` either, but I didn't want to change more than I needed to at the moment.

Carreau:
@meeseeksdev backport to 5.x

jcolonese:
This fails on Windows. Not sure but seems like Sphinx removes the "\\" from `os.path.join` and can't find files on `savefig_dir` when rendering html.

I have also set `ipython_savefig_dir` in `conf.py` and the images are saved in the right directory.

Using it right now for documenting a Windows app, so to make it work, I simply added
```python
if os.name == 'nt':
     outfile = outfile.replace('\\', '/')
```
Not sure if the best way.

Issue #6386, opened 2014-08-30
ipython.parallel in wsgi app


bsdis:
When importing IPython.parallel in a script running as a wsgi service, an error occurs:

[Sat Aug 30 21:36:52.586183 2014] [:error] [pid 7469] [remote 10.252.230.165:51802] mod_wsgi (pid=7469): Target WSGI script '/home/opta/www/env/pyramid.wsgi' cannot be loaded as Python module.
[Sat Aug 30 21:36:52.586274 2014] [:error] [pid 7469] [remote 10.252.230.165:51802] mod_wsgi (pid=7469): Exception occurred processing WSGI script '/home/opta/www/env/pyramid.wsgi'.
[Sat Aug 30 21:36:52.586316 2014] [:error] [pid 7469] [remote 10.252.230.165:51802] Traceback (most recent call last):
[Sat Aug 30 21:36:52.586338 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]   File "/home/opta/www/env/pyramid.wsgi", line 4, in <module>
[Sat Aug 30 21:36:52.586493 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]     application = get_app(ini_path, 'main')
[Sat Aug 30 21:36:52.586510 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]   File "/home/opta/www/env/lib/python3.3/site-packages/pyramid/paster.py", line 31, in get_app
[Sat Aug 30 21:36:52.586672 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]     global_conf=options)
[Sat Aug 30 21:36:52.586689 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]   File "/home/opta/www/env/lib/python3.3/site-packages/paste/deploy/loadwsgi.py", line 247, in loadapp
[Sat Aug 30 21:36:52.586988 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]     return loadobj(APP, uri, name=name, *_kw)
[Sat Aug 30 21:36:52.587005 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]   File "/home/opta/www/env/lib/python3.3/site-packages/paste/deploy/loadwsgi.py", line 272, in loadobj
[Sat Aug 30 21:36:52.587044 2014] [:error] [pid 7469] [remote 10.252.230.165:51802]     return context.create()
[...]

PR #10191, opened 2017-01-20
IOStream: Ignore missing attrs from `dir()`. #6386

cjwelborn:
In some cases calling `getattr()` on attributes found with
`dir()` will fail. This is seen in Issue #6386.

The attribute in that issue is `name`, which is not needed for
`IOStream` to work correctly. This fix will set the attribute
to `None` and move on.

Tests were added in `IPython.utils.tests.test_io.py` to catch
any regressions.

### DIFF
--- a/IPython/utils/io.py
+++ b/IPython/utils/io.py
@@ -37,7 +37,12 @@ class IOStream:
         def clone(meth):
             return not hasattr(self, meth) and not meth.startswith('_')
         for meth in filter(clone, dir(stream)):
-            setattr(self, meth, getattr(stream, meth))
+            try:
+                val = getattr(stream, meth)
+            except AttributeError:
+                pass
+            else:
+                setattr(self, meth, val)
 
     def __repr__(self):
         cls = self.__class__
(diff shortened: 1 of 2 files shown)


IPython/utils/io.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, stream, fallback=None):
         def clone(meth):
             return not hasattr(self, meth) and not meth.startswith('_')
         for meth in filter(clone, dir(stream)):
-            setattr(self, meth, getattr(stream, meth))
+            setattr(self, meth, getattr(stream, meth, None))
```
minrk: Would the right thing to do be skipping the `setattr` if no attr is found, rather than setting it as None?

e.g.

```python
try:
    attr = getattr(stream, meth)
except AttributeError:
    pass
else:
    setattr(self, meth, attr)
```

cjwelborn: I think you're right. I was going for "least invasive change" instead of what is truly right to do. I always do that when approaching large projects and I need to stop. I'm gonna fix it up.

minrk: No problem. Thanks for the PR!

cjwelborn:
Thanks @minrk.

Carreau:
@meeseeksdev backport to 5.x

PR #10193, opened 2017-01-23
shorten the url

srinivasreddy:


Carreau:
I don't think this is necessary, or a good idea. URL shortener have a tendency to disappear and  then you can't find the original thread. We already had to "unshorten" a number of url because the links were broken.

So -1, but I'll let other weight in.

takluyver:
Also -1 - URL shorteners are another thing that can break, and another thing that let someone track who's following a link. I don't have any particular problem with a long URL here.

PR #10195, opened 2017-01-23
remove dict() function call and replace with dict literal

srinivasreddy:


Carreau:
Thanks @srinivasreddy. We tend to accept "cleanup PR" too much, for some reasons. 

1) It will screw up git-blame (which can be crucial in narrowing down a bug) 
2) it can make backport more difficult
3) even if it seem trivial and obvious good, it may introduce bugs. 

For example `dict()` and `{}` have a slightly different meaning. One is that `dict()` is looked up in `locals()`, `globals()`  then `builtins`, and we _may_ override it to do some smart things (like make them an eventful namespace).  More subtil is that since Python 3.6 literal dict keep source-order insertion (as an implementation detail ) so we may end up relying on a feature which is not one. 

In general I understand that "Code cleanup" might be a way to get familiar with the code, but itis not the most helpful contribution. If you want to learn more see the following video which is really good.  
https://www.youtube.com/watch?v=wf-BqAjZb8M

I would suggest you try so actually tackle issues like the one which [need contributors](https://github.com/ipython/ipython/issues?q=is%3Aopen+is%3Aissue+label%3A%22Need+Contributor%22), then there are [bugs left for a 5.2](https://github.com/ipython/ipython/issues?q=is%3Aopen+is%3Aissue+milestone%3A5.2), and then [all left for 6.0](https://github.com/ipython/ipython/issues?utf8=%E2%9C%93&q=is%3Aopen%20is%3Aissue%20milestone%3A6.0). These will teach you more, be more pleasant, and have a bigger impact than "code cleaning". 

This one is small enough so I'm going to review and merge it, but for you I suggest you dig into actual code, and if you don't understand feel free to ask.

Carreau:
Thanks !

PR #10198, opened 2017-01-24
Correctly deprecate limit_to_all

Carreau:
At least emit a warning. 

Should likely be backported to 5.x.


Carreau:
Ok I don't like to do this, but i'm going to self merge if there is no reviews.

takluyver:
@meeseeksdev backport to 5.x

PR #10202, opened 2017-01-26
Add a ipywidgets line to install

philippegr:
To prevent an Error message 
https://github.com/jupyter/help/issues/126


takluyver:
I think this distracts from what is actually needed to install the kernelspec. The error message about widgets is not a problem.

Carreau:
That probably need to be in the ipywidgets docs, not here. As I believe ipywidgets will not be the only package affected by this. 

But adding this kind of information is indeed useful.

philippegr:
I agree with you, this is probably not the place to address the case I was describing in the issue (1 Jupyter install, several kernel in conda environment). I will thus close the pull request.

PR #10207, opened 2017-01-27
Fix deactivation of embedded instance.

### LINKED ISSUE #9761: IPython 5 - Embeded shell not reopening once exited
I recently updated to IPython 5 and am observing some strange behavior with the embedded shell functionality. If ipshell is called more than once in the code, once the first embedded shell is exited, the rest are also exited without allowing for interactivity.

So for example, from the [IPython 5 documentation on embedding](https://ipython.readthedocs.io/en/stable/interactive/reference.html#embedding), the sample code provided in embed_class_long does not work as intended. Once the first instance of ipshell is exited, IPython does not stop again when it is called. Below is an excerpt of the code from the documentation to highlight the issue.

``` python
from __future__ import print_function
from traitlets.config.loader import Config
cfg = Config()
# First import the embeddable shell class
from IPython.terminal.embed import InteractiveShellEmbed

ipshell = InteractiveShellEmbed(config=cfg,
                       banner1 = 'Dropping into IPython',
                       exit_msg = 'Leaving Interpreter, back to program.')

ipshell2 = InteractiveShellEmbed(config=cfg,
                        banner1 = 'Second IPython instance.')

print('\nHello. This is printed from the main controller program.\n')

ipshell('***Called from top level. '
        'Hit Ctrl-D to exit interpreter and continue program.\n'
        'Note that if you use %kill_embedded, you can fully deactivate\n'
        'This embedded instance so it will never turn on again')

print('\nBack in caller program, moving along...\n')

def foo(m):
    s = 'spam'
    ipshell('***In foo(). Try %whos, or print s or m:')
    print('foo says m = ',m)

# Some calls to the above functions which will trigger IPython:
print('Main program calling foo("eggs")\n')
foo('eggs')

ipshell2()
```

On my machine, when this runs, I drop into the first ipshell instance with the "Called from the top level" message. I then type exit to leave this embedded shell. I get the exit message as intended, get the print statement "Back in caller program, moving along..." and the statement "Main program calling foo("eggs")" as intended. Then on my shell, I get the "Dropping into IPython" message which is the banner for the first instance of ipshell, but IPython does not stop. Without embedding a shell, the program continues by printing the  "In foo()" statement, as well as the "Leaving Interpreter" exit message. Finally the script gets to ipshell2() and correctly opens a the second instance. 

Before in IPython 4.x, when foo('eggs') was run, a shell would be embedded. The issue is also a problem when used in for loops. If ipshell is called in a for loop and subsequently exited, all the ipshell calls afterwards will be skipped, which is annoying when you want to stop in every iteration of the for loop to inspect variables.

Here is some system information in case it is relevant.

```
'commit_hash': u'1ba246d',
'default_encoding': 'cp437',
'ipython_path': 'C:\\Anaconda\\lib\\site-packages\\IPython',
'ipython_version': '5.0.0',
'os_name': 'nt',
'platform': 'Windows-7-6.1.7601-SP1',
'sys_executable': 'C:\\Anaconda\\python.exe',
'sys_platform': 'win32',
'sys_version': '2.7.11 |Anaconda custom (64-bit)| (default, Feb 16 2016, 09:58:36) [MSC v.1500 64 bit (AMD64)]'}
```

Carreau:
While deactivation of full instances make some sens, the most common
behavior user expect when creating an instance and using
``kill_embedded`` is deactivation of the current call location.

Implement the expected on by default, add options to get previous
behavior, and add flag for direct exit and no-confirm

Fix #9761

### DIFF
--- a/IPython/terminal/embed.py
+++ b/IPython/terminal/embed.py
@@ -10,6 +10,7 @@ import sys
 import warnings
 
 from IPython.core import ultratb, compilerop
+from IPython.core import magic_arguments
 from IPython.core.magic import Magics, magics_class, line_magic
 from IPython.core.interactiveshell import DummyMod, InteractiveShell
 from IPython.terminal.interactiveshell import TerminalInteractiveShell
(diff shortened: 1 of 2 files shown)


Carreau:
@takluyver ? @minrk ?

IPython/terminal/embed.py
```diff
@@ -25,22 +26,68 @@ class KillEmbeded(Exception):pass
...
+        print(args)
+        if args.instance:
+            # let no ask
+            if not args.yes:
+                kill = ask_yes_no(
+                    "Are you sure you want to kill this embedded instance? [y/N] ", 'n')
+            else:
+                kill = True
+            if kill:
+                self.shell._disable_init_location()
```
takluyver: Haven't you got these back to front? This is the `instance` branch, but it looks like it's killing by location (and vice versa).

takluyver: Also, since you've added a distinction between call location and init location, do we need ways to kill by both? Or is that starting to get too arcane and fiddly?

takluyver: Aha, I see - setting `shell.embedded_active = False` is equivalent to killing by call location, because of the property access. I'd forgotten about that bit.

IPython/terminal/embed.py
```diff
@@ -77,44 +124,56 @@ class InteractiveShellEmbed(TerminalInteractiveShell):
...
+        InteractiveShellEmbed._inactive_locations.add(self._init_location_id)
 
     @embedded_active.setter
     def embedded_active(self, value):
-        if value :
+        if value:
             if self._call_location_id in InteractiveShellEmbed._inactive_locations:
-                InteractiveShellEmbed._inactive_locations.remove(self._call_location_id)
+                InteractiveShellEmbed._inactive_locations.remove(
+                    self._call_location_id)
```
takluyver: BTW, sets have a `.discard()` method which is like `.remove()` but doesn't raise if the item isn't in the set. So instead of checking `if a in b: b.remove(a)`, we can just do `b.discard(a)`.

Carreau: Oh ! Sweet ! Thanks !

IPython/terminal/embed.py
Carreau: yeah, it's a bit weird, killing by instance by setting an attribute on the instance does not really help if instances are created in a loop. So I do keep track of instance location and call location. 

Honestly I shouldn't have make the "embedded_active" a property and keep all these details private... but backward compatibility.  The worst case I'm trying to cover is :

```
for loop:
    ipembed = InteractiveShellEmbed()
    if option:
         ipembed()
    elseif:
         ipembed()
    else:
         ipembed()
````

You can either kill the 3 call locations or kill a single instance location.

I agree that the "instance location" is abit weirdish, but that's the behavior of 5.0 and 5.1, so I leave it as an option.


So I think I got the logic right, and I think we should keep both.

File: IPython/terminal/embed.py [hunk 3]
```diff
@@ -311,7 +379,8 @@ def embed(**kwargs):
         cls.clear_instance()
     frame = sys._getframe(1)
     shell = InteractiveShellEmbed.instance(_call_location_id='%s:%s' % (frame.f_code.co_filename, frame.f_lineno), **kwargs)
```
Carreau: indeed.

Carreau:
comments should be addressed (except the call-location/init-location thing where I don't think anything is needed)

takluyver:
Thanks!

@meeseeksdev backport to 5.x

PR #10220, opened 2017-01-30
Get GitHub stats since 5.1.0, and deduplicate mails in .mailmap

Carreau:
(forward port of 94209687855db35455b84415fcc460d559585889)


willingc:
Love the recognition of contributors :heart: :cake:

Carreau:
> Love the recognition of contributors

We do our best. 

One possible improvement to the script would be to extract/mark first time contributors. 
Also, this does not include contributors that did not contribute code which can be improved as well. 
We could try to pull the issues each PR close and count the authors /contributors in these issues.

Issue #10222, opened 2017-01-31
Tab completions error


anceschialberto:
Hi, I'm using python 2.7.13 with ipython 5.2.0 and ipdb 0.10.2 in a virtualenv. OS is windows 10.
When I run my script in the system shell and stop my code with `ipdb.set_trace()` and I try to use completions with tab, I immediately encounter this error:

```
Traceback (most recent call last):
  File "c:\python27\Lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "c:\python27\Lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "C:\web\findomestic-etl\.venv\lib\site-packages\prompt_toolkit\interface.py", line 840, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "C:\web\findomestic-etl\.venv\lib\site-packages\IPython\terminal\ptutils.py", line 45, in get_completions
    with self.shell.pt_cli.patch_stdout_context():
AttributeError: 'NoneType' object has no attribute 'pt_cli'
```

Any idea? Thank you in advance.

takluyver:
Sorry, I think that just got broken in 5.2. #10223 should fix it.

PR #10223, opened 2017-01-31
Fix error tab completing in the debugger

takluyver:
I think I broke this in 5.2.

Closes gh-10222

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -20,13 +20,16 @@ class TerminalPdb(Pdb):
         def get_prompt_tokens(cli):
             return [(Token.Prompt, self.prompt)]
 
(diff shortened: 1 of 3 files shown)


minrk:
Do you want to do a 5.2.1 shortly?

takluyver:
I wouldn't say I want to, but yes, I should.

takluyver:
@meeseeksdev backport to 5.x

takluyver:
It's released, but the upload to archive.ipython.org failed; I think the ssh key for this system isn't authorized on the server.

Carreau:
> It's released, but the upload to archive.ipython.org failed; I think the ssh key for this system isn't authorized on the server.

Uploaded for you.

PR #10228, opened 2017-02-01
Remove TemporaryDirectory class

srinivasreddy:
Remove TemporaryDirectory class as it was required for < Python 3.2


IPython/utils/tempdir.py
```diff
@@ -1,98 +1,8 @@
-"""TemporaryDirectory class, copied from Python 3.2.
-
-This is copied from the stdlib and will be standard in Python 3.2 and onwards.
```
takluyver: The module should still have a docstring. It actually grew a couple of other classes which are not in the stdlib, and those should be mentioned in the docstring too.

srinivasreddy: Done.

IPython/utils/tempdir.py
```diff
@@ -1,98 +1,14 @@
...
-                    try:
-                        self._remove(fullname)
-                    except self._os_error:
-                        pass
-            try:
-                self._rmdir(path)
-            except self._os_error:
-                pass
-
+from tempfile import TemporaryDirectory
```
takluyver: And can you leave a comment explaining that this is here in case things that need to work with earlier versions still import it. Thanks!

srinivasreddy: I am of the opinion that its not needed  because `TemporaryDirectory` is being sub-classed [here  ](https://github.com/ipython/ipython/blob/master/IPython/utils/tempdir.py#L127) . What do you think?  Thanks.

takluyver: Fair enough, makes sense.

Issue #10229, opened 2017-02-01
Crash when doing "import math; math.sin?"


jdemeyer:
In the IPython session below, I executed
```
%config Application.verbose_crash=True
```
and
```
import math; math.sin?
```
When doing this, IPython crashes.

Below, you will see ```In [3]: import math; math.sin?``` but I have no idea where that came from. That was just displayed automatically.

```
Python 2.7.13 (default, Jan 25 2017, 11:01:10) 
Type "copyright", "credits" or "license" for more information.

IPython 5.2.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: %config Application.verbose_crash=True

In [2]: import math; math.sin?
Object `math.sin` not found.

In [3]: import math; math.sin?
---------------------------------------------------------------------------
---------------------------------------------------------------------------
AssertionError   Python 2.7.13: /usr/local/src/sage-config/local/bin/python
                                                   Wed Feb  1 10:17:47 2017
A problem occurred executing Python code.  Here is the sequence of function
calls leading up to the error, with the most recent (innermost) call last.
/usr/local/src/sage-config/local/bin/ipython in <module>()
      1 #!/usr/local/src/sage-config/local/bin/python
      2 
      3 # -*- coding: utf-8 -*-
      4 import re
      5 import sys
      6 
      7 from IPython import start_ipython
      8 
      9 if __name__ == '__main__':
     10     sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
---> 11     sys.exit(start_ipython())
        global sys.exit = <built-in function exit>
        global start_ipython = <function start_ipython at 0x7fcd69088e60>

/usr/local/src/sage-config/local/lib/python2.7/site-packages/IPython/__init__.pyc in start_ipython(argv=None, **kwargs={})
    104     This is a public API method, and will survive implementation changes.
    105     
    106     Parameters
    107     ----------
    108     
    109     argv : list or None, optional
    110         If unspecified or None, IPython will parse command-line options from sys.argv.
    111         To prevent any command-line parsing, pass an empty list: `argv=[]`.
    112     user_ns : dict, optional
    113         specify this dictionary to initialize the IPython user namespace with particular values.
    114     kwargs : various, optional
    115         Any other kwargs will be passed to the Application constructor,
    116         such as `config`.
[...]

PR #10231, opened 2017-02-01
Fix set_next_input with prompt_toolkit 1.0.10

takluyver:
Closes gh-10229

Please don't merge this just yet.

@jonathanslenders I discovered after updating that we can no longer set text in the buffer with a `pre_run` function, because `pre_run_callables` reset the buffer immediately after that. I've worked around this by adding another callable to `pre_run_callables` which sets the text that I want, but I don't know if this is the right way to do it, or whether it works with previous releases of prompt_toolkit.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -12,6 +12,7 @@ from IPython.utils.terminal import toggle_set_term_title, set_term_title
 from IPython.utils.process import abbrev_cwd
 from traitlets import Bool, Unicode, Dict, Integer, observe, Instance, Type, default, Enum, Union
 
+from prompt_toolkit.document import Document
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @takluyver,
I didn't realize you were relying on this.
This merge request is correct, but it will indeed not work for earlier versions of prompt_toolkit.
You may want to test for ``hasattr(self.pt_cli, 'pre_run_callables')``.
What is actually the use case for this? Why are you setting a text?

takluyver:
Thanks, I've pushed a fix that only uses pre_run_callables if the attribute is there, and otherwise just sets the document directly.

There are a few features in IPython that put some text at the next prompt; probably the most obvious is the `%recall` magic - `%recall 1-3` will get inputs 1, 2 and 3 from your history, concatenate them and put them at the next prompt. They all go through a method `shell.set_next_input()`.

takluyver:
We're getting a number of bug reports that I believe this would fix, if someone would like to review it.

takluyver:
@meeseeksdev backport to 5.x

PR #10239, opened 2017-02-03
IPython/lib/editorhooks.py: wait for process even if wait=False

segevfiner:
The wait parameter is meant to add a prompt before returning from
the hook for editors that exit immediatly (fork/CreateProcess) but
it accidently prevented waiting at all for the process when it was
False. I think it was meant to be `not wait and proc.wait()` but we
might as well wait for the process in the `wait=True` case anyhow. It's
less confusing.

It would be nice if this is backported to 5.x.


takluyver:
Thanks, this seems reasonable.

takluyver:
@meeseeksdev backport to 5.x

PR #10241, opened 2017-02-03
Upload to PyPI in release upload script

takluyver:
There's no obvious reason that this should be a separate manual step


Carreau:
@meeseeksdev backport to 5.x

PR #10244, opened 2017-02-03
Protect against jedi failing to get the type of a completion

### LINKED ISSUE #10243: Jedi: Protect against crash on property access
Jedi Completions object can raise on attribute access (eg `completion.type`).

Completer.py ~ line 1520 should protect arount that:

```
       for jm in jedi_matches:
            delta = len(jm.name_with_symbols) - len(jm.complete)
            try:
                yield Completion(start=offset - delta, end=offset, text=jm.name_with_symbols, type=jm.type, _origin='jedi')
            except Exception:
                print('crash with jm:',jm)

````

Reproducible with 

```
## foo.py
def foo():pass
```

```
In[1]: from foo import foo
In[2] foo.__co<tab>
```

Log an error when debug is on would be a plus.

### PR DESCRIPTION
**Pull request by Carreau**
Closes #10243

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1554,11 +1554,17 @@ class IPCompleter(Completer):
         iter_jm = iter(jedi_matches)
         if _timeout:
             for jm in iter_jm:
+                try:
+                    type_ = jm.type
+                except Exception:
+                    if self.debug:
+                        print("Error in Jedi getting type of ", jm)
+                    type_ = None
                 delta = len(jm.name_with_symbols) - len(jm.complete)
                 yield Completion(start=offset - delta,
                                  end=offset,
                                  text=jm.name_with_symbols,
-                                 type=jm.type,
+                                 type=type_,
                                  _origin='jedi')
 
                 if time.monotonic() > deadline:

### CODE REVIEW

File: IPython/core/completer.py [hunk 1]
```diff
@@ -1517,7 +1517,13 @@ def _completions(self, full_text: str, offset: int)->Iterator[Completion]:
 
         for jm in jedi_matches:
             delta = len(jm.name_with_symbols) - len(jm.complete)
-            yield Completion(start=offset - delta, end=offset, text=jm.name_with_symbols, type=jm.type, _origin='jedi')
+            try:
+                type_ = jm.type
+            except Exception:
```
takluyver: Should this catch a more specific error?

takluyver (approved):
I couldn't reproduce the error, but this looks OK overall.

takluyver:
I could reproduce it once I had upgraded to jedi 0.10.

IPython/core/completer.py
Carreau: Probably not,  in case there are other similar issues I don't want to crash IPython completion, and make the user experience horrible. I prefer to silently fail (unless --debug is set).

I can make it more specific while on master for users to actually report bugs though.

minrk: I think catching Exception makes sense here.

takluyver:
@Carreau this needs a rebase, unfortunately.

Carreau:
Rebased.

Carreau:
Thanks !

PR #10245, opened 2017-02-03
Requires jedi >0.10 as it is now out.

Carreau:


takluyver:
Just spotted that installing jedi now downloads 26 MB 😕

Carreau:
> Just spotted that installing jedi now downloads 26 MB

???
```
$ pip install jedi
Collecting jedi
  Downloading jedi-0.10.0-py2.py3-none-any.whl (186kB)
    100% |...| 194kB 2.3MB/s
Installing collected packages: jedi
Successfully installed jedi-0.10.0
```

Carreau:
Confirmed in an empty env. Are you being spied on, and pip used to install backdoor on your computer ?

takluyver:
I filed a bug against Jedi and it got fixed before you tried. There was no
py3 wheel before, and the tarball had a random large wav file in it.

stonebig:
I have strange issues with Jedi-0.10.2 or jedi-trunk on Windows. Can this constraint safely be lowered to "jedi>=0.9 ?"

takluyver:
You can try installing an older version locally - we don't check the version at runtime, as far as I know. But @Carreau has been doing quite a bit of work on jedi, so don't be surprised if the older version breaks things.

Carreau:
Jedi 0.9 will definitely work until it crashes IPython. If you have issues
with master and IPython, please open bug reports, and I'll what I can do
(here or upstream).

PR #10249, opened 2017-02-06
disable `_ipython_display_` in terminal IPython

### LINKED ISSUE #10247: _ipython_display_, return value, and output prompt behavior
I might have found a strange interaction between `_ipython_display_` the in output prompt system. Given this file:

```python
# tst.py
from IPython import display

class A:
    def __repr__(self):
        return "    A\n0   B"


class B(A):
    def _ipython_display_(self):
        data = {'text/plain': repr(self)}
        display.display(data, raw=True)

class C(B):
    def _ipython_display_(self):
        super()._ipython_display_()
        return "C"
```

Here's the output:

```python
In [1]: from tst import *

In [2]: A()
Out[2]:
    A
0   B

In [3]: B()
Out[3]:     A
0   B

In [4]: C()
Out[4]:     A
0   B
```

I think the correct output for `In[3]` is just the `repr(self)`, but no `Out[3]` tag, since nothing is returned? I have no idea what the correct output of `In[4]` is, but maybe

```     
    A  # the super
0   B
"C"  # the return value of C._ipython_display_()
```
but I don't know if / where the `Out[4]` tag would go.

For background, we're working on a table schema repr for pandas in https://github.com/pandas-dev/pandas/pull/14904.

This is on IPython master from about 3 weeks ago.

cc @minrk

minrk:
We already try to disable all formatters other than text/plain. The ipython_display formatter is not in that mime-type list, so it was not being disabled with the others.

closes #10247

### DIFF
--- a/IPython/core/tests/test_formatters.py
+++ b/IPython/core/tests/test_formatters.py
@@ -407,6 +407,9 @@ def test_ipython_display_formatter():
         def _ipython_display_(self):
             raise NotImplementedError
     
+    save_enabled = f.ipython_display_formatter.enabled
+    f.ipython_display_formatter.enabled = True
+    
     yes = SelfDisplaying()
     no = NotSelfDisplaying()
     
@@ -419,6 +422,9 @@ def test_ipython_display_formatter():
     nt.assert_equal(d, {})
     nt.assert_equal(md, {})
     nt.assert_equal(catcher, [yes])
+
+    f.ipython_display_formatter.enabled = save_enabled
+
 
 def test_json_as_string_deprecated():
     class JSONString(object):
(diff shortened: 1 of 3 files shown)


takluyver:
Seems reasonable - I imagine objects with `_ipython_display_` are not intending it for the terminal.

Test failure:

```
FAIL: Objects with _ipython_display_ defined bypass other formatters
  File "/home/travis/virtualenv/python3.6.0/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/ipython/ipython/IPython/core/tests/test_formatters.py", line 419, in test_ipython_display_formatter
    nt.assert_equal(d, {})
AssertionError: {'text/plain': '<IPython.core.tests.test_f[80 chars]d0>'} != {}
```

minrk:
@takluyver thanks, fixed.

Carreau:
@meeseeksdev backport to 5.x

PR #10250, opened 2017-02-06
Speedup completions

Carreau:
Don't spend too much time in Jedi. It can be slow.
Return what we've got so far as fast as we can.

the timeout of 0.4 second was chosen by trial and error
with:

    import matplotlib
    import matplotlib.patches

    matplotblib.<tab>

keeping in mind that I have a fast machine.
There is an overhead associated with rerunning the completion so a too
short value, never leaves Jedi the chance to progress in computing more
completions.

--- 

I'll be trying to land patches on jedi to improve that, but it should not harm anyone as you should be mostly interested in only the first few completions

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -62,6 +62,7 @@ skip_doctest = True
 import __main__
 import builtins as builtin_mod
 import glob
+import time
 import inspect
 import itertools
 import keyword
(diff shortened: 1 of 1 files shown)


takluyver:
I had noticed some slowness, so let's try this out.

Carreau:
You want to try out https://github.com/davidhalter/jedi/pull/851 once it will be ready.

PR #10257, opened 2017-02-07
Finish old deprecation

Carreau:
All in title.

### DIFF
--- a/IPython/core/extensions.py
+++ b/IPython/core/extensions.py
@@ -13,6 +13,7 @@ from importlib import import_module
 
 from traitlets.config.configurable import Configurable
 from IPython.utils.path import ensure_dir_exists, compress_user
+from IPython.utils.decorators import undoc
 from traitlets import Instance
 
 try:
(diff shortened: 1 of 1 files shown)


takluyver:
I just went to check the [extensions index](https://github.com/ipython/ipython/wiki/Extensions-Index), and all the `%install_ext` instructions have been replaced with `%<deprecated command>`. I'm not very happy about this: it's punishing users rather than developers, and it doesn't give them any clear alternative to get those extensions that aren't on pip. I'm going to revert that.

I'm fine with deprecating this and removing it from docs, but I think we should leave it working for quite a long time.

Carreau:
Well install_ext has been removed in 5.0, so this is just cleaning unused
function with one last error to remind potentially other developer using
the API directly.

As for the wiki we can revert.that will just be showing non working
commands to users.

takluyver:
*Sigh* I didn't notice that one, or didn't think about it enough. I'm annoyed that we removed something that large numbers of extensions still recommend people use. I think we've been very over-aggressive with the deprecation here. But as it's done, and we don't have waves of protest, so I guess we should probably go ahead and remove this code.

Carreau:
I understand, and we deprecated it quickly to avoid this proliferation of bad habits and people copying this behavior. We started to have too many questions on how to update/have dependency, and user having `%install_ext` at the top of all their notebooks. I'm sorry you did not had a chance to ask for a longer deprecation cycle.... we can still make a pip installable fallback if you want.

takluyver:
I don't think a pip-installable option can work without having people do `%load_ext install_ext` first, and that gets a bit ridiculous. From the relative lack of issues complaining that we broke things, we got away with this, but I think it was a sadly cavalier attitude to extension authors to remove functionality. It's a long way from Linux's *don't break userspace*.

takluyver:
I'll fudge it on the wiki page :(

Carreau:
> I don't think a pip-installable option can work without having people do %load_ext install_ext first, and that gets a bit ridiculous.

Fair point. 

> From the relative lack of issues complaining that we broke things, we got away with this, but I think it was a sadly cavalier attitude to extension authors to remove functionality. It's a long way from Linux's don't break userspace.

We are a relatively fast moving project and we do have better backward compat and deprecation cycle than most. Also (hum.. Python 2 ... hum). 


> I'll fudge it on the wiki page :(

Thanks.

Speaking of "Don't break user space" I'm working on a tool to compare 2 versions of a python package and warn (fail?) when API is not backward compatible. Think roughly walking object hierarchy and dumping function signature. I'm mostly interested to catch the accidental positional-argument-renaming mistakes. I can ping you once I push a not too ugly version.

takluyver:
Regarding the tool, I'd say the important thing is to emphasise that getting a green light from the tool does not mean you're 100% backwards compatible: what happens inside the functions matters as well as the function signature, and that's much harder for a tool to check.

Carreau:
> Regarding the tool, I'd say the important thing is to emphasise that getting a green light from the tool does not mean you're 100% backwards compatible: what happens inside the functions matters as well as the function signature, and that's much harder for a tool to check.

Yes, definitively. I'm also not looking at types in the  signature. I'm looking at something that has as little false negative. Focusing for now on non intentional change of function signature (I catched a few in matplotlib 1.5->2.0 but it's too late now). And also something that can give me a list of change to review. 

A change can easily sneak through multiple PRs, as you may think a change had not been released yet and bring modification to APIs where you should not.

It does not of course remove work from the reviewer (I don't want to be without a job 😄  )

Carreau:
rebased.

PR #10258, opened 2017-02-07
Bump stack number on autoreload fail to 10.

Carreau:
This would allow for easier to debug error when encountered.


takluyver:
Want to backport this, or leave it for 6.0? Feel free to set the milestone either way.

Carreau:
6.0 is fine.we lived with 1 for a couple of years we can wait a few month :-)
But thanks for reminding me.

PR #10260, opened 2017-02-08
Fix the rest of the colors used on Windows so they are more visible

### LINKED ISSUE #10238: Issue with colors on Windows since IPython 5
I'm having some issues with colors in IPython 5 on Windows that I'm currently working around using configuration.

## Prompt Numbers
The prompt numbers are not using a bright color like they used to in IPython versions before 5. They use the same color as the In/Out text, I'm currently using this to override this: (Copied and modified from [IPython/terminal/interactiveshell.py:275-286](https://github.com/ipython/ipython/blob/e8341164d5f525a3bb3bf440cb042cc8145a5642/IPython/terminal/interactiveshell.py#L275-L286)
```py
c.TerminalInteractiveShell.highlighting_style_overrides = {
    Token.Number: '#007700',
    Token.Operator: 'noinherit',
    Token.String: '#BB6622',
    Token.Name.Function: '#2080D0',
    Token.Name.Class: 'bold #2080D0',
    Token.Name.Namespace: 'bold #2080D0',
    Token.Prompt: '#ansidarkgreen',
    Token.PromptNum: '#ansigreen bold',
    Token.OutPrompt: '#ansidarkred',
    Token.OutPromptNum: '#ansired bold',
}
```

### Before
ipython5-prompt-before

### After
ipython5-prompt-after

## Tracebacks highlighting
The highlighting used for tracebacks is weird, the code is highlighted like it used to before IPython 5 (I guess it's because of this: https://github.com/ipython/ipython/commit/7b7d560822d4e9b6ab486393e22e305c502ff6d1), But the text that decorates the traceback is using dark colors that are a bit hard on the eyes. I'm using this to override this:
```python
c.InteractiveShell.colors = 'Linux'  # Use Linux colors instead of Neutral
c.TerminalInteractiveShell.highlighting_style = 'default'  # Use default instead of monokai
```

The traceback coloring is defined here: [IPython/core/excolors.py](https://github.com/ipython/ipython/blob/master/IPython/core/excolors.py)

### Before
ipython5-tb-before

### After
ipython5-tb-after

## Live highlighting is different from tracebacks
(Not Windows related, but I want to mention this since it appears in the screenshots) 
The highlighting used for code in prompts is different from the one used in tracebacks (It's even a different library 😛 ), but I think this bit is known already. It shouldn't be too difficult to create a Pygments style which would make them the same but I haven't bothered with this.

## System Info
```python
{'commit_hash': u'be4827b',
 'default_encoding': 'cp862',
 'ipython_path': 'C:\\Python27\\lib\\site-packages\\IPython',
 'ipython_version': '5.2.1',
 'os_name': 'nt',
 'platform': 'Windows-10-10.0.14393',
 'sys_executable': 'C:\\Python27\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)]'}
```

prompt-toolkit 1.0.13

segevfiner:
This applies a hack similar to the one in
7b7d560822d4e9b6ab486393e22e305c502ff6d1 to change the prompt colors
and exception colors so they are more visible as shown in issue #10238

If this fix is acceptable it would be nice if this is backported to IPython 5 as this is very visible. Every user I see using IPython on Windows sees this and most don't seem to understand that they can change it. And on top of that this is especially annoying when using IPython on a computer you don't own or a VM. Having to reconfigure on each one...

Fixes: #10238

### DIFF
--- a/IPython/core/excolors.py
+++ b/IPython/core/excolors.py
@@ -3,6 +3,7 @@
 Color schemes for exception handling code in IPython.
 """
 
+import os
 import warnings
 
 #*****************************************************************************
@@ -155,6 +156,12 @@ def exception_colors():
         Normal = C.Normal,
         ))
 
+    # Hack: the 'neutral' colours are not very visible on a dark background on
+    # Windows. Since Windows command prompts have a dark background by default, and
+    # relatively few users are likely to alter that, we will use the 'Linux' colours,
+    # designed for a dark background, as the default on Windows.
+    if os.name == "nt":
+        ex_colors.add_scheme(ex_colors['Linux'].copy('Neutral'))
 
     return ex_colors
 
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -286,6 +286,16 @@ class TerminalInteractiveShell(InteractiveShell):
                     Token.OutPrompt: '#990000',
                     Token.OutPromptNum: '#ff0000 bold',
                 })
+
+                # Hack: Due to limited color support on the Windows console
+                # the prompt colors will be wrong without this
+                if os.name == 'nt':
+                    style_overrides.update({
+                        Token.Prompt: '#ansidarkgreen',
+                        Token.PromptNum: '#ansigreen bold',
+                        Token.OutPrompt: '#ansidarkred',
+                        Token.OutPromptNum: '#ansired bold',
+                    })
             elif legacy =='nocolor':
                 style_cls=_NoStyle
                 style_overrides = {}


IPython/core/excolors.py
```diff
@@ -155,6 +156,12 @@ def exception_colors():
         Normal = C.Normal,
         ))
 
+    # Hack: the 'neutral' colours are not very visible on a dark background on
+    # Windows. Since Windows command prompts have a dark background by default, and
+    # relatively few users are likely to alter that, we will use the 'Linux' colours,
+    # designed for a dark background, as the default on Windows.
+    if os.name == "nt":
+        ex_colors.add_scheme(ex_colors['Linux'].copy('Neutral'))
```
takluyver: Would it be easier (and more amusing) to make Windows use 'Linux' as the default scheme, even though Linux no longer does?

IPython/terminal/interactiveshell.py
```diff
@@ -286,6 +286,16 @@ def _make_style_from_name_or_cls(self, name_or_cls):
                     Token.OutPrompt: '#990000',
                     Token.OutPromptNum: '#ff0000 bold',
                 })
+
+                # Hack: Due to limited color support on the Windows console
+                # the prompt colors will be wrong without this
+                if os.name == 'nt':
+                    style_overrides.update({
+                        Token.Prompt: '#ansidarkgreen',
```
takluyver: @Carreau what version of pygments is needed for these?

segevfiner: It's actually prompt-toolkit that handles this. They are declared in [prompt_toolkit/styles/base.py:39-48](https://github.com/jonathanslenders/python-prompt-toolkit/blob/99fa7fae61c9b4ed9767ead3b4f9b1318cfa875d/prompt_toolkit/styles/base.py#L39-L48) and referenced throughout prompt-toolkit.

I tried using hex colors but prompt-toolkit seems quite stubborn to use the dark versions instead of the light versions...

IPython/core/excolors.py
segevfiner: The problem is that the Linux color scheme uses monokai syntax highlighting which is weird. 
([IPython/terminal/interactiveshell.py:266-268](https://github.com/ipython/ipython/blob/b8cfbfe27bdac730111d32a29f8aab469c5fcf45/IPython/terminal/interactiveshell.py#L266-L268)).

I think what we really want is the Linux color scheme with the Pygments default syntax highlighting with the style overrides used in Neutral. (With fixes for Windows when running on Windows)

takluyver: Maybe we should be creating a new name for the defaults we use on Windows. I'd like to hear what @Carreau thinks when he wakes up, though, because he's spent more time than me working on colour stuff.

segevfiner: Something like this https://github.com/segevfiner/ipython/commit/50c3fc12a6f8ebe55e2ea848e615974f582b5b28?

I guess @Carreau will decide?

IPython/terminal/interactiveshell.py
Carreau: > It's actually prompt-toolkit that handles this. They are declared in prompt_toolkit/styles/base.py:39-48 and referenced throughout prompt-toolkit.

Yes, though be careful, we may have custom code that special case `OutPromptNum` . Though I believe it will not be affected as it it generates `OutPrompt` Tokens and not consume the. I would have to check.

IPython/core/excolors.py
Carreau: I haven't use Windows in a while. I know there've been changes in the console, which is now white background (right) ?. This is IIRC one of the reason of "Neutral" to work both on Light and Dark Background.

But I'll trust a fellow windows user in the choice of colors. I've also tested locally and while it's not the coloring scheme I'm used too (put I'm picky and can have my own). I'm not opposed to making it the default even on non-'nt'

Carreau (approved):

takluyver:
Let's not go changing the default colours on non-Windows systems. The current ones are a hard-fought compromise, and I suspect that any change will quickly lead to more arguments.

Carreau:
> Let's not go changing the default colours on non-Windows systems. The current ones are a hard-fought compromise, and I suspect that any change will quickly lead to more arguments.

fair enough.

Carreau:
Ok let's move forward on this one. 

Thanks @segevfiner ! Merging.

segevfiner:
Thanks for merging! 😃 

You milestoned this as 6.0. Are you sure you don't want this in 5.x? Considering that https://github.com/ipython/ipython/commit/7b7d560822d4e9b6ab486393e22e305c502ff6d1 is already in 5.x and this pull request essentially completes it, and also fixes the prompt numbers so they look like they used to in IPython 4 and below.

Since 5 (LTS) is the last version to support Python 2.7, many will be using it for quite a while even after 6.0 is out, maybe even using both versions at the same time since they work with both Python 2 and Python 3. Considering this, it will be nice if IPython 5 on Windows uses the same colors as IPython 6+ on Windows, otherwise it would look quite weird for a user that uses both versions that the colors are different and it's nicer if the user doesn't have to use custom configuration to make them the same (At least as long IPython 5/6 and later use the same highlighting system).

Carreau:
Well, that might push them to upgrade to Python 3 and upgrade to IPython 6 right ?
Let's see if it backports cleanly.

@meeseeksdev backport to 5.x

Carreau:
Well it seems like MrMeeseeks likes you. See #10260

PR #10264, opened 2017-02-09
Indent on new line by looking at the text before the cursor.

### LINKED ISSUE #9283: TAB key does not indent
In the git version of IPython, type

```
def foo(a):
    pass<ENTER>
```

The new line is not correctly indented, and the tab key does not insert 4 spaces.

/cc @Carreau @takluyver

Carreau:
that would allow edit like the following to insert the right amount of
space on the newline (`|` is the cursor)

    class Bar:
        def __init__(self):|<enter>
            self.filed=0

Instead of inserting a blank line with 0 space.

Closes #9283

---

@takluyver can you have a look ?

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -100,12 +100,14 @@ def newline_or_execute_outer(shell):
                 b.cancel_completion()
             return
 
-        if not (d.on_last_line or d.cursor_position_row >= d.line_count
-            - d.empty_line_count_at_the_end()):
-            b.newline()
-            return
+        before_text = d.text[:d.cursor_position]
+        status, indent = shell.input_splitter.check_complete(before_text + '\n')
 
-        status, indent = shell.input_splitter.check_complete(d.text + '\n')
+        if not (d.on_last_line or
+                d.cursor_position_row >= d.line_count - d.empty_line_count_at_the_end()
+                ):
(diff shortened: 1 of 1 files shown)


IPython/terminal/shortcuts.py
```diff
@@ -100,12 +100,14 @@ def newline_or_execute(event):
...
-            return
+        before_text = d.text[:d.cursor_position]
+        status, indent = shell.input_splitter.check_complete(before_text + '\n')
 
-        status, indent = shell.input_splitter.check_complete(d.text + '\n')
+        if not (d.on_last_line or
+                d.cursor_position_row >= d.line_count - d.empty_line_count_at_the_end()
+                ):
+            b.insert_text('\n' + (' ' * (indent or 0)))
+            return
```
Carreau: ALso I'm unsure, but it seem to me that this block is completely redundant and can be remove. It is taken care of by the block below. (this is due to the fact that we cut the text at the cursor position, so `status` will never be 'ready'

Carreau: Scratch that. 

```
In [1]: def foo():
   ...:     pass
   ...: |<enter>
   ...: def bar():
   ...:     pass
```

Above example insert a new line with this patch. 
Execute if I remove above block.

Carreau:
@meeseeksdev backport to 5.x

Issue #9514, opened 2016-05-31
deprecate `ipython_extension_dir` ?


Carreau:
Now that we can mostly require any IPython extension to be a package and are recommending that. 

Would it make sent to deprecate, loading things from `ipython_extension_dir`, like found in `IPython/core/extensions.py`:

```
if module_str not in sys.modules:
                with prepended_to_syspath(self.ipython_extension_dir):
                    __import__(module_str)
```

Carreau:
ping @fperez

PR #10266, opened 2017-02-09
Warn about IPython extension dir

Carreau:
See #9514. 

I'm not happy with this approach, I would like to warn only if extensions are actually being loaded from this folder, so I might need to inspect he loaded module if it is loaded. 

cc @takluyver as this touch `install_ext` mechanism, I would like to have this to give useful information to people that might still be affected by older extensions. 

I'm also unsure if warnings are the best tool as it is not technically deprecated, but is simply not really advised.

### DIFF
--- a/IPython/core/extensions.py
+++ b/IPython/core/extensions.py
@@ -5,6 +5,9 @@
 # Distributed under the terms of the Modified BSD License.
 
 import os
+import os.path
+import warnings
+import textwrap
 from shutil import copyfile
 import sys
 from importlib import import_module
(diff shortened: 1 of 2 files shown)


takluyver:
I also think we should inspect modules to see if they're loaded from the extensions directory. It shouldn't be hard to use `mod.__file__`.

Carreau:
How about the following ?

takluyver:
That looks like a lot of text to throw at users for something that's (IMO) rather gently deprecated. I don't think there's any hurry to get rid of the extensions dir.

Carreau:
You want to bump to 7.0, or have a shot at a shorter text ?

takluyver:
I'll have a go at a shorter text. Let's see if I can push to your branch.

takluyver:
How's that?

Carreau:
> How's that?

Sorry I missed your comment !

Thanks !

Self-but-not-really-self-as-you-pushed-changes merging.

PR #10267, opened 2017-02-09
Improve indentation of next line when using CTRL-o in terminal.

### LINKED ISSUE #9588: In the PTK terminal, control-o should indent the newly created line

Carreau:
Closes #9588

Similar to #10264, but with a simpler logic as we never execute.

--- 

I think that @takluyver might want to have a look as well as @gforsyth.

We can probably get rid of the need for an InputSplitter instance as we process all at one. 
Would it make sens to in general remove the stateful part of InutSplitter, and open a separate issue for this ?

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -1,5 +1,16 @@
+"""
+Module to define and register Terminal IPython shortcuts with
+:any:`prompt_toolkit`
+"""
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import warnings
 import signal
 import sys
+from typing import Callable
+
 
 from prompt_toolkit.enums import DEFAULT_BUFFER, SEARCH_BUFFER
 from prompt_toolkit.filters import (HasFocus, HasSelection, Condition,
@@ -60,7 +71,7 @@ def register_ipython_shortcuts(registry, shell):
 
     registry.add_binding(Keys.ControlO,
                          filter=(HasFocus(DEFAULT_BUFFER)
-                                & EmacsInsertMode()))(newline_with_copy_margin)
+                                & EmacsInsertMode()))(newline_autoindent_outer(shell.input_splitter))
 
     registry.add_binding(Keys.F2,
                          filter=HasFocus(DEFAULT_BUFFER)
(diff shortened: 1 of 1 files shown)


takluyver:
I merged #10264 before seeing this. Is it still relevant?

Carreau:
Yep. #10264 is (was) the behavior of Enter (ie, execute or insert new line and indent moving the cursor.
This is insert newline and indent _after_ the cursor.

IPython/terminal/shortcuts.py
```diff
@@ -178,6 +198,38 @@ def newline_with_copy_margin(event):
...
+
+        text = d.text[:d.cursor_position] + '\n'
+
+        _, indent = inputsplitter.check_complete(text)
+        b.insert_text('\n' + (' ' * (indent or 0)))
+        b.cursor_up(count=1)
+        cursor_end_pos = b.document.cursor_position_col
+        if cursor_start_pos != cursor_end_pos:
+            pos_diff = cursor_start_pos - cursor_end_pos
+            b.cursor_right(count=pos_diff)
```
takluyver: Looking at the prompt_toolkit source code, `.insert_text()` can take a `move_cursor=False` parameter, which would simplify this a fair bit.

takluyver: And that has been there [since 1.0.4], which is the minimum we require, so we should be fine to use it.

takluyver: That link was: https://github.com/jonathanslenders/python-prompt-toolkit/blob/1.0.4/prompt_toolkit/buffer.py#L913

takluyver:
I suspect there are bits of InputSplitter's API we can simplify now, but that would have to be done very carefully. If you do create an issue, please assign it to me.

IPython/terminal/shortcuts.py
Carreau: Well spotted. Updated accordingly.

PR #10269, opened 2017-02-09
Fix docstring of FileLink to match parameter name

StevenMaude:
It used to be `directory_prefix`, but is currently `url_prefix`.

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -325,7 +325,7 @@ class FileLink(object):
             path to the file or directory that should be formatted
         url_prefix : str
             prefix to be prepended to all files to form a working link [default:
-            'files']
+            '']
         result_html_prefix : str
             text to append to beginning to link [default: none]
         result_html_suffix : str
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -327,7 +327,7 @@ class FileLink(object):
             prefix to be prepended to all files to form a working link [default:
             '']
         result_html_prefix : str
-            text to append to beginning to link [default: none]
+            text to append to beginning to link [default: '']
         result_html_suffix : str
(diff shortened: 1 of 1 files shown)


IPython/lib/display.py
```diff
@@ -323,7 +323,7 @@ def __init__(self,
         ----------
         path : str
             path to the file or directory that should be formatted
-        directory_prefix : str
+        url_prefix : str
             prefix to be prepended to all files to form a working link [default:
             'files']
```
takluyver: The default also appears to have changed to `''`

StevenMaude: @takluyver Just playing around with FileLink now, and spotted that too.

Will add to this PR.

IPython/lib/display.py
```diff
@@ -327,7 +327,7 @@ def __init__(self,
             prefix to be prepended to all files to form a working link [default:
             '']
         result_html_prefix : str
-            text to append to beginning to link [default: none]
+            text to append to beginning to link [default: '']
```
StevenMaude: Changed this one for consistency with `url_prefix` docstring update.

First time looking at IPython code, so don't know if the use of `none` here is idiomatic or not. If not, I can drop this commit from the PR.

takluyver: Thanks, I prefer the precision of what you changed it to.

takluyver:
Thanks for your first contribution to IPython! :cookie:

Carreau:
@meeseeksdev backport to 5.x

Carreau:
@meeseeksdev backport to 5.x

Carreau:
> backport to 5.x

He can't because the fork is gone. I need to fix that at some point.

Issue #9490, opened 2016-05-23
Remove test that rely on matplotlib, pandas.... etc


Carreau:
We have a few test that rely on external libraries we don't install and thus explicitly  skip on travis:

```
   1 @dec.skip_without("PIL.Image")
   3 @dec.skip_without('matplotlib')
   2 @dec.skip_without('nbformat')
   1 @dec.skip_without('numpy')
   1 @dec.skip_without('pandas')
```

May I suggest to either : 
  - 1) make them test requirements 
  - 2) get rid of the tests (at least if they don't pass)

Carreau:
Well, I won't try that this morning:

``
May 23, 2016
09:53 PDT Some users may experience a delay in pushes or other changes appearing on the site.
```

Anything I've pushed in the last hour are still not appearing in GitHub, so Travis do not pick it up, and I can't even make PRs...

Carreau:
So we can't do that for IPython 6.0, numpy is 3.4+ only. Bump to 7.0

takluyver:
+1 to installing extra packages on Travis where it's easy & quick, but -1 to requiring compiled extension modules for the tests to run successfully - see my comment on #10271 .

PR #10271, opened 2017-02-09
Make test dependencies non-optional (numpy, pandas)

Carreau:
Step toward #9490


Carreau:
Dang. can't do that. Numpy is 3.4+ only.

takluyver:
I'm not sure about this in general: wheels have made things easier to install, but they still don't cover e.g. ARM architectures - it's perfectly possible to run IPython on a Raspberry Pi, for instance, where pip can't easily install numpy.

PR #10277, opened 2017-02-09
Requires typing only for older python.

### LINKED ISSUE #10265: typing should not be a dependency for python>=3.5
https://github.com/ipython/ipython/commit/3ff1be2ea8ef180a6f17a6a03a3f8452303b9abe added `typing` as an `install_requires`.  This package is builtin for python>=3.5, so it should be protected by an environment marker (just like `pathlib2`) in setup.py.

Carreau:
Closes #10265

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -202,7 +202,6 @@ extras_require = dict(
 install_requires = [
     'setuptools>=18.5',
     'jedi>=0.10',
-    'typing',
     'decorator',
     'pickleshare',
     'simplegeneric>0.8',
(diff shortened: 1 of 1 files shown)


takluyver:
I personally find it clearer to use `<3.5` rather than `<=3.4`, but that's not worth worrying about.

(The reverse applies for specifying a minimum version; there I like `>=`)

Carreau:
> I personally find it clearer to use <3.5 rather than <=3.4, but that's not worth worrying about.

Yeah, but the the colon looked misaligned. :-)

Issue #9517, opened 2016-06-01
Install packages into the right environment


takluyver:
The biggest cause of issues I see when helping people with Python is confusion over installing packages in different environments and different Python installations. This typically takes the form of:

> I can do foo when running X but not when running Y (e.g. the notebook). Help!

Where the issue is that Y is running their code in a different environment or with a different Python, and that either doesn't have a module, or has a different version. This is only getting worse as using environments and multiple Python versions becomes more common.

I know we don't want to wade into package management, but I think IPython is quite well placed to provide a simple but useful interface. Something along the lines of @Carreau's [pip_magic](https://github.com/Carreau/pip_magic), but I think there are a couple of major issues with that:

1. I think it's pedagogically bad to use exactly the same interface as the `pip` command line, because we already see enough people confused about the distinction between command line and Python commands.
2. It's unaware of conda, and e.g. updating a package with pip when it's been conda installed can lead to problems.
3. You have to install it in each environment before using it, which brings you back to the original issue! This would be reduced if such a function were included in IPython.

I'd like to have something akin to R's [install.packages](http://stat.ethz.ch/R-manual/R-devel/library/utils/html/install.packages.html) function, which is the primary way to install packages inside R. Can we design something that has sufficient awareness of pip and conda packages, and virtualenvs and conda envs, that we can tell users to e.g. `install('matplotlib')`, and it will work out the best way to do that in the current environment.

takluyver:
(Tagged as 5.0 but I fully expect this to be bumped - but if we can get it sorted out, I think it's worth backporting to 5.x)

Carreau:
I asked the conda folks some time ago,  whether we could have an `import conda` and can do a `conda.install('<package>')` the answer was no, and I would **prefer** not to shell out.

@bollwyvl might be of help, anaconda navigator might have an API ?

I believe @njsmith  have some rant about packaging too, and would like a separate tool that not only can install Python packages but **actually resolve dependencies**. 

I guess within IPython we could, if `ImportError`, look at the env variable, and if `CONDA_DEFAULT_ENV`, is define, extend the warning with `did you install this in your <...> environament ?`

It's also painful not to be able to detect a package that cannot be imported from a package that is not installed. @njsmith you are at Pycon and language summit, anything you can do ? Like subclass od ImportError (`DependencyCannotBeFound`, `DependencyFailedtoLoad`) ... @njsmith if you are still with CPython crowd, maybe you can get something out of them ?

[...]

PR #10280, opened 2017-02-09
Insert a fake pip magic warning users the should use the shell.

Carreau:
Bump #9517 to 7.0

Revived by a discussion with @fperez yesterday.

I'm wondering if we should try to catch `pipX[.y]` but it seem to me like user of `pipX.y` probably know what they are doing. 

@willingc any thoughts on the phrasing ?

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -2,6 +2,7 @@
 
 
 import argparse
+import textwrap
 import io
 import sys
 from pprint import pformat
@@ -17,7 +18,8 @@ from logging import error
 
 
 class MagicsDisplay(object):
-    def __init__(self, magics_manager):
+    def __init__(self, magics_manager, ignore=None):
+        self.ignore = ignore if ignore else []
         self.magics_manager = magics_manager
     
     def _lsmagic(self):
@@ -27,10 +29,10 @@ class MagicsDisplay(object):
         mman = self.magics_manager
         magics = mman.lsmagic()
         out = ['Available line magics:',
-               mesc + ('  '+mesc).join(sorted(magics['line'])),
+               mesc + ('  '+mesc).join(sorted([m for m,v in magics['line'].items() if (v not in self.ignore)])),
                '',
                'Available cell magics:',
-               cesc + ('  '+cesc).join(sorted(magics['cell'])),
+               cesc + ('  '+cesc).join(sorted([m for m,v in magics['cell'].items() if (v not in self.ignore)])),
                '',
                mman.auto_status()]
         return '\n'.join(out)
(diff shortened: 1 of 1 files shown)


takluyver:
Do we want to come up with a way to hide it from `%lsmagic`? I don't think something that just warns 'you're doing it wrong' should show up in the list.

Also, do we want to suggest using `!pip`, or is that too confusing?

Carreau:
> Do we want to come up with a way to hide it from %lsmagic? I don't think something that just warns 'you're doing it wrong' should show up in the list.

Good point.

>Also, do we want to suggest using !pip, or is that too confusing?

Probably no. `!pip` may not do the right thing, so I prefer to avoid.

Carreau:
> Good point.

Done. If someone register another magic named pip it should not be ignored. Meaning that `pip_magic` still works as expected.

takluyver:
Fair enough re `!pip`.

I've started a discussion on distutils-sig about indexing which modules come from which distributions, so we could eventually make a tool that you can do `install_module('zmq')` instead of `pip install pyzmq`. Part of why I want this is so that one tool can select between pip and conda install options (and potentially other sources too)

Carreau:
> I've started a discussion on distutils-sig

Yes, I've seen it. It would be great. @fperez started [this](https://github.com/fperez/conda-edev) also, to have pip and conda get along better. It sure would be nice.

IPython/core/magics/basic.py
```diff
@@ -377,7 +379,24 @@ def xmode_switch_err(name):
             xmode_switch_err('user')
 
     @line_magic
-    def quickref(self,arg):
+    def pip(self, args=''):
+        """
+        Intercept usage of `pip` in the IPython shell and print an explanation.
+        """
+        print(textwrap.dedent('''
+        The following command is not meant to be run from a Python shell:
```
takluyver: I think this could be worded more clearly.

> pip commands should be run in a system shell or command prompt, not inside Python.

We could even check the platform and switch between 'system shell' for Unix and 'command prompt' for Windows.

Carreau: Hum, with bash-for-window as and other stuff I'd like to avoid this.

Carreau: Also did you mean changing the first line all the all text for your alternative ?

willingc:
```
"""
Intercept usage of `pip` in IPython and direct user to run command outside of IPython.
"""

The following command must be run outside of the IPython shell:

    $ pip {args}

The Python package manager (pip) can only be used from outside of IPython.
Please reissue the `pip` command in a separate terminal or command prompt.

See the Python documentation for more informations on how to install packages:
 
     https://docs.python.org/3/installing/'''.format(args=args)))

```

Carreau:
Apologies for the delay, phrasing updated.

Issue #10282, opened 2017-02-10
Duplicate completions in master


fperez:
Right now, I'm getting most attributes listed twice in the completions list:

![image](https://cloud.githubusercontent.com/assets/57394/22811815/fbea22ca-eef5-11e6-9941-69b7d23d6094.png)

Interestingly, the `mro` isn't duplicated, but most others are.  Particularly glaring with longer lists:

![image](https://cloud.githubusercontent.com/assets/57394/22811833/19cb995e-eef6-11e6-87db-dd109ba3836e.png)

willingc:
@fperez Did you break IPython again? ;-)

fperez:
I swear they put `if user == "fperez"` snippets in there just to keep me on my toes ;)

takluyver:
I think this is something temporary that @Carreau put in there to test the new Jedi completion system.

Carreau:
> I swear they put if user == "fperez" snippets in there just to keep me on my toes ;)

Yes we did, yes we did. If you read [reflexion on trusting trust](https://www.ece.cmu.edu/~ganger/712.fall02/papers/p761-thompson.pdf)  you'll figure out how we did it, and why you never find the reason. 

Without joking the first column in Jedi, the second is IPython. Fixing the bug is not that hard, it's just that "visually" the completion are identical, under the hood they are not. you press tab several time to highlight some of them you will see the first column should show the types. The second is that under the hood. `it.accumulate` is "Instert `.accumulate` at index 3." for left column for right one: "replace text from 0 to 2 by `it.accumulate`."

There is even a comment in the code:

```
        # Possible Improvements / Known limitation
        ##########################################
        # Completions may be identical even if they have different ranges and
        # text. For example:
        # >>> a=1
        # >>> a.<tab>
        # May returns:
        #  - `a.real` from 0 to 2
        #  - `.real` from 1 to 2
        # the current code does not (yet) check for such equivalence
```

I'll push an ugly 2 lines fix.

You will still see duplicate now and then, in particular `from foo imp<tab>` has 2 completion with and without trailing space.

fperez:
Ah, I see the difference now, thanks! I don't think there's any rush to fix this quickly, I wasn't really complaining :)  I just wanted to report it in case it wasn't flagged (I didn't see it open), if nothing else for folks that may be following master and be surprised.  As long as the issue is here open and explained, it's perfectly OK if a clean fix takes a bit of time.

[...]

PR #10284, opened 2017-02-10
Workaround for duplicate completions

Carreau:
Workaround for #10282

I consider this a workaround as this works only on the regularized
completions that have been extended with surrounding text in order to
still match old API. The deduplication should happen earlier.

I can't just either "remove" the IPython completer as it gives results
Jedi does not.


Carreau:
Scratch that I'll properly fix it

PR #10285, opened 2017-02-10
Deduplicate completions between IPython and Jedi.

Carreau:
Completions coming from different sources may be in practice identical,
but the object themselves differ.

Provide a function to properly deduplicate them.

Not perfect for #10282 in particular with trailing spaces, but can be
dealt with later.

--- 

cc @fperez

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -344,11 +344,52 @@ class Completion:
     def __hash__(self):
         return hash((self.start, self.end, self.text))
 
+
 _IC = Iterator[Completion]
 
-def rectify_completions(text:str, completion:_IC, *, _debug=False)->_IC:
+
+def _deduplicate_completions(text: str, completions: _IC)-> _IC:
     """
-    Rectify a set of completion to all have the same ``start`` and ``end``
+    Deduplicate a set of completions.
+
+    .. warning:: Unstable
+
+        This function is unstable, API may change without warning.
+
+    Parameters
+    ----------
+    text: str
+        text that should be completed.
+    completions: Iterator[Completion]
+        iterator over the completions to deduplicate
+
+
+    Completions coming from multiple sources, may be different but end up having
+    the same effect when applied to ``text``. If this is the case, this will
+    consider completions as equal and only emit the first encountered.
+
+    Not folded in `completions()` yet for debugging purpose, and to detect when
+    the IPython completer does return things that Jedi does not, but should be
+    at some point.
+    """
+    completions = list(completions)
+    if not completions:
+        return
+
+    new_start = min(c.start for c in completions)
+    new_end = max(c.end for c in completions)
+
+    seen = set()
+    for c in completions:
+        new_text = text[new_start:c.start] + c.text + text[c.end:new_end]
+        if new_text not in seen:
+            yield c
+            seen.add(new_text)
+
+
+def rectify_completions(text: str, completions: _IC, *, _debug=False)->_IC:
+    """
+    Rectify a set of completions to all have the same ``start`` and ``end``
 
     .. warning:: Unstable
 
@@ -359,7 +400,7 @@ def rectify_completions(text:str, completion:_IC, *, _debug=False)->_IC:
     ----------
     text: str
         text that should be completed.
-    completion: Iterator[Completion]
+    completions: Iterator[Completion]
         iterator over the completions to rectify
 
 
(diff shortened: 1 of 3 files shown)


Carreau:
Also I expect in the longrun for this function to disapear, but I want to keep it separate for now. 
Also brain a bit slow today, as I'm still likely sick, but I think that should work.

IPython/core/tests/test_completer.py
```diff
@@ -294,14 +294,26 @@ def _test_not_complete(reason, s, comp):
...
     yield _test_complete, 'cover duplicate completions', 'im', 'import', 0, 2
 
     yield _test_not_complete, 'does not mix types', 'a=(1,"foo");a[0].', 'capitalize'
 
+def test_deduplicate_completions():
+    """
+    Test that completions are correctly deduplicated (even if ranges are not the same)
+    """
+    ip = get_ipython()
+    ip.ex('import itertools as it')
```
fperez: I would instead write the test to run with a simpler object, like a simple custom class:

```
In [5]: class Z:
   ...:     z = 1
   ...:     

In [6]: Z.z<TAB>
```

If itertools later develops a second method starting with the letter 'a', this test would break.

IPython/terminal/ptutils.py
```diff
@@ -10,7 +10,7 @@
 import unicodedata
 from wcwidth import wcwidth
 
-from IPython.core.completer import IPCompleter, provisionalcompleter, rectify_completions, cursor_to_position
+from IPython.core.completer import IPCompleter, provisionalcompleter, rectify_completions, cursor_to_position, _deduplicate_completions
```
fperez: Line a bit too long? We used to keep line lengths ~80, though I don't know if that's being followed still :)

IPython/core/completer.py
```diff
@@ -1510,22 +1554,18 @@ def completions(self, text: str, offset: int)->Iterator[Completion]:
             fake Completion token to distinguish completion returned by Jedi
             and usual IPython completion.
 
+        .. note::
+            
+            Completions are not completely reduplicated yet. If identical
+            completions are coming from different sources this function does not
+            ensure that each completion  object will only be present once.
+
```
fperez: Spurious blank line?

IPython/core/completer.py
```diff
@@ -344,11 +344,55 @@ def __eq__(self, other)->Bool:
...
+    Parameters
+    ----------
+    text: str
+        text that should be completed.
+    completions: Iterator[Completion]
+        iterator over the completions to deduplicate
+
+
+    Completions coming from multiple source, may be different but end up having
```
fperez: s/source/sources/

IPython/core/completer.py
```diff
@@ -344,11 +344,55 @@ def __eq__(self, other)->Bool:
...
+    the same effect when applied to ``text``. If this is the case, this will
+    consider completions as equal and only emit the first encountered.
+
+    Not folded in `completions()` yet for debug purpose, and to detect when the
```
fperez: s/debug/debugging/

IPython/core/completer.py
```diff
@@ -1510,22 +1554,18 @@ def completions(self, text: str, offset: int)->Iterator[Completion]:
             fake Completion token to distinguish completion returned by Jedi
             and usual IPython completion.
 
+        .. note::
+            
+            Completions are not completely reduplicated yet. If identical
```
fperez: you mean "de"-duplicated, right?

IPython/core/completer.py
```diff
@@ -1510,22 +1554,18 @@ def completions(self, text: str, offset: int)->Iterator[Completion]:
...
+            completions are coming from different sources this function does not
+            ensure that each completion  object will only be present once.
```
fperez: double whitespace before "object"

IPython/core/completer.py
```diff
@@ -344,11 +344,55 @@ def __eq__(self, other)->Bool:
...
+    """
+    completions = list(completions)
+    if not completions:
+        return
+
+    starts = (c.start for c in completions)
+    ends = (c.end for c in completions)
+
+    new_start = min(starts)
+    new_end = max(ends)
```
fperez: Since `starts` and `ends` aren't really needed further, I think it would be clearer/simpler to write

```python
new_start = min(c.start for c in completions)
new_end = max(c.end for c in completions)
```

no?

fperez:
I left some code-specific comments, thanks! 

I was going to suggest deduplicating by making a set, but then I [saw the comments in the Completions class](https://github.com/Carreau/ipython/blob/a2cad6afcf335824a0b4ee5139beba69766fc940/IPython/core/completer.py#L336) and it seems the approach you're using is actually necessary.

Other than those easy fixes, I don't see any way to make it significantly cleaner/simpler off-hand.

IPython/core/completer.py
Carreau: done.

IPython/core/completer.py
Carreau: done.

IPython/core/completer.py
Carreau: yep, Some copy pasting where it was necessary.

IPython/core/completer.py
Carreau: Indeed, added `deduplicated` in my VIm dictionary.

IPython/core/completer.py
Carreau: Lynx eyes I see.

IPython/core/completer.py
Carreau: Yes. Blank line make me feel comfortable when things get converted o RST :-)

IPython/core/tests/test_completer.py
Carreau: Done.

IPython/terminal/ptutils.py
Carreau: Done.

Carreau:
Thanks, all comments addressed, and commits squashed.

fperez:
Great, thanks! Merging.

Issue #10283, opened 2017-02-10
Python completions come up inside strings


takluyver:
I have files called e.g. `embed-test.py`. If I type `open('embed-<TAB>`, I see this:

![screenshot from 2017-02-10 14-16-38](https://cloud.githubusercontent.com/assets/327925/22829715/d16cd64a-ef9b-11e6-9d04-fae716eced55.png)

I think this is new with the Jedi completion system. We shouldn't be calculating Python completions when we're inside a string literal.

tonyzhoujingtao:
Do you expect no completion or intelligently showing files in the current directory?

takluyver:
The files - it already does this, but those completions are getting buried amidst the erroneous completions in master.

Carreau:
> The files - it already does this, but those completions are getting buried amidst the erroneous completions in master.

That a side effect / compound effect  of Jedi + IPython completion. 
technically there is 2 bugs: 
   1) Jedi return only the part after the `-` (abs, all, and), 

But because IPython requires all completion to have the same start and end, and that IPython does return `embed-test`, then the wrapper have to "pad" jedi completion with `embed-`.

takluyver:
Right, the padding makes sense, but I don't want those completions in that
place at all.

Carreau:
Opened https://github.com/davidhalter/jedi/issues/860. I have an idea fro a workaround.

Carreau:
> Right, the padding makes sense, but I don't want those completions in that
> place at all.

Just as a note, `open('<tab>` and `open('embed-<tab>` are two different cases, one of which is a regression and simple to fix (see #10286) the other one being the current behavior of IPython since long ago, and would require quite a bit more refactor.

takluyver:
Fixed for now by #10286.

The follow up issue #10292 is to move to a cleaner fix when possible.

PR #10286, opened 2017-02-10
Avoid triggering jedi in string.

Carreau:
Rely on the fact that the net to last token is an error and starts with
", or '

@takluyver, can you try ? That should fix the `open('embed-<tab>` which is a regression,  but not the `open('<tab>` which also complete irrelevant things, but that's the case on master. 

Would that be sufficient to bump https://github.com/ipython/ipython/issues/10283 to 7.0 ?

If so I'll add tests and polish documentation.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1105,6 +1105,26 @@ class IPCompleter(Completer):
 
         interpreter = jedi.Interpreter(
             text, namespaces, column=cursor_column, line=cursor_line + 1)
+
+        try_jedi = False
+
+        try:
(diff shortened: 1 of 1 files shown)


takluyver:
Yep, that works for me. If we merge this, I'd be inclined to close #10283 - as it does fix that problem - and open a new issue to remind us to clean up the use of private Jedi APIs when possible.

Carreau:
> and open a new issue to remind us to clean up the use of private Jedi APIs when possible.

Done as #10292

takluyver:
Thanks!

PR #10294, opened 2017-02-15
remove utf-8 encode/decode input param since on python3 it is the default encoding/decoding format

srinivasreddy:


takluyver:
I'm inclined to leave these on the grounds of 'explicit is better than implicit'

Carreau:
I'm also worried as some of the requests come from network, and it's not because python builtin str/bytes are by default utf-8 that the byte-like and str-like thing we get decide to to default to utf8.

So -1 as well.

Still thank you very much for helping with the maintenance.  There are a couple of explicit task that would be more rewarding and challenging for you and users, like all the issues marked as [Need contributor targettign both 5.3 or 6.0](https://github.com/ipython/ipython/issues?utf8=%E2%9C%93&q=is%3Aopen%20is%3Aissue%20milestone%3A6.0%20milestone%3A5.3%20label%3A%22Need%20Contributor%22%20)

srinivasreddy:
Thanks. I am closing it

srinivasreddy:
@Carreau Thanks for the list.  👍

Carreau:
> Thanks. I am closing it

Thanks for understanding ! And thanks for your contributions !

PR #10301, opened 2017-02-18
FIX: re-order qt eventloop hook a bit

### LINKED ISSUE #10201: ipython 5.1 sporadically freezes when matplotlib plots are active
I've using ipython 5.1 with matplotlib 1.5.1 with the Qt4Agg backend on Mac OS X 10.9.5.  There have now been 3 occasions when the ipython terminal froze entirely - it would not accept control-c, control-z, or any command I tried.  Quitting the Qt plot windows re-activated the python terminal.

This so far has *not* affected me the first or second time I've made a plot window, but eventually it seems to affect every ipython session I've started with a plot window.

I upgraded to matplotlib 2.0, and the same thing happened, but this time (now with Qt5Agg as a backend), I get a message:

```
QEventDispatcherMac::unregisterSocketNotifier: Tried to unregister a not registered notifier
```

After that notification, the entire terminal containing ipython became unresponsive.  I think the same message was issued under ipython + matplotlib 1.5.1, but the message only appeared after closing ipython entirely, i.e.:

```
In [11]:
QEventDispatcherMac::unregisterSocketNotifier: Tried to unregister a not registered notifier
```

tacaswell:
try to reduce race conditions by:

 - connect the exit callback before enabling the notifier (this might
   not matter)
 - only bother starting the event loop if the input is not ready.

closes #10201 

I don't have a mac to test this on.  This may be complete cargo-culting.

### DIFF
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -5,6 +5,7 @@ from IPython.external.qt_for_kernel import QtCore, QtGui
 # garbage collected.
 _appref = None
 
+
 def inputhook(context):
     global _appref
     app = QtCore.QCoreApplication.instance()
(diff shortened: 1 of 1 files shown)


minrk (approved):

Carreau:
@meeseeksdev backport to 5.x

PR #10304, opened 2017-02-20
BUG: Make behavior of -i and -n independent.

### LINKED ISSUE #10303: BUG: (Unintended?) Interaction between -n and -i arguments to %run
The documentation states that `-i` would run the script in the present notebook's namespace while `-n` would prevent the `__name__` in the script from being set to `__main__`, thereby preventing execution of code hidden behind the `if __name__ == '__main__':` check.

However, [the code](https://github.com/ipython/ipython/blob/master/IPython/core/magics/execution.py#L665) only takes the `-n` into account only if `-i` is NOT passed to the `%run` (i.e. in the `else` branch).

Shouldn't it be handled before checking for `-i`?

Or is it necessary to change the `__name__` of the program set to `__main__` to run it in the namespace of the notebook?

musically-ut:
Fixes #10303.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -649,23 +649,24 @@ python-profiler package from non-free.""")
 
         sys.argv = [filename] + args  # put in the proper filename
 
+        if 'n' in opts:
+            name = os.path.splitext(os.path.basename(filename))[0]
+        else:
+            name = '__main__'
+
         if 'i' in opts:
             # Run in user's interactive namespace
             prog_ns = self.shell.user_ns
             __name__save = self.shell.user_ns['__name__']
-            prog_ns['__name__'] = '__main__'
+            prog_ns['__name__'] = name
             main_mod = self.shell.user_module
-            
+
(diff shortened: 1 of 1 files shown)


takluyver:
This seems reasonable. I've marked it for 6.0 because `%run` is quite delicate code, so I'd rather not change it in a 5.x release.

musically-ut:
Umm ... so this change is going to be in the next release?

Should I make a CHANGELOG entry somewhere?

takluyver:
Oh yeah, that's a good idea. If you want to make a new PR for it, put a rst snippet in `docs/source/whatsnew/pr/` - we have a script that later merges them into a changelog file.

musically-ut:
Okay, will do.
And any regression tests?

takluyver:
%run can be a bit awkward to test, which is why I didn't ask for one before, but if you're keen, please have a look at the existing tests in `IPython.core.tests.test_run`. Actually, this is probably easier than I thought - have a look at `test_run__name__` in that file.

PR #10306, opened 2017-02-20
reset to initial state

srinivasreddy:


srinivasreddy:
:)

takluyver:
That's not necessary - `List()` defines a *traitlet*, which uses Python's descriptor magic to create a regular Python list when the class is instantiated.

srinivasreddy:
@takluyver  Thanks.

PR #10314, opened 2017-02-21
update what's new in 5.3

Carreau:


takluyver:
I think for the next major release, I'd like to order the page the other way round, adding new minor releases at the bottom of the page instead of the top.

Carreau:
@meeseeksdev backport to 5.x

PR #10317, opened 2017-02-21
Suppress "not JSON strings" warning if `filename` argument is provided

gnestor:
I observed that this was not only affecting the `JSON` display class but any sub-class of `DisplayObject`. As a result, I would see a warning whenever trying to display data from a file:

screencap

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -655,7 +655,7 @@ class JSON(DisplayObject):
     @data.setter
     def data(self, data):
         if isinstance(data, str):
-            if getattr(self, 'filename') is None:
+            if getattr(self, 'filename', None) is None:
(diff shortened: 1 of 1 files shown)


Carreau:
Well, you can't do that, as DIsplayObject might not be text or JSON, they might be images, or actually any arbitrary files. so you have to assume they are raw data. You'll have to overwrite the behavior in the JSON case to do the right thing without affecting non-json formats.

gnestor:
Duh! Updating...

Carreau:
@gnestor I've pushed 2 extra commits to fix the test, does that suits you ?

gnestor:
Yes, merge away!

Carreau:
@meeseeksdev merge

From my phone, unsure if it works.

PR #10321, opened 2017-02-21
Update shim warnings to include IPython version.

### LINKED ISSUE #10143: Update ShimWarnings to include version numbers.
Updating ShimWarnings to include since when the modules have been moved would be helpful to indicate people to transition, or at least to inform better when they can stop using old import versions.

Carreau:
It will soon be 2 major versions, and we are dropping,
I'm considering either dropping shims, or insert a filter to make them
raise and let a last chance for users to upgrade before removing.

Closes #10143

--- 

Apparently I had this branch lying around, I'm astonished to not have opened a PR. 
Backport ?

### DIFF
--- a/IPython/frontend.py
+++ b/IPython/frontend.py
@@ -17,7 +17,7 @@ from warnings import warn
 
 from IPython.utils.shimmodule import ShimModule, ShimWarning
 
-warn("The top-level `frontend` package has been deprecated. "
+warn("The top-level `frontend` package has been deprecated since IPython 1.0. "
      "All its subpackages have been moved to the top `IPython` level.", ShimWarning)
 
 # Unconditionally insert the shim into sys.modules so that further import calls
--- a/IPython/config.py
+++ b/IPython/config.py
@@ -9,7 +9,7 @@ from warnings import warn
 
 from IPython.utils.shimmodule import ShimModule, ShimWarning
 
-warn("The `IPython.config` package has been deprecated. "
+warn("The `IPython.config` package has been deprecated since IPython 4.0. "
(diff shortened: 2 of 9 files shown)


IPython/frontend.py
```diff
@@ -17,7 +17,7 @@
 
 from IPython.utils.shimmodule import ShimModule, ShimWarning
 
-warn("The top-level `frontend` package has been deprecated. "
```
takluyver: This one was deprecated earlier, I think.

takluyver:
I'm happy for this to be backported, but I don't think there's any hurry to remove the shims. Libraries and tools often want to support several versions of IPython at once, so there's no need to make that harder for them.

IPython/frontend.py
Carreau: Indeed, the ShimWarning was introduced in 2c9013e4, which is rel-3.0.0-881-g2c9013e , but indeed, the shim is form 1.0 !

Carreau:
> I'm happy for this to be backported, but I don't think there's any hurry to remove the shims. Libraries and tools often want to support several versions of IPython at once, so there's no need to make that harder for them.

It's not a question of removing, It's a question of having good deprecations on python 2 as well. 
But anyway, I think it's fine as 6+ only.

Carreau:
> It's not a question of removing, It's a question of having good deprecations on python 2 as well.
> But anyway, I think it's fine as 6+ only.

Taking it back, I see you marked as 5.3. Fine with me as well.

takluyver:
Yup, I don't see any reason not to backport more informative warnings.

takluyver:
@meeseeksdev backport to 5.x

PR #10324, opened 2017-02-22
[Do not merge] FYI. Fix IPython on Python 3.7.

Carreau:
Recent Python 3.7 master broke the AST module.
Use this as a workaround if you are on 3.7.

See http://bugs.python.org/issue29622 for resolution upstream (if it get
resolved)

--- 

Just because I'm going to guess I might not be the only one and the error message is not obvious:

```
In [1]: import os
TypeError: Module constructor takes either 0 or 2 positional arguments
```

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2785,7 +2785,10 @@ class InteractiveShell(SingletonConfigurable):
 
         try:
             for i, node in enumerate(to_run_exec):
-                mod = ast.Module([node])
+                try:
+                    mod = ast.Module([node])
+                except TypeError:
+                    mod = ast.Module([node],'fake docstring workaround Python 3.7-master')
                 code = compiler(mod, cell_name, "exec")
                 if self.run_code(code, result):
                     return True


takluyver:
Thanks for the heads up. If 3.7 makes it to beta and it doesn't look like this is getting changed there, we'll merge a fix like this.

Carreau:
> Thanks for the heads up. If 3.7 makes it to beta and it doesn't look like this is getting changed there, we'll merge a fix like this.

We're still ~14month from release right :-) Hopefully I caught the bug 44 minutes after the offending patch was merged.

takluyver:
You're a human continuous integration service ;-)

Carreau:
> You're a human continuous integration service ;-)

nah, I was just playing with having a local conda-build of Python 3.7 that fake itself as a 3.6.99 so you can "just" `conda install python=3.6 --use-local --force`  and get a local HEAD of 3.7 with all the already available conda packages.

Carreau:
Fixed upstream.

Issue #9945, opened 2016-09-13
error in %run doesn't set ExecutionResult.success = False


minrk:
`run_cell` returns an ExecutionResult with a `success` flag to indicate whether the cell encountered an error. `%run` always results in success=True, whether there was an error or not, because it does its own error capture and display, and always returns successfully.

This is revealed now that ipykernel relies on `ExecutionResult().success` to determine if execution succeeded or not.

See https://github.com/jupyter/qtconsole/pull/141 for one symptom, though there is some strange protocol handling in the qtconsole as well that could further hide this bug.

PR #10328, opened 2017-02-23
[WIP] Don't swallow exception in %run

Carreau:
related to #9945 but will need exception to be truncated at some point
or we will see IPython internals.

--- 

Honestly that looks like a rat nest though.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2397,7 +2397,7 @@ class InteractiveShell(SingletonConfigurable):
         with self.builtin_trap:
             return eval(expr, self.user_global_ns, self.user_ns)
 
-    def safe_execfile(self, fname, *where, **kw):
+    def safe_execfile(self, fname, *where, exit_ignore=False, raise_exceptions=False, shell_futures=False):
         """A safe version of the builtin execfile().
 
         This version will never throw an exception, but instead print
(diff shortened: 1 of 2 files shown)


IPython/core/interactiveshell.py
```diff
@@ -2397,7 +2397,7 @@ def ev(self, expr):
         with self.builtin_trap:
             return eval(expr, self.user_global_ns, self.user_ns)
 
-    def safe_execfile(self, fname, *where, **kw):
+    def safe_execfile(self, fname, *where, exit_ignore=False, raise_exceptions=True, shell_futures=False):
         """A safe version of the builtin execfile().
 
         This version will never throw an exception, but instead print
```
takluyver: This docstring is now a bit... erroneous.

Carreau:
OMG it's a mess. 

I'm really unsure how to do it without changing tons of API to pass results around without changing the bowels of IPython behaves. Unless we accept that IPython shows extra stacktraces for run that show the inside of IPython, in which case it's just a matter of changing tests. 

@takluyver is there a way to re-raise omitting some stacks ?  We could do that (or generally add an option to IPython's showtraceback that skip traceback internal frames?)

takluyver:
Yeah, I think I suspected that this one was going to be tricky, and that's why I hadn't tried to fix it.

We have `UsageError` which magics can raise to fail with a one-line message rather than a full traceback. Maybe we should add an exception type like `SilentError` which they can use to indicate failure when they have already taken care of displaying the error details.

PR #10341, opened 2017-02-24
Remove the 'fresh clone' step of the release process

takluyver:
We already clean the repository in step 3. Since the big split, there are fewer complexities in the repo that can cause problems in packaging, and I think this step is overkill.

This is part of my effort to simplify the release procedure, as I feel like I'm spending a lot of time cutting releases nowadays.


Carreau:
@meeseeksdev backport to 5.x

Carreau:
Grumlgrumlgruml conflict Grumlgrumlgruml

PR #10350, opened 2017-02-27
Minor improvements to completion

### LINKED ISSUE #9835: completion should favor displaying the actually completed part of the string rather than the prefix
One screenshot says it all.
screenshot_20160804_003842

Carreau:
If object being completed is a function, append `()` to match the
convention used in cpython doc. This include methods. Append `()` only
in the _displayed_ menu, does not insert the ().

If completion is relatively to long (mostly if is does not come from
jedi), and use a fully qualified name with more than 3 parts, ellides
the middles one to `…`. (This closes #9835).

### DIFF
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -11,7 +11,7 @@ import unicodedata
 from wcwidth import wcwidth
 
 from IPython.core.completer import (
-    IPCompleter, provisionalcompleter, rectify_completions, cursor_to_position,
+    provisionalcompleter, cursor_to_position,
     _deduplicate_completions)
 from prompt_toolkit.completion import Completer, Completion
 from prompt_toolkit.layout.lexers import Lexer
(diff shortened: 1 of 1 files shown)


Carreau:
Here is how it looks like not using jedi:

screen shot 2017-02-27 at 14 38 22

It does not affect completions  that much when jedi is installed as jedi mostly returns all the completions that IPython finds with the right text:

screen shot 2017-02-27 at 14 39 13

the  `--jedi/ipython--` token is a debugging one you should not worry about. 

cc @anntzer

anntzer:
Works for me, thanks.

IPython/terminal/ptutils.py
```diff
@@ -21,6 +21,23 @@
 
 _completion_sentinel = object()
 
+def _ellide(string, *, min_ellide=30):
```
takluyver: It's `elide` with one L.

However, `ellipsis` (plural ellipses) with two Ls is correct. English, such a wonderfully consistent language ;-)

Carreau:
> However, ellipsis (plural ellipses) with two Ls is correct. English, such a wonderfully consistent language ;-)

Fixed.

Issue #10342, opened 2017-02-24
Authorisation for uploading releases


takluyver:
5.3 is once again not uploaded to the ipython.org server, because I've done the release from a computer whose SSH public key is not authorised on that server.

This is easy to fix in the short term, but I'd like to simplify this part of the release procedure so that we only need to give releasers access to Github and PyPI, not to an extra server that we only use for this.

I suggest that we:

- [ ] Fix uploading release files to Github releases
- [ ] Have a cron job/webhook/something on that server which finds the files on Github and downloads them to the correct location.

Carreau:
Can you retry to ssh ?

Carreau:
Also I think that if we could have travis do the release on tag, that would be great.

Carreau:
Au and autosubmit a PR to conda forge.

blink1073:
Auto-forge would be awesome. 👍

takluyver:
I have now succeeded in uploading the 5.3.0 files to the ipython.org server; thanks @Carreau (I assume you added my ssh key).

Carreau:
> (I assume you added my ssh key).

Yes, from github/takluyver.keys



> Auto-forge would be awesome. 👍

Auto-forge it shall be named if we write it then.

takluyver:
> Auto-forge it shall be named if we write it then.

Not "Forging aHEAD" ? :wink:

blink1073:
:picardfacepalm:

tqa236:
Hello all, it's been a while, is this issue fixed now?

PR #10355, opened 2017-02-28
Use Travis to upload to GitHub releases

takluyver:
See gh-10342

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,21 @@ after_success:
 matrix:
     allow_failures:
         - python: nightly
+
+before_deploy:
+    - rm -rf dist/
+    - python setup.py sdist
+    - python setup.py bdist_wheel
+
+deploy:
+    provider: releases
+    api_key:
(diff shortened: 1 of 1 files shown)


takluyver:
Milestoned for 6.0 - because IPython doesn't use universal wheels, this gets more complicated if we try to backport it to use with 5.x, because two different versions of Python would need to upload files.

Carreau:
Hum I thought we stopped uploading to GH since the submodules craziness
but +1..

Carreau:
Well actually, will the travis and locally built sdist ~~have the same sha1~~ be identical

takluyver:
I doubt they'll be identical. Even if we [use Travis to upload to PyPI](https://docs.travis-ci.com/user/deployment/pypi/) too, it looks like the PyPI deployment thing builds distributions itself, so I don't know if it would produce identical packages.

Carreau:
> I doubt they'll be identical. Even if we use Travis to upload to PyPI too, it looks like the PyPI deployment thing builds distributions itself, so I don't know if it would produce identical packages.

hum, ok. Fair enough. I'm happy with this for now. Merging ?

takluyver:
Let's see how it goes.

PR #10356, opened 2017-02-28
Cleanup and deprecates autoindent flags and options.

Carreau:
cc @takluyver

### DIFF
--- a/IPython/terminal/magics.py
+++ b/IPython/terminal/magics.py
@@ -84,8 +84,7 @@ class TerminalMagics(Magics):
         """Toggle autoindent on/off (deprecated)"""
         print("%autoindent is deprecated since IPython 5: you can now paste "
               "multiple lines without turning autoindentation off.")
-        self.shell.set_autoindent()
-        print("Automatic indentation is:",['OFF','ON'][self.shell.autoindent])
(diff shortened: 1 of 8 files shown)


IPython/terminal/magics.py
```diff
@@ -84,8 +84,7 @@ def autoindent(self, parameter_s = ''):
         """Toggle autoindent on/off (deprecated)"""
         print("%autoindent is deprecated since IPython 5: you can now paste "
               "multiple lines without turning autoindentation off.")
-        self.shell.set_autoindent()
-        print("Automatic indentation is:",['OFF','ON'][self.shell.autoindent])
+        print("Automatic indentation is deprecated and will have no effects.")
```
takluyver: I don't think we need two messages about the same deprecation.

takluyver:
I'm having trouble seeing tests on Travis at the moment, but the test failures on Appveyor look real.

Carreau:
@ivanov I'm going to guess we should not remove that if you'r going to bring back readline ?

unode:
I was looking for a way to disable autoindentation and found `--no-autoindent` but this option has no effect (IPython 6.1.0 & 6.2.1).
Should the deprecation in this issue also affect `--no-autoindent` and `--autoindent`? Those do not produce any warning.

PR #10357, opened 2017-02-28
Deprecate EmacsHandler as it is noop

Carreau:


takluyver:
And EmacsChecker too?

takluyver:
Looking at the file, I imagine we want to remove them from `_default_checkers` and `_default_handlers` at the bottom.

Carreau:
No, because some code still call self.gethandler('normal') and
self.getchecker('normal') I want to deprecate to CYA for the future removal
to be safe.

takluyver:
But then our own code is triggering the warnings (if filters are set to show them at that time). I don't think we should deprecate calling something that our own code is still calling.

There's no `getchecker()` method anyway. Maybe for `get_handler_by_name` we can issue a warning specifically if name is `'emacs'`.

Carreau:
Hum, I'm going to delay that then. I want to better understand the consequences.

Carreau:
Marking as 7.0 and "closed PR" tag to keep the queue clean.

PR #10365, opened 2017-03-01
Use $SHELL in _process_posix.py

xiapoy:
This PR follows the issue jupyter/notebook/issues/2242.

### DIFF
--- a/IPython/utils/_process_posix.py
+++ b/IPython/utils/_process_posix.py
@@ -22,6 +22,8 @@ import sys
 
 import pexpect
 
+import re
+
 # Our own
 from ._process_common import getoutput, arg_split
 from IPython.utils import py3compat
@@ -59,11 +61,18 @@ class ProcessHandler(object):
 
     @property
     def sh(self):
-        if self._sh is None:        
-            self._sh = pexpect.which('sh')
-            if self._sh is None:
-                raise OSError('"sh" shell not found')
-        
+        self._sh = os.environ['SHELL']
+        # ensure we got a clean shell from our environment
+        with open('/etc/shells', 'r') as s:
+            for line in s:
+                if re.match(r'(/[^/]+)+$', line):
+                    line = line.rstrip()
+                    if line == self._sh:
+                        return self._sh
+        # if shell not clean, defaults to /bin/sh
+        self._sh = pexpect.which('sh')
+        if self._sh is None:
+            raise OSError('"sh" shell not found')
         return self._sh
 
     def __init__(self, logfile=None, read_timeout=None, terminate_timeout=None):
@@ -131,7 +140,7 @@ class ProcessHandler(object):
         """
         # Get likely encoding for the output.
         enc = DEFAULT_ENCODING
-        
+
         # Patterns to match on the output, for pexpect.  We read input and
         # allow either a short timeout or EOF
         patterns = [pexpect.TIMEOUT, pexpect.EOF]


IPython/utils/_process_posix.py
```diff
@@ -59,11 +61,18 @@ class ProcessHandler(object):
 
     @property
     def sh(self):
-        if self._sh is None:        
-            self._sh = pexpect.which('sh')
-            if self._sh is None:
-                raise OSError('"sh" shell not found')
-        
+        self._sh = os.environ['SHELL']
```
ivanov: A new change like this can have unexpected consequences, we should either allow users to opt into respecting `$SHELL` or allow them to opt out of this new functionality.

IPython/utils/_process_posix.py
```diff
@@ -59,11 +61,18 @@ class ProcessHandler(object):
...
+        # ensure we got a clean shell from our environment
+        with open('/etc/shells', 'r') as s:
+            for line in s:
+                if re.match(r'(/[^/]+)+$', line):
+                    line = line.rstrip()
+                    if line == self._sh:
+                        return self._sh
+        # if shell not clean, defaults to /bin/sh
+        self._sh = pexpect.which('sh')
```
ivanov: All of the setting of `self._sh` logic should still be indented - we only want to set `self._sh` once, as before, not on every use of `.sh`.  It should only happen the first time this property is accessed, so please restore the `if self._sh is None:` above

IPython/utils/_process_posix.py
```diff
@@ -59,11 +61,18 @@ class ProcessHandler(object):
 
     @property
     def sh(self):
-        if self._sh is None:
```
ivanov: I repeat it below, but we should only set self._sh once, not on every call, so this line should be restored.

ivanov (changes requested):

naileakim:
I understand your comments. Here is another solution that may be more suitable. If sysadmins have been setting up a restricted shell for users, `self._sh` take the `$SHELL` environment variable as value, otherwise it falls back to the `sh` shell.

```python
    def sh(self):
        if self._sh is None:
            if os.environ['SHELL'] is pexpect.which('lshell') \
                    or os.environ['SHELL'] is pexpect.which('rbash') \
                    or os.environ['SHELL'] is pexpect.which('rksh'):
                self._sh = os.environ['SHELL']
            else:
                self._sh = pexpect.which('sh')
                if self._sh is None:
                    raise OSError('"sh" shell not found')

        return self._sh
```

What's your advice? Do you prefer to create an option for the user?

ivanov:
> What's your advice? 

I think this new version is to specific about possible uses, hard-coding only a few restricted shells has a code smell and is not the right direction. It would be enough to just use whatever the `$SHELL` variable is, and fall-back to the old `pexpect.which('sh')` for the rare case when `$SHELL`is not defined. I think this is what @takluyver signaled also back in #2242, is that right, Thomas?

> Do you prefer to create an option for the user?

Personally, I think it's a virtue for powerful general-purpose tools like ipython to either be backwards-compatible or to provide path for backward compatibility when rolling out new features. I realize that there can be tension between user experience and software complexity, and I'd much rather focus on keeping current users happy and giving them choices than keeping software pristine. But I also see it from the other side, having every possible thing be configurable can make it overwhelming for new users to get started and also difficult for developers who would have to consider the myriad of possibilities in a pathologically configurable system. A maze of tiny passages all alike.

In this particular case, since this does not apply to windows users at all, it feels like it's not worth making this configurable, and just using SHELL unconditionally is fine. Please add a new file in the [docs/whatsnew/pr](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr) directory documenting this change (there's a README there).

Thank you for taking the time to make IPython better.

takluyver:
Short phone email, but yes, that's what I was thinking - use $SHELL if set,
and sh as a fallback.

naileakim:
Thanks for your answers. As far as I understood, this is what you have in mind, right?

```python
@property
def sh(self):
    if self._sh is None:
        # we use $SHELL if set
        _env_shell = os.environ['SHELL']
        if _env_shell is not None:
            self._sh = _env_shell
        # POSIX sh as a fallback
        else:
            self._sh = pexpect.which('sh')
            if self._sh is None:
                raise OSError('"sh" shell not found')
        # we only setup the property once
        return self._sh
```

In my opinion, using whatever the `$SHELL` variable could be a security vulnerability. That's why my previous solutions check that the given `$SHELL` variable is clean. If this change is suitable for you, I will submit my PR and document it in [docs/whatsnew/pr](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr).

minrk:
> In my opinion, using whatever the $SHELL variable could be a security vulnerability.

In what way? IPython can already execute arbitrary code, so anything the user is able to do on the system, they are already able to do from IPython. Changing what `!` invokes via an environment variable doesn't seem like a security issue, especially when the same thing can be accomplished directly via Python code.

We should be careful with a change like this, because using `$SHELL` means we are implicitly assuming that all possible shells behave the same under `$SHELL -c 'cmd'`. I'm not sure that's true, and we should test some cases when users have shells like xonsh, fish, tcsh, etc.

Relying on `$SHELL` to lock down `!` also won't work, because the user can always do:

```python
os.environ['SHELL'] = '/bin/bash'
```

before invoking `!`.

takluyver:
We already rely on `$SHELL -c ...` for `system_raw`, which implements `!` in terminal IPython, so I think it should be OK to use it for `system_piped`, which is `!` in the kernel, as well.

takluyver:
@kimeaa are you still interested in updating this?

ivanov:
Hey there, I'm going through old issues and it seems to me that it makes sense to close this one. There's been several months of inactivity here - feel free to re-open or open a new PR if you end up picking it up again.

Thanks everyone and happy hacking! :bowtie:

PR #10369, opened 2017-03-02
Fixed #10322 - Made autocall to ignore string/byte literals

### LINKED ISSUE #10322: %autocall is conflicting with string literals
```
In [15]: name = 'foo'

In [16]: %autocall 0
Automatic calling is: OFF

In [17]: f'{name}'
Out[17]: 'foo'

In [18]: %autocall 1
Automatic calling is: Smart

In [19]: f'{name}'
-------> f('{name}')
sample function with ('{name}',), {}

In [20]: f '{name}'
-------> f('{name}')
sample function with ('{name}',), {}
```

```
→ ipython --version
5.2.2

→ python --version 
Python 3.7.0a0
```

ChillarAnand:
fixed #10322

### DIFF
--- a/IPython/core/prefilter.py
+++ b/IPython/core/prefilter.py
@@ -510,6 +510,12 @@ class AutocallChecker(PrefilterChecker):
         if not oinfo['found']:
             return None
 
+        ignored_funs = ['b', 'f', 'r', 'u', 'br', 'rb', 'fr', 'rf']
+        ifun = line_info.ifun
+        line = line_info.line
+        if ifun.lower() in ignored_funs and (line.startswith(ifun + "'") or line.startswith(ifun + '"')):
+            return None
+
         if callable(oinfo['obj']) \
                and (not self.exclude_regexp.match(line_info.the_rest)) \
                and self.function_name_regexp.match(line_info.ifun):
@@ -625,7 +631,7 @@ class AutoHandler(PrefilterHandler):
         elif esc == ESC_PAREN:
             newcmd = '%s(%s)' % (ifun,",".join(the_rest.split()))
         else:
-            # Auto-paren.       
+            # Auto-paren.
             if force_auto:
                 # Don't rewrite if it is already a call.
                 do_rewrite = not the_rest.startswith('(')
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, but I don't think this is quite the right fix; if there's something like `f 1`, it should still use autocall - especially as `f` is a common name for a function.

takluyver:
I think the better place to fix it is in `IPython.core.splitinput.split_user_input` - 

```
 2│ from IPython.core.splitinput import split_user_input

 3│ split_user_input("'foo'")
 3> ('', '', '', "'foo'")

 4│ split_user_input("f'foo'")
 4> ('', '', 'f', "'foo'")

 5│ split_user_input("12")
 5> ('', '', '12', '')
```

ChillarAnand:
```
In [2]: split_user_input("r 'a'")
Out[2]: ('', '', 'r', "'a'")

In [3]: split_user_input("r'a'")
Out[3]: ('', '', 'r', "'a'")
```

`split_user_input` wont be able to differentiate above cases. We should probably check for space?

takluyver:
I think checking for space is probably reasonable. In principle at the moment you could do `d'foo'` and it gets transformed to `d('foo')`, but that doesn't work with names or numbers, and it seems a bit pathological to do it with a string.

ChillarAnand:
@takluyver Thanks for the review. Pushed changes.

takluyver:
Thanks. I think this is close, but I still see a few issues:

- Some (not all) combinations of prefixes are valid - e.g. `br'123'`.
- Rather than checking for a space, let's explicitly check if it looks like the start of a string, e.g. `line.startswith((ifun+"'", ifun+'"')).

ChillarAnand:
@takluyver Good catch. We should check for `br` & `fr`.

IPython/core/prefilter.py
```diff
@@ -510,6 +510,12 @@ def check(self, line_info):
         if not oinfo['found']:
             return None
 
+        ignored_funs = ['b', 'f', 'r', 'u', 'br', 'rb', 'fr', 'rf']
+        ifun = line_info.ifun
+        line = line_info.line
+        if ifun.lower() in ignored_funs and (line.startswith(ifun + "'") or line.startswith(ifun + '"')):
```
takluyver: Fun fact: startswith and endswith can both take a tuple of strings, so `s.startswith((p1, p2))` is equivalent to `s.startswith(p1) or s.startswith(p2)`. That's a minor point, though, so I won't hold this PR up for it.

takluyver:
Thanks for this, merging :-)

Issue #10014, opened 2016-10-18
Console reverse search broken


maksay:
When trying to use reverse-search in `ipython console` / `jupyter console`, I am getting the following error:

```
I-search backward: Traceback (most recent call last):
  File "/Users/andriimaksai/anaconda/bin/jupyter-console", line 5, in <module>
    app.main()
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/traitlets/config/application.py", line 596, in launch_instance
    app.start()
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/jupyter_console/app.py", line 151, in start
    self.shell.mainloop()
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/jupyter_console/ptshell.py", line 463, in mainloop
    self.interact()
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/jupyter_console/ptshell.py", line 448, in interact
    code = self.prompt_for_code()
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/jupyter_console/ptshell.py", line 408, in prompt_for_code
    reset_current_buffer=True)
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/prompt_toolkit/interface.py", line 399, in run
    self._redraw()
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/prompt_toolkit/interface.py", line 346, in _redraw
    self.renderer.render(self, self.layout, is_done=self.is_done)
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/prompt_toolkit/renderer.py", line 429, in render
    extended_height=size.rows,
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/prompt_toolkit/layout/containers.py", line 142, in write_to_screen
    sizes = self._divide_heigths(cli, write_position)
  File "/Users/andriimaksai/anaconda/lib/python2.7/site-packages/prompt_toolkit/layout/containers.py", line 177, in _divide_heigths
    dimensions = [get_dimension_for_child(c, index) for index, c in enumerate(self.children)]
[...]

PR #10371, opened 2017-03-02
Attempt at fixing assertion from reverse search

takluyver:
Trying to fix gh-10014, though I can't reproduce the problem to test it.

PR against 5.x because I think this is only needed for Python 2.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -239,7 +239,7 @@ class TerminalInteractiveShell(InteractiveShell):
             # Ignore blank lines and consecutive duplicates
             cell = cell.rstrip()
             if cell and (cell != last_cell):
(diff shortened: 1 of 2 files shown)


IPython/terminal/interactiveshell.py
```diff
@@ -239,7 +239,7 @@ def prompt():
             # Ignore blank lines and consecutive duplicates
             cell = cell.rstrip()
             if cell and (cell != last_cell):
-                history.append(cell)
+                history.append(cast_unicode_py2(cell))
```
minrk: Should this be done at `history_manager.get_tail`?

takluyver: Done. I'd still like someone to validate that this actually fixes the issue, though...

PR #10372, opened 2017-03-02
Fix error in completer when the cursor position is at the start of a string.

ian-r-rose:
We should only do some inference based on the previous character if the previous character exists.


Carreau:
Oops, thanks.

Issue #10364, opened 2017-03-01
Bring back readline capability?


ivanov:
Even if it's not what is used by default, I've experienced and heard about enough other people experiencing pains with the prompt_toolkit-based input breaking people's workflows, that I think we should seriously consider bringing that code back in.

I think it was a mistake to remove it altogether, and change the default at the same time.

- [recent discussion on the Sage mailing list about this](https://groups.google.com/forum/#!topic/sage-devel/t2QucyHLQXg)
- #9816 has several users reporting breakage
- there are several users brought up prompt toolkit having regressions compared to readline in #9388 
-  and similarly #10075 contains sentiment from a user saying this:  "Multiline editing is a very nice feature. As a longtime vim and bash user, though, the switch to IPython 5.x and prompt_toolkit has been jarring."
- #10085 lists a limitation and inconsistency of prompt_toolkit which [will not be addressed upstream](https://github.com/jonathanslenders/python-prompt-toolkit/issues/432#issuecomment-266311715).
- #9944 is another issue that came up with the prompt-toolkit transition, which used to work in old readline based code, (though I understand Thomas recently proposed a fix for this in #10185, there is discussion there about the limitations of that, too).
- #10211 is an emacs user affected a change that as far as I can tell is also attributable to the transition to prompt toolkit.
- #10315 is another unexpected breakage in IPython 5, because prompt_toolkit uses a separate thread for completions (whereas completions had to be synchronous in readline, so jpype completions worked fine in IPython <= 4). 
- #9948 - user unable to paste multi-line using tmux.
- #10071 - randomly missing prompt after upgrade to IPython 5 inside docker (because terminal size was reported as 0x0)

ivanov:
I'm willing to do the work to get this back in time for 6.0, so I'll tag #10329 here.

ivanov:
I think this is a particularly important issue because long time users and advocates of IPython are finding themselves frustrated by or even unwilling to use IPython 5 as it is right now.

Carreau:
I'm happy to add that back as an option, though most of the code has been removed and that will likely be a lot of efforts to reintegrate readline.

Carreau:
If we can figure out how to make that optional (so that it can evolve faster than than core to re-fix potential bug that were introduced by the revival, that would be great.

Carreau:
We might want to delay #10356 if we bring RL back.

Carreau:
See #9260 as well.

Carreau:
[...]

PR #10373, opened 2017-03-02
Make the TerminalInteractiveShell class configurable.

Carreau:
Should allow custom frontend in separate packages.

cc @takluyver, @ivanov 

--- 

That's sufficient for me to get the basics working. 

rlipython

### DIFF
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -33,7 +33,7 @@ from IPython.extensions.storemagic import StoreMagics
 from .interactiveshell import TerminalInteractiveShell
 from IPython.paths import get_ipython_dir
 from traitlets import (
-    Bool, List, Dict, default, observe,
+    Bool, List, Dict, default, observe, Type
 )
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


takluyver:
As I said in the other issue, I wouldn't do this - I would subclass TerminalIPythonApp in `rlipython` and override `init_shell` there, so you start it with `rlipython`, not by configuring `ipython`.

Carreau:
> As I said in the other issue, I wouldn't do this - I would subclass TerminalIPythonApp in rlipython and override init_shell there, so you start it with rlipython

That implies that all IPython wrapers and extensions that exists need to be updated to take an option instead of allowing it to be done globally. That seem like quite the task, and the annoyance for users.

ivanov:
>As I said in the other issue, I wouldn't do this - I would subclass TerminalIPythonApp in rlipython and override init_shell there, so you start it with rlipython, not by configuring ipython.

Why weld something when it could be swappable? 

Please consider why we're doing this in the first place: IPython 5.0 broke a bunch of people's workflows when it unilaterally moved away from readline to prompt toolkit without giving users the ability to fall back to old behavior. #10364 has a list of many of them, and proposed bringing back that code into IPython, which you (@takluyver) are strongly against. @Carreau is already making the compromise of not putting that code back into IPython, but instead proposes a straightforward ability for users who relied on the old behavior to install a package, set a config, and continue using IPython.

IPython/terminal/ipapp.py
```diff
@@ -183,6 +183,13 @@ class TerminalIPythonApp(BaseIPythonApplication, InteractiveShellApp):
     flags = Dict(flags)
     aliases = Dict(aliases)
     classes = List()
+    
+    interactive_shell_class = Type(
+        klass=object,   # use default_value otherwise which only allow subclasses.
+        default_value=TerminalInteractiveShell,
+        help="Class to use to instantiate the TerminalInteractiveShell object. Usefull for custom Frontends"
```
jasongrout: "Useful"

Carreau:
@meeseeksdev backport

PR #10389, opened 2017-03-09
[doc] remove unused look_in_string from inputtransform doc

minrk:
cc @Carreau


Carreau:
Fine with me. 

Pinging @takluyver just in case.

Carreau:
Merging, we can revert later is needed.

PR #10390, opened 2017-03-10
Await repl

Carreau:
```
In [1]: import aiohttp

In [2]: res = aiohttp.get('https://api.github.com')

In [3]: res
Out[3]: <aiohttp.client._DetachedRequestContextManager at 0x112a28828>

In [4]: resp = await res

In [5]: resp
Out[5]:
<ClientResponse(https://api.github.com) [200 OK]>
<CIMultiDictProxy('Server...DC')>

In [6]: print(await resp.json())
{'current_user_url': 'https://api.github.com/user', 'current_user_authorizations_html_url': 'https://g...
```

Thanks to useful discussion with @njsmith and @minrk. cc @takluyver 

As await is not valid at top level, and extracting ast, or other modification of `__code__` don't work, 
when `SyntaxError` and wraping in `async def` is not syntax error use that.

Send `globals()` into the function, run the function, extract `locals()`, and update `user_ns`

### DIFF
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -14,9 +14,9 @@ import ast
 import sys
 from textwrap import dedent, indent
 
-def _asyncio_runner(function, user_ns):
+def _asyncio_runner(coro):
     import asyncio
-    return asyncio.get_event_loop().run_until_complete(function(**user_ns))
+    return asyncio.get_event_loop().run_until_complete(coro)
 
 def _curio_runner(function, user_ns):
     import curio
@@ -49,13 +49,12 @@ def _asyncify(code):
      - ``loop_runner`` is use to send the loop_runner _in_
      - ``last_expression`` to get the last expression _out_
     """
-    return dedent("""
+    res = dedent("""
         async def phony():
-        {usercode}
-            return locals()
-        interm = loop_runner(phony, user_ns)
-        user_ns, last_expr =  interm
-        """).format(usercode=indent(code,' '*4))
+            {usercode}
+            locals()
+        """).format(usercode=indent(code,' '*4)[4:])
+    return res
 
 def _should_be_async(cell:str) -> bool:
     """Detect if a block of code need to be wrapped in an `async def`
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -1,0 +1,70 @@
+"""
+Async helper function that are invalid syntax on Python 3.5 and below.
+"""
+
+import ast
+import sys
+from textwrap import dedent, indent
+
+def _asyncio_runner(function, user_ns):
+    import asyncio
+    return asyncio.get_event_loop().run_until_complete(function(**user_ns))
+
+def _curio_runner(function, user_ns):
+    import curio
+    return curio.run(function(**user_ns))
+
+if sys.version_info > (3,5):
+    # nose refuses to avoid this file.
+    s =dedent('''
+    def _trio_runner(function, user_ns):
+        import trio
+        async def loc(fun, user_ns):
+            """
+            We need the dummy no-op async def to protect from
+            trio's internal. See https://github.com/python-trio/trio/issues/89
+            """
+            return await fun(**user_ns)
+        return trio.run(loc, function, user_ns)
+    ''')
+    exec(s, globals(), locals())
+
+def _asyncify(code):
+    """wrap code in async def definition.
+
+    And setup a bit of context to run it later.
+
+    The following names are part of the API:
+
+     - ``user_ns`` is used as the name for the namespace to run code in, both
+     _in_ and _out_.
+     - ``loop_runner`` is use to send the loop_runner _in_
+     - ``last_expression`` to get the last expression _out_
+    """
+    return dedent("""
+        async def phony():
+        {usercode}
+            return locals()
+        interm = loop_runner(phony, user_ns)
+        user_ns, last_expr =  interm
+        """).format(usercode=indent(code,' '*4))
+
+def _should_be_async(cell:str) -> bool:
+    """Detect if a block of code need to be wrapped in an `async def`
+
+    Attempt to parse the block of code, it it compile we're fine.
+    Otherwise we  wrap if and try to compile.
+
+    If it works, assume it should be async. Otherwise Return False.
+    """
+    try:
+        ast.parse(cell)
+        return False
+    except SyntaxError:
+        try:
+            ast.parse(_asyncify(cell))
+        except SyntaxError:
+            return False
+        return True
+    return False
+
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -1,6 +1,14 @@
 """
 Async helper function that are invalid syntax on Python 3.5 and below.
+
+Known limitation and possible improvement. 
+
+Top level code that contain a return statement (instead of, or in addition to
+await) will be detected as requiring being wrapped in async calls. This should
+be prevented as early return will not work.
 """
+
+
 
 import ast
 import sys
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -16,7 +16,7 @@ def _curio_runner(function, user_ns):
 
 if sys.version_info > (3,5):
     # nose refuses to avoid this file.
-    s =dedent('''
+    s = dedent('''
     def _trio_runner(function, user_ns):
         import trio
         async def loc(fun, user_ns):
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -14,20 +14,21 @@ import ast
 import sys
 import inspect
 from textwrap import dedent, indent
+from types import CodeType, CoroutineType
 
-def _asyncio_runner(coro):
+def _asyncio_runner(coro:CoroutineType):
     """
     Handler for asyncio autoawait
     """
     import asyncio
     return asyncio.get_event_loop().run_until_complete(coro)
 
-def _curio_runner(function):
+def _curio_runner(coroutine:CoroutineType):
     """
     handler for curio autoawait
     """
     import curio
-    return curio.run(function)
+    return curio.run(coroutine)
 
 if sys.version_info > (3,5):
     # nose refuses to avoid this file and async def is invalidsyntax
@@ -44,20 +45,13 @@ if sys.version_info > (3,5):
     ''')
     exec(s, globals(), locals())
 
-def _asyncify(code):
+def _asyncify(code:str) -> str:
     """wrap code in async def definition.
 
     And setup a bit of context to run it later.
-
-    The following names are part of the API:
-
-     - ``user_ns`` is used as the name for the namespace to run code in, both
-     _in_ and _out_.
-     - ``loop_runner`` is use to send the loop_runner _in_
-     - ``last_expression`` to get the last expression _out_
     """
     res = dedent("""
-        async def phony():
+        async def ___wrapper___():
             {usercode}
             locals()
             return None
@@ -71,7 +65,11 @@ def _should_be_async(cell:str) -> bool:
     Otherwise we  wrap if and try to compile.
 
     If it works, assume it should be async. Otherwise Return False.
+
+    Not handled yet: If the block of code has a return statement as  the top
+    level, it will be seen as async. This is a know limitation.
     """
+
     try:
         ast.parse(cell)
         return False
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -18,22 +18,22 @@ def _asyncio_runner(coro):
     import asyncio
     return asyncio.get_event_loop().run_until_complete(coro)
 
-def _curio_runner(function, user_ns):
+def _curio_runner(function):
     import curio
-    return curio.run(function(**user_ns))
+    return curio.run(function)
 
 if sys.version_info > (3,5):
     # nose refuses to avoid this file.
     s = dedent('''
-    def _trio_runner(function, user_ns):
+    def _trio_runner(function):
         import trio
-        async def loc(fun, user_ns):
+        async def loc(coro):
             """
             We need the dummy no-op async def to protect from
             trio's internal. See https://github.com/python-trio/trio/issues/89
             """
-            return await fun(**user_ns)
-        return trio.run(loc, function, user_ns)
+            return await coro
+        return trio.run(loc, function)
     ''')
     exec(s, globals(), locals())
 
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -12,18 +12,25 @@ be prevented as early return will not work.
 
 import ast
 import sys
+import inspect
 from textwrap import dedent, indent
 
 def _asyncio_runner(coro):
+    """
+    Handler for asyncio autoawait
+    """
     import asyncio
     return asyncio.get_event_loop().run_until_complete(coro)
 
 def _curio_runner(function):
+    """
+    handler for curio autoawait
+    """
     import curio
     return curio.run(function)
 
 if sys.version_info > (3,5):
-    # nose refuses to avoid this file.
+    # nose refuses to avoid this file and async def is invalidsyntax
     s = dedent('''
     def _trio_runner(function):
         import trio
@@ -75,4 +82,3 @@ def _should_be_async(cell:str) -> bool:
             return False
         return True
     return False
-
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -14,16 +14,16 @@ import ast
 import sys
 import inspect
 from textwrap import dedent, indent
-from types import CodeType, CoroutineType
+from types import CodeType
 
-def _asyncio_runner(coro:CoroutineType):
+def _asyncio_runner(coro):
     """
     Handler for asyncio autoawait
     """
     import asyncio
     return asyncio.get_event_loop().run_until_complete(coro)
 
-def _curio_runner(coroutine:CoroutineType):
+def _curio_runner(coroutine):
     """
     handler for curio autoawait
     """
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -53,6 +53,7 @@ def _asyncify(code):
         async def phony():
             {usercode}
             locals()
+            return None
         """).format(usercode=indent(code,' '*4)[4:])
     return res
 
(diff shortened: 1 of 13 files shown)


minrk:
This is super cool, @Carreau!

Carreau:
I think I might be able to execute the outer function in a separate
namespace to avoid potential name conflicts and conditional clearing of
user_ns. I'd like to have a clean way of still going through AST
transformers. I'm thinking of setting a flag, extraction the inner ADT of
the async function and shove it back in before exec.

Both of these should also allow to set your own runner without string
manipulation.

That's delicate surgery in the core of IPython though.

Carreau:
I'm thinking of protecting it behind a flag, (on by default), so that you can disable the heuristic. 
I need to test with curio/trio and add a way to change the loop.

Carreau:
Ok, now compatible with `trio` and `curio`:

```
$ ipython --InteractiveShell.looprunner=IPython.core.interactiveshell._trio_runner
Python 3.6.0 | packaged by conda-forge | (default, Feb 10 2017, 07:08:35)
IPython 6.0.0.dev -- An enhanced Interactive Python. Type '?' for help.
```
```python
In [1]: async def child1():
   ...:     print("   child 1 go to sleep")
   ...:     await trio.sleep(1)
   ...:     print("child 1 wakes up")

In [2]: async def child2():
   ...:     print("   child 2 go to sleep")
   ...:     await trio.sleep(2)
   ...:     print("child 2 wakes up")
   ...:

In [3]: import trio

In [4]: print('parent start')
   ...: async with trio.open_nursery() as n:
   ...:     n.spawn(child1)
   ...:     n.spawn(child2)
   ...: print('joined')
parent start
   child 2 go to sleep
   child 1 go to sleep
child 1 wakes up
child 2 wakes up
joined
```

gnestor:
Nice @Carreau! I have taken a shot at implementing this in [jp-babel](https://github.com/n-riesco/jp-babel) using [babel-plugin-top-level-await](https://github.com/eponymous-labs/babel-plugin-top-level-await) but no luck yet. If I get a chance, I will revisit this and share any findings...

takluyver:
Did you look at how I did the transformation to run the code in the prototype [asyncio magic](https://github.com/Gr1N/asyncio-ipython-magic/blob/master/asynciomagic.py), now adopted by @Gr1N? I haven't thought through the implications in detail, but pulling the function body out after transformation feels neater than pushing namespaces into functions.

njsmith:
The problem is that for a library like trio, we really want a single call
to trio.run() wrapped around the whole cell, because each call basically
creates its own event loop. Consider a cell like

print("start:", trio.current_time())
async with trio.open_nursery() as nursery:
    nursery.spawn(...)

There are two showstopper issues:
- trio.current_time() is syntactically synchronous, but it can only be used
inside a call to trio.run (it uses some thread local storage thing to fetch
the current run's clock object).

- if we transform the async with to something like

# highly simplified:
tmp = trio.open_nursery()
trio.run(tmp.__aenter__)
...
trio.run(tmp.__aexit__)

then we will create the nursery in one run loop, and clean it up in a
different run loop. This is not going to work out well, as __aexit__ sits
and waits for tasks that its loop isn't even running...

njsmith:
... On the other hand tho I just realized an obvious place where the
shoving-locals-in-and-out thing doesn't work: closures.

I think with the current implementation we have:

>>> a = 1
>>> def blah():
...     print(a)
>>> blah()
1
>>> a = 2
>>> blah()  # should print 2
1

?

Carreau:
>> blah()  # should print 2

well that seem to work:

```
$ ipython
IPython 6.0.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: a = 1

In [2]: async def blah():
   ...:     print(a)
   ...:

In [3]: await blah()
1

In [4]: a = 2

In [5]: await blah()
2
```

@takluyver I didn't found it, but it's an approach I tried. The other problem with this approach is that your await call are now blocking (are each are in a loop) and my initial need to write this was to actually make use of the concurrency to fetch parallel URL in the REPL. 

I (with @njsmith  help ) went a couple of other direction (rewriting codeobjects, trying to exec only the inner body of a function) all of them a dead end.

Carreau:
> Nice @Carreau! I have taken a shot at implementing this in jp-babel using babel-plugin-top-level-await  but no luck yet. If I get a chance, I will revisit this and share any findings...

At least JS have a default event loo, so that's solved :-)

takluyver:
> The other problem with this approach is that your await call are now blocking

It should only block on `await` expressions in the top-level code, which is necessary to ensure that that is executed in the correct order. `await` expressions inside function definitions should still work as normal, so you can still use something like [asyncio.gather()](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) to run coroutines concurrently.

Carreau:
> It should only block on await expressions in the top-level code, which is necessary to ensure that that is executed in the correct order.

Fair enough, I'm unsure how the AST should rewrite that though:

```
In [18]: async def child(i):
    ...:     print("   child %s  go to sleep"%i)
    ...:     await trio.sleep(2)
    ...:     print("   child %s  wakes up"%i)

In [19]: print('parent start')
    ...: async with trio.open_nursery() as n:
    ...:     for i in range(5):
    ...:         n.spawn(child, i)
    ...: print('parent end')
```

There is no await (in the second block) and top-level `async with` will be invalid at exec-time as well. 

I'm guessing we can push the rewriting further though, I would be happy to give it a try (unsure how to treat async-with, and then you need to care about async-for as well ? and other constructs ?. I would like to not bake-in asyncio though. and in the end we should try to push for supporting this upstream instead of the crazy hack we have now.

takluyver:
Ah, right, I didn't consider `await with` or `await for`.

njsmith:
> In [5]: await blah()

I think this is working because your code doesn't "asyncify" the cell containing the `async def`? What if you do something silly like:
```python
In [5]: await asyncio.sleep(0)
  ...: def blah():
  ...:     print(a)
```
and then call `blah`?

(The thing I'm concerned about is that `blah` will become a closure that captures the value of `a` *inside the fake async function's namespace*, and will keep referring to *that* `a` even after the fake function returns.)

Carreau:
> (The thing I'm concerned about is that blah will become a closure that captures the value of a inside the fake async function's namespace, and will keep referring to that a even after the fake function returns.)

Yep, you are right, it does capture `a`. 

I'm wondering how much of on issue this is. The other possibility is to not inject anything in locals, try to mark all names as globals in the function scope, and mess-up with the ast outside of the async-def function so that all names are invalid python identifiers (I think we can do that). We would still need to update globals() with the locals created inside the function I guess. Or maybe we can figure those out and inject them as global before ?

njsmith:
BTW the `trio.current_time` thing reminds me that it really would be nice (at least for the trio version of this) to run all cells inside the async context, and to use a single async context for everything. I guess this can be done without *too* much horribleness within the framework you have now, by doing something like:

```python
import trio, threading
class TrioThread:
    def __init__(self):
        self._await_in_trio_thread = None
        self._run_in_trio_thread = None
        self._shut_down_trio_thread_event = None
        self._have_trio_things = threading.Event()
        self._trio_thread = threading.Thread(target=trio.run, args=(self._trio_main,))
        self._trio_thread.start()
        self._have_trio_things.wait()

    async def _trio_main(self):
        self._await_in_trio_thread = trio.current_await_in_trio_thread()
        self._run_in_trio_thread = trio.current_run_in_trio_thread()
        self._shut_down_trio_thread_event = trio.Event()
        self._have_trio_things.set()
        await self._shut_down_trio_thread_event.wait()

    def run_async_fn(self, async_fn):
        return self._await_in_trio_thread(async_fn)

    def close(self):
        self._run_in_trio_thread(self._shut_down_trio_thread_event.set())
        self._trio_thread.join()
```

and the idea is that you instantiate one of these when starting the REPL, and then instead of passing the asyncified thunks to `trio.run` you pass them to `trio_thread.run_async_fn`. And do this for all cells, regardless of whether they have a visible `await`.

njsmith:
@Carreau: will you around this afternoon? I have to run to a dr appt, but I'm actually wondering if we should revisit the ctypes hackery...

Carreau:
> @Carreau: will you around this afternoon? I have to run to a dr appt, but I'm actually wondering if we should revisit the ctypes hackery...

I should be.

Carreau:
> and the idea is that you instantiate one of these when starting the REPL, and then instead of passing the asyncified thunks to trio.run you pass them to trio_thread.run_async_fn. And do this for all cells, regardless of whether they have a visible await.

I'll be happy to get a prototype of this. At least, get the right hooks in IPython to be able to hook whatever is needed later on. I'm thinking this is enough of a complex problem that we might want to postpone the hooks for 6.1.

Carreau:
So @njsmith was right, there is another dance we can do, which fix the capturing problem (apparently):

```
In [9]: a = 1

In [10]: await asyncio.sleep(0)
    ...: def blah():
    ...:     print(a)

In [11]: blah()
1

In [12]: a = 2

In [13]: blah()
2
```

This implies:
  - wrap user code in `async def dummy` append `locals()` at the end to force a dict update (fast to local ?)
  - `exec` this in `user_ns`
  - pull out `dummy.__code__`
  - patch `__code__` flags with `new_code = __code__ & ~ CO_NEWLOCALS = 0x0002` (makes locals() and globals() the same)
  - `eval(new_code, user_ns)` (return a coroutine)
  - use the set `loop_runner` on the `coroutine`

I have a locally working version, but I need to figure out how to rewrite things to still display the last result of async cells.

Carreau:
Ok, almost complete refactor using @njsmith advices. Looks and feel really nice. Almost not even a crazy hack.

vidartf:
I'm a little curious about timings for certain use-cases:

```python
In [5]: async def child(i):
   ...:     print("child %s  goes to sleep"%i)
   ...:     await asyncio.sleep(200)
   ...:     print("child %s  wakes up"%i)
```

```python
In [6]: first = asyncio.ensure_future(child(1))
```

Here, the user goes for coffee, and when filled with caffeine, returns and runs:

```python
In [7]: await first
```

When will `first` actually be run? After `[7]` is executed? Or while getting coffee? Would this change if tornado was using an asyncio loop? I'm not able to see this directly from updated docs (and don't know the code well enough to see it from there).

Carreau:
Thanks @vidartf for the question. 

There is 2 way to find out: 
 - Theoretically, what would the Python REPL do? Well [5] and [6] are valid in plain Python repl, and plain Python would be happy. So IPython has not reason to change the behavior. In pure Python you couldn't schedule the coroutine as there is no default eventloop. So the future will not scheduled to run. [7] is invalid syntax in plain Python, so I expect [7] to trigger the awaiting of the future.
 -  Try it

```
In [5]: import asyncio
   ...: async def child(i):
   ...:     print("child %s  goes to sleep"%i)
   ...:     await asyncio.sleep(2)
   ...:     print("child %s  wakes up"%i)
   ...:     return 'foo'

In [6]: first = asyncio.ensure_future(child(1))

In [7]: await first
child 1  goes to sleep
child 1  wakes up
Out[7]: 'foo'
```


I've tried to convey this in the following paragraph:

```
 +As with usual Python semantic, the awaitables are started only when awaited for the
 +first time. That is to say, in first example, no network request is done between
 +``In[1]`` and ``In[2]``.
```

THe other thing to keep in mind is that there is not a single loop running, there (likely) a new loop starting and stopping every time you attempt to use await-ish code at tom level. Tornado is not involved in the above, I haven't written an integration for tornado loop, but that would likely require still tornado to start/stop after every execution. 

My guess is that what you are looking for is `async-exec` it's a long term plan, but it will be hard to do before dropping 3.3 / 3.4 as we need to convert the full internal machinery of IPython to be asynchronous. Right now there is only synchronous execution of asynchronous code. 

Does that answer some of your questions ?

vidartf:
@Carreau Thanks for pointing out the relevant doc paragraph. Now I see it! Maybe it could make it clearer that `In[1]` refers to the top-most one with the `result = aiohttp.get('https://api.github.com')` line?

Regarding the Tornado part, I asked because I was playing around with asyncio in the notebook recently, and discovered that Tornado 5.0 will use asyncio for the main loop by default (if it is available). To test this out, I made a python kernel with an asyncio loop, see the gist here: https://gist.github.com/vidartf/ca84910380b4ed819a4d5df749ab2c83

If I start such a kernel, and then run the code above, I get this output:

```python
In[1]: import asyncio

In[2]: async def child(i):
  ...:    print("child %s  goes to sleep"%i)
  ...:    await asyncio.sleep(2)
  ...:    print("child %s  wakes up"%i)
  ...:    return 'foo'

In[3]: first = asyncio.ensure_future(child(1))  # wait a little after this line

child 1  goes to sleep
child 1  wakes up

In[4]: first.result()

Out[4]: 'foo'
```

If this is unwanted behavior, I think it may be avoided by implementing a smart `get_event_loop` on the event loop policy. This could prevent `In[3]` from scheduling the new Task on the kernel main loop.

njsmith:
This is expected, because the call to ensure_future is what schedules the
coroutine on the event loop. I think the difference between your result and
what Matthias was predicting is that he was talking about running this from
the console ipython, where the event loop is paused when waiting for input,
while you're using the kernel's event loop that keeps running in the
background in between cells.

vidartf:
@njsmith Yes, it is expected, I was just suggesting that if we want to avoid this behavior changing depending on the users Tornado/Python version we should probably implement a custom event loop policy (or some other way to make sure that `ensure_future` does not get the kernel event loop).

Carreau:
> @njsmith Yes, it is expected, I was just suggesting that if we want to avoid this behavior changing depending on the user's Tornado/Python version we should probably implement a custom event loop policy (or some other way to make sure that ensure_future does not get the kernel event loop).

I don't want to get into the business of implementing our own event loop (at least not right now). 
I looked into automatically selection loop depending on whether you have native function from curio/asyncio/trio, but that seem like a rats nest. I also think the IPython is flexible enough to set your custom runner or event loop policy with a custom magic that we do not need to put that in the core now. Though, if somethings like this get developed separately and get traction/stabilises,  we can introduce that into the core.

IPython/core/async_helpers.py
```diff
@@ -0,0 +1,82 @@
...
+import inspect
+from textwrap import dedent, indent
+from types import CodeType
+
+def _asyncio_runner(coro):
+    """
+    Handler for asyncio autoawait
+    """
+    import asyncio
+    return asyncio.get_event_loop().run_until_complete(coro)
```
minrk: What would happen if ipykernel started running on asyncio? Would this cause hangs?

Carreau:
> What would happen if ipykernel started running on asyncio? Would this cause hangs?

Likely, but if we make ipykernel use asyncio, we can always change this handler and make a minor release. And maybe by that time python 3.7 is out and will support actual async exec. I'm going to guess there might  also be weird cases when you use uvloop, or if you register a loop in a different thread to have actual async execution. But I doubt this is worth trying to tackle now.

IPython/core/interactiveshell.py
```diff
@@ -2641,13 +2758,33 @@ def error_before_exec(value):
...
+        _run_async = False
+
         with self.builtin_trap:
             cell_name = self.compile.cache(cell, self.execution_count)
 
             with self.display_trap:
                 # Compile to bytecode
                 try:
-                    code_ast = compiler.ast_parse(cell, filename=cell_name)
+                    if _should_be_async(cell) and self.autoawait:
```
minrk: I'd reverse these since autoawait is a bool and `_should_be_async` doesn't need to parse if the bool is False.

IPython/core/interactiveshell.py
```diff
@@ -2826,6 +3005,8 @@ def run_code(self, code_obj, result=None):
           A compiled code object, to be executed
         result : ExecutionResult, optional
           An object to store exceptions that occur during execution.
+        async :  Bool (Experimental)
```
minrk: General question: with async the almost-a-keyword, is there any indication that `async` as a variable name should be discouraged?

IPython/core/magics/basic.py
```diff
@@ -379,6 +379,65 @@ def xmode_switch_err(name):
...
+
+        If the passed parameter does not match any of the above and is a python
+        identifier, get said object from user namespace and set it as the
+        runner, and activate autoawait.
+
+        If the object is a fully qualified object name, attempt to import it and
+        set it as the runner, and activate autoawait."""
+
+        param = parameter_s.strip()
+        d = {True:"on",False:"Off"}
```
minrk: mixed case: 'on' 'Off'. Presumably both should be lowercase.

IPython/core/interactiveshell.py
```diff
@@ -220,6 +317,23 @@ class InteractiveShell(SingletonConfigurable):
...
+    autoawait = Bool(True, help=
+        """
+        Automatically run await statement in the top level repl.
+        """
+    ).tag(config=True)
+
+    loop_runner_map ={
+        'asyncio':_asyncio_runner,
+        'curio':_curio_runner,
+        'trio':_trio_runner,
```
minrk: It seems a little odd that tornado, the only async runner that we actually use in IPython, isn't supported here. Is there a reason for that?

IPython/core/magics/basic.py
```diff
@@ -379,6 +379,65 @@ def xmode_switch_err(name):
...
+            self.shell.loop_runner = param
+            self.shell.autoawait = True
+            return None
+
+        if param in self.shell.user_ns :
+            self.shell.loop_runner = self.shell.user_ns[param]
+            self.shell.autoawait = True
+            return None
+
+        from traitlets.utils.importstring import import_item
```
minrk: import to top-level of module

IPython/core/tests/test_async_helpers.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+
+import sys
+import nose.tools as nt
+from textwrap import dedent
+
+ip = get_ipython()
+iprs = lambda x: ip.run_cell(dedent(x))
+
+if sys.version_info > (3,5):
```
minrk: Rather than making the test definitions themselves conditional, why not skip the tests?

IPython/core/interactiveshell.py
njsmith: Good catch – the plan is indeed to make `async` into a real keyword in 3.7, so it shouldn't be used as a name.

IPython/core/async_helpers.py
```diff
@@ -0,0 +1,82 @@
...
+def _asyncify(code:str) -> str:
+    """wrap code in async def definition.
+
+    And setup a bit of context to run it later.
+    """
+    res = dedent("""
+        async def ___wrapper___():
+            {usercode}
+            locals()
+            return None
```
njsmith: Just happened to look at this again, and noticed that I think this should be
```python
async def __wrapper__():
    try:
        {usercode}
    finally:
        locals()
```

Carreau:
Closing in favor of #11155 (at least for now)

PR #10391, opened 2017-03-10
Expand a bit the documentation about transformers.

Carreau:
Mostly answering questions I had myself while trying to write some stuff
and while trying to get a simple example working.

### DIFF
--- a/IPython/core/inputsplitter.py
+++ b/IPython/core/inputsplitter.py
@@ -24,6 +24,7 @@ import sys
 import tokenize
 import warnings
 
+from IPython.utils.py3compat import cast_unicode
 from IPython.core.inputtransformer import (leading_indent,
                                            classic_prompt,
                                            ipy_prompt,
@@ -686,6 +687,7 @@ class IPythonInputSplitter(InputSplitter):
         """
 
         # We must ensure all input is pure unicode
+        lines = cast_unicode(lines, self.encoding)
         # ''.splitlines() --> [], but we need to push the empty line to transformers
         lines_list = lines.splitlines()
         if not lines_list:
--- a/IPython/core/inputsplitter.py
+++ b/IPython/core/inputsplitter.py
@@ -24,7 +24,6 @@ import sys
 import tokenize
 import warnings
 
-from IPython.utils.py3compat import cast_unicode
 from IPython.core.inputtransformer import (leading_indent,
                                            classic_prompt,
                                            ipy_prompt,
@@ -687,7 +686,6 @@ class IPythonInputSplitter(InputSplitter):
         """
 
         # We must ensure all input is pure unicode
-        lines = cast_unicode(lines, self.encoding)
         # ''.splitlines() --> [], but we need to push the empty line to transformers
         lines_list = lines.splitlines()
         if not lines_list:
--- a/docs/source/config/inputtransforms.rst
+++ b/docs/source/config/inputtransforms.rst
@@ -77,8 +77,8 @@ Transforming a full block
 .. warning::
 
     Transforming a full block at once will break the automatic detection of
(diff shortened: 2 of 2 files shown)


IPython/core/inputsplitter.py
```diff
@@ -687,7 +686,6 @@ def push(self, lines):
         """
 
         # We must ensure all input is pure unicode
-        lines = cast_unicode(lines, self.encoding)
```
takluyver: Why is this changed in a doc PR? If you're sure removing it is OK, please remove the comment too.

docs/source/config/inputtransforms.rst
```diff
@@ -71,6 +71,32 @@ The decorator returns a factory function which will produce instances of
...
 function.
 
+Transforming a full block
+-------------------------
+
+Transforming a full block of python code is possible by implementing a
+:class:`~IPython.core.inputtransformer.Inputtransformer` and overwriting the
+``push`` and ``reset`` methods. The reset method should send the full block of
+transformed text. As an example a transformer the reversed the lines from last
+to first.
```
takluyver: We should say that this will break the automatic 'is this input complete' detection in console interfaces - with a transformer like this added, it will always think more input is needed, and you will need to use a shortcut to force it to execute.

docs/source/config/inputtransforms.rst
```diff
@@ -79,9 +105,28 @@ sent each line in turn, followed by ``None`` to reset it. It can yield lines, or
...
+
+    ip = get_ipython()
+
+    ip.input_splitter.logical_line_transforms.append(noop())
+    ip.input_transformer_manager.logical_line_transforms.append(noop())
+
 This code in IPython strips a constant amount of leading indentation from each
 line in a cell::
 
+    from IPython.core.inputtransformer import CoroutineInputTransformer
```
takluyver: Do you want to add the import to your new example above too?

docs/source/config/inputtransforms.rst
```diff
@@ -79,9 +105,28 @@ sent each line in turn, followed by ``None`` to reset it. It can yield lines, or
 ``None`` if it is accumulating text to yield at a later point. When reset, it
 should give up any code it has accumulated.
 
+You may use :meth:`CoroutineInputTransformer.wrap` to simplify the creation of
+such a transformer.
+
+Here is a simple :class:`CoroutineInputTransformer` that can be though of be
```
willingc: s/though/thought

willingc: ~~be~~

docs/source/config/inputtransforms.rst
```diff
@@ -71,6 +71,39 @@ The decorator returns a factory function which will produce instances of
 :class:`~IPython.core.inputtransformer.StatelessInputTransformer` using your
 function.
 
+Transforming a full block
+-------------------------
+
+.. warning::
+
+    Transforming a full block at once will break the automatic detection of
+    wether a block of code is complete in interface relying on this
```
takluyver: - whether
- interfaces

docs/source/config/inputtransforms.rst
```diff
@@ -71,6 +71,39 @@ The decorator returns a factory function which will produce instances of
...
+    functionality, as for example terminal IPython. You will need to use a
```
takluyver: 'as for example' doesn't work in English. You could use 'such as', 'including', or 'for example'.

('such as, for example, foo' is valid, but it's too verbose to read nicely here, IMO)

Carreau: done,

docs/source/config/inputtransforms.rst
Carreau: done.

PR #10403, opened 2017-03-20
Improve timeit message to fix ambuiguity.

### LINKED ISSUE #10359: Timeit results does not show -r value
```
In [1]: timeit -o -r7 -n11 1+1
11 loops, average of 7: 33.1 ns +- 8.77 ns per loop (using standard deviation)
Out[1]: <TimeitResult : 11 loops, average of 7: 33.1 ns +- 8.77 ns per loop (using standard deviation)>
```

Should show r=7.

Carreau:
10 loops, best of 7 is a bit ambiguous, be clearer about what is meant.

Closes #10359 

--- 

Thanks @fperez for helping with the wording.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -85,14 +85,20 @@ class TimeitResult(object):
         return (math.fsum([(x - mean) ** 2 for x in self.timings]) / len(self.timings)) ** 0.5
 
     def __str__(self):
-        return (u"%s loop%s, average of %d: %s +- %s per loop (using standard deviation)"
-                   % (self.loops,"" if self.loops == 1 else "s", self.repeat,
-                      _format_time(self.average, self._precision),
-                      _format_time(self.stdev, self._precision)))
+        return (
+            u"{mean} ± {std} per loop (mean ± std. dev. of {runs} run{run_plural}, {loops} loop{loop_plural} each)"
+                .format(
+                    runs = self.repeat,
+                    loops = self.loops,
+                    loop_plural = "" if self.loops == 1 else "s",
+                    run_plural = "" if self.repeat == 1 else "s",
+                    mean = _format_time(self.average, self._precision),
+                    std = _format_time(self.stdev, self._precision))
+                )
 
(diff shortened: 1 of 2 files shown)


willingc (approved):
Thanks @Carreau. A nice improvement!

IPython/core/magics/execution.py
```diff
@@ -85,14 +85,20 @@ def stdev(self):
         return (math.fsum([(x - mean) ** 2 for x in self.timings]) / len(self.timings)) ** 0.5
 
     def __str__(self):
-        return (u"%s loop%s, average of %d: %s +- %s per loop (using standard deviation)"
-                   % (self.loops,"" if self.loops == 1 else "s", self.repeat,
-                      _format_time(self.average, self._precision),
-                      _format_time(self.stdev, self._precision)))
+        return (
+            u"{mean} ± {std} per loop (mean ± std. dev. of {runs} run{run_plural}, {loops} loop{loop_plural} each)"
```
fperez: Are we comfortable assuming that ± will render fine everywhere? I recall looking at this ages ago and deciding for the more pedestrian/ugly +- precisely because ± didn't render very reliably.

fperez (approved):
I'm ok approving as-is, I trust you to check the ± rendering issue, but I don't want to make the review conditioned on that.

thanks!

IPython/core/magics/execution.py
takluyver: For comparison, have a look at what we do in `_format_time()` at the bottom of this file to print `μs` (with mu) if we determine it's safe and `us` otherwise.

Carreau: > For comparison, have a look at what we do in _format_time() at the bottom of this file to print μs (with mu) if we determine it's safe and us otherwise.

Done. Though I think we are starting to support recent enough systems that it is likely not an issue, and it would be nice to push people to utf8. But fair enough.

takluyver: It's probably not an issue most of the time, but there'll be that one weird corner case where it is, and someone's left looking at a traceback.

PR #10404, opened 2017-03-22
JSON: Include kwargs in metadata

gnestor:
This cleans up both the `JSON` and `GeoJSON` display classes by including `kwargs` in the bundle’s `metadata`. This allows subclasses of `JSON`, such as `GeoJSON`, to support _arguments as metadata_ by default vs. requiring subclasses override several JSON methods.

Usage examples:

```py
# JSON
JSON({
    'string': 'string',
    'array': [1, 2, 3],
    'bool': True,
    'object': {
        'foo': 'bar'
    }
}, expanded=True)

# GeoJSON
GeoJSON({
    "type": "Feature",
    "geometry": {
        "type": "Point",
        "coordinates": [-118.4563712, 34.0163116]
    }
}, url_template="https://api.tiles.mapbox.com/v4/{id}/{z}/{x}/{y}.png?access_token=pk.eyJ1Ijoic2Vub3JuZXN0b3IiLCJhIjoiY2l4MHU4MWJsMDIwcjJ0cGwybWQzbnhpeiJ9.beQ7B_WHe3K7_YMUJ684yg", 
layer_options={
    "id": "mapbox.streets",
    "attribution" : '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
    "minZoom" : 0,
    "maxZoom" : 18,
})

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -622,7 +622,7 @@ class JSON(DisplayObject):
     """
     # wrap data in a property, which warns about passing already-serialized JSON
     _data = None
-    def __init__(self, data=None, url=None, filename=None, metadata=None, **kwargs):
+    def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=None, **kwargs):
         """Create a JSON display object given raw data.
 
         Parameters
@@ -640,9 +640,11 @@ class JSON(DisplayObject):
         metadata: dict
             Specify extra metadata to attach to the json display object.
         """
-        self.metadata = kwargs
+        self.metadata = {'expanded': expanded}
         if metadata:
             self.metadata.update(metadata)
+        if kwargs:
+            self.metadata.update(kwargs)
         super(JSON, self).__init__(data=data, url=url, filename=filename)
 
     def _check_data(self):
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -622,7 +622,7 @@ class JSON(DisplayObject):
     """
     # wrap data in a property, which warns about passing already-serialized JSON
     _data = None
-    def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=None):
+    def __init__(self, data=None, url=None, filename=None, metadata=None, **kwargs):
         """Create a JSON display object given raw data.
 
         Parameters
@@ -640,8 +640,9 @@ class JSON(DisplayObject):
         metadata: dict
             Specify extra metadata to attach to the json display object.
         """
-        self.expanded = expanded
-        self.metadata = metadata
+        self.metadata = kwargs
+        if metadata:
+            self.metadata.update(metadata)
         super(JSON, self).__init__(data=data, url=url, filename=filename)
 
     def _check_data(self):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: IPython/core/display.py [hunk 1]
```diff
@@ -622,7 +622,7 @@ class JSON(DisplayObject):
     """
     # wrap data in a property, which warns about passing already-serialized JSON
     _data = None
-    def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=None):
+    def __init__(self, data=None, url=None, filename=None, metadata=None, **kwargs):
```
Carreau: Removing `expanded=False` is a change of API as it's also the 4th positional parameter of the __init__ function.

Carreau:
That seem reasonable. 
I think we just need to do an extra dance with `expanded=`, just to not break usage where positional version of the API is used.

gnestor:
I see how this could break if the `expanded` argument is not provided by name. If I'm not mistaken though, `data`, `url`, and `filename` are mutually exclusive in that a user will only use one (at position 1). If that's the case, then a user could only define `expanded` by position if done like `JSON({'key': 'value'}, None, None, True)` or `JSON(None,'https://gist.githubusercontent.com/gnestor/ea4b5786c21d592924d616ffce61b72e/raw/50d725e899fa601b767a0ec41a81d74aeb9d3346/vega-3-spec.json', None, True)`?

Carreau:
> I see how this could break if the expanded argument is not provided by name. If I'm not mistaken though, data, url, and filename are mutually exclusive in that a user will only use one (at position 1). If that's the case, then a user could only define expanded by position if done like JSON({'key': 'value'}, None, None, True) or JSON(None,'https://gist.githubusercontent.com/gnestor/ea4b5786c21d592924d616ffce61b72e/raw/50d725e899fa601b767a0ec41a81d74aeb9d3346/vega-3-spec.json', None, True)?

Yep, that's right. I know it's rare, but if it's too complicated, then we likely change the API.
If we change API, we could make these kwarg only.

IPython/core/display.py
```diff
@@ -622,7 +622,7 @@ class JSON(DisplayObject):
     """
     # wrap data in a property, which warns about passing already-serialized JSON
     _data = None
-    def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=None):
+    def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=None, **kwargs):
```
gnestor: Ok, this will prevent API breakage :+1:

IPython/core/display.py
```diff
@@ -640,8 +640,11 @@ def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=
         metadata: dict
             Specify extra metadata to attach to the json display object.
         """
-        self.expanded = expanded
-        self.metadata = metadata
+        self.metadata = {'expanded': expanded}
+        if metadata:
+            self.metadata.update(metadata)
+        if kwargs:
+            self.metadata.update(kwargs)
```
gnestor: I've moved the metadata merging logic from `_data_and_metadata` to init because I don't think`metadata` or `expanded` can change after init.

PR #10408, opened 2017-03-24
Speed up shutdown by eliminating unnecessary sleeping

lpsinger:
Eliminate an annoying split-second pause when closing a shell session by reducing unnecessary sleeps while killing background scripts. If there are no bacgkround scripts running, or if they shut down on the first or second attempt, then we can eliminate some pauses.


takluyver:
Thanks!

takluyver:
@meeseeksdev backport

PR #10410, opened 2017-03-24
Remove %profile in perspective of using it in later IPython versions.

Carreau:
Raise an informative error message with both legacy and new behavior.

Prompted by discussion with @stefanv who is not the only one that get confused by `%profile` and think it profiles code. I'm thinking on letting it raise for about a year or two, and then remove/replace by a %prun-like magic (or maybe something like Robert Kern line_profiler)

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -294,13 +294,9 @@ Currently the magic system has the following functions:""",
         prun : run code using the Python profiler
                (:meth:`~IPython.core.magics.execution.ExecutionMagics.prun`)
         """
-        warn("The `%profile` magic has been deprecated since IPython 2.0. Please use "
-             "the value of `get_ipython().profile` instead to see current profile in use.")
-        from IPython.core.application import BaseIPythonApplication
-        if BaseIPythonApplication.initialized():
-            print(BaseIPythonApplication.instance().profile)
-        else:
-            error("profile is an application-level value, but you don't appear to be in an IPython application")
+        raise DeprecationWarning("The `%profile` magic has been deprecated since IPython 2.0. "
+            "and removed in IPython 6.0. Please use the value of `get_ipython().profile` instead "
+            "to see current profile in use. Perhaps you meant to use `%prun` to profile code ?")
 
     @line_magic
     def pprint(self, parameter_s=''):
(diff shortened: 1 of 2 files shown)


IPython/core/magics/basic.py
```diff
@@ -294,13 +294,9 @@ def profile(self, parameter_s=''):
...
                (:meth:`~IPython.core.magics.execution.ExecutionMagics.prun`)
         """
-        warn("The `%profile` magic has been deprecated since IPython 2.0. Please use "
-             "the value of `get_ipython().profile` instead to see current profile in use.")
-        from IPython.core.application import BaseIPythonApplication
-        if BaseIPythonApplication.initialized():
-            print(BaseIPythonApplication.instance().profile)
-        else:
-            error("profile is an application-level value, but you don't appear to be in an IPython application")
+        raise DeprecationWarning("The `%profile` magic has been deprecated since IPython 2.0. "
```
minrk: Do you mean `warnings.warn(...)`? Raise gives a traceback, which I don't think we want.

IPython/core/magics/basic.py
```diff
@@ -294,13 +294,9 @@ def profile(self, parameter_s=''):
...
+            "and removed in IPython 6.0. Please use the value of `get_ipython().profile` instead "
+            "to see current profile in use. Perhaps you meant to use `%prun` to profile code ?")
```
minrk: No space before `?`

IPython/core/magics/basic.py
Carreau: No, I meant raise. It has been printing a warning for already 4 versions.
The alternative is to just remove it, but then there is no explanation message. 
I want at least 1 version where calling this function is an error (raise exception or not present). To not have 2 consecutive versions where `%profile` is a valid magic.

IPython/core/magics/basic.py
Carreau: Removed. I don't think I'll ever get used to this english convention.

IPython/core/magics/basic.py
minrk: raising a DeprecationWarning seems odd, and shows a largely uninformative error. It's also not a deprecation anymore if it doesn't work. Maybe raise UsageError, then?

Carreau: Done.

Carreau:
Thanks.

PR #10414, opened 2017-03-27
update author email

minrk:
now that list is on python.org


Carreau:
@meeseeksdev backport

PR #10419, opened 2017-03-29
Fix is_event_loop_running_qt4 and is_event_loop_running_wx for ipytho…

### LINKED ISSUE #9974: is_event_loop_running_qt4 seems broken in IPython 5

bergtholdt:
…n kernels

basically using the "old way" of detecting

fixed #9974

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,10 @@ python:
     - 3.3
 sudo: false
 before_install:
-    - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
 install:
-    - pip install "setuptools>=18.5" pip --upgrade
-    - pip install -f travis-wheels/wheelhouse -e file://$PWD#egg=ipython[test] --upgrade
+    - pip install setuptools pip --upgrade
+    - pip install -e file://$PWD#egg=ipython[test] --upgrade
     - pip install codecov --upgrade
 script:
     - cd /tmp && iptest --coverage xml && cd -


minrk:
I think it's not in the `is_event_loop_running` where the check is wrong, but rather in `start_event_loop`. I think It'll be more sustainable to fix this by not calling `start_event_loop` in ipykernel, instead calling a dedicated function in the ipykernel package.

minrk:
With https://github.com/ipython/ipykernel/issues/241, I think guisupport can be left alone. Thanks for tracking this down!

PR #10426, opened 2017-04-06
Ignore all exceptions in getargspec()

jdemeyer:
Sage overrides `getargspec()` and it can raise all kind of exceptions, not just `TypeError` or `AttributeError`.


Carreau:
Thanks

PR #10436, opened 2017-04-10
Polish release docs.

Carreau:


Carreau:
That's just docs.

Travis is happy on 4 out of 5 Python, so I'm merging and will do an RC.

Issue #10438, opened 2017-04-11
Outreach re: ipython 6.0.0rc1 breaking builds of related projects on Python 2.7


nealmcb:
After the [ipython 6.0.0rc1 release](https://github.com/ipython/ipython/releases/tag/6.0.0rc1), the [python-fire](https://github.com/google/python-fire) project ran in to Travis CI failures for internal builds, as noted in https://github.com/google/python-fire/issues/57

It would be helpful to have more readily visible or even proactive advice on how to configure `setup.py`, pip, and other dependency-management systems for a wide variety of use cases.  E.g. how to configure a project to depend on ipython 6 for supported python versions, but version 5 otherwise.

nealmcb:
The discussion in https://github.com/ipython/ipython/pull/9900 is pointed to from the README, and was indeed proactive and very helpful. But it only covers the case "I'm installing IPython itself".  Extending that effort with tips for the other projects which depend on IPython (and also have naive users doing source builds) is what I'm talking about here.

A blog post about 6.0.0rc1 and its impact would be most welcome.

nealmcb:
It looks offhand like [pypi-data](https://github.com/nathforge/pypi-data) could be used to find most of the pip-installable packages that depend on IPython.

willingc:
Hi @nealmcb,

I'll let @takluyver and @carreau provide more details on the possibility of editing error messages/install messages on discontinuing 2.7 support beyond the IPython 5 LTS release.

>It would be helpful to have more readily visible or even proactive advice on how to configure setup.py, pip, and other dependency-management systems for a wide variety of use cases. E.g. how to configure a project to depend on ipython 6 for supported python versions, but version 5 otherwise.

We can definitely look at the docs and README to further highlight Python 2 support and IPython 5 LTS.

> A blog post about 6.0.0rc1 and its impact would be most welcome.

Thanks for the suggestion. We typically do blog posts for major releases and mailing list posts for release candidates.

Thanks!

takluyver:
Many packages are only uploaded as source packages, and the setup script can ask for different dependencies depending on the environment (e.g. require x only on Windows, or only on Python 2). So you can't have a static list of all a packages dependencies in general.

There are better ways to handle conditional dependencies now - you can make a static list with 'environment markers' to filter which ones are used. If packages provide the information, PyPI can record dependencies (e.g. [astsearch](https://pypi.python.org/pypi/astsearch) shows the dependency on astcheck). But many packages don't.

[...]

PR #10439, opened 2017-04-11
Add an important directive for Py2/Py3 support

willingc:
@Carreau and @takluyver, here is some suggested wording to address comments in #10438. The rst directive didn't render well on GitHub so feel free to modify. (I'm traveling today).


takluyver:
I've made it not a directive, because for the README I think the rendering on Github matters. I've also made the description about 5.x refer only to Python 2.7 - we already dropped support for 2.6 some time ago, and we weren't explicitly supporting Py 3.2, whether or not it worked.

willingc:
Perfect @takluyver. Ship it!

PR #10446, opened 2017-04-12
Setup.py cleanup

takluyver:
I went to polish the pip version check @Carreau added yesterday, and then I noticed that we also have code lying around to build and upload Windows installers, which we haven't done since [IPython 0.10](https://pypi.python.org/pypi/ipython/0.10) in 2010.

I don't think we're going to go back to building Windows installers with distutils. Packaging tools have improved the experience for Windows users (e.g. pip can create `ipython.exe` so IPython is accessible at the command line), and we often point people to all-in-one bundles such as Anaconda. I do want to keep thinking about how people get our software, but I don't think GUI installers for a single package make that easier.


setup.py
```diff
@@ -17,24 +17,26 @@
...
         import pip
         pip_version = tuple([int(x) for x in pip.__version__.split('.')[:3]])
-        print(pip_version)
         if pip_version < (9, 0, 1) :
-            pip_message = 'You pip version is out of date, please install pip >= 9.0.1.'\
+            pip_message = 'Your pip version is out of date, please install pip >= 9.0.1. '\
             'pip {} detected.'.format(pip.__version__)
+        else:
+            # pip is new enough - it must be something else
+            pip_message = ''
```
Carreau: I think it might be that the pip you find is more recent than the pip you install if pip get upgraded as same time as IPython. But fair enough.

PR #10453, opened 2017-04-14
minor changes

RaymondDavidson:
dded minor changes and try/except on imports

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -30,27 +30,27 @@ https://docs.python.org/2/license.html
 try:
     import bdb
 except:
-    print "unable to import bdb"
+    print ("unable to import bdb")
 try:
     import functools
 except:
-    print "unable to import functools"
+    print ("unable to import functools")
 try:
     import inspect
 except:
-    print "unable to import inspect"
+    print ("unable to import inspect")
 try:
     import linecache
 except:
-    print "unable to import linecache"
+    print ("unable to import linecache")
 try:
     import sys
 except:
-    print "unable to import sys"
+    print ("unable to import sys")
 try:
     import warnings
 except:
-    print "unable to import warnings"
+    print ("unable to import warnings")
 from IPython import get_ipython
 from IPython.utils import PyColorize
 from IPython.utils import coloransi, py3compat
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -26,13 +26,31 @@ https://docs.python.org/2/license.html
 #
 #*****************************************************************************
 
-import bdb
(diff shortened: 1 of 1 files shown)


Carreau:
Hi, 

Thanks you for the pull requests, though you are using print statements (the current code source is Python 3 only so need parenthesis)  And I'm unsure printing and silencing exception is the right things. 

What is your thought process ? Or are you trying to debug something ? 

Thanks.

RaymondDavidson:
i thought it might be helpful, also, it would keep the script from crashing on a failed import

RaymondDavidson:
i just had an idea, so i suggested it

Carreau:
> i thought it might be helpful, also, it would keep the script from crashing on a failed import
> View changes

Well, if there is a failed import then the script will not work. Preventing the error message at import is just asking for trouble later. Plus all of these are from standard library modules that can't be missing. 
If they are missing then something is deeply wrong with your Python install. 

A better way of having optional modules is to import them in the function using them:

```
def foo():
     import mod
     mod.dostuff()
```

Then `mod` will be imported only if `foo()` if used, so `mod` can be an optional dependency.
I can see this maybe of use for `functools` case for this file, but it's part of the standard lib as well on Py3 so no need to protect or delay it. 

Note that you can do the following to see what modules are already loaded at Python startup:

```
$ python3
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
>>> import sys
>>> sys.modules['functools']
<module 'functools' from '/Users/bussonniermatthias/anaconda/lib/python3.6/functools.py'>
>>> sys.modules['warnings']
<module 'warnings' from '/Users/bussonniermatthias/anaconda/lib/python3.6/warnings.py'>
>>>
```

You can see that warning, and functools are already available are are thus required for Python to work. 
You can also use `sys.modules.keys()` to see the full list. 

For these catching import error will never trigger at is is a no-op (except binding the name)
If we were to go this route (which we probably won't unless special case), you might want to separate the "ImportError" from "ModuleNotFound" (like package not installed). And a good practice is to reraise, or at least print the current stack (even if you don't reraise) as the stack get the __reason__ for the error.

Carreau:
I doubt we want to get something like that in IPython (we can revisit if other disagree) and we are close enough from 6.0 that I don't want to commit to that so close to release. 

I'm going to close (to keep the PR Queue small), but I look forward to receive more contribution from you. We do have a number of  issue that need work and [should not be too hard](https://github.com/ipython/ipython/issues?q=is%3Aopen+is%3Aissue+label%3A%22Need+Contributor%22).

Thanks !

PR #10475, opened 2017-04-21
Fix bug in generating shortcuts

frenzymadness:
Hello.
I found this bug during building RPM package for Fedora with Python 2 only. I think that key from KeyBindingManager registry can be `unicode` and not only `str`.

At least, this fixes my issue during build which failed with:
```
/usr/bin/python2 autogen_shortcuts.py
  File "autogen_shortcuts.py", line 63, in <module>
    shortcut = ' '.join([k if isinstance(k, str) else k.name for k in kb.keys])
AttributeError: 'unicode' object has no attribute 'name'
make: *** [Makefile:90: autogen_shortcuts] Error 1
```

### DIFF
--- a/docs/autogen_shortcuts.py
+++ b/docs/autogen_shortcuts.py
@@ -60,7 +60,7 @@ for kb in ipy_bindings:
     if not doc or doc in dummy_docs:
         continue
 
-    shortcut = ' '.join([k if isinstance(k, str) else k.name for k in kb.keys])
+    shortcut = ' '.join([k if isinstance(k, (str, unicode)) else k.name for k in kb.keys])
     shortcut += shortcut.endswith('\\') and '\\' or ''
     if hasattr(kb.filter, 'filters'):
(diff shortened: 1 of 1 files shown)


takluyver:
That would break building the docs on Python 3, however, which is what we usually do. Is it practical to build the docs with Python 3 in Fedora? IPython 6 requires Python 3 anyway.

frenzymadness:
Yes, you are right. Sorry for breaking compatibility with Python 3.

In Fedora, we are splitting IPython packages to two separated packages - one with the latest version (for Python 3) and the second one with IPython version 5.x for Python 2. So I would prefer that package with IPython version 5.x depends only on Python 2 packages.

If you agree, I can fix compatibility using `string_types` from `IPython.utils.py3compat`.

takluyver:
That sounds good. Could you make the PR against the 5.x branch? Thanks!

frenzymadness:
Thank you, new PR is here: https://github.com/ipython/ipython/pull/10476

PR #10476, opened 2017-04-21
Fix bug in generating shortcuts

frenzymadness:
Key from KeyBindingManager registry can be `unicode` and not only `str`.

The fix is compatible with Python 2 and 3 with using `string_types` from `py3compat`.


takluyver:
Thanks, this looks fine to me.

Issue #10425, opened 2017-04-06
IPython 5.3 terminal: enter should execute not insert line break


bmerry:
This is a regression from 5.2.2 to 5.3 when using the terminal interface: if the cursor is in the middle of a line and I press enter, a line break is inserted into the line. On 5.2.2, pressing enter had the same effect as if the cursor was at the end of the line i.e. it executes the line I've just entered.

To reproduce: type `foo(2, 3)` (without pressing enter), move the cursor to be on the 3, then press enter.

It's extremely rare that one wants to break lines in an interactive session, but quite common to go back to a line from history, modify the contents, and then immediately want to execute it. Please bring back the old behaviour.

gatagat:
I also find this new behavior annoying. If reverting it completely is not possible, would it work to have `Shift+Enter` to execute (as in notebooks) when in the middle of a line? At the end of the line, both `Shift+Enter` and `Enter` would execute.

Carreau:
> If reverting it completely is not possible, would it work to have Shift+Enter to execute (as in notebooks) when in the middle of a line?

Unfortunately terminals technical limitation prevent us to do that. You can use `Esc` followed by `Enter` though.

carlsmith:
> It's extremely rare that one wants to break lines in an interactive session...

Personally, since IPython 5 moved to using Prompt Toolkit, I often write multiline code interactively. It actually annoys me that IPython executes so eagerly. I find larger chunks of code easier to manage than lots of separate lines. Each to their own.

bmerry:
Indeed, you'll never keep everyone happy. @carlsmith: out of interest, when writing multiline code are you breaking lines in the middle of statements (e.g. inside parentheses) or between statements? The behaviour I'm finding annoying is IPython inserting line breaks when the cursor is between parentheses within a statement.

carlsmith:
I've not really thought about it, but it would probably be both.

It's one of those things that everyone approaches differently, and as it makes no sense to promote a specific coding style for working in a shell, it's difficult to know what would work best for most people.

iblis17:
Is there any way to configure this behavior?
I expect the command should be executed instantly as well...

takluyver:
#10489 attempts to improve this, but I'm not sure that it's quite right.

PR #10489, opened 2017-04-26
Prefer execution when there's only a single line entered

takluyver:
Closes gh-10425

The heuristic here is to treat a single line specially, and always evaluate it as if the cursor was at the end. An alternative heuristic could be to do this if the cursor is on the last line of the input.

This could also cause some weird effects if you e.g. type `for a in range(5):`, move the cursor back a few places and press enter - you'll get a newline inserted in the text, but it will indent as if it were after the colon. I'm still trying to think if there's a better way to approach it.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -108,8 +108,13 @@ def newline_or_execute_outer(shell):
                 b.cancel_completion()
             return
 
-        before_text = d.text[:d.cursor_position]
-        status, indent = shell.input_splitter.check_complete(before_text + '\n')
+        # If there's only one line, treat it as if the cursor is at the end.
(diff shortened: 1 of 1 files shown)


minrk (approved):

minrk:
AOK by me. Do you want to merge this as an incremental improvement, or keep thinking before landing this one?

takluyver:
I think let's do this as an incrememental improvement (@Carreau seems enthusiastic, going by #10498 ), and I'll try to keep thinking about it.

Carreau:
Ha, I sent the command on the wrong issue hard to review from phone

@meeseeksdev merge :-)

Carreau:
Ok, no smiley after merge.

@meeseeksdev merge

Carreau:
@meeseeksdev[bot] backport

PR #10497, opened 2017-05-02
Fix virtualenv warning for Cygwin

Igonato:
Getting `Attempting to work in a virtualenv...` warnings when running in Cygwin and starting ipython in a virtualenv. (Also, using a virtualenvwrapper, not sure if it makes a difference)

Turns out `p_venv` looks like this `\cygdrive\c\users\igonato\envs\myenv` when `sys.executable` is like this `c:\users\igonato\envs\myenv\scripts\python.exe`.

This should make the check a bit more robust when running in Cygwin

Cheers!

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -699,7 +699,14 @@ class InteractiveShell(SingletonConfigurable):
             p = os.path.normcase(os.path.join(os.path.dirname(p), os.readlink(p)))
             paths.append(p)
         p_venv = os.path.normcase(os.environ['VIRTUAL_ENV'])
-        if any(p.startswith(p_venv) for p in paths):
+        
+        # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
+        if p_venv.startswith('\\cygdrive'):
+            p_venv = p_venv[11:]
+        else if p_venv[1] == ':':
+            p_venv = p_venv[2:]
+
+        if any(p_venv in p for p in paths):
             # Running properly in the virtualenv, don't need to do anything
             return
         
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks for the PR I'm unfortunately at conferences for the next 2 weeks, so may have issue having time to review it. 

One quick comment, can you deactivate the auto-trimming of whitespace on you editor? That screw up usage of git blame.

Thanks !

Igonato:
> One quick comment, can you deactivate the auto-trimming of whitespace on you editor

For sure, I noticed only after the push. Should I revert the changes or just keep it in mind for the future? It's only affected empty lines, (except one at 716). I'm willing to take the blame for any of them if they cause any problems ;)

Carreau:
> For sure, I noticed only after the push. Should I revert the changes or just keep it in mind for the future?

Yes, that would be great, it's ok if you don't I think that would be fine. We should probably have something more strict to avoid reintroducing space. But for now that's mostly historical reasons.

Thanks !

Igonato:
>> For sure, I noticed only after the push. Should I revert the changes or just keep it in mind for the future?
>
> Yes, that would be great, ...

"Yes" as "Yes, revert", right? Anyway, done!

Carreau:
Great thanks !

I'm going to guess that affects IPython 5.x as well, worth a backport ? (@takluyver?)

Igonato:
I'm not using IPython under Cygwin very often, so this was the first time I saw the warning, but yeah, whenever this venv detection logic was added, you should be able to safely backport the patch. 

On the other hand, it's just a wrong warning, everything should work just fine. Personally, I wouldn't bother with a backport.

Carreau:
Ok, tagging as 6.1 then. We can still change our minds.

Carreau:
Thanks !

Issue #10498, opened 2017-05-03
Always executing input with Enter?


Kwpolska:
In my opinion, this behavior is confusing (`|` is cursor):

```
>>> foo()| <enter>
foo runs
>>> foo(|) <enter>
a newline is added and editing continues
>>> foo(
|)
```

Could there be an option to make it work the old way (before 5.3.0)? I know I can work around it by using esc-enter (but I found out about this only by searching the issues and finding #10374 , so it’s undocumented!), however that’s not good enough for me (especially because I don’t need to it in my shell). I change parameters to functions more often than I write multi-line stuff in IPython.

via #9283 #10264 #10275 #10374

(IPython v6.0.0, Python 3.6.1, macOS Sierra)

takluyver:
#10489 would make it do what you expect in this case.

Carreau:
I like that as is and we should get it in or we won't try it.

@meeseeksdev merge

Thanks Thomas!

carlsmith:
Is there any chance this could become 'officially' configurable please?

It'd be cool if users could pass a function in to the configuration that gets called every time Enter is pressed, that has a reference to the cli instance or whatever makes sense, so it could update state, adding newlines and indentation, moving the cursor position etc. then return a bool, indicating whether to execute or not.

Maybe there's a better API that can reuse PTK's transform machinery, and Pygments' tokens. I'm not sure, but would like a hook into the Enter key's action.

Having a few gists kicking about, that we can hack on as users, that cover the three or four ways this could work, would also stop us from debating which is best.

takluyver:
> Is there any chance this could become 'officially' configurable please?

Sounds reasonable, #10500 would add a config option for it. The function you give it should work like [this one](https://github.com/ipython/ipython/blob/fcbde47323fac4809350e5ee21687027d42519f3/IPython/terminal/shortcuts.py#L97), which is the default implementation.

Carreau:
closed by #10500

PR #10500, opened 2017-05-04
Allow configuring a function to handle return in the terminal

takluyver:
As suggested on gh-10498


Carreau:
@meeseeksdev merge

+1, maybe a line in the docs somewhere in a subsequent PR with an example
if someone read this.

PR #10507, opened 2017-05-05
Prototype: show signature with Jedi.

Carreau:
That (unsurprisingly), change the APIs that were marked as unstable for 6.0. 
Not in a big way. 

In the terminal I stick the signature of functions (not callables, sorry numpy ufuncs..) into the meta-display (ie, next to the type), which make multicolumn compact (you see the sig only when selecting), and column a bit ugly, sig still in the type in the right column.  

sig

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1,4 +1,3 @@
-# encoding: utf-8
 """Completion for IPython.
 
 This module started as fork of the rlcompleter module in the Python standard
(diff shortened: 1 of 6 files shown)


takluyver:
Neat :-)

takluyver:
Do you want to land this as is so we can start playing with it, or do more on it first? There's nothing obviously amiss with the code, but I haven't used Jedi enough to really understand everything on that side.

Carreau:
I need to check it works with jedi stable (at least don't crash) and cleanup a bit/document/test.
I'm at the same time working on a patch to ipykernel (and nteract) to show this... but UI is hard.

Carreau:
Updated, cleaned, rebased, documented, tested. 

Now that 6.1 is out I'm happy to get that in .

Carreau:
Test fixed, just a small typo. 
I'm going to get that in to have a chance to get if tested a bit.

PR #10508, opened 2017-05-06
Fix Crash: Eliding string with consecutive dots.

Carreau:
Spiting a string with consecutive dots would lead to empty parts, which
you can't subslice when recombining and lead to `IndexError`.

Should not happened on 6.0, but could if the completer is refactored to
display completions that can contain the ellipsis symbol in Python 3 or
arbitrary string in default arguments.


Carreau:
Thanks!

PR #10509, opened 2017-05-06
Cleaning of Oinspect.

Carreau:
Mostly reading and cleaning in view of a refactor/ documentation.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -587,7 +587,21 @@ class Inspector(Colorable):
         return bundle
 
     def _get_info(self, obj, oname='', formatter=None, info=None, detail_level=0):
-        """Retrieve an info dict and format it."""
+        """Retrieve an info dict and format it.
+        
+        Parameters
+        ==========
+
+        obj: any
+            Object to inspect and return info from
+        oname: str (default: ''):
+            Name of the variable pointing to `obj`.
+        formatter: callable
+        info:
+            already computed informations
+        detail_level: integer
+            Granularity of detail level, if set to 1, give more informations.
+        """
 
         info = self._info(obj, oname=oname, info=info, detail_level=detail_level)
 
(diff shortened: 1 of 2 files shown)


IPython/core/oinspect.py
```diff
@@ -587,7 +587,21 @@ def format_mime(self, bundle):
...
-        """Retrieve an info dict and format it."""
+        """Retrieve an info dict and format it.
+        
+        Parameters
+        ==========
+
+        obj: any
+            Object to inspect and return info from
+        oname: str
+            Passed to self._info as is
```
takluyver: As `self._info` as a private method, I think this should explain the parameter rather than referring to another method. Though I'm still not entirely sure why we're deprecating `self.info()`.

IPython/core/oinspect.py
```diff
@@ -696,24 +710,31 @@ def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
...
-          precomputed already.
-
-        - detail_level: if set to 1, more information is given.
+        Parameters
+        ==========
+
+        obj: any
+            An object to find information about
+        oname: str (default: ''):
+            Nname of the variable pointing to the obj`.
```
takluyver: Nname -> Name

Also, stray backtick

takluyver (changes requested):
Needs a rebase, presumably since I merged your other PR.

IPython/core/oinspect.py
Carreau: _get_info is also private :-)  but fair.

takluyver (approved):

takluyver:
Can you check the merge conflict? I had a quick look, but it's not 100% obvious, and you've been looking at that code already.

Carreau:
Rebased.

IPython/core/oinspect.py
```diff
@@ -696,24 +710,31 @@ def info(self, obj, oname='', formatter=None, info=None, detail_level=0):
                       DeprecationWarning, stacklevel=2)
         return self._info(obj, oname=oname, info=info, detail_level=detail_level)
 
-    def _info(self, obj, oname='', info=None, detail_level=0):
+    def _info(self, obj, oname='', info=None, detail_level=0) -> dict:
```
takluyver: :-)

PR #10511, opened 2017-05-07
Clarify method deprecation

Carreau:


takluyver:
Why is the `info()` method deprecated when we're still using it? :-S

Carreau:
> Why is the info() method deprecated when we're still using it? :-S

To remove it at some point, I guess. I'm unsure if the full method is deprecated or just the `formatter` kwarg, which is what the warning warns about. I'm digging through this piece of code, and IIRC it's becaue we want to replace `info` by something that return a mimebundle, and that you'd better cann the method on the shell object instead of the inspector itself.

CAM-Gerlach:
For what its worth, the deprecation was done in one of the original PRs that added Docrepr and the warning is explicitly limited to only if the `formatter` argument is passed, but the docstring wasn't updated to match. `Inspector.info()` is used in the Docrepr test suite, and the confusing docstring sent me on a wild goose chase to figure out why it was deprecated and what it should be replaced with. Is there currently a replacement, other than using the private internal only function (which we obviously want to avoid)? It doesn't really make sense to deprecate a public API before a replacement is available, unless there is a pressing need to remove it entirely. Clarification would be much appreciated, thanks!

Carreau:
Can you open a separate issue ? I was agressive in 8.0.0a1 removal to catch some of those questions.

CAM-Gerlach:
Sure, I was going to after tracing the Git history and reading the original PRs since it seemed like an oversight, and I even forked and cloned the repo to prep a PR to adjust it, but then I found this one and I wasn't so sure.

PR #10165, opened 2017-01-17
Don't install numpy to test on Python 3.3

takluyver:

PR #10512, opened 2017-05-07
Test on travis with numpy installed.

Carreau:
Apparently we got a large drop of coverage when we removed this from the
[test] dependency in setup.py.

It's still not a dep, but no reasons not to test with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
 install:
     - pip install setuptools pip --upgrade
     - pip install -e file://$PWD#egg=ipython[test] --upgrade
+    - pip install numpy matplotlib
     - pip install codecov --upgrade
 script:
     - cd /tmp && iptest --coverage xml && cd -
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,7 +3,7 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 10
+        threshold: 5
     patch:
       default:
         target: 0%


Carreau:
Seem more complicated but apparently `./IPython/core/tests/test_pylabtools.py` is not run, or at least not included in coverage.

Carreau:
Occurred between #10163 (71% coverage) and #10165 (66% coverage).  @takluyver thoughts ? You wrote #10165

takluyver:
It looks like it's already trying to install numpy, and it's already there anyway: [Travis build log on master, w/ Py 3.6](https://travis-ci.org/ipython/ipython/jobs/229939011)

```
Requirement already up-to-date: numpy in /home/travis/virtualenv/python3.6.1/lib/python3.6/site-packages (from ipython)
```

And indeed this PR doesn't increase coverage: `66.58% (-0.01%) compared to a209dd0 `

I only dropped it from the dependencies for 3.3, because a new release of numpy didn't have wheels for 3.3, so it was trying to compile from source. It's still a test dep on Py >= 3.4.

takluyver:
I think something must have changed in the coverage-measuring systems around that time. [Here's the big coverage drop](https://codecov.io/gh/ipython/ipython/compare/6bf8bd1f5cf6a35a805b74985b3cce6559d59b26...9d28afa0dbbb80b8e8f16686b0de1c4cf589067e/tree), on the merge commit of #10163, which didn't change any code. If you look at the changes section in that comparison, it is seeing a lot of 'new' files that weren't actually new. So the decrease in % coverage is mostly from it suddenly adding a lot of lines to the denominator.

takluyver:
Closing this since it's not an improvement - we can keep discussing here if you think there's something we need to fix with the coverage.

Carreau:
Ah thanks ! That make sens. Let's keep it closed then. Still strange some files that are tested appear with 0 coverage...

PR #10513, opened 2017-05-07
Cleanup oinspect II

Carreau:
Mostly formatting, and removal of comment and debug statements.

I remove a handling of "special case", as far as I can't tell these
can't occur as we are looking up things by name, and these special case
are invalid python name. Also they appear as not be tested (lines not in
coverage). Also eval-ing these seem to be a bit evil, a simple mapping
seem better suited if we need to reintegrate them.


Carreau:
Thanks!

PR #10515, opened 2017-05-07
Standardize traceback.

Carreau:
Compress user and location of stdlib/sitepackage.
That should make traceback identical on all machines, hopefully making
Google a bit better at finding them, and a little less wide in many
case.

The drawback is that filename are not copy-pastable to edit files, unless
situated somewhere not in site packages but that should be the case for
__most__ dev install, so I expect that to not bother many people.

---- 
```
In [1]: import numpy; numpy.histogram('a')
With $STDLIB='/Users/bussonniermatthias/anaconda/lib/python3.6':
<ipython-input-1-e5edd817948c> in <module>()
----> 1 import numpy; numpy.histogram('a')

$STDLIB/site-packages/numpy/lib/function_base.py in histogram(a, bins, range, normed, weights, density)
    659             mn, mx = 0.0, 1.0
    660         else:
--> 661             mn, mx = a.min() + 0.0, a.max() + 0.0
    662     else:
    663         mn, mx = [mi + 0.0 for mi in range]

$STDLIB/site-packages/numpy/core/_methods.py in _amin(a, axis, out, keepdims)
     27
     28 def _amin(a, axis=None, out=None, keepdims=False):
---> 29     return umr_minimum(a, axis, None, out, keepdims)
     30
     31 def _sum(a, axis=None, dtype=None, out=None, keepdims=False):

TypeError: cannot perform reduce with flexible type
```
---- 

@takluyver I feel like you will be opinionated on this one.

I'm flexible on the naming and what we compress, but I think it would be great to do that.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -870,7 +870,7 @@ class VerboseTB(TBTools):
                     pass
 
         file = py3compat.cast_unicode(file, util_path.fs_encoding)
-        link = tpl_link % file
+        link = tpl_link % util_path.compress_user(file)
         args, varargs, varkw, locals = inspect.getargvalues(frame)
 
         if func == '?':


takluyver:
I'm happy enough to use the `compress_user` function to abbreviate the home directory as `~`. That saves some space, and you can still copy and paste the path to open it, at least on Unix.

I'm not convinced on `$STDLIB`, `$SITE-PACKAGES`, etc. I don't really like the separate delimited block at the top explaining variables. I don't like the $SUGGESTION that these are environment variables or shell variables. And both the standard library and `site-packages` can be in multiple locations (e.g. virtualenvs, Debian's `dist-packages` directories), so the simplifications won't work consistently.

> @takluyver I feel like you will be opinionated on this one.

It's like you know me! :stuck_out_tongue: 

## Counterproposal

What makes the line so long is having the path and the function signature in one line. From your example, `histogram(a, bins, range, normed, weights, density)` is already 51 characters, so with the ` in `, you only have 25 characters for the path before you hit 80 columns. What about putting them on separate lines:

```
~/anaconda/lib/python3.6/site-packages/numpy/lib/function_base.py
  in histogram(a, bins, range, normed, weights, density)
```

There the first line is only 65 characters.

We could either do this on every frame (simply and consistent), or only where the single line would exceed some length (so `foo.py in bar()` would still be on one line).

Carreau:
Yeah, I though about a large refactoring of traceback, you can even omit the filename if two consecutive frames have the same file, and I don't want to go this route (for now), counting line length with current architecture would be a pain.

I've push a simpler commit that just `compress_user`, and I'm fine delaying any other refactoring for later. If you think of any better layout please let me know. 

How does that sounds ?

takluyver:
Sounds reasonable. Let's go with compress_user and see how we get on.

Issue #10510, opened 2017-05-07
interactiveshell: almost indentical methods


Carreau:
interactive shell seem to have object_inspect, object_inspect_text and object_inspect_mime

The two first ones seem to be almost identical (one does some extra formatting though). 
If would be good to simplify the logic.

srinivasreddy:
I am taking it.

Carreau:
> I am taking it.

Thanks ! Be careful, I'm unsure this one is easy. It might be a rat'snest.

takluyver:
I'm not sure it's necessarily a problem. There are a few different methods doing similar things but returning different types of data. Maybe the docstrings just need to make it clearer what they are?

- `object_inspect()` returns a dictionary with semantic info, some of which is only meaningful in Python.
- `object_inspect_mime()` returns formatted info as a mimebundle.
- `object_inspect_text()` returns formatted info with only plain text.

PR #10517, opened 2017-05-08
remove un-necessary method

srinivasreddy:


Carreau:
Hi, thanks !

Unfortunately you can't "just" remove `object_inspect_mime`, it's part of the public api, so other can require it. So it would need to be deprecated first. 

Second we are more likely to move toward using `object_inspect_mime`  as it return more information than to remove it.

If you look closely it's more `object_inspect` and `object_inspect_text` that are similar. Not identical, but similar. You'll see that `inspector._get_info` does a bit more than `inspector._info`.

takluyver:
They also return different kinds of things, as I described in #10510. I don't think it's necessary to remove any of them, maybe just better document the difference.

srinivasreddy:
close this?

PR #10527, opened 2017-05-10
Document _repr_mimebundle_

Carreau:
Unfinished as display_* are broken with `_repr_mimebundle_`

### DIFF
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -960,7 +960,7 @@ class MimeBundleFormatter(BaseFormatter):
             if method is not None:
                 d = {}
                 d['include'] = include
-                d['exclude'] = include
+                d['exclude'] = exclude
                 return method(**d)
             return None
         else:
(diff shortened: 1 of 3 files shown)


dhirschfeld:
`_repr_mimebundle_` seems very cool! Just wondering if there are any plans to have the displayed repr to be user-selectable in the frontend?

What I'm thinking of is the user being able to select between e.g. an html table repr or a chart for a DataFrame.

...or is that not something it's designed for and would be more appropriate for an extension?

Carreau:
Yes it has been a long term plan, nobody bothered to implement it.

Carreau:
Rebased and fixed, new section `Custom Mimetypes with _repr_mimebundle_` in teh notebook. Closes #https://github.com/ipython/ipython/issues/10586 once backported to 5.x

rgbkrk:
> What I'm thinking of is the user being able to select between e.g. an html table repr or a chart for a DataFrame.

This is where I want to take the new data resource stuff with pandas. It won't be a toggle between mimetypes though -- it will be a component that allows switching views between the table and some light charting with vegalite (or otherwise).

takluyver:
@meeseeksdev backport

Issue #10531, opened 2017-05-11
?? show less info than ? 


blackmagic02881:
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:52:12)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: 1
Out[1]: 1

In [2]: _?
Type:        int
String form: 1
Docstring:
int(x=0) -> integer
int(x, base=10) -> integer

Convert a number or string to an integer, or return 0 if no arguments
are given.  If x is a number, return x.__int__().  For floating point
numbers, this truncates towards zero.

If x is not a number or if base is given, then x must be a string,
bytes, or bytearray instance representing an integer literal in the
given base.  The literal can be preceded by '+' or '-' and be surrounded
by whitespace.  The base defaults to 10.  Valid bases are 0 and 2-36.
Base 0 means to interpret the base from the string as an integer literal.
>>> int('0b100', base=0)
4

In [3]: _??
Type:        int
String form: 1

Carreau:
Closed by #10532

PR #10532, opened 2017-05-11
?? shall show docstring if source info is not available

blackmagic02881:
this is to fix the issue #10531. the reason is that when there is no source info. detail_level > 0 shall still show docstring.


Carreau:
Seem reasonable to me.

Carreau (approved):

takluyver:
And to me :-)

Carreau:
Mind opening an issue to add a test? I'm on my phone.thanks!

Carreau:
Thanks ! Back on a keyboard ! I had already forgotten :-)

Issue #8032, opened 2015-03-12
Non-deterministic tests in IPython.core.magic


takluyver:
While debugging other tests, @Carreau noticed that running `iptest IPython.core.magics` sometimes runs 9 tests, and sometimes 10.

Carreau:
That seem to be doctests of functions that have no examples in doctests. So likely not important.
Still it might mean we find the wrong number of doctests here and there.

It also can go as low at 8 and as high as 11 tests for `iptest IPython.core.magics`.

takluyver:
Fixed, we hope, by #10538.

PR #10538, opened 2017-05-11
Fix non-deterministic tests in IPython.core.magics

rtilk89:
See https://github.com/ipython/ipython/issues/8032

### DIFF
--- a/IPython/core/magics/code.py
+++ b/IPython/core/magics/code.py
@@ -80,7 +80,6 @@ def extract_code_ranges(ranges_str):
         yield (start, end)
 
 
-@skip_doctest
 def extract_symbols(code, symbols):
     """
     Return a tuple  (blocks, not_found)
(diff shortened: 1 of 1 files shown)


Carreau:
Wow. Nice finding. I'm baffled. 

cc @takluyver, I'd like to figure out why the `@skipdoctest` was triggering that.

kr1m:
Yes - I'm equally as baffled ;)

takluyver:
Weird. I'm totally guessing, but maybe it sometimes gets picked up by our doctest plugin, which looks for the flag set by `@skipdoctest`, and sometimes by the standard nose doctest collector, which might not.

Congratulations on finding it, but I'd like to find a fix that doesn't obscure the distinction between input and output. Maybe we can tweak the code so that it is a valid doctest?

kr1m:
The multiline code block seems to confuse the doctest. 

Doesn't my change follow the pattern above? https://github.com/ipython/ipython/pull/10538/files#diff-6cba74f9bcf04d4fb403e97ebeff3d3aR54

kr1m:
Hi @takluyver - See my latest pull request which contains a working doctest.

takluyver:
Thanks :-)

PR #10539, opened 2017-05-11
Respect the ignore and include keyword in display

### LINKED ISSUE #10526: display_* not respected if _repr_mimebundle_ defined.
See https://github.com/ipython/ipython/pull/10496#issuecomment-300641192
screen shot 2017-05-10 at 16 33 23

Carreau:
Fixes #10526 , but as noted in there I think we should think about passing extra arguments to `_repr_mimetype_` I even think we __should__ require that it takes arbitrary kwargs to be able to pass more informations later on without breaking API. 

cc @minrk and @rgbkrk

### DIFF
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -91,7 +91,7 @@ class DisplayFormatter(Configurable):
 
         By default all format types will be computed.
 
-        The following MIME types are currently implemented:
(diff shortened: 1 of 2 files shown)


Carreau:
@meeseeksdev backport

Carreau:
retry: 

@meeseeksdev backport

PR #10540, opened 2017-05-11
Verbose error message for syntax error occurring at runtime

piotrzielinski:
Fix for: https://github.com/ipython/ipython/issues/208

Added running_compiled_code flag to showbacktrace and showsyntaxerror to distinguish between compile and runtime case.


Carreau:
Thanks !

Issue #9796, opened 2016-07-23
Make term_title customizable


vbraun:
Right now you can only switch it on or off; It would be nice if it could be made to set the title to something else than a hardcoded "IPython: ..". E.g. Sage terminals would the easier to distinguish from plain IPython.

anantkaushik89:
Hello @vbraun, @Carreau  could you please give some more information about this? I would like to work on this issue and any pointers would be very helpful.

PR #10541, opened 2017-05-11
Add --term-title-format flag to customize terminal title format

toryhaavik:
Adds the feature described in #9796

### DIFF
--- a/IPython/core/magics/osm.py
+++ b/IPython/core/magics/osm.py
@@ -352,7 +352,7 @@ class OSMagics(Magics):
             try:
                 os.chdir(os.path.expanduser(ps))
                 if hasattr(self.shell, 'term_title') and self.shell.term_title:
-                    set_term_title('IPython: ' + abbrev_cwd())
+                    set_term_title(self.shell.term_title_format.format(cwd=abbrev_cwd()))
             except OSError:
                 print(sys.exc_info()[1])
             else:
@@ -365,7 +365,7 @@ class OSMagics(Magics):
         else:
             os.chdir(self.shell.home_dir)
             if hasattr(self.shell, 'term_title') and self.shell.term_title:
-                set_term_title('IPython: ' + '~')
+                set_term_title(self.shell.term_title_format.format(cwd="~"))
             cwd = os.getcwd()
             dhist = self.shell.user_ns['_dh']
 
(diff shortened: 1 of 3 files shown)


IPython/terminal/ipapp.py
```diff
@@ -154,6 +154,7 @@ def make_report(self,traceback):
 
 aliases = dict(base_aliases)
 aliases.update(shell_aliases)
+aliases['term-title-format'] = 'TerminalInteractiveShell.term_title_format'
```
Carreau: You actually do not need to do this, as you did `.tag(config=True)` the CLI parser has autogenerated a flag called `--TerminalInteractiveShell.term_title_format`,  and I'm unsure we need a short for of this one. @takluyver @ellisonbg thoughts ?

toryhaavik: Ah, that makes sense.  I agree a short option is probably unnecessary.

toryhaavik: Actually, I don't see that flag when I remove the alias, both in the CLI help and the profile config.  Is there something else I'm missing to get it to auto-generate the flag?

Carreau: It does not show up, but it does work 
screen shot 2017-05-11 at 13 36 05

So that's a separate bug.

Carreau: D'oh, I missed-grep'ed:

```
 ~/dev/ipython origin/pr/10541 $ ipython --help-all | grep -e 'term_title_format' -A3
--TerminalInteractiveShell.term_title_format=<Unicode>
    Default: 'IPython: {cwd}'
    Customize the terminal title format.  This is a python format string.
    Available substitutions are: {cwd}.

~/dev/ipython origin/pr/10541 $ git diff
diff --git a/IPython/terminal/ipapp.py b/IPython/terminal/ipapp.py
index 11c1503a4..19f419999 100755
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -154,7 +154,6 @@ def make_report(self,traceback):

 aliases = dict(base_aliases)
 aliases.update(shell_aliases)
-aliases['term-title-format'] = 'TerminalInteractiveShell.term_title_format'

 # Main classes and functions
```

You probably did not knew about the --help-all flag. 
It's missing from the configuration file as the configuration file only get the commented options IPython knew about at config file creation. You you would need to nuke your file , and re-do an `ipython profile create`

toryhaavik: Ah, yep.  I'll update the PR with the change.

Carreau: Thanks ! I'm make the `--help-all`  [more visible](https://github.com/ipython/traitlets/pull/404)

Carreau:
Thanks !

Issue #9404, opened 2016-04-15
Make tab completion of kwargs context-sensitive


andreas-h:
I'm running IPython 4.0.0 under Python 3.5.1.

Tab completion of kwarg definitions in function signatures should be more aware of the cursor's surroundings.  More specifically, when using Tab completion on a kwarg in a function definition, and the cursor is at the `=` sign of the kwarg definition, the Tab completion should *not* insert another `=` as part of the completion.

I'm not sure I described the situation properly; here's my use case:

Say I have defined a function

    def func(arg1, config=None): print(configfile)

During the interactive IPython session, I'm calling this function repeatedly:

    a1 = func(1, 2)
    a2 = func(2, 3)

and so on.  At some later stage, I change the function to read

    def func(arg1, configfile=None): print(configfile)

so I renamed the kwarg `config` to `configfile`.  If now I use UP and DOWN arrows to move through history, insert the `f` of `configfile` after the `config` that's in my history, and press ENTER, the tab completion inserts `ile=` so my command line has two `=` signs:

    In [1]: def func(arg1, configfile=None): print(configfile)
    
    In [2]: func(None, config=1) # press TAB when cursor at =
    %config      configfile=  
    
    In [2]: func(None, configf=1) # press TAB when cursor at =
    
    In [2]: func(None, configfile==1)

PR #10547, opened 2017-05-11
Do not duplicate '=' when tab completing a kwarg with the cursor on '='

bartz:
Fix for #9404:

Adjust completion text when the last character is an '=' and the cursor is on a '='

### DIFF
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -43,6 +43,11 @@ def _elide(string, *, min_elide=30):
     return '{}.{}\N{HORIZONTAL ELLIPSIS}{}.{}'.format(parts[0], parts[1][0], parts[-2][-1], parts[-1])
 
 
+def _adjust_completion_text_based_on_context(text, body, offset):
+    if text.endswith('=') and len(body) > offset and body[offset] is '=':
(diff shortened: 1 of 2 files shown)


IPython/terminal/ptutils.py
```diff
@@ -43,6 +43,11 @@ def _elide(string, *, min_elide=30):
     return '{}.{}\N{HORIZONTAL ELLIPSIS}{}.{}'.format(parts[0], parts[1][0], parts[-2][-1], parts[-1])
 
 
+def _adjust_completion_text_based_on_context(text, body, offset):
```
takluyver: Could the naming be a bit clearer about what this function does? 'adjust based on context' is rather generic. Or maybe add a comment explaining the idea.

takluyver:
Thanks. Do you also want to add a brief description in `docs/source/whatsnew/pr/`, to go into the release notes.

Carreau:
> Thanks. Do you also want to add a brief description in docs/source/whatsnew/pr/, to go into the release notes.

I think we can do that ourself when we write the release notes. It should be about time to release IPyhon 6.1 in a week or so as I believe we have a couple of bugfixes.

PR #10549, opened 2017-05-11
ENH: Add alias_magic parameters

### LINKED ISSUE #2965: %alias_magic to accept argument to %magic
It would be great to have an ability to pass argument to `%alias_magic`. For example, `%alias_magic h history -l 30` so I can just type `h` to get 30 lines of last history.

nickweseman:
Added functionality and tests to add the ability to save a magic function with parameters via `alias_magic`.  It was previously only possible to store the magic function itself with `alias_magic`.

Please see `aliasmagicparams-feature.rst` for details.

Fixes #2965.

### DIFF
--- a/IPython/core/magic.py
+++ b/IPython/core/magic.py
@@ -426,7 +426,7 @@ class MagicsManager(Configurable):
         setattr(self.user_magics, magic_name, func)
         record_magic(self.magics, magic_kind, magic_name, func)
 
-    def register_alias(self, alias_name, magic_name, magic_kind='line'):
+    def register_alias(self, alias_name, magic_name, magic_kind='line', magic_params=None):
         """Register an alias to a magic function.
 
         The alias is an instance of :class:`MagicAlias`, which holds the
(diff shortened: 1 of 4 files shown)


IPython/core/magic.py
```diff
@@ -675,6 +676,9 @@ def __call__(self, *args, **kwargs):
                              "magic aliases cannot call themselves.")
         self._in_call = True
         try:
-            return fn(*args, **kwargs)
+            if self.magic_params:
```
takluyver: It looks like an alias created with parameters can't be called with any more parameters - you either pass parameters in while creating the alias,  or when you call it, but not both. That may well be a reasonable restriction, but I would have expected that any parameters specified when you call it would be added after the parameters defined with the alias.

nickweseman: @takluyver, thanks for taking a look at my pull request!  Yea, that's the interpretation I had, but after thinking about it, I like your interpretation better.  It gives the user more flexibility and closer mimics the unix shell behavior.

I've updated the pull request to include the behavior you cited above which adds any called parameters after the alias parameters.  Can you take a look?

Carreau:
Awesome ! Thanks !

PR #10530, opened 2017-05-11
Adding a -q option to the logstart magic to suppress output of log state

joshuastorck:
Also added a command line option. Resolves #3398.

Issue #3398, opened 2013-06-04
Option to disable printing logstate on log activation


jagguli:
/usr/lib/python2.7/site-packages/IPython/core/magics/logging.py Line 135

printing the log state everytime shell is opened is annoying, please make it configurable.

PR #10552, opened 2017-05-12
Adding whatsnew documentation for the -q flag to logstart

joshuastorck:
This is for #3398 which was resolved in #10530.


takluyver:
Thanks :-)

PR #10558, opened 2017-05-13
Let nbformat take care of opening the file

takluyver:
Notebook files should always be read as utf-8, but the default for `io.open()` is to use a platform-dependent default encoding. The easiest way around this is to pass the filename to nbformat.read() and let it open the file correctly.

Bug identified at: http://stackoverflow.com/questions/43915006/encoding-error-in-jupyter-when-run-another-notebook


Carreau:
@meeseeksdev backport

PR #10559, opened 2017-05-13
Fix test for changes in traitlets

takluyver:
The output from '%config TerminalInteractiveShell' changed slightly, causing a spurious test failure on the jack-of-none builds on Jenkins.


Carreau:
Thanks  !

PR #10561, opened 2017-05-13
Add completion for %colors magic

### LINKED ISSUE #8901: Autocompletion of %colors
No autocompletion is implemented for %colors currently.  Arguably a low-priority item.

tomowind:
Fixes #8901

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -136,6 +136,7 @@ from traitlets.config.configurable import Configurable
 from IPython.core.error import TryNext
 from IPython.core.inputsplitter import ESC_MAGIC
 from IPython.core.latex_symbols import latex_symbols, reverse_latex_symbol
+from IPython.core.oinspect import InspectColors
 from IPython.utils import generics
 from IPython.utils.dir2 import dir2, get_real_method
 from IPython.utils.process import arg_split
@@ -1049,11 +1050,14 @@ class IPCompleter(Completer):
         self.matchers = [
                          self.python_matches,
                          self.file_matches,
-                         self.magic_config_matches,
                          self.magic_matches,
                          self.python_func_kw_matches,
                          self.dict_key_matches,
                          ]
+        self.magic_arg_matchers = [
+            self.magic_config_matches,
+            self.magic_color_matches,
+        ]
 
         # This is set externally by InteractiveShell
         self.custom_completers = None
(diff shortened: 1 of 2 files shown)


IPython/core/completer.py
```diff
@@ -1879,6 +1891,14 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
         self.line_buffer = line_buffer
         self.text_until_cursor = self.line_buffer[:cursor_pos]
 
+        # Do magic arg matches
+        for matcher in self.magic_arg_matchers:
+            matches = [(m, matcher.__qualname__) for m in matcher(line_buffer)]
```
takluyver: I don't think we need to put `matcher.__qualname__` into the list - it's going to be the same for every item anyway.

IPython/core/completer.py
```diff
@@ -1879,6 +1891,14 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
...
+            if matches:
+                matches2 = [m[0] for m in matches]
+                origins = [m[1] for m in matches]
+                return text, matches2, origins, {}
```
takluyver: It looks like the fourth return value below is a list or a tuple, not a dict.

IPython/core/completer.py
```diff
@@ -1182,13 +1186,11 @@ def magic_matches(self, text):
 
         return [cast_unicode_py2(c) for c in comp]
 
-    def magic_config_matches(self, text):
+    def magic_config_matches(self, line_buffer):
```
Carreau: nitpicking but changing a positional parameter name, is a change of API as it can be used a a keyword argument.

Carreau: Adding on that, but because it's new and has not been released it ok :-)  Just noting it.

IPython/core/completer.py
Carreau: Hum, I missed that it can return 4 things, I'l have to dig into that. The type annotation is wrong on this function for sure then.

IPython/core/completer.py
Carreau: You actually get two of these: 
```
self.magic_arg_matchers = [
            self.magic_config_matches,
            self.magic_color_matches,
        ]
```

Origin is not super important but useful for debug, and to deduplicate completions in a smart ways. It's cheap enough o keep.

Carreau:
Rebased and fix a couple of return type where `{}` was a set. I also added a couple of type annotations at the same time.

tomowind:
Thanks. Let me pull the changes and revisit this one.

IPython/core/completer.py
takluyver: Right, but you never mix entries from the two matchers. You could achieve the same thing like this:

```python
for matcher in self.magic_arg_matchers:
    matches = matcher(line_buffer)
    if matches:
        origins = [matcher.__qualname__] * len(matches)
        return text, matches, origins, ()
```

i.e. skip putting the origins into a list-of-tuples and then separating them back out.

Carreau:
This work, I'm going to merge ahead, and we can still make subsequent PRs if necessary.

PR #10569, opened 2017-05-17
Some extra info about tab completion in documentation.

Carreau:


ivanov:
I added some minor edits, primarily to use the `:kbd:` directive for the key combinations. Thanks, @Carreau !

Carreau:
<3

ivanov:
Hmmm... the GitHub render for ```:kbd:`Ctrl-O`.``` seems to be unhappy... I wonder if it wants the `O` to be lower case...

ivanov:
Oh, oops, it's not that picky, I had swapped `kdb` for `kbd` - that's why it wasn't rendering... - fixed and force pushed now...

ivanov:
Oh, and by the way - should we  continue to write " if ``jedi`` is installed, "? Seems like [we install it unconditionally](https://github.com/ipython/ipython/blob/master/setup.py#L188)

PR #10570, opened 2017-05-18
Remove python2 code

srinivasreddy:
revert 147b245

### DIFF
--- a/IPython/core/page.py
+++ b/IPython/core/page.py
@@ -106,18 +106,6 @@ def _detect_screen_size(screen_lines_def):
         # can fail on Linux 2.6, pager_page will catch the TypeError
         raise TypeError('termios error: {0}'.format(err))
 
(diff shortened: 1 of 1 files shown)


takluyver:
The issue describing why this code was added (http://bugs.python.org/issue10144 ) says that the problem wasn't seen on Python 3, but I don't understand it well enough to be totally confident that it can't come up on Python 3. There may be some corner case where it still comes up.

srinivasreddy:
may be a test case to keep watch on that doesn't occur again makes this PR merged?

takluyver:
I don't think it's easy to test, especially if it affects some corner case that a straightforward test is unlikely to exercise. We may just decide to merge it anyway, as the original problem discovered didn't affect Python 3.

Carreau:
I think it's been long enough. Let's try it.

PR #10574, opened 2017-05-18
remove python2 specific cast_unicode_py2 function

srinivasreddy:


IPython/core/ultratb.py
```diff
@@ -382,8 +381,6 @@ def _format_traceback_lines(lnum, index, lines, Colors, lvals=None,  _line_forma
     i = lnum - index
 
     for line in lines:
-        line = py3compat.cast_unicode(line)
```
takluyver: Careful, this one *isn't* a no-op on Python 3. `cast_unicode` will check for bytes on both Python versions.

IPython/core/ultratb.py
```diff
@@ -688,7 +685,7 @@ def _format_exception_only(self, etype, value):
         have_filedata = False
         Colors = self.Colors
         list = []
-        stype = py3compat.cast_unicode(Colors.excName + etype.__name__ + Colors.Normal)
```
takluyver: In this case, I don't think it could ever have been bytes on Python 3, because `Colors.excName` will always be a str.

IPython/core/ultratb.py
```diff
@@ -704,10 +701,10 @@ def _format_exception_only(self, etype, value):
                     textline = ''
                 list.append('%s  File %s"%s"%s, line %s%s%s\n' % \
                             (Colors.normalEm,
-                             Colors.filenameEm, py3compat.cast_unicode(value.filename), Colors.normalEm,
```
takluyver: But here are several more `cast_unicode` which might be doing something.

takluyver:
Mostly fine, but there are a few cases where you've removed `cast_unicode` as well as `cast_unicode_py2` - the former may still be doing something useful on Python 3.

srinivasreddy:
rebase made the PR to close.  :disappointed:

srinivasreddy:
just wondering why appveyor tests weren't triggered?

takluyver:
No idea, but I'm happy enough with this.

PR #10590, opened 2017-05-24
Fix a couple of types annotation and actual types errors.

Carreau:
Running basic mypy just on the completer.


Carreau:
There is a reminding bytes+str/str+bytes warning if you run 

```bash
$ mypy IPython/core/completer.py --ignore-missing-imports  --follow-import skip
```

PR #10591, opened 2017-05-24
remove unused  imports

srinivasreddy:


IPython/extensions/tests/test_autoreload.py
```diff
@@ -132,7 +132,6 @@ class TestAutoreload(Fixture):
 
     @skipif(sys.version_info < (3, 6))
     def test_reload_enums(self):
-        import enum
```
takluyver: @Carreau you wrote this test; is it possible that it relies on enum being imported before the rest of the test runs, for some reason? I know that testing autoreload can be a bit fiddly.

IPython/utils/_process_cli.py
```diff
@@ -11,16 +11,11 @@
 in the details. 
 """
 
-# Import cli libraries:
-import clr
```
takluyver: Let's leave this alone; it's possible that it needs to import this first to use CLI libraries. I don't know much about how IronPython works, and our CI doesn't cover it.

IPython/utils/_process_cli.py
```diff
@@ -11,16 +11,11 @@
...
-import System
 
-# Import Python libraries:
+import System
 import os
-
-# Import IPython libraries:
 from IPython.utils import py3compat
-from ._process_common import arg_split
```
takluyver: This needs to be there - `process.py` expects to be able to import it from this file.

IPython/extensions/tests/test_autoreload.py
Carreau: unsure. I think it's more a safeguard that we can import enum.

IPython/paths.py
```diff
@@ -9,8 +9,7 @@
 from IPython.utils.importstring import import_item
 from IPython.utils.path import (
     get_home_dir, get_xdg_dir, get_xdg_cache_dir, compress_user, _writable_dir,
-    ensure_dir_exists, fs_encoding, filefind
-)
+    ensure_dir_exists, fs_encoding)
```
Carreau: I'm worries of removing `filefind` as this is user exposed.

IPython/extensions/tests/test_autoreload.py
takluyver: Do you think it should be restored? This test is already skipped on Python < 3.6, so we should always be able to import enum. Might be nicest to have a clear error if that goes wrong for some reason, though.

Carreau:
Thanks for this, but please this is not the interesting part of working on IPython, that's mostly blindly following tools, and I will suggest to watch [behind pep 8](https://www.youtube.com/watch?v=wf-BqAjZb8M). There is a number os issues tagged [Need contributors](https://github.com/ipython/ipython/issues?q=is%3Aopen+is%3Aissue+label%3A%22Need+Contributor%22) which are better for you to learn, and will bring new features and fix bugs for users. They might be harder to solve, but they will actually exercise your skills. You should look at these.

IPython/paths.py
takluyver: It is [documented](http://ipython.readthedocs.io/en/stable/api/generated/IPython.utils.path.html#IPython.utils.path.filefind) as part of `IPython.utils.path`, though, not this module. Do we know of people using it from here? I think that if we start worrying in general about things that are accessible because they're imported into that module, we'll go mad.

srinivasreddy:
@Carreau Thanks for write up and i agree with you.
Will fix those issues you mentioned.

srinivasreddy:
@takluyver  @Carreau  Done

Carreau:
Thanks !

PR #10593, opened 2017-05-24
remove py2 code

Carreau:
MyPy get confused as the same variable get 2 types and that's for Python2, so remove.

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -484,11 +484,6 @@ class GroupQueue(object):
         except ValueError:
             pass
 
-try:
-    _baseclass_reprs = (object.__repr__, types.InstanceType.__repr__)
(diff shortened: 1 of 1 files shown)


IPython/lib/pretty.py
```diff
@@ -484,10 +484,7 @@ def remove(self, group):
         except ValueError:
             pass
 
-try:
-    _baseclass_reprs = (object.__repr__, types.InstanceType.__repr__)
-except AttributeError:  # Python 3
-    _baseclass_reprs = (object.__repr__,)
```
takluyver: As it's a private variable, we could get rid of it entirely and just check `is not object.__repr__` below.

Carreau: done.

PR #10594, opened 2017-05-24
Prevent recustion in autoreload.

### LINKED ISSUE #8771: autoreload module, and new instance methods?
Is there a reason why update_class in ipython/autoreload.py doesn't add new instance methods to the class dict?  Wouldn't that work?

Thanks

### LINKED ISSUE #10214: Fix for infinite recursion in autoreload
I have a class which memoizes contains a link to itself as an attribute. This is used as a to ensure that a class specific memoized computation is unique and not inherited from bases.

It turns out that this causes an infinite recursion between update_generic and update_class in autoreload.

This is fixed with a quick check to prevent this case. added after the "prevent recursion" comment.

Furthermore in this code, the for loop also loops on "new" object attributes, which will fix #8771

```python
def update_class(old, new):
    """Replace stuff in the __dict__ of a class, and upgrade
    method code objects"""
    #for key in list(old.__dict__.keys()):
    for key in set(list(old.__dict__.keys()) + list(new.__dict__.keys())):
        old_obj = getattr(old, key)

        try:
            new_obj = getattr(new, key)
        except AttributeError:
            # obsolete attribute: remove it
            try:
                delattr(old, key)
            except (AttributeError, TypeError):
                pass
            continue

        #prevent recursion 
        if old_obj not in [old, new] and new_obj not in [old, new]:
            if update_generic(old_obj, new_obj): continue

        try:
            setattr(old, key, getattr(new, key))
        except (AttributeError, TypeError):
            pass # skip non-writable attributes
```

Carreau:
closes #10214, closes #8771

We don't have a test case :-( but if it fixes things for someone, I'm
inclined to get that in.

### DIFF
--- a/IPython/extensions/autoreload.py
+++ b/IPython/extensions/autoreload.py
@@ -272,7 +272,7 @@ def update_function(old, new):
 def update_class(old, new):
     """Replace stuff in the __dict__ of a class, and upgrade
     method code objects"""
-    for key in list(old.__dict__.keys()):
+    for key in set(list(old.__dict__.keys()) + list(new.__dict__.keys())):
         old_obj = getattr(old, key)
         try:
             new_obj = getattr(new, key)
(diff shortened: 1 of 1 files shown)


Carreau:
Well apparently that does break things.. :-( bumping to figure that out.

Carreau:
closing for now to keep PR queue clean I can reopen later.

PR #10596, opened 2017-05-24
Inject display into builtins

Carreau:
display is an underutilized feature of IPython which is relatively hard
to access to. You need to explicitly import from `IPython.display`,
which not only is long to type, but users get confused if they should
use `IPython.display` or `IPython.display.display` also many tutorials
tell users to import from IPython.core.display which is not right.

So inject that directly in builtin module – so that user can overwrite
and delete without issues – and document display a lot more.

cc @rgbkrk, @ivanov

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -145,6 +145,9 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     By default all representations will be computed and sent to the frontends.
     Frontends can decide which representation is used and how.
 
+    In terminal IPython this will be similar to using `print`, for use in richer
+    frontends see Jupyter notebook examples with rich display logic.
+
     Parameters
     ----------
     objs : tuple of objects
@@ -173,19 +176,109 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
         If given as True, generate a new display_id
     kwargs: additional keyword-args, optional
         Additional keyword-arguments are passed through to the display publisher.
-    
+
     Returns
     -------
-    
+
     handle: DisplayHandle
         Returns a handle on updatable displays, if display_id is given.
         Returns None if no display_id is given (default).
+
+    Examples
+    --------
+
+    >>> class Json(object):
+    ...     def __init__(self, json):
+    ...         self.json = json
+    ...     def _repr_pretty_(self, pp, cycle):
+    ...         import json
+    ...         pp.text(json.dumps(self.json, indent=2))
+    ...     def __repr__(self):
+    ...         return str(self.json)
+    ...
+
+    >>> d = Json({1:2, 3: {4:5}})
+
+    >>> print(d)
+    {1: 2, 3: {4: 5}}
+
+    >>> display(d)
+    {
+      "1": 2,
+      "3": {
+        "4": 5
+      }
+    }
+
+    >>> def int_formatter(integer, pp, cycle):
+    ...     pp.text('I'*integer)
+
+    >>> plain = get_ipython().display_formatter.formatters['text/plain']
+    >>> plain.for_type(int, int_formatter)
+    <function _repr_pprint at 0x...>
+    >>> display(7-5)
+    II
+
+    >>> del plain.type_printers[int]
+    >>> display(7-5)
+    2
+
+    See Also
+    --------
+
+    `update_display`
+
+    Notes
+    -----
+
+    In Python, objects can declare their textual representation using the
+    `__repr__` method. IPython expands on this idea and allows objects to declare
+    other, rich representations including:
+
+      - HTML
+      - JSON
+      - PNG
+      - JPEG
+      - SVG
+      - LaTeX
+
+    A single object can declare some or all of these representations; all are
+    handled by IPython's display system.
+
+    The main idea of the first approach is that you have to implement special
+    display methods when you define your class, one for each representation you
+    want to use. Here is a list of the names of the special methods and the
+    values they must return:
+
+      - `_repr_html_`: return raw HTML as a string
+      - `_repr_json_`: return a JSONable dict
+      - `_repr_jpeg_`: return raw JPEG data
+      - `_repr_png_`: return raw PNG data
+      - `_repr_svg_`: return raw SVG data as a string
+      - `_repr_latex_`: return LaTeX commands in a string surrounded by "$".
+      - `_repr_mimebundle_`: return a full mimebundle containing the mapping
+      from all mimetypes to data
+
+    When you are directly writing your own classes, you can adapt them for
+    display in IPython by following the above approach. But in practice, you
+    often need to work with existing classes that you can't easily modify.
+
+    You can refer to the documentation on IPython display formatters in order to
+    register custom formatters for already existing types.
+
+    Since IPython 5.4 and 6.1 display is automatically made available to the
+    user without import. If you are using display in a document that might be
+    used in a pure python context or with older version of IPython, use the
+    following import at the top of your file::
+
+        from IPython.display import display
+
     """
     raw = kwargs.pop('raw', False)
     if transient is None:
         transient = {}
     if display_id:
-        if display_id == True:
+        if display_id is True:
             display_id = _new_id()
         transient['display_id'] = display_id
     if kwargs.get('update') and 'display_id' not in transient:
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -145,7 +145,7 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     By default all representations will be computed and sent to the frontends.
     Frontends can decide which representation is used and how.
 
-    In terminal IPython this will be similar to using `print`, for use in richer
+    In terminal IPython this will be similar to using :func:`print`, for use in richer
     frontends see Jupyter notebook examples with rich display logic.
 
     Parameters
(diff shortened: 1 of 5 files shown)


rgbkrk:
🙇  Thank you!

IPython/core/display.py
```diff
@@ -173,19 +176,108 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
...
+    ...     pp.text('I'*integer)
+
+    >>> plain = get_ipython().display_formatter.formatters['text/plain']
+    >>> plain.for_type(int, int_formatter)
+    >>> display(7 - 5)
+    II
+
+    >>> del plain.type_printers[int]
+    >>> display(3+6)
+    9
```
rgbkrk: This is so great to have these examples.

IPython/core/interactiveshell.py
```diff
@@ -618,6 +619,7 @@ def init_builtins(self):
         # removing on exit or representing the existence of more than one
         # IPython at a time.
         builtin_mod.__dict__['__IPYTHON__'] = True
+        builtin_mod.__dict__['display'] = display
```
rgbkrk: away we go!

rgbkrk (approved):

IPython/core/display.py
Carreau: They are also in the Custom Display logic notebook.

rgbkrk:
I was recently alerted to the fact that pixiedust exposes `display` as a top level for use in the notebook, likely to match up with zeppelin/databricks/etc., which is also why I said we should expose `display` by default.

The extra thing most of these other displayers do though is hook the display for data frames (spark and pandas) and Spark GraphFrames.

Carreau:
> I was recently alerted to the fact that pixiedust exposes display as a top level for use in the notebook, likely to match up with zeppelin/databricks/etc., which is also why I said we should expose display by default.

Where is it injected ? in user_ns or builtins ? AFAICT that will lead to user having:

```
Pixiedust didn't find any vizualization plugins that can render this object: <__main__.Foo object at 0x7fd58600d810>
```

Or whatever object is given to it. I think that pixiedust should fallback on ipython's display. 
It should even to hook into IPython formatter. Right now if you `display(df)` you get a nice rich representation. If you just return it as the last statemtn of the cell you get a `DataFrame[Colors: string, %: bigint]` output, which is less nice. 

Registering formatter should allow to have rich representation in both case which should make the experience more uniform for users. 

The other things is that with this PR IPython's display is restored as soon as you delete any things that shadows it:

```
$ ipython
IPython 6.1.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: display
Out[1]: <function IPython.core.display.display>

In [2]: display = lambda x:X

In [3]: display
Out[3]: <function __main__.<lambda>>

In [4]: del display

In [5]: display
Out[5]: <function IPython.core.display.display>
```

IPython/core/display.py
ivanov: Totally minor nitpick, but it'd be even more clear if the display calls both used the same expression inside... 
Make both `display(3+6)` for example, so the top one is even more clear about how it's doing the formatting...

IPython/core/display.py
```diff
@@ -173,19 +176,109 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
...
+      from all mimetypes to data
+
+    When you are directly writing your own classes, you can adapt them for
+    display in IPython by following the above approach. But in practice, you
+    often need to work with existing classes that you can't easily modify.
+
+    You can refer to the documentation on IPython display formatters in order to
+    register custom formatters for already existing types.
+
+    since IPython 5.4 and 6.1 display is automatically made availlable to the
```
ivanov: " **S**ince IPython 5.4 and 6.1 **,** "  (capital S, comma after 6.1)

IPython/core/display.py
```diff
@@ -173,19 +176,109 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
...
+    user without import. If you are using display in a document that might be
+    used in a pure python context or with older version of IPython use the
```
ivanov: comma after IPython

IPython/core/display.py
```diff
@@ -173,19 +176,109 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
...
+
+        from IPython.display import display
+
     """
     raw = kwargs.pop('raw', False)
     if transient is None:
         transient = {}
     if display_id:
-        if display_id == True:
+        if display_id is True:
```
ivanov: Nice bug fix! `display_id=1` would have always generated a new ID for this.

IPython/core/displayhook.py
```diff
@@ -16,6 +16,8 @@
 from traitlets import Instance, Float
 from warnings import warn
 
+from IPython.display import display
```
ivanov: Not sure why this import is here. Could you document that if it needs to stay here?

ivanov (commented):
Looks sweet, Matthias! With the display_update stuff, I think `display` is an even more powerful feature now that should available more prominently.

rgbkrk:
> The other things is that with this PR IPython's display is restored as soon as you delete any things that shadows it:

Excellent!

IPython/core/display.py
```diff
@@ -145,6 +145,9 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     By default all representations will be computed and sent to the frontends.
     Frontends can decide which representation is used and how.
 
+    In terminal IPython this will be similar to using `print`, for use in richer
```
ivanov: Comma after `IPython` here as well.

IPython/core/interactiveshell.py
DTAIEB: > I think that pixiedust should fallback on ipython's display

Makes sense, I'll enter a Github issue in PixieDust repo

minrk (approved):

minrk:
Great!

@ivanov since you mention the update display stuff, do you think `update_display` should be included, or is that too much? I realize builtins is a pretty precious namespace.

Minor technical detail: should this go in `.user_ns_hidden` instead of builtins, so it doesn't show up as available in module code?

Carreau:
> Minor technical detail: should this go in .user_ns_hidden instead of builtins, so it doesn't show up as available in module code?

It can, I did that originally, it's more complicated if we want to re-inject it when user delete it/shadow it, but doable.

I believe `update_display` is more a technical under the hood detail, while everyone should know about (and use) display.

This is still a relatively heavy usability change so I'm probing what other are thinking. I'd like @fperez input on this one as well (not on the code itself, but happy to get review) but wether it's ok to have `display()` available by default.

IPython/core/interactiveshell.py
Carreau: > Makes sense, I'll enter a Github issue in PixieDust repo

Thanks !

Or, if we can figure out a way to use IPython machinery for pixeldust (as we already have dispatch per object types) that would be great. Feel free to crosslink here or ping me.

rgbkrk:
I agree that we shouldn't surface update_display. We can still use update = True on display anyways.

Carreau:
On the `builtins` vs `user_ns_hidden`:

When in user_ns_hidden it happily accept to delete it but the object will reapear:

```
$ ipython
IPython 6.1.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: display
Out[1]: <function IPython.core.display.display>

In [2]: display = 1

In [3]: del display

In [4]: display
Out[4]: <function IPython.core.display.display>

In [5]: del display

In [6]: display
Out[6]: <function IPython.core.display.display>
```

If in builtins, il will complain  that object can't be found if trying to delete it:

```
$ ipython
IPython 6.1.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: display
Out[1]: <function IPython.core.display.display>

In [2]: display = 1

In [3]: display
Out[3]: 1

In [4]: del display

In [5]: display
Out[5]: <function IPython.core.display.display>

In [6]: del display
<ipython-input-6-78e18c419bea> in <module>()
----> 1 del display

NameError: name 'display' is not defined
```

The user_ns_hidden also have the case that the object reapears only after each cell execution so cutting cell may change behavior. For example NameError  vs NoError
```
In [1]: display=1
   ...: print(display)
   ...: del display
   ...: print(display)
   ...:
1
<ipython-input-2-bb2e1216e38f> in <module>()
      2 print(display)
      3 del display
----> 4 print(display)

NameError: name 'display' is not defined

In [2]: display = 1
   ...: print(display)
   ...: del display
1

In [3]: print(display)
<function display at 0x102dfbd08>
```

We can also not reinject it, but I believe it is the wrong solution.

Carreau:
Or, we can change the exec for `exec(code_obj, self.user_global_ns, ChainMap(self.user_ns, hidden_ns))`

(you can't ChainMap global, exec refuses), that should avoid a lot of the hidden_ns dance we do.

Carreau:
Rebased, updated, tested, documented.

rgbkrk (approved):

Carreau:
`@fperez` seem to +1, 
`@jasongrout` was a good devil advocate and seem to agree this is a good idea
`@willingc` seem to thing for users. 

There is one  (longstanding) issue that calling display in a pure Python terminal break the repl by instantiating an InteractiveShell (which has crazy side effects like nuking your namespace and setting sys.ps1), I'm unsure it's a bloker for this PR>

takluyver:
But that will only happen if you start a plain Python shell after starting IPython in the same process, right? I don't think that comes up very often.

Carreau:
> But that will only happen if you start a plain Python shell after starting IPython in the same process, right? I don't think that comes up very often.

Not even, it break if you just call display in a pure python repl. We can fix that, the only question if what to do. If Interactiveshell._instance is None fallback on print ? Even if raw=True ? 

```
$ python
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
>>> from IPython.display import display
>>> a = 1
>>> display(a)
1
In : a
NameError: name 'a' is not defined
In :
```

But that's another issue.

rgbkrk:
@meseeksdev backport

rgbkrk:
@meeseeksdev backport

takluyver:
Oh, I see, you mean if code explicitly imports it. That's not affected by injecting display into builtins when IPython runs, though.

Carreau:
> Oh, I see, you mean if code explicitly imports it. That's not affected by injecting display into builtins when IPython runs, though.

Yes, sorry the missing datapoint is:
If we inject it into builtins, we might want `nbconvert --to script` to add the import at the top.

takluyver:
Ah, that makes sense. Yeah, that's tricky.

PR #10598, opened 2017-05-24
Add new interactivity mode for ast.

Carreau:
Allow to display the result of the last assignment of it has a unique
target.

meh, I should write tests for that.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -79,6 +79,9 @@ from warnings import warn
 from logging import error
 import IPython.core.hooks
 
+from typing import List as ListType
+from ast import AST
+
 # NoOpContext is deprecated, but ipykernel imports it from here.
 # See https://github.com/ipython/ipykernel/issues/157
 from IPython.utils.contexts import NoOpContext
@@ -101,6 +104,13 @@ class ProvisionalWarning(DeprecationWarning):
     Warning class for unstable features
     """
     pass
+
+if sys.version_info > (3,6):
+    _assign_nodes         = (ast.AugAssign, ast.AnnAssign, ast.Assign)
+    _single_targets_nodes = (ast.AugAssign, ast.AnnAssign)
+else:
+    _assign_nodes         = (ast.AugAssign, ast.Assign )
+    _single_targets_nodes = (ast.AugAssign, )
 
 #-----------------------------------------------------------------------------
 # Globals
@@ -376,11 +386,12 @@ class InteractiveShell(SingletonConfigurable):
         """
     ).tag(config=True)
 
-    ast_node_interactivity = Enum(['all', 'last', 'last_expr', 'none'],
+    ast_node_interactivity = Enum(['all', 'last', 'last_expr', 'none', 'last_expr_or_assign'],
                                   default_value='last_expr',
                                   help="""
-        'all', 'last', 'last_expr' or 'none', specifying which nodes should be
-        run interactively (displaying output from expressions)."""
+        'all', 'last', 'last_expr' or 'none', 'last_expr_or_assign' specifying
+        which nodes should be run interactively (displaying output from expressions).
+        """
     ).tag(config=True)
 
     # TODO: this part of prompt management should be moved to the frontends.
@@ -2749,7 +2760,7 @@ class InteractiveShell(SingletonConfigurable):
         return node
                 
 
-    def run_ast_nodes(self, nodelist, cell_name, interactivity='last_expr',
+    def run_ast_nodes(self, nodelist:ListType[AST], cell_name:str, interactivity='last_expr',
                         compiler=compile, result=None):
         """Run a sequence of AST nodes. The execution mode depends on the
         interactivity parameter.
@@ -2762,11 +2773,13 @@ class InteractiveShell(SingletonConfigurable):
           Will be passed to the compiler as the filename of the cell. Typically
           the value returned by ip.compile.cache(cell).
         interactivity : str
-          'all', 'last', 'last_expr' or 'none', specifying which nodes should be
-          run interactively (displaying output from expressions). 'last_expr'
-          will run the last node interactively only if it is an expression (i.e.
-          expressions in loops or other blocks are not displayed. Other values
-          for this parameter will raise a ValueError.
+          'all', 'last', 'last_expr' , 'last_expr_or_assign' or 'none',
+          specifying which nodes should be run interactively (displaying output
+          from expressions). 'last_expr' will run the last node interactively
+          only if it is an expression (i.e. expressions in loops or other blocks
+          are not displayed) 'last_expr_or_assign' will run the last expression
+          or the last assignment. Other values for this parameter will raise a
+          ValueError.
         compiler : callable
           A function with the same interface as the built-in compile(), to turn
           the AST nodes into code objects. Default is the built-in compile().
@@ -2780,6 +2793,21 @@ class InteractiveShell(SingletonConfigurable):
         """
         if not nodelist:
             return
+
+        if interactivity == 'last_expr_or_assign':
+            if isinstance(nodelist[-1], _assign_nodes):
+                asg = nodelist[-1]
+                if isinstance(asg, ast.Assign) and len(asg.targets) == 1:
+                    target = asg.targets[0]
+                elif isinstance(asg, _single_targets_nodes):
+                    target = asg.target
+                else:
+                    target = None
+                if isinstance(target, ast.Name):
+                    nnode = ast.Expr(ast.Name(target.id, ast.Load()))
+                    ast.fix_missing_locations(nnode)
+                    nodelist.append(nnode)
+            interactivity = 'last_expr'
 
         if interactivity == 'last_expr':
             if isinstance(nodelist[-1], ast.Expr):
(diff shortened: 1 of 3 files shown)


Carreau:
@rgbkrk though on that ? I though about that because display() is long to type and I often do 
```
df = some_operation(df)
df 
```
Just to display it.

docs/source/whatsnew/pr/interactive_assignment.rst
```diff
@@ -0,0 +1,18 @@
+IPython can now trigger the display hook on last assignment of cells. 
+Up until 6.0 the following code wouldn't show the value of the assigned
+variable:
+
+    >>> var = "something"
```
rgbkrk: My brain almost broke because it was in JS mode while reading this.

Carreau: I can change for `let` if you want ?

rgbkrk:
If this behavior exists, out of old Matlab habit I will append `;` in hopes that it will not show the result by default.

I too do the operation then display it pretty often. I'd hate to have to figure out how to suppress output though.

rgbkrk:
Oh I see, this isn't the default setting, just a configurable. :cool:

Carreau:
yeah, of course it's not the default :-) 

```
$ ipython --InteractiveShell.ast_node_interactivity='last_expr_or_assign'
IPython 6.1.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: a = 1;

In [2]: a = 1
Out[2]: 1

In [3]:
```

`;` works, you can also do a newline / `pass` or newline / `None`

IPython/core/interactiveshell.py
```diff
@@ -2781,6 +2784,15 @@ def run_ast_nodes(self, nodelist, cell_name, interactivity='last_expr',
         if not nodelist:
             return
 
+        if interactivity == 'last_expr_or_assign':
+            if isinstance(nodelist[-1], ast.Assign):
+                asg = nodelist[-1]
+                if len(asg.targets) == 1:
+                    nnode = ast.Expr(ast.Name(asg.targets[0].id, ast.Load()))
```
ssanderson: I think this will crash if you try to do a multiple assignment.  When I parse something like `(a, b) = (1, 2)`, I get an assignment with a length-1 list of targets whose only entry is an instance of `ast.Tuple`, which has no `id`:

```
In [5]: sys.version_info
Out[5]: sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)

In [6]: %%ast
   ...: (a, b) = (1, 2)
   ...:
Module(
  body=[
    Assign(
      targets=[
        Tuple(
          elts=[
            Name(id='a', ctx=Store()),
            Name(id='b', ctx=Store()),
          ],
          ctx=Store(),
        ),
      ],
      value=Tuple(
        elts=[
          Num(1),
          Num(2),
        ],
        ctx=Load(),
      ),
    ),
  ],
)

Comment (ssanderson) [inline IPython/core/interactiveshell.py]: (That `%%ast` magic comes from https://github.com/llllllllll/codetransformer/blob/master/codetransformer/__init__.py#L23-L35, btw.  That might be relevant to your interests if you're working on ast stuff.)

Comment (Carreau) [inline IPython/core/interactiveshell.py]: Oh, sweet. Thanks. Does it also define a pretty formatter for ast objects that are returned ?

Comment (ssanderson) [inline IPython/core/interactiveshell.py]: It doesn't at the moment, but that could be added pretty easily.  The logic that does the actual formatting lives here: https://github.com/llllllllll/codetransformer/blob/master/codetransformer/utils/pretty.py#L31.

Comment (Carreau) [inline IPython/core/interactiveshell.py]: Hey ! ... it's not in the released version ! It's been merged a year ago, why no release? 
1pns0e

Comment (ssanderson) [inline IPython/core/interactiveshell.py]: Oh, TIL.  @llllllllll we should probably do a codetransformer release...

Comment (ssanderson) [inline IPython/core/interactiveshell.py]: [codetransformer 0.7.0 has been released](https://pypi.python.org/pypi/codetransformer/0.7.0)

**Comment by Carreau** [comment]
> codetransformer 0.7.0 has been released

Thanks ! Though I'm on 3.6, so can't import it :-( 

Pushed new commits that test the various behavior.

File: docs/source/whatsnew/pr/interactive_assignment.rst [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+You would have to actually make it the last statement:
+
+    >>> var = "someting else"
+    ... var 
+    "something else"
+
+With the option ``InteractiveShell.ast_node_interactivity='last_expr_or_assign'``
+you can now do::
+
+    >>> df = pandas.load_csv(...)
```
takluyver: Make this example match the others above? And use IPython prompts instead of `>>>`?

IPython/core/interactiveshell.py
```diff
@@ -2781,6 +2787,15 @@ def run_ast_nodes(self, nodelist, cell_name, interactivity='last_expr',
         if not nodelist:
             return
 
+        if interactivity == 'last_expr_or_assign':
+            if isinstance(nodelist[-1], ast.Assign):
```
takluyver: I think this is fine for getting this feature in and letting people experiment with it, but be aware that it won't catch `a += 1` (which is an `AugAssign` node), or `a : int = 1` (new in Python 3.6, `AnnAssign` node).

takluyver (commented):
Nice work :+1:

IPython/core/interactiveshell.py
Carreau: Good call. Fixed. It should work on all of these.

docs/source/whatsnew/pr/interactive_assignment.rst
Carreau: done.

IPython/core/tests/test_displayhook.py
```diff
@@ -53,3 +53,52 @@ def test_underscore_no_overrite_builtins():
...
+
+        ip.run_cell('d = 11', store_history=True)
+        with AssertPrints('12'):
+            ip.run_cell('d += 1', store_history=True)
+
+        with AssertNotPrints('42'):
+            ip.run_cell('(u,v) = (41+1, 43-1)')
+
+    finally:
+        ip.run_cell('a = 1+1', store_history=True)
```
takluyver: Why is this in the finally block?

Carreau: No reasons, bad copy past likely. Fixed.

Carreau:
@fperez would prefer for this to not assign to `Out[]` and `_`. 
I'm unsure about that. So I can rewrite the Ast to actually do `display(ast_name)`.
I'll see what I can do.

Carreau:
Hum inject something at the end like that:
```
In [10]: %%ast
    ...: __import__('IPython').display.display(whatever)
Module(
  body=[
    Expr(
      value=Call(
        func=Attribute(
          value=Attribute(
            value=Call(
              func=Name(id='__import__', ctx=Load()),
              args=[
                Str(
                  s='IPython',
                ),
              ],
              keywords=[],
            ),
            attr='display',
            ctx=Load(),
          ),
          attr='display',
          ctx=Load(),
        ),
        args=[
          Name(id='whatever', ctx=Load()),
        ],
        keywords=[],
      ),
    ),
  ],
)
```

At the same time when the expression of a of a cell display something there is always an `Out[]`, so I"m unsure.

takluyver:
If we're injecting `display()` into the builtins, that could be simpler. Although then I guess it would be susceptible to someone overriding `display =`.

Maybe there's a nicer way to do this - instead of modifying the AST, inspect it to find the target of the final assignment, and then record some information which says 'if this code completes successfully, print the value of `foo`'.

Carreau:
> Maybe there's a nicer way to do this - instead of modifying the AST, inspect it to find the target of the final assignment, and then record some information which says 'if this code completes successfully, print the value of foo'

Well that's almost exactly what the above do. it calls display on the last assignment target. 


Here is my argument against having it displayed and not triggerd as part of display hook.

One the the two below methods are doing 

```python
def load_something():
    display(...)
```
The other
```python
def load_somethingelse():
    return ...
```





```python
$ ipython
IPython 6.1.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: %config TerminalInteractiveShell.ast_node_interactivity='display_last_expr_or_assign'
In [2]: from dummymod import load_cars, load_iris

In [3]: a = load_iris()

=================
 Id    |  types
=================
1      |    A
2      |    B
3      |    A
None

In [4]: b = load_cars()

=================
 Id    |  types
=================
I      |    E
II     |    F
III    |    G
```

With the current implementation that would be disambiguated by having (or not) an Out:

```python
In [6]: a = load_iris()

=================
 Id    |  types
=================
1      |    A
2      |    B
3      |    A

In [7]: b = load_cars()
Out[7]:

=================
 Id    |  types
=================
I      |    E
II     |    F
III    |    G
```

Well now both are implemented so I can push 2 options.

rgbkrk:
Ok I'm totally going to set this in my personal defaults.

Carreau:
Note, the current code will also not work if the LHS is an arbitrary assignment like `a.b = 1`

Carreau:
So `@fperez` agree with the original proposal. I'm going to revert ca387c8

docs/source/whatsnew/pr/interactive_assignment.rst
```diff
@@ -0,0 +1,22 @@
...
+
+    In[2]: xyz = "something else"
+    ...  : xyz
+    "something else"
+
+With the option ``InteractiveShell.ast_node_interactivity='last_expr_or_assign'``
+you can now do::
+
+    In[2]: xyz = "something else"
+    "something else"
```
takluyver: This example (and the second one above) should have an `Out[2]:` prompt, right?

Carreau: Yes ! I `s/>>>/In[1]` stupidly.

Carreau:
Now that 6.1 is released, I'm getting this in to play with it.

PR #10605, opened 2017-05-25
[Question] shadow namespace

Carreau:
We do a lot of dance with user_ns and our injected variable.
In particular _,__,___ and rotate these while still making sure the user
can overwrite and if _ is user defined not push this one to __, but the
actual older value.

It seem to me that ChainMap would be a way cleaner way of doing this,
with a slightly different semantic as you can't delete things from the
shadow namespace directly.

Do anybody see an obvious reason not to use ChainMap ? If so I might
experiment with it and move some logic there. Thoughts ?

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1018,6 +1018,12 @@ class InteractiveShell(SingletonConfigurable):
         # we can list later only variables defined in actual interactive use.
         self.user_ns_hidden = {}
 
+        # This shadow namespace complement user_ns with undeletable variables 
+        # that will be shadowed by user defined ones up until they are deleted.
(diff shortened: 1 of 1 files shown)


takluyver:
Weird things happen when you `exec()` with different globals and locals, because the way Python compiles closures doesn't work nicely with dynamic code execution. In normal usage, `user_ns` and `user_global_ns` are references to the same object, so we avoid a range of issues which we've never managed to fix when `embed()` is called inside a function, resulting in different locals and globals. I'd definitely be concerned that this would introduce such bugs in regular use of IPython.

Carreau:
OK, i'll experiment with this, maybe try to make it an experiment flag and enable it on my system. I'll poke around at embed.

PR #10606, opened 2017-05-25
Support multiline input in --simple-prompt 

### LINKED ISSUE #9816: Add multiline coding in ipython 5 with --simple-prompt
As emacs user I often use emacs repl with **ipython**, but after update to **ipython 5** I need to use `--simple-prompt` flag for my work. But when I try to create new def or class I get this error message

``` python
In [1]: def new():
  File "<ipython-input-1-ce93a60072ff>", line 1
    def new():
              ^
SyntaxError: unexpected EOF while parsing
```

infinity0:
fixes: #9816

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -101,7 +101,7 @@ class TerminalInteractiveShell(InteractiveShell):
     _pt_app = None
 
     simple_prompt = Bool(_use_simple_prompt,
-        help="""Use `raw_input` for the REPL, without completion, multiline input, and prompt colors.
+        help="""Use `raw_input` for the REPL, without completion and prompt colors.
 
             Useful when controlling IPython as a subprocess, and piping STDIN/OUT/ERR. Known usage are:
             IPython own testing machinery, and emacs inferior-shell integration through elpy.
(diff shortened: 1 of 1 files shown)


infinity0:
Actually, ideally I would take the prompt_text and prompt_continuation from the underlying application config - if I change them to `sage: ` and `....: ` respectively, then the cantor sage backend *does* work with sage 7.6 / ipython 5.

infinity0:
Some help fixing the tests would be appreciated, but at least my own cantor-sage-backend works now with this patch applied to IPython (plus a corresponding patch in sagemath to force `simple_prompt` when called by Cantor).

Carreau:
Have you tried to use [rlipython](https://github.com/ipython/rlipython) ? This allow IPython to reuse readline as a frontend, and should and you will not need to wait for a new release of IPython.

Carreau:
So it seem you miss-committed, as the only missing thing was an import of cast_unicode_py2. Which is unnecessary as IPython 6 is Python 3 only, but that will make backport easier.

infinity0:
Oh right, thanks! Yes I had been testing this against 5.1.0 from Debian, and didn't notice the import had been removed from git master.

I'll let the Sage developers know about rlipython. It would be good to have this patch in IPython as well though, so that multiline input works even if not using rlipython. And it's easier for me to backport stuff to Debian after it's been accepted upstream, rather than waiting for Sage to add a new feature.

Carreau:
ok, let's get that one in and backport.

Carreau:
@meeseeksdev backport

infinity0:
Hm, unfortunately this breaks a lot of doctests in Sage. I'll investigate and try to fix...

infinity0:
That is because sage expects the prompt to be "In [0]" but I changed it with this commit... should be quite simple to fix.

infinity0:
OK, confirmed I can un-break the Sage doctests with a 1-line fix, there is no need to further patch IPython.

PR #10614, opened 2017-05-30
Add all the metadata (esp. SVG)

mpacer:
The aim of this PR is to get it so that DisplayObject has a metadata API that can be inherited by it's subclasses.

One of the challenges in designing this is to not destroy those subclasses that had already defined their own metadata API (such as `Image`).  Fortunately, @minrk helped me work through that design challenge (along with @Carreau helping a lot with with debugging whether any of this worked at all). 

Technically, this also adds a way for SVG objects to surface metadata when returned or displayed. If we want this to work more generally it will require adding a function like `_data_and_metadata` to other subclasses. 

### Open question: returning metadata information more generally

Should we try to figure out a way to implement a `_data_and_metadata` method on the base class so that that is just inherited? If we do so that would likely require a more substantial refactoring. Alternatively, we could keep with the current approach. That would entail handling this in a custom fashion for each of the potential Mimetype specific DisplayObject subclasses. 

Which approach would be preferred?

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -574,6 +574,7 @@ class DisplayObject(object):
 
     _read_flags = 'r'
     _show_mem_addr = False
+    metadata = None
 
     def __init__(self, data=None, url=None, filename=None, metadata=None):
         """Create a display object given raw data.
@@ -607,10 +608,11 @@ class DisplayObject(object):
         self.data = data
         self.url = url
         self.filename = filename
+
         if metadata is not None:
-            self.metadata=metadata
-        else:
-            self.metadata={}
+            self.metadata = metadata
+        elif self.metadata is None:
+            self.metadata = {}
 
         self.reload()
         self._check_data()
@@ -725,9 +727,19 @@ class SVG(DisplayObject):
             pass
         svg = cast_unicode(svg)
         self._data = svg
+    
+    def _data_and_metadata(self):
+        """shortcut for returning metadata with shape information, if defined"""
+        md = {}
+        if self.metadata:
+            md.update(self.metadata)
+        if md:
+            return self.data, md
+        else:
+            return self.data
 
     def _repr_svg_(self):
-        return self.data
+        return self._data_and_metadata()
 
 
 class JSON(DisplayObject):
@@ -1071,8 +1083,8 @@ class Image(DisplayObject):
         self.height = height
         self.retina = retina
         self.unconfined = unconfined
-        self.metadata = metadata
-        super(Image, self).__init__(data=data, url=url, filename=filename)
+        super(Image, self).__init__(data=data, url=url, filename=filename, 
+                metadata=metadata)
 
         if retina:
             self._retina_shape()
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -281,6 +281,8 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     raw = kwargs.pop('raw', False)
     if transient is None:
         transient = {}
+    if metadata is None:
+        metadata={}
     if display_id:
         if display_id is True:
             display_id = _new_id()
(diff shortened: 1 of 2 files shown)


IPython/core/display.py
```diff
@@ -460,8 +467,9 @@ class DisplayObject(object):
 
     _read_flags = 'r'
     _show_mem_addr = False
+    metadata = None
 
-    def __init__(self, data=None, url=None, filename=None):
+    def __init__(self, data=None, url=None, filename=None, metadata=None):
```
takluyver: Don't forget to add it to the docstring

IPython/core/display.py
```diff
@@ -607,9 +620,19 @@ def data(self, svg):
             pass
         svg = cast_unicode(svg)
         self._data = svg
+    
+    def _data_and_metadata(self):
+        """shortcut for returning metadata with shape information, if defined"""
+        md = {}
+        if self.metadata:
+            md.update(self.metadata)
```
takluyver: Aren't these lines equivalent to `md = self.metadata.copy()`?

IPython/core/display.py
```diff
@@ -607,9 +620,19 @@ def data(self, svg):
...
+        if md:
+            return self.data, md
+        else:
+            return self.data
 
     def _repr_svg_(self):
-        return self.data
+        return self._data_and_metadata()
```
takluyver: Does returning a 2-tuple from our repr methods actually work? I've never seen this before.

IPython/core/display.py
```diff
@@ -199,6 +202,10 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
         format = InteractiveShell.instance().display_formatter.format
 
     for obj in objs:
+        if isinstance(obj, DisplayObject) and obj.metadata:
+            temp_dict = deepcopy(obj.metadata)
+            temp_dict.update(metadata)
+            metadata.update(temp_dict)
```
takluyver: This seems awkward. I'm guessing it's part of the compatibility thing you mentioned you worked out with Min - if so, can we have a comment explaining why it's necessary?

IPython/core/display.py
minrk: Yes, we do this in Image, matplotlib inline, etc. to set width/height.

IPython/core/display.py
Carreau: Yes that's seem to be the same. I'm also unsure if a helper method to have `_repr_svg_` be one line is necessary.

Carreau:
> Technically, this also adds a way for SVG objects to surface metadata when returned or displayed. If we want this to work more generally it will require adding a function like _data_and_metadata to other subclasses.

Maybe make it a helper function outside of the class to not have to duplicate the code if you decide to keep it ?

Carreau:
I'm supposing this should target 5.4 ?

IPython/core/display.py
mpacer: Not sure what the expository part of the docstring was originally intended to be, but it's now part of the signature markup.

IPython/core/display.py
mpacer: I was copying the idiom in `JSON` and `Image` where the one liner technically accomplishes even less.

mpacer: I think it would be equivalent to deepcopy(self.metadata) not self.metadata.copy() (technically). I'll change that.

takluyver: No, if we're being technical, `.update()` only does a shallow copy :-P

```
 1│ from copy import deepcopy

 2│ a = {'b': []}

 3│ c = {}; c.update(a)

 4│ d = deepcopy(a)

 5│ c, d
 5> ({'b': []}, {'b': []})

 6│ a['b'].append(3)

 7│ c, d
 7> ({'b': [3]}, {'b': []})
```

takluyver: Obviously it's fine for the code to use deepcopy. But 'technically' is just too good a challenge to resist :wink:

IPython/core/display.py
```diff
@@ -617,6 +628,15 @@ def _check_data(self):
         """Override in subclasses if there's something to check."""
         pass
 
+    def _data_and_metadata(self):
+        """shortcut for returning metadata with shape information, if defined"""
+        if self.metadata:
+            md = deepcopy(self.metadata)
+        if md:
+            return self.data, md
```
takluyver: I think the lines above could be condensed into:

```python
if self.metadata:
    return self.data, deepcopy(self.metadata)
```

takluyver:
One minor code style point, otherwise :+1:

PR #10620, opened 2017-05-30
Add code helpers tag to README.rst

mraduldubey:
The label will help beginners to get started with contributing to Ipython: [Ipython@CodeTriage](https://www.codetriage.com/ipython/ipython).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,6 +10,8 @@
 .. image:: https://img.shields.io/travis/ipython/ipython.svg
     :target: https://travis-ci.org/ipython/ipython
 
+.. image:: https://www.codetriage.com/ipython/ipython/badges/users.svg
+    :target: https://www.codetriage.com/ipython/ipython/
 
 ===========================================
  IPython: Productive Interactive Computing


Carreau:
Hum, I'm unsure. I see the use case, but I don't know about how many people are using that and definitively receiving 3 random issues among IPython's one will not be that useful for users. 

We do tag issues for newcomers, and we have often have other blocking things like reviewing etc... 

I'm going to stay neutral and wait for other dev to weigh in.

takluyver:
The idea of dozens of people unfamiliar with the project each getting a random issue emailed to them every day with an invitation to try and fix it does not exactly fill me with joy, as I sit here trying to go through my inbox with ~40 issues updated overnight.

If this service has value to us, I think it would be as a kind of new contributor stress test - forcing us to change the ways we work so that new contributors can get started more easily. 

I'd be concerned about dredging up old issues, if it really is selecting them at random. I think it's not unusual for 'would be nice' issues to linger because there isn't much demand for a feature, and it may be tricky to implement without either breaking something else or making deeper changes to the architecture. If new users try to tackle these, I could see it requiring lots of communication for relatively little tangible benefit.

Carreau:
I tend to agree with @takluyver after this night as I (at least) got some ping from users of CodeTriage in the last couple of hours, that were not really good experience.

So I'm -1 ing today.

mraduldubey:
OK. I get it. You guys should close this thread now.

takluyver:
I'll leave it a couple of days in case other developers want to make the case for doing this - we do want to make the project more welcoming to new contributors, and I feel a bit bad about saying no to something that's trying to help with that, even if I'm not convinced by its approach to the issue.

Carreau:
> I'll leave it a couple of days in case other developers want to make the case for doing this - we do want to make the project more welcoming to new contributors, and I feel a bit bad about saying no to something that's trying to help with that, even if I'm not convinced by its approach to the issue.
> Merge state

Ok, then let's get this in, and remove it if it appear to not be a good experience.

mraduldubey:
@Carreau That is a valid point. :)

Issue #10616, opened 2017-05-30
remove filehash in IPython/utils/path.py


ivanov:
This bug is deliberately intended for a **first-time contributor**. If you're a witch or wizard who is comfortable conjuring up incantations of the serpent constrictor, this is *not* the bug :bug: for you. Don't despair, there are [plenty of others in need of you and your pointy hat](https://github.com/ipython/ipython/issues), work your :sparkles: magic :sparkles: on them. 💫 

Alright, it's just us, now - welcome to the beginning of your IPython adventure!

[Here's a chunk of code](https://github.com/ipython/ipython/blob/master/IPython/utils/path.py#L369,L374) that was deprecated in IPython 4.0. Now that we've released 6.0 (two release cycles), it's fair game to remove it altogether. Be sure to remove the `md5` import itself, as well.

Why do this? Well, one reason to do this is to not use hardcoded `md5` in our codebase, because it can be an issue for OpenSSL in FIPS mode, which will not have `md5`  (see #10615). And in general, this is a good "weeding the garden" practice, removing stale, unused code. 🌱

When making the pull request, be sure to mention this bug :bug: by number in either the commit message or the description, so that it gets cross linked. You can also @ mention me, and I'll help shepherd your contribution into the code base. 

Happy hacking! :bowtie:

PR #10624, opened 2017-05-31
removed filehash in IPython/utils/path.py #10616

adityausathe:
removed filehash method and md5 import from IPython/utils/path.py.
Bug #10616 
@ivanov


ivanov:
awesome, thank's for your contribution and welcome to the wonderful world of IPython and Jupyter contributors @adityausathe !

Issue #10617, opened 2017-05-30
display() instantiates a shell in pure Python


Carreau:
See #10596 conversation near the end.

We should likely fallback to print (with a warning) on pure Python.

mraduldubey:
OK

adityausathe:
Hey, I'm a newbie to IPython. I am trying to work on this issue, but I don't know how to distinguish calls to display() from Pure Python interpreter and IPython interpreter.
@Carreau

mraduldubey:
#10634 
This should fix the issue. I have not issued a warning to users. Just printed the objects.

srinivasreddy:
@takluyver  safely closed?

takluyver:
Yup, thanks

PR #10630, opened 2017-06-01
fixed - display() instantiates a shell in pure Python #10617 

adityausathe:
I kept a variable to keep track whether the Interactive shell have been instantiated. Based on that, we get to know from where display() has been called. In response to a call from Pure Python shell a warning message will be shown.
Currently, I'm working on implementing default print() functionality after displaying warning message. #10617 
@Carreau

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -21,6 +21,8 @@ import warnings
 from IPython.utils.py3compat import cast_unicode
 from IPython.testing.skipdoctest import skip_doctest
 
+from warnings import warn
+
 __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
 'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json',
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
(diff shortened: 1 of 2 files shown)


takluyver:
We've actually already got a way to check this - it's `InteractiveShell.initialized()`.

Can we also ask you to make the changes with your editor configured not to strip trailing whitespace? The changes on lots of empty lines can cause merge conflicts, and they make tools like `git blame` less useful.

IPython/core/display.py
```diff
@@ -278,6 +280,14 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
         from IPython.display import display
 
     """
+    # Check whether display() has been called from IPython shell or from Pure Python shell
+    from IPython.core.interactiveshell import InteractiveShell
```
Carreau: We already import `InteractiveShell `  line 303, id you need it here please also remove the import below.

IPython/core/display.py
```diff
@@ -278,6 +280,14 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
...
+    if not InteractiveShell.is_instantiated:
+        # Warn user who are calling display() from Pure Python Shell
+        #TODO:Add standard print() functionality
+        warn('display() no longer supports calls from Pure Python shell. To experience it\'s features use IPython shell instead.')
```
Carreau: I think we might want to have `print(*objs)` directly, no need to warn I think, but that just my opinion.

PR #10634, opened 2017-06-01
fix of display() instantiates a shell in pure Python #10617

mraduldubey:
Directly fallback on print if InteractiveShell.initialized() is None. We are talking this Issue: #10617.
Though, this means that regardless of data being raw or not, we would be using print() in both cases.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -280,7 +280,7 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     """
     from IPython.core.interactiveshell import InteractiveShell
     
-    if not InteractiveShell.is_instantiated:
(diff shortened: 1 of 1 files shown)


mraduldubey:
Though, this means that even if data is raw or not, we would be using print() itself.

takluyver:
LGTM

Carreau:
Great ! Thanks !

rgbkrk:
Can this one be backported to Python2? I'm running into issues where running `IPython.display.display` is creating an interactive shell.

rgbkrk:
@meeseeksdev backport

rgbkrk:
@meeseeksdev backport to 5.x

Issue #10636, opened 2017-06-02
Need better error message for 'return' keyword in builtin magic functions


adityausathe:
I'm using 'ipython_version': '6.1.0.dev'. Whenever, I use `return` in **inbuilt** `%line_magic` or `%%cell_magic`, I get a clumsy error with stack traceback. We can display a syntax error, instead.
PS: `continue`, `break` work fine with inbuilt magic, though used in absence of loop/conditional statement.
**Code**
```
In [8]: %%timeit -n1
   ...: pass
   ...: return
   ...: 
```
**Output**
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-b0ecc2043232> in <module>()
----> 1 get_ipython().run_cell_magic('timeit', '-n1', 'pass\nreturn')

~/github/ipython_bug2/ipython/IPython/core/interactiveshell.py in run_cell_magic(self, magic_name, line, cell)
   2106             magic_arg_s = self.var_expand(line, stack_depth)
   2107             with self.builtin_trap:
-> 2108                 result = fn(magic_arg_s, cell)
   2109             return result
   2110 

<decorator-gen-61> in timeit(self, line, cell)

~/github/ipython_bug2/ipython/IPython/core/magic.py in <lambda>(f, *a, **k)
    185     # but it's overkill for just that one bit of state.
    186     def magic_deco(arg):
--> 187         call = lambda f, *a, **k: f(*a, **k)
    188 
    189         if callable(arg):

~/github/ipython_bug2/ipython/IPython/core/magics/execution.py in timeit(self, line, cell)
   1083 
   1084         all_runs = timer.repeat(repeat, number)
-> 1085         best = min(all_runs) / number
   1086         worst = max(all_runs) / number
   1087         timeit_result = TimeitResult(number, repeat, best, worst, all_runs, tc, precision)

TypeError: unorderable types: NoneType() < NoneType()
```

takluyver:
Hmm. The way timeit works is that it compiles your code inside a function like this:

```python
def inner(_it, _timer):
    _t0 = _timer()
    for _i in _it:
        # Your code here
    _t1 = _timer()
    return _t1 - _t0
```

Any code that's valid outside a function should work inside it, but we didn't think about people writing code that's only valid inside the function. I think I can see a way to fix that.

takluyver:
#10637 should fix this. The traceback is still a bit longer than it needs to be, but at least you get the right error at the end of it.

[...]

PR #10637, opened 2017-06-02
Give SyntaxError for naked return in timeit

takluyver:
We were missing invalid syntax because we put the code into a function before compiling it, leading to confusing error messages.

Closes gh-10636


Carreau:
@meeseeksdev backport

Carreau:
@meeseeksdev backport

(just testing it will fail but should apply a needs manual backport label)

Issue #10327, opened 2017-02-23
InteractiveShell.magic is deprecated but is what we use in history.


Carreau:
All in title.

srinivasreddy:
I do not understand, and could you please care to explain?

takluyver:
There's a method called `InteractiveShell.magic` that we use somewhere in the history code (not sure where - you could have been more precise, @Carreau ;-). We're now encouraging people to use `InteractiveShell.run_line_magic` instead.

Carreau:
You can see that magic is used in a couple of places (interaciveshell, line 2029) we call `self.magic('color %s` % self.colors)` instead of `self.run_line_magic('color', self.colors)`.

And if you run the hist magic:
```
In [1]: time 1+1
CPU times: user 3 µs, sys: 1 µs, total: 4 µs
Wall time: 5.72 µs
Out[1]: 2

In [2]: hist -t
get_ipython().magic('time 1+1')
get_ipython().magic('hist -t')`
```

You see that what was executed is `get_ipython().magic(...)` instead of `get_ipython().run_line_magic()`

Carreau:
(Note, one possible thing is to un-deprecate magic if it's too painful to update.

srinivasreddy:
I am taking it

adityausathe:
We need to also make changes to a lot of core test cases, if we replace `get_ipython().magic(..)` with `get_ipython().run_line_magic(..)`. We can simply un-deprecate `get_ipython().magic(..)`.

Carreau:
> We need to also make changes to a lot of core test cases, if we replace get_ipython().magic(..) with get_ipython().run_line_magic(..). We can simply un-deprecate get_ipython().magic(..).

Yes, we can also undeprecate `magic()`, in general we should figure out why (and when?) it was deprecated. It is still likely something we want to keep for backward compatibility for a while anyway.

adityausathe:
Please, have a look at PR #10638. I have resolved the issue and also kept backward compatibility support for `get_ipython().magic()`

PR #10638, opened 2017-06-04
fix for #10327 : replacing support for get_ipython().magic()

adityausathe:
There was no direct calls made to `get_ipython().run_line_magic()`, instead we were calling `get_ipython().magic()` first which calls `get_ipython().run_line_magic()` from it's body. That was the reason `hist -t` showing `get_ipython().magic()` instead of `get_ipython().run_line_magic()` because that's what was happening in there.

I replaced calls to `get_ipython().magic()` by `get_ipython().run_line_magic()`. But, there are some tests in `IPython/core/tests/..` which still make calls to `get_ipython().magic()`. I haven't changed them.
#10327

### DIFF
--- a/IPython/core/inputtransformer.py
+++ b/IPython/core/inputtransformer.py
@@ -198,11 +198,14 @@ def _make_help_call(target, esc, lspace, next_input=None):
                 else 'psearch' if '*' in target \
                 else 'pinfo'
     arg = " ".join([method, target])
+    #Prepare arguments for get_ipython().run_line_magic(magic_name, magic_args)
+    t_magic_name, _, t_magic_arg_s = arg.partition(' ')
+    t_magic_name = t_magic_name.lstrip(ESC_MAGIC)
     if next_input is None:
-        return '%sget_ipython().magic(%r)' % (lspace, arg)
+        return '%sget_ipython().run_line_magic(%r, %r)' % (lspace, t_magic_name, t_magic_arg_s)
     else:
-        return '%sget_ipython().set_next_input(%r);get_ipython().magic(%r)' % \
-           (lspace, next_input, arg)
+        return '%sget_ipython().set_next_input(%r);get_ipython().run_line_magic(%r, %r)' % \
+           (lspace, next_input, t_magic_name, t_magic_arg_s)
     
 # These define the transformations for the different escape characters.
 def _tr_system(line_info):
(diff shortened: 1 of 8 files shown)


adityausathe:
It seems that the tests should also need to be changed in order to accommodate the changes. The checks are failing at those points. Are we changing the test cases?
@Carreau

Carreau:
Hi there, thanks you probably want to synchronize with #10612 which is also trying to update this. 
Yes the test need to be updated, but we need to make sure `get_ipython().magic()` still works at least until it is removed.

adityausathe:
Summary: 
- Replaced direct calls to `get_ipython().magic()` with `get_ipython().run_line_magic()`
- Updated test cases to accommodate the change.
- Added backward compatibility support for `get_ipython().magic()`, by updating `stack_depth`

@Carreau 
@takluyver

IPython/core/inputtransformer.py
```diff
@@ -198,11 +198,14 @@ def _make_help_call(target, esc, lspace, next_input=None):
                 else 'psearch' if '*' in target \
                 else 'pinfo'
     arg = " ".join([method, target])
```
srinivasreddy: this line can be removed.

srinivasreddy (approved):
LGTM. thanks.

IPython/core/interactiveshell.py
```diff
@@ -2044,7 +2044,7 @@ def register_magic_function(self, func, magic_kind='line', magic_name=None):
         self.magics_manager.register_function(func, 
                                   magic_kind=magic_kind, magic_name=magic_name)
 
-    def run_line_magic(self, magic_name, line):
+    def run_line_magic(self, magic_name, line, back_compat=False):
```
takluyver: Maybe make the parameter `_stack_depth=1` rather than `back_compat=False` - that's a bit clearer about what it means. The leading underscore would suggest that it's not for normal use.

IPython/core/tests/test_inputtransformer.py
```diff
@@ -57,13 +57,13 @@ def transform_checker(tests, transformer, **kwargs):
 
        assign_magic =
        [(i,py3compat.u_format(o)) for i,o in \
-       [(u'a =% who', "a = get_ipython().magic({u}'who')"),
-        (u'b = %who', "b = get_ipython().magic({u}'who')"),
-        (u'c= %ls', "c = get_ipython().magic({u}'ls')"),
+       [(u'a =% who', "a = get_ipython().run_line_magic({u}'who', {u}'')"),
```
takluyver: So long as we're targeting IPython 6.x, we can get rid of the `{u}` things in the tests - that's for Python 2 compatibility, which we've dropped now.

adityausathe:
@takluyver
Thanks for the feedback. 
Incorporated suggested changes.

takluyver:
Thanks

PR #10644, opened 2017-06-07
system-wide startup files and env config

minrk:
- adds `{sys.prefix}/etc/ipython` to config-file path, which is commonly used for Jupyter applications in envs
- support startup files in system-wide config directories (e.g. `/etc/ipython/startup`, now `{sys.prefix}/etc/ipython/startup`)

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -43,6 +43,14 @@ else:
         "/etc/ipython",
     ]
 
+
+ENV_CONFIG_DIRS = []
+_env_config_dir = os.path.join(sys.prefix, 'etc', 'ipython')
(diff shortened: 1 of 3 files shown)


Carreau:
Do you wish a backport to 5.x ?

Carreau (approved):

minrk:
Either way. It's a new feature, so we don't have to, but it's also small and simple, so shouldn't be a problem. I guess I'd be for backport unless someone thinks we shouldn't.

takluyver (approved):

Carreau:
@meeseeksdev backport

PR #10651, opened 2017-06-09
Safe method to get the repr of an object.

Carreau:
- Protect against recursion.
 - <del>Incorrect because of bugs in IPython, in particular the priority of
   formatters is not well respected.</del>

First step toward #10647

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -18,6 +18,7 @@ import struct
 import sys
 import warnings
 from copy import deepcopy
+from collections import namedtuple
 
 from IPython.utils.py3compat import cast_unicode
 from IPython.testing.skipdoctest import skip_doctest
@@ -150,24 +151,19 @@ class RecursiveObject:
     """
 
     def __init__(self, already_seen):
+        self.seen = already_seen
         pass
 
     def __repr__(self):
-        return '<recursion ...>'
+        return '<recursion ... {}>'.format(str(self.seen))
 
     def _repr_html_(self):
-        return '&lt;recursion ...&gt;'
+        import html
+        return '&lt;recursion ... {}&gt;'.format(html.escape(str(self.seen)))
 
 
-class DataMetadata(tuple):
 
-    @property
-    def data(self):
-        return self[0]
-
-    @property
-    def metadata(self):
-        return  self[1]
+DataMetadata = namedtuple('DataMetadata', ('data','metadata'))
 
 
 class ReprGetter:
@@ -226,27 +222,40 @@ class ReprGetter:
 
         When :any:`get_repr_mimebundle` detect it is recursively called, it will
         attempt to return the representation of :class:`RecursiveObject`. You
-        may register extra formatter for :class:`RecursiveObject`
+        may register extra formatter for :class:`RecursiveObject`.
 
         If you are computing objects representation in a concurrent way (thread,
         coroutines, ...), you should make sure to instanciate a
         :class:`ReprGetter` object and use one per task to avoid race conditions.
+
+        If a specific mimetype formatter need to call `get_repr_mimebundle()`
+        for another mimeformat, then it must pass the mimetypes values it desire
+        to `include` in order to correctly avoid recursion  
 
         See Also
         --------
             :func:`display`, :any:`DisplayFormatter.format`
 
         """
+        from there import print
         from IPython.core.interactiveshell import InteractiveShell
+        if isinstance(include, str):
+            include = (include,)
+        if not include:
+            keys = {(id(obj), None)}
+        else:
+            keys = {(id(obj), f) for f in include}
         fmt = InteractiveShell.instance().display_formatter.format
-        if id(obj) in self._objs:
-            return DataMetadata(fmt(on_recursion(obj), include=include, exclude=exclude))
+        if id(obj) == id(object):
+            return DataMetadata({'text/plain':"<class 'object'>"}, {})
+        if self._objs.intersection(keys):
+            return DataMetadata(*fmt(on_recursion(obj), include=include, exclude=exclude))
         else:
             try:
-                self._objs.add(id(obj))
-                return DataMetadata(fmt(obj, include=include, exclude=exclude))
+                self._objs.update(keys)
+                return DataMetadata(*fmt(obj, include=include, exclude=exclude))
             finally:
-                self._objs.discard(id(obj))
+                self._objs.difference_update(keys)
 
 # Expose this for convenience at the top level. Similar to what the random
 # module in python does. If you want to avoid weird behavior from concurrency:
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -28,8 +28,8 @@ __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
 'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'JSON', 'GeoJSON', 'Javascript',
 'Image', 'clear_output', 'set_matplotlib_formats', 'set_matplotlib_close',
-'publish_display_data', 'update_display', 'DisplayHandle', 'Video', 'get_repr_mimebundle',
-'RecursiveObject', 'ReprGetter']
+'publish_display_data', 'update_display', 'DisplayHandle', 'Video', 'get_repr_mimebundle'
+]
 
 #-----------------------------------------------------------------------------
 # utility functions
@@ -237,7 +237,6 @@ class ReprGetter:
             :func:`display`, :any:`DisplayFormatter.format`
 
         """
-        from there import print
         from IPython.core.interactiveshell import InteractiveShell
         if isinstance(include, str):
             include = (include,)
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -27,7 +27,8 @@ __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
 'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'JSON', 'GeoJSON', 'Javascript',
 'Image', 'clear_output', 'set_matplotlib_formats', 'set_matplotlib_close',
-'publish_display_data', 'update_display', 'DisplayHandle', 'Video']
+'publish_display_data', 'update_display', 'DisplayHandle', 'Video', 'get_repr_mimebundle',
+'RecursiveObject', 'ReprGetter']
 
 #-----------------------------------------------------------------------------
 # utility functions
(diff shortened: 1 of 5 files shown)


IPython/core/display.py
```diff
@@ -140,6 +141,88 @@ def _new_id():
...
+        return '&lt;recursion ...&gt;'
+
+
+DataMetadata = namedtuple('DataMetadata', ['data', 'metadata'])
+
+_objs = set()
+
+
+
+def get_repr(obj, include=None, exclude=None, *, on_recursion=_RecursiveObject):
```
takluyver: Naming: should this be something like `get_repr_mimebundle()` or `get_rich_reprs()`? It seems like a good idea to clearly distinguish our rich repr system from Python's text-based reprs.

IPython/core/display.py
```diff
@@ -140,6 +141,88 @@ def _new_id():
...
+        data dict. If this is set all format types will be computed,
+        except for those included in this argument.
+    on_recursion: callable
+        Return an object to compute the representation when recursion is
+        detected.
+
+
+    Returns
+    -------
+    (format_dict, metadata_dict) : named tuple of two dicts
```
takluyver: I think the names here should match what they are in the named tuple, i.e. `(data, metadata)`, rather than involving a different pair of names for them.

IPython/core/display.py
```diff
@@ -140,6 +141,88 @@ def _new_id():
...
+
+        0/data: format_dict is a dictionary of key/value pairs, one of each
+        format that was generated for the object. The keys are the format
+        types, which will usually be MIME type strings and the values and
```
takluyver: > usually be MIME type strings

Are there circumstances in which they're not? If so, we should make at least a brief mention of what they are. If we just mean that they might be unofficial mime types, I think that's OK - unregistered mime types beginning with either `x-` or `vnd.` are common enough that they don't need special mention.

Should also mention:

- For most mime types (other than `application/*json`), values will be strings.
- Binary data will be base64 encoded strings (assuming this handles that step), and it's up to calling code to know which are binary. [aside: should we standardise a field in metadata to indicate which mime types are base64 encoded?]

IPython/core/display.py
```diff
@@ -140,6 +141,88 @@ def _new_id():
...
+        format_dict.
+
+    Note
+    ----
+
+    When `get_repr` detect it is recursively called, it will attempt to return
+    the representation of `_RecursiveObject`
+
+    """
+    global _objs
```
takluyver: I'm not keen on the use of a global store here - you could get weird results if there's more than one thread or coroutine using this function at once. I'll try to think of a better way to handle the recursion thing.

IPython/core/display.py
Carreau: that's copy/pasted from the formatter module, I'm guessing we should either synchronize, of get the information in a single place.

IPython/core/display.py
Carreau: I don't see any way that does not require a global-ish state, I can move it into a class (that would be better).
But we can't really enforce users to pass variable down. 

I'm still expecting this to almost never go on and recurse. 
If you are concern about threading I can use  `threading.local()`. Coroutines: there is no tasklocals in Python (yet?) and having a task-switch would require having coroutiness based reprs, which I'm unsure we do support. 

Worse case is I believe we will go down the recursion path where there is no recursions, (AFAICAT), and I think it is a reasonable tradeoff.

Carreau:
Refactored, cleaned and documented with some tests. Still some polish to do. 
I moved the "get_repr_mimebundle" into a class method with "use your own instance of the class if you have any chance of having concurrency", Of course you can shoot your self in the foot, by creating a instance everytime you compute the repr, but he, you can shoot yourself in the foot in so many different way, and you can also call the formatters yourself.

I now also "Fake" the `__module__` object of everything in IPython.core.display as otherwise Sphinx can't cross link (I'm partially unhappy because it changes the `repr`.)

The DataMetadata now have `.data` and `.metadata` attributes instead of being a named tuple.

The `IPython.display` module have some top-level docs.

IPython/core/display.py
```diff
@@ -139,6 +141,117 @@ def _new_id():
...
+
+    def _repr_html_(self):
+        return '&lt;recursion ...&gt;'
+
+
+class DataMetadata(tuple):
+
+    @property
+    def data(self):
+        return self[0]
```
takluyver: Sorry, I didn't follow why this is preferable to a NamedTuple?

IPython/core/display.py
```diff
@@ -1338,3 +1448,15 @@ def set_matplotlib_close(close=True):
...
+
+
+
+for oname in __all__:
+    # set the module name to IPython.display or sphinx cannot correctly cross
+    # link. Mostly it figures out that objets are in IPython.core.display, but
+    # that's a module we explicitly do not document.
+    try:
+        # using hasattr to check still raises an AttributeError.
+        locals()[oname].__module__ = 'IPython.display'
```
takluyver: Eugh. Maybe we can figure out improvements to Sphinx...

IPython/core/display.py
Carreau: I had issue with namedtuple, where it keep saying in soma place the `.data` was not defined but `[0]` was working. I'm confused as to exactly why but I can retry.

IPython/core/display.py
Carreau: Or, we re-allow `IPython.core.display` to be documented. Right now sphinx can't link because there is no `IPython.core.display` page.

takluyver: What happens if we cross reference to e.g. 

    :func:`IPython.display.display`

Carreau:
Not sure, but that won't help with the base class that are automatically
inferred.

Carreau:
TODO: Look at wether a custom formatter can say "I actually don't know how to represent this".

Carreau:
meh, I'm un-happy with the actual API when using it. I keep using `get_repr_mimebundle(obj).data.get('text/html', None) or escape(repr(obj))` to build my own nested repr. I defined a helper function, but that seem less than optimal. Though it allows to do some nice things. I'm going to think a bit about the API.

IPython/core/display.py
```diff
@@ -139,6 +141,117 @@ def _new_id():
...
+            try:
+                self._objs.add(id(obj))
+                return DataMetadata(fmt(obj, include=include, exclude=exclude))
+            finally:
+                self._objs.discard(id(obj))
+
+# Expose this for convenience at the top level. Similar to what the random
+# module in python does. If you want to avoid weird behavior from concurrency:
+#   Instantiate your own.
+get_repr_mimebundle = ReprGetter().get_repr_mimebundle
```
minrk: If `get_repr_mimebundle` instantiated an internal ReprGetter at call time rather than being a method on a global object, there shouldn't be a potential for race conditions.

IPython/core/display.py
```diff
@@ -27,7 +27,8 @@
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
 'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'JSON', 'GeoJSON', 'Javascript',
 'Image', 'clear_output', 'set_matplotlib_formats', 'set_matplotlib_close',
-'publish_display_data', 'update_display', 'DisplayHandle', 'Video']
+'publish_display_data', 'update_display', 'DisplayHandle', 'Video', 'get_repr_mimebundle',
+'RecursiveObject', 'ReprGetter']
```
minrk: RecursiveObject and ReprGetter should probably *not* be added to `__all__`, since they are private, internal implementation details.

IPython/core/formatters.py
```diff
@@ -139,6 +139,9 @@ def format(self, obj, include=None, exclude=None):
             precedence and the corresponding `_repr_*_` for this mimetype will
             not be called.
 
+            TODO: Figure out and document whether user registered formatters do
```
Carreau: Yes, in theory it seem to not be the case with `_repr_mimebundle_`, I should have been clearer in the comment.

IPython/core/display.py
Carreau: but that would kill the detection of recursion, as every call would create a new repr_getter that would track independently which objects have been seen.

Carreau:
Note to self, the recursion stuff also does not work if a formatter need another. Like for example, the HTML, being "just" a bold version of the text/plain.

Carreau:
> Note to self, the recursion stuff also does not work if a formatter need another. Like for example, the HTML, being "just" a bold version of the text/plain.

Ok, I got that to work.

IPython/core/display.py
Carreau: Fair enough for it to be clean, but I still expect people to need to register formatter for `RecursiveObject` if they want nice representation for format it does not support currently, so you need a handle on it.  ReprGetter also need to be public if you are computing recursive Repr on multiple task/threads

Carreau: removed.

IPython/core/display.py
minrk: It shouldn't. The recursion would be tracked for the duration of the recursion, and the instance disposed of when the outermost call finishes.

Carreau: I'm confused. 


Like so ?
```python
def get_repr_mimebundle(*args, **kwargs):
      return ReprGetter.instance().get_repr_mimebundle(*args, **kwargs)
```
?

I don't see how that prevent race condition. 

If it unconditionally create an instance, and an object have:

```python
from IPython.display import get_repr_mimebundle
class MyContainer:

    def _repr_html_(self):
          return '|'.join(get_repr_mimebundle(obj)['text/html'] for obj in self.objects)
```
Then in breaks recursion detection, and is not thread/task safe as you get the global get_repr_mimebundle anyway. 

Or I'm missing something.

IPython/core/display.py
minrk: Functions defined in one module can be referenced in any module in which they are importable, we just have to make sure that they are present in the autodoc module for the desired target.

I do this [in pyzmq](https://github.com/zeromq/pyzmq/blob/master/docs/source/api/zmq.rst), where docs treat everything as being in top-level `zmq` module, but classes are actually defined all over the place.

IPython/core/display.py
minrk: Another option is to use threadlocal storage for the global. Then there is no way to have a race condition because all of the cooperative concurrency libraries (asyncio, etc.) can't have a race because this is a blocking call.

IPython/core/display.py
```diff
@@ -139,6 +142,124 @@ def _new_id():
...
+    def __repr__(self):
+        return '<recursion ... {}>'.format(str(self.seen))
+
+    def _repr_html_(self):
+        import html
+        return '&lt;recursion ... {}&gt;'.format(html.escape(str(self.seen)))
+
+
+
+DataMetadata = namedtuple('DataMetadata', ('data','metadata'))
```
minrk: You were asking about the issue of conveniently retrieving mime-keys with the `('data', 'metadata')` tuple. One option is a method on this object:

```python
def get_mime(self, mimetype):
    """Extract a single mime-type representation from a mimebundle"""
    if mimetype not in self.data:
        return (None, None)
    else:
        return self.data[mimetype], self.metadata.get(mimetype)
```

This could also be wrapped in a single function:

```python
def get_repr_mime(obj, mimetype):
    """Get the representation of an object as a single mime-type"""
    return get_repr_mimebundle(obj).get_mime(mimetype)
```

A `metadata=False` flag could also prevent it from returning the metadata part, if you want to avoid getting the first item for the data part.

Issue #10556, opened 2017-05-12
%%timeit could use local scope.


Carreau:
You can't use timeit in a function (which is useful to test variable  inputs of several function doing the same things, and plot graphs.

I believe a well placed `@needs_local_scope` and modification to timeit should allow that.

adityausathe:
In current scenario, the `inputSplitter` and `inputTransformer` treat `%%timeit ... ` as line-magic and replaces it with a call to`%sget_ipython().magic(%r)`  function,  which just ignores the line. Do we need to distinguish local scoped `%%timeit... ` while parsing input or is there any other way?

takluyver:
I'm not sure what you mean about the parsing - `%%timeit` definitely works.

You can't have a cell magic in a local scope because it has to be a whole cell - only the line magic form can be in a local scope.

adityausathe:
I was working on it and I was able to get it work(ran some basic tests). Only problem was, I had to enter two blank lines to complete the cell.
My implementation considers all the remaining lines in the cell starting from `%%timeit`, as magic body. Please, tell me if I'm doing wrong here.

takluyver:
> Only problem was, I had to enter two blank lines to complete the cell.
> My implementation considers all the remaining lines in the cell starting from %%timeit, as magic body. 

I think that's normal for a cell magic. Is it any different from `%%timeit` in master?

Carreau:
I was likely refering to the line magic `%timeit` not the cell magic. I tried to make a %timeit, in a function in a for loop. For example:

```
def fib(n):
    return 1 if n < 2 else fib(n-1)+fib(n-2)

def timefib(P):
    t = []
    for i in range(P):
        res = %timeit -o fib(i)
        t.append(res)
    return t
```

And this timeit would need local scope.

I believe that what I was referring to and  inadvertantly have put 2 % sign.

adityausathe:
ooh. :disappointed:

adityausathe:
I worked on it and was able to narrow down the needed changes. We need to pass `local_ns` to `exec(code, self.shell.user_ns, local_ns)` call. But, it was still giving `NameError`. I think that is due to custom `inner` function
```
timeit_ast_template = ast.parse('def inner(_it, _timer):\n'
                                        '    setup\n'
                                        '    _t0 = _timer()\n'
                                        '    for _i in _it:\n'
                                        '        stmt\n'
                                        '    _t1 = _timer()\n'
                                        '    return _t1 - _t0\n')
```
[...]

PR #10655, opened 2017-06-10
Feature req #10556: %timeit could use local scope

adityausathe:
Enabled `@needs_local_scope` for `timeit`. Merged `local_ns` and `global_ns` and passed to `exec` as a `global_ns`. Also, handled conflicting namespace variables.
#10556 
Can we go with this approach?
@takluyver 
@Carreau

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -940,7 +940,8 @@ python-profiler package from non-free.""")
 
     @skip_doctest
     @line_cell_magic
-    def timeit(self, line='', cell=None):
+    @needs_local_scope
+    def timeit(self, line='', cell=None, local_ns=None):
         """Time execution of a Python statement or expression
 
         Usage, in line mode:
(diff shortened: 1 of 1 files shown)


IPython/core/magics/execution.py
```diff
@@ -1074,7 +1075,16 @@ def timeit(self, line='', cell=None):
...
+        glob = self.shell.user_ns
+        # handles global vars with same name as local vars. We store them in conflict_globs.
+        if local_ns is not None:
+            conflict_globs = {}
+            for var_name, var_val in glob.items():
+                if var_name in local_ns:
+                    conflict_globs[var_name] = var_val
+            glob.update(local_ns)
+            
+        exec(code, glob, ns)
```
Carreau: how would this work if we were to chose a way more unlikely name for `inner`  and pass the value of `ns.merge(local_ns)` instead of `ns` ?

Carreau:
Really good work, thanks a lot for that. I haven't touch this code in a while, and we may be able to use `ns`, but if we can't +1 for this.

IPython/core/magics/execution.py
adityausathe: I tried using `ns.merge(local_ns)`, but again it leads to same problem. The vars in `ns` are not in `locals()` inside `inner()`, but they are in local namespace, if we access them outside `inner`. Hence, same problem persists. 
If we somehow get rid of that extra stack_frame due to `inner` then we can retain our standard approach.

takluyver: Yeah, the combination of scoping and dynamic execution in Python is a pain, unfortunately.

Carreau:
Thanks ! That was really good and impressive dive in not-so easy part of IPython. If you like you can make a separate PR that drops a `rst` file in `docs/source/whatsnew/pr/`, which will be merged into the what's new notes at release time.

Issue #10405, opened 2017-03-23
ENTER within brackets of one-liner gives newline instead of executing


mspacek:
This could very well be due to the move away from readline, but I couldn't quite find an existing open issue on this, only an indirect mention of it [here](https://github.com/ipython/ipython/issues/2493#issuecomment-264454032):

I've noticed a change of behaviour of the ENTER key in recent versions of terminal IPython. It used to be when I hit ENTER at any point on a line, the line would be executed. Now, if the cursor falls within some kind of brackets (parentheses, curly, or square), ENTER inserts a new line.
```
In [1]: a = (1 + 1)
# highlight the + sign with the cursor and hit ENTER:
In [2]: a = (1     
   ...: + 1)
```
This behaviour only really depends on the presence of the first bracket, not the second. More generally, when the cursor is to the right of an open bracket, but not at the end of the one-liner, the only way to make it execute instead of inserting a newline is to hit ALT+ENTER.

For one-liners, I'd prefer the opposite key mapping: ALT+ENTER inserts a newline, while ENTER executes regardless of cursor position (except maybe when the line ends with a colon and the cursor is at the end of the line).
```
{'commit_hash': '5a894b9',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/usr/local/lib/python3.5/dist-packages/IPython',
 'ipython_version': '5.3.0',
 'os_name': 'posix',
 'platform': 'Linux-4.4.0-64-generic-x86_64-with-Ubuntu-16.04-xenial',
 'sys_executable': '/usr/bin/python3',
 'sys_platform': 'linux',
 'sys_version': '3.5.2 (default, Nov 17 2016, 17:05:23) \n[GCC 5.4.0 20160609]'}

Name: prompt-toolkit
Version: 1.0.13
```

takluyver:
I'd like to keep the shortcuts consistent (enter to guess what to do, alt-enter to force execution - unfortunately we can't detect Ctrl-enter), but if you want to customise what they do yourself, see here:
http://ipython.readthedocs.io/en/stable/config/details.html#keyboard-shortcuts

Unfortunately it requires a bit of digging around in prompt_toolkit to find the functions you need.

mspacek:
Thanks, though I'm not quite sure what you mean by "enter to guess what to do"...

Hmm. So this is considered a feature, not a bug? Is there a discussion somewhere that justifies the change in behaviour from previous versions of IPython?

[...]

PR #10663, opened 2017-06-13
Improved example code in handle_return docs.

carlsmith:
This just improves the example code in the docs for the new `handle_return` feature.

The old example code only showed how to execute the input or insert a newline, while this new example shows how to execute the input or input a newline *with auto-indentation*.

Most users only want to change *when* IPython executes or inserts a newline, not *how* it does it, so they will normally want IPython to automatically indent their code still.

The new example has a comment that shows how to *only* insert a newline, and from there, it's obvious how to insert arbitrary text, so that stuff is still explained as well.

This example also seems a bit more realistic, showing how to auto-execute when the input begins with a bang or modulo character, rather than when the current line doesn't end with a bar.

This PR was recommended while discussing #10405.


takluyver:
Thanks :-)

carlsmith:
No worries.

PR #10670, opened 2017-06-22
Update tutorial.rst

The-Judge:
Removed the underscore from ``data = ['Number of users', 123_456]``, since this leads to an syntax error.


Carreau:
Sorry about that, `_` has become valid in  literals with Python 3.6, but I agree we should keep tutorial valid for earlier version of Python. Thanks.

PR #10676, opened 2017-06-26
avoid using private traitlets APIs in load_default_config

minrk:
private APIs can change, and this one may in traitlets 5, which could break IPython.embed.

Probably for backport to 5.x


Carreau:
@meeseeksdev baclport

Carreau:
@meeseeksdev backport

Issue #10406, opened 2017-03-24
In terminal, stdout output of background thread is truncated, not wrapped.


zpincus:
In previous versions of IPython and in the standard python interpreter running in a terminal window, text sent to stdout from a background thread will be wrapped to the width of the terminal. Example:
```python
import threading
def p():
    print(''.join(map(str, range(120))))
t = threading.Timer(0.1, p)
t.start()
```

However, the above code in IPython 5 generates text that runs off the edge of the window and is simply truncated. (Resizing the window does not make more text appear.)

I presume this is due to the mode that prompt_toolkit puts the terminal window into. Among other more minor annoyances, this makes debugging crashes in background threads (or in a GUI event loop running with IPython integration) difficult as stack traces usually get cut off.

No idea if this can be fixed, or if it's unavoidable. @jonathanslenders, is this a prompt_toolkit limitation, or is the issue somewhere else?

takluyver:
Possibly we should use `patch_stdout_context` for the whole time that the prompt_toolkit interface runs. I don't know if that has any adverse effects, though.

jonathanslenders:
Hi, this is indeed a consequence of the terminal flags that are set by prompt_toolkit.

Using ``patch_stdout_context`` does indeed solve this, but it's important to set this as soon as possible, because this replaces ``sys.stdout`` by a proxy object. It needs to be set as early as possible, because every import of ``sys`` that was done before patching stdout will still refer to the original object. The disadvantage is of course the monkey patch. ``sys.stdout`` won't refer to the real output.


A better, but much more complex solution I can think of is by using ``os.openpty()`` to create a new pty, then use ``os.dup2`` and fiddle with the file descriptors to make sure that ``prompt_toolkit`` writes to the original stdout, while all stdout objects will point to a proxy object while the interface is visible. This would need a reader for the PTY master FD in the event loop. It is a solution that would work on Posix systems. (Mac+Linux.) But this is something I prefer to implement in prompt_toolkit itself.

Can somebody please create a feature request in prompt_toolkit for this? I think I can include this at least in the 2.0 branch.

ngg:
Hi! I have this same issue, and I tried to use `patch_stdout_context` but it doesn't seem to work for me.

I'm using IPython 5.4.1 (python2.7), I have inserted it into the `prompt_for_code` function in `TerminalInteractiveShell`. It definitely helps, now the prompt is working great when there is output from background threads. However the lines are still truncated.

Can you help me find a workaround for this?

[...]

PR #10686, opened 2017-07-01
Patch stdout in interactive shell

ngg:
If a background thread prints text while the prompt is active then the text would overlap with the prompt which makes it kind of unusable. `patch_stdout_context` solves this issue, all texts are now printed above the prompt and completers remain perfectly usable. Unfortunately this does not solve issue #10406.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -1,5 +1,6 @@
 """IPython terminal interface using prompt_toolkit"""
 
+from contextlib import contextmanager
 import os
 import sys
 import warnings
@@ -236,6 +237,11 @@ class TerminalInteractiveShell(InteractiveShell):
                     prompt_text = prompt_continuation
                 return isp.source_reset()
             self.prompt_for_code = prompt
+
+            @contextmanager
+            def dummy_context(*args, **kwargs):
+                yield
+            self.patch_stdout_context = dummy_context
             return
 
         # Set up keyboard shortcuts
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @ngg,

would it be possible for you to try this branch: https://github.com/jonathanslenders/python-prompt-toolkit/pull/520
patch_stdout is not required and can be disabled, but this should make sure that line wrapping is not disabled for background threads.

ngg:
@jonathanslenders: I've tried your branch jonathanslenders/python-prompt-toolkit#520, and it seems to solve the linewrap issue for me (#10406).
Without `patch_stdout_context` the background prints are wrapped correctly, but they screw up the prompt and completers.
With both our patches, it seems to work perfectly, the lines are wrapped correctly, and the prompt remains intact.

jonathanslenders:
Even with readline (like before), the prompt will be screwed up.
Patching stdout could potentially break other libraries (I'm not sure). What if a certain library has a ``if sys.__stdout__ == sys.stdout`` test? I'm not sure it's worth doing it. It's a risk. Just press Control-L if a background thread printed to stdout. But I leave this up to the IPython developers to decide.

ngg:
I think this can be a really useful feature, but it might have side-effects. If you think it doesn't worth it, then probably we can hide this behind an optional flag. Or probably checking if stdout is modified can work too.

ngg:
https://github.com/jonathanslenders/python-prompt-toolkit/issues/507 is kind of a blocker issue. Its fix was already merged but not yet released.

ngg:
Actually this has lots of unwanted sideeffects I do not reallly understand, for example when there are lots of background prints then `self.pt_cli.run()` sometimes returns with None and whole ipython dies. Something like this would be really good to have, but not solution.

PR #10699, opened 2017-07-13
fix: Clear breakpoints before running any script with debugger.

### LINKED ISSUE #10565: IPython 5.3.0 debug mode breakpoint issue
Hi IPython Team,

I recently upgraded Python to version 2.7.8 (from 2.6.6) and IPython to version 5.3.0.  I noticed that when using the magic command (%run -d) in IPython, subsequent runs of a program seem to fail with a "KeyError", apparently due to the debugger's failure to clear breakpoints from previous runs of the program.

I attached a sample of output from one such failed run (see below), where a breakpoint that had been set at line 56 in a previous call caused the program to crash, though in the current call, only one breakpoint at line 70 is requested.

For the time being, I've downgraded to IPython version 4.2.0, which seems to work as expected.

Cheers,
Taylor

[ipython_error.txt](https://github.com/ipython/ipython/files/1004089/ipython_error.txt)

kgritesh:
Fixes #10565

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -849,6 +849,7 @@ python-profiler package from non-free.""")
         bdb.Breakpoint.next = 1
         bdb.Breakpoint.bplist = {}
         bdb.Breakpoint.bpbynumber = [None]
+        deb.clear_all_breaks()
         if bp_line is not None:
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks!

@meeseeksdev backport

Issue #10469, opened 2017-04-20
Jedi inspected functions cannot be queried for docstring


stefanv:
```
In [1]: d = {'f': 'asd'}

In [2]: d['f'].capitalize?
Object `capitalize` not found.
```

Carreau:
:-) that's unrelated to jedi, it just never worked. But I can see why someone would expect it to work now.

takluyver:
Closing for now, because I don't see a way to do this that's not essentially reimplementing a parser badly. See #10703 for more discussion.

PR #10703, opened 2017-07-16
fix: Allow pinfo magic command to work for dictionary access.

kgritesh:
Closes: #10469

### DIFF
--- a/IPython/core/inputtransformer.py
+++ b/IPython/core/inputtransformer.py
@@ -277,7 +277,8 @@ _initial_space_re = re.compile(r'\s*')
 
 _help_end_re = re.compile(r"""(%{0,2}
                               [a-zA-Z_*][\w*]*        # Variable name
-                              (\.[a-zA-Z_*][\w*]*)*   # .etc.etc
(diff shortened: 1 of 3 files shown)


takluyver:
Hi, sorry it's taken us so long to look at this!

takluyver:
I'm wary of this, not because I think it's wrong, but because it's extra complexity in part of the code that I suspect is already a bit wobbly.

I'm also not sure where it should stop. If we can get dictionary keys `d['foo']`, why not list elements `d[0]`? Literals `[1, 2].index`? Calling builtin methods `s.lower().count`? Builtin functions `str(x).lower`? This is one step towards a Python parser, and I don't think we want to build a Python parser.

kgritesh:
@takluyver i think i get your concern why it may not be a good idea  to have this as its kind of a rabbit hole. In that case, it would make sense then to close the issue and PR right? Or you think there can be a better way to solve this issue?

takluyver:
Yup, sorry, I am going to close it. I can't think of any way to do what you want without going down the parser rabbit hole.

Issue #10702, opened 2017-07-14
Post execute event handlers do not have access to exceptions for failed cells


drdarshan:
If you write a post-execute event handler:
```py
class MyEventHandler(object):
    def __init__(self, ip):
        self.shell = ip
       
    def post_execute(self):
        ec = self.shell.execution_count
        input = self.shell.user_ns["In"][ec]
        output = self.shell.user_ns["Out"].get(ec, None)
        succeeded = self.shell.last_execution_succeeded
        ## stack_trace = ???
        
def load_ipython_extension(ip):
    eh = MyEventHandler(ip)
    ip.events.register("post_execute", eh.post_execute)
```

The handler can get access to the cell that was just run and the outputs if the run succeeded. However, if the run failed, there is no way for the handler to get hold of the stack trace that was produced. It would be ideal if the handler was passed in the result of ``run_cell`` or if there was a way to get hold of the result from the history manager.

takluyver:
I think the easiest way to do this would be to attach the `ExecutionResult` object to `shell` before we call the hooks. Changing what's passed to a callback is tricky, unfortunately.

PR #10704, opened 2017-07-17
Store last execution result in the interactive shell instance

drdarshan:
## Overview
Please see  #10702 for more context on this PR. 

The shell contains a ``last_execution_succeeded`` field which (for instances) lets event handlers know if a cell execution succeeded or not. 

However, if the execution failed, there is no way for an event handler to get to the error itself.  

This PR adds a ``last_execution_result`` field to the shell that stores a reference to the ``ExecutionResult`` object resulting from the evaluation of the last cell. If the execution failed, there is a way to get to the error using ``shell.last_execution_result.error_in_exec``.

## Validation
I added a simple unit test to verify both successful and failed runs based on existing tests. This is my first contribution to IPython so please let me know if I'm not doing something right :-) 

## Issues
I would like to access the full stack trace in addition to the final error. However, this seems like a fairly involved change. I'm guessing one might do this by adding the output of ``_get_exc_info`` to the ``ExecutionResult`` but I'm open to suggestions.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -431,6 +431,8 @@ class InteractiveShell(SingletonConfigurable):
 
     last_execution_succeeded = Bool(True, help='Did last executed command succeeded')
 
+    last_execution_result = Instance('IPython.core.interactiveshell.ExecutionResult', help='Result of executing the last command', allow_none=True)
+
     def __init__(self, ipython_dir=None, profile_dir=None,
                  user_module=None, user_ns=None,
                  custom_exceptions=((), None), **kwargs):
@@ -2613,7 +2615,7 @@ class InteractiveShell(SingletonConfigurable):
         result = ExecutionResult()
 
         if (not raw_cell) or raw_cell.isspace():
-            self.last_execution_succeeded = True
+            self.last_execution_succeeded, self.last_execution_result = True, result
             return result
         
         if silent:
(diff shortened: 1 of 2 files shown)


IPython/core/interactiveshell.py
```diff
@@ -2613,7 +2615,7 @@ def run_cell(self, raw_cell, store_history=False, silent=False, shell_futures=Tr
         result = ExecutionResult()
 
         if (not raw_cell) or raw_cell.isspace():
-            self.last_execution_succeeded = True
+            self.last_execution_succeeded, self.last_execution_result = True, result
```
takluyver: Let's set these on two lines, rather than doing two things on one line. It's a bit longer, but it makes it clear and simple.

takluyver (commented):
Thanks, this looks pretty good to me.

You should be able to get the traceback from `exc.__traceback__`, I think.

There are a couple of failing tests we'll need to sort out. This is because we're adding a new place to store results, and we IPython has some functions to clear those - `%reset` should clear everything, and `%xdel foo` should delete all references to foo. We need to add the new field to the things they clear/check. Have a go at that, and let us know if it's not clear. :-)

IPython/core/interactiveshell.py
drdarshan: Thanks, I'll clean this up, make sure the ``%reset`` and ``xdel`` tests pass and resubmit the PR.

takluyver: There's no need to resubmit - just push more commits to the same branch, the pull request will update and the tests will re-run. :-)

drdarshan: OK, looks like the tests passed. 

I reformatted the code per your input. Please let me know if this looks good to be merged in.

Thank you!

takluyver:
Thanks, that looks great :-)

PR #10706, opened 2017-07-19
Fix typo

nonatomiclabs:
Just some typo with the word "embedded" 😉

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1410,7 +1410,7 @@ class IPCompleter(Completer):
             pass
         elif not (inspect.isfunction(obj) or inspect.ismethod(obj)):
             if inspect.isclass(obj):
-                #for cython embededsignature=True the constructor docstring
+                #for cython embeddedsignature=True the constructor docstring
                 #belongs to the object itself not __init__
                 ret += self._default_arguments_from_docstring(
                             getattr(obj, '__doc__', ''))
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1410,7 +1410,7 @@ class IPCompleter(Completer):
             pass
         elif not (inspect.isfunction(obj) or inspect.ismethod(obj)):
             if inspect.isclass(obj):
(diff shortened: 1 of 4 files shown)


IPython/core/completer.py
```diff
@@ -1410,7 +1410,7 @@ def _default_arguments(self, obj):
             pass
         elif not (inspect.isfunction(obj) or inspect.ismethod(obj)):
             if inspect.isclass(obj):
-                #for cython embededsignature=True the constructor docstring
+                #for cython embeddedsignature=True the constructor docstring
```
takluyver: This should actually say `embedsignature` - see the Cython docs here: http://cython.readthedocs.io/en/latest/src/reference/compilation.html

IPython/terminal/embed.py
```diff
@@ -19,7 +19,7 @@
 from traitlets import Bool, CBool, Unicode
 from IPython.utils.io import ask_yes_no
 
-class KillEmbeded(Exception):pass
+class KillEmbedded(Exception):pass
```
takluyver: In this specific case, I think this is OK, but in general we should be very careful about renaming exposed objects, because it would break any code using them. I don't think this is likely to be used outside our code, though.

takluyver (commented):
Thanks, this looks good - just one thing to fix.

IPython/core/completer.py
nonatomiclabs: Thank you, I looked for more info about that one but did not find anything

IPython/terminal/embed.py
nonatomiclabs: I agree on both points 🙂

takluyver: Actually, sorry, I've changed my mind on this one - I think it may be used by third party code, e.g. like this

```python
try:
    for x in range(many):
        f(x)
        embed()
except KillEmbeded:
    pass
```

So I think we should leave the existing name as an alias for a release or two.

takluyver:
Thanks both :-)

PR #10707, opened 2017-07-20
Document magic profile as deprecated

### LINKED ISSUE #10693: %profile magic is deprecated but still in the documentation
If I try to display my current profile using `%profile`, I get a nice deprecation warning:

```python
In [1]: %profile
UsageError: The `%profile` magic has been deprecated since IPython 2.0. and removed in IPython 6.0. Please use the value of `get_ipython().profile` instead to see current profile in use. Perhaps you meant to use `%prun` to profile code?
```

This is all well and good, except the only reason I tried it is because of the [current docs](http://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-profile). Should I submit a PR removing that line?

Incidentally, I came across this while trying to execute my profile within ptipython. So any suggestions for doing that are welcome!

Carreau:
Closes #10693

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -303,12 +303,14 @@ Currently the magic system has the following functions:""",
 
     @line_magic
     def profile(self, parameter_s=''):
-        """Print your currently active IPython profile.
+        """DEPRECATED since IPython 2.0.
+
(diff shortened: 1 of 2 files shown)


takluyver:
@meeseeksdev backport

jni:
Oops, sorry @Carreau, I made the commit but never opened the PR! Thanks for taking it on board!

Issue #10045, opened 2016-11-07
elegant way to display gif file in notebook?


hainm:
Hi,

Is there any way to display gif file in super short way? I have at two ways but none of them are really what I want

- Using `HTML`
```python
def display_gif(fn):
    from IPython import display
    return display.HTML('<img src="{}">'.format(fn))
```

This does not play nicely with browser's cache. If I remake an image with the same filename, the new image won't be updated.

- Rename `.gif` to `.png` and then use `display.Image(fn)`.

thanks.

takluyver:
Anyone know if data urls work across browsers with gifs? Maybe we just need to add it to `IPython.display.Image._ACCEPTABLE_EMBEDDINGS`.

vbraun:
Data uris of course work with gif, or any other mime type that the browser supports:

    '<img src="data:image/gif;base64,{0}"/>'.format(base64.b64encode(gif))

gnestor:
As discussed, there are couple different ways to do this:

### HTML

```py
from IPython.display import HTML
HTML('<img src="../giphy.gif">')
```
### `Image(filename=*)` or `Image(url=*)` by adding a `png` extension to GIF file

```py
from IPython.display import Image
Image(filename="../giphy.gif.png")
```

### `Image(data=*)`

```py
with open('../giphy.gif','rb') as f:
    display(Image(data=f.read(), format='png'))
```

Ideally, we want to be able to `Image(filename=*)` or `Image(url=*)` without being required to rename the file and be able to use `format="gif"`. Is there any reason that GIF is not supported format yet? If not, I say we support it 👍

ferreirafabio:
how to arrange them horizontally?

gnestor:
Using `HTML`, it's pretty straight-forward:

```py
from IPython.display import display, HTML

HTML('''<div style="display: flex; justify-content: row;">
    <img src="../giphy.gif">
    <img src="../giphy.gif">
    <img src="../giphy.gif">
</div>''')
```

denfromufa:
Only the third option works in Colab notebooks:

https://github.com/ipython/ipython/issues/10045#issuecomment-318202267

Miladiouss:
Just to make it easier to copy, I will repeat the method that works in IPython, Jupyter, and CoLab:

```Python
from IPython.core.interactiveshell import InteractiveShell
InteractiveShell.ast_node_interactivity = "all"
from IPython import display
```

It is important to run the above cell _before_ running the cell below.

```Python
[...]

PR #10716, opened 2017-07-27
Support gif files, urls, and format

gnestor:
This supports the following usage:

```py
from IPython.display import Image, HTML, display

Image(filename="../giphy.gif")

Image(url="https://media.giphy.com/media/xUA7b6vHi778bTElqg/giphy.gif")

with open('../giphy.gif','rb') as f:
    display(Image(data=f.read()), format="gif")
```

It uses the png format because it wasn't working otherwise.

Depends on a new ipykernel release that includes https://github.com/ipython/ipykernel/pull/254.

Closes https://github.com/ipython/ipython/issues/10045

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -23,7 +23,7 @@ from IPython.utils.py3compat import cast_unicode
 from IPython.testing.skipdoctest import skip_doctest
 
 __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
-'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json',
+'display_svg', 'display_png', 'display_jpeg', 'display_gif', 'display_latex', 'display_json',
 'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
 'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'JSON', 'GeoJSON', 'Javascript',
 'Image', 'clear_output', 'set_matplotlib_formats', 'set_matplotlib_close',
@@ -96,6 +96,7 @@ def publish_display_data(data, metadata=None, source=None, *, transient=None, **
     * application/javascript
     * image/png
     * image/jpeg
+    * image/gif
     * image/svg+xml
 
     Parameters
@@ -256,6 +257,7 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
       - `_repr_json_`: return a JSONable dict
       - `_repr_jpeg_`: return raw JPEG data
       - `_repr_png_`: return raw PNG data
+      - `_repr_gif_`: return raw PNG data
       - `_repr_svg_`: return raw SVG data as a string
       - `_repr_latex_`: return LaTeX commands in a string surrounded by "$".
       - `_repr_mimebundle_`: return a full mimebundle containing the mapping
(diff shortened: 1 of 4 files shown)


IPython/core/display.py
```diff
@@ -256,6 +257,7 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
       - `_repr_json_`: return a JSONable dict
       - `_repr_jpeg_`: return raw JPEG data
       - `_repr_png_`: return raw PNG data
+      - `_repr_gif_`: return raw PNG data
```
takluyver: PNG -> GIF

IPython/core/display.py
```diff
@@ -927,7 +945,7 @@ def __init__(self, data=None, url=None, filename=None, lib=None, css=None):
             raise TypeError('expected sequence, got: %r' % css)
         self.lib = lib
         self.css = css
-        super(Javascript, self).__init__(data=data, url=url, filename=filename)
+        superg(Javascript, self).__init__(data=data, url=url, filename=filename)
```
takluyver: Typo

IPython/core/display.py
```diff
@@ -1059,6 +1078,9 @@ def __init__(self, data=None, url=None, filename=None, format=None,
                     format = self._FMT_JPEG
                 if ext == u'png':
                     format = self._FMT_PNG
+                if ext == u'gif':
+                    # use PNG format until we understand why GIF is not working
+                    format = self._FMT_PNG
```
takluyver: Probably because the frontend needs to handle the gif mimetype.

takluyver (commented):
I think this probably needs some gif handling in the notebook frontend so we can write this code correctly (i.e. not pretending it's a png).

IPython/core/display.py
gnestor: I think you're right. I created [a branch](https://github.com/gnestor/notebook/tree/image-gif) for this very purpose but the issue I am running into is that when running the following:

```py
with open('../giphy.gif','rb') as f:
    display(Image(data=f.read(), format="gif"))
```

the data-uri looks like `data:image/gif;base64,GIF89a�w�2MekdWE71Sv�������gWN` 

When using `format="png"`, the data-uri looks like `data:image/png;base64,R0lGODlh9AF3AfcAADJNZWtkV0U3MVN2n6qai+rWyWd...`. 

For some reason, the PNG format encoding the data different than the GIF format but I can't find any code in ipython or the notebook that does any encoding. 

Any ideas?

takluyver: It's in ipykernel, which is a rather icky bit of tight coupling. :-(

https://github.com/ipython/ipykernel/blob/8acaee80c83c9a64a22afde522f9df2d0c3638e4/ipykernel/jsonutil.py#L51

rgbkrk (approved):

takluyver (commented):
I think this should wait for ipython/ipykernel#254, then a new release of ipykernel - if people get this without the corresponding upgrade to ipykernel, then displaying gifs will silently fail.

So I've rather conservatively marked it for 7.0, so we don't merge it too quickly. We can probably bring it back to some 6.x release once we've got a new ipykernel out.

IPython/core/display.py
```diff
@@ -1100,10 +1121,12 @@ def _retina_shape(self):
...
         if not self.embed:
             return
-        if self.format == 'png':
+        if self.format == self._FMT_PNG:
             w, h = _pngxy(self.data)
-        elif self.format == 'jpeg':
+        elif self.format == self._FMT_JPEG:
             w, h = _jpegxy(self.data)
+        elif self.format == self._FMT_GIF:
+            w, h = _pngxy(self.data)
```
gnestor: No, it's not! I just figured out how to do it for gif 👍

IPython/core/tests/test_display.py
```diff
@@ -77,8 +77,8 @@ def test_base64image():
 def test_image_filename_defaults():
     '''test format constraint, and validity of jpeg and png'''
     tpath = ipath.get_ipython_package_dir()
-    nt.assert_raises(ValueError, display.Image, filename=os.path.join(tpath, 'testing/tests/badformat.gif'),
-                     embed=True)
+    # nt.assert_raises(ValueError, display.Image, filename=os.path.join(tpath, 'testing/tests/badformat.gif'),
```
gnestor: Yes, perfect 👍

minrk:
With #10739, this doesn't need anything from any other packages.

We could also do this without defining `_repr_gif_` or `GIFFormatter` by using `_repr_mimebundle_`. Part of the point of adding that was eliminating the need to continue defining new formatters and magic methods when we support new mime-types. I would prefer if GIFs were implemented that way, without the new class and magic method.

takluyver:
Finished in #10748.

tohnperfect:
Is this working now?
I have iPython Version: 6.2.1

but got `publish() got an unexpected keyword argument 'format'` error when I try to display a gif image

PR #10719, opened 2017-07-29
Update object inspector colorscheme when needed.

### LINKED ISSUE #10465: difference in source highlighting between 5.3 and 6.0rc1 using %colors Linux
The following screenshot shows the difference in source highlighting between the two versions of IPython (note the difference in version between left and right):
screenshot_20170419_152049
The "null" profile is just an empty, freshly created profile.

Arch Linux, pip-installed ipython.

anntzer:
Fixes #10465.


takluyver:
Thanks - do I understand correctly from the issue that this is only needed for 6.x?

anntzer:
I guess so.

PR #10722, opened 2017-07-29
[Prototype] Generalize magic machinery to allow completion.

Carreau:
In particular useful for multi language integration. One of defect is
that you can't really complete. But what you can do is actually
dispatch the completion to the Magic Class (if you really interested in
integration anyway you go with a magic Class. And dispatch to the class.

So far it completes only for cell magics. I need to poke at
InputSpliter to know if we can figure out we're completing a line magic
which is not at the beginning of the line.

There are a few question remaining:
  - How to tell IPython to give-up (or not) on completing using jedi and
  Python completion. Indeed for %%timeit you just want to __extend__
  completion with your own. For %%sql, you don't want to include Python
  completions.
  - What API should we provide ?
    - here is line/cell/cursor figure things out with a single function
    - Allow user to register a separate completer for the first line,
    and the core of the cell independently.

I'd like to have a prototype that can forward the completion to another
kernel (hook into the python2 magic and have a Python 3 kernel that
control a Python2 kernel ?

-- 
CC @ivanov

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1802,6 +1802,15 @@ class IPCompleter(Completer):
         have lots of processing to do.
 
         """
+        if full_text.startswith('%%') and '\n' in full_text:
+            line, *rest = full_text.split('\n')
+            magic = line.split(' ')[0][2:]
+            magic_completer = self.shell.magics_manager.magics['completer'].get(magic, None)
+            if magic_completer:
+                for c in magic_completer(line, '\n'.join(rest), offset):
+                    assert c.start <= offset
+                    yield c
+            return
         deadline = time.monotonic() + _timeout
 
 
--- a/IPython/core/magic.py
+++ b/IPython/core/magic.py
@@ -17,15 +17,21 @@ import sys
 from getopt import getopt, GetoptError
 
 from traitlets.config.configurable import Configurable
+
 from IPython.core import oinspect
 from IPython.core.error import UsageError
 from IPython.core.inputsplitter import ESC_MAGIC, ESC_MAGIC2
+
 from decorator import decorator
+
 from IPython.utils.ipstruct import Struct
 from IPython.utils.process import arg_split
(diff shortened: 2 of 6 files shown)


Carreau:
Working on this with @unnamedplay-r during sprints. 

We though a bit about the API to register completions, you have to take into consideration a couple of things: 
  - Many magics can be defined in a single class, 
  - even if the name of the function is for example `sql` you **can** register it with another name when you call register_magic. 

We had idea of a couple of way to register completer:

```
@completer
def mycompleter(self, line, cell, offset):
    # let author dispatch to the right cell magic depending on what `line` starts with
    # and wether the cursor is on the first line or not.
    for...
        yield completion
```



```
@sql.completer
def mycompleter(self, line, cell, offset):
    # let author dispatch wether the cursor is on the first line or not.
    for... 
      yield completion
```

This second one would automatically dispatch for SQL (in this example), but this might need metaclass to work properly or Python might say "Sql" not define, this mimic the `@property` decorator.

We can also use named conventions, 

```
def sql_completer(....):
    ...
```

Appending `_completer` to a function would make it the completer for the corresponding function.
and the more explicit one:

```
@completer_for('sql')
def mycompleter(self, line, cell, offset):
    ...
```

Where the author need to be explicit. We then to like option 4 (`@completer_for('sql')`) easy to grep and relatively obvious and easy to implement.

For each of these there is another alternative to split the completer into 2 completer function, for wether the cursor is on the `line` or in the `cell`. We're not sure this is a good idea, as some argument of the line, may affect the cell completion so getting both at once could be useful, and it's relatively easy to provide convenience function to strip the first line and correct the cursor offset. 

We still need to exactly figure out how to map the completer correctly in case the magic get registered with a different name.. but that's an implementation detail I guess.

Carreau:
Note to self: have a look at `strdispatchers`/`complete_command` that cover some of this roles.

Issue #10724, opened 2017-07-31
runtime configuration when embedding ipython


blueForestIcarus:
It seems that there is no way to set configuration options from a Python source file before embedding (either with .embed or .start_ipython). For example, I would like to set exec_lines right before calling start_ipython or embed. The versatility and simplicity of a python based configuration system are lost if the programmer does not have the ability to set configuration options in his own code just as they are set in the config file. If this is possible in the current system, then it should be documented (I looked through the IPython and Traitlets doc's and haven't seen anything), otherwise, I believe it should be added.

takluyver:
Something like this:

```python
from traitlets.config import Config
import IPython
c = Config()
c.Class.config_value = 'foo'
IPython.start_ipython(config=c)
```

blueForestIcarus:
After seeing your response and looking back at the [docs](https://ipython.readthedocs.io/en/stable/interactive/reference.html) I noticed similar code In one of the embed examples; my bad, I should have looked closer.

takluyver:
Yeah, docs can be tricky - there's a million things people might want to do, and it's not obvious where to look, or where to put things for people to find. Where were you looking for this? Is there somewhere where an example would clearly benefit from showing it?

blueForestIcarus:
The best place for an example would probably be here https://ipython.readthedocs.io/en/stable/config/intro.html. That was where I first looked for config documentation, and it explained three ways to set configuration but lacked any information on how to set configuration when embedding IPython. A fourth section to this page with an example would be great, and a link to it from [this](https://ipython.readthedocs.io/en/stable/interactive/reference.html#embedding-ipython) page would be useful as well.

takluyver:
Sounds sensible. Do you want to make a PR adding a short section about it? The docs are in the `docs/source` folder of this repo.

PR #10727, opened 2017-08-01
Configuration example for embedded IPython

blueforesticarus:
Created an example file and added it to the docs to illustrate how one might configure IPython in an embedded context.

Issue: #10724

### DIFF
--- a/docs/source/config/intro.rst
+++ b/docs/source/config/intro.rst
@@ -112,6 +112,18 @@ At present, this only affects the current session - changes you make to
 config are not saved anywhere. Also, some options are only read when
 IPython starts, so they can't be changed like this.
 
+.. _configure_embedded:
+
+Configuring embedded IPython
+----------------------------
+
+If you are :ref:`Embedding`, you can set configuration options the same
+way as in a config file by creating a traitlets config object and passing
+it to start_ipython like in the example below.
+
+.. literalinclude:: ../../../examples/Embedding/embed_configured.py
+    :language: python
+
 .. _profiles:
 
 Profiles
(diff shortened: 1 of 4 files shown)


docs/source/config/intro.rst
```diff
@@ -112,6 +112,18 @@ At present, this only affects the current session - changes you make to
 config are not saved anywhere. Also, some options are only read when
 IPython starts, so they can't be changed like this.
 
+.. _configure_embedded:
+
+Configuring embedded IPython
```
takluyver: 'Embedding' has a particular meaning in IPython: when you call `IPython.embed()`, you can see the variables from the scope where it was called in the shell. `start_ipython()`, which your example uses, starts IPython normally without this. We encourage people to use `start_ipython()` if they don't need that, because it's more robust.

So maybe we should call this something like 'Configuration for launching IPython from Python code'? Better ideas welcome :-)

blueForestIcarus: Doesn't embed() do the same thing as start_ipython, except it grabs the global and local namespaces automatically?
Anyway, since the file is entirely to do with configuration, a heading of "Starting IPython from Python" would be sufficiently obvious. Other candidates could be something all the lines of "Using a Python caller" or "Configuring directly from caller code" but I think "Starting IPython from Python" makes the most sense. 

Also, doesn't the first code block under the embedding section of the [IPython reference](https://ipython.readthedocs.io/en/stable/interactive/reference.html#embedding-ipython) use start_ipython?

takluyver: Yeah, we added that because we found that quite a few people were embed()-ing when they didn't really need that, because we hadn't made start_ipython() obvious before. It's meant to be an 'instead of embedding, try this...' note, but it does make it a bit confusing what we actually mean by embedding.

"Starting IPython from Python" sounds good to me. :-)

blueForestIcarus:
Okay, I changed references to embedding to "running ipython from python."

takluyver:
Thanks!

PR #10730, opened 2017-08-02
Deactivate sorting on dict keys when pretty printing.

### LINKED ISSUE #10110: python3.6 dict printing
IPython's sorting of dictionary keys before pretty printing them made sense before Python 3.6 dictionaries.

Now that dictionaries in 3.6 preserve insertion order, is it still necessary to sort them?

This would prevent odd behavior such as:

```
In [1]: d = dict(a = 1, z = 2, c = 3)

In [2]: d
Out[2]: {'a': 1, 'c': 3, 'z': 2}

In [3]: print(d)
{'a': 1, 'z': 2, 'c': 3}

```

Carreau:
Closes #10110

That is a backward incompatible change, but as the dictionary are (as an
implementation details of Python as of 3.6) ordered and teaching Python
often need to show that (for kwargs), it is (IMHO) a better choice to
remove the sorting.

If one id unhappy with that it is still possible to register its own
pretty printer.

-- 
Needs what's new, doc and test ?

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -600,13 +600,6 @@ def _dict_pprinter_factory(start, end, basetype=None):
         step = len(start)
         p.begin_group(step, start)
         keys = obj.keys()
-        # if dict isn't large enough to be truncated, sort keys before displaying
-        if not (p.max_seq_length and len(obj) >= p.max_seq_length):
-            try:
-                keys = sorted(keys)
-            except Exception:
-                # Sometimes the keys don't sort.
-                pass
         for idx, key in p._enumerate(keys):
             if idx:
                 p.text(',')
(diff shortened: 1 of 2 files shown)


minrk:
I don't think we should do this. When looking at a dictionary interactively, it's extremely useful for the keys to be sorted so that the same dictionary always looks the same (different key order is *not* a different dictionary).

I don't think displaying the dictionary is the right way to indicate that key order can change in certain Python implementations (why would you assume that the repr takes the key order into account rather than sorting)? I think if you want to illustrate key order, explicitly iterating through the dict keys is a better way to illustrate this idea.

takluyver:
Explicitly iterating through it can be as simple as `list(d)`.

Carreau:
Closing this then, we can revisit later. Thanks for the input.

PR #10732, opened 2017-08-04
Smarter documentation of config options

takluyver:
* Make config options individually linkable (I've wanted this for ages)
  * As a side effect, they also show up as objects in Sphinx's search.
* Only show default values when they're interesting (not like `''` or `[]`)
* Show any command line aliases which map to a config option.

Before:

screenshot from 2017-08-04 15-55-10

After:

screenshot from 2017-08-04 15-55-17

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -76,7 +76,7 @@ source/interactive/magics-generated.txt: autogen_magics.py
 
 autoconfig: source/config/options/config-generated.txt
 
-source/config/options/config-generated.txt:
+source/config/options/config-generated.txt: autogen_config.py
 	$(PYTHON) autogen_config.py
 	@echo "Created docs for config options"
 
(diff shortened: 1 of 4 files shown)


takluyver:
It probably makes sense to migrate most of this functionality to traitlets once we're happy with it, so we can use it in other docs. But it's easiest to work on it in one repo which uses it.

Carreau (approved):

minrk (approved):

takluyver:
Has the merge button started eating mice or something? :stuck_out_tongue: 

The Travis errors are unrelated - I think there was a temporary glitch in Github, which seems to have got it stuck in a bad state.

mpacer:
Could you add the ability to hover over where the link would pop up by keeping it as an element but just hiding it (or not displaying it, I forget which one maintains the boundaries of the element for hovers)?

takluyver:
That certainly makes sense to me, but it's up to the HTML theme how that works throughout the docs. For IPython, we're using [sphinx_rtd_theme](https://github.com/rtfd/sphinx_rtd_theme) at present, though in some of the other projects we've switched to a Jupyter sphinx theme (e.g. [JupyterHub](http://jupyterhub.readthedocs.io/en/latest/)).

PR #10739, opened 2017-08-08
return base64-encoded image data from image repr methods

minrk:
avoids any need for ipykernel to detect ambiguous str/bytes data.

With this, #10716 needs nothing from ipykernel.

I'll have a PR to ipykernel as well that should clarify that the b64 image detection there is legacy Python 2 code that shouldn't be needed on Python 3.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -5,12 +5,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 
-try:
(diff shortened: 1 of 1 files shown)


gnestor (approved):
I cherry-picked this commit into https://github.com/ipython/ipython/pull/10716 to test and it's working nicely with ipykernel master and notebook master 👍

PR #10740, opened 2017-08-09
preserve priority of user-registered per-mime renderers

minrk:
fixes case where `_repr_mimebundle_` method had higher priority than custom user-registered formatters for single mime-types.

There's one case that's ambiguous: user-registered *both* mime-bundle and single-mime for the same mime-type.

- Prioritizing mime-bundle matches the priority of methods
- Prioritizing single-mime allows the "more specific" choice to take precedence (current state)

I think both make sense, but I would lean slightly toward the second option, which this PR currently implements.

Brought up in comments in #10651

cc @Carreau


Carreau:
+1

PR #10743, opened 2017-08-15
Limit number of completions returned

takluyver:
If there are a very large number of completions (e.g. tens of thousands of files), the notebook frontend can have trouble handling them - see https://github.com/jupyter/help/issues/198 . I don't see any practical need for so many completions in the UI - if there are that many, you're likely to type some more letters and try completing again.

This limits the returned completions to 500. It's not configurable, but there's a clear constant so someone could patch it locally if there was an unexpected need for a larger number of completions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ _build
 docs/man/*.gz
 docs/source/api/generated
 docs/source/config/options
+docs/source/config/shortcuts/*.csv
 docs/source/interactive/magics-generated.txt
 docs/gh-pages
 jupyter_notebook/notebook/static/mathjax
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -161,6 +161,9 @@ if sys.platform == 'win32':
 else:
     PROTECTABLES = ' ()[]{}?=\\|;:\'#*"^&'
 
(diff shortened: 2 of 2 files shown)


Carreau:
JupyterLab May not like that as (IIRC) they use client-side completion filtering for as-you type. returning only the fist 500 may be having undesirable side effects.

minrk:
Is it clear where it has difficulty? If it's in the display and not the message, maybe this filtering should be done client-side rather than kernel-side. If it's in the transfer/serialization of the message itself, then doing it in the kernel makes sense.

takluyver:
I didn't investigate that, but the issue was about 70k files appearing as completions. If it takes 15 characters for each completion, that's sending over 1MB of data. Even if the main slowdown is from display when it's local, if you're accessing a remote server over a low-bandwidth connection, that's a lot of data to be transferring when you hit tab.

minrk:
Makes sense, thanks

takluyver:
Do we want this backported to 5.x?

Carreau:
> Do we want this backported to 5.x?

*want* ? Not really. Maybe we should for consistency though.

takluyver:
I'm inclined to think we should.

@meeseeksdev backport

PR #10748, opened 2017-08-18
implement GIF support without registering a new formatter

minrk:
finishes up PR #10716

we shouldn't need to define a new Formatter class or `_repr_foo_` method ever again now that we have `_repr_mimebundle_`

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -23,7 +23,7 @@ from IPython.utils.py3compat import cast_unicode
 from IPython.testing.skipdoctest import skip_doctest
 
 __all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
-'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json',
+'display_svg', 'display_png', 'display_jpeg', 'display_gif', 'display_latex', 'display_json',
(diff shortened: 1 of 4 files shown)


IPython/core/display.py
```diff
@@ -961,20 +951,30 @@ def _jpegxy(data):
             # read another block
             idx += 2
 
-    h, w = struct.unpack('>HH', data[iSOF+5:iSOF+9])
-    return w, h
+    return struct.unpack('>HH', data[iSOF+5:iSOF+9])
```
takluyver: The code this is replacing swaps round the variables - it gets `h, w` and returns `w, h`. That seems like something it should keep.

minrk: Yup, I just saw that. Inherited from the other PR. Fixed now.

IPython/core/display.py
```diff
@@ -81,17 +81,7 @@ def publish_display_data(data, metadata=None, source=None, *, transient=None, **
...
-    * text/plain
-    * text/html
-    * text/markdown
-    * text/latex
-    * application/json
-    * application/javascript
-    * image/png
-    * image/jpeg
-    * image/svg+xml
+    Keys of data and metadata can be any mime-type.
```
rgbkrk: I love how much simpler this makes working with the classes down below.

PR #10750, opened 2017-08-18
Give the example for custom magics in the docs.

Carreau:
The docs are registering magics by getting the current instance of
IPython and registering directly against it. A better approach is to
actually show the example on how to use the `load_ipython_extension`
function to correctly define a magic.

### DIFF
--- a/docs/source/config/custommagics.rst
+++ b/docs/source/config/custommagics.rst
@@ -84,12 +84,17 @@ IPython object:
 
 
     # In order to actually use these magics, you must register them with a
-    # running IPython.  This code must be placed in a file that is loaded once
-    # IPython is up and running:
-    ip = get_ipython()
-    # You can register the class itself without instantiating it.  IPython will
-    # call the default constructor on it.
-    ip.register_magics(MyMagics)
+    # running IPython.
+
+    def load_ipython_extension(ipython):
+        """
+        Any nodule file that define a function named load_ipython_extension can
+        be loaded via `load_ext` or be configured to be autoloaded by IPython at
+        startup time.
+        """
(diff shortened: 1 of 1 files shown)


docs/source/config/custommagics.rst
```diff
@@ -84,12 +84,17 @@ IPython object:
...
-    ip = get_ipython()
-    # You can register the class itself without instantiating it.  IPython will
-    # call the default constructor on it.
-    ip.register_magics(MyMagics)
+    # running IPython.
+
+    def load_ipython_extension(ipython):
+        """
+        Any nodule file that define a function named load_ipython_extension can
+        be loaded via `load_ext` or be configured to be autoloaded by IPython at
```
takluyver: Let's put `%load_ext` to make it clearer that it's a magic.

docs/source/config/custommagics.rst
```diff
@@ -108,11 +113,16 @@ instantiate the class yourself before registration:
...
-    ip.register_magics(magics)
+    def load_ipython_extension(ipython):
+        """
+        Any nodule file that define a function named load_ipython_extension can
+        be loaded via `load_ext` or be configured to be autoloaded by IPython at
+        startup time.
+        """
+        # This class must then be registered with a manually created instance,
+        # since its constructor has different arguments from the default:
+        magics = StatefulMagics(ip, some_data)
```
takluyver: ip -> ipython as that's what it's called in the function definition.

Carreau:
Feedback from sprints, it's unclear that load_ipython_extension should be in `__init__.py` and maybe link to flit on how to publish a package.

Carreau:
Thanks !

PR #10756, opened 2017-08-26
Mpl magics

tacaswell:
The first commit is (I think) non-controversial

The second is something I would like, but may be too aggressive.


Carreau:
+1 to both.

Carreau (approved):

tacaswell:
Can I also request that these be backported as far as is practical?

minrk:
@meeseeksdev backport to 5.x

Carreau:
@meeseeksdev backport

PR #10760, opened 2017-08-27
Small fix: Bang should be after the assignment operator

erenhalici:


takluyver:
Thanks! We normally put it after the space as well, `= !ls` so it's clear that it's separate from the equals sign.

erenhalici:
You're welcome. Makes sense. If you want I can update the documentation accordingly.

takluyver:
Yup, go ahead and update that line, and any other examples you spot.

erenhalici:
Done :)

takluyver:
Thanks!

@meeseeksdev backport

PR #10767, opened 2017-08-31
Sort on string representation if sorting fails

jdemeyer:


Carreau:
@meeseeksdev backport

PR #10787, opened 2017-09-11
Do not shadow explicit magic completion.

### LINKED ISSUE #10754: `%matplotlib` is not showing when pressing tab for autocompletation after `%pylab`
The issue appears in IPython 6.1, but not in 6.0
I first realize about it using Spyder: spyder-ide/spyder#5020

Nevertheless, even if it is not shown, `%matplotlib`can be executed and it works.

Carreau:
If a binding has the same name than a magic, the magic should be
shadowed; unless the use is explicitly requesting a magic with a leading
percent sign.

Closes #10754

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1221,7 +1221,9 @@ class IPCompleter(Completer):
         cell_magics = lsm['cell']
         pre = self.magic_escape
         pre2 = pre+pre
-        
+
+        explicit_magic = text.startswith(pre)
+
         # Completion logic:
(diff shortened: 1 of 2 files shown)


takluyver:
Not rerunning the failed test because I'm satisfied that it's random, and that failure is just too good to wipe out (see #10788 for details).

I'm a bit concerned that the completion code is accumulating special cases and if/else branches - this doesn't feel like a way to build reliable code that we can confidently modify. But for now, it solves a real problem.

PR #10788, opened 2017-09-11
Fix random test failure.

Carreau:
The filename was generated taking the last digits of a randomly
generated number between 0 and 1, if the number is small enough, it's
repr ends with `e-5`, or less (the probability is actually close to
1/1000 because the biggest number we can get with e-5 is (1e-4 -
epsilon).

We actually did hit that on our CI:

- https://ci.appveyor.com/project/IPython/ipython/build/1.0.839/job/mk4l08s295ynkfs5

```
ERROR: test_run_submodule_with_absolute_import (IPython.core.tests.test_run.TestMagicRunWithPackage)
  File "c:\python36-x64\lib\site-packages\IPython\core\tests\test_run.py", line 455, in test_run_submodule_with_absolute_import
    self.check_run_submodule('absolute')
  File "c:\python36-x64\lib\site-packages\IPython\core\tests\test_run.py", line 450, in check_run_submodule
    self.assertEqual(_ip.user_ns['x'], self.value,
KeyError: 'x'
-------------------- >> begin captured stdout << ---------------------
  File "C:\Users\appveyor\AppData\Local\Temp\1\tmp4asmvv33\tmpqpswwe9n\tmp065832886952393e-05\absolute.py", line 2
    from tmp065832886952393e-05.sub import x
                            ^
SyntaxError: invalid syntax
```

Actually pick a random string now we should have less chance of being
invalid identifier.

### DIFF
--- a/IPython/core/tests/test_run.py
+++ b/IPython/core/tests/test_run.py
@@ -17,6 +17,7 @@ import functools
 import os
 from os.path import join as pjoin
 import random
+import string
 import sys
 import textwrap
 import unittest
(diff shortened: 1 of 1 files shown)


takluyver:
Ha, well found! Let's see if meeseeks is on the case for backporting now.

takluyver:
It looks like meeseeks made the commit, but didn't open a PR. I'll let you debug!

Carreau:
hum, I need to fix the labelling. If there are labels applied, it wont duplicate the labels on the new PR. Thanks ! I retrigered the webhook and the PR was created, so it should work now !

PR #10792, opened 2017-09-12
Avoid crashing if interrupt the kernel during completion.

### LINKED ISSUE #10733: Attempting to interrupt the kernel while autocomplete is thinking crashes the runtime
Backstory: we've added some custom completers, some of which do things that amount to making a network request (say to find out the potential arguments to a function). Sometimes, this can be slow, and users are tempted to interrupt their kernel. Doing so then crashes the kernel.

Here's a really goofy repro:

```
def bad_completer(*args, **kwargs):
    import time
    time.sleep(10)
    return ['houses']

get_ipython().set_hook('complete_command', bad_completer, re_key='%%file')

%%file <TAB>
```

and then interrupt while you're waiting.

A naive fix would be to simply catch any exception [here](https://github.com/ipython/ipython/blob/268a37740656a2cfb033e6803e2459fdf4ece516/IPython/core/completer.py#L1703), but I'm not sure if that should happen in a more central place. 

In particular, if a user does interrupt a completer, do we want to continue with the next one, or cancel completion completely?

Carreau:
This is likely due to completion taking too long, so just return what we
have so far, and ignore the interrupt.

closes #10733

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1769,11 +1769,16 @@ class IPCompleter(Completer):
                       category=ProvisionalCompleterWarning, stacklevel=2)
 
         seen = set()
-        for c in self._completions(text, offset, _timeout=self.jedi_compute_type_timeout/1000):
-            if c and (c in seen):
-                continue
-            yield c
-            seen.add(c)
+        try:
+            for c in self._completions(text, offset, _timeout=self.jedi_compute_type_timeout/1000):
+                if c and (c in seen):
+                    continue
+                yield c
+                seen.add(c)
+        except KeyboardInterrupt:
+            """if completions take too long and users send keyboard interrupt,
+            do not crash and return ASAP. """
+            pass
 
     def _completions(self, full_text: str, offset: int, *, _timeout)->Iterator[Completion]:
         """


takluyver:
You've marked this for 5.5, but it looks like the method you're modifying is not even in 5.x. Should it be for 6.x only?

Carreau:
> You've marked this for 5.5, but it looks like the method you're modifying is not even in 5.x. Should it be for 6.x only?

Oh, the issue is marked as 5.5, i might have missed something then. Let me check.

Carreau:
Meh, you are right, it need to be fixed in multiple location and I need a different patch for 5.5

Carreau:
See #10797 that fix only the case reported on 5.x and master. I'll let this one open to investigate a proper global fix and bump #10733  to 6.x/wishlist if #10797 get merged.

takluyver:
@Carreau I'm a bit confused by your last comment. Do you still want this change in for 6.3? Or are we leaving it for further investigation?

takluyver:
Ping @Carreau - still happy for this to go in for 6.3?

takluyver:
OK, I'm assuming you're happy for this to go in.

Issue #10774, opened 2017-09-02
`post_execute` and `post_run_cell` hooks not always triggered


fniephaus:
Due to various early returns in the code below, `post_execute` and `post_run_cell` hooks are not triggered in some cases. Since the pre hooks are always called, this is either a bug or the [docs](http://ipython.readthedocs.io/en/stable/config/callbacks.html#post-run-cell) should mention that post hooks are not guaranteed to be triggered.

https://github.com/ipython/ipython/blob/ca86b79bfd5cdd633ea6cb9c18bb2c1c493402a1/IPython/core/interactiveshell.py#L2694-L2739

/cc @craigcitro

Carreau:
Thanks for the report, you are right that we should split the post_execution hooks into hooks that execute after every requests, or only after every valid Python request. I don't think we should fix the current ones, but provide variants that also trigger even when the pre-execute fails.

craigcitro:
Hey @Carreau -- I'm curious about the details of what you're proposing.

* Do you mean that you want to allow users to set separate hooks for the various ways a cell can fail, eg `post_syntax_error_hook`, `post_preprocessing_error_hook`, that sort of thing? I feel like that's a lot of complexity. For my money, I'd rather provide a way to customize the `error_before_exec` function that gets used.

* Is the motivation for adding a new hook simply to avoid a subtle behavior change? As it is, I'm betting that most (if not all) users of the `post_*` hooks actually *expect* that they're always being executed, and would be surprised by the behavior @fniephaus ran into here. But of course I'm saying that based on 0 data. 😀 😀 😀  Do you know of a use case for the current behavior?

* If we do go the route of new hooks, do you imagine having something like a `post_execute_exception_hook`, where we know exactly one of that and `post_execute_hook` will be triggered? Or something more like a `finally_hook`, where it's executed no matter what?

Thinking it through, I could see an argument that the right answer here is to say "we will be explicit about documenting and sticking to the current behavior, because it's a subtle behavior change. If you want something else, you should override `run_cell` yourself." I think that would work for our use case, but I bet we aren't the first and won't be the last to stumble on this, so we wanted to look at fixing it at the root first. 😉

Carreau:
> Do you mean that you want to allow users to set separate hooks for the various ways a cell can fail

Probably not _that_ granular. 

> Is the motivation for adding a new hook simply to avoid a subtle behavior change?

Yes. 

> But of course I'm saying that based on 0 data

[...]

PR #10795, opened 2017-09-13
Ensure post event callbacks are always triggered.

fniephaus:
This PR addresses #10774 by introducing two *finally* event callbacks: `finally_execute` and `finally_run_cell`. They work similar to the corresponding *post* callbacks, but are guaranteed to be triggered (even when, for example, a ``SyntaxError`` was raised). Also, the execution result is provided as an argument for further inspection.

The PR includes:
- Code and tests
- Docs (incl. example)
- PR feature description

/cc @craigcitro


takluyver:
API question for anyone interested: does it make more sense to have 'finally' callbacks that fire regardless of whether code is executed, or 'error' callbacks that only fire when the 'post' ones wouldn't? It should be possible to do the same things with either API,  but I wonder if there's any design reason to prefer one over the other.

fniephaus:
First, 'post' callbacks don't get the result as an argument, 'error' callbacks would probably just expect the error as an argument. Therefore, the current version of 'finally' callbacks is more powerful as it allows to inspect the execution result.
Also, the docs on 'post' callbacks states that they can be used for, for example, cleanups etc. Let's just say a 'pre' callback creates a file. If you want that to be removed, you'd need to add the logic to both, 'post' *and* 'error' callbacks. 'Finally' callbacks seem to be an easy way to cover all use cases, as it's simple to check whether an error was thrown or not ([see example](https://github.com/fniephaus/ipython/commit/1ce57fed5113a195dab05a3022f8b887c43ed512#diff-e41e73b39515e5e4cf42b43dd06a3ab5)).

docs/source/config/callbacks.rst
```diff
@@ -64,16 +71,32 @@ skipping the history/display mechanisms, in which cases ``pre_run_cell`` will no
 post_run_cell
 -------------
 
-``post_run_cell`` runs after interactive execution (e.g. a cell in a notebook).
-It can be used to cleanup or notify or perform operations on any side effects produced during execution.
-For instance, the inline matplotlib backend uses this event to display any figures created but not explicitly displayed during the course of the cell.
-
+``post_run_cell`` runs after successful interactive execution (e.g. a cell in a
+notebook, but, for example, not when a ``SyntaxError`` was raised).
```
takluyver: If we leave the `post_` events as they are (I think we should consider changing them), this documentation needs to be a bit more precise. They still fire after a normal error (e.g. from `1/0`), it's only if preprocessing/parsing/compiling the code fails, so we don't actually execute it, that they don't fire.

takluyver (commented):
Thanks, I'm convinced about doing `finally_` events rather than `error_` events. A few other things I'd like to consider:

1. Maybe it's actually fine to make the `post_` events fire in all the cases where the new `finally_` events do. I don't know if there are any real use cases that would be broken by firing them after a syntax error.
2. We could actually add parameters to existing events... I wrote a little package a few years back to deal with precisely that: [backcall](https://pypi.python.org/pypi/backcall). I don't think it's ever been seriously used, so it might be rough, but I was thinking of IPython's events API when I wrote it.
3. Naming: `finally_execute` will fire in cases where code has not been executed. Maybe it should just be `finally_run_cell` and take `silent` as another parameter.

docs/source/config/callbacks.rst
fniephaus: Sure, will update this in the next commit...

fniephaus:
> 1. Maybe it's actually fine to make the `post_` events fire in all the cases where the new `finally_` events do. I don't know if there are any real use cases that would be broken by firing them after a syntax error.
> 2. We could actually add parameters to existing events... I wrote a little package a few years back to deal with precisely that: backcall. I don't think it's ever been seriously used, so it might be rough, but I was thinking of IPython's events API when I wrote it.

I'm not familiar with IPython's backwards compatibility policy, but introducing `finally_` events appears to be safer than changing the semantics of existing events. Also, adding parameters to existing events would cause method signature mismatches, which might also not be great in terms of compatibility. What other parameters were you thinking about?

> 3. Naming: `finally_execute` will fire in cases where code has not been executed. Maybe it should just be `finally_run_cell` and take `silent` as another parameter.

Purely for consistency reasons, I'm still in favor of adding both, `finally_execute` and `finally_run_cell`.

takluyver:
> adding parameters to existing events would cause method signature mismatches

That's precisely what backcall is designed to handle - you can add parameters  and callbacks that don't accept them will still work. I agree that it's reasonable to just add new events, though.

minrk:
I'd be happy if this were the behavior of the existing events and `backcall` were used for backward-compatibility of the signature. I imagine that most people writing current event handlers would expect this behavior rather than the current. I know that most times that I've written execute hooks I wished that the cell text were part of the callback signature.

fniephaus:
Okay, so what do we want to do? Modify and extend existing callbacks or introducing new ones?
I'm personally not a big fan of adding more complexity in cases like this, so I'd still be in favor of adding new callbacks. Otherwise, I'm happy to incorporate `backcall`'s approach to provide backwards compatibility.

takluyver:
I'd be in favour of trying out backcall to extend existing callbacks, but it's my weird side project, so I shouldn't be the one making that call. @minrk? @Carreau ?

minrk:
👍 to giving backcall a go and applying this to the existing events.

fniephaus:
Ok, cool. Will update this PR accordingly later this week...

fniephaus:
PTAL, I have:

- Adjusted the behavior of the *post* callbacks as discussed;
- Removed the *finally* callbacks again;
- Added the `result` object as an argument to both *pre* and *post* callbacks. Because we would need to use a backport of `inspect.signature` for Python2, I've decided to use a much simpler approach instead: the previous versions of all *pre* and *post* callbacks had no arguments, so compatibility wrappers are only used for callback functions without an argument.

Carreau:
Thanks all for restarting that and following up, I'm slowly getting back to review PR and push IPython forward, apologies for the delay. I'll try to catch up with the discussion. 

> Because we would need to use a backport of inspect.signature for Python2,

Unless you want this PR back ported to the 5.x branch that will be unnecessary as IPython is already 3.3+ already. I do understand though that backporting could help to make external code easier to maintain on two version of IPython.

Carreau (approved):
+1, and I believe that should be relatively easy to backport.

fniephaus:
> Unless you want this PR back ported to the 5.x branch that will be unnecessary as IPython is already 3.3+ already. I do understand though that backporting could help to make external code easier to maintain on two version of IPython.

I don't think we need this code to be back ported to the 5.x branch. But since the current version uses a simple import fallback for Python2 compatibility and does not depend on `inspect.signature`, it should be easy to do as you already mentioned.

IPython/core/events.py
```diff
@@ -90,29 +125,53 @@ def _define_event(callback_proto):
 # ------------------------------------------------------------------------------
 
 @_define_event
-def pre_execute():
+def pre_execute(result):
```
takluyver: It seems strange for the *pre* events to carry a result object. Also, the code in `interactiveshell` that triggers these events does not give them the result.

IPython/core/events.py
```diff
@@ -56,11 +78,24 @@ def register(self, event, function):
...
             raise TypeError('Need a callable, got %r' % function)
-        self.callbacks[event].append(function)
+
+        callback_proto = available_events.get(event)
+        if (isfunction(callback_proto) and isfunction(function) and
+            len(getfullargspec(callback_proto).args) > 0 and
+            len(getfullargspec(function).args) == 0):
+            # `callback_proto` has args but `function` does not, so a
+            # compatibility wrapper is needed.
+            self.callbacks[event].append(_compatibility_wrapper_for(function))
```
takluyver: Is there a good reason to implement this functionality within IPython rather than adding a dependency on backcall and using that? We can improve backcall if needed, but it's more flexible than this implementation - e.g. you can have a callback that gets only some of the arguments available.

fniephaus: I didn't know that adding `backcall` as a dependency is an option and that you'd prefer that. I just wanted to replicate the behavior we needed for this feature here without adding to much overhead. Will update the PR again when I have some time...

IPython/core/events.py
fniephaus: Opps, looks like I forgot to provide the result as an argument. Anyway, if you don't want the *pre* events that carry as result object, I can clean that up again. I just thought it would be useful, so user for example can "memorize" a result in a *pre* event and then match them in the corresponding *post* event (stupid example: pre event allocates memory for a computation, post event frees that corresponding memory allocation again if that makes sense).
Remove or keep the result argument?

takluyver: Oh, just using it as a handle to match up events? I guess so, but I find it hard to think of a real case where you'd want that. @minrk?

minrk: I would expect the `pre` events would take the requested code to be executed (basically the ExecuteRequest from the message perspective), i.e. a dict of `run_cell`'s kwargs. I've encountered several places where people want pre-execute notifiers of executions about to start.

For matching things up, since execution is ordered, there should always be one post_execute after each pre_execute, never two pre_executes in a row, so I'm not sure there is a use case for the pre_execute getting the incomplete result object. I, too, found it a little surprising to have a `result` object prior to the execution occurring.

fniephaus: Thanks for the comments. I've introduced `ExecuteRequest` which does what @minrk suggested. Additionally, I've added a reference to the corresponding `ExecuteRequest` to `ExecuteResult`s which allows to retrieve for example the cell code in *post* events as well (and also to match up events if this is actually a use case).

fniephaus:
`backcall` is now used as a dependency, so PTAL :)

minrk (approved):

minrk:
I'll give @takluyver the last word, but looks AOK to me. Thanks!

takluyver:
The API looks good, but I think the implementation could be simplified a little bit.

This maintains a global mapping of `(event, function)` pairs to the wrapped functions. Am I right in thinking that the only need for this is to allow unregistering functions again, by looking up the wrapped function to remove?

If so, I think we can do it without keeping a mapping by using the `.__wrapped__` attribute of the wrapper function, something like this:

```python
def unregister(self, event, function):
    if function in self.callbacks[event]:
        return self.callbacks[event](remove)

    for callback in self.callbacks[event]:
        if callback.__wrapped__ is function:
            return self.callbacks[event].remove(callback)

    raise ValueError("Function {!r} is not registered as a {} callback".format(function, event)
```

takluyver:
Also, a naming question. I think `ExecutionRequest` could cause confusion with an `execute_request` message - it might have come from such a message, but it doesn't directly represent one.

I'm not sure what a better name is, though. `ExecutionParams`?

takluyver:
Ohhh, also, I've just remembered something. We can add these parameters to `pre_run_cell` and `post_run_cell`, but we can't add them to `pre_execute` and `post_execute`; the latter events are also fired around the callbacks for comm messages, which happen without all the context of `run_cell`.

https://github.com/ipython/ipykernel/blob/71c44fe80e79126de4f27e6713fda1d274132d77/ipykernel/comm/comm.py#L152-L161

fniephaus:
re https://github.com/ipython/ipython/pull/10795#issuecomment-336415949: yes, you are right @takluyver. Will make that better...

re https://github.com/ipython/ipython/pull/10795#issuecomment-336416420: Oh right, I forgot about `execute_request`. `ExecutionParams` might be a little weird, because it would read `pre_run_cell(params)`, I would expect `pre_run_cell(raw_cell, silent, ...)`. On the other hand, the latter would be inconsistent with regard to the *post* event. What do you think about something more generic like `ExecutionInfo` or `ExecutionData`?

re https://github.com/ipython/ipython/pull/10795#issuecomment-336417456: I didn't know about `handle_msg`. Shouldn't there be `pre_handle_msg` and `post_handle_msg` which take the `msg` as an argument? Or shall we provide an `ExecutionRequest` with a reference to `msg` instead?

takluyver:
I'd be OK with `ExecutionInfo`, or maybe `ExecutionInput`. @minrk @Carreau - any brainwaves for the name of the type to store data about a call to `run_cell()`, for event handlers?

I think we'll just have to only add the data to `pre_run_cell` and `post_run_cell`. Changing the data for the `*_execute` events is going to be too messy now that they're called from two different projects with separate release schedules, and there's no good analogue of the `ExecutionResult` or `ExecutionRequest` (whatever that gets called) when we're handling a comm message.

fniephaus:
@takluyver great, so I've renamed the class to `ExecutionInfo`, removed it as an argument from the `*execute` events and refactored the code as you suggested. Is there anything else you want me to change?

takluyver:
Yes, I'm happy with this. Thanks for your patience with all the changes I asked for!

fniephaus:
Thanks for the comments!

PR #10797, opened 2017-09-13
Let keyboard interrupt stop custom dispatch of completer.

Carreau:
See #10733, interruption during custom completer can crash the
kernel. Technically we should likely even protect normal completion
(like jedi taking a while), but let's get something that fix an actual
bug.

This can lead to some inconsistencies in the frontend, as you interrupt
the kernel in Command mode, and interrupting the current custom
completer will lead to normal completion being (still) returned and the
completer poping up in command mode. It's not optimal but at least we do
not loose user state.

Alternative to #10792, I like to have a better, more general fix, but that will actually fix a case with crash where we do have bug reports, and can be backported to 5.x

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1715,6 +1715,12 @@ class IPCompleter(Completer):
                     return [r for r in res if r.lower().startswith(text_low)]
             except TryNext:
                 pass
+            except KeyboardInterrupt:
+                """
+                If custom completer take too long,
+                let keyboard interrupt abort and return nothing.
+                """
+                break
 
         return None
 


IPython/core/completer.py
```diff
@@ -1704,7 +1704,14 @@ def dispatch_custom_completer(self, text):
                  try_magic,
                  self.custom_completers.flat_matches(self.text_until_cursor)):
             try:
-                res = c(event)
+                try:
+                    res = c(event)
+                except KeyboardInterrupt:
```
takluyver: Does this need a nested try/except, or can we just add another except clause to the existing try/except?

Carreau: We can add an Except clause, but then it's hard to convey exactly where care about catching the keyboard interrupt. The only thing that could be slow is `c(event)`.But if you think that would be better I can move the clause.

takluyver: It feels neater to me to avoid another level of nesting, but I acknowledge there's also a benefit to your approach. I don't feel strongly about it, so let's leave it how you prefer it.

Carreau:
Does that make you happier ? I can also directly `return None`, instead of `break`

takluyver:
That seems neater to me. Let's see if meeseeks can do the backport. ;-)

Carreau:
Damn, having auto backport is good !

takluyver:
Nice work on the bot!

Carreau:
I've fixed the auto-backport this morning to have the correct PR title. Still a lot of things hardcoded :-)

PR #10809, opened 2017-09-15
Exit code 1 if non-interactive execution fails

### LINKED ISSUE #10807: non-zero exit status not returned when error happens with `-c`
This should return an exit status of 1 IIUIC

```bash
$ ipython -c "import blah"
Using matplotlib backend: Qt5Agg
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-253b2e59e51a> in <module>()
----> 1 import blah

ModuleNotFoundError: No module named 'blah'
$ echo $?
0
```

This works for the normal python interpreter

```bash
$ python -c "import blah"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'blah'
$ echo $?
1
```

and if you use a file instead of `-c`

```bash
$ echo "import blah" > test.py
$ ipython test.py 
Using matplotlib backend: Qt5Agg
ModuleNotFoundError                       Traceback (most recent call last)
~/Desktop/civis-jupyter-notebook/test.py in <module>()
----> 1 import blah

ModuleNotFoundError: No module named 'blah'
$ echo $?
1
```

takluyver:
Closes #10807.

This works for both `ipython -c "print(a)"` and `ipython error.py`.

I also did some cleanup while I was looking at the files to reduce the amount of warnings Pycharm shows me.

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -27,7 +27,7 @@ from IPython.core.profiledir import ProfileDir, ProfileDirError
 from IPython.paths import get_ipython_dir, get_ipython_package_dir
 from IPython.utils.path import ensure_dir_exists
 from traitlets import (
-    List, Unicode, Type, Bool, Dict, Set, Instance, Undefined,
+    List, Unicode, Type, Bool, Set, Instance, Undefined,
     default, observe,
 )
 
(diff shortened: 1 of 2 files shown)


Carreau (approved):
Do you want that in 6.2 ?

takluyver:
I think it's safe enough, but if you're doing 6.2 today, it's probably fine to leave this for 6.3.

Carreau:
Ok, 6.3 it is then.
We can also target 5.6 maybe, for consistency, even if i'm not thrilled with backports.

Issue #10812, opened 2017-09-16
Progress bar iteration


henryiii:
The progress bar added in the latest versions is really nice, but a bit clunky to use, especially interactively. Iteration would be a trivial and welcome addition. For example, the current code:

```python
bar = ProgressBar(200)
bar.display()
for i in range(200):
    bar.progress=i
    sleep(.01)
bar.progress=200
```

Could become:

```python
for i in ProgressBar(200):
    sleep(.01)
```

An example of how this could be done:

```python
    def __iter__(self):
        self.display()
        return self
    def __next__(self):
        progress = self.progress
        self.progress += 1
        if self.progress < self.total:
            return progress
        else:
            raise StopIteration()
```

If there's interest, I could make a PR.

Carreau:
I would try to get inspiration from [tqdm](https://pypi.python.org/pypi/tqdm), one of the difficulties is that you rarely loop on a range, you look on another iterable, so you need to take that into account. Then this iterable need to provide a length., etc. As we can't change API as often as we'd like, we need to be careful. 

I'm happy also providing also a minimal implementation as well, and tell users that if they need something more complex, they should see tqdm (or other); I prefer having bare bone where you need to switch to external dependency soon, than let user build complex program where switching becomes hard. 

Also we should mention tqdm in the docs for progressbar.

But +1 to make thing slightly more general, but not too much either.

cc @rgbkrk

henryiii:
I've previously worked on [Plumbum's progress bar](http://plumbum.readthedocs.io/en/latest/cli.html#terminal-utilities), which provides a `.range` and a `.wrap` facility. Since IPython's progress bar implementation is simple, this seemed like a simple addition inline with the way one might expect it to work (it already has a size).

For wrapping, how about a class method (just an example):
```python
@classmethod
def wrap(cls, iterable, length=None)
    if length is None:
        length = len(iterable)
    self = cls(length)
    self._iter = iterable # add to __iter__ method
    return self
```

Or, the constructor could try `iter()` on the argument, and then treat it like `total` if that throws an exception (int).

takluyver:
Should we encourage people to use the progress bar API from a popular library like tqdm or [progressbar2](https://pypi.python.org/pypi/progressbar2), and just write an IPython backend for it? It seems like we're reinventing the wheel a bit.

henryiii:
[...]

PR #10813, opened 2017-09-18
Adding iteration to ProgressBar to make it a more useful utility

henryiii:
As discussed in #10812, this simply adds the ability to iterate over a ProgessBar (no other API changes). Added a test and a mention in the relevant example.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -774,6 +774,19 @@ class ProgressBar(DisplayObject):
         self._progress = value
         self.update()
 
+    def __iter__(self):
+        self.display()
+        self._progress = -1 # First iteration is 0
+        return self
+
+    def __next__(self):
+        """Returns current value and increments display by one."""
+        self.progress += 1
+        if self.progress < self.total:
+            return self.progress
+        else:
+            raise StopIteration()
+
 class JSON(DisplayObject):
     """JSON expects a JSON-able dict or list
 
--- a/IPython/core/tests/test_display.py
+++ b/IPython/core/tests/test_display.py
@@ -196,10 +196,17 @@ def test_displayobject_repr():
 
 def test_progress():
     p = display.ProgressBar(10)
-    nt.assert_true('0/10' in repr(p))
+    nt.assert_in('0/10',repr(p))
     p.html_width = '100%'
     p.progress = 5
     nt.assert_equal(p._repr_html_(), "<progress style='width:100%' max='10' value='5'></progress>")
+
+def test_progress_iter(capsys):
+    for i in display.ProgressBar(5):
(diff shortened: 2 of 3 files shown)


IPython/core/display.py
```diff
@@ -774,6 +774,19 @@ def progress(self, value):
...
+    def __iter__(self):
+        self.display()
+        self._progress = -1 # First iteration is 0
+        return self
+
+    def __next__(self):
+        """Returns current value and increments display by one."""
+        self.progress += 1
+        if self.progress < self.total:
+            return self.progress
```
mariusvniekerk: Increment afterwards, that way you don't need to initialize to - 1.

henryiii: This is just a function, not a generator, so anything after the return will not run. You have to either have a -1 initial value (which is never printed), or you have to add a new variable to tell the difference between before-iteration 0 and iteration 0. You also really want the final iteration to be 4/5, to match the range, and to provide a visual indication that it's finished by changing to 5/5.

IPython/core/tests/test_display.py
```diff
@@ -196,11 +196,18 @@ def test_displayobject_repr():
...
     p.progress = 5
     nt.assert_equal(p._repr_html_(), "<progress style='width:100%' max='10' value='5'></progress>")
 
+def test_progress_iter(capsys):
+    for i in display.ProgressBar(5):
+        out, err = capsys.readouterr()
+        nt.assert_in('{0}/5'.format(i), out)
+    out, err = capsys.readouterr()
+    nt.assert_in('5/5', out)
+
```
henryiii: Oops, I guess this is nose running nose tests, not pytest running nose tests.

The nose project hasn't receive a commit in about a year, I think pytest would be a good idea :) I'll fix this.

henryiii:
Is there a way to test in a normal terminal environment? It seems that the tests run in a environment where display is not supported.

takluyver:
I don't think it's the environment that's involved there: it looks like `CapturingDisplayPublisher` (which is used by `capture_output`) hasn't been updated with the `transient=` keyword argument that the regular `DisplayPublisher` has. I guess you'll need to update that, or use `capture_output(display=False)` and let it capture the output on stdout.

henryiii:
I've used the `display=False` to make sure this works for IPython 5 or 6, and separately patched the display issue for IPython 5 and 6.

takluyver:
@mariusvniekerk @Carreau are you happy with this?

mariusvniekerk (approved):
Looks good

Carreau:
+1, I'm gessing backport to have identical functionality on Python2.

henryiii:
Yes, it should be in both.

PR #10818, opened 2017-09-20
Change some if -> elif in display Image code

takluyver:
I just happened to notice this. I'm pretty sure they were meant to be elifs; otherwise the value set for jpeg/png images is immediately overridden by the `else` clause below.


takluyver:
Done backport manually in 5f1cd28

PR #10830, opened 2017-09-25
Document the process of having a reproducible build.

Carreau:
Next step should be to enforce the building with a set SOURCE_DATE_EPOCH
value.

Right now I'm extra careful in the process as I haven't tried it,
but we should make it a bit more strait forward.

-- 
cc @takluyver because you have experience with flit and were IIRC involved in the Python discussions, and @mpacer because I'm sure you'll want to do the same with Nbconvert. 

When we make our first reproducible build, I'd like to have a blog post about that. Thoughts?
Wondering if we should also set the same timestamp in the cond-forge recipes.

### DIFF
--- a/docs/source/coredev/index.rst
+++ b/docs/source/coredev/index.rst
@@ -105,6 +105,13 @@ for the release you are actually making::
     VERSION=5.0.0
     BRANCH=master
 
+For `reproducibility of builds <https://reproducible-builds.org/specs/source-date-epoch/>`_,
+we recommend setting ``SOURCE_DATE_EPOCH`` prior to running the build; record the used value
+of ``SOURCE_DATE_EPOCH`` as it may not be available from build artifact. You
+should be able to use ``date +%s`` to get a formatted timestamp::
+
+    SOURCE_DATE_EPOCH=$(date +%s)
+
 
 2. Create GitHub stats and finish release note
 ----------------------------------------------
@@ -215,8 +222,11 @@ We encourage creating a test build of the docs as well.
 
 Commit the changes to release.py::
 
-    git commit -am "release $VERSION"
+    git commit -am "release $VERSION\n\nReleased with SOURCE_DATE_EPOCH=$SOURCE_DATE_EPOCH"
     git push origin $BRANCH
+
+Check that the commit message have the actual interpolated ``SOURCE_DATE_EPOCH``
+value.
 
 Create and push the tag::
 
@@ -246,6 +256,16 @@ folder. Be sure to test the ``wheels``  and the ``sdist`` locally before
 uploading them to PyPI. We do not use an universal wheel as each wheel
 installs an ``ipython2`` or ``ipython3`` script, depending on the version of
 Python it is built for. Using an universal wheel would prevent this.
+
+Check the shasum of files with::
+
+    shasum -a 256 dist/*
+
+and takes notes of them you might need them to update the conda-forge recipes.
+Rerun the command and check the hash have not changed::
+
+    ./tools/release
+    shasum -a 256 dist/*
 
 Use the following to actually upload the result of the build::
 
--- a/tools/release
+++ b/tools/release
@@ -79,12 +79,9 @@ else:
     sh('mv ipython-*.tgz %s' % ipbackupdir)
 
     # Build release files
-    sh('./build_release %s' % ipdir)
+    sh('./build_release')
 
     cd(ipdir)
-
-    # Upload all files
-    sh(sdists)
 
     buildwheels()
     print("`./release upload` to upload source distribution on PyPI and ipython archive")
--- a/tools/retar.py
+++ b/tools/retar.py
@@ -1,0 +1,54 @@
+"""
+Un-targz and retargz a targz file to ensure reproducible build. 
+
+usage: 
+
+    $ export SOURCE_DATE_EPOCH=$(date +%s)
+    ...
+    $ python retar.py <tarfile.gz>
+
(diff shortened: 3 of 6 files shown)


takluyver:
Have we checked that this actually works? Sdists, wheels, or both? What are the minimum versions of setuptools, wheel, or any other packages necessary?

Carreau:
> Have we checked that this actually works? Sdists, wheels, or both? What are the minimum versions of setuptools, wheel, or any other packages necessary?

I haven't checked the exact minimum versions; but I had it to work here yesterday; now I might likely have overlooked something as indeed it does not work anymore if I just follow the instructions.

I'm worried of enforcing these things at build time; as as we know; pip builds a wheel, and I don't want that to fail on user machines if they don't have the right versions.

takluyver:
I don't think we should enforce it at build time. Reproducible builds are not something everyone requires.

Carreau:
Hum, actually there is a bug in `wheel`...

Carreau:
> Hum, actually there is a bug in wheel...

Maybe not....

Carreau:
Ok, I've fixed it for wheels; now I need to figured it out for tgz.

Carreau:
@takluyver would you care for a re-review ; and if possible trying the instructions ?

tools/release
```diff
@@ -79,13 +79,10 @@ else:
...
 
     # Build release files
-    sh('./build_release %s' % ipdir)
+    sh('./build_release')
 
     cd(ipdir)
 
-    # Upload all files
-    sh(sdists)
-
```
Carreau: this was actually re-building the sdist

docs/source/coredev/index.rst
```diff
@@ -105,6 +105,13 @@ for the release you are actually making::
     VERSION=5.0.0
     BRANCH=master
 
+For `reproducibility of builds <https://reproducible-builds.org/specs/source-date-epoch/>`_,
+we recommend setting ``SOURCE_DATE_EPOCH`` prior to running the build; record the used value
+of ``SOURCE_DATE_EPOCH`` as it may not be available from build artifact. You
```
takluyver: This actually gives me an idea for flit; when SDE is set, it could record somewhere in the resulting file:

> Built with SOURCE_DATE_EPOCH=123 and flit version 0.12. See https://reproducible-builds.org/ for more info.

docs/source/coredev/index.rst
```diff
@@ -247,6 +257,16 @@ uploading them to PyPI. We do not use an universal wheel as each wheel
 installs an ``ipython2`` or ``ipython3`` script, depending on the version of
 Python it is built for. Using an universal wheel would prevent this.
 
+Check the shasum of files with::
+
+    shasum -a 256 dist/*
+
+and takes notes of them you might need them to update the conda-forge recipes.
+Rerun the command and check the hash have not changed::
```
takluyver: I think this paragraph should refer back to the earlier one.

tools/retar.py
```diff
@@ -0,0 +1,54 @@
+"""
+Un-targz and retargz a targz file to ensure reproducible build.
```
takluyver: Wait, we have to do this ourselves? I thought when you opened this PR that setuptools had gained the ability to make properly reproducible builds itself. I like the idea of reproducible builds, but if it's something we need custom tooling for, I wouldn't bother. I want our release process to be less special, not more special.

IPython doesn't have a particular need for reproducible builds beyond what packages in general have. In fact, it's arguably less important for pure Python packages, because you can unpack them and diff the source tree against that from the git tag.

takluyver (commented):
I didn't realise this still required custom code. As detailed inline, I'd favour leaving it until the packaging tools can do it for us without a special script.

tools/retar.py
Carreau: I though too as well; but it's actually just for Sdist. 
We can wait; but that will require changes in core CPython distutils, gzip and tarfile module.
Basically tar does not pass kwargs as options to gzip

So if we want to have that only using core python then we need to wait for 3.7; maybe 3.8

I'm happy to not get that in; but at least this give us a minimal idea of what is missing in core Python; I can try to start a thread on distutils-sig if you like; but with the usual bikesheding...

docs/source/coredev/index.rst
Carreau: I was thinking about that as well, in particular in IPython we have a  file that record the built commit; we could also record the timestamp.

tools/retar.py
takluyver: I had a look at this for flit at some point - I think it's doable without changes to the stdlib, if  the packaging tool (flit/setuptools) duplicates a few lines of code from `tarfile` so that it creates the gzip object itself.

Carreau:
https://pypi.python.org/pypi/distutils644 is relevant as well.

jzf2101:
@Carreau just to be clear, these are builds would be reproduced by the developer locally?

takluyver:
Reproducible builds are a pretty new idea, but as I understand it, the idea is that a third party can verify that a given package is the accurate result of building from a given git checkout. This is useful for security - it makes it harder to slip bad code in unnoticed when building a release.

I wouldn't expect typical developers installing IPython to care about reproducing the build themselves.

As I think more about it, I suspect that reproducible builds are only really useful once there's standard metadata to indicate how to reproduce a package, so that they can be reproduced and checked automatically.

Carreau:

Carreau:
see https://bugs.python.org/issue38632

PR #10831, opened 2017-09-25
Protect release script against non dist files.

Carreau:
On some OSes, for example OS X / macOS, if the dist/ folder is opened in
the GUI while during the build process, hidden files will be created,
and thus attempted to be uploaded.

This protect about that by restricting upload to tgz and whl files.

### DIFF
--- a/tools/release
+++ b/tools/release
@@ -6,6 +6,7 @@ This should ONLY be run at real release time.
 from __future__ import print_function
 
 import os
+from glob import glob
 from subprocess import call
 import sys
 
@@ -49,9 +50,10 @@ cd(tooldir)
 
 if 'upload' in sys.argv:
     cd(distdir)
-    #print( 'Uploading distribution files to GitHub...')
 
-    for fname in os.listdir('.'):
+    # do not upload OS specific files like .DS_Store
+    to_upload = glob('*.whl')+glob('*.tgz')
+    for fname in to_upload:
         # TODO: update to GitHub releases API
         continue
         print('uploading %s to GitHub' % fname)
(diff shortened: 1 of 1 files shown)


tools/release
```diff
@@ -49,9 +50,10 @@ cd(tooldir)
 
 if 'upload' in sys.argv:
     cd(distdir)
-    #print( 'Uploading distribution files to GitHub...')
 
-    for fname in os.listdir('.'):
+    # do not upload OS specific files like .DS_Store
+    to_upload = glob('*.whl')+glob('*.tgz')
```
takluyver: I think it should be `*.tar.gz` rather than `*.tgz`.

Carreau: The rest of the script have things like `sh('scp `ls -1tr *tgz | tail -1` %s' % backup_site)` does that mean it's wrong as well ?

takluyver: The tarballs that get uploaded to PyPI are `.tar.gz`. The script also makes a 'backup' tarball using `git archive` (this is done in `make_tarball.py`). That's what has the `.tgz` extension. It gets uploaded to http://archive.ipython.org/release/

I think the 'backup' files are a remnant from the distant past before IPython was in proper version control, to preserve a copy of the code as of that release. I'd like to make our release process simpler and more standard - we trust the combination of Github and PyPI to preserve releases of all our other projects. But more backups always seem like a good idea, and it's mostly automated (once you have the necessary SSH key...), so I've not pushed for removing it.

takluyver: @Carreau ping?

takluyver:
@Carreau I've tweaked the extension as I think it needs to be - are you OK with that?

Carreau:
Yes, sorry I've been busy with lots of other stuff. Thanks for taking care of this. 

+1

takluyver:
No problem :-)

Issue #10161, opened 2017-01-13
When do we want to drop Python 3.3 support?


takluyver:
Now that we're dropping 2.7 support, I've started to notice the points where supporting Python 3.3 requires extra work. For instance, 3.4 introduced `pathlib` and `unittest.mock` into the standard library, so we have a couple of dependencies which are only needed on 3.3.

It also means tests take longer, especially on Appveyor which only runs one build at a time.

Python 3.4 was released in March 2014, so it's not yet 2 years old. So I don't think we should drop 3.3 yet, but I wanted to open the discussion.

willingc:
Perhaps match the end-of-life date that is published by CPython: 2017-09-29

Currently, CPython releases are typically end-of-life at 5 years from original release date.
[Reference](https://docs.python.org/devguide/#status-of-python-branches)

takluyver:
That sounds like a reasonable date. Thanks!

Carreau:
Agreed, that's seam reasonable. 

We have 2 ways of seeing that (in general):
  - Stop support slightly after, cause their might still be bugfixes, and we usually release version with  support for X.y after X.y got released, so all is delayed. 
  - Stop supporting a bit before to give a signal to Python core that they can safely stop support as no new release will target X.y.

Also I don't see any reason, nor think we should stop support in between minor revisions. 

Also I'm going to guess this is prompted by #10159. I did not meant to drop support for 3.3 in it, I'll put that on the jetlag.

minrk:
👍 for EOL + ϵ as the time to start considering it. I'd like to keep the bar for dropping any kind of support as "maintaining support for X is causing a problem".

Carreau:
> for EOL + ϵ as the time to start considering it. I'd like to keep the bar for dropping any kind of support as "maintaining support for X is causing a problem".

I think that's also a question of what do we call "support" we can perfectly go through:
- Not officially supported in docs
- Not tested
- No active bugfixes, but PR welcome
- actively incompatible.

Also "is causing a problem" can be quite subjective :-)

willingc:
Looks like numpy may have already dropped 3.3 support [Travis build on recent IPython doc PR](https://travis-ci.org/ipython/ipython/jobs/192415150) cc/@takluyver

Carreau:
> Looks like numpy may have already dropped 3.3 support Travis build on recent IPython doc PR  cc/@takluyver numpy changelog

[...]

PR #10833, opened 2017-09-26
Drop support for Python 3.3

takluyver:
This makes 3.4 the minimum required version of Python. See discussion on #10161.

### DIFF
--- a/IPython/__init__.py
+++ b/IPython/__init__.py
@@ -30,9 +30,9 @@ import sys
 if sys.version_info < (3,4):
     raise ImportError(
 """
-IPython 6.3+ supports Python 3.4 and above.
+IPython 7.0+ supports Python 3.4 and above.
 When using Python 2.7, please install IPython 5.x LTS Long Term Support version.
-Python 3.3 was supported up to IPython 6.2.
+Python 3.3 was supported up to IPython 6.x.
 
 See IPython `README.rst` file for more information:
 
--- a/IPython/__init__.py
+++ b/IPython/__init__.py
@@ -27,12 +27,12 @@ import sys
 #-----------------------------------------------------------------------------
 
 # Don't forget to also update setup.py when this changes!
-if sys.version_info < (3,3):
+if sys.version_info < (3,4):
     raise ImportError(
 """
-IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.
+IPython 6.3+ supports Python 3.4 and above.
 When using Python 2.7, please install IPython 5.x LTS Long Term Support version.
-Beginning with IPython 6.0, Python 3.3 and above is required.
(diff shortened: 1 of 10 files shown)


IPython/__init__.py
```diff
@@ -27,12 +27,12 @@
 #-----------------------------------------------------------------------------
 
 # Don't forget to also update setup.py when this changes!
-if sys.version_info < (3,3):
+if sys.version_info < (3,4):
     raise ImportError(
 """
-IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.
+IPython 6.3+ supports Python 3.4 and above.
```
hugovk: There's some similar text to update in setup.py

https://github.com/takluyver/ipython/blob/68559994a2c1efcbd408fa63adaf1867389e22df/setup.py#L47

takluyver: Thanks, updated.

willingc (approved):

Carreau:
I'm Ok to drop 3.3 for 6.3; we just have to be careful, I don't think we've ever dropped a Python version a minor release though.

willingc:
I think that we can provide guidance to pre-6.3 for anyone on Python 3.3.

takluyver:
I'm also fine with leaving it until we split 6.x from master, and targetting it for 7.0. I don't think there's any hurry to do this - I don't know of anything we're eagerly waiting to do that's only possible from 3.4.

Carreau:
I'm happy to wait a bit and see how much we are planning to keep the 6.x branch. We can fork master toward 7.0; though that might mean having a 5.x 6.x and master branch.

takluyver:
I wouldn't start 7.0 for this; it's a simple set of changes, it's not worth having to backport everything for. Let's just let it sit until we have another breaking change that gives us a reason to aim for a 7.0 release.

hugovk (approved):

jzf2101:
See #10911

takluyver:
I think it's fine to stop testing with 3.3 in 6.x releases, and then make the changes that we know will break it for 7.0.

Carreau:
> I think it's fine to stop testing with 3.3 in 6.x releases, and then make the changes that we know will break it for 7.0.

Ok, fair enough. This need rebase now. I think a nov/dec release would be welcome. Do we wait for Dec25 ?

takluyver:
I've rebased the changes and updated the version numbers in various docs and messages to say that 7.0 (not 6.3) drops support for Python 3.3.

Carreau:
@takluyver  do you want to merge this, so that we start master being a 7.0-dev ?

Carreau (approved):

takluyver:
That's my plan. I've already updated the version number in master.

Carreau (approved):

PR #10834, opened 2017-09-26
Add Julia to default script magics.

Ismael-VC:
I have been adding this line manually to my local copies of IPython/Jupyter and even suggest to do so in a Stack Overflow question, it would be nice if this worked by default.

https://stackoverflow.com/questions/24091373/best-way-to-run-julia-code-in-an-ipython-notebook-or-python-code-in-an-ijulia-n/24106944#24106944

Should I also add R or any other language? Is this not the correct way to do it? Please let me know, cheers!

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -89,6 +89,7 @@ class ScriptMagics(Magics):
             'python2',
             'python3',
             'pypy',
+            'julia',
         ]
         if os.name == 'nt':
             defaults.extend([


Ismael-VC:
I'm looking into how to test this with CI.

takluyver:
Thanks! I don't think it's worth getting Julia installed in CI to test this - there are already tests for the script magics machinery. I take it you've run this manually and it works? I.e. Julia accepts code from a pipe connected to stdin?

Ismael-VC:
Yes, I've tested this manually and it works, Julia does accept code from a pip connected to stdin:

```julia
C:\Users\Ismael                                   
λ echo foo | julia -e "write(STDOUT, read(STDIN))"
foo                                               
                                                  
C:\Users\Ismael                                   
λ                                                 
```

jupy

It currently prints each line of code though, not just the last one.

takluyver:
Thanks!

Naereen:
Hi,
I think we should use this opportunity to add more in the scripts alias, maybe ruby, perl etc
Any suggestions?

takluyver:
ruby & perl are already there ;-)

Carreau:
I'm unsure we want julia by default; julia is not a language that is installed that much compared to other (like perl, ruby that are basically already installed on all linux machines) so if user were to see the julia magic and it would not work that would be confusing. 

That would raise confusion with the [official julia magic](https://pypi.python.org/pypi/julia) which has way may more capability than the script magic as it can shares memory with Python. 

Plus you can already use `%%script julia` which is only 7 more characters.

I'm going to open a PR to revert, and continue the discussion.

Naereen:
@Carreau that makes sense.

However, based on the (limited) documentation of the `julia` pypi package you linked, I cannot see that it gives a `%%julia` ipython magic after installation. Does it really?

Carreau:
Yes, and unlike the script magic it is way more capable.

Naereen:
Great, then I agree that adding julia to the default script magic is not useful;

Carreau:
If you are using both Julia and Python the you are likely a good candidate to maintain (and document) the pyjulia project; it seem there is not much love from the Julia developers. 

A bit of documentation show how to use the julia magic would be great ! It is under appreciated and not enough advertized.

Naereen:
I would love to, but I don't feel skilled enough in Julia, and I will probably don't have time to maintain the project correctly... Sorry.

Carreau:
> I would love to, but I don't feel skilled enough in Julia, and I will probably don't have time to maintain the project correctly... Sorry.

No problem; I think you are skilled enough; it works, it's mostly a question of write some docs and advertising. 

Anyway, thanks for the PRs !

Ismael-VC:
Thank you guys for the discussion, I'll update my Stack Overflow answer with `%%script julia`, etc. and check out how the pyjulia package has been going.

I thought the pyjulia magic was dead, since there is no mention of it in the README anymore, but now I see the file:

https://github.com/JuliaPy/pyjulia/blob/master/julia/magic.py

Which is indeed better. I'll test it out and make some docs in pyjulia.

Issue #10814, opened 2017-09-18
IPython completer crashes when invoked at the start of a line


ian-r-rose:
Most frontends don't invoke the completer at the start of a new line, so this doesn't come up much, but I came across it when tracking down a different bug.

If the completer is invoked at the start of a new line, it crashes. It seems that this is due to it identifying the previous line as the one for which it needs completions, rather than the current (empty) line. The trouble looks like it is in `IPython.utils.tokenutil.line_at_cursor`.

The bug may be reproduced by running:
```python
import IPython
c = IPython.get_ipython().Completer
txt = 'One\nTwo\n'

c.complete(text=None, line_buffer=txt, cursor_pos=4)
```

Edit: as the example shows, it doesn't need to be an empty line, just the start of a line.

takluyver:
Yup, I think you're right. It's counting the newline character as part of the preceding line.

takluyver:
I made #10835 to fix this, but it doesn't fix the example script you gave, because the completer code doesn't use `line_at_cursor` at all. That is done in ipykernel. Is there an easy way to trigger this for manual testing?

ian-r-rose:
You can also test this using JupyterLab 0.27.0, which was mistakenly invoking the completer at the start of the line.

PR #10835, opened 2017-09-27
Identify position after a newline as the start of the next line

takluyver:
Closes gh-10814

### DIFF
--- a/IPython/utils/tests/test_tokenutil.py
+++ b/IPython/utils/tests/test_tokenutil.py
@@ -108,6 +108,12 @@ def test_line_at_cursor():
     assert line == "", ("Expected '', got %r" % line)
     assert offset == 0, ("Expected '', got %r" % line)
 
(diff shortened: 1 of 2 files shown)


ian-r-rose:
This fixes the IPython `ValueError`, but it breaks completion on the last line of a cell (due to not correctly identifying the right line).
If I have a cell with contents:
```python
pri
pri
```
invoking the completer on the first line successfully gives me `print`, but on the second line it does not.
(In case you missed it on #10814, you can test this in JLab 0.27, which was mistakenly invoking the completer at the start of lines).

takluyver:
How about now?

ian-r-rose:
Works great, thanks for chasing that down @takluyver !

ian-r-rose (approved):

Carreau:
Thanks ! Awesome !

PR #10836, opened 2017-09-27
Revert "Add Julia to default script magics."

Carreau:
Reverts ipython/ipython#10834

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -89,7 +89,6 @@ class ScriptMagics(Magics):
             'python2',
             'python3',
             'pypy',
-            'julia',
         ]
         if os.name == 'nt':
             defaults.extend([


takluyver:
Hmm, I thought it only set up those magics if the command was found, i.e. `which('julia')`. But it looks like I was wrong; I can't see any code that does that.

I don't feel strongly about this, in any case. I guess another reason to revert it is that it might interfere with the more powerful julia magic provided by another package.

Carreau:
> I guess another reason to revert it is that it might interfere with the more powerful julia magic provided by another package.

That's the main thing that I'm afraid of; there is already a wide use of  the julia magic provided by pyjulia; and there is precedent of us removing magics (like R) form the core because they were in separate package.

takluyver:
Merging since further discussion on #10834 seems to have led to agreement that it wasn't the way to go. Thanks Matthias!

Carreau:
Thanks !

PR #10837, opened 2017-09-27
Fixing display publisher (Python 3)

henryiii:
This fixes the CapturingDisplayPublisher by matching the signature with the regular DisplayPublisher (bug surfaced in #10813).

### DIFF
--- a/IPython/utils/capture.py
+++ b/IPython/utils/capture.py
@@ -14,14 +14,17 @@ from io import StringIO
 
 
 class RichOutput(object):
-    def __init__(self, data=None, metadata=None):
+    def __init__(self, data=None, metadata=None, transient=None, update=False):
         self.data = data or {}
         self.metadata = metadata or {}
-    
+        self.transient = transient or {}
+        self.update = update
+
     def display(self):
         from IPython.display import publish_display_data
-        publish_display_data(data=self.data, metadata=self.metadata)
-    
+        publish_display_data(data=self.data, metadata=self.metadata,
+                             transient=self.transient, update=self.update)
+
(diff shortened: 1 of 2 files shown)


takluyver:
I suspect that `capture_output()` should also record the data passed as `transient=` and `update=`, rather than accepting them and discarding them. @minrk? In any case, making the signature valid is already an improvement over throwing an error.

minrk:
@takluyver exactly right

takluyver:
@henryiii do you want to have a go at that as part of this PR, now that it's no longer required for #10813? If you're not sure about how to do it, we can also just merge this and open an issue for storing the extra fields.

henryiii:
Is there anywhere the captured info is processed (other than stderr and stdout)? It looks to be simple unless something uses the outputs and assumes it'a a two element list.

takluyver:
As far as I can see, it's only published in one place, by `IPython.utils.capture.CapturedIO`.

IPython/utils/capture.py
```diff
@@ -14,14 +14,17 @@
 
 
 class RichOutput(object):
-    def __init__(self, data=None, metadata=None):
+    def __init__(self, data=None, metadata=None, transient=None, update=False):
         self.data = data or {}
         self.metadata = metadata or {}
-    
+        self.transient = metadata or {}
```
takluyver: `transient or  {}`

takluyver:
Other than the one note I made, this looks good to me.

IPython/utils/capture.py
henryiii: Fixed

PR #10838, opened 2017-09-27
Fixing display publisher (Python 2)

henryiii:
This fixes the CapturingDisplayPublisher by matching the signature with the regular DisplayPublisher (bug surfaced in #10813).


takluyver:
By the way, you don't generally need to make two pull requests - we have a bot (made by @Carreau ) which can backport a pull request from master to 5.x once it's merged. I don't know if it will work in this case, though.

takluyver:
Ah, I see you've done the args differently for Py2. Thanks!

PR #10840, opened 2017-09-28
Always reset after `%run -i` in tests.

craigcitro:
Currently, `IPython/core/tests/run_tests.py` has a few tests that do `run -i`,
which has side-effects on the IPython environment shared by all tests. In
particular, setting `__file__` in the global namespace leads to traceback
handling in `ultratb` getting confused, since it no longer knows which file to
look in.

This fixes existing uses of `run -i`, and adds a note reminding future
travelers to be careful when adding new cases.

PTAL @Carreau @takluyver


takluyver:
Seems fine to me, thanks.

PR #10841, opened 2017-09-28
Silence tokenization errors in ultratb.

### LINKED ISSUE #6864: Exceptions thrown in cell with multiline strings
It seems that when an exception is raised from a cell with a multiline string, we get a code-red error.

Try:

```
raise Exception(r'''foo
''')
```

It will report:

```
ERROR: An unexpected error occurred while tokenizing input
The following traceback may be corrupted or invalid
The error message is: ('EOF in multi-line string', (1, 0))

<ipython-input-15-8d28bd905cfc> in <module>()
      1 raise Exception(r'''foo
----> 2 ''')

Exception: foo
```

The first three lines are worrying for the casual user.

craigcitro:
Currently, the ultratb module attempts to tokenize the code it finds in
tracebacks, and issues a scary-sounding warnings about potentially invalid
tracebacks if it fails. However, this tokenization is known to fail in some
cases, notably if the reported error line is the terminating line of a
multiline, triple-quoted string literal.

This change fixes the issue by switching this to a `logging.debug` message,
which silences it by default without completely removing it for debugging
purposes.

Fixes #6864.

**Note:** This is actually blocked on #10840, and currently includes a subset of that PR. I'll rebase and repush this once that's in, but the rest of the files are ready to go.

PTAL @takluyver @Carreau

### DIFF
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -2,6 +2,7 @@
 """Tests for IPython.core.ultratb
 """
 import io
+import logging
 import sys
 import os.path
 from textwrap import dedent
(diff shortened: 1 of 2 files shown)


craigcitro:
OK, rebased on top of #10840 and now this should be ready to go.

takluyver:
Thanks!

PR #10857, opened 2017-10-11
Remove broken downloads badge

hugovk:
It's showing:

image

See https://github.com/badges/shields/issues/716 for more info.


takluyver:
Thanks for noticing this and taking the time to deal with it!

PR #10858, opened 2017-10-11
Update version classifiers

hugovk:
Assumes https://github.com/ipython/ipython/pull/10833 (for https://github.com/ipython/ipython/issues/10161) will be merged and Python 3.4 is the minimum.

### DIFF
--- a/IPython/core/release.py
+++ b/IPython/core/release.py
@@ -114,5 +114,10 @@ classifiers = [
     'License :: OSI Approved :: BSD License',
     'Programming Language :: Python',
     'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: Only',
+    'Programming Language :: Python :: Implementation :: CPython',
     'Topic :: System :: Shells'
     ]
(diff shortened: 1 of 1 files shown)


takluyver:
I don't like using classifiers this specific: it's easy to forget to update them, and e.g. it will probably work with Python 3.7 when it comes out. Now that we have the [python_requires](https://github.com/ipython/ipython/blob/0dbfdb9c9e2eafa4a4ddaf76532cfae482a554b8/setup.py#L235) option to specify the supported version of Python properly, I don't think there's any need to use classifiers for this.

IPython is also not particularly specific to CPython; it should work on other implementations so long as the dependencies do.

hugovk:
Maybe just these then?
```
 +    'Programming Language :: Python :: Only',
 +    'Programming Language :: Python :: Implementation :: CPython',
```

takluyver:
`Programming Language :: Python :: Only` is not a [valid classifier](https://pypi.python.org/pypi?%3Aaction=list_classifiers). There's `Programming Language :: Python :: 3 :: Only`, which I'm OK with using but don't feel strongly about.

Naming CPython specifically implies that it's not meant to work with other Python interpreters, which is wrong. I don't know how well it does work with the others, but it's not meant to be CPython specific.

hugovk:
Oops, `Programming Language :: Python :: Only` was a typo!

takluyver:
Thanks

PR #10859, opened 2017-10-12
Don't list pygments styles on import

takluyver:
Listing the styles forces pygments to load plugins, which slows startup and may cause unexpected errors from other packages. We shouldn't be calling this on import.

The list of available styles will also depend on what packages are installed, so building our config docs with one list may be misleading. I've made the help string point to `pygmentize -L styles` to see available styles on your system.

Closes jupyter/notebook#2926


takluyver:
With this change, I see a reduction in cold-start time from ~11 seconds to ~6 seconds (this is on a spinning metal hard drive), and a reduction in warm start from ~1.1s to ~0.9s.

takluyver:
I'll merge this tomorrow unless someone says otherwise.

PR #10863, opened 2017-10-19
Add signature to display of functions in the shell

### LINKED ISSUE #9310: Define a better displayFormatter of function and methods
I think it should be possible to define a default formatter for functions that would be better than the defautl one. 

The following is not really useful:

```
In [1]: from inspect import signature
In [2]: signature
Out[2]: <function inspect.signature>
```

Something that would display less than `signature?` or `signature??`, but still a bit better than plain function/methods REPR. 

```
In [1]: from inspect import signature
In [2]: signature
Out[2]: <function inspect.signature(obj, *, follow_wrapped=True)>
```

Seem like a nice standalone project that could be useful. 

Thoughts ?

Madhu94:
Closes #9310

### DIFF
--- a/IPython/core/tests/test_oinspect.py
+++ b/IPython/core/tests/test_oinspect.py
@@ -305,8 +305,13 @@ def test_empty_property_has_no_source():
 
 
 def test_property_sources():
-    import zlib
-
+    import posixpath 
+    # A simple adder whose source and signature stays
+    # the same across Python distributions
+    def simple_add(a, b):
+        "Adds two numbers"
+        return a + b
+    
     class A(object):
         @property
         def foo(self):
@@ -314,18 +319,18 @@ def test_property_sources():
 
         foo = foo.setter(lambda self, v: setattr(self, 'bar', v))
 
-        id = property(id)
-        compress = property(zlib.compress)
+        dname = property(posixpath.dirname)
+        adder = property(simple_add) 
 
     i = inspector.info(A.foo, detail_level=1)
     nt.assert_in('def foo(self):', i['source'])
     nt.assert_in('lambda self, v:', i['source'])
 
-    i = inspector.info(A.id, detail_level=1)
-    nt.assert_in('fget = <function id>', i['source'])
-
-    i = inspector.info(A.compress, detail_level=1)
-    nt.assert_in('fget = <function zlib.compress>', i['source'])
+    i = inspector.info(A.dname, detail_level=1)
+    nt.assert_in('def dirname(p)', i['source'])
+    
+    i = inspector.info(A.adder, detail_level=1)
+    nt.assert_in('def simple_add(a, b)', i['source'])
 
 
 def test_property_docstring_is_in_info_for_detail_level_0():
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -88,7 +88,7 @@ from warnings import warn
 
 from IPython.utils.decorators import undoc
 from IPython.utils.py3compat import PYPY
-
+from IPython.utils.signatures import signature
 
 __all__ = ['pretty', 'pprint', 'PrettyPrinter', 'RepresentationPrinter',
     'for_type', 'for_type_by_name']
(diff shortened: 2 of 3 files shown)


takluyver:
This seems like a reasonable addition to me, but you'll need to update the test that's currently failing.

Madhu94:
Sure, I'm really sorry, next time I'll make sure to check the tests.

takluyver:
No problem, that's what CI is for ;-)

Madhu94:
@takluyver Do you have any suggestions for dealing with outputs that vary across Python distributions? 

The output of ``inspect.signature(zlib.compress)`` is different in Python versions 3.4/3.5 and 3.6. That's why the updated test case passes in only 3.5 (which is what I use)

For Python 3.4 and Python3.5
```  
>>> str(inspect.signature(zlib.compress))    
'(bytes, level=-1, /)'
```

Python 3.6
```  
>>> str(inspect.signature(zlib.compress))
'(data, /, level=-1)'
```

Likewise for Python versions < 3.5, ``inspect.signature(id)`` will throw a ``ValueError``.

takluyver:
A function defined in Python is more likely to have a reliable signature than a compiled one, I think. Try e.g. `posixpath.dirname`.

Madhu94:
By "compiled one", you mean a ``builtin`` writtin in C ?  Got it, thanks.

takluyver:
Yup, exactly.

takluyver:
Ping! Do you think you'll get a chance to fix the test soon?

Madhu94:
I had issues with my laptop and was offline for a while, will get to it
ASAP. It is a ridiculously small PR to be hanging around for so long.

takluyver:
pickle has compiled functions, so you're still running into the same problem. I'd suggest `posixpath.dirname` as the example function - that's pure Python, it only takes one argument, and it's so basic that it's unlikely to change in the future.

takluyver:
Or I guess you could define a function inside the test.

Madhu94:
Yep. This time I will test on 3.3, 3.4, 3.5, 3.6 before pushing...:-/

Madhu94:
@takluyver Are these tests fine ? I went with posixpath.dirname and my own function.

IPython/core/tests/test_oinspect.py
```diff
@@ -38,14 +40,14 @@
 # defined, if any code is inserted above, the following line will need to be
 # updated.  Do NOT insert any whitespace between the next line and the function
 # definition below.
-THIS_LINE_NUMBER = 41  # Put here the actual number of this line
+THIS_LINE_NUMBER = 43  # Put here the actual number of this line
```
takluyver: I think this and the bits above are the result of a mistaken merge with another PR that got merged - can you set them back so they're unchanged in this PR?

takluyver (commented):
Thanks, this is looking good for fixing the existing tests, but we should still add a test for the new functionality in IPython.lib.tests.test_pretty.

Madhu94:
I had a question about that - why is [this test](https://github.com/ipython/ipython/blob/master/IPython/lib/tests/test_pretty.py#L223) called ``test_**un**bound_method`` ?

takluyver:
```python
class MyClass:
    def amethod(self):
        pass

MyClass.amethod  # unbound method

instance = MyClass()
# bound method - it's bound to the instance, which becomes self when it is called
instance.amethod
```

In Python 2, unbound methods had their own type. In Python 3, they are just considered functions.

takluyver:
Closing and reopening to re-run tests

(For some reason I don't have access to rebuild on Appveyor - @Carreau if you have access, can you see a way to add me?)

Madhu94:
@takluyver It might have been better if I had gotten everything to work, and then asked you to merge - you did not have to check each commit. :)

takluyver:
Sorry! I saw this was almost there, and didn't know if you would be available to finish it quickly.

Madhu94:
I'll have a faster turnaround time on my next PR to ipython

takluyver:
It hasn't been a problem, I was just a bit eager. Anyway, the tests are passing now, so I'm going to merge this. Thanks for your contribution!

Issue #9145, opened 2016-01-19
[serious] LTS 4.x and drop 2.7 on 5.x ?


Carreau:
Well all is in the title.

I'm not sure IPython 4.x is missing a lot of features, we have some annoying bugs that have languished for years, and we more and more tend to [commit code that uses python 3 only features and fails on travis](https://github.com/ipython/ipython/pull/9144#issuecomment-172985454) 

So real question, is it time to start thinking about dropping 2.7 ? 

Obviously there are several level of "dropping support". I'll on purpose do not list change that will On purpose break things on Python 2.7:
   - Drop 2.7 from supported python, at least officially, by which I mean, we remove mention of 2.7 compatibility, but still have unit-test... etc 
   - allow-failure on 2.7, not actively working on fixing 2.7 bugs accept PRs.
   - not release 2.x wheels.
   - marking 5.0 as `requires-python >=3.3` (still leave ability to install from source manually).

I know that people are still attached to 2.7 so I propose to make 4.x "LTS", in the sens of still accepting PRs that fixes things and release at least regularly.

takluyver:
Nah, not for IPython. 5.x is too soon, and I don't want to be stuck supporting the readline stuff that we'll hopefully be dropping from the codebase in 5 or 6.

I think we should consider dropping Python 2 support in the notebook server for maybe version 5 or 6 of that - it would remain possible to run Python 2 notebooks, but you'd have to have Python 3 for the server itself. Once we've done that, we can start to think about dropping Python 2 further down the stack.

ivanov:
Hey gang - doing the standard issues cleanup - closing this one since we've dropped support for 2.7 in master.

:cake: and happy hacking! :bowtie:

PR #10865, opened 2017-10-21
Remove redundant Python 2 code

hugovk:
IPython 6 dropped support for Python 2.7, 3.0, 3.1 and 3.2 (eg. https://github.com/ipython/ipython/issues/9145, https://github.com/ipython/ipython/pull/9900, https://github.com/ipython/ipython/pull/10131).

This removes a lot of old Python 2 code still in the codebase.

TODO:
* [x] Restore trailing whitespace
* [x] Restore `cast_unicode`
* [x] Resolve merge conflict
* [x] Restore `clr` import
* [x] Restore `arg_split` import
* [x] Keep `cast_bytes` in `py3compat`
* [x] Restore all non-Python 2 stuff in `py3compat`
* [x] Add deprecation docstring to `py3compat`


takluyver:
I haven't checked in detail, but I'm not 100% comfortable removing all those `cast_unicode` calls. We have a `cast_unicode_py2` function for when we *know* it's redundant on Python 3, so if we're using `cast_unicode` instead, there's a chance that that is necessary even on Python 3.

Also, as with your PR to my fork, can we ask you to do it with your editor configured to not strip whitespace? It can cause merge conflicts and problems with `git blame`.

hugovk:
Sure, coming up.

hugovk:
@takluyver This PR updated.

.editorconfig
```diff
@@ -0,0 +1,19 @@
+# Top-most EditorConfig file
```
takluyver: @Carreau @gnestor thoughts on adding this file? It seems reasonable to me, but I know people sometimes take issue with ever more config files in the repo root directory.

IPython/utils/_process_cli.py
```diff
@@ -12,23 +12,21 @@
 """
 
 # Import cli libraries:
-import clr
```
takluyver: I think this might be a special thing that needs to be imported first for IronPython to work. Best to leave this module alone, I think.

IPython/utils/_process_cli.py
```diff
@@ -12,23 +12,21 @@
...
 import System
 
 # Import Python libraries:
 import os
 
 # Import IPython libraries:
 from IPython.utils import py3compat
-from ._process_common import arg_split
```
takluyver: This is there to be re-exposed by the module. Again, let's leave this as is.

IPython/utils/py3compat.py
```diff
@@ -27,310 +22,84 @@ def cast_unicode(s, encoding=None):
         return decode(s, encoding)
     return s
 
-def cast_bytes(s, encoding=None):
-    if not isinstance(s, bytes):
-        return encode(s, encoding)
-    return s
```
takluyver: I'm a bit wary of removing things from `py3compat` in case other modules like `ipykernel` rely on them.

takluyver:
Thanks, this mostly looks good now. A few comments inline.

IPython/utils/py3compat.py
hugovk: Sure, I'll make these changes. Seeing as Python 3 is now the only supported version, how about somehow marking `py3compat` as deprecated?

takluyver:
Yeah, I think it could have a 'deprecated' note in the docstring now, and then in the future when we think we've got rid of all uses of it, add a warning on import.

For now,  let's leave all the functions exposed by `py3compat` that were there before (not just the one my previous comment was on the line of) :-)

Carreau:
First thing I want to say is  I'm worried about finishing up cleaning py2 compat code; this has been painful before for automatic backport of bug fixes, and lead to subtle bugs. We've stop removing things on purpose some time ago. 

I'll look at it in more details.

.editorconfig
Carreau: I'm ok adding it, but I don't know of a lot of people using that. It may likely just rot as the tox.ini file which is never used and still test on Python 27 as shown further in this PR.

.gitignore
```diff
@@ -21,5 +21,6 @@ __pycache__
 .DS_Store
 \#*#
 .#*
+.cache
```
Carreau: Should that also be in your system global `.gitignore` if you feel the need to add it ?

IPython/utils/py3compat.py
```diff
@@ -1,22 +1,17 @@
 # coding: utf-8
```
Carreau: _if_ we remove the compatibility layer python2/python3, should this file still exists ? The way I see it either this file is / or is not. If there are function with use often they probably should be in utils. @takluyver thoughts ?

IPython/utils/tokenize2.py
```diff
@@ -91,11 +91,7 @@ def maybe(*choices): return group(*choices) + '?'
 Floatnumber = group(Pointfloat, Expfloat)
 Imagnumber = group(r'[0-9]+[jJ]', Floatnumber + r'[jJ]')
 Number = group(Imagnumber, Floatnumber, Intnumber)
-
-if sys.version_info.minor >= 3:
-    StringPrefix = r'(?:[bB][rR]?|[rR][bB]?|[uU])?'
-else:
-    StringPrefix = r'(?:[bB]?[rR]?)?'
+StringPrefix = r'(?:[bB][rR]?|[rR][bB]?|[uU])?'
```
Carreau: if I remember correctly this file is base on Python 3.2 stdlib, does it make sens to check whether the bugs have been fixed upstream on 3.4+ instead of having it diverge more ?

setupbase.py
```diff
@@ -40,15 +40,9 @@ def oscmd(s):
...
-
-try:
-    execfile
-except NameError:
-    def execfile(fname, globs, locs=None):
-        locs = locs or globs
-        exec(compile(open(fname).read(), fname, "exec"), globs, locs)
+def execfile(fname, globs, locs=None):
+    locs = locs or globs
+    exec(compile(open(fname).read(), fname, "exec"), globs, locs)
```
Carreau: +1

tox.ini
```diff
@@ -19,7 +19,7 @@ toxworkdir = /tmp/tox_ipython
 ; Other IPython/testing dependencies should be in setup.py, not here
 deps =
   pypy: https://bitbucket.org/pypy/numpy/get/master.zip
-  py{36,35,34,33,27}: matplotlib
+  py{36,35,34,33}: matplotlib
```
Carreau: Considering if we should just remove the `.tox` file as it is never used.

IPython/utils/py3compat.py
Carreau: We can mark it as deprecated once we are not using it anymore. If we mark it as deprecated and still use it here and there people complain we don't even follow our own deprecations.

Carreau:
I don't have much to say, we'll need to do that at some point (and that will simplify the code. My guts guts just make me worried that there will be subtle bugs, and that it may create conflicts later. 

I've been willing to start a 7.x branch at some point; and would be more comfortable to get that into a major release than a minor. @takluyver thoughts ? You've been more active than me on IPython in the last few weeks.

.gitignore
hugovk: Perhaps, but it's in the common Python one --https://github.com/github/gitignore/blob/master/Python.gitignore#L43 -- and ignored in some other Python projects I just checked. And it showed up after running the tests for this project (`iptest --coverage xml `), so it can show up for others too.

takluyver: Yeah, I'm not sure exactly what creates it, but I see it quite often when running tests, so I'm fine with adding it to gitignore.

IPython/utils/py3compat.py
takluyver: Thoughts: leave it in for now, except for the if/else branch that only runs on Py2. Add a deprecation note in the docstring. Then, over a succession of future releases:

1. Eliminate all remaining usage of it in IPython
2. Add a deprecation warning when it's loaded
3. (Eventually) get rid of it

It doesn't cost much to leave it sitting around unused for a while, and other code might be importing it. We did [document it](http://ipython.readthedocs.io/en/4.x/development/pycompat.html) in earlier versions, though the [current version](http://ipython.readthedocs.io/en/stable/development/pycompat.html) tells people not to rely on it.

IPython/utils/tokenize2.py
takluyver: I suspect this change is reverting a divergence that was previously introduced. Ultimately it would be good to get rid of it, but that's a more complex job for another day.

takluyver:
I think that if we make this a bit more conservative, it's fine for a 6.x release. I already had a look over it and I'm happy with most of the changes. The main thing is that I'd preserve the py3compat API at least through the 6.x series, and probably longer.

hugovk:
Updated to:

* retain all of py3compat.py as is, except for the if/else that only runs on Python 2
* update py3compat.py docstring to note it is deprecated and will be removed in a future version
* update pycompat.rst to note it will be removed in a future version
* delete unused tox.ini

takluyver:
This looks OK to me as it stands. I'm sure someone will complain about removing tox.ini, but it's evidently not in use, or people would have already removed Python 2 from the list. So I'd say that not having it is better than having it inaccurate.

takluyver:
Thanks @hugovk !

PR #10866, opened 2017-10-22
Simplify oinspect.getdoc.

anntzer:
AFAICT the dance with catching the exception and rethrowing it had
become pointless.


takluyver:
Looks like it was probably leftover debugging, but git blame shows it was added five years ago (#1715), so it shows that it's probably safe to let the exception raise.

PR #10867, opened 2017-10-23
FIX: bare functions in callback registry on bogus unregister

tacaswell:
7de483b5fdaa99fa289b418ce0efc5102287ba64 added checking for the case
where the callback was wrapped by backcall, however if there are
un-wrapped functions registered and a non-registered function is
attempted to be unregistered then this fails by raise an uncaught
AttributeError.

Where should a test for this go?

### DIFF
--- a/IPython/core/events.py
+++ b/IPython/core/events.py
@@ -69,8 +69,11 @@ class EventManager(object):
 
         # Remove callback in case ``function`` was adapted by `backcall`.
         for callback in self.callbacks[event]:
-            if callback.__wrapped__ is function:
-                return self.callbacks[event].remove(callback)
+            try:
+                if callback.__wrapped__ is function:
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! Tests in `IPython.core.tests.test_events`, please.

tacaswell:
I can drop the last commit if you want, just being a difficult busy body / making my editor less angry....

takluyver:
The wrapping is fine, but can you fix the test? ;-)

tacaswell:
cargo-culted how to test raising....

takluyver:
Needs a bit more cargo ;-)

```python
import nose.tools as nt
```

tacaswell:
3rd times the charm?

takluyver:
Seems it is!

PR #10868, opened 2017-10-23
ENH: add the non-interactive backends

tacaswell:
For completeness and to support future work of the Figure class having
`_repr_*_` methods.


takluyver:
Seems reasonable. I'll merge soon if no-one objects.

Issue #10869, opened 2017-10-24
Docs for IPython.sphinxext.ipython_directive


shoyer:
This module is still documented [internally](https://github.com/ipython/ipython/blob/master/IPython/sphinxext/ipython_directive.py), but I can't find a public doc page for it anymore in the latest release.

The most recent page I can find is for IPython 3, which is several years old at this point:
https://ipython.org/ipython-doc/3/api/generated/IPython.sphinxext.ipython_directive.html

takluyver:
The API docs generator excludes it with a comment that "This isn't API", which is kind of true. But we can include the module docstring into a page somewhere.

takluyver:
#10870 adds it

PR #10870, opened 2017-10-24
Document IPython.sphinxext

takluyver:
It's not really API, but it's actually got quite a detailed docstring, which it's worth exposing in the docs.

Closes gh-10869


shoyer:
Thanks @takluyver !

PR #10878, opened 2017-11-02
Fix GUI integration for support for pyqt5 apps that import QtWebEngineWidgets 

### LINKED ISSUE #10877: QTC qt5 integration broken for pyqt5 apps that import QtWebEngineWidgets
The `%gui qt5` magic plugs the IPython QtConsole into the Qt/pyqt5 event loop, so that PyQt5 application can be run from within IPython without blocking the kernel.

QT5 include the new modules QtWebEngine/QtWebEngineWidgets for displaying web content.
Loading a PyQt5 application within IPython currently throws with:
```
QtWebEngineWidgets must be imported before a QCoreApplication instance is created 
```

It seems IPython QtC creates a `QCoreApplication` without first importing `QtWebEngineWidgets`, so by the time the user's code tries to import  `QtWebEngineWidgets`, it's too late.

Problem code to [here](https://github.com/ipython/ipython/blob/53c132361f3ba145f93b9836fa26d328e6f7136e/IPython/external/qt_loaders.py#L265).

JeremySikes73:
closes https://github.com/ipython/ipython/issues/10877

### DIFF
--- a/IPython/external/qt_loaders.py
+++ b/IPython/external/qt_loaders.py
@@ -263,6 +263,13 @@ def import_pyqt5():
     import sip
 
     from PyQt5 import QtCore, QtSvg, QtWidgets, QtGui
+    try:
+        # If QtWebEngineWidgets will be used, Qt5 requires that it be imported
+        # before any QCoreApplication instance is created.
+        from PyQt5 import QtWebEngineWidgets
+    except ImportError:
+        # Not available, so the user can't use it anyway - Ignore.
+        pass
 
     # Alias PyQt-specific functions for PySide compatibility.
     QtCore.Signal = QtCore.pyqtSignal
--- a/IPython/external/qt_loaders.py
+++ b/IPython/external/qt_loaders.py
@@ -258,7 +258,7 @@ def import_pyqt5():
     """
     Import PyQt5
 
-    ImportErrors rasied within this function are non-recoverable
+    ImportErrors raised within this function are non-recoverable
     """
     import sip
 


takluyver:
It doesn't seem ideal to load it regardless of whether or not it's needed. @ccordoba12 , you use Qt more than most of us - it would be good to get your thoughts.

JeremySikes73:
In terms of ease of use, it *is* ideal. The cost is about 100MB of memory (size of the DLL) iff qt5 integration is requested. QWebEngine is the qt5 replacement of QtWebkit, the "Qt web view" widget. I expect it's fairly common in modern Qt Applications.

There are two alternatives:
1. Change the `%gui` magic to take arguments. This would also require changes to to config file, so the change can be made permanent.
2. Use a hack like:
```python
IPYTHON=1

def ensure_QApp(*args,**kwargs):
    app=None
    app = QApplication.instance() or QApplication([''])

    if IPYTHON:
        try:
            import IPython.terminal.pt_inputhooks.qt as ipqt
            ipqt._appref = app
        except:
            pass
    return app

app = ensure_QApp([''])
window = None

def main():
    global app, window
    window = MyMainWindow() 
    window.show() 

if __name__ == '__main__':  
    main()  
    app.exec_()  
```

Then, in IPython:
```
>>> import myapp # create the QApplication and shove it into IPython internals
>>> %gui qt5 # activate integration, will use the event loop of OUR QApplication
>>> foo.main() # show the main window, doesn't block now
```

This is cumbersome, touches IP internals, requires IP-specific changes to the application and the little dance has to be redone on every kernel restart. It doesn't require changes to IP though.

takluyver:
100 MB of memory is quite a lot, and I think plenty of Qt code doesn't need a web engine, e.g. matplotlib's Qt backend probably doesn't.

Maybe an easy if ugly workaround is to make a `%gui qt5-web` option that loads QWebEngine.

Carreau:

PR #10883, opened 2017-11-03
Fix broken jedi auto-completion with newest jedi

### LINKED ISSUE #10882: IPython 6.2.1 breaks dictionary greedy completion
Using IPython v6.2.0 and setting `c.Completer.greedy = True` in the ipython_config.py, I had been able to auto complete dictionary keys like this:
```
# define some dictionary
D = { 'a': 1, 'b': 2, 'c': 3 }
```
Now, hitting D[<tab> showed me relevant completion:
```
D['a']
D['b']
D['c']
```

With version v6.2.1 the list of completion is much longer showing me the list of all globals and builtins. There are still dictionary keys in the completion list, but they are at the end of the list.

I liked the previous behaviour and used it a lot, now it's not as useful anymore.

barlik:
Inside `IPCompleter:_jedi_matches()`, we're including `jedi.parser.tree`, which has been removed from jedi version 0.11. As a **temporary fix**, specify jedi 0.10 as a requirement. This fixes: https://github.com/ipython/ipython/issues/10882

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1350,7 +1350,13 @@ class IPCompleter(Completer):
 
         try:
             # should we check the type of the node is Error ?
-            from parso.tree import ErrorLeaf
+            try:
+                # jedi >= 0.11
+                from parso.tree import ErrorLeaf
+            except ImportError:
+                # jedi < 0.11
+                from jedi.parser.tree import ErrorLeaf
+
             next_to_last_tree = interpreter._get_module().tree_node.children[-2]
             completing_string = False
             if isinstance(next_to_last_tree, ErrorLeaf):
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1350,11 +1350,11 @@ class IPCompleter(Completer):
 
         try:
             # should we check the type of the node is Error ?
-            from jedi.parser.tree import ErrorLeaf
+            from parso.tree import ErrorLeaf
             next_to_last_tree = interpreter._get_module().tree_node.children[-2]
             completing_string = False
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks!

Carreau:
I'll recheck later, but I'm pretty sure there are many other part that do
work with Jedi  > 0.11, and work way better than with 0.11 or below.

Happy to do that as a stopgap, but we should not forget to fix it instead
of pinning Jedi.

I won't have much time to review before next week.

barlik:
@Carreau This is just a temporary regression fix and should ideally disappear within the next couple of days.

I'll have a look over the weekend and try to create a proper solution. But firstly I need to get familiar with the jedi codebase. Hopefully it won't take too long.

takluyver:
If you're going to investigate this weekend, let's leave this open but not merge it for now - it's unlikely that there will be a release of IPython in that time anyway.

barlik:
It turned out that the fix has been much simpler than what I have expected. The parser package hasn't been removed, but just refactored into a separate package called `parso` on which jedi itself is dependent.

ccordoba12:
Please make the fix so that both Jedi 0.10 and 0.11 are supported. Else, through our dependency on IPython, Spyder will also end up depending on `>=0.11` as well, but we want to give our users the possibility to use 0.10 as well.

barlik:
@ccordoba12 done

Carreau:
Thanks, that's look good to me; I really appreciate you looking into jedi/parso; I was aware David had split the two but was just missing the time to find what the new API was. 

Thanks !

Issue #10889, opened 2017-11-08
Nbcovert-pdfvialatex-table format


Rhadadi:
When I export as pdf via latex from jupyter notebook menu, it creates a nice pdf documents but Pandas output tables do not produced nice. Is there any way that I customise the table format, like the font size, headers, ...?

takluyver:
[Pandas options](http://pandas.pydata.org/pandas-docs/stable/options.html#available-options) include some settings to control its latex output - look for the ones starting `display.latex`

Rhadadi:
Thanks, Thomas!

takluyver:
You can register a custom Latex formatter to produce whatever you want. The code would look something like this:

```python
import pandas as pd

def dataframe_latex(df):
    return '... latex ...'

get_ipython().display_formatter.formatters['text/latex'].for_type(
    pd.DataFrame, dataframe_latex)
```

You might want to look at pandas' `DataFrame.to_latex()` method as a base for writing your own formatter.

Rhadadi:
Thank you for your detailed comment.

takluyver:
No problem. I've opened PR #10890 to have an example of registering a formatter function in the docs.

PR #10890, opened 2017-11-10
Extend docs on customising object display

takluyver:
I just answered a question on #10889 with information that I should be able to point to in the docs, but I couldn't find it anywhere. So here's a bit more information on customising how objects are displayed.

### DIFF
--- a/docs/source/config/integrating.rst
+++ b/docs/source/config/integrating.rst
@@ -19,6 +19,8 @@ returns a list of objects which are possible keys in a subscript expression
 .. versionadded:: 5.0
    Custom key completions
 
+.. _integrating_rich_display:
+
 Rich display
 ============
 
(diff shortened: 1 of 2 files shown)


docs/source/config/integrating.rst
```diff
@@ -40,6 +40,38 @@ For example::
...
+
+    def foo_html(obj):
+        return '<marquee>Foo object %s</marquee>' % obj.name
+
+    html_formatter = get_ipython().display_formatter.formatters['text/html']
+    html_formatter.for_type(Foo, foo_html)
+
+    # Or register a type without importing it - this does the same as above:
+    html_formatter.for_type_by_name('bar.Foo', foo_html)
+
```
Carreau: I think these are also documented in the `display()` docstring. Should it be cross linked there ?

Carreau (approved):
+1, feel free to merge if you want to add the crosslink to display later.

docs/source/config/integrating.rst
takluyver: Done.

Carreau: Sweet, last read and merge.

Issue #10903, opened 2017-11-16
Key completion on class


DavidMStraub:
The dictionary key completion feature and the possibility to define custom completion methods via `_ipython_key_completions_` are great, but I have a use case where it doesn't work: I have a class that keeps track of its instances, that all have (string) names, via a dictionary attribute. By defining `__getitem__` on the class's metaclass, I can access instances via their name like so:

```python
my_instance = NamedInstanceClass['name of my instance']
```

[See here for the definition](https://github.com/flav-io/flavio/blob/master/flavio/classes.py#L18-L68). Now, the reason why key completion doesn't work in this case are [these two lines](https://github.com/ipython/ipython/blob/master/IPython/utils/dir2.py#L63-L64) in `dir2` that exclude completion for classes. Indeed if I monkey-patch that method like

```python
def my_getattr(obj, method):
    return getattr(obj, method, None)
IPython.core.completer.get_real_method = my_getattr
```

then key completion works perfectly by defining a class method

```python
@classmethod
def _ipython_key_completions_(cls):
  return cls.instances.keys()
```

on the `NamedInstanceClass`.

If you agree that this is something that should/can be fixed, I would be happy to submit a PR.

System info:
```python
{'commit_hash': 'ca5443062',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/usr/local/lib/python3.6/dist-packages/IPython',
 'ipython_version': '6.2.1',
 'os_name': 'posix',
 'platform': 'Linux-4.13.0-17-generic-x86_64-with-Ubuntu-17.10-artful',
 'sys_executable': '/usr/bin/python3',
 'sys_platform': 'linux',
 'sys_version': '3.6.3 (default, Oct  3 2017, 21:45:48) \n[GCC 7.2.0]'}
```

takluyver:
I did a bit of git-blaming: the check was added in #7006, and it did fix some problems at the time.

I think the nature of Python can make it tricky to tell whether we're getting a class method or an unbound instance method, especially where different types of classes are involved (e.g. classes from compiled modules).

@minrk do you remember more about your thinking at the time you wrote the check?

minrk:
I suspect the check I added was naïve, not handling the case of classes properly. But #7006 is about display methods, not completer methods, though the logic is probably similar and with similar reasoning behind it.

[...]

PR #10905, opened 2017-11-16
Allow key completion for classes (#10903)

DavidMStraub:
Based on the discussion in #10903 (if I understood it correctly), I made a simple-minded modification of `get_real_method`: instead of ignoring methods when `obj` is a class, it is now only ignored if it is not a class method (i.e. if it is not an instance of `types.MethodType`).

I also added a unit test with a stripped-down version of my concrete use case

### DIFF
--- a/IPython/core/tests/test_completer.py
+++ b/IPython/core/tests/test_completer.py
@@ -939,6 +939,34 @@ def test_object_key_completion():
     nt.assert_in('qwick', matches)
 
 
+class NamedInstanceMetaclass(type):
+    def __getitem__(cls, item):
+        return cls.get_instance(item)
+
+class NamedInstanceClass(object, metaclass=NamedInstanceMetaclass):
+    def __init__(self, name):
+        if not hasattr(self.__class__, 'instances'):
+            self.__class__.instances = {}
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks for the pull request and thanks for the patience waiting for review, I was/am traveling and have low bandwidth. There seem to be some spurious test failure; I'll restart the build to see if they were random, and will review the code. Thanks !

Carreau:
Looks fine with a quick review. I'll try to understand it more in depth later.

As a side note, can you please deactivate the auto-trimming of trailing white-space of your editor. We don't mind trimming white-spaces, but when it's all-over it tends to screw git-blame and auto-backport. While in a perfect wold it wold be good to have it, until our code base is clean we try to only modify wite-space to lines close enough to code we change. 

Thanks !

DavidMStraub:
Oh, right. I restored the whitespace and did a rebase. Diff looks better now.

takluyver:
Thanks @DavidMStraub !

Carreau:
Thanks @takluyver for taking care of this. Apologies it completely fell under my radar. 

Thanks @DavidMStraub for contributing !

PR #10907, opened 2017-11-17
BUG sphinxext: ensure sys.stdout is used for logging (#10904)

### LINKED ISSUE #10904: IPython sphinx directive no longer reports errors in the code block
In pandas we are trying to switch from our vendored version to the upstream version of `IPython.sphinxext.ipython_directive`, but we noticed that if the code block contains an error, and thus when the code is run an exception is raised, this is no longer reported in the sphinx log.

I also identified the potential reason, and I think it is related to https://github.com/ipython/ipython/pull/9823 (cc @minrk)
This PR replaced a `io.stdout` with `sys.stdout` (https://github.com/ipython/ipython/pull/9823/files#diff-6420af2c1a12b55af7351690ec339cedL298), thus overwriting `sys.stdout` with `self.cout` of the `EmbeddedSphinxShell` (which is a StringIO object, I suppose to capture output of the embedded interactive shell).
However, `sys.stdout` is also used to print errors and warnings to the sphinx log:

https://github.com/ipython/ipython/blob/8c7ab85c0f4677b687cbb3d025189a86d7982619/IPython/sphinxext/ipython_directive.py#L505-L511

and those use cases seems to conflict with each other, because the result is that the error message is no longer written to the sphinx log.

I am not very familiar with stdout at al, but it seems this replacement was done because `IPython.utils.io` is deprecated? So I can't just revert it?

Another option is to better use the Sphinx logging utilities instead of writing the log warnings to stdout (so write the informative messages about errors to the actual sphinx logging). However, currently there is no dependency on Sphinx (no import of sphinx, only of docutils). Is this deliberate? (is there a usecase of this ipython directive outside of sphinx ?) Or can a dependency be added? (it should only be an optional dependency, which is IMO implicitly already the case for users of the ipython directive) 

If a sphinx import can be added, I can do a PR to use the actual sphinx logging.

jorisvandenbossche:
Closes #10904

Didn't add tests, as I think there are no tests for the sphinxext? (and adding is probably quite a bit more complicated than this fix)

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -287,14 +287,6 @@ class EmbeddedSphinxShell(object):
         IP = InteractiveShell.instance(config=config, profile_dir=profile)
         atexit.register(self.cleanup)
 
-        sys.stdout = self.cout
-        sys.stderr = self.cout
-
-        # For debugging, so we can see normal output, use this:
-        #from IPython.utils.io import Tee
-        #sys.stdout = Tee(self.cout, channel='stdout') # dbg
-        #sys.stderr = Tee(self.cout, channel='stderr') # dbg
-
         # Store a few parts of IPython we'll need.
         self.IP = IP
         self.user_ns = self.IP.user_ns


takluyver:
Thanks! This looks fine to me, following our discussion in #10904.

I think it's fine not to add tests in this PR, but if you wanted to add some as a separate PR, that would be welcome.

jorisvandenbossche:
Yes, I am doing a PR to sphinx itself now as well, and looking how they do tests. If that's not too hard to do something similar here, I will try that.

Issue #10649, opened 2017-06-09
Rethink completer limit to __all__


Carreau:
See https://github.com/ipython/ipykernel/issues/129

We have various possibility, but it seem like showing them by default is annoying people.

andportnoy:
Is anything blocking progress here?

PR #10914, opened 2017-11-21
Un-deprecate IPComplete.limit_to__all__ (#10649)

flying-sheep:
Since people want the to-be-removed functionality as default, we should at least keep the ability to configure it!

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1019,10 +1019,7 @@ class IPCompleter(Completer):
         """
     ).tag(config=True)
     limit_to__all__ = Bool(False,
-        help="""
-        DEPRECATED as of version 5.0.
-        
-        Instruct the completer to use __all__ for the completion
+        help="""Instruct the completer to use __all__ for the completion
         
         Specifically, when completing on ``object.<tab>``.
         


takluyver:
@Carreau ?

Carreau:
One of the issue is that (IIRC) we need to teach jedi about this. Which is the initial reason to have deprecated it. So un-deprecating it will just only partially limit to `__all__` which may be worse than just ignoring it as you will kinda-ignore-all but not for everything.

flying-sheep:
So this was deprecated for technical reasons as opposed to ideological ones, which means that once we solve the problem, there’s nothing standing in the way of keeping it.

Could you please elaborate on why such a partial limiting would happen? Which part?

Carreau:
There are 2 schools: 

 - People who want to hide elements for completions to not disturb normal user with internal details. 
 - Advance users that "know what they are doing" and want to be able to complete all the things, even private. 

Hence mechanism like limit_to_all, and also if you look carefully, `.<tab>` should not complete to private attributes, unless you actually start with a `_`, hence `._<tab>` will complete to more stuff that start with `_`.

Recently we added to our completer marchinery can now have several sources. The two sources are 
 - The core IPython completion machinery (historical, use Python introspection mechanism, has lots of historical baggage and options). 
 - Jedi, that can do static analysis on things like `[1, 'b'][0].<tab>` but know nothing about the historical completer options

The completer machinery does its best to meld the results of the multiple completer sources. even if the historical core completer could be made to follow the `__all__` convention, if jedi does not, then the completer will merge the two results, and what you-thought-you-had-excluded will reappear. 

While it's likely technically possible to figure out that without modifying jedi, it would be a really large effort. Modifying Jedi is as well possible, and would also require a lot of effort. And in the end a lot of what IPython does (and jedi does not), should be folded into jedi (unless it's really IPython specific). And in a perfect world, we would just register extension w/ jedi so that it can understand IPython specificity. ( I had also a prototype that added Kite completions, but did not ended up finishing and there was all this debate about the Company tracking users...)

So mostly, limit_to_all is deprecated because jedi (and potential other completer) does not understand it. 

Now there is also a reason for not limiting the completion to __all__, you can always ignore what you see, but you can hardly guess what you don't or maybe not even realize it exists. So as a default, not ignoring (IMHO) make sens.

Now, if we have a highly motivated contributor that really figure out how to make limit_to_all work correctly and help maintain the completer I'm pretty sure we could reinstate it – I believe spending time on jedi would be a better spent time that the old completer of IPython. But I prefer for now telling users that a specific function is deprecated and has no effects, than to have a half bakes solution that does not really work.

Now I haven't dived into Jedi and the completer in a long time, so things may have change a bit and I may be mistaken, I would be happy to be proved wrong.

Sorry for this a bit out of order explanation, I hope it make sens.

takluyver:
Doesn't sound ideal, because then the filters have to understand the code again in order to figure out what object is being completed on and check its `__all__`. The point of using jedi is to outsource that understanding where possible.

Carreau:
Completer is a real mess and this has not been touched in quite some time. 

To try to decrease the number of in-flight PR I'll tag this with the "closed-pr" label, meaning there is valuable code and discussion and it's ok to reopen.

I hope we can find some funding or manpower to rework the completion more in depth and please everyone.

PR #10912, opened 2017-11-20
Add docs about building the docs.

douglatornell:
re: #10895

PR #10918, opened 2017-11-21
Improve local docs building docs to ensure all deps are installed

douglatornell:
re: https://github.com/ipython/ipython/pull/10912#pullrequestreview-78043112

@jzf2101 @Carreau Working in a clean conda environment locally I found that it is necessary to do both a dev install and then add the deps from `docs/requirements.txt`. So, this change links to the dev install instructions and then the requirements install. 

Works for either conda env or venv, but doesn't make sense to me in the debian package manager context, so I took that out (assuming we don't actually want to encourage system-wide installs for dev work - though I guess it's relevant if you are provisioning a VM or building a container, but then you probably don't need to be told, right?)

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -16,22 +16,17 @@ Requirements
 
 The documentation must be built using Python 3.
 
-The following tools are needed to build the documentation:
+In additions to :ref:`devinstall`,
+the following tools are needed to build the documentation:
 
  - sphinx
  - sphinx_rtd_theme
+ - docrepr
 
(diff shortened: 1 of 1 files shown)


takluyver:
+1, I think it's better to point to one simple way of building than three. All the packages needed, as far as I know, can easily be installed with pip.

Carreau:
Agreed, let's keep things simple.

Carreau:
Thanks !

jzf2101:
Super thorough thanks!

PR #10940, opened 2017-12-07
sphinx: turn warnings into errors

### LINKED ISSUE #10916: Use the Sphinx option which causes an error if any warning occurs
see @takluyver 's comment in #10895

cmccandless:
Resolves #10916

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -32,12 +32,16 @@ import inspect
 import linecache
 import sys
 import warnings
+import re
 
 from IPython import get_ipython
 from IPython.utils import PyColorize
 from IPython.utils import coloransi, py3compat
 from IPython.core.excolors import exception_colors
 from IPython.testing.skipdoctest import skip_doctest
+
+
+RGX_EXTRA_INDENT = re.compile('(?<=\n)\s+')
 
 
 prompt = 'ipdb> '
@@ -175,6 +179,10 @@ class Tracer(object):
         self.debugger.set_trace(sys._getframe().f_back)
 
 
+def strip_indentation(multiline_string):
+    return RGX_EXTRA_INDENT.sub('', multiline_string)
+
+
 def decorate_fn_with_doc(new_fn, old_fn, additional_text=""):
     """Make new_fn have old_fn's doc string. This is particularly useful
     for the ``do_...`` commands that hook into the help system.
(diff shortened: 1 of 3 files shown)


takluyver:
So as expected, this is causing a failure on Travis, because we haven't yet eliminated all warnings. We'll need to finish clearing them up before this can be merged.

If possible, I'd also like to use this behaviour only when building the docs on Travis, not every time we build the docs locally. We can already see warnings, and I think it's still useful to be able to build the docs when there are warnings.

takluyver:
Thanks, that looks good for the infrastructure.

Now we just need to get the docs building without warnings. Do you want to have a go at that?

cmccandless:
If understand the code correctly, docstrings are being imported from the official pdb and applied to the internal implementation of the Pdb class. One of these is the malformed docstring that is causing the warning. However, [looking at the source code](https://github.com/python/cpython/blob/ede157331b4f9e550334900b3b4de1c8590688de/Lib/pdb.py#L1106-L1109) for the actual docstring, I don't see any issues, so I'm not entirely sure what the issue is. It's possible that there is an issue with the `decorate_fn_with_doc` function. I'll keep digging.

I am having issues building the docs locally, so I'm going to add [WIP] to the PR title while I test things. Please ignore commits to this PR until I have removed the [WIP] tag from the title.

takluyver:
The `do_quit` docstring is a bit weird:

```python
    def do_quit(self, arg):
        """q(uit)\nexit
        Quit from the debugger. The program being executed is aborted.
        """
```

There's a newline between quit and exit, so the string is:

```
q(uit)
exit
        Quit from the debugger. The program being executed is aborted.
```

We could stop inheriting this, and write a nicer docstring in our own code. Or we could make the `decorate_fn_with_docstring` utility smarter - but I don't know exactly what it should do with that.

cmccandless:
@takluyver I went with "make `decorate_fn_with_docstring` smarter": indentation is now removed from imported docstrings.

IPython/core/debugger.py
```diff
@@ -40,6 +41,9 @@
 from IPython.testing.skipdoctest import skip_doctest
 
 
+RGX_EXTRA_INDENT = re.compile('(?<=\n)\s+')
```
takluyver: Style point - let's put this near the function that uses it, so it's easier to see them both together.

takluyver:
Thanks!

Issue #10943, opened 2017-12-14
autoreload of __mp_main__ still fails when running script


rmrao:
There appear to have been two attempts to fix this problem (#8568 and #10523), however I am still getting this problem consistently. I have been able to narrow down the cause somewhat. First of all, the problem appears platform-specific. It occurs whenever you run a script that contains `import multiprocessing` or a number of 3rd party libraries (e.g. some scipy libraries), but only on Mac. Running the same script with the same ipython version number on Linux does not appear to cause this problem.

On my Mac, I get the following output:
```python
In [2]: sys.modules.get('__mp_main__')
Out[2]: <module '?' from '/Users/roshan_rao/class/deeprl/l2l-gradfree/source/empty.py'>
```
which means that, despite the key in `sys.modules` being `'__mp_main__'`, the actual attribute `__name__` does not exist. However on Linux, the script doesn't even appear to be loaded as a module (which I guess makes sense?), so the problem doesn't arise. For completion, here's the output of the same command:
```python
In [13]: sys.modules.get('__mp_main__')
Out[13]: <module '__main__' from '/home/rmrao/anaconda3/bin/ipython'>
```

The fixes proposed have resulted in the following addition to the code:
```python
if getattr(module, '__name__', None) in ['__mp_main__', '__main__']:
            # we cannot reload(__main__) or reload(__mp_main__)
            return None, None
```
Unfortunately, if the `__name__` attribute does not exist, then `getattr(module, '__name__', None)` returns `None`, which isn't actually checked for. I think the solution proposed in https://github.com/ipython/ipython/issues/8564#issuecomment-150321942 would work. My local workaround has been explicitly adding `None` to the list, which would give
```python
if getattr(module, '__name__', None) in [None, '__mp_main__', '__main__']:
            # we cannot reload(__main__) or reload(__mp_main__)
            return None, None
```
I can submit a pull request in a few moments.

System info:
```
{'commit_hash': 'd86648c5d',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/roshan_rao/anaconda/lib/python3.6/site-packages/IPython',
 'ipython_version': '6.1.0',
 'os_name': 'posix',
 'platform': 'Darwin-17.2.0-x86_64-i386-64bit',
 'sys_executable': '/Users/roshan_rao/anaconda/bin/python',
 'sys_platform': 'darwin',
[...]

PR #10944, opened 2017-12-14
make sure module actually has __name__ attribute

rmrao:
Fix for issue #10943


takluyver:
This makes sense. Thanks for the detailed investigation!

PR #10948, opened 2017-12-18
Avoid PyGIDeprecationWarning when using Gtk3 backend.

pkgw:
Since pygobject 3.7.3, which was [released ~5 years ago](https://piware.de/2012/12/pygobject-3-7-3-released/), the function `GLib.io_add_watch` has preferred a call signature other than the one that our input backends were using.

(Because the new API has been around for so long, I think it's OK to stop supporting older versions of `pygobject`.)

This seems to work fine in local testing.


takluyver:
Thanks! It sounds like you know more about this, so I'll trust you that that's the preferred signature.

PR #10956, opened 2017-12-20
Swap the check for Jedi versions around

takluyver:
It's possible to have Parso installed alongside an older version of Jedi which doesn't use it. I had this situation, and it was messing up my tab completions because it couldn't detect when it was in a string.

Jedi 0.11 doesn't have the `jedi.parser` subpackage, so I think checking it this way should be more reliable.


minrk (approved):

Carreau:
Thanks !

PR #10958, opened 2017-12-21
Don't sort dicts for display in Python >= 3.6

takluyver:
Closes gh-10110

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -95,7 +95,9 @@ __all__ = ['pretty', 'pprint', 'PrettyPrinter', 'RepresentationPrinter',
 
 
 MAX_SEQ_LENGTH = 1000
-DICT_IS_ORDERED = sys.version_info >= (3, 7)
+# The language spec says that dicts preserve order from 3.7, but CPython
+# does so from 3.6, so it seems likely that people will expect that.
+DICT_IS_ORDERED = sys.version_info >= (3, 6)
 _re_pattern_type = type(re.compile(''))
 
 def _safe_getattr(obj, attr, default=None):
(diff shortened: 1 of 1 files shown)


IPython/lib/pretty.py
```diff
@@ -95,6 +95,7 @@ def _repr_pretty_(self, p, cycle):
 
 
 MAX_SEQ_LENGTH = 1000
+DICT_IS_ORDERED = sys.version_info >= (3, 7)
```
njsmith: Maybe:

```python
DICT_IS_ORDERED = (
    sys.version_info >= (3, 7)
    or platform.python_implementation() == "PyPy"
    or (platform.python_implementation() == "CPython" and sys.version_info >= (3, 6))
)

takluyver: I'm inclined to use 3.7, because that's when Python dicts are defined as being ordered, rather than it being up to the implementation.

If we want to push it faster, I'd rather say 3.6 than start checking which implementation it is. But following the language definition feels like the better thing to do.

njsmith: I was going to suggest making it `sys.version_info >= (3, 6)`, and then guessed that you would give the pedantically correct response that technically there could someday be a 3.6 valid implementation that doesn't have ordered dicts, so I switched to the exact pedantically accurate check you see above :-).

I'm not sure how it helps end-users to order their dicts differently on 3.6 versus 3.7, given that in practice the actual interpreter behavior is identical. Most users don't understand the complex metaphysical question of how the exact same behavior is different depending on whether it's nominally accidental or nominally intentional. OTOH we do have multiple reports from people that the sorting on 3.6 is causing them pain right now.

tacaswell: I lobby for not sorting on 3.6+ on the practicality over purity platform.

takluyver: OK, I've set it to 3.6.

PR #10959, opened 2017-12-21
Defining __repr__ should override pretty printer for parent class

### LINKED ISSUE #10950: OrderedDict output differs in ipython from python (and official documentation)
[collections — Container datatypes](https://docs.python.org/3/library/collections.html#collections.OrderedDict)

An ordered dictionary can be combined with the Counter class so that the counter remembers the order elements are first encountered:

```
from collections import Counter, OrderedDict

class OrderedCounter(Counter, OrderedDict):
    'Counter that remembers the order elements are first encountered'

    def __repr__(self):
        return '%s(%r)' % (self.__class__.__name__, OrderedDict(self))

    def __reduce__(self):
        return self.__class__, (OrderedDict(self),)
```
This is based off a Raymond Hettinger - Super considered super! - PyCon 2015, and his blog post [Python’s super() considered super!](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/).

My confusion is IPython returns a result that differs from the Python it's built on.  The ordering of the output is important and the recipe in Hettinger's blog and the official documentation preserves the insertion order but IPython 6.1.0 built on Python 3.6.3 does not.  However if I execute directly in Python 3.6.3 I receive the expected desired result.

`oc = OrderedCounter('abracadabra')`

output in Python 3.6.3 (and Python 2.7.14)
`OrderedCounter(OrderedDict([('a', 5), ('b', 2), ('r', 2), ('c', 1), ('d', 1)]))`

output in IPython 6.1.0 on Python 3.6.3 Anaconda
`OrderedCounter({'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2})`  //ordering is not preserved, seems sorted by key

Would someone kindly explain why the output is different in IPython 6.1.0

takluyver:
I think that the pretty printer should always prefer a repr defined for an earlier class in the MRO, whether that is as a registered function, a `_repr_pretty_` method or a `__repr__` method. I.e. if you define:

```python
class A:
    def _repr_pretty_(self, pprint, cycle):
        pprint.text("<A>")

class B(A):
    def __repr__(self):
        return "<B>"
```

Then instances of B should display as `<B>`, not `<A>`.

Closes #10950.

This is related to #452, though it actually reverts the implementation from there. That PR implemented a similar check within the pretty-printers for certain basic types, like lists and dicts. This one aims to do the same thing in a more general way, which should make the earlier implementation redundant.

### DIFF
--- a/IPython/core/tests/test_formatters.py
+++ b/IPython/core/tests/test_formatters.py
@@ -49,7 +49,7 @@ def test_pretty():
     f = PlainTextFormatter()
     f.for_type(A, foo_printer)
     nt.assert_equal(f(A()), 'foo')
-    nt.assert_equal(f(B()), 'foo')
+    nt.assert_equal(f(B()), 'B()')
     nt.assert_equal(f(GoodPretty()), 'foo')
     # Just don't raise an exception for the following:
     f(BadPretty())
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -392,6 +392,10 @@ class RepresentationPrinter(PrettyPrinter):
                             meth = cls._repr_pretty_
                             if callable(meth):
                                 return meth(obj, self, cycle)
+                        if cls is not object \
+                                and callable(cls.__dict__.get('__repr__')):
+                            return _repr_pprint(obj, self, cycle)
+
             return _default_pprint(obj, self, cycle)
         finally:
             self.end_group()
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -541,17 +541,12 @@ def _default_pprint(obj, p, cycle):
     p.end_group(1, '>')
 
 
-def _seq_pprinter_factory(start, end, basetype):
+def _seq_pprinter_factory(start, end):
     """
     Factory that returns a pprint function useful for sequences.  Used by
     the default pprint for tuples, dicts, and lists.
     """
     def inner(obj, p, cycle):
-        typ = type(obj)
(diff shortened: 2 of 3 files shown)


rkern:
Sounds reasonable. Might also be worth adding a pprinter for `OrderedDict`s and `defaultdict`s while we're at it. Here's mine:

```
def defaultdict_pprinter(obj, p, cycle):
    p.text(type(obj).__name__)
    p.text('(')
    p.pretty(obj.default_factory)
    p.text(',')
    p.breakable()
    if cycle:
        return p.text('{...})')
    keys = obj.keys()
    try:
        keys.sort()
    except Exception:
        # Sometimes the keys don't sort.
        pass
    with p.group(2, '{', '})'):
        p.breakable()
        for i, key in enumerate(keys):
            if i > 0:
                p.text(',')
                p.breakable()
            p.pretty(key)
            p.text(': ')
            p.pretty(obj[key])


def ordered_dict_pprinter(obj, p, cycle):
    p.text(type(obj).__name__)
    p.text('(')
    if cycle:
        return p.text('[...]')
    with p.group(2, '[', '])'):
        p.breakable()
        for i, item in enumerate(obj.items()):
            if i > 0:
                p.text(',')
                p.breakable()
            p.pretty(item)
```

takluyver:
Hi @rkern , I didn't realise you were paying attention to these issues. :-)

There are already printers in there for ordered dicts and default dicts, though the code looks slightly different to yours.

PR #10962, opened 2017-12-23
added --isolated option for %%html magic

sjdv1982:
Fix for https://github.com/jupyter/notebook/issues/3153

### DIFF
--- a/IPython/core/magics/display.py
+++ b/IPython/core/magics/display.py
@@ -16,6 +16,7 @@ from IPython.core.display import display, Javascript, Latex, SVG, HTML, Markdown
 from IPython.core.magic import  (
     Magics, magics_class, cell_magic
 )
+from IPython.core import magic_arguments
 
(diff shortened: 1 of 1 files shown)


Carreau:
Seem reasonable to me. Thanks !

IPython/core/magics/display.py
```diff
@@ -59,12 +60,23 @@ def svg(self, line, cell):
...
     @cell_magic
     def html(self, line, cell):
         """Render the cell as a block of HTML"""
-        display(HTML(cell))
-        
-    @cell_magic    
+        args = magic_arguments.parse_argstring(self.html, line)
+        html = HTML(cell)
+        if args.isolated:
+            display(html, metadata={'text/html':{'isolated':True}})
```
rgbkrk: We should make note of this metadata in http://jupyter.readthedocs.io/en/latest/reference/mimetype.html or wherever seems like a good choice. /cc @willingc

PR #10964, opened 2017-12-29
[WIP] Upgrade to prompt_toolkit 2.0

jonathanslenders:
There are still some bugs and a couple of things to fix.
- async autocompletion.
- result always appears below output prompt
- debugger has not yet been tested.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -18,12 +18,15 @@ from traitlets import (
 from prompt_toolkit.document import Document
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
 from prompt_toolkit.filters import (HasFocus, Condition, IsDone)
+from prompt_toolkit.formatted_text import PygmentsTokens
 from prompt_toolkit.history import InMemoryHistory
-from prompt_toolkit.shortcuts import create_prompt_application, create_eventloop, create_prompt_layout, create_output
-from prompt_toolkit.interface import CommandLineInterface
-from prompt_toolkit.key_binding.manager import KeyBindingManager
+from prompt_toolkit.shortcuts import PromptSession, CompleteStyle
+from prompt_toolkit.output.defaults import create_output
+from prompt_toolkit.key_binding import KeyBindings
 from prompt_toolkit.layout.processors import ConditionalProcessor, HighlightMatchingBracketProcessor
-from prompt_toolkit.styles import PygmentsStyle, DynamicStyle
+from prompt_toolkit.patch_stdout import patch_stdout
+from prompt_toolkit.styles import DynamicStyle, merge_styles
+from prompt_toolkit.styles.pygments import style_from_pygments_cls, style_from_pygments_dict
 
 from pygments.styles import get_style_by_name
 from pygments.style import Style
@@ -34,7 +37,7 @@ from .magics import TerminalMagics
 from .pt_inputhooks import get_inputhook_name_and_func
 from .prompts import Prompts, ClassicPrompts, RichPromptDisplayHook
 from .ptutils import IPythonPTCompleter, IPythonPTLexer
-from .shortcuts import register_ipython_shortcuts
+from .shortcuts import create_ipython_shortcuts
 
 DISPLAY_BANNER_DEPRECATED = object()
 
@@ -93,12 +96,11 @@ class TerminalInteractiveShell(InteractiveShell):
                                                   'to reserve for the completion menu'
                             ).tag(config=True)
 
-    def _space_for_menu_changed(self, old, new):
-        self._update_layout()
+#    def _space_for_menu_changed(self, old, new):
+#        self._update_layout()
 
-    pt_cli = None
+    pt_app = None
     debugger_history = None
-    _pt_app = None
 
     simple_prompt = Bool(_use_simple_prompt,
         help="""Use `raw_input` for the REPL, without completion and prompt colors.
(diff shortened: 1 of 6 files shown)


jonathanslenders:
See also: https://github.com/jonathanslenders/python-prompt-toolkit/issues/568

IPython/terminal/interactiveshell.py
```diff
@@ -444,9 +450,9 @@ def pre_prompt(self):
             # after the buffer is reset.
             s = self.rl_next_input
             def set_doc():
-                self.pt_cli.application.buffer.document = Document(s)
-            if hasattr(self.pt_cli, 'pre_run_callables'):
-                self.pt_cli.pre_run_callables.append(set_doc)
+                self.pt_app.application.buffer.document = Document(s)
+            if hasattr(self.pt_app.application, 'pre_run_callables'):
```
takluyver: IIRC this `hasattr()` check was for compatibility with older prompt_toolkit versions. As this will require 2.x anyway, we can probably remove it and only use `pre_run_callables`.

IPython/terminal/interactiveshell.py
```diff
@@ -340,44 +332,58 @@ def _make_style_from_name_or_cls(self, name_or_cls):
...
-        """
-        if self._pt_app:
-            self._pt_app.layout = create_prompt_layout(**self._layout_options())
+#    def _update_layout(self):
+#        """
+#        Ask for a re computation of the application layout, if for example ,
+#        some configuration options have changed.
+#        """
+#        if self.pt_app:
+#            self.pt_app.layout = create_prompt_layout(**self._extra_prompt_options())
```
takluyver: Is this no longer necessary, or do we need a different way of doing it? There are some settings we can change at runtime, and we needed to do this for them to take effect.

takluyver:
Thanks for making a start on this!

I remember you said that PT 2 would add a way to do tab completions on the main thread. I'm looking forward to this, because we've had a number of reports of completion problems resulting from it using a separate thread. Is that API in place? Will it be easy to integrate?

IPython/terminal/interactiveshell.py
jonathanslenders: Right. Fixed now.

IPython/terminal/interactiveshell.py
jonathanslenders: This is indeed no longer necessary, unless I miss something. All the ``extra_prompt_options`` are passed to each ``prompt()`` call. This should update them every time.

jonathanslenders:
prompt_toolkit 2.0 does indeed support synchronous completion. This is even the default, because apparently, that's what most people want, and is the easiest. Asynchronous completion is possible by passing ``complete_in_thread=True`` to the ``prompt()`` function.

IPython/terminal/interactiveshell.py
jonathanslenders: What is actually the use for this ``pre_prompt`` function? What is ``rl_next_input``?

takluyver:
Great, so we should get synchronous completion without doing anything else.

IPython/terminal/interactiveshell.py
takluyver: Various functions can set some text to appear at the next prompt, e.g. I can run `%load foo.py` to get the contents of `foo.py` at the next prompt. There's also `%recall` which can retrieve entries from history. That gets stored in `rl_next_input` and then inserted when the next prompt is being prepared.

takluyver:
Hi @jonathanslenders ! How's v2 coming along? :-)

jonathanslenders:
Hi @takluyver,
Thanks for asking. It's going well actually, I still have to do a bit of testing, but it's pretty solid already.
It should be possible to release very soon. Would you merge this for IPython 7, or simply the next 6.x release?
Are you coming to PyCon?

takluyver:
My normal inclination would be to do this for 7.0, but I know that people have had problems with tab completion running in a separate thread, and so we may end up doing it as a bugfix, possibly even back to 5.x. The remaining Python 2 users now can't go past 5.x, so there's some extra pressure to backport things.

takluyver:
I'm not coming to Pycon (I'm avoiding travelling to the US as a kind of personal protest), but I'm hoping to be at EuroPython later this year.

gpotter2:
Hello,

Is there any release date of `prompt_toolkit>=2.0.0` and its `IPython` integration? Yet we can't have both IPython and new prompt_toolkit features running at the same time.

Thanks a lot for both your projects, there are just amazing and really useful!

Carreau:
If we want to make a smallish-7.x release that simply update to `prompt_toolkit>=2.0.0` I would be ok with that.

gpotter2:
Seems like `prompt_toolkit` 2.0.1 has just been released ! (btw great job !)

Hope to see an IPython integration soon :)

takluyver:
Nice one @jonathanslenders :-) Are you going to get back to this PR, or would you rather someone else picked it up?

And well spotted @gpotter2 - it hasn't even been out one hour yet!

screenshot from 2018-06-02 19-28-05

jonathanslenders:
@takluyver: I'm not sure I have the time for all the required testing. I think all the relevant changes are done, so maybe somebody can take it from here.

A couple of things:
- Maybe add an option to configure whether completion needs to be async or not. (Check for ``complete_in_thread`` in this PR.)
- Use colors in the completion menu depending on the type (function, attribute, file, etc..). See: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/examples/prompts/auto-completion/colored-completions.py
- If people defined custom key bindings for IPython somehow, they have to update their code.
- Right now, this pull request still uses Pygments tokens everywhere. This will keep working, but it would be good in the future to move to ``(style_string, text)`` tuples. See: https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/styling.html and https://python-prompt-toolkit.readthedocs.io/en/master/pages/printing_text.html#formatted-text

takluyver:
I've rebased and fixed up a few problems in #11177.

gpotter2:
@takluyver Thanks a lot for your work !

gpotter2:
Any reason to keep this PR open ?

takluyver:
Nope, it was superseded by #11177.

PR #10966, opened 2017-12-30
In obj??, display the docstring if it is not in the source.

anntzer:
Currently, `obj??` does not display obj's docstring because it is likely
redundant with the source (which usually also includes the docstring).
However, when the docstring had been dynamically added, then it should
also be displayed by `obj??` (because it is not redundant in this case).

A reasonable way to check for this condition is to parse the source and
compare the result of `ast.get_docstring()` to the result of
`inspect.getdoc()`, as done by this PR.

No tests yet but would be happy to add them if I can get some pointers as to where to add them.
See https://github.com/matplotlib/matplotlib/pull/9475#issuecomment-354178459 and following comments for original motivation.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -13,6 +13,7 @@ reference the name under which an object is being read.
 __all__ = ['Inspector','InspectColors']
 
 # stdlib modules
+import ast
 import inspect
 from inspect import signature
 import linecache
@@ -630,10 +631,10 @@ class Inspector(Colorable):
             # Functions, methods, classes
             append_field(_mime, 'Signature', 'definition', code_formatter)
             append_field(_mime, 'Init signature', 'init_definition', code_formatter)
+            append_field(_mime, 'Docstring', 'docstring', formatter)
             if detail_level > 0 and info['source']:
                 append_field(_mime, 'Source', 'source', code_formatter)
             else:
-                append_field(_mime, 'Docstring', 'docstring', formatter)
                 append_field(_mime, 'Init docstring', 'init_docstring', formatter)
 
             append_field(_mime, 'File', 'file')
@@ -821,7 +822,7 @@ class Inspector(Colorable):
                 pass
 
         # Add docstring only if no source is to be shown (avoid repetitions).
-        if ds and out.get('source', None) is None:
+        if ds and not self._source_contains_docstring(out.get('source'), ds):
             out['docstring'] = ds
 
         # Constructor docstring for classes
@@ -935,6 +936,23 @@ class Inspector(Colorable):
                     argspec_dict['varkw'] = argspec_dict.pop('keywords')
 
         return object_info(**out)
+
+    @staticmethod
+    def _source_contains_docstring(src, doc):
+        """
+        Check whether the source *src* contains the docstring *doc*.
+
+        This is is helper function to skip displaying the docstring if the
+        source already contains it, avoiding repetition of information.
+        """
+        try:
+            def_node, = ast.parse(dedent(src)).body
+            return ast.get_docstring(def_node) == doc
+        except Exception:
+            # The source can become invalid or even non-existent (because it
+            # is re-fetched from the source file) so the above code fail in
+            # arbitrary ways.
+            return False
 
     def psearch(self,pattern,ns_table,ns_search=[],
                 ignore_case=False,show_all=False):


IPython/core/oinspect.py
```diff
@@ -821,8 +822,22 @@ def _info(self, obj, oname='', info=None, detail_level=0) -> dict:
...
+                    def_node, = ast.parse(dedent(out['source'])).body
+                    node_docstring = ast.get_docstring(def_node)
+                except Exception:
+                    # The source can become invalid (because it is re-fetched
+                    # from the source file) so the parse can fail in arbitrary
+                    # ways.
+                    pass
+                else:
+                    if node_docstring == ds:
+                        skip_ds = True
```
eric-wieser: Contents of this `if` could be a helper function `skip_ds = docstring_matches_source(out)`

anntzer: A helper for an equality check seems a bit overkill, no?

eric-wieser: I mean the entire try/catch

anntzer: I think the whole method needs some refactoring, in particular so that we can compare with the *actual* docstring and not some of the autogenerated fill-in docstrings that come from the top of the method...

eric-wieser: All the more reason to wrap this into a function, so that you can easily swap out how the docstring is obtained later

IPython/core/oinspect.py
```diff
@@ -630,10 +631,10 @@ def code_formatter(text):
             # Functions, methods, classes
             append_field(_mime, 'Signature', 'definition', code_formatter)
             append_field(_mime, 'Init signature', 'init_definition', code_formatter)
+            append_field(_mime, 'Docstring', 'docstring', formatter)
             if detail_level > 0 and info['source']:
                 append_field(_mime, 'Source', 'source', code_formatter)
             else:
-                append_field(_mime, 'Docstring', 'docstring', formatter)
                 append_field(_mime, 'Init docstring', 'init_docstring', formatter)
```
eric-wieser: I wonder if a similar argument can be used for moving this to become unconditional (although it's perhaps out of scope for this change)

anntzer: Again, this would mostly make sense with a proper refactoring so that the same AST test can be applied to the init docstring, but that'll have to go in another PR.

IPython/core/oinspect.py
anntzer: done

IPython/core/oinspect.py
```diff
@@ -821,7 +822,7 @@ def _info(self, obj, oname='', info=None, detail_level=0) -> dict:
                 pass
 
         # Add docstring only if no source is to be shown (avoid repetitions).
-        if ds and out.get('source', None) is None:
+        if ds and not self._source_contains_docstring(out.get('source'), ds):
```
eric-wieser: This should probably handle `out.get('source')` is none, rather than passing `None` onto the function.

anntzer: We already have to catch all possible parse failures (or parsing into something unexpected) in _source_contains_docstrings (because the source file is re-read when getsource is called, so it could have been modified in between), doesn't seem to be a stretch to also handle the case of None in it.

IPython/core/oinspect.py
```diff
@@ -936,6 +937,27 @@ def _info(self, obj, oname='', info=None, detail_level=0) -> dict:
...
+
+        This is is helper function to skip displaying the docstring if the
+        source already contains it, avoiding repetition of information.
+        """
+        try:
+            def_node, = ast.parse(dedent(src)).body
+            node_docstring = ast.get_docstring(def_node)
+        except Exception:
+            # The source can become invalid (because it is re-fetched
+            # from the source file) so the parse can fail in arbitrary
```
eric-wieser: Right now you're catching `TypeError`s from `src` being `None` here, which is nothing to do with `parse` failing (see above comment)

IPython/core/oinspect.py
eric-wieser: It would be nice to handle that case explicitly, rather than secretly in an `except Exception`

IPython/core/oinspect.py
```diff
@@ -936,6 +937,27 @@ def _info(self, obj, oname='', info=None, detail_level=0) -> dict:
...
+            # ways.
+            pass
+        else:
+            if node_docstring == doc:
+                return True
```
eric-wieser: May as well be `return node_docstring == doc`

IPython/core/oinspect.py
```diff
@@ -936,6 +937,27 @@ def _info(self, obj, oname='', info=None, detail_level=0) -> dict:
...
+        return False
```
eric-wieser: This can replace the `pass` above

IPython/core/oinspect.py
anntzer: This already has to catch at least (SyntaxError, ValueError, TypeError) (syntaxerror because of parse failures, valueerror because `body` may have become more than a single node (and I want to fail in that case too), typeerror because the single node in the body may now be e.g. an Expr node, and in that case ast.get_docstring *will* raise a TypeError.

Distinguishing catching a TypeError that comes from source == None and a TypeError that comes from the source having become a single expression seems like splitting hairs to me.

IPython/core/oinspect.py
anntzer: edited accordingly

IPython/core/oinspect.py
eric-wieser: My complaint is that the comment below about the parse failing is technically incorrect - in this case, the parse was never even attempted. Updating the comment is the other way to resolve that

anntzer: done

minrk:
Thanks!

PR #10969, opened 2018-01-03
Skip our own Python completions when using Jedi

takluyver:
cc @Carreau 

At present, we combine our own completions with those from Jedi. This means we won't catch anything that Jedi is missing, making it harder to eventually drop our own completion machinery.

This makes the `use_jedi` option switch between Jedi and our own completion machinery, so by default the Jedi completions will be used, and our own completions will not. If people experience problems, they can configure `Completer.use_jedi = False` to go back to our completion machinery for now (and hopefully file issues so we're aware of the problems).

I also got rid of a parameter which doesn't appear to be used anywhere.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1102,13 +1102,20 @@ class IPCompleter(Completer):
         #= re.compile(r'[\s|\[]*(\w+)(?:\s*=?\s*.*)')
 
         # All active matcher routines for completion
-        self.matchers = [
-                         self.python_matches,
-                         self.file_matches,
-                         self.magic_matches,
-                         self.python_func_kw_matches,
-                         self.dict_key_matches,
-                         ]
+        if self.use_jedi:
+            self.matchers = [
+                self.file_matches,
+                self.magic_matches,
+                self.dict_key_matches,
+            ]
+        else:
+            self.matchers = [
+                self.python_matches,
+                self.file_matches,
+                self.magic_matches,
+                self.python_func_kw_matches,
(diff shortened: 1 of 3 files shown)


Carreau:
Seem reasonable to me. 
The test failures looks like jedi is not working. Stange.

takluyver:
Some tests assumed our own completion machinery was in use. I've fixed them, but not very elegantly.

Carreau:
Do you want to have that on the 5.x branch ? or 6.x only ?

Carreau (approved):
Feel free to self-merge after choosing 5.x or 6.x

takluyver:
5.x doesn't use Jedi, right? So this should be for 6.x.

Carreau:
> 5.x doesn't use Jedi, right? So this should be for 6.x.

Ah you are right. I need to wake up.

gabriellima:
I think there's another "bug". Ipython is doing case-**insensitive** completion.

If I do `%config Completion.use_jedi = False`  I "temporaly" fix this issue. 

Ipython 6.2.1
ipykernel 4.8.0
jedi 0.11.1

takluyver:
Can you open a new issue about that? I suspect it's deliberate-ish that Jedi does case-insensitive completions, but ideally we could be smarter about it. I think IPython's own machinery tries to do case-sensitive completions first, and then case-insensitive if there were no matches.

data83:
This issue has still not been resolved as of:
iPython 6.2.1
iPyKernel 4.8.1
jedi 0.11.1

Issue #10627, opened 2017-05-31
QXcbConnection: Could not connect to display


timothydmorton:
When I create a completely fresh conda envirnoment:
```
conda create --name test ipython matplotlib
```
and then try to `import matplotlib.pyplot` from within ipython, I get the following:
```
In [1]: import matplotlib.pyplot

In [2]: QXcbConnection: Could not connect to display
Aborted
```
There is no problem when I run the same in the vanilla python shell.  Happens for both py2.7.13 (ipython 5.3.0) and py3.6.1 (ipython 6.0.0).

I don't know whether this is an ipython bug or a conda bug, so I raised an issue first with [anaconda](https://github.com/ContinuumIO/anaconda-issues/issues/1806), but I wanted to note it here too.

takluyver:
Can you check that `os.environ['DISPLAY']` is the same in IPython as in the vanilla Python shell?

I guess it might be matplotlib picking a different backend, though I'm not sure why it would do that.

timothydmorton:
Right-- I forgot to mention that I'm running this on a cluster head node that doesn't have a display.  So there is no $DISPLAY set (`os.environ['DISPLAY']` gives a `KeyError` on both shells).

timothydmorton:
And if I explicitly set the backend to `'agg'` I also don't get the error, but I'd like to avoid having to do that everything I launch ipython and want to import something that imports matplotlib.

takluyver:
Are you using the IPython terminal interface or IPython as a kernel in a Jupyter interface? I think IPython as a kernel sets the `MPLBACKEND` environment variable to produce inline plots by default, but that shouldn't happen with IPython in the terminal. And I don't think the inline plots need an X server, anyway. :-/

timothydmorton:
I was just using the ipython terminal interface.  It would seem that if a `DISPLAY` is not available then the backend should default to something that doesn't need it, but that doesn't seem to be happening.

takluyver:
@tacaswell any idea why IPython might be affecting the default backend here?

ccordoba12:
This happens because the default backend for Matplotlib in Anaconda is Qt5. We're considering to change it to Tk to avoid this kind of problems.

tacaswell:
Changing to Tk would still have the same problem.

I think it is showing up immediately with IPython vs python because if pyplot is imported inside of IPython it notices and sets up the event loop integration.   In the plain prompt you would get an issue eventually when the user tried to make a plot.

[...]

PR #10974, opened 2018-01-09
IPython breaks import matplotlib.pyplot when there is no display

boeddeker:
First: This is more an Issue than a PR, I made a PR to show where the error occurs and suggest a workaround.

Related:
#10627
https://github.com/ContinuumIO/anaconda-issues/issues/1806

When someone starts ipython from the command line interface on a machine without a display and executes `import matplotlib.pyplot` the python interpreter breaks
```
In [1]: import matplotlib.pyplot
In [2]: QXcbConnection: Could not connect to display 
Abgebrochen (Speicherabzug geschrieben)
```
Since some libraries that have visualization functions require matplotlib and do not use a lazy import, in my mind IPython should handle this case correctly. At least with a Python exception and not an C/C++ Error.

### DIFF
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -22,9 +22,8 @@ def inputhook(context):
                     _already_warned = True
                     warnings.warn(
                         'The DISPLAY enviroment variable is not set or empty '
-                        'and qt requires this enviroment variable. '
-                        'Deaktivate qt code.\n'
-                        'Backend: {}'.format(QtGui)
+                        'and Qt5 requires this enviroment variable. '
+                        'Deaktivate Qt5 code.'
                     )
                 return
         _appref = app = QtGui.QApplication([" "])
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -13,8 +13,8 @@ def inputhook(context):
     if not app:
         if sys.platform == 'linux':
             try:
-                os.environ['DISPLAY']  # DISPLAY is set
-                assert os.environ['DISPLAY'] != ''  # DISPLAY is not empty
+                # DISPLAY or WAYLAND_DISPLAY is set and not empty
+                assert os.environ.get('DISPLAY') or os.environ.get('WAYLAND_DISPLAY')
             except Exception:
                 import warnings
                 global _already_warned
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -12,18 +12,16 @@ def inputhook(context):
     app = QtCore.QCoreApplication.instance()
     if not app:
         if sys.platform == 'linux':
-            try:
-                # DISPLAY or WAYLAND_DISPLAY is set and not empty
-                assert os.environ.get('DISPLAY') or os.environ.get('WAYLAND_DISPLAY')
-            except Exception:
+            if not os.environ.get('DISPLAY') \
+                    and not os.environ.get('WAYLAND_DISPLAY'):
                 import warnings
                 global _already_warned
                 if not _already_warned:
                     _already_warned = True
                     warnings.warn(
-                        'The DISPLAY enviroment variable is not set or empty '
-                        'and Qt5 requires this enviroment variable. '
-                        'Deactivate Qt5 code.'
+                        'The DISPLAY or WAYLAND_DISPLAY enviroment variable is '
+                        'not set or empty and Qt5 requires this enviroment '
+                        'variable. Deactivate Qt5 code.'
                     )
                 return
         _appref = app = QtGui.QApplication([" "])
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -23,7 +23,7 @@ def inputhook(context):
                     warnings.warn(
                         'The DISPLAY enviroment variable is not set or empty '
                         'and Qt5 requires this enviroment variable. '
-                        'Deaktivate Qt5 code.'
+                        'Deactivate Qt5 code.'
                     )
                 return
         _appref = app = QtGui.QApplication([" "])
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -1,15 +1,32 @@
 import sys
+import os
 from IPython.external.qt_for_kernel import QtCore, QtGui
 
 # If we create a QApplication, keep a reference to it so that it doesn't get
 # garbage collected.
 _appref = None
-
+_already_warned = False
 
 def inputhook(context):
     global _appref
     app = QtCore.QCoreApplication.instance()
     if not app:
+        if sys.platform == 'linux':
+            try:
+                os.environ['DISPLAY']  # DISPLAY is set
+                assert os.environ['DISPLAY'] != ''  # DISPLAY is not empty
+            except Exception:
+                import warnings
+                global _already_warned
+                if not _already_warned:
+                    _already_warned = True
+                    warnings.warn(
+                        'The DISPLAY enviroment variable is not set or empty '
+                        'and qt requires this enviroment variable. '
+                        'Deaktivate qt code.\n'
+                        'Backend: {}'.format(QtGui)
+                    )
+                return
         _appref = app = QtGui.QApplication([" "])
     event_loop = QtCore.QEventLoop(app)
 


takluyver:
I'd say it's up to matplotlib to pick a better backend in that situation.

cc @tacaswell

boeddeker:
I agree that the default backend in this situation should be another one.
But it is still the problem when a user wants to use Qt as backend.
When simulations run on a remote server and use the same code base, the user can hit this bug while debugging.
The current error msg is annoying and does not really help.
It would be better to have a try except around the Qt code, but I do not know if this is possible.

So an exception with information about the cause with a solution is an alternate.

takluyver:
Is it possible to use Qt as a backend in a headless environment at all? I assumed that was just impossible.

boeddeker:
It depends on how you define use.

In my mind, the following code should always work in any python interpreter and any environment.
```python
import matplotlib
matplotlib.use('Qt5Agg')
import matplotlib.pyplot
```
Because it is typical code at the top of files that contain visualization functions, that may be imported, but not executed.
Or at least raise an appropriate error msg.

The following should raise an exception, depending if is possible to render the plot
```python
matplotlib.pyplot.plot(...)
```

dsblank:
We had to work around this issue. We did this:

```python
if (("DISPLAY" not in os.environ) or
    (os.environ["DISPLAY"] == "")):
    if (matplotlib.get_backend() in [
            'module://ipykernel.pylab.backend_inline',
            'NbAgg',
            ]):
        pass  ## Don't change if server has no DISPLAY but is connected to notebook
    else:
        matplotlib.use('Agg')
```

ccordoba12:
> In my mind, the following code should always work in any python interpreter and any environment.

The Qt5 package from Anaconda and the conda-forge one too are linked against X libraries, so I really don't understand how setting the Qt5Agg backend is going to "work" on a headless server.

> We had to work around this issue. We did this

I find @dsblank solution better than the OP one.

boeddeker:
> The Qt5 package from Anaconda and the conda-forge one too are linked against X libraries, so I really don't understand how setting the Qt5Agg backend is going to "work" on a headless server.

The server where I have access use the same operating system as workstations and anaconda is installed on a network drive, so the X libraries are installed.

> I find @dsblank solution better than the OP one.

I find @dsblank solution is a nice workaround, but it shouldn't be in any user code.
What happens when someone uses another jupyter backend? Or the name `module://ipykernel.pylab.backend_inline` changes?
Or someone uses matplotlib to produce image files with a special backend?

So from my viewpoint, the solution should be on the library side (matplotlib, Qt, IPython, Anaconda).

Nevertheless, it is annoying, that there is no python exception that explains what happens wrong and how to solve it.

So on the IPython side, an exception instead a warning in `inputhook` is better?

ccordoba12:
> So on the IPython side, an exception instead a warning in inputhook is better?

No, of course not. I just think that you should use @dsblank code instead of one you proposed in this PR.

> What happens when someone uses another jupyter backend?

I think the only backend that fails when no DISPLAY is available is Qt5Agg. Could you test it, please?

> Or someone uses matplotlib to produce image files with a special backend?

Well, I guess special backends (like Agg) doesn't need DISPLAY. That's what makes them special.

boeddeker:
According to https://stackoverflow.com/a/5092255/5766934 a subset of all matplotlib backends are
```
>>> matplotlib.rcsetup.interactive_bk
['GTK', 'GTKAgg', 'GTKCairo', 'MacOSX', 'Qt4Agg', 'Qt5Agg', 'TkAgg', 'WX', 'WXAgg', 'GTK3Cairo', 'GTK3Agg', 'WebAgg', 'nbAgg']
>>> matplotlib.rcsetup.non_interactive_bk
['agg', 'cairo', 'gdk', 'pdf', 'pgf', 'ps', 'svg', 'template']
```
TkAgg and nbAgg work.
```
import matplotlib; matplotlib.use('TkAgg'); import matplotlib.pylab
import matplotlib; matplotlib.use('nbAgg'); import matplotlib.pylab
```

The others do not work (Missing libs), but all oft them throw an exception (Not a C++ Error).
```python
>>> import matplotlib; matplotlib.use('GTK'); import matplotlib.pylab
ImportError: Gtk* backend requires pygtk to be installed.
# Same for 'GTKAgg', 'GTKCairo', 'MacOSX'
>>> import matplotlib; matplotlib.use('Qt4Agg'); import matplotlib.pylab
ModuleNotFoundError: No module named 'PyQt4'
>>> import matplotlib; matplotlib.use('WX'); import matplotlib.pylab
ImportError: Matplotlib backend_wx and backend_wxagg require wxPython >=2.8.12
>>> import matplotlib; matplotlib.use('WebAgg'); import matplotlib.pylab
NameError: name 'FigureManagerWebAgg' is not defined
```

So on my server, only Ot5 does not work with IPython and that backend is the only one that closes the ipython shell.

> No, of course not. I just think that you should use @dsblank code instead of one you proposed in this PR.

As a solution on the user side, it is a bad idea (As workaround it is fine) because what when I want to use the SVG backend? It will be overwritten.
I want to have a solution and not a workaround.

> I think the only backend that fails when no DISPLAY is available is Qt5Agg. Could you test it, please?

Yes, only Qt5Agg have a C++ Error. The two other interactive backends that run on the server are importable.

> Well, I guess special backends (like Agg) doesn't need DISPLAY. That's what makes them special.

Yes, but there are more than Agg for example SVG. So with above workaround, you need to add all non-interactive backends. 
And `'module://ipykernel.pylab.backend_inline'` is an example that is not listed from matplotlib.
Further the backends are not case sensitive, for example, my notebooks have `nbAgg` as the backend, while dsblank has `NbAgg`. This needs also be handled.

The only thing I want is to get rid of the C++ exception and get instead a Python exception or it is silent.

The next point after this is to figure out where Qt5 is set as default and change it for linux (and mac?) to 
```
if (("DISPLAY" not in os.environ) or
    (os.environ["DISPLAY"] == "")):
        matplotlib.use('Agg')
else:
        matplotlib.use('Qt5Agg')
```

boeddeker:
To clarify, I am not concerning that the Qt5Agg backend does not work, I am concerning that the python interpreter stops to work with a C++ exception and not a Python Exception.

takluyver:
That makes sense. But is the C++ error coming from something we do with Qt, or something matplotlib does?

ccordoba12:
> The next point after this is to figure out where Qt5 is set as default and change it for linux (and mac?)

The backend is set directly by the Matplotlib Anaconda packages:

https://github.com/AnacondaRecipes/matplotlib-feedstock/blob/master/recipe/cfg_qt5agg.patch

> But is the C++ error coming from something we do with Qt, or something matplotlib does?

I think it comes directly from PyQt5,

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,15 +1,32 @@
...
+                os.environ['DISPLAY']  # DISPLAY is set
+                assert os.environ['DISPLAY'] != ''  # DISPLAY is not empty
+            except Exception:
+                import warnings
+                global _already_warned
+                if not _already_warned:
+                    _already_warned = True
+                    warnings.warn(
+                        'The DISPLAY enviroment variable is not set or empty '
+                        'and qt requires this enviroment variable. '
```
ccordoba12: Change qt to Qt5

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,15 +1,32 @@
...
+                        'Deaktivate qt code.\n'
```
ccordoba12: Deaktivate qt code -> Deactivating Qt5 backend

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,15 +1,32 @@
...
+                        'Backend: {}'.format(QtGui)
```
ccordoba12: Is this really necessary?

ccordoba12:
After giving it a bit of thought, I think @boeddeker is a good compromise.

@takluyver?

IPython/terminal/pt_inputhooks/qt.py
boeddeker: No, it comes from my debugging code, the line below breaks.

boeddeker:
> That makes sense. But is the C++ error coming from something we do with Qt, or something matplotlib does?

I forgot to mention that the line below my modifications breaks `QtGui.QApplication([" "])`.
It has something to do with the prompt_toolkit and hooks that are executed.
I am not sure why the code is executed, but with the activated Qt backend (i.e. `import matplotlib.pyplot`) IPython gets this Qt hook.
I only investigated where the error happens, not why the function is called (I went through functions from prompt_toolkit, traitlet and Ipython).

> The backend is set directly by the Matplotlib Anaconda packages:
https://github.com/AnacondaRecipes/matplotlib-feedstock/blob/master/recipe/cfg_qt5agg.patch

Thanks for the link, it saves me many searches. I was hoping the backend is set in a python file and not in an ini file.

Depending what the IPython developers decide (Exception/Warning/Silent) it may not be necessary to change the default backend on a server.

IPython/terminal/pt_inputhooks/qt.py
boeddeker: I am not sure if `backend` is the correct word. Because this code prevents IPython to exit the interpreter. This position in the code should normally never reached, but it should be there to prevent the interpreter to exit. It simply disables some code.

boeddeker:
Any plans to merge this or fix this at another position? 
It is always annoying when the interactive interpreter exits.

ccordoba12:
+1 from me to merge.

takluyver:
I think we need to find a better way of checking whether we can use a GUI. $DISPLAY is for X, but Qt can also run on top of Wayland, which doesn't need that environment variable. I'm already running Wayland on Fedora - at the moment DISPLAY is still set for the XWayland compatibility layer, but I bet that before long someone will be running Wayland without the compatibility layer, and then the check is wrong.

Ideally Qt itself should provide some way of checking if it has a display to draw to, but at a minimum we should check:

```python
os.environ.get('DISPLAY') or os.environ.get('WAYLAND_DISPLAY')
```

boeddeker:
Ok, I added WAYLAND_DISPLAY.

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,15 +1,31 @@
...
+_already_warned = False
 
 def inputhook(context):
     global _appref
     app = QtCore.QCoreApplication.instance()
     if not app:
+        if sys.platform == 'linux':
+            try:
+                # DISPLAY or WAYLAND_DISPLAY is set and not empty
+                assert os.environ.get('DISPLAY') or os.environ.get('WAYLAND_DISPLAY')
```
tacaswell: Why not use `if...else` instead of `assert` here?

tacaswell:
I am :+1: on this being fixed at the IPython level.  At the Matplotlib level if the user asks to use qt5agg then we should respect that (even if it causes exceptions).

Similar logic should be put into all of the GUI input hooks?

IPython/terminal/pt_inputhooks/qt.py
boeddeker: I find the logic easier. (There should be no runtime difference.)
```python
if not (os.environ.get('DISPLAY') or os.environ.get('WAYLAND_DISPLAY')):
    ...
# vs 
try:
    assert os.environ.get('DISPLAY') or os.environ.get('WAYLAND_DISPLAY'))
except Exception:
    ...
````
But if you prefare `if`, I can change the code.

Comment (takluyver) [reply IPython/terminal/pt_inputhooks/qt.py]: I'd prefer an `if`; it feels strange to throw an error only to immediately catch it again.

File: IPython/terminal/pt_inputhooks/qt.py [hunk 5]
```diff
@@ -1,15 +1,31 @@
...
+            except Exception:
+                import warnings
+                global _already_warned
+                if not _already_warned:
+                    _already_warned = True
+                    warnings.warn(
+                        'The DISPLAY enviroment variable is not set or empty '
```
takluyver: Don't forget to update the message.

takluyver:
I think the issue that made this particularly a problem with Qt is that the error crashes the process. I don't know if the integration with other toolkits fails as hard in a headless environment, but it would definitely be good if someone could check.

isaacovercast:
Any chance this will get backported to the 5.x branch? In my case the behavior was very difficult to debug (ipyparallel engines failing silently was driving me insane). The fix seems reasonable, simple, and it should provide more transparent behavior. Also, not sure if anybody else saw this, but in my case ipyparallel engines were actually segfaulting D: 

Hey, thanks for all your awesome work on ipython! It's a really really awesome and useful!

IPython/terminal/pt_inputhooks/qt.py
boeddeker: ok, I changed it to if

IPython/terminal/pt_inputhooks/qt.py
boeddeker: Ok, done

boeddeker:
Ok, I applied the comments

IPython/terminal/pt_inputhooks/qt.py
```diff
@@ -1,15 +1,29 @@
...
 def inputhook(context):
     global _appref
     app = QtCore.QCoreApplication.instance()
     if not app:
+        if sys.platform == 'linux':
+            if not os.environ.get('DISPLAY') \
+                    and not os.environ.get('WAYLAND_DISPLAY'):
+                import warnings
+                global _already_warned
+                if not _already_warned:
```
takluyver: I don't think we need this check - the warnings module already takes care of only printing a warning the first time it occurs.

boeddeker:
```python
In [1]: import matplotlib.pyplot

.../ipython/IPython/terminal/pt_inputhooks/qt.py:22: UserWarning: The DISPLAY or WAYLAND_DISPLAY enviroment variable is not set or empty and Qt5 requires this enviroment variable. Deactivate Qt5 code.
  'The DISPLAY or WAYLAND_DISPLAY enviroment variable is '
In [2]: 

.../ipython/IPython/terminal/pt_inputhooks/qt.py:22: UserWarning: The DISPLAY or WAYLAND_DISPLAY enviroment variable is not set or empty and Qt5 requires this enviroment variable. Deactivate Qt5 code.
  'The DISPLAY or WAYLAND_DISPLAY enviroment variable is '
In [2]: 

.../ipython/IPython/terminal/pt_inputhooks/qt.py:22: UserWarning: The DISPLAY or WAYLAND_DISPLAY enviroment variable is not set or empty and Qt5 requires this enviroment variable. Deactivate Qt5 code.
  'The DISPLAY or WAYLAND_DISPLAY enviroment variable is '
In [2]: 
```

I do not know how the warning module works, but at least on my system it does not work properly. When I disable the `_already_warned` I get above output (I pressed enter) with multiple warnings. When I activate it, I only get one warning.

Carreau:
Ok, I reviewed the discussion, and this is just adding warnings, and deactivating Qt. There is the small disagreement about warning showing everytime but I'm favor of moving forward with at least an explanation for users.

We can still update the behavior later. 
Merging. 

> Any chance this will get backported to the 5.x branch?

I'm marking as backport, just to see if this will apply cleanly, if not someone will have to send a PR against 5.x.

tacaswell:
As a late follow up, as of 3.0 Matplotlib does a better job of not, by default, selecting a GUI backend on a headless machine (but will still respect an rcparams file that asks for it).

PR #10978, opened 2018-01-11
Add display class for syntax-highlighted code

takluyver:
Preparing some workshop materials, I've found myself wanting a better way to display a separate code file in a notebook.

* `!cat` doesn't highlight code, and it only works if cat is available.
* `%pycat` displays code in the pager, not as regular output, and its highlighting is specific to Python code.

There are many ways this could be extended - to add more output formats, or to let the user pick a pygments style, for instance. But I think what's there probably meets the 80% use case.

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -5,10 +5,10 @@ Authors : MinRK, gregcaporaso, dannystaple
 from os.path import exists, isfile, splitext, abspath, join, isdir
 from os import walk, sep
 
-from IPython.core.display import DisplayObject
+from IPython.core.display import DisplayObject, TextDisplayObject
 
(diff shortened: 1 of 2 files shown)


takluyver:
I did have another idea how to achieve this: send plain text with some metadata identifying the language, and let the frontend do the highlighting if it can: https://github.com/jupyter/nbconvert/issues/363

That proposal is more elegant: it only sends the data once, it allows frontends to highlight in formats the kernel many not know about, and it works nicely for converting to formats like rst and markdown, which don't include highlighted code themselves, but may be rendered into a format which does.

However, this idea can be implemented in IPython and be useful immediately, whereas the other one requires changes to several different projects.

Carreau (approved):

PR #10993, opened 2018-01-24
fix execution count when custom exception is caught during compile

minrk:
e.g. `set_custom_exc(SyntaxError)`

was causing history thread error due to repeated lines

consolidate handling of compilation errors to make deviations less likely in the future

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2752,23 +2752,25 @@ class InteractiveShell(SingletonConfigurable):
 
             with self.display_trap:
                 # Compile to bytecode
+                ast_error = None
                 try:
                     code_ast = compiler.ast_parse(cell, filename=cell_name)
                 except self.custom_exceptions as e:
                     etype, value, tb = sys.exc_info()
                     self.CustomTB(etype, value, tb)
-                    return error_before_exec(e)
+                    ast_error = e
                 except IndentationError as e:
                     self.showindentationerror()
-                    if store_history:
-                        self.execution_count += 1
-                    return error_before_exec(e)
+                    ast_error = e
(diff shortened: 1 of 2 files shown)


IPython/core/tests/test_interactiveshell.py
```diff
@@ -922,3 +922,14 @@ def test_no_dep(self):
         with tt.AssertNotPrints("I AM  A WARNING"):
             ip.run_cell("wrn()")
         ip.run_cell("del wrn")
+
+
+def test_custom_exc_count():
```
willingc: Just verifying my understanding of the test:
- set a custom exception handler for syntax errors
- grab cell execution count before running the cell with a syntax error
- run a cell with a syntax error
- restore default exception handler
- verify the hook to the handler was called only once
- verify the execution count was incremented by one

willingc (approved):
@minrk Looks good to me. I'll leave for you or @takluyver to merge.

IPython/core/interactiveshell.py
```diff
@@ -2752,23 +2752,25 @@ def error_before_exec(value):
...
-                    return error_before_exec(e)
+                    ast_error = e
                 except (OverflowError, SyntaxError, ValueError, TypeError,
                         MemoryError) as e:
                     self.showsyntaxerror()
+                    ast_error = e
+
+                if ast_error is not None:
                     if store_history:
                         self.execution_count += 1
```
takluyver: Would it make sense to put these two lines inside the `error_before_exec` closure?

takluyver: ping @minrk

minrk: Sorry about that. Updated.

Issue #10926, opened 2017-11-28
[Completions] dict key completion occluded by kitchen sink


JeremySikes73:
I find the following IPython behaviour unfortnuate:

```
In [1]: d={'a':1,'bdfglkjdfg':2}
In [2]: d['<TAB>
```

Returns not only the keys, but also a list of every magic and python global in the environment, ordered alphabetically. The result is that the signal (dictionary keys) is completely swamped by the noise.
 
Granted the general case is complex, but the user intent in this case is very clear. In cases where the goal is not to spare keystrokes but to discover/explore, this "everything,everything AND in alphabetical order! " makes the completion list of little use.

By default, IPython merges https://github.com/ipython/ipython/blob/474d48d9541e3efd3b8d4a6ea6a5276f7751e977/IPython/core/completer.py#L1002

from its multiple completion sources https://github.com/ipython/ipython/blob/474d48d9541e3efd3b8d4a6ea6a5276f7751e977/IPython/core/completer.py#L1105

(Whoa, awesome new inline snippets by Github)

the `dict_key_matches` source does a fair bit of work to figure out its context, but arguably, this context-sensitivity belongs higher up in the call stack. Sources (filenames,python, dict keys,...) should be included/excluded/result-ordered according the line context. In the example above I would expect either that only dict keys would be displayed, or that the keys would somehow be visually singled out as the likely candidates.

Some completion systems list completions by section/source. I even seem to recall IPython did this one, but I might be wrong.

flying-sheep:
When completing, all of the matchers do their thing:

https://github.com/ipython/ipython/blob/474d48d9541e3efd3b8d4a6ea6a5276f7751e977/IPython/core/completer.py#L1105-L1111

… as well as jedi. The jedi completions and matcher completions are emitted:

https://github.com/ipython/ipython/blob/474d48d9541e3efd3b8d4a6ea6a5276f7751e977/IPython/core/completer.py#L1863-L1865

https://github.com/ipython/ipython/blob/474d48d9541e3efd3b8d4a6ea6a5276f7751e977/IPython/core/completer.py#L1884-L1885

The jedi completer knows not to return anything when we’re in a string (https://github.com/ipython/ipython/pull/10996#issuecomment-378288569). I think the problem is that the “magic” and “python” matchers get run even if we’re in a string.

PR #10996, opened 2018-01-26
Detect string completion when no syntax error

bjoernpollex:
Previously the behaviour of completing would be different for `d['`
compared to `d['']`, because disabling of `jedi` relied on detecting a
syntax error, which isn't present in the latter. With this fix, this is
correctly detected.

Unfortunately, the change is very difficult to test, because there is
no easy way to test for _absence_ of any `jedi` completions.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1355,7 +1355,7 @@ class IPCompleter(Completer):
                     raise ValueError("Don't understand self.omit__names == {}".format(self.omit__names))
 
         interpreter = jedi.Interpreter(
-            text, namespaces, column=cursor_column, line=cursor_line + 1)
+            text[:cursor_column], namespaces, column=cursor_column, line=cursor_line + 1)
         try_jedi = True
 
         try:


bjoernpollex:
This fixes an issue described [here][1], that was ultimately tracked down to IPyhon.

[1]: https://github.com/ipython/ipykernel/issues/299

takluyver:
cc @Carreau 

I suspect that `cursor_column` is only right if it's on the first line of a cell - otherwise we need to translate the line+column position into an index (which we have a function to do somewhere).

takluyver:
Thanks, I've taken this idea and made #11063, which handles multiple lines better (although there's another issue with completing inside a compound statement, see #11064 ).

takluyver:
#11063 was merged, so I'll close this, but thanks @bjoernpollex for finding where the fix was needed!

jmozmoz:
Unfortunately, the problem with completion of keys of a dict still exists, even after updating to ipython 6.3.0.

    In [1]: a = {'x':1, 'y':2}
    a['<press TAB>


shows the content of the current working directory, ipython commands (and also the keys). So the original issues https://github.com/ipython/ipykernel/issues/299 an #10926 are still not fixed.

takluyver:
Is your Jedi version up to date?

jmozmoz:
> Is your Jedi version up to date?
I think so:
```
$ pip show jedi
Name: jedi
Version: 0.11.1
Summary: An autocompletion tool for Python that can be used for text editors.
Home-page: https://github.com/davidhalter/jedi
Author: David Halter
Author-email: davidhalter88@gmail.com
License: MIT
Location: c:\opt\python36\lib\site-packages
Requires: parso
```

takluyver:
Oh, I see what's going on. The Jedi matcher is getting disabled correctly, but the completions that are upsetting you are not the ones from jedi - magic and filename completions are generated separately.

As @JeremySikes73  said on #10926:

> the dict_key_matches source does a fair bit of work to figure out its context, but arguably, this context-sensitivity belongs higher up in the call stack. Sources (filenames,python, dict keys,...) should be included/excluded/result-ordered according the line context.

At present, it's up to each individual completer to decide when it's relevant. What we really need is some piece that runs before them, figures out what context is relevant to a completion, and which completers should be used (e.g. `a['<TAB>` -> only call the key completer).

Unfortunately, this is hard for a variety of reasons: our completion code is a thicket of code that people have added over the years to fix specific problems they're seeing, and there's no formal grammar or parser for IPython special syntax (like `%magic` functions). And every time anything changes in completions, people queue up to complain about it, because completion is *meant* to be magic that works out what you want.

jmozmoz:
@takluyver thank you for your explanation. So which of the related bugs should stay open/be reopened to trace any work on implementing this magic?

takluyver:
#10926 seems like what you're after, and it's still open, so let's use that one.

Issue #10988, opened 2018-01-20
IPython doing case-insensitive completion


gabriellima:
Ipython is doing case-insensitive completion when using Jedi.

If I do `%config Completion.use_jedi = False` I "temporaly" fix this issue. 
Or permanently fix it updating ipython_config.py

May be related to #10969 

![image](https://user-images.githubusercontent.com/487779/35187162-7c67a4de-fe06-11e7-8bde-070fc0342142.png)

Environment:
```
Python 3.5.1
Ipython 6.2.1
ipykernel 4.8.0
jedi 0.11.1
```

takluyver:
@Carreau is this something we can configure through Jedi? It seems to be hitting a few people (also #10728).

Carreau:
There seem to be a `settings.case_insensitive_completion = True` when grepping through the source code. So it can probably be set.

cottrell:
I might PR this. 

https://github.com/cottrell/ipython/commit/c07e32e1f996966e1322b679f6734816cfae567b

palxex:
...what should I do if I WANT case insensitive completion? For macOS and Windows user that's better, since the filesystem is case-insensitive too.
I think its better introduce a option that configureable in ipython-settings.py, in order to control it.

MrMino:
If someone wants case-insensitive completion, see https://github.com/ipython/ipython/issues/10728#issuecomment-921419106.

The case insensitivity for `%cd` magic has been started within #11850, but it's probably something that is going to (if not already has been) superseeded by Jedi's filename completion.

Issue #10728, opened 2017-08-02
tab completion now ignores case?


cottrell:
I looks like the new tab completion now ignores case. In the old versions of IPython this was not the case. Not a big deal really, but probably should be reverted if not intended.

 $ cat j.py
def this_one():
    pass

def this_ONE():
    pass

In [2]: j.this_O<TAB>
                 this_ONE()
                 this_one()

Carreau:
Yes, I believe that is recent. You can also have a look at https://github.com/ipython/ipython/pull/10537/ which also add some fuzzy matching. I don't remember which PR made completion case insensitive though.

cottrell:
Anyone know how to turn this off? I would vote for case sensitive being default given past behaviour and usage patterns in python.

cottrell:
This seems to get rid of the bad column displays where it doesn't show you all the options.

`c.TerminalInteractiveShell.display_completions = 'readlinelike'
`

pitrou:
I would also like to turn this off. False positives due to case insensitive matching are annoying as they lead to more typing.

cottrell:
https://github.com/cottrell/ipython/commit/c07e32e1f996966e1322b679f6734816cfae567b

GF-Huang:
4 years, any progress?

GF-Huang:
> I would also like to turn this off. False positives due to case insensitive matching are annoying as they lead to more typing.

How did you turn it on?

GF-Huang:
I want to turn it on, how to do that???????????????

MrMino:
@GF-Huang Instead of duplicating your comments and abusing the question mark key, you could've just look into commits linked in this issue:

https://github.com/cottrell/ipython/commit/2903df30ec5ce51ac0abcdf12a90ce5d5fb380c1#diff-f10e4b6051cca1d8ce27936518d6908de43b5fd861902cd3923a4342e8d4fc5aR146-R147

![image](https://user-images.githubusercontent.com/6691643/133717259-9049c720-fa1c-4a6a-88e4-44518648f34c.png)

⬇

![image](https://user-images.githubusercontent.com/6691643/133717357-f33bfcf8-544e-40a5-95a0-f934b3ef6443.png)

Simple stuff. Read the code. Don't expect others to do it for you. Don't be a nuisance.

MrMino:
I think this issue has been settled. The status-quo has been agreed upon in #10999, if someone wants to challange it - please raise a separate issue.

[...]

PR #10999, opened 2018-02-04
Change defaults to be consistent with original defaults and REPL perf…

cottrell:
Happy to hear feedback on why whatever is currently going on is better but seems like generally it is best to not dramatically change the defaults. <dot>.<tab> is one of the most important operations in python (IPython) from a human perspective so everyone is quite sensitive about this stuff.

Case-insensitivity break tab-completion in the case when there is one case-sensitive choice but several case-insensitive choices.  Would be interesting to consider second <tab> to be case insensitive or something like that. Then you don't break that UX on the first tab completion.  

…ormance. #10728 #10988

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -144,6 +144,7 @@ skip_doctest = True
 
 try:
     import jedi
+    jedi.settings.case_insensitive_completion = False
     import jedi.api.helpers
     import jedi.api.classes
     JEDI_INSTALLED = True


takluyver:
I'm happy with the bit setting jedi to do case-sensitive completion, but I'd rather leave the default presentation as 'multicolumn'.

cc @Carreau

cottrell:
K, I've just done the case-sensitive one-line change.

takluyver:
Thanks, merging. We may want to refine this further (e.g. falling back to case-insensitive search when we find no matches), but it sounds like this will be an improvement for most users.

That's always a dangerous thing to say. If there are people who don't like this change, they won't be complaining yet. 🤞 Fingers crossed...

cottrell:
Ah, yes falling back in the case of no-matches ... I will try to remember to have a look if it is quick to do. This functionality could actually save me time when teaching people new things.

PR #11014, opened 2018-02-18
making prompt_tolkit history search configurable

### LINKED ISSUE #9961: Make prompt_toolkit history search configurable
We currently hard-code it on, but people may want to disable it.

https://github.com/ipython/ipython/blob/8008eda4a856dd2e8c6dcae21d651d36ce52c683/IPython/terminal/interactiveshell.py#L245

ghost:
resolve #9961 
make use of os.environ to read from an envvar called enable_history_search which must be true or false, and if it's not present the default is set to True as it was before.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -57,8 +57,6 @@ _style_overrides_linux = {
             Token.OutPromptNum: '#bb0000 bold',
 }
 
-
-
 def get_default_editor():
     try:
         return os.environ['EDITOR']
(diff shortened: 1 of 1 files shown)


willingc:
Thanks @ShailynOrtiz for the PR. We'll take a look and review this. 

cc/ @minrk @takluyver Submitted at the PyCaribbean sprint organized by @jzf2101

willingc (commented):
Thanks @ShailynOrtiz. This looks good to me. Leaving for @takluyver or @minrk to merge.

IPython/terminal/interactiveshell.py
```diff
@@ -271,7 +269,11 @@ def patch_stdout(**kwargs):
                             history=history,
                             completer=IPythonPTCompleter(shell=self,
                                                     patch_stdout=patch_stdout),
-                            enable_history_search=True,
+                            enable_history_search=(
```
minrk: Thanks! Can you make this a `Bool(True, config=True)` trait rather than an environment variable? That's how we make things configurable in IPython. See `highlight_matching_brackets` above on L194 for a similar example (it has a default of False, but this should default to True). Then here it can be `enable_history_search=self.enable_history_search`.

minrk: Perfect, thanks!

jzf2101:
Thanks! @ghost

willingc:
Thanks!

PR #11018, opened 2018-02-20
Set stacklevel for warnings

### LINKED ISSUE #9981: Audit our warnings
Our warnings should likely be audited to check the following (mostly for deprecation warnings):

Warnings messages should try to mention: 
- Since which version something is deprecated.
- Whether it currently has an effect (or not), and in both case by what it could (should?) be replaced.
- a `stacklevel` value (except maybe at import time, in order to appear at in the right stackframe)

That should be easy for someone to get into the code.

souravsingh:
Fixes #9981

### DIFF
--- a/IPython/consoleapp.py
+++ b/IPython/consoleapp.py
@@ -7,6 +7,6 @@ Shim to maintain backwards compatibility with old IPython.consoleapp imports.
 from warnings import warn
 
 warn("The `IPython.consoleapp` package has been deprecated. "
-     "You should import from jupyter_client.consoleapp instead.")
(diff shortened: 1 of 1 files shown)


jzf2101:
Is this ready for review?

takluyver:
Setting the stacklevel makes sense, but I'd rather not use DeprecationWarning, because those are invisible in many common circumstances. Also, we have several other deprecated modules in the IPython package - it would be good to keep them consistent.

Carreau:
> Setting the stacklevel makes sense,

It may not as import machinery has an undefined stack number  across python versions.
Still updated to remove `DeprecationWarning` and added since when it was deprecated.

Carreau (approved):

Issue #11032, opened 2018-03-06
Autocomplete regression in IPython 6.x vs 5.x


jni:
Autocomplete is now too smart for its own good, and has actually dramatically reduced in utility as far as I'm concerned. Specifically, IPython 5 was perfectly able to find directories when using tab completion:

<img width="347" alt="screen shot 2018-03-06 at 1 58 09 pm" src="https://user-images.githubusercontent.com/492549/37012021-b189b40a-2146-11e8-8ca0-910e0ec58919.png">

IPython 6, in contrast, "helpfully" gives me additional options:

<img width="487" alt="screen shot 2018-03-06 at 1 44 44 pm" src="https://user-images.githubusercontent.com/492549/37012042-c773d2d2-2146-11e8-8d55-8b80a210f90e.png">

(To clarify, I typed the same thing in both instances, `directory = '~/pr<TAB>'`.)

Is a fix a config switch away, or is this a deeper issue?

takluyver:
(sigh) more issues with the Jedi integration. I think #11035 should fix this case.

adrn:
In case it's useful, just adding a note that I see the same issue with ipython 6.2.1, jupyter 4.4.0, jedi 0.11.1

![image](https://user-images.githubusercontent.com/583379/37868633-4341362e-2f80-11e8-9987-b008f0444947.png)

Issue goes away when I add `c.IPCompleter.use_jedi = False` to my `ipython_config.py`.

PR #11035, opened 2018-03-07
Fix check for being inside string for jedi completions

takluyver:
Completing something like `a = "/home/<tab>` gives a token ` "/home"` starting with a space, which wasn't being caught.

Closes gh-11032

@Carreau do you think you'll have time soon to get back to working on the jedi integration? I like the idea, but it seems to be making the user experience worse and the code feels hackish (we're using a private method to access Jedi's parse tree to decide when to ignore its suggestions entirely). If we can't commit to making it work well, I think we should drop it and go back to relying on the battle-tested IPython completion machinery.


takluyver:
Ping @Carreau - do you want to look at this before I merge it?

Issue #11026, opened 2018-03-01
infinite loop 


kmike:
When I try to execute this code: https://gist.github.com/kmike/91ef29a3fd2b94ff8b47a5a9ebfa7be2 in an IPython cell (both terminal and notebook), IPython goes to an infinite loop. Code just defines a nested dict; there are unicode symbols. Here is the traceback on interruption:

```
ERROR:tornado.general:Uncaught exception in ZMQStream callback
Traceback (most recent call last):
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/zmq/eventloop/zmqstream.py", line 432, in _run_callback
    callback(*args, **kwargs)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/tornado/stack_context.py", line 277, in null_wrapper
    return fn(*args, **kwargs)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 283, in dispatcher
    return self.dispatch_shell(stream, msg)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 233, in dispatch_shell
    handler(stream, idents, msg)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 399, in execute_request
    user_expressions, allow_stdin)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/ipykernel/ipkernel.py", line 208, in do_execute
    res = shell.run_cell(code, store_history=store_history, silent=silent)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/ipykernel/zmqshell.py", line 537, in run_cell
    return super(ZMQInteractiveShell, self).run_cell(*args, **kwargs)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 2654, in run_cell
    cell = self.input_transformer_manager.transform_cell(raw_cell)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/IPython/core/inputsplitter.py", line 659, in transform_cell
    self.push(cell)
  File "/Users/kmike/envs/deep-deep/lib/python3.6/site-packages/IPython/core/inputsplitter.py", line 702, in push
    out = self.push_line(line)
[...]

Issue #10972, opened 2018-01-07
ipython incorrectly suppresses backslashes at end of lines in raw multiline strings


anntzer:
Hopefully the following screenshot should illustrate the issue (note the different interpretation of the raw multiline string by python and ipython):
![screenshot_20180106_172634](https://user-images.githubusercontent.com/1322974/34645610-13c14458-f307-11e7-98ee-def430226e79.png)

Issue #10272, opened 2017-02-09
Simplify InputSplitter


Carreau:
InputSplitter was originally developed with a stateful API in mind mostly because Readline was feeding input one line after another. Now that we do not use readline, we should be able to get rid of its statefulness.

That's no hurry, it's mostly code cleaning. Assigning @takluyver, ~~but as marking that we welcome anyone that would like to try it.~~ 

We might want a deprecation period as well before actually removing.

takluyver:
This is one that I actually don't want to invite people to try. That code is old, tricky and critical - exactly the sort of thing where someone with good intentions could come in to do a helpful bit of cleanup and cause subtle breakage. I asked to be assigned on it to gently ward people off. I'd rather one of the core team did this - and as I'm the most familiar with that part of the code, I'm the first choice to look at it. ;-)

Carreau:
Fair enough. I stoke the comment i posted initially  then.

Carreau:
Closing, has this have been fixed by #11041.

PR #11041, opened 2018-03-11
New input transformation machinery

takluyver:
See #10272. This is a major change which I expect to be the trigger for IPython 7.0.

## Main changes

* `IPython.core.inputtransformer2` is added, containing the new machinery to transform input before we parse it.
* `IPython.core.inputsplitter` and `IPython.core.inputtransformer` are deprecated. We can't remove them immediately because nbconvert relies on them.
* `IPython.utils.tokenize2` is removed. This was a copy of the stdlib `tokenize` module with a couple of issues patched. One of them ([bpo 12691](https://bugs.python.org/issue12691)) was fixed in Python 3.4.1, which we will require for IPython 7 (see #10833). The other ([bpo 17061](https://bugs.python.org/issue17061)) turned out to be intended behaviour, so I've handled it another way in inputtransformer2.

## Rationale

This will break any code that extends the current input transformation machinery (and the interface was documented), and any change this big is bound to cause some unexpected bugs. So why should we do this?

* Maintainability: I wrote the current input transformation machinery when I'd just learned about coroutines, and I think I got carried away. I don't think anyone else wants to touch that code, and I'm wary of doing so myself. I don't think I can make the code to do this *simple* - it's just too complex a task - but I believe the new code is more quite a bit *simpler* than what it replaces.
  * Code size is not a good measure of complexity, but FWIW, 550 lines in inputtransformer2 replace 1300 lines in inputsplitter+inputtransformer.
* Weird quirks: I hope that the new framework, which uses tokens to recognise our special syntax, will be more robust against corner cases involving comments, multiline strings, etc. #10972 appears to be fixed. Hopefully #8987 is too, though I can't currently reproduce it on master.
* Performance: despite some improvements to the existing code in #10898, it still gets pathologically slow with long enough code, as discussed in #11026. That latter issue contains a sample of ~1500 lines of plain Python code (a giant dictionary literal), which takes ~20s to run through inputsplitter on master before it can be executed. The new transformation machinery in this PR processes it in about 30ms, a speedup of three orders of magnitude. :sunglasses: 

## Limitations

* Worst case performance: Every piece of special syntax requiring token based transformation (`%line_magics`, `!system commands`, `help?`, etc.) results in another tokenisation pass after the transformation. So a large code cell with many instances of special syntax could take longer to parse than it currently does - although the current code isn't very efficient either (see above). I haven't tried to measure this. My intuition is that large cells tend to be regular Python code, as in #11026, in which case the new implementation can comfortably beat the old one.
* Extensibility of 'is this complete?' checking: the old code allows transformers to 'accumulate' code to indicate that they'd like further lines of input from the user. The new system currently has no equivalent. It's a source of complexity which I'm not sure if anyone actually needs, so I propose we skip it for now and see how many people complain.


Carreau:
I like simplifications in general ! I'll trust you on this. 
I'd like to think about how that would/could be integrated with a non-blocking sans-io IPython.

takluyver:
As this is a breaking change, and the system I built last time round was rather convoluted, I hope someone can look at it a bit rather than trusting me. ;-) A good place to start might be the [updated doc page](https://github.com/takluyver/ipython/blob/96b86fb221893f13ba26644098edc4a20cdc0669/docs/source/config/inputtransforms.rst) about the transformation framework.

I think our input transformation machinery has always used something like a Sans-IO design, even before that was a thing.

willingc:
@takluyver I've pulled it down and am running it locally. It will take a while to review. Big PR :smile:

takluyver:
Thanks Carol! It is a big PR.

takluyver:
ping @willingc were you able to keep running with this branch? Did you encounter any anomalies?

willingc:
@takluyver I still need to loop back to this. I haven't run it very much. Will try to kick the tires some more this afternoon.

takluyver:
Thanks Carol :-)

jasongrout:
Sage uses the input transformation framework, so CC a few people from Sage: @williamstein, @jdemeyer (feel free to CC others too)

IPython/core/inputsplitter.py
```diff
@@ -1,4 +1,6 @@
-"""Input handling and transformation machinery.
+"""DEPRECATED: Input handling and transformation machinery.
+
+This module was deprecated in IPython 7.0, in favour of inputtransformer2.
```
Carreau: Emit deprecation warnign at import time ?

IPython/core/inputtransformer.py
```diff
@@ -1,16 +1,18 @@
...
+This module was deprecated in IPython 7.0, in favour of inputtransformer2.
 
 This includes the machinery to recognise and transform ``%magic`` commands,
 ``!system`` commands, ``help?`` querying, prompt stripping, and so forth.
 """
 import abc
 import functools
 import re
+import tokenize
+from tokenize import generate_tokens, untokenize, TokenError
```
Carreau: generate_token is deprecated/not public. Maybe figure out how to use tokenize directly ? Or complainc to CPython to undeprecate it cause it is annoying to use.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+    If the first line starts with a spaces or tabs, the same whitespace will be
+    removed from each following line.
+    """
+    m = _indent_re.match(lines[0])
+    if not m:
+        return lines
+    space = m.group(0)
+    n = len(space)
+    return [l[n:] if l.startswith(space) else l
+            for l in lines]
```
Carreau: I'm just going to write some +1/ok here and there to have a mark for myself as to what I have already reviewed.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+        ``...`` is a valid Python expression in Python 3, so shouldn't be stripped.
+    
+    If initial_re and prompt_re differ,
+    only initial_re will be tested against the first line.
+    If any prompt is found on the first two lines,
+    prompts will be stripped from the rest of the block.
+    """
+    def __init__(self, prompt_re, initial_re=None):
+        self.prompt_re = prompt_re
+        self.initial_re = initial_re or prompt_re
```
Carreau: Potential improvement. let use pass string and compile is is str?
Can be done later, just thinking at loud.

IPython/core/inputtransformer.py
takluyver: I did ask them to make it official: https://bugs.python.org/issue12486#msg313591

It's been there unchanged for quite a few releases, so I doubt it's in imminent danger of removal. I'll try opening a PR to make it public.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+    # We don't try to support multiple special assignment (a = b = %foo)
+    paren_level = 0
+    for i, ti in enumerate(token_line):
+        s = ti.string
+        if s == '=' and paren_level == 0:
+            return i 
+        if s in '([{':
+            paren_level += 1
+        elif s in ')]}':
+            paren_level -= 1
```
Carreau: How does that behave with unmatched parenthesis ? Does ti need to behave well ?

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+def cell_magic(lines):
+    if not lines[0].startswith('%%'):
+        return lines
+    if re.match('%%\w+\?', lines[0]):
+        # This case will be handled by help_end
+        return lines
+    magic_name, _, first_line = lines[0][2:-1].partition(' ')
+    body = ''.join(lines[1:])
+    return ['get_ipython().run_cell_magic(%r, %r, %r)\n'
+            % (magic_name, first_line, body)]
```
Carreau: Ok.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+        self.initial_re = initial_re or prompt_re
+
+    def _strip(self, lines):
+        return [self.prompt_re.sub('', l, count=1) for l in lines]
+
+    def __call__(self, lines):
+        if self.initial_re.match(lines[0]) or \
+                (len(lines) > 1 and self.prompt_re.match(lines[1])):
+            return self._strip(lines)
+        return lines
```
Carreau: ok

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+    Uses 0-indexed line numbers.
+    """
+    end_line = start_line
+    while lines[end_line].endswith('\\\n'):
+        end_line += 1
+        if end_line >= len(lines):
+            break
+    return end_line
+
+def assemble_continued_line(lines, start: Tuple[int, int], end_line: int):
```
Carreau: I see you are trying types annotations :-)

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+    """Assemble pieces of a continued line into a single line.
+
+    Uses 0-indexed line numbers. *start* is (lineno, colno).
+    """
+    parts = [lines[start[0]][start[1]:]] + lines[start[0]+1:end_line+1]
+    return ' '.join([p[:-2] for p in parts[:-1]]  # Strip backslash+newline
+                    + [parts[-1][:-1]])         # Strip newline from last line
```
Carreau: Hum, as is, out of context this one give me a headache. I guess the example later will help. Maybe one example in the docstring ?

I need to re-read this one later.

IPython/core/inputtransformer2.py
Carreau: update from `Find the first assignment` to `return the index of the first `=` token in an assignemnt line...

I _guess_ the parenthesis thing allow unpacking `(a,b) = foo` ?

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+
+    def sortby(self):
+        return self.start_line, self.start_col, self.priority
+
+    def __init__(self, start):
+        self.start_line = start[0] - 1   # Shift from 1-index to 0-index
+        self.start_col = start[1]
+
+    def transform(self, lines: List[str]):
+        raise NotImplementedError
```
Carreau: Hum, Ok. Will reread that one first pass is done.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+        
+        Returns (line, column) of the % if found, or None. *line* is 1-indexed.
+        """
+        for line in tokens_by_line:
+            assign_ix = _find_assign_op(line)
+            if (assign_ix is not None) \
+                    and (len(line) >= assign_ix + 2) \
+                    and (line[assign_ix+1].string == '%') \
+                    and (line[assign_ix+2].type == tokenize.NAME):
+                return cls(line[assign_ix+1].start)
```
Carreau: You _may_ be able to make that more readable by using multiple ifs and a `continue` ?

The other question that come to mind about `Find the first magic assignment`, is what if there is multiple magics assignments ?

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+    def find(cls, tokens_by_line):
+        """Find the first system assignment (a = !foo) in the cell.
+
+        Returns (line, column) of the ! if found, or None. *line* is 1-indexed.
+        """
+        for line in tokens_by_line:
+            assign_ix = _find_assign_op(line)
+            if (assign_ix is not None) \
+                    and (len(line) >= assign_ix + 2) \
+                    and (line[assign_ix + 1].type == tokenize.ERRORTOKEN):
```
Carreau: premature opt is the root of all evil, still the following seem like it _may_ be a repetitive pattern:
```
        for line in tokens_by_line:
            assign_ix = _find_assign_op(line)
            if (assign_ix is not None) \
                    and (len(line) >= assign_ix + 2) \
					and (<token check>):
```

If this is used by other external transformers, maybe figure out how to extract this logic.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+ESC_HELP   = '?'     # Find information about object
+ESC_HELP2  = '??'    # Find extra-detailed information about object
+ESC_MAGIC  = '%'     # Call magic function
+ESC_MAGIC2 = '%%'    # Call cell-magic function
+ESC_QUOTE  = ','     # Split args on whitespace, quote each as string and call
+ESC_QUOTE2 = ';'     # Quote all args as a single string, call
+ESC_PAREN  = '/'     # Call first argument with rest of line as arguments
+
+ESCAPE_SINGLES = {'!', '?', '%', ',', ';', '/'}
+ESCAPE_DOUBLES = {'!!', '??'}  # %% (cell magic) is handled separately
```
Carreau: ok

Carreau:
I've read part of this, I'll let it settle a bit in my head and come back to it. 
How easy would it be to compare current/new transformations; discarding the old one, but comparing and raising a warning if there is a difference ?

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+                              (\.[a-zA-Z_*][\w*]*)*   # .etc.etc
+                              )
+                              (\?\??)$                # ? or ??
+                              """,
+                              re.VERBOSE)
+
+class HelpEnd(TokenTransformBase):
+    # This needs to be higher priority (lower number) than EscapedCommand so
+    # that inspecting magics (%foo?) works.
+    priority = 5
```
Carreau: I'm starting to be worried of attaching priorities to class themselves.
Should we register them with a priority ? that seem more flexible. Not sure yet, just Gut Feeling.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+    try:
+        for token in tokenize.generate_tokens(iter(lines).__next__):
+            tokens_by_line[-1].append(token)
+            if (token.type == NEWLINE) \
+                    or ((token.type == NL) and (parenlev <= 0)):
+                tokens_by_line.append([])
+            elif token.string in {'(', '[', '{'}:
+                parenlev += 1
+            elif token.string in {')', ']', '}'}:
+                parenlev -= 1
```
Carreau: We likely want to make sure `parenlev` always `>0` (same in other similar method). Not sure if it is possible though.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+
+def show_linewise_tokens(s: str):
+    """For investigation"""
+    if not s.endswith('\n'):
+        s += '\n'
+    lines = s.splitlines(keepends=True)
+    for line in make_tokens_by_line(lines):
+        print("Line -------")
+        for tokinfo in line:
+            print(" ", tokinfo)
```
Carreau: I guess that's for debug, it's unused I think.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+            with warnings.catch_warnings():
+                warnings.simplefilter('error', SyntaxWarning)
+                res = compile_command(''.join(lines), symbol='exec')
+        except (SyntaxError, OverflowError, ValueError, TypeError,
+                MemoryError, SyntaxWarning):
+            return 'invalid', None
+        else:
+            if res is None:
+                return 'incomplete', find_last_indent(lines)
+        return 'complete', None
```
Carreau: Ok.

IPython/core/interactiveshell.py
```diff
@@ -334,15 +334,17 @@ def _exiter_default(self):
...
-    # running them. It allows cell magics to contain blank lines.
-    input_transformer_manager = Instance('IPython.core.inputsplitter.IPythonInputSplitter',
-                                         (), {'line_input_checker': False})
+    @property
+    def input_splitter(self):
+        """Make this available for compatibility
+
+        ipykernel currently uses shell.input_splitter.check_complete
+        """
+        return self.input_transformer_manager
```
Carreau: Emit Deprecation ?

IPython/utils/tokenutil.py
```diff
@@ -99,12 +99,12 @@ def token_at_cursor(cell, cursor_pos=0):
...
         
-        if tok.token == tokenize2.NAME and not iskeyword(tok.text):
-            if names and tokens and tokens[-1].token == tokenize2.OP and tokens[-1].text == '.':
+        if tok.token == tokenize.NAME and not iskeyword(tok.text):
+            if names and tokens and tokens[-1].token == tokenize.OP and tokens[-1].text == '.':
                 names[-1] = "%s.%s" % (names[-1], tok.text)
             else:
                 names.append(tok.text)
-        elif tok.token == tokenize2.OP:
+        elif tok.token == tokenize.OP:
```
Carreau: ok.

IPython/core/inputtransformer2.py
takluyver: Not worth adding extra internal complexity to this to save a bit of typing in the 2 places it's used, IMO.

IPython/core/inputtransformer2.py
takluyver: Yup, parentheses do allow unpacking, as well as things like `d[key] = foo`.

It shouldn't need to handle unmatched closing parentheses, because as far as I know something like `a) = ...` would always be a syntax error. The 'token line' passed in to this function is a logical Python line, including multiline statements, so something like this would be processed as one line:

```python
(
 a, b
) = ...
```

IPython/core/inputtransformer2.py
takluyver: Yeah, I'll see if I can make this any clearer.

IPython/core/inputtransformer2.py
takluyver: I don't think it would help. It could be more readable with a pattern matching DSL - this is finding something like `.* = % .*`. But then we'd have to define and implement a DSL for this, which is a load of extra complexity for a relatively few use cases.

> what if there is multiple magics assignments ?

Then it finds the first one! Because magics don't have to be valid Python syntax, we can't be sure that the tokenisation will make any sense for subsequent magics. So we find the first one, transform it to valid Python syntax, retokenise, find the new first one (i.e. the second in the original code), and repeat.

IPython/core/inputtransformer2.py
takluyver: These are the sites that recognise assignment from `a = %magic` and `b = !system`. Those are the only two we support, and given how complicated this is, I honestly wouldn't encourage anyone else to try adding more such transformers. It is a bit clumsy, but for n=2, I don't think it's worth trying to abstract it away.

IPython/core/inputtransformer2.py
takluyver: It is unused, but I'm inclined to leave it there for debugging. It's not causing any harm by being there.

IPython/core/inputtransformer2.py
takluyver: It shouldn't come up in valid Python code, and if the code is invalid, all bets are off. But I'll add a check anyway.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,556 @@
...
+            return False, lines
+
+        transformer = min(candidates, key=TokenTransformBase.sortby)
+        return True, transformer.transform(lines)
+
+    def do_token_transforms(self, lines):
+        while True:
+            changed, lines = self.do_one_token_transform(lines)
+            if not changed:
+                return lines
```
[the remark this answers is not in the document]
takluyver: Good point.

takluyver:
Deprecation warnings: I'd like to make at least one release with the new machinery in place before the old ones start warning. Otherwise it's a pain for packages that need to support the old and new versions for some period of time. Nbconvert is one such package, for instance.

Concerns about extensibility, e.g. priority attributes: I know, it's not great. To be honest, although I hope this is an improvement on the old API, it's still complicated enough that I would advise anyone to avoid extending the syntax unless there's really no other option. Maybe we should define a more restrictive API that's safer to build?

Would it be worth doing a survey of who currently extends input transformation? There are only a couple of use cases I'm aware of that really need *text* transformation as opposed to AST transformation:

- Converting float literals to decimals before they are parsed and represented as a binary float. This is transforming already valid Python syntax into something different, which is much easier than what we need to do in transforming code that's not already valid Python.
- There's a [very old extension](https://bitbucket.org/birkenfeld/ipython-physics) that allows for calculations with uncertainties and units: `m = 80 +/- 5 kg`. IMO, at least the units part would be more idiomatic as multiplication (`80 * kg`) than a special syntax. Though I've got no good answer for the uncertainties part.

On reflection, I'd like IPython to get out of being a framework for people to build custom Python-like languages. It's a pretty bad framework for that, and doing it well is a hard problem that we're not really trying to solve. There are much smarter people than me working on parser technology.

Carreau:
> Deprecation warnings: I'd like to make at least one release with the new machinery in place before the old ones start warning.

Aren't PendingDeprecationwWarning for that ? "We're going to deprecate that, but not yet, please have a look at new mechanism, and give us feedback". 

Are you panning thus to re-add some of what was removed ? 

Fair enough for the rest. At least we know these questions were discussed. 

We should try to send a request for feedback on the mailing list ?

takluyver:
If PendingDeprecationWarning shows by default, the same argument applies to it - it's annoying to have tools warning when they can't yet migrate. If they don't show by default... I don't think there's much point in putting them there.

I think I just don't have your faith that clever use of deprecation warnings will help anything. ;-)

takluyver:
I've come back to this and simplified the extension interface. Now there are just two points to plug in string transformers which work on a list of lines - plus AST transformers, which are unchanged.

The complex token-based transformation machinery, which turns `%magics` and `!system` commands into valid Python code, is now essentially a black box; there is no supported interface to extend or modify that. That gives us the chance to change it in the future without needing to worry about breaking people's code again.

IPython/core/inputtransformer.py
takluyver: My PR was accepted, so `generate_tokens` is documented for Python 3.8. https://docs.python.org/3.8/library/tokenize.html#tokenize.generate_tokens

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
+"""Input transformer machinery to support IPython special syntax.
+
+This includes the machinery to recognise and transform ``%magic`` commands,
+``!system`` commands, ``help?`` querying, prompt stripping, and so forth.
```
willingc: Perhaps add to comment:

Added: IPython 0.7. Replaces inputsplitter and inputtransformer which were deprecated in 0.7.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+from typing import List, Tuple
+import warnings
+
+_indent_re = re.compile(r'^[ \t]+')
+
+def leading_indent(lines):
+    """Remove leading indentation.
+    
+    If the first line starts with a spaces or tabs, the same whitespace will be
+    removed from each following line.
```
willingc: ...line in the block.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    return [l[n:] if l.startswith(space) else l
+            for l in lines]
+
+class PromptStripper:
+    """Remove matching input prompts from a block of input.
+    
+    Parameters
+    ----------
+    prompt_re : regular expression
+        A regular expression matching any input prompt (including continuation)
```
willingc:
```
continuation `...`)
```

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    initial_re : regular expression, optional
+        A regular expression matching only the initial prompt, but not continuation.
+        If no initial expression is given, prompt_re will be used everywhere.
+        Used mainly for plain Python prompts, where the continuation prompt
```
willingc: plain Python prompts (typically >>>), where the

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+        return lines
+    if re.match('%%\w+\?', lines[0]):
+        # This case will be handled by help_end
+        return lines
+    magic_name, _, first_line = lines[0][2:-1].partition(' ')
+    body = ''.join(lines[1:])
+    return ['get_ipython().run_cell_magic(%r, %r, %r)\n'
+            % (magic_name, first_line, body)]
+
+# -----
```
willingc: Stray line?

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+
+def _find_assign_op(token_line):
+    # Get the index of the first assignment in the line ('=' not inside brackets)
```
willingc: Perhaps change to docstring

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    # We don't try to support multiple special assignment (a = b = %foo)
```
willingc: Note: We don't...

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    """
+    end_line = start_line
+    while lines[end_line].endswith('\\\n'):
+        end_line += 1
+        if end_line >= len(lines):
+            break
+    return end_line
+
+def assemble_continued_line(lines, start: Tuple[int, int], end_line: int):
+    """Assemble pieces of a continued line into a single line.
```
willingc:
```
"""Assemble a single line from multiple continued line pieces (lines ending in `\` and the line following the last `\` in the block).

For example, this code continues over multiple lines:

if (assign_ix is not None) \
                     and (len(line) >= assign_ix + 2) \
                     and (line[assign_ix+1].string == '%') \
                     and (line[assign_ix+2].type == tokenize.NAME):

This statement contains four continued line pieces. Assembling these pieces into a single line would be:

if (assign_ix is not None) and (len(line) >= assign_ix + 2) and (line[assign_ix+1].string == '%') and (line[assign_ix+2].type == tokenize.NAME):
```

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+
+    Uses 0-indexed line numbers. *start* is (lineno, colno).
+    """
+    parts = [lines[start[0]][start[1]:]] + lines[start[0]+1:end_line+1]
+    return ' '.join([p[:-2] for p in parts[:-1]]  # Strip backslash+newline
+                    + [parts[-1][:-1]])         # Strip newline from last line
+
+class TokenTransformBase:
+    # Lower numbers -> higher priority (for matches in the same location)
```
willingc: docstring for when to use TokenTransformBase

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+
+    def __init__(self, start):
+        self.start_line = start[0] - 1   # Shift from 1-index to 0-index
+        self.start_col = start[1]
+
+    def transform(self, lines: List[str]):
+        raise NotImplementedError
+
+class MagicAssign(TokenTransformBase):
+    @classmethod
```
willingc: class docstring perhaps

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+        for line in tokens_by_line:
+            assign_ix = _find_assign_op(line)
+            if (assign_ix is not None) \
+                    and (len(line) >= assign_ix + 2) \
+                    and (line[assign_ix+1].string == '%') \
+                    and (line[assign_ix+2].type == tokenize.NAME):
+                return cls(line[assign_ix+1].start)
+    
+    def transform(self, lines: List[str]):
+        """Transform a magic assignment found by find
```
willingc: Transform a magic assignment which has been found by the class method `find()` in a cell block's lines.

Not sure that the above is technically correct but "found by find" is a bit 😕

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+
+                while ix < len(line) and line[ix].type == tokenize.ERRORTOKEN:
+                    if line[ix].string == '!':
+                        return cls(line[ix].start)
+                    elif not line[ix].string.isspace():
+                        break
+                    ix += 1
+
+    def transform(self, lines: List[str]):
+        """Transform a system assignment found by find
```
willingc: Similar to comment above re: "found by find"

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    t_magic_name = t_magic_name.lstrip(ESC_MAGIC)
+    if next_input is None:
+        return 'get_ipython().run_line_magic(%r, %r)' % (t_magic_name, t_magic_arg_s)
+    else:
+        return 'get_ipython().set_next_input(%r);get_ipython().run_line_magic(%r, %r)' % \
+           (next_input, t_magic_name, t_magic_arg_s)
+
+def _tr_help(content):
+    "Translate lines escaped with: ?"
+    # A naked help line should just fire the intro help screen
```
willingc: I would just add the comment into the docstring.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    if not content:
+        return 'get_ipython().show_usage()'
+
+    return _make_help_call(content, '?')
+
+def _tr_help2(content):
+    "Translate lines escaped with: ??"
```
willingc: Same comment as prior function.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    name, _, args = content.partition(' ')
+    return 'get_ipython().run_line_magic(%r, %r)' % (name, args)
+
+def _tr_quote(content):
+    "Translate lines escaped with: ,"
+    name, _, args = content.partition(' ')
+    return '%s("%s")' % (name, '", "'.join(args.split()) )
+
+def _tr_quote2(content):
+    "Translate lines escaped with: ;"
```
willingc: with a semicolon: ;

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    name, _, args = content.partition(' ')
+    return '%s("%s")' % (name, args)
+
+def _tr_paren(content):
+    "Translate lines escaped with: /"
```
willingc: with a forward slash:

Since the function names don't match the character it's better to be explicit IMHO.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+        Returns (line, column) of the escape if found, or None. *line* is 1-indexed.
+        """
+        for line in tokens_by_line:
+            ix = 0
+            while line[ix].type in {tokenize.INDENT, tokenize.DEDENT}:
+                ix += 1
+            if line[ix].string in ESCAPE_SINGLES:
+                return cls(line[ix].start)
+
+    def transform(self, lines):
```
willingc: Add a brief docstring

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+_help_end_re = re.compile(r"""(%{0,2}
+                              [a-zA-Z_*][\w*]*        # Variable name
+                              (\.[a-zA-Z_*][\w*]*)*   # .etc.etc
+                              )
+                              (\?\??)$                # ? or ??
+                              """,
+                              re.VERBOSE)
+
+class HelpEnd(TokenTransformBase):
+    # This needs to be higher priority (lower number) than EscapedCommand so
```
willingc: Let's make a class docstring here.

Relative priority of the different classes could be added to the module docstring or mentioned there.

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    # that inspecting magics (%foo?) works.
+    priority = 5
+
+    def __init__(self, start, q_locn):
+        super().__init__(start)
+        self.q_line = q_locn[0] - 1  # Shift from 1-indexed to 0-indexed
+        self.q_col = q_locn[1]
+
+    @classmethod
+    def find(cls, tokens_by_line):
```
willingc: docstring

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+        for line in tokens_by_line:
+            # Last token is NEWLINE; look at last but one
+            if len(line) > 2 and line[-2].string == '?':
+                # Find the first token that's not INDENT/DEDENT
+                ix = 0
+                while line[ix].type in {tokenize.INDENT, tokenize.DEDENT}:
+                    ix += 1
+                return cls(line[ix].start, line[-2].start)
+
+    def transform(self, lines):
```
willingc: docstring

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+                if parenlev > 0:
+                    parenlev -= 1
+    except tokenize.TokenError:
+        # Input ended in a multiline string or expression. That's OK for us.
+        pass
+    
+    return tokens_by_line
+
+def show_linewise_tokens(s: str):
+    """For investigation"""
```
willingc: For investigation and debugging

IPython/core/inputtransformer2.py
```diff
@@ -0,0 +1,564 @@
...
+    lines = s.splitlines(keepends=True)
+    for line in make_tokens_by_line(lines):
+        print("Line -------")
+        for tokinfo in line:
+            print(" ", tokinfo)
+
+# Arbitrary limit to prevent getting stuck in infinite loops
+TRANSFORM_LOOP_LIMIT = 500
+
+class TransformerManager:
```
willingc: docstring """Manages the different transforms applied to a cell or code block""" ?

IPython/core/interactiveshell.py
```diff
@@ -334,15 +334,26 @@ def _exiter_default(self):
...
+        return self.input_transformer_manager.cleanup_transforms
+
+    input_transformers_post = List([],
+        help="A list of string input transformers, to be applied after IPython's "
+             "own input transformations."
+    )
+
+    @property
+    def input_splitter(self):
+        """Make this available for compatibility
```
willingc: Make this available for backward compatibility (pre-0.7 release) with existing code.

For example, ipykernel...

New code should use input_transformer_manager.

IPython/core/interactiveshell.py
```diff
@@ -2657,6 +2668,7 @@ def run_cell(self, raw_cell, store_history=False, silent=False, shell_futures=Tr
         -------
         result : :class:`ExecutionResult`
         """
+        result = None
```
willingc: :smile: Good catch

IPython/core/interactiveshell.py
```diff
@@ -2799,6 +2800,24 @@ def error_before_exec(value):
             self.execution_count += 1
 
         return result
+
+    def transform_cell(self, raw_cell):
```
willingc: Add docstring explaining static input transformation vs dynamic transformations

IPython/core/tests/test_inputtransformer2.py
```diff
@@ -0,0 +1,189 @@
+import nose.tools as nt
```
willingc: Add a docstring similar to this:

```
"""Tests for the token-based transformers in IPython.core.inputtransformer2

Line-based transformers are the simpler ones; token-based transformers are
more complex.
"""
```

IPython/terminal/shortcuts.py
```diff
@@ -68,9 +68,8 @@ def create_ipython_shortcuts(shell):
                           & insert_mode
                           & cursor_in_leading_ws
                         ))(indent_buffer)
-
     kb.add('c-o', filter=(has_focus(DEFAULT_BUFFER)
-                  & emacs_insert_mode))(newline_autoindent_outer(shell.input_splitter))
+                  & emacs_insert_mode))(newline_autoindent_outer(shell.input_transformer_manager))
```
willingc: A comment on style here. Perhaps going to a longer line length would improve readability and understanding. i.e. line 74 is more understandable than 71-72 and 66-70

willingc (commented):
@takluyver This looks really good. The code looks clean. Most of my suggestions are related to docstrings and improving understanding. After merging we should probably go back and do another pass with docs. Thank you 👍

IPython/core/interactiveshell.py
Carreau: @willingc Your  `7` and `0` keys are either swapped, or you're used to software below 1.0 :-) I've edited a couple of your comments to say `7.0` instead of `0.7` as I think that's what you meant.

Carreau:
I'll re-read all of that later, flying tomorrow. If I have wifi in the plane, I'll do that then.

IPython/core/interactiveshell.py
willingc: Yes, I had a backwards moment :wink:

Carreau:
I push a docstring and a couple of deprecation warnings. 
Feel free to remove if you like.

takluyver:
Thanks @willingc for the detailed review! I think I've basically applied all of your suggestions.

Carreau:
cc @dakoop, @colinjbrown from https://github.com/dataflownb/dfkernel  I believe you are using input transformer. Just to make sure this will not affect you too much.

IPython/core/inputsplitter.py
Carreau: Done.

IPython/core/inputtransformer2.py
Carreau: Done.

dakoop:
@Carreau, thanks for the heads-up. Shouldn't affect us much, and it looks like a nice improvement.

willingc:
@takluyver Traveling today but trust that you covered all of my suggestions. I'll give a look tomorrow.

Carreau:
I've fixed the conflicts.

Carreau:
> I've fixed the conflicts.

I've fixed new conflict due to the reactivation of autoindent.

takluyver:
Thanks for fixing up the conflicts; I would be happy to see this land (though I'm not looking forward to finding out what the change has broken).

Carreau:
> I would be happy to see this land

So be it, 

> though I'm not looking forward to finding out what the change has broken

I'll help you fix stuff, worse case we revert and re-iterate.

There is only a couple of things left for 7.0 (and one test regression I'd like to backport https://github.com/ipython/ipython/pull/11277 for 5.9) and we can do a beta.

Carreau:
🎉 🍸 :pinot grigio:

Carreau:
I'm sure even Mr Meeseeks is excited...

takluyver:
:-P

Does saying the bots name and the word 'excited' in the same comment trigger that?

Carreau:
No, but the bot can read inside HTML comments, that aren't shown in the UI!

Issue #11042, opened 2018-03-11
Make the crash handler tell users to provide more information


slel:
Currently the crash handler provides a crash log and tells users to send it
to developers. It would be useful to also tell users to provide information
about their operating system and how they installed the application.

jdemeyer:
+1. Somehow the existence of the crash log makes users think that they do not have to provide any extra information. We often get bug reports in Sage containing literally only the crash log.

slel:
Pull request #11043 aims to solve this issue by suggesting users
to include some more information when sending a crash report.

takluyver:
Closed by #11043.

PR #11043, opened 2018-03-11
Update crashhandler.py to solve #11042

slel:
Tell users to provide information about their operating system
and how they installed the application.

### DIFF
--- a/IPython/core/crashhandler.py
+++ b/IPython/core/crashhandler.py
@@ -45,16 +45,23 @@ A crash report was automatically generated with the following information:
 It was left in the file named:
 \t'{crash_report_fname}'
 If you can email this file to the developers, the information in it will help
-them in understanding and correcting the problem. Please include information
-about your operating system: is it Linux, macOS, Windows, other, and which
-version, for example Ubuntu 16.04, macOS 10.13.2, Windows 10 Pro 64-bit...
-Please also include information on how you installed {app_name}.
+them in understanding and correcting the problem.
 
 You can mail it to: {contact_name} at {contact_email}
 with the subject '{app_name} Crash Report'.
 
 If you want to do it now, the following command will work (under Unix):
 mail -s '{app_name} Crash Report' {contact_email} < {crash_report_fname}
+
+Please also include in your email:
+- information about your operating system: is it Linux, macOS, Windows, other,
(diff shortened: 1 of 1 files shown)


jdemeyer:
Info about operating system is already in the crash log. What's mostly missing is *context*: what did users do to end up with that crash log?

IPython/core/crashhandler.py
```diff
@@ -53,6 +53,16 @@
 If you want to do it now, the following command will work (under Unix):
 mail -s '{app_name} Crash Report' {contact_email} < {crash_report_fname}
 
+Please also include in your email:
+- information about your operating system: is it Linux, macOS, Windows, other,
+  and which version, for example Ubuntu 16.04, macOS 10.13.2, Windows 10 Pro,
+  is it 32-bit or 64-bit, ...
+- information on how you installed {app_name}: from binary, compiled from source,
```
takluyver: 'compiled from source' doesn't really apply to pure Python projects like IPython. I'd give examples like "using pip or conda, from GitHub, or as part of a Docker container."

IPython/core/crashhandler.py
```diff
@@ -53,6 +53,16 @@
...
+  via Docker, on a virtual machine...
+- information on how to reproduce the crash: what exact sequence of instructions
```
takluyver: Since all three points start with "information on", let's phrase the bit at the top "Please also include information on:" and cut those words from the bullet points.

Also, I'd like the bullet points to start with capital letters  ("How to reproduce...").

IPython/core/crashhandler.py
```diff
@@ -53,6 +53,16 @@
 If you want to do it now, the following command will work (under Unix):
 mail -s '{app_name} Crash Report' {contact_email} < {crash_report_fname}
 
+Please also include in your email:
+- information about your operating system: is it Linux, macOS, Windows, other,
+  and which version, for example Ubuntu 16.04, macOS 10.13.2, Windows 10 Pro,
+  is it 32-bit or 64-bit, ...
```
takluyver: Ending a paragraph with `...` to signify that these are examples seems a bit too casual for the tone of this message. Let's find a way to make it read a bit better.

takluyver:
Thanks!

Issue #11059, opened 2018-03-26
FileLink should accept PathLike objects


ciupicri:
Starting with [Python 3.6][1], the [`os.PathLike`][1] interface was introduced so that functions which operate on paths would accept not only `str` or `bytes`, but also [`pathlib.PurePath`][3] for example. 

`IPython.display.FileLink` does not accept `PathLike` objects yet:

    [In]: IPython.display.FileLink(pathlib.Path(r'C:\Windows\System32\drivers\etc\hosts'))
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    C:\mnt\Anaconda\5.1.0\lib\site-packages\IPython\core\formatters.py in __call__(self, obj)
        343             method = get_real_method(obj, self.print_method)
        344             if method is not None:
    --> 345                 return method()
        346             return None
        347         else:
    
    C:\mnt\Anaconda\5.1.0\lib\site-packages\IPython\lib\display.py in _repr_html_(self)
        355                     "incorrect path." % self.path)
        356 
    --> 357         return self._format_path()
        358 
        359     def __repr__(self):
    
    C:\mnt\Anaconda\5.1.0\lib\site-packages\IPython\lib\display.py in _format_path(self)
        341 
        342     def _format_path(self):
    --> 343         fp = ''.join([self.url_prefix,self.path])
        344         return ''.join([self.result_html_prefix,
        345                         self.html_link_str % (fp, self.path),
    
    TypeError: sequence item 1: expected str instance, WindowsPath found

This could be implemented by calling [`os.fspath`][4] on the parameter, but care should be taken because the function did not exist before Python 3.6.

Also for what it's worth `pathlib` objects have a method named [`as_uri()`][5].

 [1]: https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep519
 [2]: https://docs.python.org/3/library/os.html#os.PathLike
 [3]: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath
 [4]: https://docs.python.org/3/library/os.html#os.fspath
 [5]: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_uri

ciupicri:
This could be implemented by calling [`os.fspath`][1] on the parameter, but care should be taken because the function does not exist before Python 3.6.

[1]: https://docs.python.org/3/library/os.html#os.fspath

takluyver:
Sounds good. Maybe we can handle the compatibility something like this:

[...]

PR #11060, opened 2018-03-27
FileLink: accept path-like objects (#11059)

ciupicri:
Accept path-like objects on Python >= 3.6.


takluyver:
Thanks!

takluyver:
A lot of attributes in our code are like that - not explicitly private, but not documented either. So it's not really defined whether users can rely on them or not. In practice, we'd try not to break something like changing `filelink.path` unless there was a good reason to.

PR #11061, opened 2018-03-28
FileLink: escape HTML unsafe characters from path

ciupicri:


IPython/lib/display.py
```diff
@@ -342,7 +343,7 @@ def __init__(self,
     def _format_path(self):
         fp = ''.join([self.url_prefix,self.path])
```
takluyver: It probably needs to be escaped on this line too - this is what goes into the `href`.

takluyver:
Thanks, looks good.

PR #11062, opened 2018-03-29
Misc. typos

luzpaz:
Found via `codespell -q 3`


IPython/core/completer.py
```diff
@@ -949,7 +949,7 @@ def _formatparamchildren(parameter) -> str:
 
     Jedi does not expose a simple way to get `param=value` from its API.
 
-    Prameter
+    Parameter
     ========
```
takluyver: Can you extend the underline here? I think it needs to be the same length as the heading.

takluyver (commented):
Thanks!

luzpaz:
Revised and rebased

takluyver:
Thanks!

PR #11063, opened 2018-03-31
Get jedi completions on only text before the cursor

### LINKED ISSUE #10781: Invalid path completion when not at end of string
As of 6.1 and master (with an empty profile)
```
with IPython.core.completer.provisionalcompleter():
    print(list(ip.Completer.completions("Path('/tmp/", 11)))
```
i.e. complete `Path('/tmp/_` where _ indicates the cursor position, returns a correct list of subpaths from /tmp, but
```
with IPython.core.completer.provisionalcompleter():
    print(list(ip.Completer.completions("Path('/tmp/')", 11)))
```
i.e. complete `Path('/tmp/_')` (which is common in a style of typing where you immediately close your paired delimiters/brackets) returns a bunch of useless completions (basically the entire global namespace, and also the correct completions too).

The behavior is also seen by manually triggering the completion, of course.

takluyver:
Closes gh-10781

This might make completions less smart in some situations, if Jedi can do intelligent things with the text after the cursor, but it fixes an important case that has been causing annoyances.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1340,7 +1340,6 @@ class IPCompleter(Completer):
             namespaces.append(self.global_namespace)
 
         completion_filter = lambda x:x
-        # cursor_pos is an it, jedi wants line and column
         offset = cursor_to_position(text, cursor_line, cursor_column)
         # filter output if we are completing for object members
         if offset:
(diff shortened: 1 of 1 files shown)


Carreau:
+1 as a workaround, but we should have a proper fix at some point.

takluyver:
I agree. I put some ideas about how to do this check more robustly in #11064.

takluyver:
Thanks! This was the last thing open for 6.3. Is there anything else you can see that should happen before a release? If not, I'll try to get 6.3 and 5.5 out tomorrow.

Issue #11029, opened 2018-03-01
position_to_cursor won't accept offset of zero


zooba:
[completer.py](https://github.com/ipython/ipython/blob/48eab58c7a96a66b67b15b926a6143db86d16a69/IPython/core/completer.py#L872) has this assertion in its `position_to_cursor` function, which raises if you pass `0`.

```
    assert 0 < offset <= len(text) , "0 < %s <= %s" % (offset , len(text))
```

I don't see any reason why zero should be an invalid offset here, and it interferes with requesting all completions at the start of a cell (e.g. `client.complete("")` via Jupyter client).

takluyver:
Makes sense. #11065 should fix it to allow 0.

PR #11065, opened 2018-04-02
Allow offset 0 in position_to_cursor

takluyver:
Closes gh-11029


takluyver:
Merging this because I'm trying to make a release, and it seems harmless (famous last words)

PR #11067, opened 2018-04-02
remove some py3compat usage

Carreau:
kept some doctest_refactor_print in testing/ipydoctest.

### DIFF
--- a/IPython/utils/io.py
+++ b/IPython/utils/io.py
@@ -17,7 +17,6 @@ from warnings import warn
 
 from IPython.utils.decorators import undoc
 from .capture import CapturedIO, capture_output
-from .py3compat import input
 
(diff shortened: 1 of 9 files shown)


IPython/utils/io.py
```diff
@@ -172,7 +172,7 @@ def ask_yes_no(prompt, default=None, interrupt=None):
     ans = None
     while ans not in answers.keys():
         try:
-            ans = input(prompt+' ').lower()
+            ans = builtins.input(prompt+' ').lower()
```
takluyver: I don't think we actually need `builtins` here - we can just use `input()` directly, I think.

Carreau: Yeah, good point. Fixed.

PR #11072, opened 2018-04-03
Remove deprecated profile line magic.

Carreau:
This was deprecated since 2.0, was made no-op in 6.0 + print warning.

I suggest removing that in 7.0, and leave open the possibility to have
%profile as an alias or alternative for %prun in 8.0+


takluyver:
:+1:

PR #11073, opened 2018-04-03
Deprecate some methods from IPython.utils.io

Carreau:
stdin/err/out defined in this module ado not seem to be used and are
based on IOStream which is deprecated since 5.0. I'm considering that
for removal as well.


minrk:
As with everything, I think we shouldn't *remove* anything deprecated until its presence causes some kind of problem. But deprecating these makes sense to me.

Issue #11070, opened 2018-04-03
Tab completion stopped working in IPython 6.3.0


rayosborn:
I have a class with the ```__dir__``` function defined to enable autocompletion. It works well in IPython 6.2.1, but has stopped working in IPython 6.3.0. I tested it in a newly updated Jupyter QtConsole in Python 3.6.
```
Jupyter QtConsole 4.3.1
Python 3.6.3 | packaged by conda-forge | (default, Nov  4 2017, 10:13:32) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.3.0 -- An enhanced Interactive Python. Type '?' for help.
```
Do I have to update any other modules that are not listed as a dependency?

Edit: This is for tab-completion of class attributes. Tab-completion seems to work for dictionaries.

takluyver:
It's an issue in Jedi, the library we're moving towards for using completions. It's been fixed, but it needs a new release:

https://github.com/davidhalter/jedi/issues/1027

In the meantime, you can configure `Completer.use_jedi = False` to switch back to completions found by IPython's own machinery.

rayosborn:
Thanks for the quick response. Should I leave this issue open for now to alert others? You're welcome to close it if you prefer.

takluyver:
May as well leave it open for now, because people will probably open more if they don't see it.

takluyver:
I've released IPython 6.3.1 which switches the default back to our own completions, so this issue should go away.

PR #11076, opened 2018-04-05
Disable Jedi completions by default

takluyver:
The release of IPython 6.3.0 changed to relying on jedi completions, whereas we had previously combined Jedi completions with our own (which caused issues with duplicate completions in some circumstances).

Unfortunately this exposed a bug in Jedi where it doesn't find dynamic attributes from `__dir__()`. This means that, among other things, completing `df.Column` on pandas dataframes is broken (e.g. https://github.com/jupyter/notebook/issues/2435#issuecomment-378940525 , #11070). It has been fixed in Jedi, but a new release is apparently still a few weeks away.

I'm therefore proposing that we release IPython 6.3.1 quickly with this change, disabling Jedi by default and switching back to our ugly but battle-tested completion machinery. That fixes the immediate problems for users, and we can reevaluate the options for future releases.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -578,7 +578,7 @@ class Completer(Configurable):
 
     use_jedi = Bool(default_value=False,
                     help="Experimental: Use Jedi to generate autocompletions. "
-                    "Default to True if jedi is installed").tag(config=True)
+                    "Off by default.").tag(config=True)
 
     jedi_compute_type_timeout = Int(default_value=400,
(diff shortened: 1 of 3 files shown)


takluyver:
I agreed this course yesterday with @Carreau , who has been driving the completions work. I've added the 6.3.1 changelog (single entry) to this PR for simplicity.

takluyver:
Reopening to rerun tests on Appveyor, because I don't have a button to do that directly.

Has anyone ever worked out Appveyor's permissions? :-/

Issue #11068, opened 2018-04-03
Cleanup extra logic that handle python <=3.3


Carreau:
https://github.com/ipython/ipython/pull/10833 dis most of that but there seem to be some logic still handling python 3.3

https://github.com/ipython/ipython/blob/dda523895e5a9c4185de763a25f038dc6fff76b8/IPython/core/extensions.py#L17-L21

https://github.com/ipython/ipython/blob/dda523895e5a9c4185de763a25f038dc6fff76b8/IPython/utils/openpy.py#L117-L121

https://github.com/ipython/ipython/blob/dda523895e5a9c4185de763a25f038dc6fff76b8/IPython/utils/text.py#L512-L519

(3.4.2+)

https://github.com/ipython/ipython/blob/dda523895e5a9c4185de763a25f038dc6fff76b8/IPython/core/magics/execution.py#L145-L149

There are possibly other places, this should be relatively easy to fix for a new contributor.

Each of these removal can be a separate PR.

subhrm:
I would try to clean them up with separate PRs for each fix.

subhrm:
I have now created two pull requests for fixes in IPython/core/extensions.py and IPython/utils/openpy.py. Based on the feedback from the maintainers, I would proceed to clean up rest of the files.

PR #11093, opened 2018-04-13
Issue #11068 : Cleanup extra logic that handle python <=3.3 in IPython/utils/openpy.py

subhrm:
Removed deprecated imports from IPython/utils/openpy.py .

Please review and merge.


takluyver:
Thanks!

PR #11101, opened 2018-04-17
Added fix for display hook call output format

MSeal:
Hit an issue were under complex conditions, that are hard to express in a simple notebook, the displayhook was building output array elements in the old tuple format instead of the kwarg format. This causes (ignore py2 package path in the stack trace example):

```
TypeErrorTraceback (most recent call last)
<ipython-input-7-dcf49c89b1af> in <module>()
----> 1 output.outputs[0]

/usr/local/lib/python2.7/dist-packages/IPython/utils/capture.pyc in outputs(self)
    112                 display(o)
    113         """
--> 114         return [ RichOutput(**kargs) for kargs in self._outputs ]
```

Tracing down the fix was to adjust the callable which was not updated. Tracing the capture code throughout the project was a real headache as the global state is touched a lot along the way, so I made a spec just to prove that the output format conforms rather than express where that hook callable is made consequentially.

Py2 backport PR: https://github.com/ipython/ipython/pull/11102

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -317,4 +317,4 @@ class CapturingDisplayHook(object):
         if result is None:
             return
         format_dict, md_dict = self.shell.display_formatter.format(result)
-        self.outputs.append((format_dict, md_dict))
+        self.outputs.append({ 'data': format_dict, 'metadata': md_dict })
--- a/IPython/core/tests/test_displayhook.py
+++ b/IPython/core/tests/test_displayhook.py
@@ -1,4 +1,7 @@
+import sys
 from IPython.testing.tools import AssertPrints, AssertNotPrints
+from IPython.core.displayhook import CapturingDisplayHook
+from IPython.utils.capture import CapturedIO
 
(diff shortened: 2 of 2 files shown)


Carreau:
> Failure seems unrelated to PR? FAIL: IPython.core.tests.test_completer.test_omit__names?

That might be due to a recent release of Jedi.  I'll investigate.

MSeal:
@Carreau Any update on this?

takluyver:
I think this makes sense. The corresponding change to displaypub was in #10837, which shipped in 6.3. Reopening to rerun tests.

takluyver:
...which probably won't help, because they're failing on master as well. @Carreau , did you get anywhere with the `test_omit__names` failure?

Carreau:
Yes we've commented the tests.

I'm flabbergasted our bot did not react to the PR merge. I guess we also need to backport that on 6.x. and do a 5.x, and 6.x release.

Carreau:
Ah, no my bad, I merged the backport, and not on master that's why.

PR #11102, opened 2018-04-17
Added fix for display hook call output format

MSeal:
Hit an issue were under complex conditions, that are hard to express in a simple notebook, the displayhook was building output array elements in the old tuple format instead of the kwarg format. This causes:

```python-traceback
TypeErrorTraceback (most recent call last)
<ipython-input-7-dcf49c89b1af> in <module>()
----> 1 output.outputs[0]

/usr/local/lib/python2.7/dist-packages/IPython/utils/capture.pyc in outputs(self)
    112                 display(o)
    113         """
--> 114         return [ RichOutput(**kargs) for kargs in self._outputs ]
```

Tracing down the fix was to adjust the callable which was not updated. Tracing the capture code throughout the project was a real headache as the global state is touched a lot along the way, so I made a spec just to prove that the output format conforms rather than express where that hook callable is made consequentially.

Master variant of the same PR: https://github.com/ipython/ipython/pull/11101

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -306,4 +306,4 @@ class CapturingDisplayHook(object):
         if result is None:
             return
         format_dict, md_dict = self.shell.display_formatter.format(result)
-        self.outputs.append((format_dict, md_dict))
+        self.outputs.append({ 'data': format_dict, 'metadata': md_dict })
--- a/IPython/core/tests/test_displayhook.py
+++ b/IPython/core/tests/test_displayhook.py
@@ -1,4 +1,7 @@
+import sys
 from IPython.testing.tools import AssertPrints, AssertNotPrints
+from IPython.core.displayhook import CapturingDisplayHook
+from IPython.utils.capture import CapturedIO
 
 ip = get_ipython()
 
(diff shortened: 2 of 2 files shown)


rgbkrk:
This ended up affecting both nbconvert and papermill with Python 2.x kernels.

Carreau (approved):
+1, we can try to slowly deprecate (warn ?) on the old format. But this can be done separately. 

Does that need backport ?

Carreau:
Or is that just on 5.x ? that is to say no forward port to master ?

MSeal:
The old format was removed from 5.x with the 5.6 release -- this fixes a bug with that change. I also added the same bug fix against master in the linked PR

Carreau:
> Master variant of the same PR: #11101
> I also added the same bug fix against master in the linked PR

Ah sorry I missed that. Was probably tired. You can skip making PR on old branches (if applying the patch is trivial) we have a bot that take care of it.

rgbkrk:
This will target a `5.6.x` milestone right?

Carreau:
> This will target a 5.6.x milestone right?

Sorry, habbit of tagging for  Major-1. Changed to 5.7 (which should backport to 5.x), then we can see for a 5.6.x or 5.7.

PR #11110, opened 2018-05-01
Comment out jedi parts of test_omit__names

takluyver:
We're not using Jedi by default for now, and the test failures are holding up unrelated PRs.


Carreau:
@meeseeksdev backport to 6.x

PR #11128, opened 2018-05-10
Traceback printing change `in <module>()` to `in <module>`.

Carreau:
Modules are usually not callable


Carreau:
@willingc on the education side, do you have thoughts ?

willingc:
@Carreau Let me make sure that I understand the change. Are there changes beyond the traceback printing out  `<module>` instead of `<module>()`?

Overall, I think it is cleaner without the parens.

willingc (approved):

Carreau:
No, in this pull-request no change beyond that.
For me the `()` mean you call something, and you (usually) can't call a module.

I'm working on refactoring the tracebacks code – without visual change – but once I'm ready to do visual changes I'll do an extensive proposal.

PR #11135, opened 2018-05-12
Fix test for Python 3.7+

Carreau:
A bare string is interpreted as a module docstring and does not trigger
the usual side-effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: python
 python:
     - "nightly"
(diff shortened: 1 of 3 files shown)


Carreau:
Also start testing on 3.7.

Carreau:
actually I think I should port that to 6.x as 7.0 will not be release for Python 3.7

@meeseeksdev backport to 6.x

PR #11139, opened 2018-05-16
signature: use inspect instead of type checks

jdemeyer:
Using `inspect.isfunction` is more future-compatible in case that the implementation of `inspect.isfunction` ever changes.

In SageMath, we actually monkey-patch `inspect.isfunction` to allow for Cython functions. So while this patch doesn't change anything at all in vanilla Python, it is important for Cython support.

More context:
- https://opendreamkit.org/2017/06/09/CythonSphinx
- https://bugs.python.org/issue30071
- https://www.python.org/dev/peps/pep-0579/

### DIFF
--- a/IPython/utils/_signatures.py
+++ b/IPython/utils/_signatures.py
@@ -21,6 +21,7 @@ import itertools
 import functools
 import re
 import types
+import inspect
 
 
 # patch for single-file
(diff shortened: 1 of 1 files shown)


Carreau:
+1.

Do you think it is worth backporting to 5.x (Python 2.7), or can this be a carrot for 3.x users ?

jdemeyer:
It's a pull request to the `5.x` branch. It *only* makes sense on Python 2.

Carreau:
Oh, sorry, I missread. Thanks.

Carreau:
@jdemeyer released on PyPI, I'll update the conda recipe later.

PR #11142, opened 2018-05-17
Some cleanup of Pycolorize.

Carreau:
Style, remove of unused variable... etc. In planning of refactor.


takluyver:
This all looks fine, though I don't know why we'd want to plan a refactor; AFAIK `Pycolorize` has been deprecated for some time in favour of pygments.

takluyver:
I think some refactoring of ultratb makes sense. If that needs refactoring of Pycolorize at the same time, I think it might be better just to switch ultratb away from Pycolorize at the same time. But you're the one looking at it, so I'll trust your judgment.

PR #11158, opened 2018-05-25
simplify Python 3.7 bare string workaround

minrk:
recompile with 'single' instead of reaching in and modifying the AST

possibly simpler alternative to #11135 for #11133


Carreau:
The problem with this is if we have any further downstream AST transformers looking at the first node then this won't work as the first node will be gone in multiline statements.

Carreau:
Closing as CPython removed the workaround.

Issue #8608, opened 2015-07-05
Volume normalization in IPython.display.Audio should be optional


mmcdan:
I am manipulating audio using numpy with IPython Notebook and want to use IPython.display.Audio for listening to the numpy arrays. Unfortunately, auto-normalization tampers with the results. Example:

```
# Generate a sound
import IPython
import numpy as np
framerate = 44100
t = np.linspace(0,5,framerate*5)
tone = np.sin(2*np.pi*220*t)
antitone = np.sin(2*np.pi*220*t + np.pi)
IPython.display.Audio(tone+antitone, rate=framerate)
```

Adding a sin wav to itself shifted by 180 deg should give total silence. Instead, auto-normalization amplifies the floating point errors. The problem is in the IPython.lib.display.Audio _make_wav method, which always normalizes the numpy array(see 'scaled' variable). I think that we should have a normalize keyword argument, so that Audio can be used for audio analysis. Something like:

```
Audio(tone+antitone, rate=framerate, normalize=False)
```

takluyver:
That seems reasonable. Do you want to make a PR?

mmcdan:
Sure, will make a PR in the near future.

dsblank:
I am also interested in making a couple of enhancements to Audio:

1. Consider passing in a `maxvalue` value rather than a `normalize` flag. That way, you can still normalize, but to a different value. In the antitone example above, maxvalue would be 1.
2. Adding a `save` method to Audio.

I have code that does the first.

The second can accomplished with:

```python
def save_wav(data, rate, filename):
    audio = Audio(data, rate=rate)
    with open(filename, "wb") as fp:
        fp.write(audio.data)
```
but would be handy to have as a defined method on Audio:

```python
    def save(self, filename):
        with open(filename, "wb") as fp:
            fp.write(self.data)
```

drscotthawley:
@mmcdan @takluyver : Any progress on this?   
The Audio object is distorting my output, even adding a DC offset.  
Have been searching for a while for a bug in my code, before finding this issue. 

Plotting the waveform and listening to it via Audio is totally different -- 
You can directly compare the difference between downloading from Audio, vs. saving via librosa.
How can we get Audio to just leave the signal unaltered?

drscotthawley:
Woops, missed a spot. New PR.

musicnerd:
[...]

PR #11161, opened 2018-05-25
added 'norm' kwarg to make Audio scaling normalization optional

drscotthawley:
Following up on Issue https://github.com/ipython/ipython/issues/8608

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -139,7 +139,7 @@ class Audio(DisplayObject):
             if norm:
                 scaled = np.int16(data/np.max(np.abs(data))*32767).tolist()
             else:
-                scaled = np.int16(data*32767).tolist()
+                scaled = np.int16(np.clip(data,-1,1)*32767).tolist()
 
         except ImportError:
             # check that it is a "1D" list
@@ -155,7 +155,7 @@ class Audio(DisplayObject):
                 maxabsvalue = float(max([abs(x) for x in data]))
                 scaled = [int(x/maxabsvalue*32767) for x in data]
             else:
-                scaled = [int(x*32767) for x in data]
+                scaled = [int(np.clip(x,-1,1)*32767) for x in data]
 
             nchan = 1
 
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -54,6 +54,9 @@ class Audio(DisplayObject):
     autoplay : bool
         Set to True if the audio should immediately start playing.
         Default is `False`.
+    norm : bool
+        Normalize (rescale) audio to maximum possible range.
+        Default is `True`.
 
     Examples
     --------
@@ -83,7 +86,7 @@ class Audio(DisplayObject):
     """
     _read_flags = 'rb'
 
-    def __init__(self, data=None, filename=None, url=None, embed=None, rate=None, autoplay=False):
+    def __init__(self, data=None, filename=None, url=None, embed=None, rate=None, autoplay=False, norm=True):
         if filename is None and url is None and data is None:
             raise ValueError("No image data found. Expecting filename, url, or data.")
         if embed is False and url is None:
@@ -97,7 +100,7 @@ class Audio(DisplayObject):
         super(Audio, self).__init__(data=data, url=url, filename=filename)
 
         if self.data is not None and not isinstance(self.data, bytes):
-            self.data = self._make_wav(data,rate)
+            self.data = self._make_wav(data,rate,norm)
 
     def reload(self):
         """Reload the raw data from file or URL."""
@@ -112,7 +115,7 @@ class Audio(DisplayObject):
         else:
             self.mimetype = "audio/wav"
 
-    def _make_wav(self, data, rate):
+    def _make_wav(self, data, rate, norm):
         """ Transform a numpy array to a PCM bytestring """
         import struct
         from io import BytesIO
@@ -144,8 +147,12 @@ class Audio(DisplayObject):
             except TypeError:
                 # this means it's not a nested list, which is what we want
                 pass
-            maxabsvalue = float(max([abs(x) for x in data]))
-            scaled = [int(x/maxabsvalue*32767) for x in data]
+            if (norm):
+                maxabsvalue = float(max([abs(x) for x in data]))
+                scaled = [int(x/maxabsvalue*32767) for x in data]
+            else:
+                scaled = [int(x*32767) for x in data]
+
             nchan = 1
 
         fp = BytesIO()
@@ -253,7 +260,7 @@ class YouTubeVideo(IFrame):
 
     Other parameters can be provided as documented at
     https://developers.google.com/youtube/player_parameters#Parameters
-    
+
     When converting the notebook using nbconvert, a jpeg representation of the video
     will be inserted in the document.
     """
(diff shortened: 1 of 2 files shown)


takluyver:
BTW, there's no need to make a new PR if you spot another thing you need to change. Just push to the same branch, and the existing PR is updated. This is how we work if a reviewer asks for a change as well.

drscotthawley:
@takluyver  Sure thing!  Just did.  
Also, am now clipping the non-normed audio just to be on the safe side; otherwise you get runtime errors if the abs() value of the audio exceeds 1.

takluyver:
Hmm, the thing about clipping makes me wonder if this is actually a good API. E.g. if you pass in data as 16-bit integers, which appears to be the native data type of .wav files, it will clip them to +/- 1 and you won't hear anything useful.

Maybe if you don't want the data normalised by max absolute value, you should pass in a number to use instead? I.e. in your case, you would pass in `1.0`, to say that your waveform is within the range +/- 1. If your data goes outside that range, it would then throw an error.

I don't know what to call that parameter, or if that even makes any sense.

drscotthawley:
Oh, ok, interesting.  I've only ever used floating-point -1 to 1.  I looked at the code for scipy.io.wavfile.write, and it looks as though one could clip or scale by the appropriate maximum-possible value(s), based on detecting the datatype of the input signal.  I could work on that in the next day or two.

eranegozy:
I was just about to look into this myself, when I saw this PR. Would love to have this merged in if possible.

takluyver:
Sorry for the delay. I'm happy with the API, but I'm not actually sure that this would raise an error with out-of-range values - it looks like numpy's type conversion accepts out-of-range values and wraps around:

```
In [15]: np.int16([2**18], wrap=False)                                          
Out[15]: array([0], dtype=int16)
```

Presumably this would result in distorted audio or noise depending on how far beyond the range the numbers are. I think throwing an error on unexpected input would be preferable. I can't see any functionality in numpy to cast and check for overflows, so we may have to construct the array first, check it for overflows, and then cast to int16.

takluyver:
And I think an error is preferable to clipping. If you're outside either range (+-1 or int16), there's a good chance that you're on totally the wrong scale, not just a little bit off, so the resulting audio will be terrible. It's much easier to figure out the problem from an error message than from mangled audio.

eranegozy:
Hmm.. I personally think that if norm=False, internally, clipping is the right thing to do, like with
`np.clip(x, -1, 1)`
But really, if a user sets norm=False, they are then responsible for understanding their data's range, and what the API expects. I don't think that the current proposed implementation, of accepting values either (-1, 1) for float or (-2^15, +2^15) for non-float is quite right.

One way around this is to have the user specify the max value through the API. So instead of `norm=<True | False>`, pass in `max-value`. So `max-value=None` would be the current default, meaning "Audio calculates the max value from the data and normalizes by that" and `max-value=<number>` would mean use this number as the max-value.

astrocox:
I was just digging to see if it's possible to disable normalization, so 👍 for this whole thread!
I agree with @eranegozy that clipping would be preferable to an error if a user explicitly sets `norm=False`, but all these options seem reasonable and it would be awesome to see any of them merged 😄

matangover:
I would like to help with this, but unsure how. Thanks @drscotthawley for your previous efforts! My understanding of the current state:
 * Right now `Audio` supports arrays (or lists) of several data types (floating-point, signed ints), because it always normalizes by the largest input sample. Unsigned ints (PCM8) are not supported. See [common data formats](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.write.html#id1).

 * Adding `norm=False` requires `Audio` to know the input data boundaries (min/max sample value, or only max if supporting only signed data types). Possible solutions:

   1. Guess the data boundaries based on the numpy array dtype. This is what `scipy.io.wavfile.write` does. Cons: implicit and might be confusing. Requires detailed documentation. Also, for list input we can either guess based on the list values (as @drscotthawley did), which is even more implicit, or throw an error for `norm=False`.

  2. Support only floating-point audio in range [-1, 1] for `norm=False`. Easiest to document and implement. Very straightforward for users. Seems like this is what most users want, although I didn't conduct a survey :)

  3. Require user to specify data format ('Explicit is better than implicit'...). This is what PySoundFile does ([subtype is set manually](https://pysoundfile.readthedocs.io/en/0.9.0/#soundfile.read)). @takluyver and @eranegozy have suggested adding a `maxvalue` parameter. I suggest instead adding a `wav_format` parameter that takes values 'float', 'PCM_16', 'PCM_24', 'PCM_32', 'PCM_U8' (values from ```soundfile.available_subtypes('wav')```). Pros: more readable than 'maxvalue', supports all common formats including unsigned PCM8 (maxvalue=1337 is possible but definitely not needed). `wav_format` could be set to `float` by default.

Regarding clipping, I think that if the boundaries are made explicit then an error makes sense.

@takluyver would you accept a PR for option 2 or 3?

matangover:
Or if you prefer I can just continue @drscotthawley's work and fix as per your comments for out-of-range values (option 1).

drscotthawley:
Hey folks, I'm ok with whatever you decide to do.  For my own work I've moved on, and I just always use a modified version of the Audio class where I've disabled the normalization.  I understand that may not fit in the grand scheme of what all other users need. 
If you want to delete/close this PR and submit a different one, that's fine with me, or keep this going... Your call.

eranegozy:
FWIW, my usage case is indeed floating-point [-1, 1] for norm=False case, so I would be happy with option 2. But I can see option 3 being more general. If we do go with that, I think wav_format=float by default is a good idea.

Also, I am not a fan of raising an error if values are out of bound. I would advocate clipping to the min/max values in that case. This is what some audio drivers would do anyway.

Eran

takluyver:
I'm fine with option 2 (only support floats -1 to +1). Option 3 strikes me as offputting complexity, as I don't know about the details of the wav format - but if that's preferred by others, I'm happy to review a PR.

I still maintain that out-of-range values should cause an error rather than be clipped. As above, if your data goes beyond +- 1, it may well be on the wrong scale completely, not just slightly out.

eranegozy:
> On Mar 12, 2019, at 4:38 PM, Thomas Kluyver <notifications@github.com> wrote:
> I'm fine with option 2 (only support floats -1 to +1). Option 3 strikes me as offputting complexity, as I don't know about the details of the wav format - but if that's preferred by others, I'm happy to review a PR.
> 
I agree with this.
> I still maintain that out-of-range values should cause an error rather than be clipped. As above, if your data goes beyond +- 1, it may well be on the wrong scale completely, not just slightly out.
> 
That’s fine. That argument makes sense.
-Eran

matangover:
OK, great. Thanks for your input everyone. Working on option 2.

matangover:
@drscotthawley feel free to close this one as #11650 has been merged 🎉

drscotthawley:
@matangover  ok, done!

musicnerd:
This is still not in the docs, and I get an error using the most up-to-date version that "__init__() got an unexpected keyword argument 'norm'". Am I missing something?

matangover:
Hi @musicnerd, the argument name is `normalize`. I can see it [here](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.Audio) in the docs. Does this work?

musicnerd:
Thanks @matangover I was on an older version of the docs. Oops! Thank you :)

PR #11164, opened 2018-05-30
pretty-rendering for os.environ by default

minrk:
it's dict-like, but not a dict so doesn't get caught by the default dict printer


takluyver:
That makes sense.

PR #11173, opened 2018-06-07
Correctly tearDown some test cases with multiple inheritance.

Carreau:
On Python 3.7+ this leads to Resources warnings at testing suite
shutdown. Fix this at least for IPython.core.tests.test_interactiveshell.

-- 

I'm not really sure why the teardown is not done correctly when the first class we inherit from is `unittest.TestCase` as it does not have a tearDown.

### DIFF
--- a/IPython/core/tests/test_interactiveshell.py
+++ b/IPython/core/tests/test_interactiveshell.py
@@ -546,6 +546,10 @@ class ExitCodeChecks(tt.TempFileMixin):
             else:
                 del os.environ['SHELL']
 
+    def tearDown(self):
+        tt.TempFileMixin.tearDown(self)
+
+
 class TestSystemRaw(unittest.TestCase, ExitCodeChecks):
     system = ip.system_raw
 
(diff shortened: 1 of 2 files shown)


IPython/core/tests/test_interactiveshell.py
```diff
@@ -546,6 +546,10 @@ def test_exit_code_signal_csh(self):
             else:
                 del os.environ['SHELL']
 
+    def tearDown(self):
```
minrk: Isn't the right fix to re-order the inheritance so the Mixin comes first?

Carreau: > I can retry though.

Hum, some seem to work, It may have been a random failure when I tried. I'll re-check and do that where it works.

Carreau: Pfff, PEBKAC, I was testing without #11174 merged and so was still seing ResourceWarnings.

Carreau:
Done by swapping base class.  The resource warnings were not showing on 3.6 but showing on nightly.

PR #11177, opened 2018-06-07
Upgrade to prompt_toolkit 2.0

takluyver:
Continuation of #10964

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -8,15 +8,14 @@ from .ptutils import IPythonPTCompleter
 from .shortcuts import suspend_to_bg, cursor_in_leading_ws
 
 from prompt_toolkit.enums import DEFAULT_BUFFER
-from prompt_toolkit.filters import (Condition, HasFocus, HasSelection,
-    ViInsertMode, EmacsInsertMode)
-from prompt_toolkit.keys import Keys
-from prompt_toolkit.key_binding.manager import KeyBindingManager
+from prompt_toolkit.filters import (Condition, has_focus, has_selection,
+    vi_insert_mode, emacs_insert_mode)
+from prompt_toolkit.key_binding import KeyBindings
 from prompt_toolkit.key_binding.bindings.completion import display_completions_like_readline
-from prompt_toolkit.token import Token
-from prompt_toolkit.shortcuts import create_prompt_application
-from prompt_toolkit.interface import CommandLineInterface
+from pygments.token import Token
+from prompt_toolkit.shortcuts.prompt import PromptSession
 from prompt_toolkit.enums import EditingMode
+from prompt_toolkit.formatted_text import PygmentsTokens
 
 
 class TerminalPdb(Pdb):
@@ -26,11 +25,8 @@ class TerminalPdb(Pdb):
         self.pt_init()
 
     def pt_init(self):
-        def get_prompt_tokens(cli):
+        def get_prompt_tokens():
             return [(Token.Prompt, self.prompt)]
-
-        def patch_stdout(**kwargs):
-            return self.pt_cli.patch_stdout_context(**kwargs)
 
         if self._ptcomp is None:
             compl = IPCompleter(shell=self.shell,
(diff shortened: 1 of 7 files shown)


takluyver:
The remaining failure is in generating the shortcuts list for the docs, if anyone wants to update that script:

https://github.com/ipython/ipython/blob/master/docs/autogen_shortcuts.py

gpotter2:
Hi !
It may not be the real correct way, but here's a custom fix that works

```
diff --git a/docs/autogen_shortcuts.py b/docs/autogen_shortcuts.py
index 95c8d608b..8535f1c28 100755
--- a/docs/autogen_shortcuts.py
+++ b/docs/autogen_shortcuts.py
@@ -1,7 +1,6 @@
 from os.path import abspath, dirname, join
 
-from IPython.terminal.interactiveshell import KeyBindingManager
-
+from IPython.terminal.shortcuts import create_ipython_shortcuts
 
 def name(c):
     s = c.__class__.__name__
@@ -42,8 +41,14 @@ def multi_filter_str(flt):
 log_filters = {'_AndList': 'And', '_OrList': 'Or'}
 log_invert =  {'_Invert'}
 
-kbm = KeyBindingManager.for_prompt()
-ipy_bindings = kbm.registry.key_bindings
+class _DummyTerminal(object):
+    """Used as a buffer to get prompt_toolkit bindings
+    """
+    handle_return = None
+    input_splitter = None
+    display_completions = None
+
+ipy_bindings = create_ipython_shortcuts(_DummyTerminal()).bindings
 
 dummy_docs = []  # ignore bindings without proper documentation
 

```

takluyver:
Thanks, I applied that patch.

gpotter2:
@takluyver Is there anything else to do for prompt_toolkit 2.0 I could help with ?

gpotter2:
@takluyver Also, 

> My normal inclination would be to do this for 7.0, but I know that people have had problems with tab completion running in a separate thread, and so we may end up doing it as a bugfix, possibly even back to 5.x. The remaining Python 2 users now can't go past 5.x, so there's some extra pressure to backport things.

Are you still willing to backport this ? I see it was only added to 7.0

takluyver:
I'm not sure :confused: . It's ended up being a fairly big change, which probably won't backport cleanly, and may cause other problems.

There's a larger shift in the Python world towards each application having its own dependencies installed in a separate environment. IPython doesn't entirely fit with that, because it's not just any application - you want it installed in the same environment as the modules you're working with. But then prompt_toolkit is a module that's largely useful only for applications.

So for now, I think I'm inclined to push forwards towards 7.0, and see how things go. IPython 5.7 correctly specifies the versions of prompt_toolkit it works with, so it's not a bug that it's incompatible with newer versions.

gpotter2:
Ok, this will be an additional point to make people switch to Python 3. Also it won’t break any app using both IPython and prompt_toolkit 1.0

Only problem we had with python 2.7 is that when we wanted to add prompt_toolkit 2.0 features to scapy, we were blocked by IPython.
We don’t have separate branches as IPython has, but that’s pretty easy to handle anyway

stonebig:
when will we have a chance to have a Ipython compatible version ? Ipython-7.0 seems months away

stonebig:
any hope for today ?

2sn:
yes, it would be great to have IPython work with current versions of modules.  As you note, prompt-tooolkit 2.x is incompatible with version 1,0.x, so if you have other applications that have been updated, you have to pick which module you prefer or need - IPython or the other ... !  Maybe you need a branch using the new version; users who need the old prompt-toolkit for other apps - and there may be some, can continue use the 6.x series of IPython ...

ProbonoBonobo:
(Cross-posted from https://github.com/jonathanslenders/python-prompt-toolkit/issues/647)

For anyone needing a temporary workaround, I can confirm that reverting to v.1.x mostly resolves the errors that prevent ipython, ptpython, etc. from initializing:

```

git clone -b 1.0  --single-branch https://github.com/jonathanslenders/python-prompt-toolkit.git
cd python-prompt-toolkit
python3 setup.py install
```


You might get the following error when launching ptpython if you're using the [default `config.py` template](https://github.com/jonathanslenders/ptpython/blob/master/examples/ptpython_config/config.py)...

```
~/dev/git/clones/python-prompt-toolkit$ ptpython
  File "/home/kevin/.local/lib/python3.6/site-packages/ptpython/repl.py", line 244, in run_config
    six.exec_(code, namespace, namespace)
  File "/home/kevin/.ptpython/config.py", line 8, in <module>
    from prompt_toolkit.key_binding.key_processor import KeyPress
ModuleNotFoundError: No module named 'prompt_toolkit.key_binding.key_processor'

```
...but it's not fatal, and can be safely ignored, or you could delete the offending lines. A quick glance at the code suggests KeyPress is only used to map the 'jj' key sequence to 'Esc' in vim mode.

PR #11190, opened 2018-06-15
Fix config option TerminalInteractiveShell.display_completions

infmagic2047:
The extra comma in pt_complete_style() causes the option to be always
recognized as "column", because it compares equal to none of the three
values.


takluyver:
Thanks!

takluyver:
Looks like the problem was introduced by the prompt_toolkit 2 work for 7.0, so it doesn't affect 6.x.

PR #11192, opened 2018-06-16
Fixes the default value of `-r<R>` and improves documentation

alphaCTzo7G:
- Changes the default value of `-r<R>` from 3 to 7
- adds clarifications for `-r<R>`

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -962,11 +962,12 @@ python-profiler package from non-free.""")
           body has access to any variables created in the setup code.
 
         Options:
-        -n<N>: execute the given statement <N> times in a loop. If this value
-        is not given, a fitting value is chosen.
+        -n<N>: execute the given statement <N> times in a loop. If <N> is not
+        provided, <N> is determined so as to get sufficient accuracy.
 
-        -r<R>: repeat the loop iteration <R> times and take the best result.
-        Default: 3
+        -r<R>: number of repeats <R>, each consisting of <N> loops, and take the
+        best result.
+        Default: 7
 
         -t: use time.time to measure the time, which is the default on Unix.
         This function measures wall time.


IPython/core/magics/execution.py
```diff
@@ -965,8 +965,10 @@ def timeit(self, line='', cell=None, local_ns=None):
         -n<N>: execute the given statement <N> times in a loop. If this value
         is not given, a fitting value is chosen.
 
-        -r<R>: repeat the loop iteration <R> times and take the best result.
-        Default: 3
+        -r<R>: number of runs<R>, each consisting of <N> loops, and take the
+        best result. If <N> is not provided, <R> is determined so as to get
```
takluyver: It's `<N>` that changes to get sufficient accuracy, not `<R>` - and that note should be in the description of the `-n` flag.

Also, you've got rid of the word *repeat*, which is what the R stands for.

takluyver (commented):
Thanks, we do need to update the default. I think the proposed clarification needs a bit more work, though.

Carreau:
Also please be careful, your editor seem to have been set to trim all whitespaces, which adds a lot of irrelevant changes. Tanks.

alphaCTzo7G:
Thanks for the feedback.. I will work on this..

IPython/core/magics/execution.py
alphaCTzo7G: @takluyver.. when I was reading the documentation in the code.. I was a bit confused because when you run something like `%timeit -r3 100**100`, it will print out something like 

`12.4 ns ± 0.187 ns per loop (mean ± std. dev. of 3 runs, 100000000 loops each)`. 

So it seemed that `run` was a better description of what `-r` stands for. Otherwise.. would it be better to printout something like

`12.4 ns ± 0.187 ns per loop (mean ± std. dev. of 3 repeats, 100000000 loops each)`.

In that case I can change `runs` to `repeats`: https://github.com/alphaCTzo7G/ipython/blob/b26f764c594aa83d2bb1f6efcd6c94045f3f5fe0/IPython/core/magics/execution.py#L97?

If you guys think, its better to keep the word `repeat` in the documentation and `run` in the string printed out in `%timeit`.. I can just keep it as is and reword  the commit as well

alphaCTzo7G:
@Carreau @takluyver, I have corrected the whitespace issue, moved the message about <N> to its designated location and put the `repeat` back.. 

I added `number of repeats <R>, each consisting of <N> loops, and take the best result.` because when I originally read the string initially (`repeat the loop iteration <R> times and take the best result. Default: 3`), it wasn't very clear to me what was the `loop iteration`. 

If you guys think it would be easier to understand the message if we use `repeat` instead of `run` in the message that is printed out `12.6 ns ± 0.152 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)`, let me know.. I can make the necessary changes here: https://github.com/ipython/ipython/pull/11192#discussion_r199327320

alphaCTzo7G:
@Carreau , @takluyver .. any comments.. I forced pushed it to the same branch..

Carreau:
Thanks.

Issue #11198, opened 2018-06-22
Script (e.g. Bash) magics doesn't have an error on status code


MSeal:
I ran into a use-case where I wanted to call a small bash script magic in a notebook, but couldn't halt execution easily based on exit status.

======= Cell 1 =======
```
%%bash
not_a_valid_command
```

======= Cell 2 =======
```
print("I don't want to run this if the bash command didn't return status code 0")
```

It seems like the [script magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html) and the [code](https://github.com/ipython/ipython/blob/ec3d1a11bf26a0962cb2cf55ba263b12ac023183/IPython/core/magics/script.py#L63) don't have any options to inform the subprocess to error on non-zero status code. This would be nice to be configurable and/or defaulted.

MSeal:
I've been having a lot of feedback from other users at my company who get frustrated that this isn't the default behavior (both for `%%script` and `!` patterns). For example `!pip install -U package-that-does-not-exist` will not halt execution, but this is a common pattern that cannot assume success to continue execution.

How disruptive would it be if we changed the default behavior to raise an error for nonzero exit codes by default?

MSeal:
@Carreau thoughts on this topic?

Carreau:
Thanks for pinging me. 
I think that make sens for consistency. It would be relatively easy to add a configuration option to change the behavior. 

Would that be ok only for the next major version ? 

Note that it should be relatively easy to backport as a custom package/extension for IPYthon, where you can just overwrite magics, if you want to have it on earlier versions.

MSeal:
Ok let me see if I can make a PR to change the default behavior to make use of this for the 7.0 milestone.

PR #11199, opened 2018-06-22
create new --raise-error argument to raise an error if nonzero exitcode

mpacer:
This creates a way for users to stop executing cells if a script magic raises an error.

e.g., 

```
%%bash --raise_error
exit 1
```

Would raise an error and stop execution instead of continuing execution if one ran all cells.

@MSeal brought this up with #11198

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -8,7 +8,7 @@ import os
 import sys
 import signal
 import time
-from subprocess import Popen, PIPE
+from subprocess import Popen, PIPE, CalledProcessError
 import atexit
 
 from IPython.core import magic_arguments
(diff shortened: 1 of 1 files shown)


IPython/core/magics/script.py
```diff
@@ -54,6 +54,12 @@ def script_args(f):
             This is used only when --bg option is given.
             """
         ),
+        magic_arguments.argument(
+            '--raise_error', action="store_true",
```
Carreau: `--raise-error` (dash instead of underscore) seem more common for flags ?

Carreau (approved):

IPython/core/magics/script.py
MSeal: That seems reasonable

mpacer:
Hi @Carreau are there any other changes you want to see happen before this can be merged?

Carreau:
Looks good, thanks !

PR #11211, opened 2018-07-01
Allow specifying the name of the root element of a JSON display object

dhirschfeld:


dhirschfeld:
PR as suggested by @gnestor in https://github.com/jupyterlab/jupyterlab/pull/4436#issuecomment-384466742

I guess the benefit of of this is in explicitly documenting this capability (for frontends which support it)

gnestor (changes requested):
Looks good to me but it looks like the `test_json` tests are failing because they're not expecting a `root` arg: https://github.com/ipython/ipython/blob/master/IPython/core/tests/test_display.py#L213-L240

dhirschfeld:
^^^will look into that.

gnestor (approved):
Looks good!

Carreau:
Thanks !

PR #11234, opened 2018-07-22
Upgrade pip before setuptools for Python compat.

Carreau:
Setuptools recently dropped Python 3.3 and with an out of date pip we
know install an incompatible setuptools on the 6.x branch.

This should avoid the same error on future python drops.

Likely to be backported to 5.x and 6.x


Carreau:
Thanks ! Backported to 5.x manually and merged into 6.x

PR #11250, opened 2018-08-03
Use raw string when invalid escape sequence present.

Carreau:
Those emit deprecation warnings at import/compile time so unfortunately
we can't trap that with iptest as modules are already imported.

The way to test that is to start IPython via `python -We -m IPython`,
and we should try to likely do that in travis.

But first fix all places that raise an error


Carreau:
Thanks !

PR #11253, opened 2018-08-04
Fix breaking newline in Index.ipynb

parente:
Fixes https://github.com/jupyter/nbviewer/issues/787


Carreau:
Oops. Good catch ! And good job tracking things down.

PR #11255, opened 2018-08-04
Add .python-version to gitignore

alyssawhitwell:
- Add `.python-version` from `pyenv` to the repo's gitignore file

cc: @ivanov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ __pycache__
 *.swp
 .vscode
 .pytest_cache
+.python-version


Carreau:
Thanks, 

If you use PyEnv it would be preferable to set up a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) as using pyenv is a decision you make and not one IPython did. 

Still merging this at it will be of help for other, but setting up your global gitignore is usually a good idea.

PR #11256, opened 2018-08-04
Refactor modules_path.py to use importlib instead of imp

alyssawhitwell:
**In this PR:**
- [x] Deprecated use of `imp` library since it is pending deprecation
- [x] Refactored module to use `importlib.util.find_spec`: https://docs.python.org/3/library/importlib.html#importlib.util.find_spec
- [x] Condensed module from three functions to one function, `find_mod`, which includes all logic and edge cases for returning `None` instead of the filepath
- [x] Updated tests

**Questions:**
- [ ] The original code in `modules_path.py` indicates that we want to ignore bytecode: `Return full path of module or None if module is missing or does not have .py or .pyw	extension. We are not interested in running bytecode.`. However, there was an existing unit test that explicitly checked that `.pyc` filepaths **were** returned, rather than returning `None`. **Is it correct to assume that we want to ignore filepaths that lead to `.pyc` files?** This assumption has been made in this PR.

cc: @ivanov @Carreau

### DIFF
--- a/IPython/utils/module_paths.py
+++ b/IPython/utils/module_paths.py
@@ -2,14 +2,7 @@
 
 Utility functions for finding modules on sys.path.
 
-`find_mod` finds named module on sys.path.
-
-`get_init` helper function that finds __init__ file in a directory.
-
-`find_module` variant of imp.find_module in std_lib that only returns
-path to module and not an open file object as well.
-
-
+`find_module` returns a path to module or None, given certain conditions.
 
 """
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Carreau:
> he original code in modules_path.py indicates that we want to ignore bytecode: Return full path of module or None if module is missing or does not have .py or .pyw extension. We are not interested in running bytecode.. However, there was an existing unit test that explicitly checked that .pyc filepaths were returned, rather than returning None. Is it correct to assume that we want to ignore filepaths that lead to .pyc files? This assumption has been made in this PR.

I believe ignoring pyc is the right thing. This appear to be use in `deepreload` and `%run`, which you likely want to run on modified `.py`  files. So that seem the right call. 

It also seem to be used _only_ there, so at some point we may want to move that somewhere else. 

Awesome work, that looks good to me, so merging and we'll see if we get complaints :-)

PR #11265, opened 2018-08-14
Prototype async REPL using IPython, take III

Carreau:
This is a squash and a rebase of a large number of commits from Min and
I. For simplicity of managing it, history has been reduced to a single
commit, but more historical versions can be found, in particular in
PR #11155, or commit aedb5d6d3a441dcdb7180ac9b5cc03f91329117b to be more
exact.

-- 

I'm expecting some errors in the squash and resolve of conflicts, so will iterate on it.


Carreau:
I'm thinking of dropping 3.4, as maintaining compat with it will be painful:

> There are no specific plans for the next release of Python 3.4. However, the expectation is that Python 3.4.10 will be released in March of 2019, and this will be the final release of Python 3.4.

We'll also be supporting 3.5, 3.6 and 3.7.

@minrk  and @takluyver thoughts ?

minrk:
asyncio support is pretty valuable and long overdue, so if 3.4 makes it too hard, I think the trade-off is worth it.

setup.py
```diff
@@ -201,6 +201,7 @@
 
 extras_require.update({
     ':python_version == "3.4"': ['typing'],
+    ':python_version >= "3.5"': ['trio', 'curio'],
```
minrk: Neither of these are dependencies, right? We shouldn't require these packages.

Carreau: One of them is necessary if we want to have  nested `embed()`.  I can make that optional, I probably added them here to have the test passing.

dalejung:
@Carreau I ran into an issue running python files that have their own loop calls. So if I have a file that calls `run_until_complete`, it'll break when %run since it nests within the loop_runners own call.

Can we change the run_cell_async coro to first yield interactivity so it can call the pseudo sync when async is not required?

https://github.com/dalejung/ipython/commit/8d38b3ce41953d8d9048092cc283eb7662a768ed

is a working example. `await` statements in the shell work while running or calling funcs/files that invoke their own loops work as well.

Carreau:
I have to better understand the implication. It is likely possible, though it will likely work only on terminal IPython. 

It might also not stay forever, as in the end I want IPython to always run on an event loop. Though, I guess we could run only on the pseudo-sync, or a non-asyncio one. 

Once that hapend we can figure out how to patch `%run` in order to patch `run_until_complete`

njsmith:
The way I would imagine it working in the long run is:

* The `async` annotations spread out from `run_cell` to infect the rest of IPython, so that the whole program runs under a single call to the event loop
* By default IPython starts out in 'sync' mode, which means:
  * Using the pseudo-sync runner
  * Raise `SyntaxError` if someone tries to use a top-level `await` in a cell
* You can switch from 'sync' mode to 'asyncio' mode, 'trio' mode etc., which switches IPython to using these loops
  * Internally, this loop switcheroo can be accomplished by yielding a special value to the pseudo-sync runner, which then resumes the same coroutine object under the new loop. This is serious black magic but should work...
  * Also it sets a flag internally to start allowing top-level `await` inside cells
* You can use configuration, pass a command-line flag, etc., to preemptively switch into one of the async modes at startup

Of course there's a ways to go before we use async annotations everywhere like that. But we need to eventually reach a point of keeping the same loop alive over multiple cells, if we want this to be really useful. In any case, I think the basic idea of tracking whether we're in sync or async mode, and turning the magic parser on-or-off as appropriate, is a good one to expose to the user. And that's something that can be done now and would solve @dalejung's problem.

njsmith:
...of course when writing the above I forgot about prompt-toolkit. Besides it needing some work to integrate with all these different loops, I'm not sure that it'll be happy switching between loops mid-stream...

Well, the architecture I described above ought to work for the kernel, or for `jupyter console`, at least :-). (With no GUI integration enabled.)

Carreau:
> I forgot about prompt-toolkit

Yep, that was one of my main concern, but we can see once we're there. Seeing that it took us 2 years between prototype and now, maybe by then code Python will have a better solution for `async-exec` and multiple loop.

tacaswell:
> (With no GUI integration enabled.)

GUI integration is pretty key....

The way the input hooks are implemented now they are very much preemptive rather than cooperative which seems like a hard square to fit into a round hole. 

The way we deal with this at my day-job is having a self-re-triggering task that exhausts the GUI task queue and then schedules its self for some small time in the future. 

@Carreau How through of a review do you want on this (on a scale of "theory sounds good" to "picky implementation")?

IPython/core/interactiveshell.py
```diff
@@ -258,6 +358,43 @@ class InteractiveShell(SingletonConfigurable):
         """
     ).tag(config=True)
 
+    autoawait = Bool(True, help=
+        """
+        Automatically run await statement in the top level repl.
+        """
+    ).tag(config=True)
+
+    loop_runner_map ={
```
pganssle: Note: For another PR, it would be a good idea to have a way to register new names for this.

docs/source/interactive/autoawait.rst
```diff
@@ -0,0 +1,210 @@
...
+
+
+You can use the ``c.InteractiveShell.autoawait`` configuration option and set it
+to :any:`False` to deactivate automatic wrapping of asynchronous code. You can also
+use the :magic:`%autoawait` magic to toggle the behavior at runtime::
+
+    In [1]: %autoawait False
+
+    In [2]: %autoawait
+    IPython autoawait is `Off`, and set to use `IPython.core.interactiveshell._asyncio_runner`
```
pganssle: This seems like it is exposing an implementation detail. it would probably be worth defining a `__str__` or something on the runners so that they reflect more closely how they were activated.

pganssle:
Testing this out, it seems like this breaks `%paste`, even without autoawait on:

```python
$ ipython
Python 3.7.0 (default, Jul 30 2018, 14:36:22) 
IPython 7.0.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: %paste                                                                                                                              
def parseprint(code, filename="<string>", mode="exec", **kwargs):
    """Parse some code from a string and pretty-print it."""
    node = parse(code, mode=mode)   # An ode to the code
    print(dump(node, **kwargs))
    
def pn(tree):
    print(dump(tree))

## -- End pasted text --
<ipython-input-1-cda6375ef3e5> in <module>
----> 1 get_ipython().run_line_magic('paste', '')

....

    509         if self.is_running():
--> 510             raise RuntimeError('This event loop is already running')
    511         if events._get_running_loop() is not None:
    512             raise RuntimeError(

RuntimeError: This event loop is already running
```

Carreau:
@dalejung  I've added your commit to this pull request, it does fix @pganssle use case as well as %paste and %cpaste were internally running `run_cell`, which was thus nested.

IPython/core/tests/test_async_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+            """
+            )
+
+        def test_autoawait(self):
+            ip.run_cell("%autoawait False")
+            ip.run_cell("%autoawait True")
+            iprc(
+                """
+                from asyncio import sleep
+                await.sleep(0.1)
```
pganssle: This is an example of where the test isn't working right - `await.sleep` is incorrect.

That said, it turns out that:

`iprc("""
from asyncio import sleep
await sleep(0.1)
""")

is *also* a `SyntaxError`, for inscrutable reasons.

pganssle: Found the problem:

```python
In [1]: from IPython.core.async_helpers import _asyncify                                                                                    
In [2]: from textwrap import dedent
In [3]: print(_asyncify(dedent(""" 
   ...: import asyncio 
   ...: await sleep(0.001) 
   ...: """)))                                                                                                                              

async def __wrapper__():
    try:
         import asyncio
        await sleep(0.001)

    finally:
        locals()
```

Note the erroneous indentation on the `from asyncio`. I think `_asyncify` is transforming the `\n` into a space somewhere.

pganssle:
@Carreau I've pushed two commits to the branch adding support for properly detecting syntax errors.

Also, it seems that the tests weren't actually failing if the test failed, because errors weren't being raised, and indeed several of the tests *should* have been failing. I've pushed two more commits fixing the tests and a bug in `_asyncify` that was causing test failures.

Carreau:
Todo, catch exception when the runner crash, EG, when using `autoawait trio`, and then `await asyncio.sleep(0)`.

Carreau:
Timeit, time, and prun magics will need updates to work with async code. 
They still work for sync code though, so I don't think it is a blocker.

docs/source/whatsnew/pr/deprecations.rst
```diff
@@ -1,3 +1,6 @@
+Depreations
```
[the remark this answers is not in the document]
Carreau: Thanks, Fixed.

IPython/core/interactiveshell.py
```diff
@@ -2780,9 +2984,14 @@ def error_before_exec(value):
...
                 # Execute the user code
-                interactivity = 'none' if silent else self.ast_node_interactivity
-                has_raised = self.run_ast_nodes(code_ast.body, cell_name,
-                   interactivity=interactivity, compiler=compiler, result=result)
+                interactivity = "none" if silent else self.ast_node_interactivity
+                if _run_async:
+                    interactivity = 'async'
+                # yield interactivity so let run_cell decide whether to use
+                # an async loop_runner
+                yield interactivity
```
minrk: Adding `yield interactivity` here makes this a not-entirely-valid coroutine, doesn't it? When I try to call `await run_cell_async` from ipykernel, I get a `bad yield: 'last_expr'`. How can we communicate this required info without faking that this isn't actually a coroutine? Does it need to be an actual async generator?

Carreau:
That's right. You need to do a small dance. I believe I have a patch on my
machine for ipykernel to make this work.
You need to advance the coro by sending None to it once. We could do it in
another way, but at some point we'll always hit the fact that python does
not have a real async-exec and that we need to distinguish 2 path of
running sync code, or asyc one.

Carreau:
Ok, seem like we've reached quasi static state and most of the latest updates were documentation. Plus ipykernel works with that. So I'm going to merge (thanks all involved !), and we can refine if necessary.

minrk:
#11289 makes run_cell_async a regular coroutine by moving the 'should it be async' check to a separate method.

Since in general, running it with a coroutine runner ought to work (I know there are edge cases for now), skipping this check should be safe.

PR #11266, opened 2018-08-16
Fix `up` through generators in postmortem debugger; closes #6251

### LINKED ISSUE #6251: %debug magic "up" fails to pass through generator stack
Consider the following example "pdbtest.py":

```
def f(x):
    raise Exception

gen = (f(x) for x in [0])

for x in gen:
    pass
```

If I run this with `python -i pdbtest.py`, when the exception is thrown I can go up the stack all the way to the module-level with pdb:

```
  File "pdbtest.py", line 6, in <module>
    for x in gen:
  File "pdbtest.py", line 4, in <genexpr>
    gen = (f(x) for x in [0])
  File "pdbtest.py", line 2, in f
    raise Exception
Exception
>>> import pdb; pdb.pm()
> /home/antony/tests/pdbtest.py(2)f()
-> raise Exception
(Pdb) u
> /home/antony/tests/pdbtest.py(4)<genexpr>()
-> gen = (f(x) for x in [0])
(Pdb) u
> /home/antony/tests/pdbtest.py(6)<module>()
-> for x in gen:
(Pdb) u
*** Oldest frame
(Pdb) 
>>> 
```

However, if I try to do the same with ipython's %debug magic, I cannot go beyond the generator frame -- which is particularly annoying when you think the actual bug happens earlier:

```
In [1]: %debug
> /path/to/pdbtest.py(2)f()
      1 def f(x):
----> 2     raise Exception
      3 

ipdb> u
> /path/to/pdbtest.py(4)<genexpr>()
      3 
----> 4 gen = (f(x) for x in [0])
      5 

ipdb> u
*** Oldest frame
ipdb>
```

This happens with ipython 2.1.0 and python 3.4.

rhelmot:
By passing the lowest traceback element into the debugger, we require it to use frame.f_back to find older frames. However, [f_back is always None for generator frames in the postmortem context](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Objects/genobject.c#L226-L230). By providing a higher-up traceback element, pdb can traverse down the traceback.tb_next links, which do work correctly across generator calls.

...or at least, that's my understanding of the situation after spending half an hour researching the issue. This fix seems to work for my use cases, but I don't really understand the depths of pdb and can't tell if this may cause adverse affects elsewhere.

(I would also love to write a test for this, but I have no idea if we can or want to drive the debugger automatically for tests?)

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -1203,7 +1203,7 @@ class VerboseTB(TBTools):
                 if etb and etb.tb_next:
                     etb = etb.tb_next
                 self.pdb.botframe = etb.tb_frame
-                self.pdb.interaction(self.tb.tb_frame, self.tb)
+                self.pdb.interaction(None, etb)
 
         if hasattr(self, 'tb'):
             del self.tb
--- a/IPython/terminal/tests/test_debug_magic.py
+++ b/IPython/terminal/tests/test_debug_magic.py
@@ -1,0 +1,74 @@
+"""Test embedding of IPython"""
+
+#-----------------------------------------------------------------------------
+#  Copyright (C) 2013 The IPython Development Team
+#
+#  Distributed under the terms of the BSD License.  The full license is in
+#  the file COPYING, distributed as part of this software.
+#-----------------------------------------------------------------------------
+
+#-----------------------------------------------------------------------------
+# Imports
+#-----------------------------------------------------------------------------
+
+import os
+import sys
+from IPython.testing.decorators import skip_win32
+
+#-----------------------------------------------------------------------------
+# Tests
+#-----------------------------------------------------------------------------
+
+@skip_win32
+def test_debug_magic_passes_through_generators():
+    """
+    This test that we can correctly pass through frames of a generator post-mortem.
+    """
+    import pexpect
+    import re
+    in_prompt = re.compile(b'In ?\[\\d+\]:')
+    ipdb_prompt = 'ipdb>'
+    env = os.environ.copy()
+    child = pexpect.spawn(sys.executable, ['-m', 'IPython', '--colors=nocolor', '--simple-prompt'],
+                          env=env)
+    child.timeout = 2
+
+    child.expect(in_prompt)
+    child.sendline("def f(x):")
+    child.sendline("    raise Exception")
+    child.sendline("")
+
+    child.expect(in_prompt)
+    child.sendline("gen = (f(x) for x in [0])")
+    child.sendline("")
+
+    child.expect(in_prompt)
+    child.sendline("for x in gen:")
+    child.sendline("    pass")
+    child.sendline("")
+
+    child.expect('Exception:')
+
+    child.expect(in_prompt)
+    child.sendline(r'%debug')
+    child.expect('----> 2     raise Exception')
+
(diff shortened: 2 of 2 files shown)


rhelmot:
...any updates on this? Is there something I can provide to help move this along?

Carreau:
Ok I was wrong for above, the only way I can see is to use subprocess and run python -i and check the output. Something similar to [IPython/terminal/tests/test_embed.py](https://github.com/ipython/ipython/blob/a85a08ccc63726c1a2ee8bde8fa05a3cb2955dd9/IPython/terminal/tests/test_embed.py)

I'll see if I can write a scaffold for that.

Carreau:
Ok, I'm writing some test, but this seem to not be _quite_ right:

```
In [1]: def f(x):
   ...:     raise Exception

In [2]: gen = (f(x) for x in [0])

In [3]: for x in gen:
   ...:     pass
<ipython-input-3-5b2155f9c4b2> in <module>
----> 1 for x in gen:
      2     pass

<ipython-input-2-b7f6c0ea3068> in <genexpr>(.0)
----> 1 gen = (f(x) for x in [0])

<ipython-input-1-a13efb0dffe1> in f(x)
      1 def f(x):
----> 2     raise Exception

Exception:


In [4]: %debug
> <ipython-input-1-a13efb0dffe1>(2)f()
      1 def f(x):
----> 2     raise Exception


ipdb> u
> <ipython-input-2-b7f6c0ea3068>(1)<genexpr>()
----> 1 gen = (f(x) for x in [0])

ipdb> u
> <ipython-input-3-5b2155f9c4b2>(1)<module>()
----> 1 for x in gen:
      2     pass

ipdb> u                                         |
> <ipython-input-1-a13efb0dffe1>(2)f()          |
      1 def f(x):                               |
----> 2     raise Exception                     | These two frames are wrong.
                                                |
ipdb> u                                         |
> <ipython-input-2-b7f6c0ea3068>(1)<genexpr>()  |
----> 1 gen = (f(x) for x in [0])               |
                                                
ipdb> u
*** Oldest frame
```

It show frames in the wrong place (or re-go through frames it should not)

I'm going to try to polish the tests framework to test that, and we can figure out what to do. I'm going to likely push thing for 7.1

rhelmot:
Okay, I read the pdb/bdb source more carefully and force pushed what should be a real fix. Explanation in the commit message.

The takeaway here for future reference is that the only place the frame passed into `interaction` is used is in `Bdb.get_stack`, which walks the stack backwards from the frame and forwards from the traceback to construct the full list of stack frames. The frame can be None and this will still work, but if the traceback and frame don't correspond to the same frame, it will duplicate parts of the stack.

Carreau:
Thanks, I'll have a look, thy to commit a test framework (or at least a draft of it).

Carreau:
Ok, I just pushed what should be test for this patch. 

It's a bit cumbersome to write, as you will see, but let's not optimize until we actually have need to write this often.

Carreau:
Once test passes i'll merge. 
Many thanks for your patience. 

@meeseeksdev[bot] untag help wanted
@meeseeksdev[bot] tag hacktoberfest

Carreau:
Tests are passing, let's get that in, many thanks !

PR #11267, opened 2018-08-16
Drop jquery dependency in IPython.display.Javascript

gnestor:
Closes https://github.com/jupyterlab/jupyterlab/issues/1818

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -852,17 +852,23 @@ class JSON(DisplayObject):
     def _repr_json_(self):
         return self._data_and_metadata()
 
-_css_t = """$("head").append($("<link/>").attr({
-  rel:  "stylesheet",
-  type: "text/css",
-  href: "%s"
-}));
+_css_t = """var link = document.createElement("link");
+	link.ref = "stylesheet";
+	link.type = "text/css";
+	link.href = "%s";
+	document.head.appendChild(link);
 """
 
-_lib_t1 = """$.getScript("%s", function () {
+_lib_t1 = """var script = document.createElement("script");
+	script.type = "text/javascript";
+	script.src = "%s";
+	script.onload = script.onreadystatechange = function() {
+    if (!this.readyState || this.readyState == 'complete') {
 """
-_lib_t2 = """});
-"""
+
+_lib_t2 = """}
+  };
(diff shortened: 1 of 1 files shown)


asford:
Should this include a doc update in [IPython/core/display.py#L953-L955](https://github.com/gnestor/ipython/blob/ac931b8c787704b969247ae0ca4d634be8a727c4/IPython/core/display.py#L953-L955)?

IPython/core/display.py
```diff
@@ -852,18 +852,24 @@ def _data_and_metadata(self):
...
 """
 
-_lib_t1 = """$.getScript("%s", function () {
-"""
-_lib_t2 = """});
+_lib_t1 = """var script = document.createElement("script");
+	script.type = "text/javascript";
+	script.src = "%s";
+	script.onload = script.onreadystatechange = function() {
+    if (!this.readyState || this.readyState == 'complete') {
```
rgbkrk: Is there a chance that this will never run the script, where `onload` or `onreadystatechange` are called yet `this.readyState` is not complete?

rgbkrk (approved):

IPython/core/display.py
```diff
@@ -852,18 +852,24 @@ def _data_and_metadata(self):
...
+	link.ref = "stylesheet";
+	link.type = "text/css";
+	link.href = "%s";
+	document.head.appendChild(link);
 """
 
-_lib_t1 = """$.getScript("%s", function () {
-"""
-_lib_t2 = """});
+_lib_t1 = """var script = document.createElement("script");
```
blois: Currently $.getScript returns a 'thenable' which can be awaited when the script has been executed.

An alternative which could preserve this is:
new Promise((resolve, reject) => {
  const script = document.createElement("script");
  script.onload = resolve;
  script.onerror = reject;
  script.src = "%s";
  document.head.appendChild(script);
}).then(() => {
  // data
});

This is a 'nice to have' as it primarily helps ordering and error handling at higher levels, but would require using the result when this item is displayed.

rgbkrk (approved):

gnestor:
@Carreau Care to merge?

rgbkrk:
_Fantastic_

Issue #11257, opened 2018-08-07
Cannot turn off autoindent at cli in Ipython 6


williamjameshandley:
running 
```bash
ipython --no-autoindent
````
does not turn off autoindentation.

neither does running
```ipython
In[1] %autoindent%
autoindent is deprecated since IPython 5: you can now paste multiple lines without turning autoindentation off.
Automatic indentation is: OFF
```
in the ipython interpreter.

I would like to be able to turn off autoindentation for the purposes of sending text to ipython in a sendkeys-like interface for [this project](https://github.com/williamjameshandley/vimteractive). Currently the only way I have of doing this is to start ipython in 
```bash
ipython --simple-prompt
```
but then I lose functionality like coloring/autocompletion.

Carreau:
Does something like https://github.com/ipython/rlipython fit your needs ? 
My guess is you actually want to bypass all the  IPython Gui and just send code to be executed.

williamjameshandley:
Unfortunately this doesn't seem to fix my issue, as when I try start ipython with
`ipython --TerminalIPythonApp.interactive_shell_class=rlipython.TerminalInteractiveShell`, and then try to send either individual lines, or multiple lines at once via vim's `term_send_keys`, ipython still tries to insert additional tabs

Carreau:
Does #11273 fit your needs ?

williamjameshandley:
Yes! That seems to fix all my use cases just by passing `--no-autoindent`

Carreau:
Closed by #11273

eromoe:
How do I config this in ipython config file ??

PR #11273, opened 2018-08-29
Attempt to reactivate autoindent

Carreau:
See #11257 

Will complete the pull-request and undeprecate what needs to be undeprecated if actually fixes use case for #11257.


Carreau:
I don't like this, but I"m going to self-merge. I'm going to try to use the bot to do that out of shame.

PR #11278, opened 2018-09-01
Remove Python 2 shim.

Carreau:
There was some recent bugs where doctests were not ran, so slowly
clearing and checking other doctest for later refactor


Carreau:
Thanks !

Issue #11281, opened 2018-09-02
ipython autoreload does not work with class new methods


oscar6echo:
As described in this googlegroup jupyter thread [Hybrid dev Jupyter + VS Code & IPython magic autoreload](https://groups.google.com/forum/#!msg/jupyter/heWQ2ce04MI/xsZYe5GvAgAJ;context-place=forum/jupyter), and then later discussed with @Carreau at JCon, this [ipython-autoreload-explo](https://github.com/oscar6echo/ipython-autoreload-explo) repo gives an example, and maybe a fix ?

If you run the [test-magic-autoreload notebook](https://github.com/oscar6echo/ipython-autoreload-explo/blob/master/test-magic-autoreload.ipynb) with the standard `%autoreload 2` the [cube](https://github.com/oscar6echo/ipython-autoreload-explo/blob/9ce4faf9a1d132de281e4a98a71c23d65a797411/mypackage/myfile.py#L15) and [power](https://github.com/oscar6echo/ipython-autoreload-explo/blob/9ce4faf9a1d132de281e4a98a71c23d65a797411/mypackage/myfile.py#L19) function of [MyClass](https://github.com/oscar6echo/ipython-autoreload-explo/blob/9ce4faf9a1d132de281e4a98a71c23d65a797411/mypackage/myfile.py#L4), when uncommented still cannot be accessed.

If you run the same [test-magic-autoreload notebook](https://github.com/oscar6echo/ipython-autoreload-explo/blob/master/test-magic-autoreload.ipynb)  with `%myautoreload 2` then they are available.

The myautoreload magic is essentially the standard [autoreload](https://github.com/ipython/ipython/blob/master/IPython/extensions/autoreload.py), with a short [fix](https://github.com/oscar6echo/ipython-autoreload-explo/blob/9ce4faf9a1d132de281e4a98a71c23d65a797411/magics/myautoreload.py#L298).

Would you accept a PR ?

To give some context, the autoreload magic is particularly useful when a notebook producer prototypes in the notebook, because it's convenient, and then gradually moves the code to a properly organised package edited in a full fledged IDE. This package typically contains classes that start lean and grow.

Carreau:
Apologies, I'm travelling and brain not functionning well while awake yet.
If it fix a usecase I'd say send a PR i'll try to get it in for 7.0 thtat I want to do (at least a beta) for end of semptember.

oscar6echo:
Done.  
Please see PR #11283.

[...]

PR #11283, opened 2018-09-05
Add new methods in update_class()

oscar6echo:
Reference #11281.

Currently the autoreload magic only upgrades existing methods but ignores new ones.  
This PR enables to captures new methods - very useful in incremental dev.

### DIFF
--- a/IPython/testing/iptest.py
+++ b/IPython/testing/iptest.py
@@ -386,6 +386,12 @@ def run_iptest():
         monkeypatch_xunit()
 
     arg1 = sys.argv[1]
+    if arg1.startswith('IPython/'):
+        if arg1.endswith('.py'):
+            arg1 = arg1[:-3]
+        sys.argv[1] = arg1.replace('/', '.')
+    
+    arg1 = sys.argv[1]
     if arg1 in test_sections:
         section = test_sections[arg1]
         sys.argv[1:2] = section.includes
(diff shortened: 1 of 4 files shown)


Carreau:
let's see if we can find how to fold your test case into the test suite.

Carreau:
I've tried adding some tests. 
Let me know if that's what you were looking for.

I haven't asserted the values of c.a c.b c.toto and I am jetlagged, so may have done stupid things.

IPython/testing/iptest.py
```diff
@@ -385,6 +385,12 @@ def run_iptest():
     if '--with-xunit' in sys.argv and not hasattr(Xunit, 'orig_addError'):
         monkeypatch_xunit()
 
+    arg1 = sys.argv[1]
+    if arg1.startswith('IPython/'):
+        if arg1.endswith('.py'):
+            arg1 = arg1[:-3]
+        sys.argv[1] = arg1.replace('/', '.')
```
Carreau: That allows to run iptest with a filepath which is a little more confenient for bash autocompletion.

oscar6echo:
The tests look fine to me. Thank you, I would not have been able to write them.  
If you think the whatsnew/pr/improve-autoreload.md is too verbose, feel free to cull.

Carreau:
> The tests look fine to me. Thank you, I would not have been able to write them.

It's ok they were not easy to write, testing this in one of the tricky bits. 


> If you think the whatsnew/pr/improve-autoreload.md is too verbose, feel free to cull.

all the `watsnew/pr/*` will be merged in the changelog at release time, and well be edited then.

PR #11287, opened 2018-09-06
Switch scripts cell magic default to raise if error happens.

### LINKED ISSUE #11198: Script (e.g. Bash) magics doesn't have an error on status code

Carreau:
Close #11198

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -55,7 +55,7 @@ def script_args(f):
             """
         ),
         magic_arguments.argument(
-            '--raise-error', action="store_true",
+            '--no-raise-error', action="store_false", dest='raise_error',
             help="""Whether you should raise an error message in addition to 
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks !

Issue #11180, opened 2018-06-09
ipython matplot hangs ubuntu


user3477071:
I am using Python 3.5.2, IPython 6.4.0 and matplot 2.2.2 on Ubuntu Xenial. When I try to run a simple plot using pylab, the desktop hangs with

$ uname -a
Linux voyager 4.4.0-127-generic #153-Ubuntu SMP Sat May 19 10:58:46 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
$ ipython
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.
%autoindent is deprecated since IPython 5: you can now paste multiple lines without turning autoindentation off.
Automatic indentation is: OFF

In [1]: pylab
Using matplotlib backend: TkAgg
Populating the interactive namespace from numpy and matplotlib

In [2]: x = linspace(-10., 10., 1000)

In [3]: plot(x, sin(x))
ipython process using 100% of cpu. No plot comes up. I have to kill the ipython process.

Carreau:
try issuing a `In[1]: %gui tk` first to enable the tk eventloop integration.

user3477071:
Enabling 'tk eventloop' at beginning did not help. Still have the same issue
```
$ ipython
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.
%autoindent is deprecated since IPython 5: you can now paste multiple lines without turning autoindentation off.
Automatic indentation is: OFF

In [1]: %gui tk

In [2]: %pylab
Using matplotlib backend: TkAgg
Populating the interactive namespace from numpy and matplotlib

In [3]: x = linspace(-10., 10., 1000)

In [4]: plot(x, sin(x))

```
I have a system hang at this point.

anntzer:
I can reproduce the same issue with the qt5agg backend (Arch Linux, "empty" ipython profile, `%matplotlib qt5` then `plt.plot()` (with matplotlib properly configured to use qt5agg)).  This bisects back to 8e256bd (upgrade to PTK2).

Carreau:
Can you retry on master ? seem like #11288 should have fixed that.

anntzer:
works for me

minrk:
[...]

PR #11288, opened 2018-09-07
Pass inputhook to prompt_toolkit

tkf:
From simple grep searches, it look like `TerminalInteractiveShell.inputhook` used only for `TerminalPdb`.  Don't you need to pass it to `PromptSession` when creating `TerminalInteractiveShell.pt_app`?  This fixes `%matplotlib tk` for me.

ref #11180


Carreau:
Seem reasonable to me.
Thanks !

tacaswell:
Thanks @tkf !  This had been on my list of things to track down.

tkf:
You are welcome! It's just that IPython 7 is so great and I can't wait the GUI support to be back any longer :)

PR #11289, opened 2018-09-07
Make run_cell_async a real coroutine

minrk:
follow-up to #11265

- removes `yield interactivity` that required run_cell_async to be called in a special way with `gen.send(None)`
- adds `should_run_async` public method

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2830,32 +2830,52 @@ class InteractiveShell(SingletonConfigurable):
             shell_futures=shell_futures,
         )
 
-        # run_cell_async is async, but may not actually need and eventloop.
+        # run_cell_async is async, but may not actually need an eventloop.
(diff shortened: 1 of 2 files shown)


Carreau:
Yep, seem like a better solution than current yield thing.

I guess there is likely some extra refactor we can do as we seem to apply the transformers and check for async both in run_cell_async and should_run_async, which tend to show there is likely a better abstraction, but I'm happy to move forward with this.

Carreau (approved):

minrk:
yeah, we are running the transforms twice here in order to do the check.

I think the long term fix is to get to the point where it's always async and we don't need to check, so I'm happy with this as the current state.

Carreau:
> I think the long term fix is to get to the point where it's always async and we don't need to check, so I'm happy with this as the current state.

Ok, let's get that in. Hopefully at some point we'll have async-exec in core Python :-)

PR #11291, opened 2018-09-07
catch BaseException in coroutine runner

minrk:
so we catch KeyboardInterrupt and the like, not just Exceptions


Carreau (approved):
+1 good catch!

PR #11299, opened 2018-09-11
cleanup build process

Carreau:
Some code in the release process seem to be artifacts of old time.

 - `execfile` does not exist in Python3, and was used in `setup.py`
 - `development` in index.rst is not taking care of automatically with a `.. only:` directive.

### DIFF
--- a/tools/build_release
+++ b/tools/build_release
@@ -4,20 +4,13 @@
 import os
 from shutil import rmtree
 
-from toollib import sh, pjoin, get_ipdir, cd, execfile, sdists, buildwheels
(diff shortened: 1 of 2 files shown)


minrk:
Not milestone 7.0?

Carreau:
> Not milestone 7.0?

If @ivanov is comfortable with it, sure. But the less modif between beta/rc/final the better no ?

ivanov (approved):
I'm fine with targeting this to 7.0 - it's just release related machinery, not user level code cleanup.

PR #11301, opened 2018-09-11
Remove nonexistent include from MANIFEST

Carreau:


minrk:
not 7.0?

Carreau:
> not 7.0?

Sure, we can, but not critical.

PR #11308, opened 2018-09-12
correct installation instructions for ipython 7.0.0b1

yarko:


yarko:
That is, change the 7.0.0.b1 installation instruction to:
```
pip install ipython --upgrade --pre
```

ivanov:
Thank you for your PR, @yarko - and for communicating on the twitter thread for how you were able to try out the pre-release beta1. However, the documentation here will be used for the actual 7.0.0 release, so we do no to want to put in the `--pre` flag into these docs. If you want to retarget this PR to add the `--pre` flag for beta and RC releases to the ["Draft a short release announcement" section of https://github.com/ipython/ipython/blob/7.0.0b1/docs/source/coredev/index.rst#8-draft-a-short-release-announcement](https://github.com/ipython/ipython/blob/7.0.0b1/docs/source/coredev/index.rst#8-draft-a-short-release-announcement) that will help us not lose track of this detail in the future. 

How does that sound?

ivanov (changes requested):
It would be good to document the `--upgrade --pre` flags for the announcement section of making RC and Beta releases in `docs/source/coredev/index.rst`, but this section is used for the full official release so should be left as just `pip install ipython --upgrade`

Carreau:
One of the things we can do is put the "how to upgrade betas" in a `.. only:: ipydev` directive, so it will only display while building the "latest" documentation. 

Best of both world ?

yarko:
I like @Carreau's suggestion which now brings a stylistic question.

I'm seeing other places in the docs where `.. only::` is used.

Would you prefer an _additional_ note with an `.. only:: ipydev` directive,
```
.. only:: ipydev

For beta (pre-release) installations, use:
  .. code::

     pip install ipython --upgrade --pre
```
or an alternating pair of:
```
.. only:: ipydev

  .. code::

     pip install ipython --upgrade --pre

.. only:: not ipydev

  .. code::

     pip install ipython --upgrade
```

ivanov:
I like that first version, where you're adding additional context, instead of using `ipydev` to switch between two alternatives.

yarko:
I think this should be good to go now.   Let me know what you think.

Carreau:
Thanks ! Looks great !

PR #11313, opened 2018-09-14
Fix magic directive and role.

Carreau:
We were directly modifying state instead of using blessed API (also some
api will change so be future proof with sphinx master).

This should restore some ability of the :magic: and :cellmagic:
directive to both support link to magics name when the text in the
directive contains (or not) leading % and %%, it will also automatically
prepend the right number of % for magics and cell magics.

A couple of other related fixes.

### DIFF
--- a/docs/source/interactive/autoawait.rst
+++ b/docs/source/interactive/autoawait.rst
@@ -58,7 +58,7 @@ Should behave as expected in the IPython REPL::
 
 You can use the ``c.InteractiveShell.autoawait`` configuration option and set it
 to :any:`False` to deactivate automatic wrapping of asynchronous code. You can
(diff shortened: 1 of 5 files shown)


docs/source/interactive/autoawait.rst
```diff
@@ -127,7 +127,7 @@ Effects on IPython.embed()
 IPython core being asynchronous, the use of ``IPython.embed()`` will now require
 a loop to run. By default IPython will use a fake coroutine runner which should
 allow ``IPython.embed()`` to be nested. Though this will prevent usage of the
-``autoawait`` feature when using IPython embed. 
+:magic:`autoawait` feature when using IPython embed.
```
minrk: This one has no leading `%` but others do. Presumably only one of those is right, so they should be made consistent.

Otherwise 👍

PR #11317, opened 2018-09-15
Remove implicit dependency to ipython_genutils.

Carreau:
This was installed because we rely on traitlets.

`indent` behavior is _slightly_ different in the sens that white lines
may not have the same number of whitespace.

### DIFF
--- a/docs/autogen_config.py
+++ b/docs/autogen_config.py
@@ -11,7 +11,36 @@ here = abspath(dirname(__file__))
 options = join(here, 'source', 'config', 'options')
 generated = join(options, 'config-generated.txt')
 
-from ipython_genutils.text import indent, dedent
+import textwrap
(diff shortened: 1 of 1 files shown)


docs/autogen_config.py
```diff
@@ -11,7 +11,36 @@
 options = join(here, 'source', 'config', 'options')
 generated = join(options, 'config-generated.txt')
 
-from ipython_genutils.text import indent, dedent
+import textwrap
+indent = lambda text,n: textwrap.indent(text,n*' ')
+
+def dedent(text):
```
anntzer: Is this not just inspect.cleandoc?

Carreau: looks like it, do you want to make a PR?

anntzer: done in #11461

PR #11318, opened 2018-09-15
Allow completing dict keys only

blink1073:
I have a custom python kernel where I'd like to only show dictionary keys if there are any available.  It is confusing to our users to show anything but dictionary keys in this case:

image

This patch allows me to override `do_complete`, check for dictionary key matches, use those if they are available, and otherwise use all of the matches.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -350,7 +350,7 @@ class Completion:
     Completion object used and return by IPython completers.
 
     .. warning:: Unstable
-        
+
         This function is unstable, API may change without warning.
         It will also raise unless use in proper context manager.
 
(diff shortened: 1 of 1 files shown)


blink1073:
Prior to https://github.com/ipython/ipython/pull/10969/files, I was able to set `matchers` directly to accomplish this.

blink1073:
In the classic notebook, we don't even see our dict keys in the default view:

image

blink1073:
I my ideal world this would be the default behavior :smile:.

dsblank:
Yes, please! One should only be able to select keys in a key context, only files in a string context, and never files when not in a string context. So confusing for everyone. I hope that this can be refined.

blink1073:
Happy to take a crack at implementation if there's consensus.

blink1073:
We'd also want files in a magic since we use them in shell magics for example.

Carreau:
Thanks, looks simple enough.

astrofrog:
@blink1073 - just to check, what's the recommended way to use this? As a developer of a package, is there a way to enable ``dict_keys_only=True`` for specific classes by default, or is this just settable on a user-level in IPython?

Carreau:
> just to check, what's the recommended way to use this? As a developer of a package, is there a way to enable `dict_keys_only=True` for specific classes by default, or is this just settable on a user-level in IPython?

This is only global at IPython level . I am way overdue to write a whitepaper and request funding to push jedi forward, and make IPython tab completion on par with what other IDE have. I  can involve you if you want to help with that.

Issue #11320, opened 2018-09-18
ipython directive seems to be broken


tacaswell:
Found this while trying to build the Matplotlib docs.  Can try to chase it down better tomorrow, but posting now in case I do not get back to it.

```
Exception occurred:
  File "/home/tcaswell/.virtualenvs/bleeding/lib/python3.8/site-packages/IPython/sphinxext/ipython_directive.py", line 333, in process_input_line
    if self.IP.check_complete()[0] != 'incomplete':
TypeError: check_complete() missing 1 required positional argument: 'code'
The full traceback has been saved in /tmp/sphinx-err-ruydljl4.log, if you want to report the issue to the developers.

```

23:56 $ ipython --version
7.0.0.dev

<details>

```
23:51 $ cat /tmp/sphinx-err-ruydljl4.log
# Sphinx version: 1.8.0
# Python version: 3.8.0a0 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.10
# Last messages:
#   reading sources... [ 97%] tutorials/advanced/path_tutorial
#   
#   reading sources... [ 97%] tutorials/advanced/patheffects_guide
#   
#   reading sources... [ 97%] tutorials/advanced/transforms_tutorial
#   
#   reading sources... [ 97%] tutorials/colors/colorbar_only
#   
#   reading sources... [ 97%] tutorials/colors/colormapnorms
#   
# Loaded extensions:
#   sphinx.ext.mathjax (1.8.0) from /home/tcaswell/.virtualenvs/bleeding/lib/python3.8/site-packages/sphinx/ext/mathjax.py
#   alabaster (0.7.11) from /home/tcaswell/.virtualenvs/bleeding/lib/python3.8/site-packages/alabaster/__init__.py
#   sphinx.ext.autodoc (1.8.0) from /home/tcaswell/.virtualenvs/bleeding/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py
#   sphinx.ext.autosummary (1.8.0) from /home/tcaswell/.virtualenvs/bleeding/lib/python3.8/site-packages/sphinx/ext/autosummary/__init__.py
#   sphinx.ext.doctest (1.8.0) from /home/tcaswell/.virtualenvs/bleeding/lib/python3.8/site-packages/sphinx/ext/doctest.py
[...]

PR #11321, opened 2018-09-18
Fix the sphinx_ipython directive.

Carreau:
Also use it in our docs so that there is some kind of tests.
Imperfect so far.

Fixes gh-11320

### DIFF
--- a/docs/source/whatsnew/pr/incompat-switching-to-perl.rst
+++ b/docs/source/whatsnew/pr/incompat-switching-to-perl.rst
@@ -1,1 +1,7 @@
+Incompatible change switch to perl
+----------------------------------
+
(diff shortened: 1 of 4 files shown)


docs/source/whatsnew/pr/incompat-switching-to-perl.rst
```diff
@@ -1 +1,7 @@
+Incompatible change switch to perl
```
Carreau: I know it's unrelated, but I was annoyed to have to fix this on my local machine, with I'm running sphinx with warnings as errors.

Carreau:
I'm not going to change this for 7.0, but I think ":okexcept:" should raise if there is no exceptions.

PR #11331, opened 2018-09-25
re-add the rprint and rprinte alias.

Carreau:
They are used in IPykernel 4.9 and I can see users upgrading IPython w/o
upgrading Ipykernel.


minrk:
Agreed. I don't think there's a reason to ever remove these. The only reason to remove deprecated APIs is if their presence is actively causing a problem.

Issue #11336, opened 2018-09-27
Weird character after first import


albireox:
I'm running python 7.0.1 Everything seems to work fine but after the first import I consistently get a weird symbol (see attached screenshot). When I press enter again the symbol goes away and doesn't reappear. I'm using fish shell on an iterm2 terminal on mac.

![screenshot 2018-09-27 at 13 51 00](https://user-images.githubusercontent.com/568775/46174104-fb146e00-c25c-11e8-9c46-2c99b7e578ad.png)

[Update]

Upgrade prompt_toolkit to 2.0.6 should fix the issue.

Carreau:
Hum I've seen this one but with `^[[43;1R`, but I assumed it was due to me being on master of my terminal emulator. Not sure where this come from

davidbrochart:
I've seen similar things in latest https://github.com/jonathanslenders/pymux (not released, uses prompt-toolkit 2). Maybe @jonathanslenders has some idea?

matthew-brett:
I get the same:

```
$ ipython
Python 3.6.5 (default, Mar 30 2018, 06:41:53) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.0.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os                                                                                             

^[[19;1RIn [2]:
```

If I then try to tab complete:

```
In [2]: os.p                                                                                                  
os.pardir         os.pathconf       os.pathsep        os.popen          os.putenv         
os.path           os.pathconf_names os.pipe           os.pread          os.pwrite         
^[[19;1RIn [2]: os.p
```

This makes tab completion almost unusable.

jonathanslenders:
Is it possible that this was fixed by the following commit: https://github.com/jonathanslenders/python-prompt-toolkit/commit/e226d640177aa1d2cf293e4de382f171586173a2 ?
It's merged in prompt_toolkit master, but I'm about to release a new prompt_toolkit 2.0 which includes it.

jonathanslenders:
No, I'm pretty sure this isn't IPython. I'll try to reproduce with iterm2

matthew-brett:
I'm sure I'm not understanding the complexities here, but just for information, this does occur with IPython 7.0 and 7.1, and doesn't occur with 6.0 or 6.5, testing in the same virtualenv.

Carreau:
Can you all post which terminal you've seen it in? 
[...]

PR #11347, opened 2018-09-30
[experiment] Async shell (run everything under asyncio)

Carreau:
prompted by #11336, in which I realized that `patch_stdout` was mostly meant for async based application. 

This run the all IPython shell into an eventloop, which has the side-effect of making async bg task work.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2842,7 +2842,6 @@ class InteractiveShell(SingletonConfigurable):
         try:
             return await coro
         except BaseException as e:
-            print('IN BE')
             info = ExecutionInfo(raw_cell, store_history, silent, shell_futures)
             result = ExecutionResult(info)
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Keep in mind that this could break the input hooks for UI toolkits which IPython has right now. (Did you test that?)

If so, I think a solution would be to use the asyncio even loop with a custom "selector". The selector is the blocking call that waits for any kind of IO. Within the selector you can run another event loop, and glue them together.

In any case, I think I like the direction where this is going!

Carreau:
> Keep in mind that this could break the input hooks for UI toolkits which IPython has right now. (Did you test that?)

No I did not test, and this likely also bypass the pre and post execute hooks.

This was mostly an experiment to see how much change was needed, and think about what internal API to deprecate. As we just released 7.0 I doubt I want to merge this in the 7.x branch.

PR #11350, opened 2018-10-01
Fix #9343: warn when using HTML instead of IFrame

### LINKED ISSUE #9343: HTML vs Iframe
I was again the witness of people miss-using HTML to display IFrames, and having issues. 

Could we detect in  HTML the usage of `<iframe>....</iframe>` and warn user that there is an iframe object ?

mpenkov:


IPython/core/display.py
```diff
@@ -666,6 +666,11 @@ def _repr_pretty_(self, pp, cycle):
 
 class HTML(TextDisplayObject):
 
+    def __init__(self, data=None, url=None, filename=None, metadata=None):
+        if data and "<iframe " in data and "</iframe>" in data:
+            warnings.warn("Consider using IPython.display.IFrame instead")
+        super(HTML, self).__init__(data=data, url=url, filename=filename, metadata=metadata)
```
Carreau: We're py3 only,  so now you can just use `super()`

IPython/core/tests/test_display.py
```diff
@@ -195,6 +195,14 @@ def test_displayobject_repr():
     j._show_mem_addr = False
     nt.assert_equal(repr(j), '<IPython.core.display.Javascript object>')
 
+@mock.patch('warnings.warn')
+def test_encourage_iframe_over_html(m_warn):
+    display.HTML('<br />')
+    m_warn.assert_not_called()
```
Carreau: Ah I would have use [`assert_warns`](https://docs.pytest.org/en/latest/warnings.html#warns), but that looks good !

IPython/core/display.py
```diff
@@ -666,6 +666,11 @@ def _repr_pretty_(self, pp, cycle):
 
 class HTML(TextDisplayObject):
 
+    def __init__(self, data=None, url=None, filename=None, metadata=None):
+        if data and "<iframe " in data and "</iframe>" in data:
```
Carreau: Wondering if we should use `startswith` and `endswith` that are more efficient and likely more correct, but that looks good.

Carreau:
Thanks !

IPython/core/display.py
mpenkov: Yeah, I had the same thought. There’s a difference between rendering a single iframe vs rendering HTML with an iframe in it. Startswith would be a better idea. We could also account for case differences because HTML is case insensitive.

Let me know if you’d like those changes, or if things are fine as they are.

IPython/core/tests/test_display.py
mpenkov: Didn’t know about that one, thanks!

IPython/core/display.py
Carreau: Up to you. I'll reopen the linked issue add link to this section for further improvements.

IPython/core/tests/test_display.py
mpenkov: I had a look at the link that you posted, it seems specific to pytest, which IPython doesn't use. Did you mean to send [this link](https://docs.python.org/3/library/warnings.html#testing-warnings)? Let me know if I misunderstood something.

Carreau: Ah, yes, sorry, and [this](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns) I just get confused between test runners.

ssxcho:
Hi everyone, I am having trouble on this , I need to use '<iframe srcdoc =' , any suggestions on how to bypass this warning ?

Issue #11328, opened 2018-09-23
Add height and width parameters to core.display.Video


jwkvam:
I'm viewing some videos in a notebook. The videos are larger than the width of my notebook so I can't see the entire frame unless I scroll left and right. I would like to specify a height and width so I can view the entire frame at once. I think it'd be pretty cool if it could resize automatically, but I'd be happy with the ability to set the height and width.

I can post some pictures if it helps.
```
ipython --version
6.5.0
jupyter lab --version
0.34.12
```

I've been doing this work around:
```
HTML(
    '<video controls width=500>'
    '<source src="video.mp4" type="video/mp4"/>'
    '</video>'
)
```

It seems like a simple PR, I could make it if you all like this.

Carreau:
Sure feel free to send a PR if you looked at the code and it seem easy enough. 
Feel free to ask if you want some pointers. 

You can also wait a few days if you are participating to [hacktoberfest](http://hacktoberfest.digitalocean.com/).

Carreau:
Hi @IpshitaC 

The source code for the current Video object can be found here:

https://github.com/ipython/ipython/blob/94b894de4cef5b0e3cfb666d5328749205980d37/IPython/core/display.py#L1257-L1317

You will likely need to search how to update the `_repr_html_` of it, or it' superclass to  have a `width=` parameter

Carreau:
see #11353, I leave that open, as some improvements are possible,  mainly if users are passing string like `10%`, `12em` it should still be valid sizes.

PR #11353, opened 2018-10-01
Add support for width and height arguments when displaying Video

finaiized:
This PR adds `width` and `height` attributes to `IPython.display.Video`. The current behaviour is preserved if neither width nor height are specified. Otherwise, the specified integer will be used to set the width/height of the video in pixels, much like `IPython.display.Image`. This addresses the feature request asked for in #11328.

Example usage:

```
Video('https://archive.org/download/Sita_Sings_the_Blues/Sita_Sings_the_Blues_small.mp4', width=600, height=200)
```

Here's a notebook that exercises these attributes:
video-width-height-test

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1256,7 +1256,8 @@ class Image(DisplayObject):
 
 class Video(DisplayObject):
 
-    def __init__(self, data=None, url=None, filename=None, embed=False, mimetype=None):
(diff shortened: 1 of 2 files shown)


Carreau:
Looks good  !

As an improvement, you could try to see how to let users pass strings like `40%`, `12em`, but that's a good first step !

PR #11354, opened 2018-10-01
Fix #11334: Cannot make multi-line code blocks in ipython

### LINKED ISSUE #11334: Cannot make multi-line code blocks in ipython
I just did a `pip install ipython` while teaching a class today and typed an if statement and then hit enter after the first line and the code executed.

This seems like a bug.

I did a `pip install ipython==6.5.0` and hitting Enter in a code block properly showed me the next indented line to type out.

I'm on Ubuntu 18.04 running ipython within tmux, though I doubt tmux is the issue here.

bxsx:
When codeop.compile_command() returns None it actually says "at least
some part of the code was compiled successfully" which is not really important
for checking if it's complete or not.
Once we haven't got any errors during compilation process, we just want
to check if there will be another nested block of code or not by checking
a colon.

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -636,7 +636,7 @@ class TransformerManager:
                 MemoryError, SyntaxWarning):
             return 'invalid', None
         else:
-            if res is None:
+            if not lines[-1].strip().endswith(':'):
                 return 'incomplete', find_last_indent(lines)
         return 'complete', None
 
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -631,7 +631,7 @@ class TransformerManager:
         try:
             with warnings.catch_warnings():
                 warnings.simplefilter('error', SyntaxWarning)
-                res = compile_command(''.join(lines), symbol='exec')
+                compile_command(''.join(lines), symbol='exec')
         except (SyntaxError, OverflowError, ValueError, TypeError,
                 MemoryError, SyntaxWarning):
             return 'invalid', None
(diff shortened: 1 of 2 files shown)


bartskowron:
Yeah, I have already fixed it. Trying to run a test suit locally but I get a lot of errors and unfortunately can't understand why (is there any dev-guide how to develop ipython locally?).

Anyway, push the fix, to see what CI will say...

bartskowron:
> That seem incorrect, now `a = 1` is considered as incomplete, while it's not.

Seems it's OK now

Carreau:
There seem to be some more edge cases:

```
if True:
    1
<no spaces>
``` 
Appear as incomplete and adds a new line with 4 spaces. 

```
if True:
    1
	<4 spaces>
``` 

Appear as complete an executes.

bartskowron:
I am very busy today but this looks like an another bug to me that we haven't noticed before because of this bug.

If there is at least one space in the last line (blank) then the line is complete.

```
if True:
    1
<at least 1 space>
```

this seem to work correctly.

The bug is in a code before ```check_complete()``` is being called as it's already missed in the ```cell``` argument passed to the function.

I can take a look in it closer but not before weekend.

bartskowron:
```InteractiveShell.check_complete()``` invokes ```TransferManager.check_complete()``` and seems the passing ```code``` argument doesn't have a trailing '\n' in any situation. I believe this is for some reasons and is correct, @Carreau ?

If so, this helped me to focus back to the ```TransferManager.check_complete()``` and seems like removing the condition fixes the issue but provides a regression in another test. This actually surprised me honestly.

Going off for a day or two, leaving the comment for notes, but feel free to share some information if you have or finish the fix obviously ;-)

PS. @Carreau is there any guide how to develop and test locally IPython? When I want to use IPython installed globally (to meet all requirements etc.) I have no access to the IPython-dev directory as all imports are already done for starting IPython itself. I can figure it out how to do this but maybe there is a note about it somewhere?

PR #11355, opened 2018-10-02
Fix #11328 add height and width parameters to core.display.Video

### LINKED ISSUE #11328: Add height and width parameters to core.display.Video

nestoralonso:
Passes width and height attributes to the video tag

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -238,15 +238,21 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     want to use. Here is a list of the names of the special methods and the
     values they must return:
 
-      - `_repr_html_`: return raw HTML as a string
-      - `_repr_json_`: return a JSONable dict
-      - `_repr_jpeg_`: return raw JPEG data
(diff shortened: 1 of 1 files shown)


Carreau:
Apologies, I just merged #11353 that fixes the same issue. There are still improvements possible, like letting users pass string like '12%', '11pt'. Do you want to attempt this ?

nestoralonso:
Yeah, if I have some spare time I will try it

PR #11357, opened 2018-10-02
Math display: change $$...$$ -> $\displaystyle ...$

mgeier:
This makes math expressions align left when converted to LaTeX with e.g. `nbconvert`.

Also, it should avoid problems like https://github.com/jupyter/nbconvert/issues/275.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -693,7 +693,7 @@ class Markdown(TextDisplayObject):
 class Math(TextDisplayObject):
 
     def _repr_latex_(self):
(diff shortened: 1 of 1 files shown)


asmeurer:
See the discussion at https://github.com/jupyter/notebook/issues/4060

Carreau:
I do not have any objection. If there is no objection from the nbconvert/notebook/lab side of things i"m happy to get that in.

blink1073 (approved):
LGTM, thanks!

PR #11365, opened 2018-10-06
os.scandir

### LINKED ISSUE #9914: investigate os.scandir
Starting with Python 3.5 os.scandir is available and is faster than listdir. It is also availlable on PyPI there : https://pypi.python.org/pypi/scandir

Looking as the various usage of listdir and os.walk on our codebase and seeing whether we can replace their usage by scandir/scandir walk should be a relatively simple task that might lead to some performance improvements and maybe even code simplification. 

We should also investigate this on jupyter/notebook and jupyter/nbconvert likely (cc @michaelpacer)

kd2718:
FIX #9914 
I added os.scandir in two locations. I also updated how osm.py checked for executable `isexec`. I took advantage of some type hinting. I don't think type hinting is used elsewhere in the code, so I can remove this if needed.

### DIFF
--- a/IPython/core/magics/osm.py
+++ b/IPython/core/magics/osm.py
@@ -218,8 +218,9 @@ class OSMagics(Magics):
                         os.chdir(pdir)
                     except OSError:
                         continue
+
+                    # use with notation for python 3.6 onward
                     dirlist = os.scandir(path=pdir)
-                    #with os.scandir(pdir) as dirlist:
                     for ff in dirlist:
                         if self.isexec(ff):
                             fname = ff.name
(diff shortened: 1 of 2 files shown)


IPython/core/magics/osm.py
```diff
@@ -31,6 +31,51 @@ class OSMagics(Magics):
...
+        else:
+            try:
+                winext = os.environ['pathext'].replace(';','|').replace('.','')
+            except KeyError:
+                winext = 'exe|com|bat|py'
+            
+            self.execre = re.compile(r'(.*)\.(%s)$' % winext,re.IGNORECASE)
+
+        # call up the chain
+        super(OSMagics, self).__init__(shell=shell, **kwargs)
```
minrk: This can be just `super()` since IPython requires Python 3

IPython/core/magics/osm.py
```diff
@@ -481,6 +520,7 @@ def dhist(self, parameter_s=''):
 
         dh = self.shell.user_ns['_dh']
         if parameter_s:
+            args = []
```
minrk: This empty `args` is never used

kd2718: Yes, it's overwritten on line [525](https://github.com/ipython/ipython/pull/11365/files#diff-e72931d7a47ab4a04ef88193e513a39fR525) and the `return` on line [528](https://github.com/ipython/ipython/pull/11365/files#diff-e72931d7a47ab4a04ef88193e513a39fR528) prevents a case where `args` is not in scope. 

However, the linter wasn't fully aware of this and got worried that `args` may not be implemented at the time of use. Looking forward, if the except is changed to handle a specific exception, rather than a generic exception, then this will prevent `args` from being out of scope.

This may be over protective, so I can still remove the line in question if preferred.

IPython/core/magics/osm.py
kd2718: thanks. I will update this

IPython/core/magics/osm.py
```diff
@@ -160,65 +205,59 @@ def rehashx(self, parameter_s=''):
...
-            isexec = lambda fname:os.path.isfile(fname) and execre.match(fname)
         savedir = os.getcwd()
 
         # Now walk the paths looking for executables to alias.
         try:
             # write the whole loop for posix/Windows so we don't have an if in
             # the innermost part
-            if os.name == 'posix':
+            if self.is_posix:
+                print(path)
```
Carreau: forgotten debug print ?

IPython/core/magics/osm.py
```diff
@@ -160,65 +205,59 @@ def rehashx(self, parameter_s=''):
...
             else:
                 no_alias = Alias.blacklist
                 for pdir in path:
                     try:
                         os.chdir(pdir)
-                        dirlist = os.listdir(pdir)
                     except OSError:
                         continue
+
+                    # use with notation for python 3.6 onward
```
Carreau: I guess you mean types annotations that you removed , I thing that can be trimmed now.

Carreau:
Not all your machines are setup with the same email, and/or github does not have all your emails so some commits do not appear as yours. 

I believe 3.5 have function signature types, so these would have been ok to leave in I think. 

There is 2 small leftover 1 debug statement and 1 comment. And we're good to merge IMHO.

Carreau:
@meeseeksdev tag hacktoberfest

IPython/core/magics/osm.py
kd2718: yup, thanks, removed

Carreau:
Thanks.

PR #11367, opened 2018-10-07
Improve IFrame warnings in IPython/core/display.py

### LINKED ISSUE #9343: HTML vs Iframe

mpenkov:
- Continuation of #11350 
- Fix #9343


Carreau:
LGTM.

Good work.

Issue #11375, opened 2018-10-08
trade deprecated time.clock for time.perf_counter


minrk:
We still have some calls to `time.clock`, which is deprecated. We should switch all such calls to `time.perf_counter`, which is available in all supported Python versions.

victor-eds:
I'll take a look

ammarmallik:
Done with it. Please review this: [Pull Request](https://github.com/ipython/ipython/pull/11376)

Carreau:
closed in #11376

Carreau:
Thanks, nice work.

PR #11376, opened 2018-10-08
Replace Deprecated time.clock with time.perf_counter

ammarmallik:
Ref: [Issue#11375](https://github.com/ipython/ipython/issues/11375)
This pull request contains following:

- Replaced deprecated time.clock with time.perf_counter
- Updated comments around it

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -912,7 +912,7 @@ python-profiler package from non-free.""")
             Number of times to execute `run`.
 
         """
-        twall0 = time.time()
+        twall0 = time.perf_counter()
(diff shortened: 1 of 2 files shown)


minrk:
Thanks! There is also some use in `IPython/core/magics/execution.py` in the timeit magic. I think the calls there to `time.time()` should also be replaced with `time.perf_counter()`.

ammarmallik:
> Thanks! There is also some use in `IPython/core/magics/execution.py` in the timeit magic. I think the calls there to `time.time()` should also be replaced with `time.perf_counter()`.

Done.

Carreau (approved):

Issue #10079, opened 2016-12-06
%run with single-quoted path doesn't work in IPython 5.1


scottcode:
Using `%run '<script_path>'` to run another script (single-quoted) fails in IPython 5.1, whereas it used to work in IPython 4.2. Using double-quotes still works, as well as not using quotes at all. 

# My setup

* Windows 7
* IPython 5.1
* Python 2.7

Output of `python -c "import IPython; print(IPython.sys_info())"`:

```
{'commit_hash': u'5c9c918',
 'commit_source': 'installation',
 'default_encoding': 'cp437',
 'ipython_path': 'C:\\Users\\myusername\\AppData\\Local\\Continuum\\Anaconda2\\lib\\site-packages\\IPython',
 'ipython_version': '5.1.0',
 'os_name': 'nt',
 'platform': 'Windows-7-6.1.7601-SP1',
 'sys_executable': 'C:\\Users\\myusername\\AppData\\Local\\Continuum\\Anaconda2\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '2.7.12 |Anaconda 4.1.1 (64-bit)| (default, Jun 29 2016, 11:07:13) [MSC v.1500 64 bit (AMD64)]'}
```

# Test case

1. Copy this snippet into a file called "dummy.py":
```
x = range(100)
```

2. Open IPython console and type:
```
%run 'dummy.py'
```
3. I get ``ERROR:root:File `u"'dummy.py'.py"` not found.``
4. Try `%run "dummy.py"` or `%run dummy.py` instead --> no error.
5. The `x` variable from "dummy.py" can now be used within the IPython shell (if double- or no-quotes was used)

```
In [3]:  print x[:5]
Out[3]: [0, 1, 2, 3, 4]
```

anntzer:
It's very likely.  The problem is that single quotes are valid characters in the Windows console, and the normal way to escape filenames in Windows is always to use double quotes (which are *not* valid filenames).  I don't think coming up with our own escaping mechanism is really better (e.g., backslashes used to be a problem as they are separators on Windows).

takluyver:
Does using single quotes work in cmd/Powershell? E.g. can you run `python 'foo.py'? If that doesn't work, I'd say it's OK for it not to work in IPython either.

anntzer:
Single quotes do not work in cmd.
[...]

PR #11377, opened 2018-10-09
update to %run

kd2718:
#10079
Just started work. Haven't tested to confirm I put this in the correct place


Carreau:
I think that's good, Thanks !

Issue #11329, opened 2018-09-24
Switch to vi mode in open shell


robertocarta:
Hi, I use ipython for my django shell in shared machines. With readline you can "set editing_mode=vi" or something along those lines, is there a way to do it with the current ipython with prompt_toolkit?

Carreau:
Yes, you can use `--TerminalInteractiveShell.editing_mode=vi` or set `c.TerminalInteractiveShell.editing_mode='vi'` in your configuration option. 

It is not perfect, but ressemble VI.

You can switch dynamically with something like:

```
In [1]: ip= get_ipython()
In [2]: ip.editing_mode
Out[2]: 'emacs'
In [3]: ip.editing_mode='vi'
```

There is no mode indicator, I'm sure there is an issue somewhere about that. If you figure it out feel free to send PR, or at least pointer. And/Or send updates to the docs. 

In general `ipython --help-all` and search through the output can be useful to figure out these can of things.

robertocarta:
I knew about c.TerminalInteractiveShell.editing_mode, but I needed to do it dynamically on an open ipython. I tried your method, but it doesn't seem to work. Maybe your config includes something that auto reloads settings on change?
(but, I'll definitely work on a PR, having an easy switch cold be useful to many)

robertocarta:
Ok the missing bit was ip.init_prompt_toolkit_cli()

bitan2988:
Yes, Use these following commands --TerminalInteractiveShell.editing_mode=vi or set c.TerminalInteractiveShell.editing_mode='vi' in your configuration option.



You can switch dynamically with the help of these command:

In [1]: ip= get_ipython()
In [2]: ip.editing_mode
Out[2]: 'emacs'
In [3]: ip.editing_mode='vi'

Carreau:
See #11390 which fixes the dynamic switching with `%config TerminalInteractiveShell.editing_mode='vi'`

Carreau:
closed by #11390

PR #11378, opened 2018-10-09
Fix documentation for changing configuration dynamically

2DSharp:
Documented the configuration changing process dynamically in the shell and making it live,
pointed out in https://github.com/ipython/ipython/issues/11329.

### DIFF
--- a/docs/source/config/intro.rst
+++ b/docs/source/config/intro.rst
@@ -111,6 +111,22 @@ At present, this only affects the current session - changes you make to
 config are not saved anywhere. Also, some options are only read when
 IPython starts, so they can't be changed like this.
 
+The output of `%config IPCompleter.greedy` will return true.
+
+To make the changes live using config magic dynamically in the shell,
+run::
+
+  get_ipython().init_prompt_toolkit_cli()
+
+Another way to change the configuration dynamically::
+
+    ip= get_ipython()
+    ip.editing_mode='vi'
+    
+    ip.init_prompt_toolkit_cli()
+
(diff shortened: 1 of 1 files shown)


Carreau:
Hum, the think in #11329 is a hacks, and you should'n re-call `init-ptk-app()`.

You most likely want to dynamically set `ip.pt_app.editing_mode` via a magic that does the right checks., and/or add listeners to `ip.editing_mode` that propagate the value to `ip.pt_app.editing_mode`.

2DSharp:
Sorry, but even if I set the magic config, it doesn't work unless I call `ip.init_prompt_toolkit_cli()` atleast in my version. Am I missing something?

Carreau:
I think that something like #11390 would be better suited to avoid having to poke at internals.

Then you just have to `ip.editing_mode` to switch, or use `%config TerminalInteractiveShell.editing_mode = 'vi'` which now works (and prepend `[ins]`/`[nav]`/`[...]` to the prompt. 

We just don't have that many vim user, so of course we don't fix bugs.

2DSharp:
@Carreau I don't think setting the configuration using magics is the problem. The last time I have been trying to get that working in a sprint, had me require to use `init_prompt_toolkit_cli()` to actually make the changes work.

Carreau:
Please re-read the rest of the comments and #11390.

I does work here locally if you update the right attributes of the right class instance..
 
You should not call `init_prompt_toolkit_cli()` many time, that will make internal IPython state inconsistent.

Carreau:
closing, fixed by #11390 feel free to ask if this need to be reopened.

PR #11382, opened 2018-10-09
Fix #11312: nonlocal does not raise SyntaxError

### LINKED ISSUE #11312: async-detection: nonlocal does not raise syntaxerror
See https://github.com/ipython/ipython/issues/11303#issuecomment-421297197

```
In [1]: x = 1
   ...: def f():
   ...:     nonlocal x
   ...:     x = 10000
```
Should raise but does not.

It's minor, but may be good to fix as behavior is likely undefined.

felixzhuologist:
As we traverse the AST, keep track of the depth in terms of the function call stack so that we can check for different invalid AstNodes at each depth and terminate early when possible. Fixes #11312

### DIFF
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -97,24 +97,26 @@ class _AsyncSyntaxErrorVisitor(ast.NodeVisitor):
     the implementation involves wrapping the repl in an async function, it
     is erroneously allowed (e.g. yield or return at the top level)
     """
-    def __init__(self, is_toplevel=True):
-        self.is_toplevel = is_toplevel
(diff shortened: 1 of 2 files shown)


Carreau:
running test is a bit weird as we are a python shell. 
You need to not be in IPython directory and run `iptest`

To run a single test I would for example run 
```
$ cd ; iptest IPython.core.tests.test_inputtransformer2:test_check_complete; cd -
```

You can also just push on the branch and let Travis CI work for you :-)

felixzhuologist:
Awesome, that works. Thanks!

minrk:
Thanks!

PR #11387, opened 2018-10-12
core/tests/test_magic to use %%writefile instead of %%file

hongshaoyang:
%%file is deprecated


Carreau:
Thanks ! That looks good !

PR #11390, opened 2018-10-13
Give some love to the VI mode.

Carreau:
Improve #11329, still likely need a magic to make editign mode easier to toggle

### DIFF
--- a/IPython/terminal/prompts.py
+++ b/IPython/terminal/prompts.py
@@ -14,6 +14,8 @@ class Prompts(object):
         self.shell = shell
 
     def vi_mode(self):
+        if not hasattr(self.shell.pt_app, 'editing_mode'):
+            return ''
         if self.shell.pt_app.editing_mode == 'VI':
             return '['+str(self.shell.pt_app.app.vi_state.input_mode)[3:6]+'] '
         return ''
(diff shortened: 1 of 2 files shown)


Carreau:
@meeseeksdev[bot] tag hacktoberfest

IPython/terminal/prompts.py
```diff
@@ -13,8 +13,17 @@ class Prompts(object):
     def __init__(self, shell):
         self.shell = shell
 
+    def vi_mode(self):
+        if not hasattr(self.shell.pt_app, 'editing_mode'):
+            return ''
+        if self.shell.pt_app.editing_mode == 'VI':
+            return '['+str(self.shell.pt_app.app.vi_state.input_mode)[3:6]+'] '
```
anntzer: Can this be made optional?  I nearly always know what mode I'm in (and in case I forgot I'm more likely to just reach for Escape just to be sure rather than check what's at the beginning of the line), so this feature is mostly just eating away 6 characters of my terminal width...
Thanks!

Carreau: Sure. Do you want to send a PR ? I guess we can either let the user define a value for both or a format string.

anntzer: done in #11492.

PR #11392, opened 2018-10-13
Cleaning up the interactive shell codebase

2DSharp:
Basic clean up, formatting, unnecessary if/else removal and putting the style overrides separately to make the functions smaller.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -227,11 +227,9 @@ class TerminalInteractiveShell(InteractiveShell):
     def init_term_title(self, change=None):
         # Enable or disable the terminal title.
         if self.term_title:
-            toggle_set_term_title(True)
             set_term_title(self.term_title_format.format(cwd=abbrev_cwd()))
-        else:
-            toggle_set_term_title(False)
-
+        toggle_set_term_title(self.term_title)
+        
     def init_display_formatter(self):
         super(TerminalInteractiveShell, self).init_display_formatter()
         # terminal only supports plain text
@@ -247,9 +245,11 @@ class TerminalInteractiveShell(InteractiveShell):
                 prompt_text = "".join(x[1] for x in self.prompts.in_prompt_tokens())
                 lines = [input(prompt_text)]
                 prompt_continuation = "".join(x[1] for x in self.prompts.continuation_prompt_tokens())
+
                 while self.check_complete('\n'.join(lines))[0] == 'incomplete':
                     lines.append( input(prompt_continuation) )
                 return '\n'.join(lines)
+
             self.prompt_for_code = prompt
             return
 
@@ -353,8 +353,8 @@ class TerminalInteractiveShell(InteractiveShell):
 
         return {
                 'complete_in_thread': False,
-                'lexer':IPythonPTLexer(),
-                'reserve_space_for_menu':self.space_for_menu,
+                'lexer': IPythonPTLexer(),
+                'reserve_space_for_menu': self.space_for_menu,
                 'message': get_message,
                 'prompt_continuation': (
                     lambda width, lineno, is_soft_wrap:
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -227,8 +227,10 @@ class TerminalInteractiveShell(InteractiveShell):
     def init_term_title(self, change=None):
         # Enable or disable the terminal title.
         if self.term_title:
+            toggle_set_term_title(True)
             set_term_title(self.term_title_format.format(cwd=abbrev_cwd()))
-        toggle_set_term_title(self.term_title)
+        else:
+            toggle_set_term_title(False)
         
     def init_display_formatter(self):
         super(TerminalInteractiveShell, self).init_display_formatter()
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -42,8 +42,6 @@ DISPLAY_BANNER_DEPRECATED = object()
 
 class _NoStyle(Style): pass
 
-
-
 _style_overrides_light_bg = {
             Token.Prompt: '#0000ff',
             Token.PromptNum: '#0000ee bold',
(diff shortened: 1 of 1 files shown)


2DSharp:
@Carreau Please review this.

IPython/terminal/interactiveshell.py
```diff
@@ -69,26 +87,24 @@ def get_default_editor():
...
-        return 'notepad' # same in Windows!
+    return 'notepad' # same in Windows!
 
 # conservatively check for tty
 # overridden streams can result in things like:
 # - sys.stdin = None
 # - no isatty method
 for _name in ('stdin', 'stdout', 'stderr'):
     _stream = getattr(sys, _name)
+    _is_tty = True
```
Carreau: you can actually move that out of the for loop if you really want.

Carreau:
In general we try to avoid cleaning up for the sake of cleaning up, it has many disadvantages: 

 - It often prevent from backporting patches, as now there are conflicts. 
 - It screw up git blame, and git bisect to find the root cause of a bug or the reason for change. 
 - It often may look like the same code but there may be some subtle behavior changes that are hard to track down. 

Especially since that both do some purely stylistic change and some code refactor I'm a bit worried. There is also some stylistic changes that are both incoherent with the rest of the codebase (and against pep8, but we only partially follow that.)

I'll try to review the code inline.

IPython/terminal/interactiveshell.py
```diff
@@ -58,6 +56,26 @@ class _NoStyle(Style): pass
...
+            Token.OutPromptNum: '#ff0000 bold'
+}
+
+_style_overrides_windows = {
+            Token.Prompt: '#ansidarkgreen',
+            Token.PromptNum: '#ansigreen bold',
+            Token.OutPrompt: '#ansidarkred',
+            Token.OutPromptNum: '#ansired bold'
+}
+
```
Carreau: THat seem more uniform, though that will conflict will already existing PR it would be better to do as separate changes. I'll see which PR is easier to apply first.

IPython/terminal/interactiveshell.py
```diff
@@ -69,26 +87,24 @@ def get_default_editor():
 
     if os.name == 'posix':
         return 'vi'  # the only one guaranteed to be there!
-    else:
-        return 'notepad' # same in Windows!
+    return 'notepad' # same in Windows!
```
Carreau: That's typically a change that's generally considered good, though this function is relatively simple, and that will make any further backport or history digging around this file more difficult. I dont' think it is really necessary.

IPython/terminal/interactiveshell.py
```diff
@@ -69,26 +87,24 @@ def get_default_editor():
...
         break
-else:
-    _is_tty = True
-
 
 _use_simple_prompt = ('IPY_TEST_SIMPLE_PROMPT' in os.environ) or (not _is_tty)
 
 class TerminalInteractiveShell(InteractiveShell):
-    space_for_menu = Integer(6, help='Number of line at the bottom of the screen '
+    space_for_menu = Integer(6, help ='Number of line at the bottom of the screen '
```
Carreau: I dont' think a space is needed here.

IPython/terminal/interactiveshell.py
```diff
@@ -211,11 +227,9 @@ def _displayhook_class_default(self):
     def init_term_title(self, change=None):
         # Enable or disable the terminal title.
         if self.term_title:
-            toggle_set_term_title(True)
             set_term_title(self.term_title_format.format(cwd=abbrev_cwd()))
-        else:
-            toggle_set_term_title(False)
-
+        toggle_set_term_title(self.term_title)
```
Carreau: Seem Ok, here but not strictly equivalent. What if term_title is false-ish ? like empty string,  None, or empty list in a subclass ? Does `toggle_set_term_title` allow non Boolean values ?

IPython/terminal/interactiveshell.py
```diff
@@ -289,30 +305,15 @@ def _make_style_from_name_or_cls(self, name_or_cls):
...
                 # the prompt colors will be wrong without this
                 if os.name == 'nt':
-                    style_overrides.update({
-                        Token.Prompt: '#ansidarkgreen',
-                        Token.PromptNum: '#ansigreen bold',
-                        Token.OutPrompt: '#ansidarkred',
-                        Token.OutPromptNum: '#ansired bold',
-                    })
-            elif legacy =='nocolor':
-                style_cls=_NoStyle
```
Carreau: That seem strait forward here, but actually the semantics are different, and I even believe some other branches in the code are wrong, In particular:

```
style_overrides = _style_overrides_linux
``` 

and 

```
style_overrides = _style_overrides_light_bg
```

Indeed, later on we do `style_overrides.update(self.highlighting_style_overrides)`, so if you assign something to `style_override` instead of using `update(...)` you will mutate the original. `_style_overrides_light_bg`, `_style_overrides_linux` and with your patch `_style_overrides_windows` which is surprising. 

So yes to deport them as global variable, but we need to be sure of what update() vs  assing will do.

Carreau:
This mostly looks good, but as I said above we try to avoid stylistic fixes for the sake of stylistic fixes, and try to get changes separately as otherwise you have these tiny semantic issues that crop up, and make increase the risk of introducing a bug.

IPython/terminal/interactiveshell.py
2DSharp: You are right, was about remove the space. Forgot about it.

IPython/terminal/interactiveshell.py
2DSharp: True, returning into the original state.

IPython/terminal/interactiveshell.py
2DSharp: Yea, didn't notice that. Thanks.

IPython/terminal/interactiveshell.py
2DSharp: So, does that mean the `_style_overrides_windows` part is not going to work?

aryeksa:
Closed

PR #11400, opened 2018-10-15
Fixes latex symbols in the completer (closes #11399)

### LINKED ISSUE #11399: Questionable choice for the greek codepoints in IPython.core.latex_symbols
The `IPython.core.latex_symbols` seems to contain odd choices for the greek letters. To make an example, the codepoint for epsilon is `GREEK LUNATE EPSILON SYMBOL` instead of the more obvious `GREEK SMALL LETTER EPSILON`. 

To check this, just run

```python
>>> import unicodedata
>>> from IPython.core.latex_symbols import latex_symbols
>>> unicodedata.name(latex_symbols['\epsilon'])
'GREEK LUNATE EPSILON SYMBOL'
>>> unicodedata.name('ε')
'GREEK SMALL LETTER EPSILON'
```

If I'm not wrong, the *lunate* version appeared in Unicode 3.1.0 (March, 2001), while the more common codepoint is in Unicode 1.1.0 (June, 1993).

LaTeX symbols seems to be autogenerated from Julia's one via the `/tools/gen_latex_symbols.py` so the issue probably derives from Julia code.  

Is there a reason for this choice?

mapio:
This PR addresses the tool to derive the list of LaTeX symbols from the corresponding Julia file.

FIrst of all, the location of the Julia source file has been updated from `https://raw.githubusercontent.com/JuliaLang/julia/master/base/latex_symbols.jl` to `https://raw.githubusercontent.com/JuliaLang/julia/master/stdlib/REPL/src/latex_symbols.jl`. 

Then the "parsing" (may I be forgiven for using such a term here) of such file has been made a bit more robust, and takes into account prefix definitions (via constants).

Finally, the tool has been run to update the list of symbols.

This apparently closes #11399.

Observe that many unicode names have been changed in the Python file, this seems to be in accordance to the names in Julia documentation https://docs.julialang.org/en/v1/manual/unicode-input/index.html.


Carreau:
Wow, indeed many changes. I'm find with getting that in.

I'll try to generate a table of the changes first though, to add in the documentation in case of questions.

mapio:
Most of the changes are due to a change in the (prefix of) the name of some variants (like the boldface, for instance). The change followed the one in Julia. Maybe it would be better to document such a "general" difference, instead of publishing a table with so many entries. Just my $.02

mapio:
Most of the changes are due to a change in the (prefix of) the name of some variants (like the boldface, for instance). The change followed the one in Julia. Maybe it would be better to document such a "general" difference, instead of publishing a table with so many entries. Just my $.02

Carreau:
Taking nodes for myself:

The characters corresponding to the following have been removed:
```
{'\\Elzxl', '\\textasciicaron', '\\underbar', '\\varepsilon'}
```

The characters corresponding to the following have been added:

```
{'\\jmath', '\\underbar', '\\underleftrightarrow'}
```

450 have the leading `m` removed., 216 have `\\mbfit' -> '\\bi`

Stoping there so far:

```
from IPython.core import latex_symbols as ls
from IPython.core import old_latex_symbols as ols
intersect = set(ls.reverse_latex_symbol.keys()).intersection(set(ols.reverse_latex_symbol.keys()))
n=0; m=0; k=0
for u in sorted(intersect):
    old = ols.reverse_latex_symbol[u]
    new = ls.reverse_latex_symbol[u]
    if old != new:
       if old.startswith('\\m') and old[3:]==new[2:]:
           m+=1
       elif old.replace('\\mbfit', '\\bi')==new:
           k+=1
       else:
           n+=1
           print(old, 'has been renamed to ', new)
```

I'm thinking we should support (at least for the 7-0 series the (oldname-> unicode) + potential warning. I'll see if it can write that up.

IPython/core/latex_symbols.py
```diff
@@ -102,68 +103,68 @@
...
-    "\\Elztesh" : "ʧ",
+    "\\hvlig" : "ƕ",
+    "\\nrleg" : "ƞ",
+    "\\doublepipe" : "ǂ",
+    "\\trna" : "ɐ",
+    "\\trnsa" : "ɒ",
+    "\\openo" : "ɔ",
+    "\\rtld" : "ɖ",
+    "\\schwa" : "ə",
+    "\\varepsilon" : "ε",
```
mapio: Regarding https://github.com/ipython/ipython/pull/11400#issuecomment-432867724 `\varepsilon` is still present.

mapio: It is changed the Unicode codepoint, now it is the *lunate* version as explained in https://github.com/ipython/ipython/issues/11399#issuecomment-431696925

Carreau: Yes, I understand the point I was just doing and dumb listing of the changes from a user point of view, and it still lead to the `ε` character being untypable with the latex equivalent. You can still use `\Greek Small Letter Epsilon<tab>` but I want to properly document it. Appologies if it was a confusing comment

Carreau:
I added a file. What do you think ?

mapio:
Looks fine for me! Thank you for your time, indeed.

stevengj:
For reference, most of the changes here were discussed in JuliaLang/julia#25980

Issue #11362, opened 2018-10-04
IPython directive parser broken for functions


TomAugspurger:
On IPython master, it seems like the `..ipython:: python` sphinx directive has trouble parsing function definitions

Some setup:

The `index.rst`

```rst
# file: index.html
Welcome to test's documentation!
================================

.. ipython:: python

   def f(x):
       x = 2
       return x

   f(2)


Testing
```

and the conf.py

```python
# file: conf.py
project = 'test'
copyright = '2018, tom'
author = 'tom'

# The short X.Y version
version = ''
# The full version, including alpha/beta/rc tags
release = ''

extensions = [
    'IPython.sphinxext.ipython_directive',
    'IPython.sphinxext.ipython_console_highlighting',
]

source_suffix = '.rst'
master_doc = 'index'
```

To build

```
$ sphinx-build -M html . _build
Running Sphinx v1.8.1
making output directory...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: 1 added, 0 changed, 0 removed
reading sources... [100%] index
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index
generating indices... genindex
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/html.
```

![screen shot 2018-10-04 at 8 52 45 am](https://user-images.githubusercontent.com/1312546/46478570-e41cd100-c7b2-11e8-93b5-dd52b16a2364.png)

Haven't done any debugging yet. Will take a look later.

TomAugspurger:
In `process_block`

```
(Pdb) pp data
(None, 'def f(x):\n    x = 2\n    return x', '')
(Pdb) self.process_input(data, input_prompt, lineno)
(['In [1]: def f(x):', '   ...:     x = 2', '   ...:     return x', '   ...: ', '  File "<ipython-input-2-50e9b99ff1e2>", line 1\n    return x\n            ^\nSyntaxError: \'return\' outside function\n\n'], ['def f(x):', '    x = 2', '    return x', ''], '  File "<ipython-input-2-50e9b99ff1e2>", line 1\n    return x\n            ^\nSyntaxError: \'return\' outside function\n\n', False, None, None, None)
```

So something in `process_input` in going awry.

dsblank:
[...]

PR #11402, opened 2018-10-16
Restore some functionality of the sphinx directive.

Carreau:
See #11362

The issue is in 2 part, before IPython 7.0 the input splitter was
state full, this was (in part) due to readline.

The second part is that because of this, we had to be a bit adressive of
what was considered complete code (it had to have 2 new line). This is
now not required anymore as we can submit stuff as a whole.

I hope that this fixes that. I have another fix in mind that count (and
reset) the number of consecutive blank line, but that will be more
complicated end code.

### DIFF
--- a/docs/source/development/ipython_directive.rst
+++ b/docs/source/development/ipython_directive.rst
@@ -1,0 +1,435 @@
+
+.. _ipython_directive:
+
+========================
+IPython Sphinx Directive
+========================
+
+.. note::
+
+   This has been salvadged from history, so information may be approximate or
+   duplicated. Fixes welcome.
+
+The ipython directive is a stateful ipython shell for embedding in
+sphinx documents.  It knows about standard ipython prompts, and
+extracts the input and output lines.  These prompts will be renumbered
+starting at ``1``.  The inputs will be fed to an embedded ipython
+interpreter and the outputs from that interpreter will be inserted as
+well.  For example, code blocks like the following::
+
+  .. ipython::
+
+     In [136]: x = 2
+
+     In [137]: x**3
+     Out[137]: 8
+
+will be rendered as
+
+.. ipython::
+
+   In [136]: x = 2
+
+   In [137]: x**3
+   Out[137]: 8
+
+.. note::
+    
+   This tutorial should be read side-by-side with the Sphinx source
+   for this document because otherwise you will see only the rendered 
+   output and not the code that generated it.  Excepting the example 
+   above, we will not in general be showing the literal ReST in this 
+   document that generates the rendered output.
+   
+
+The state from previous sessions is stored, and standard error is
+trapped. At doc build time, ipython's output and std err will be
+inserted, and prompts will be renumbered. So the prompt below should
+be renumbered in the rendered docs, and pick up where the block above
+left off.
+
+.. ipython::
+  :verbatim:
+
+  In [138]: z = x*3   # x is recalled from previous block
+
+  In [139]: z
+  Out[139]: 6
+
+  In [142]: print z
+  --------> print(z)
+  6
+
+  In [141]: q = z[)   # this is a syntax error -- we trap ipy exceptions
+  ------------------------------------------------------------
+     File "<ipython console>", line 1
+       q = z[)   # this is a syntax error -- we trap ipy exceptions
+	     ^
+  SyntaxError: invalid syntax
+
+
+The embedded interpreter supports some limited markup.  For example,
+you can put comments in your ipython sessions, which are reported
+verbatim.  There are some handy "pseudo-decorators" that let you
+doctest the output.  The inputs are fed to an embedded ipython
+session and the outputs from the ipython session are inserted into
+your doc.  If the output in your doc and in the ipython session don't
+match on a doctest assertion, an error will be
+
+
+.. ipython::
+
+   In [1]: x = 'hello world'
+
+   # this will raise an error if the ipython output is different
+   @doctest
+   In [2]: x.upper()
+   Out[2]: 'HELLO WORLD'
+
+   # some readline features cannot be supported, so we allow
+   # "verbatim" blocks, which are dumped in verbatim except prompts
+   # are continuously numbered
+   @verbatim
+   In [3]: x.st<TAB>
+   x.startswith  x.strip
+
+
+Multi-line input is supported. 
+
+.. ipython::
+   :verbatim:
+
+   In [130]: url = 'http://ichart.finance.yahoo.com/table.csv?s=CROX\
+      .....: &d=9&e=22&f=2009&g=d&a=1&br=8&c=2006&ignore=.csv'
+
+   In [131]: print url.split('&')
+   --------> print(url.split('&'))
+   ['http://ichart.finance.yahoo.com/table.csv?s=CROX', 'd=9', 'e=22',
+
+You can do doctesting on multi-line output as well.  Just be careful
+when using non-deterministic inputs like random numbers in the ipython
+directive, because your inputs are ruin through a live interpreter, so
+if you are doctesting random output you will get an error.  Here we
+"seed" the random number generator for deterministic output, and we
+suppress the seed line so it doesn't show up in the rendered output
+
+.. ipython::
+
+   In [133]: import numpy.random
+
+   @suppress
+   In [134]: numpy.random.seed(2358)
+
+   @doctest
+   In [135]: numpy.random.rand(10,2)
+   Out[135]:
+   array([[0.64524308, 0.59943846],
+          [0.47102322, 0.8715456 ],
+          [0.29370834, 0.74776844],
+          [0.99539577, 0.1313423 ],
+          [0.16250302, 0.21103583],
+          [0.81626524, 0.1312433 ],
+          [0.67338089, 0.72302393],
+          [0.7566368 , 0.07033696],
+          [0.22591016, 0.77731835],
+          [0.0072729 , 0.34273127]])
+
+
+Another demonstration of multi-line input and output
+
+.. ipython::
+   :verbatim:
+
+   In [106]: print x
+   --------> print(x)
+   jdh
+
+   In [109]: for i in range(10):
+      .....:     print i
+      .....:
+      .....:
+   0
+   1
+   2
+   3
+   4
+   5
+   6
+   7
+   8
+   9
+
+
+Most of the "pseudo-decorators" can be used an options to ipython
+mode.  For example, to setup matplotlib pylab but suppress the output,
+you can do.  When using the matplotlib ``use`` directive, it should
+occur before any import of pylab.  This will not show up in the
+rendered docs, but the commands will be executed in the embedded
+interpreter and subsequent line numbers will be incremented to reflect
+the inputs::
+
+
+  .. ipython::
+     :suppress:
+
+     In [144]: from pylab import *
+
+     In [145]: ion()
+
+.. ipython::
+   :suppress:
+
+   In [144]: from pylab import *
+
+   In [145]: ion()
+
+Likewise, you can set ``:doctest:`` or ``:verbatim:`` to apply these
+settings to the entire block.  For example,
+
+.. ipython::
+   :verbatim:
+
+   In [9]: cd mpl/examples/
+   /home/jdhunter/mpl/examples
+
+   In [10]: pwd
+   Out[10]: '/home/jdhunter/mpl/examples'
+
+
+   In [14]: cd mpl/examples/<TAB>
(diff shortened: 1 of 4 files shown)


TomAugspurger:
This fixed things for the small test in #11362 

I'm giving it a try on pandas docs.

TomAugspurger:
It seems like exceptions fail the build. With the same setup from #11362, 

```python
# index.rst
Welcome to test's documentation!
================================

.. ipython:: python

   def f(x):
       x = 2
       raise ValueError('bad')
       return x

   f(2)

.. ipython:: python

   print('hi')


Testing
```

then `make html` results in 

```
Running Sphinx v1.8.1
making output directory...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: 1 added, 0 changed, 0 removed
reading sources... [100%] index

>>>-------------------------------------------------------------------------
Exception in /Users/taugspurger/sandbox/sphinx-test/index.rst at block ending on line 12
Specify :okexcept: as an option in the ipython:: block to suppress this message
<ipython-input-2-c510dc86724b> in <module>
----> 1 f(2)

<ipython-input-1-fe8b2c77be70> in f(x)
      1 def f(x):
      2     x = 2
----> 3     raise ValueError('bad')
      4     return x

ValueError: bad
<<<-------------------------------------------------------------------------


Exception occurred:
  File "/Users/taugspurger/sandbox/IPython/IPython/sphinxext/ipython_directive.py", line 562, in process_input
    raise RuntimeError('Non Expected exception in `{}` line {}'.format(filename, lineno))
RuntimeError: Non Expected exception in `/Users/taugspurger/sandbox/sphinx-test/index.rst` line 12
The full traceback has been saved in /var/folders/hz/f43khqfn7b1b1g8z_z6y3bsw0000gp/T/sphinx-err-6psc98eb.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 2
```

I I include `okexcept` as an argument to the directive, things do finish. IIRC, the behavior on 6.x was to print a warning to stdout with a message telling you to use `okexcept`.

Carreau:
> I I include `okexcept` as an argument to the directive, things do finish. IIRC, the behavior on 6.x was to print a warning to stdout with a message telling you to use `okexcept`.

Yes but that was changed explicitley in https://github.com/ipython/ipython/pull/11321 (should maybe have been better documented) to actually start to catch regression in documentation.... 

I can try to make it configurable though. One other discussion was to make `okexcept` fail if there is actually no exception...

Carreau:
You can now set `ipython_strict_fail = False` in your `config.py` to get back previous behavior. 

if you have a better name...

TomAugspurger:
Thanks, that name seems fine, but if you want to match Sphinx I think it would be `ipython_warning_is_error` to match `warning-is-error`.

Carreau:
> `ipython_warning_is_error` to match `warning-is-error`

Done, 

SyntaxError is now detected as a error as well, I restored the old documentation from 0.13... that may need some polish.

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
+
+.. _ipython_directive:
+
+========================
+IPython Sphinx Directive
+========================
+
+.. note::
+
+   This has been salvadged from history, so information may be approximate or
```
choldgraf: I'd just say something like "The IPython Sphinx Directive is in 'beta' and currently under active development. Improvements to the code or documentation are welcome!"

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+  SyntaxError: invalid syntax
+
+
+The embedded interpreter supports some limited markup.  For example,
+you can put comments in your ipython sessions, which are reported
+verbatim.  There are some handy "pseudo-decorators" that let you
+doctest the output.  The inputs are fed to an embedded ipython
+session and the outputs from the ipython session are inserted into
+your doc.  If the output in your doc and in the ipython session don't
+match on a doctest assertion, an error will be
```
choldgraf: error will be...

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+
+
+.. ipython::
+
+   In [1]: x = 'hello world'
+
+   # this will raise an error if the ipython output is different
+   @doctest
```
choldgraf: is @doctest explicitly documented somewhere? or only in reference here?

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+
+  In [141]: q = z[)   # this is a syntax error -- we trap ipy exceptions
+  ------------------------------------------------------------
+     File "<ipython console>", line 1
+       q = z[)   # this is a syntax error -- we trap ipy exceptions
+	     ^
+  SyntaxError: invalid syntax
+
+
+The embedded interpreter supports some limited markup.  For example,
```
choldgraf: I'd split each of these sub-sections by a header. E.g.

```
Persisting the Python session across IPython directive blocks
=============================================================

...

Adding documentation tests to your IPython directive
====================================================

...

Multi-line input
================

etc

File: docs/source/development/ipython_directive.rst [hunk 5]
```diff
@@ -0,0 +1,435 @@
...
+   In [130]: url = 'http://ichart.finance.yahoo.com/table.csv?s=CROX\
+      .....: &d=9&e=22&f=2009&g=d&a=1&br=8&c=2006&ignore=.csv'
+
+   In [131]: print url.split('&')
+   --------> print(url.split('&'))
+   ['http://ichart.finance.yahoo.com/table.csv?s=CROX', 'd=9', 'e=22',
+
+You can do doctesting on multi-line output as well.  Just be careful
+when using non-deterministic inputs like random numbers in the ipython
+directive, because your inputs are ruin through a live interpreter, so
```
choldgraf: *run

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+if you are doctesting random output you will get an error.  Here we
```
choldgraf: I think that the "random output" comparison step can be a bit clearer. Something like:

```
Testing directive outputs
=========================

The IPython Sphinx Directive makes it possible to test the outputs that you provide with your code. To do this, decorate the contents in your directive block with one of the following:

  * list directives here

If an IPython doctest decorator is found, it will take these steps when your documentation is built:

1. Run the *input* lines in your IPython directive block against the current Python kernel (remember that the session persists across IPython directive blocks).
2. Compare the *output* of this with the output text that you've put in the IPython directive block (what comes after `Out[NN]`)
3. If there is a difference, the directive will raise an error and your documentation build will fial.

File: docs/source/development/ipython_directive.rst [hunk 7]
```diff
@@ -0,0 +1,435 @@
...
+text, and then resave 
+
+.. ipython::
+
+
+   In [151]: ylabel('number')
+
+   In [152]: title('normal distribution')
+
+   @savefig hist_with_text.png width=4in
```
choldgraf: maybe this should be called `display_fig` or `show_fig` instead of savefig? From a user's standpoint I imagine they'd think "I want to show this figure" not "I want to save this figure"

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+   print(foo)
+   foo = 2
+   foo**2
+
+We can even plot from python, using the savefig decorator, as well as, suppress
+output with a semicolon
+
+.. ipython:: python
+
+   @savefig plot_simple_python.png width=4in
```
choldgraf: will this plot a figure (because of `@savefig`) or will it plot nothing (because of the semicolon)?

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+
+If you don't see the next code block then the options work.
+
+.. ipython:: python
+   :suppress:
+
+   ioff()
+   ion()
+
+Multi-line input is handled.
```
choldgraf: People could confuse this with multi-line input in the ipython section above. Maybe call this "splitting Python statements across lines"?

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+
+.. ipython:: python
+
+   line = 'Multi\
+           line &\
+           support &\
+           works'
+   print(line.split('&'))
+
+Functions definitions are correctly parsed
```
choldgraf: I'd have a section called "syntax highlighting" and just step through several major use-cases there. E.g.:

```
Syntax highlighting
===================

The Sphinx IPython directive can highlight most Python structures and patterns. This section shows some of the most common ones.

Multi-line statements
---------------------
EXAMPLE CODE

Function definitions
--------------------
EXAMPLE CODE

etc
```

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+
+@verbatim
+
+    insert the input and output block in verbatim, but auto-increment
+    the line numbers. Internally, the interpreter will be fed an empty
+    string, so it is a no-op that keeps line numbering consistent.
+    Also, can be applied to the entire ``.. ipython`` block as a
+    directive option with ``:verbatim:``.
+
+@savefig OUTFILE [IMAGE_OPTIONS]
```
choldgraf: Ah ok - so the decorators are here. In that case you should always link to this section whenever you mention a decorator above. Something like `for more information on decorator usage, see :ref:`directive_decorators` and then add the label `.. _directive_decorators:` to this section

docs/source/development/ipython_directive.rst
```diff
@@ -0,0 +1,435 @@
...
+@doctest
+
+    Compare the pasted in output in the ipython block with the output
+    generated at doc build time, and raise errors if they don't
+    match. Also, can be applied to the entire ``.. ipython`` block as a
+    directive option with ``:doctest:``.
+
+Configuration Options
+=====================
+
```
choldgraf: Specify that these are the configurations you'd provide underneath the directive with `:option:`

choldgraf (commented):
In general it looks quite good - I have a few thoughts for clarifying/structuring the documentation!

Carreau:
Thanks @choldgraf for the feedback, I've open #11410 as a easy first issue to follow up on. 

I'm going to get that in as it fixes some functionalities.

docs/source/development/ipython_directive.rst
Carreau: Yes, feel free to make any replacement you feel is better.

docs/source/development/ipython_directive.rst
LucianaMarques: Could someone please tell me more about @doctest? Where is it documented?

LucianaMarques: Ok, now I see there's a definition at the bottom of the file. But some more info on it would be nice, I find it a little confusing.

docs/source/development/ipython_directive.rst
LucianaMarques: Hi @choldgraf , I don't know if I understood quite well what you meant here. 

Do you mean creating another section with a whole set example code blocks or rename an existing section for Syntax Highlighting and use the code that has already been written?

Thank you

docs/source/development/ipython_directive.rst
LucianaMarques: Is this supposed to create some kind of a link? I didn't understand much how am I supposed to add this reference. But good idea, this certainly helps.

Carreau: No, that is just an example on how to use `savefig`. Do what you can, even partially. If it's annoying or confusing send what you have and feel free to move on to something more interesting.

docs/source/development/ipython_directive.rst
Carreau: I think chris was suggesting refactoring and being clearer with heading/subheading. 

You can see the rendered version [there](https://ipython.readthedocs.io/en/latest/development/ipython_directive.html), There is a lot of duplicate information and organisational issues.

docs/source/development/ipython_directive.rst
LucianaMarques: Hey Carreau, I think I had this doubt because I was not familiar with reStructureText until today. Thank you for your reply!

Carreau:
No problem, restructured text can be complicated at time.

PR #11411, opened 2018-10-18
try to fix nightly builds on travis.

Carreau:


Carreau:
That fix the parso thing, now we can see regression in nightly.

PR #11412, opened 2018-10-18
5.x

sgpritam:
remove IPython.core.prompts
In master, yes, but not in a release. ipykernel 4.4 was released in August. In general, master of every individual IPython or Jupyter package should work with the stable version of everything else unless something has gone wrong. Beyond that, every breaking change should really solve a problem. I don't think cleaning up is a good enough reason to break compatibility in general, but we've done two releases of ipykernel that will work, and IPython 6 is still some time off, so I think this is probably okay.


minrk:
Looks like this PR was opened by mistake.

PR #11414, opened 2018-10-19
Update setupbase.py

ansalm:
added libraries


minrk:
Thanks for the PR, but I don't think we should make this change.

- numpy and maptlotlib are not used during installation, and cause installation to fail
- reordering a list doesn't appear to serve any purpose

PR #11419, opened 2018-10-20
removed expansion of function name when completing kw-args

### LINKED ISSUE #11408: Tab completion for method arguments shows other functions args.
With the below example class the args list that shows up when tab completing has arguments from other methods on the class that share the same root name.

```python
class Test:
    def meth(self, meth_arg1):
        print("meth")
    
    def meth_1(self, meth1_arg1, meth1_arg2):
        print("meth1")
    
    def meth_2(self, meth2_arg1, meth2_arg2):
        print("meth2")
```

jupyter_autocomplete

Using jupyter==1.0.0, ipython==7.0.1

elyashiv:
fix #11408 

There seems to be a function expansion when searching for kw-args, and that causes the bug.
I removed it and it solved it, but I'm not sure why it's there, so I'm not sure it's a complete solution.

Also, I'm not quite sure how to run tests, so I haven't done any regression tests yet.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1538,24 +1538,20 @@ class IPCompleter(Completer):
 
             usedNamedArgs.add(token)
 
-        # lookup the candidate callable matches either using global_matches
-        # or attr_matches for dotted names
-        if len(ids) == 1:
-            callableMatches = self.global_matches(ids[0])
-        else:
-            callableMatches = self.attr_matches('.'.join(ids[::-1]))
         argMatches = []
-        for callableMatch in callableMatches:
-            try:
-                namedArgs = self._default_arguments(eval(callableMatch,
-                                                        self.namespace))
-            except:
-                continue
+        try:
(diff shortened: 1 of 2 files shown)


elyashiv:
Now I see there is some CI... :)

IPython/core/completer.py
```diff
@@ -1538,24 +1538,20 @@ def python_func_kw_matches(self,text):
...
         argMatches = []
-        for callableMatch in callableMatches:
-            try:
-                namedArgs = self._default_arguments(eval(callableMatch,
-                                                        self.namespace))
-            except:
-                continue
+        try:
+            callableObj = '.'.join(ids[::-1])
+            print(callableObj)
```
Carreau: Forgotten debug ? 
```suggestion

```

Carreau:
Thanks ! I'll have to look at that more in depth, but it would be good to have a test to not have regressions :-)

Carreau (approved):

tot0:
Thanks so much for fixing this!

PR #11422, opened 2018-10-20
Bit nitpicky about rst markup, find more errors.

Carreau:


Carreau:
I remove `nitpicky=True` there are too many errors.

PR #11425, opened 2018-10-20
Fix indentation for nested code block

### LINKED ISSUE #11416: Wrong autoindent for  class method.
```
class F:
    def a(self):<enter>
```

Only 4 spaces are inserted.

haivp3010:
Fix #11416.

The problem was not just with class method but a nested block in general, for example an if else block inside a function definition. I'm not sure whether the workaround for the inconsistency between python's and cpython's module was perfect, but it works for the moment :)

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -655,7 +655,15 @@ class TransformerManager:
         if len(tokens_by_line) == 1 and not tokens_by_line[-1]:
             return 'incomplete', 0
 
-        if tokens_by_line[-1][-1].string == ':':
+        new_block = False
+        for token in reversed(tokens_by_line[-1]):
+            if token.type == tokenize.DEDENT:
+                continue
+            elif token.string == ':':
+                new_block = True
+            break
+
+        if new_block:
             # The last line starts a block (e.g. 'if foo:')
             ix = 0
             while tokens_by_line[-1][ix].type in {tokenize.INDENT, tokenize.DEDENT}:
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -481,6 +481,17 @@ def make_tokens_by_line(lines):
     if not tokens_by_line[-1]:
         tokens_by_line.pop()
 
+    # Convert if using cpython tokenize
+    if (list(map(lambda x: x.type, tokens_by_line[-1])) == 
+            [tokenize.DEDENT] * (len(tokens_by_line[-1]) - 1) + [tokenize.ENDMARKER]):
+        if (
+            len(tokens_by_line) > 1 and
+            len(tokens_by_line[-2]) > 0 and
+            tokens_by_line[-2][-1].type == tokenize.NEWLINE
+        ):
+            tokens_by_line[-2].pop()
+            tokens_by_line[-2] += tokens_by_line[-1]
+            tokens_by_line.pop()
 
     return tokens_by_line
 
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks, I'll have a look and try to understand. 

> between python's and cpython's

What do you mean by that ? do you mean `ipython's` ?  if so we probably want to fix out tokenize method itself.

Carreau:
OK, that make sens. So one question, would it make sens to move that inside `make_tokens_by_line()` so that `make_tokens_by_line()`  return consistent value regardless of the version of Python ? That feels cleaner to me.  

Thoughts  ?

haivp3010:
Yes thank you, I missed that. It definitely makes sense to move the logic into `make_tokens_by_line()`. I've made another commit for that.

IPython/core/inputtransformer2.py
```diff
@@ -481,6 +481,17 @@ def make_tokens_by_line(lines):
     if not tokens_by_line[-1]:
         tokens_by_line.pop()
 
+    # Convert if using cpython tokenize
```
Carreau:
```suggestion
    # Convert if using cpython tokenize
    # upstream bug was fixed in Python 3.7.1, so once we drop 3.7 this can likely be removed. 
```

Carreau:
> Yes thank you, I missed that. It definitely makes sense to move the logic into `make_tokens_by_line()`. I've made another commit for that.

Ok, great, I added a suggestion of comment to remind us when to remove that in a couple of years. 
Once this is done i'm +1

Carreau (approved):

IPython/core/inputtransformer2.py
```diff
@@ -481,6 +481,18 @@ def make_tokens_by_line(lines):
...
+    # upstream bug was fixed in Python 3.7.1, so once we drop 3.7 this can likely be removed.
```
takluyver: Can we link to the bugs.python.org URL for it for reference? I don't know what bug we're working round.

IPython/core/inputtransformer2.py
```diff
@@ -481,6 +481,18 @@ def make_tokens_by_line(lines):
     if not tokens_by_line[-1]:
         tokens_by_line.pop()
 
+    # Convert if using cpython tokenize
+    # upstream bug was fixed in Python 3.7.1, so once we drop 3.7 this can likely be removed. 
+    if (list(map(lambda x: x.type, tokens_by_line[-1])) == 
+            [tokenize.DEDENT] * (len(tokens_by_line[-1]) - 1) + [tokenize.ENDMARKER]):
```
takluyver: Not quite as precise, because it ignores the order, but maybe we could simplify the check using sets:

```python
if {t.type for t in tokens_by_line[-1]} == {tokenize.DEDENT, tokenize.ENDMARKER}:
```

Carreau:
@takluyver  ! Good to see you ! 

I would suggest waiting for GH to be back on green: https://status.github.com/messages
some messages have been lost, and PR unmerged.

IPython/core/inputtransformer2.py
Carreau: If you want to keep the oder I believe you can compare list comprehensions:
```
[t.type for t in tokens_by_line[-1]][-2:] == [tokenize.DEDENT, tokenize.ENDMARKER]
```

```
In [4]: [0, 1,2][-2:] == [1, 2]
Out[4]: True
```

takluyver:
Hello :-)

Yup, it's an unfortunate time to be busy on Github. I've been keeping an eye on the status most of the day; it seems to be mostly back to normal now.

Carreau:
> it seems to be mostly back to normal now.

For *you* you are probably going through a EU server :-)

IPython/core/inputtransformer2.py
takluyver: That's a looser check, though - you're ignoring anything but the last two things in the line.

haivp3010:
```
        """
        # Remember if the lines ends in a new line.
        ends_with_newline = False
        for character in reversed(cell):
            if character == '\n':
                ends_with_newline = True
                break
            elif character.strip():
                break
            else:
                continue

        if ends_with_newline:
            # Append an newline for consistent tokenization
            # See https://bugs.python.org/issue33899
            cell += '\n'
```
This is the code at the top of the check_complete, I think the last `if` statement should have been `if not end_with_newline:`. If so, the fix could be further simplified. Also, currently the logic of the fix is to convert from the recently fixed module in python 3.7.1 back to the old form in prior python version. It would be more appropriate to change the fix to address the problem in older python version, rather than manipulating the now fixed implementation. Any thoughts?

Carreau:
I would have to read the code, but converting to new version seem the right
move. Let's try to change and see if the test are failling ?

haivp3010:
> I would have to read the code, but converting to new version seem the right move. Let's try to change and see if the test are failling ?
OK I've made the changes and the tests do pass, so hopefully we can stick with this implementation.

Carreau:
Thanks for working with us on this, deeply appreciated !

You may still be able to get a T-shirt for [Hacktoberfest](https://hacktoberfest.digitalocean.com/) !

haivp3010:
Thank you! On my way to get the T-shirt \m/

PR #11446, opened 2018-10-25
set ls aliases according to colors configuration (closes #11431)

### LINKED ISSUE #11431: `ls` magic alias to `ls -F --color`
I have many different profiles used in linux. I use one profile when my terminal does not support any color.
Is there any option to control the `ls` aliases so that it uses `ls -F` instead of `ls -F --color`?

I already used `c.InteractiveShell.colors = 'NoColor'` but it does not affect the `ls` alias

liushapku:
In Linux, a set of aliases are set to list files, such as `ls` and `ll` which uses `--color` even when `c.InteractiveShell.colors=NoColor` is used. This is to have the aliases respect the colors setting so that it will not mess up the terminal which do not support color.

### DIFF
--- a/IPython/core/alias.py
+++ b/IPython/core/alias.py
@@ -204,6 +204,8 @@ class AliasManager(Configurable):
     def init_aliases(self):
         # Load default & user aliases
         for name, cmd in self.default_aliases + self.user_aliases:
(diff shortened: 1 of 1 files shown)


Carreau:
thanks !

PR #11450, opened 2018-10-25
Corrected path to InteractiveShell

authalic:
changed TerminalInteractiveShell references to Terminal.InteractiveShell

### DIFF
--- a/docs/source/config/details.rst
+++ b/docs/source/config/details.rst
@@ -31,7 +31,7 @@ which defines the defaults. The required interface is like this:
 .. class:: MyPrompts(shell)
 
    Prompt style definition. *shell* is a reference to the
-   :class:`~.TerminalInteractiveShell` instance.
+   :class:`~.Terminal.InteractiveShell` instance.
 
    .. method:: in_prompt_tokens(cli=None)
                continuation_prompt_tokens(self, cli=None, width=None)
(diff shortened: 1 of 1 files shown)


docs/source/config/details.rst
```diff
@@ -111,7 +111,7 @@ You can see the Pygments styles available on your system by running::
     import pygments
     list(pygments.styles.get_all_styles())
 
-Additionally, ``TerminalInteractiveShell.highlighting_style_overrides`` can override
+Additionally, ``Terminal.InteractiveShell.highlighting_style_overrides`` can override
```
Carreau: I'm Thanks, but I'm pretty at least some of these are correct: 

```
In [1]: from IPython.terminal.interactiveshell import TerminalInteractiveShell, InteractiveShell

In [2]: TerminalInteractiveShell.highlighting_style_overrides
Out[2]: <traitlets.traitlets.Dict at 0x105691eb8>

In [3]: InteractiveShell.highlighting_style_overrides
<ipython-input-3-e268df97a2b4> in <module>
----> 1 InteractiveShell.highlighting_style_overrides

AttributeError: type object 'InteractiveShell' has no attribute 'highlighting_style_overrides'
```

note that there is __both__ TerminalInteractiveShell and InteractiveShell, the former one used in terminal IPython, the later one inherited from in IPykernel to connect to – for example – the notebook.

PR #11452, opened 2018-10-27
Kd2718/better windows warnings

### LINKED ISSUE #11451: Better warning for windows users when using %run
After working on #10079, I thought it would be better to give windows users an explanation as to why they are receiving that error. Currently it just shows them a standard error with code that looks odd, but works on *nix systems:
`ERROR:root:File "'../test.py'.py" not found.`

I added a `warn` call to try to help:
```
  warn('For Windows, use double quotes to wrap a filename: %run "mypath\\myfile.py"')
ERROR:root:File `"'../test.py'.py"` not found.
```

I hopped that this would point them into the right direction. Especially if they are new users.

I am not sure if using a `warn()` is the best way to handle this, so let me know any thoughts.

kd2718:
fixes #11451

I wasn't sure on the use of `warn()` or the wording. Let me know if there is a better way to handle this.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -16,6 +16,7 @@ import sys
 import time
 import timeit
 import math
+import re
 from pdb import Restart
 
 # cProfile was added in Python2.5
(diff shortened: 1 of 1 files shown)


Carreau:
LGTM.

Carreau:
Thanks.

PR #11458, opened 2018-10-27
fix rehashx on macos

Carreau:
cc @bollwyvl


Carreau:
See https://github.com/conda-forge/ipython-feedstock/pull/63

PR #11461, opened 2018-10-28
Use inspect.cleandoc in autogen_config.

anntzer:
... instead of a custom implementation of the same thing.

xref https://github.com/ipython/ipython/pull/11317#discussion_r228745652.


Carreau (approved):

Carreau:
+1 thanks !

PR #11464, opened 2018-10-29
fix for PySide2 cannot be activated since it is directly deactivated …

bergtholdt:
…again in the else part

Before I got this error:

ImportError: 
    Importing PySide2 disabled by IPython, which has
    already imported an Incompatible QT Binding: pyside2


Carreau:
Thanks !

PR #11471, opened 2018-10-31
Add OSX on 3.6 and 3.7 to travis, use addons.apt:packages for graphviz

bollwyvl:
This adds 3.6/3.7 OSX builds to the travis-ci matrix. No doubt this comes at greater cost to shared resources per PR, but so it goes!

In reading the [Travis docs](https://docs.travis-ci.com/user/multi-os/), this also moves the `apt` install up to the addon level, since we weren't pinning or anything. ~~The equivalent `brew` command is also moved up to `before_install`.~~ (not building docs on osx)

c/f https://github.com/conda-forge/ipython-feedstock/pull/63#issuecomment-433657448

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ group: edge
 before_install:
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update                    ; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install pyenv             ; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then pyenv install "$PYENV_VERSION" ; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then pyenv local "$PYENV_VERSION"   ; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ group: edge
 before_install:
     - 'if [[ $GROUP != js* ]]; then COVERAGE=""; fi'
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update                    ; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update pyenv              ; fi
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew upgrade pyenv             ; fi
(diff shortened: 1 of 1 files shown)


bollwyvl:
New matrix looks like I didn't screw anything major up... yet:
screenshot from 2018-10-30 21-18-30

bollwyvl:
Har:
```
3.6 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/osx/10.13/x86_64/python-3.6.tar.bz2
$ curl -sSf -o python-3.6.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 403 Forbidden
Unable to download 3.6 archive. The archive may not exist. Please consider a different version.
```

bollwyvl:
Great, looks like the OSX CI is going through. The allowed `nightly` fail appears to not be related to this.

The first [working 3.6 build](https://travis-ci.org/ipython/ipython/builds/448665979?utm_source=github_status&utm_medium=notification) took 8 minutes, and was slower to start anyway. Investigating whether caching the built pythons makes a net difference (of course, would need to succeed....)

bollwyvl:
Oh, I love you, first :heavy_check_mark:. Ooh, and coverage went up!

bollwyvl:
Ugh, looks like the cache situation isn't great: appears to have cached `bin/ipython` and friends, and we have no interest in that. So, we're looking at 8-9 minutes for the OSX builds at present. Thoughts?

minrk:
Updating homebrew and building Python from source with pyenv both take a pretty long time. I added a commit with a couple small changes:

- only install Python if not present
- install Python binaries from Python.org instead of from source
- avoid updating homebrew

The mac builds now complete in 3 minutes.

bollwyvl:
Fantastic. I don't have a Mac so was stumbling in the dark a bit. Also +1
to yaml formatting... api_key might still be indented too far.

Moving to a virtualenv may have missed a few test cases, but I can't
interpret the codecov results properly on my phone!

Carreau:
Do we still have a JS group ? 

+1 will look at it closer from my phone.

Carreau:
*from my desktop

bollwyvl:
> Do we still have a JS group ?

No, removed!

bollwyvl:
Here's the only virtualenv coverage change:

https://codecov.io/gh/ipython/ipython/compare/70d51e3...a39b70da8c26968e72342f231f9349cc52fd69b5/changes

Probably not worth worrying about.

bollwyvl:
Hooray!

Issue #11475, opened 2018-11-02
inputsplitter doesnot work properly in --simple-prompt mode after ctrl-c interrupt


liushapku:
The followings applies to `IPython 6.5.0`. I haven't tested it for `IPython 7`

# Actions to generate the symptom
1.  run `ipython --simple-prompt`
2. insert

    def f():
      print(
      # hit ctrl-c here

3. then `ipython` will not be able to execute any code after displaying `KeyboardInterrupt escaped interact()`

# Cause 
1. in `TerminalInteractiveShell.mainloop`, when `KeyboardInterrupt` is encountered, `ipython` will enter another call to `self.interact()`, which calls `self.prompt_for_code()`.
 
2. When `--imple-prompt` is used, `prompt_for_code` is defined as
```
def prompt():
    isp = self.input_splitter
    prompt_text = "".join(x[1] for x in self.prompts.in_prompt_tokens())
    prompt_continuation = "".join(x[1] for x in self.prompts.continuation_prompt_tokens())
    while isp.push_accepts_more():
        line = input(prompt_text)
        isp.push(line)
        prompt_text = prompt_continuation
    return isp.source_reset()
self.prompt_for_code = prompt
```

3. the internal status of `self.input_splitter` is corrupted when `KeyboardInterrupt` is raised. It will keep asking for new code and never be able to run new code

# Possible solution
call `isp.reset()` inside `prompt()` before asking for code.

PR #11476, opened 2018-11-02
fix input_splitter error after KeyboardInterrupt with in --simple-prompt mode (issue #11475)

liushapku:
addresses [issue 11475](https://github.com/ipython/ipython/issues/11475)

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -232,6 +232,7 @@ class TerminalInteractiveShell(InteractiveShell):
             # This is very limited, and only accepts a single line.
             def prompt():
                 isp = self.input_splitter
+                isp.reset()
                 prompt_text = "".join(x[1] for x in self.prompts.in_prompt_tokens())
                 prompt_continuation = "".join(x[1] for x in self.prompts.continuation_prompt_tokens())
                 while isp.push_accepts_more():


liushapku:
have no idea why the integration for python 3.3 failed.

Carreau:
The failure  on Python 3 is unrelated. It's just that the new version of Pip is incompatible with Python 3.

I'm happy to get that in, but there will likely not be any further release of IPython 6.x. Do you have any specific reason to want to fix 6.x ?

liushapku:
It is just because our team uses a specific `conda` environment which stick to 6.x. It might take time to migrate to 7. Is there any compatibility issue between 6 and 7? What is the major difference between 6 and 7?

liushapku:
If there is no further release, how can you get it in? Should we just ignore all these and use 7?

Carreau:
For all the difference between 6.x and 7.x see [the what's new](https://ipython.readthedocs.io/en/stable/whatsnew/version7.html). 

There is no major compatibility issue if you don't use custom extensions. 

I'm happy to get the fix in the 6.x branch, but it's likely never going to end up on PyPI or Conda. 
You can still install source from a specific commit if that helps. (That is unless there is interest for someone to maintain 6.x, but the current core dev are doing that on their free time and already plenty to do).

But in general yes as an open source project we only support the latest branch (in this case 7.x) and use the latest unless you you really need to stick to an old one. If you need support for an old version you can also discuss support with Anaconda Inc.

liushapku:
That's fairly reasonable. Thanks for having created such a superb software that having benefitted the whole community.

Carreau:
Many thanks for your contribution. I'm going to close this as it is likely to never be released. Still looking forward to more PR coming from you.

PR #11477, opened 2018-11-03
pin pip on old python

Carreau:


Carreau:
@meeseeksdev backport to 5.x

Issue #11432, opened 2018-10-22
Show subclasses when using pinfo/pinfo2/?/??


Carreau:
See https://github.com/python/cpython/pull/5066

We should probably match that.

Extra points, when possible display the tree hierarchy ?

Carreau:
Our goal here is to get something like the following. Once it is working we can refine a nd get a better User interface: 

```
$ ipython
Python 3.6.5 | packaged by conda-forge | (default, Apr  6 2018, 13:44:09)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: class Foo:
   ...:     pass
   ...:

In [2]: class Bar(Foo):
   ...:     pass
   ...:

In [3]: Foo?
Init signature: Foo()
Docstring:      <no docstring>
Type:           type
Subclasses:      - Bar    <- that is new and useful
```

Here we have only one subclass, we probably want to limit to the first few subclasses if we have hundreds. 


## How to implement that. 

1) make sure you have IPython working on your machine. We recommend installing anaconda. And then `conda install ipython`

2) clone this repository:

```
$ git clone https://github.com/ipython/ipython
```

Cd into the repository and do a dev-install : 
```
$ cd ipython
$ pip install -e . 
```

now when you start `$ ipython` , the third live should say something like `7.2.0.dev` if you see dev, then all s good. This mean that you run IPython from the source, and noot from the install packages. You are officially a developper now. 

Time to do modifications to the source code. They will be taken into account everytime  you restart IPython . 

In our case we want to modify `IPython/core/oinspect.py`. Around line 830 find the line:

```
if inspect.isclass(obj):
```
Under this block we want to modify `out`,  to contain the list of subclasses. 

We can get all the subclasses of `obj` by using `obj__subclasses()`
As we want a something that look nice to the user we are actually interested in in each subclass name, and not the subclass them selves. That is to say we want to have:

`Subclasses:      - Bar` and not `Subclasses:      - <class bar at 0x123456>`
so let's so around line 862: 

```
names = [sub.__name__ for sub in obj.__subclasses__()]
```

We now want to join all the names by commas for example:
```
all_names = ', '.join(names)
```

and let's now put that into `out` under the names `subclasses`.::

```
out['subclasses'] = all_names
```

Good ! We've done the hard work. 

Now close to line 642 find `elif info['isclass']`, 

Modify close to the end of the block to look like : 

```diff
[...]

PR #11486, opened 2018-11-08
Added ability to show subclasses when using pinfo and other utilities

cmentzel:
addresses issue #11432 with thanks to @Carreau for basically doing everything

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -22,7 +22,7 @@ import os
 from textwrap import dedent
 import types
 import io as stdlib_io
-from itertools import zip_longest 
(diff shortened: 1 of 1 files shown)


Carreau:
The error is unrelated to what you did, we just added a `subclasses` field and other part of IPython is not handling that correctly. I've added one commit that hopefully should fix that.

Many thanks and awesome work ! Let's wait for the test to pass (just to make sure I did not forget anything), and then we can merge what you did into master.

Carreau:
All is green ! congrats !

cmentzel:
Thanks!

PR #11492, opened 2018-11-11
Make inclusion of vi mode in prompt togglable.

anntzer:
See discussion in https://github.com/ipython/ipython/pull/11390#discussion_r232088166.

If we really want to make the prompt configurable I guess the config value could be a function taking an `InputMode` as argument and returning a string, but that feels a bit overkill?

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -225,6 +225,10 @@ class TerminalInteractiveShell(InteractiveShell):
         help="Allows to enable/disable the prompt toolkit history search"
     ).tag(config=True)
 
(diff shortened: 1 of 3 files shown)


Carreau:
LGTM.

Evidlo:
For posterity, the option to disable is `TerminalInteractiveShell.prompt_includes_vi_mode`, which can be set in `ipython_config.py` as documented [here](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-TerminalInteractiveShell.prompt_includes_vi_mode).

``` python
c.TerminalInteractiveShell.prompt_includes_vi_mode = False
```

PR #11499, opened 2018-11-18
Fix extension detection

chbrown:
If the path leading up to the Python installation's `site-packages` contains a `.`, this check does not do what it says it does :(

For instance, on macOS, the `site-packages` directory is located at `/usr/local/lib/python3.7/site-packages`, which means that trying to use the magic `%run -m my.module` will always fail, because `split_path[1]` will always start with `7/site-packages`.

There are better ways to check that a file extension matches expectations, but I thought it was cute that I could fix this bug by inserting a single character :)

My current workaround looks like this:
```python
import importlib
import_path = importlib.util.find_spec('my.module').origin
%run $import_path
```

### DIFF
--- a/IPython/utils/module_paths.py
+++ b/IPython/utils/module_paths.py
@@ -64,7 +64,7 @@ def find_mod(module_name):
         return None
     else:
         split_path = module_path.split(".")
-        if split_path[1] in ["py", "pyw"]:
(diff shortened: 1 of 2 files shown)


Carreau:
THanks that seem reasonable, I'll see if I can add a test to make sure there is no regressions.

PR #11502, opened 2018-11-21
add CWD to sys.path *after* stdlib

minrk:
this deviates from Python's own interactive behavior, but seems to avoid problems where folks create stdlib-names like `email.py`.

They can still collide with packages, but that seems less common than generic names like `email` and `code`.

I'm not entirely sure if it's a good idea to deviate from Python like this, but it would solve a problem that confuses our users from time to time.


Carreau:
+1. Do you want to add a test ?

Carreau:
Probably rare enough to not need test.

PR #11505, opened 2018-11-21
Render signatures ourselves if they’re too long

### LINKED ISSUE #11504: Inspected functions’ signatures undecipherable when type hints are in play
In the `_getdef` method, the function signature is rendered:

https://github.com/ipython/ipython/blob/5e7f930d355f9bb65b22ed8e8b735ea75c994d33/IPython/core/oinspect.py#L365-L366

I have formatted it well in my source:

```py
def neighbors(
    adata: AnnData,
    n_neighbors: int = 15,
    n_pcs: Optional[int] = None,
    use_rep: Optional[str] = None,
    knn: bool = True,
    random_state: Optional[Union[int, RandomState]] = 0,
    method: str = 'umap',
    metric: Union[str, Callable[[np.ndarray, np.ndarray], float]] = 'euclidean',
    metric_kwds: Mapping[str, Any] = {},
    copy: bool = False
) -> Optional[AnnData]:
    ...
```

However, `inspect.Signature`’s `__str__` method just combines it all [into one unreadable line](https://github.com/python/cpython/blob/d170e594b2736c579d12b96b129a8e58975a9c0a/Lib/inspect.py#L3072):

`Signature: sc.pp.neighbors(adata: anndata.base.AnnData, n_neighbors: int = 15, n_pcs: Union[int, NoneType] = None, use_rep: Union[str, NoneType] = None, knn: bool = True, random_state: Union[int, mtrand.RandomState, NoneType] = 0, method: str = 'umap', metric: Union[str, Callable[[numpy.ndarray, numpy.ndarray], float]] = 'euclidean', metric_kwds: Mapping[str, Any] = {}, copy: bool = False) -> Union[anndata.base.AnnData, NoneType]`

We should render the signature ourselves in a clearer fashion if it’s too long.

flying-sheep:
Fixes #11504

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -362,7 +362,7 @@ class Inspector(Colorable):
         If any exception is generated, None is returned instead and the
         exception is suppressed."""
         try:
-            hdef = self._render_signature(signature(obj), oname)
+            hdef = _render_signature(signature(obj), oname)
             return cast_unicode(hdef)
         except:
             return None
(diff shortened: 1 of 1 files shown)


IPython/core/oinspect.py
```diff
@@ -1016,3 +1016,47 @@ def psearch(self,pattern,ns_table,ns_search=[],
             search_result.update(tmp_res)
 
         page.page('\n'.join(sorted(search_result)))
+
+
+def render_signature(obj, oname):
```
minrk: `obj` is a signature, not the object whose signature we are rendering. Maybe give it a name like `obj_signature`

IPython/core/oinspect.py
```diff
@@ -1016,3 +1016,47 @@ def psearch(self,pattern,ns_table,ns_search=[],
...
+    Look there for the comments.
+    The only change is to add linebreaks when this gets too long.
+    """
+    if not isinstance(obj, inspect.Signature):
+        return oname + str(obj)
+
+    result = []
+    pos_only = False
+    kw_only = True
+    for param in self.parameters.values():
```
minrk: I assume `self` here should be the signature object

flying-sheep: yup, whoops.

IPython/core/oinspect.py
flying-sheep: done, thank you!

flying-sheep:
Do you know where the failures come from? I have no idea how they are related to my change…

flying-sheep:
@minrk I think this should be good to go, what do you think?

minrk:
Looks great, thanks @flying-sheep!

flying-sheep:
awesome!

PR #11506, opened 2018-11-21
Reenable jedi if it’s installed

### LINKED ISSUE #11503: Reenable jedi by default
8 months ago, in #11076, Jedi was disabled because an issue was still present that was in a new release “weeks later”. We should reenable it already!

flying-sheep:
Fixes #11503

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -576,9 +576,9 @@ class Completer(Configurable):
         """
     ).tag(config=True)
 
-    use_jedi = Bool(default_value=False,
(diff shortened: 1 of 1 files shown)


Carreau:
Looks good. Many thanks !

flying-sheep:
great!

dandavison:
Jedi completion doesn't seem to be working on python objects in various common scenarios, see https://github.com/ipython/ipython/issues/11609. Would it make sense to remove `jedi` from `install_requires` so that it's not on by default?

PR #11509, opened 2018-11-25
Add new exception-reporting mode and %tb mode option.

danielballan:
This adds a new mode to [`%xmode`](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-xmode) tentatively named `'Minimal'` that shows only the exception --- no traceback.

```python                                                                       
In [1]: %xmode minimal                                                          
Exception reporting mode: Minimal                                               
                                                                                
In [2]: def f():                                                                
    ...:     3/0                                                                
    ...:                                                                        
                                                                                
In [3]: f()                                                                     
```

In many contexts, you might say this leaves out too much information. Here is the context that I feel this will be useful in. Novice Python users are using IPython to execute rote procedures in a framework that happens to produce very deep tracebacks. They are completely lost in / overwhelmed by the full tracebacks and --- during the short time they grabble with IPython --- unable to absorb the rule, "Just scroll to the bottom." Therefore, an aggressively succinct `%xmode` setting may be less disorienting. Multiple experienced IPython users at my facility have suggested that this setting would be help elicit more useful bug reports.

Of course, if the problem is not immediately clear, it will be essential to access the full traceback. This could be done by switching to a more verbose `%xmode` and then using `%tb` to show the most recent traceback. In order to streamline that, this PR also adds an optional parameter to `%tb` that temporarily switches the exception-reporting mode for that single use.

```python
In [4]: %tb verbose                                                             
<ipython-input-3-c43e34e6d405> in <module>                                      
----> 1 f()                                                                     
        global f = <function f at 0x10e52de18>                                  
                                                                                
<ipython-input-2-cd4172d70b5b> in f()                                           
        1 def f():                                                               
----> 2     3/0                                                                 
        3                                                                       
                                                                                
In [5]: f()  # The default reporting mode has not been changed....              
```

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -580,7 +580,7 @@ class InteractiveShell(SingletonConfigurable):
     separate_out = SeparateUnicode('').tag(config=True)
     separate_out2 = SeparateUnicode('').tag(config=True)
     wildcards_case_sensitive = Bool(True).tag(config=True)
-    xmode = CaselessStrEnum(('Context','Plain', 'Verbose'),
+    xmode = CaselessStrEnum(('Context', 'Plain', 'Verbose', 'Minimal'),
                             default_value='Context',
                             help="Switch modes for the IPython exception handlers."
                             ).tag(config=True)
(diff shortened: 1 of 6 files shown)


Carreau:
I'm +1 in principle. Will review when I have time.

I'll also love to have an "exception" mimebundle that could be handled by the frontend with collapsible section and stuff.

danielballan:
I like that minebundle idea a lot. I am a little bothered by how much this omits (though it might be good to have as an option). Something interactive that shows the first couple and last couple with an expandable middle would be ideal.

Carreau (approved):

Carreau:
+1. I'm going to try to get a release soon.

danielballan:
Thanks!

Carreau:
(and 7.2 is out, currently writing announcements)

PR #11516, opened 2018-11-29
Don't expand user variables in execution magics

minrk:
Adds `@magic.no_var_expand` decorator to mark a magic as opting out of variable expansion.

Most useful for magics that take Python code on the line, e.g. %timeit, etc.

I applied it to the execution magics (timeit, time, debug, prun). I'm not sure if there are other magics that should get this treatment.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2273,10 +2273,14 @@ class InteractiveShell(SingletonConfigurable):
             # Note: this is the distance in the stack to the user's frame.
             # This will need to be updated if the internal calling logic gets
             # refactored, or else we'll be expanding the wrong variables.
-            
+
             # Determine stack_depth depending on where run_line_magic() has been called
             stack_depth = _stack_depth
-            magic_arg_s = self.var_expand(line, stack_depth)
+            if getattr(fn, magic.MAGIC_NO_VAR_EXPAND_ATTR, False):
+                # magic has opted out of var_expand
+                magic_arg_s = line
+            else:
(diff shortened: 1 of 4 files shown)


Carreau:
That's sounds reasonable to me. Does it has any reason to break user code ?

Carreau:
I'm going to leave that for 7.3 I'm worried about the consequences and would like it on master for a little longer. I'm going to do a 7.2 today. Hope that's ok with you.

minrk:
> Does it has any reason to break user code ?

I don't think so, but you never know. The one valid use case for variable expansion that I can think of is shell-style *args*, leaving the code block unparsed. e.g. this technically makes sense:

```python
dest = "myresult"
%timeit -o $dest some_python_code()
print(myresult)
```

But with this patch (and the magics implementation in general) variable expansion is all or nothing. We could try a deeper fix to try to *only* disable parsing of the code block, but that would require pretty deep changes, as the decision to implement var expansion would need to happen at a level where arguments are understood, violating the "magics are passed the line as a string" concept.


Another technically valid use could be building the code to time in a Python variable, and then running:

```python
code = "np.sin(x)"
%timeit {code}
```

I've never seen it before, but it might exist in the wild. In both cases, I think the tradeoff of never mangling the Python code is worth it.


> I'm going to leave that for 7.3 

👍 I had no intention of sneaking this under the wire, I just hadn't noticed that 7.2 was ready for release. 7.3 is perfect.

Carreau:
Sounds good, let's get that in.

Issue #11481, opened 2018-11-05
Why not have %pip do the right thing?


dsblank:
On the page:

https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-pip

it says:

```
Intercept usage of pip in IPython and direct user to run command outside of IPython.
```

But why? Couldn't it be this: .... (deleted hacky code) how about this:

https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/

Carreau:
Thanks for the question.  It's because "the right thing" is not always super obvious.

I did write https://pypi.org/project/pip_magic/ which does it, but one of the problem is how to make it work in the notebook. 
If pip ask for input then there is no way to "press" Y or N so the user process get stuck. It also let the user in a state where they believe that using pip from within IPython of Notebook "works". While it may work for installing missing modules it will not work for updating and a couple of other case. So we've settle on just giving an advice message. 

I'll be happy to experiment with other method but we're afraid it would be more disruptive than anything else

dsblank:
Instructing users how to install and/or upgrade is really painful in IPython (including notebook, lab, console, and otherwise). Of course, pip is painful in Python generally. But we could do a lot to make this much better. "pip" is a Python package, so we can import it, and decide what the right thing to do is. Or, even better, detect whether we are in anaconda and decide what the best course of action would be for a given package, complete with questions for the user.

I think a general "%install" magic would help so much in this space. But first, we need to understand the issue with pip. And, if it can't handle 100% of cases, that's fine too. We just need to give a message, and maybe point to a URL with more details.

Here is the start of an IPython Package Manifesto:

1. We need to recognize that packages are hard to install/upgrade in Jupyter because it is not clear what version of Python is running nor how to get to that environment from the shell. We need to recognize that this is a big, important problem.
2. We can get all of the information we need from the running environment to install/upgrade in anaconda and/or pip
3. We can make suggestions for the anaconda and pip Python libraries to make it better integrated with Jupyter/IPython. We don't have to solve this completely from one end.
4. We can put a layer over these APIs to make it much easier to manage the majority of packages. It could even: "I'm going to update the following package. Select Yes to continue." Can also handle `--upgrade-strategy` appropriately.
5. Should be a standard part of IPython/Jupyter (eventually). Doesn't have to start out that way, but should have the support of the entire organization.
[...]

PR #11524, opened 2018-12-06
Add %pip and %conda magic functions

jakevdp:
This PR adds ``%pip`` and ``%conda`` magic functions which automatically install packages into the currently-running kernel in an IPython or Jupyter notebook session.

screen shot 2018-12-05 at 9 41 51 pm

Why? Correctly installing Python packages within a running kernel has long been a point of confusion within the Jupyter ecosystem. While this might seem like an anti-pattern for the traditional local kernel use-case, the recent popularity of cloud-backed Jupyter notebook services has made this more relevant.

For some background on why ``!pip install`` or ``!conda install`` are incorrect in ways that can be very subtle and confusing, see my blog post from last year: [Installing Python Packages From a Jupyter Notebook](https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/).

There are some existing issues requesting this sort of feature (see #9517 & #11481) as well as some prior art on pip and conda magic function implementations by @minrk (https://github.com/minrk/condamagic/) and @Carreau (https://github.com/Carreau/pip_magic), but the fact that these third-party functions must be manually installed limits their usefulness to people who are having trouble installing things :smile:.

I'm hoping that including functionality like this within IPython itself will address the bulk of the confusion many of us have seen from users of the Jupyter stack.

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -66,6 +66,7 @@ class PackagingMagics(Magics):
           %pip install [pkgs]
         """
         self.shell.system(' '.join([sys.executable, '-m', 'pip', line]))
+        print("Note: you may need to restart the kernel to use updated packages.")
 
     @line_magic
     def conda(self, line):
@@ -99,3 +100,4 @@ class PackagingMagics(Magics):
             extra_args.extend(["--prefix", sys.prefix])
 
         self.shell.system(' '.join([conda, command] + extra_args + args))
+        print("\nNote: you may need to restart the kernel to use updated packages.")
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -1,0 +1,101 @@
+"""Implementation of packaging-related magic functions.
+"""
+#-----------------------------------------------------------------------------
+#  Copyright (c) 2018 The IPython Development Team.
+#
+#  Distributed under the terms of the Modified BSD License.
+#
+#  The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+import os
+import re
+import shlex
+import sys
+from subprocess import Popen, PIPE
+
+from IPython.core.magic import Magics, magics_class, line_magic
+
+
+def _is_conda_environment():
+    """Return True if the current Python executable is in a conda env"""
+    # TODO: does this need to change on windows?
+    conda_history = os.path.join(sys.prefix, 'conda-meta', 'history')
+    return os.path.exists(conda_history)
+
+
+def _get_conda_executable():
+    """Find the path to the conda executable"""
+    # Check if there is a conda executable in the same directory as the Python executable.
+    # This is the case within conda's root environment.
+    conda = os.path.join(os.path.dirname(sys.executable), 'conda')
+    if os.path.isfile(conda):
+        return conda
+
+    # Otherwise, attempt to extract the executable from conda history.
+    # This applies in any conda environment.
+    R = re.compile(r"^#\s*cmd:\s*(?P<command>.*conda)\s[create|install]")
+    for line in open(os.path.join(sys.prefix, 'conda-meta', 'history')):
+        match = R.match(line)
+        if match:
+            return match.groupdict()['command']
+    
+    # Fallback: assume conda is available on the system path.
+    return "conda"
+
+
+CONDA_COMMANDS_REQUIRING_PREFIX = {
+    'install', 'list', 'remove', 'uninstall', 'update', 'upgrade',
+}
+CONDA_COMMANDS_REQUIRING_YES = {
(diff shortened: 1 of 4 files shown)


jakevdp:
Needs some work on tests, but that can wait until there's some consensus on whether this feature should be added or not!

minrk:
I think this is a great thing to add, personally.

dsblank:
Yes! #11481 I'll take a look, and, going forward, will always try to help maintain keep refined. This is important.

dsblank:
Very clever on scanning the history, looking for the path! Never would have thought of that.

What will the advice be to users who don't know if a package can be install by conda or pip? I was originally thing a general `%install` would try conda first, and then pip if not available.

ccordoba12:
We're also having problems with people using `!pip` and `!conda` in Spyder, so this is great! Thanks @jakevdp!

Could it be possible to intercept the use of `!pip`/`!conda` and print a message saying to people to use the magics instead? Or does the magics addition do that automatically?

rgbkrk:
Thank you so much @jakevdp

IPython/core/magics/packaging.py
```diff
@@ -0,0 +1,101 @@
...
+import sys
+from subprocess import Popen, PIPE
+
+from IPython.core.magic import Magics, magics_class, line_magic
+
+
+def _is_conda_environment():
+    """Return True if the current Python executable is in a conda env"""
+    # TODO: does this need to change on windows?
+    conda_history = os.path.join(sys.prefix, 'conda-meta', 'history')
```
rgbkrk: @dhirschfeld would this correctly detect being in a conda environment?

IPython/core/magics/packaging.py
```diff
@@ -0,0 +1,101 @@
...
+    
+    # Fallback: assume conda is available on the system path.
+    return "conda"
+
+
+CONDA_COMMANDS_REQUIRING_PREFIX = {
+    'install', 'list', 'remove', 'uninstall', 'update', 'upgrade',
+}
+CONDA_COMMANDS_REQUIRING_YES = {
+    'install', 'remove', 'uninstall', 'update', 'upgrade',
```
rgbkrk: Love this organization around args needing `--y` or the `--prefix`

IPython/core/magics/packaging.py
MSeal: A better way might be to check if `Anaconda` is in `sys.version`. That should work cross-platform.

IPython/core/magics/packaging.py
```diff
@@ -0,0 +1,101 @@
...
+    """Magics related to packaging & installation"""
+
+    @line_magic
+    def pip(self, line):
+        """Run the pip package manager within the current kernel.
+
+        Usage:
+          %pip install [pkgs]
+        """
+        self.shell.system(' '.join([sys.executable, '-m', 'pip', line]))
```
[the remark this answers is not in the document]
Carreau: +1 We likely want to remind user they might need to restart the kernel.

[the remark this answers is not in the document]
jakevdp: Done in 594d000

dhirschfeld:
There *might* be an issue on Windows where you can silently corrupt your installation by trying to update a compiled extension when Windows is holding a reference to the DLL. Certainly, with pandas that used to often happen. I still kill all Python processes before doing any upgrades, but maybe that's no longer required and conda has worked around the original issue.

I'll try to check the current behaviour and report back...

rgbkrk:
@Carreau @mseal `self.shell.system` is one of the calls that doesn't error when a shell command happens right? We should make sure this causes an error so that use of this magic will stop execution in nbconvert, papermill, and other run-all use cases.

IPython/core/magics/packaging.py
MSeal: Ahh yes your right

MSeal:
I need to dig into that but yes one of the function magic options isn't erroring as one would expect.

Carreau:
I guess we could also use `subprocess.check_call` or something. I didn't even remembered we had `shell.system`

Carreau:
I fixed the tests. 
Let's get that in. I'll try to make a release close to end of month to keep the monthly release cadence.

ceceshao1:
Can't wait for this! Was just experiencing confusion today around this specific action

dsblank:
@Carreau Time for a 7.3 release?

dsblank:
@Carreau Just saw your note. What "steps remaining" do you mean?

Carreau:
Basically a todo-list, with list of stem necessary before/during/after the release and list of issues /changes that need to be added in the what's new. 
I did it yesterday (see #11571). I unfortunately risk to not be able to do any of that before this week-end, so any help welcome.

IPython/core/magics/packaging.py
mpacer: It really is gloriously nice.

PR #11525, opened 2018-12-06
Fix signature rendering

flying-sheep:
I’m so sorry, I thought I had this properly tested!

Here I fix a bug that I introduced in a last-minute change in #11505 and add tests (which I should have done in the first place)


Carreau:
No worries, noone else complained.

flying-sheep:
Only because I’m the only one who was eager to check it out *right* after 7.2 came out :sweat_smile:

PR #11528, opened 2018-12-12
Fix #11527 initialize posix aliases after a %reset

### LINKED ISSUE #11527: %clear doesn't work after %reset
**Problem:** Normally, when I use `%clear` on IPython it works perfectly, clears the screen, but when I use `%reset` to reset my IPython shell, and then I try to use `%clear` it says:

> UsageError: Line magic function `%clear` not found.

Also, before using the `%reset` command, when I type `%` and press tab for auto completion, I can see `%clear` is one of the suggestions, but after the `%reset` command, `%clear` is not in the suggestions anymore..

**System info:**

```
{'commit_hash': '<not found>',
 'commit_source': '(none found)',
 'default_encoding': 'UTF-8',
 'ipython_path': '/usr/lib/python3.7/site-packages/IPython',
 'ipython_version': '7.1.1',
 'os_name': 'posix',
 'platform': 'Linux-4.19.8-arch1-1-ARCH-x86_64-with-arch-Arch-Linux',
 'sys_executable': '/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.7.1 (default, Oct 22 2018, 10:41:28) \n[GCC 8.2.1 20180831]'}
```

jwidner:
Changes the behavior of %reset by initializing the posix aliases `clear`, `less`, `more`, and `man` during a reset. It's not clear to me what the intended behavior should be, but it fixes #11527.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1433,6 +1433,13 @@ class InteractiveShell(SingletonConfigurable):
         self.alias_manager.clear_aliases()
         self.alias_manager.init_aliases()
 
+        # Now define aliases that only make sense on the terminal, because they
+        # need direct access to the console in a way that we can't emulate in
+        # GUI or web frontend
+        if os.name == 'posix':
+            for cmd in ('clear', 'more', 'less', 'man'):
+                self.alias_manager.soft_define_alias(cmd, cmd)
+
         # Flush the private list of module references kept for script
         # execution protection
         self.clear_main_mod_cache()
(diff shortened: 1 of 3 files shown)


Carreau:
That seem reasonable to me. Thanks.

red8012:
It seems that the change causes the following message being printed whenever a reset is done.
```
ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.
ERROR:root:Invalid alias: The name more can't be aliased because it is another magic command.
ERROR:root:Invalid alias: The name less can't be aliased because it is another magic command.
ERROR:root:Invalid alias: The name man can't be aliased because it is another magic command.
```

See https://stackoverflow.com/questions/54971047/spyder-reports-invalid-alias-when-running-any-script. The issue happens not only on Spyder but also on Jupyter notebook.

Carreau:
Can you please open a new issue for this ? Thanks.

PR #11529, opened 2018-12-12
Allow the IPython command line to run *.ipynb files

tonyfast:
This PR allows _exec_file in the shell app to pass ipynb files to safe_execfile_ipy in the interactiveshell. safe_execfile_ipy already has the machinery to consume notebooks.

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -312,7 +312,7 @@ class InteractiveShellApp(Configurable):
                 # behavior.
                 with preserve_keys(self.shell.user_ns, '__file__'):
                     self.shell.user_ns['__file__'] = fname
-                    if full_filename.endswith('.ipy'):
+                    if full_filename.endswith('.ipy') or full_filename.endswith('.ipynb'):
                         self.shell.safe_execfile_ipy(full_filename,
                                                      shell_futures=shell_futures)
                     else:


tonyfast:
This approach will only work for an IPython interactive shell.  The run_cell method is invoked so all of the interactive shell rules apply.  

The current way of running notebooks is using `nbconvert` with the `--execute` flag.  The [__runipy__](https://github.com/paulgb/runipy) project is deprecated, and recommends the `nbconvert` approach.  Some other ways to acheive a similar result is with [__importnb__](https://github.com/deathbeds/importnb) or [__papermill__](https://github.com/nteract/papermill/).

Carreau:
Thanks ! Looks good.

IPython/core/shellapp.py
```diff
@@ -312,7 +312,7 @@ def _exec_file(self, fname, shell_futures=False):
                 # behavior.
                 with preserve_keys(self.shell.user_ns, '__file__'):
                     self.shell.user_ns['__file__'] = fname
-                    if full_filename.endswith('.ipy'):
```
Carreau: Tip: endswith support multiple values, so you could do `full_filename.endswith(('.ipy', 'ipynb', '.py'))`...

tonyfast: 👍

PR #11542, opened 2019-01-04
Apply @needs_local_scope to cell magics.

Carreau:
While technically this will not be completely needed (at cell magics
cannot be in nested scope), it make the API a tiny bit more consistent
between line and cell magics.

(Bug report I recieved personally that will be submitted later)


takluyver:
Did you get round to filing the bug report about this?

Carreau:
No, I need to translate it from french. I'm slowly coming back the cluster was under maintnance for the last few days so it was taking all my time.

PR #11545, opened 2019-01-06
Fix ResourceWarning: unclosed file

BoboTiG:
Also uniformize usage of the `with` context manager to prevent resource leaks.

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


Carreau:
Thanks, that looks good to me.

PR #11564, opened 2019-01-17
More readable .n, .s, .p reference

techtonik:


docs/source/interactive/shell.rst
```diff
@@ -186,13 +186,11 @@ First, capture output of "hg status"::
 
 Now we can just remove these files by doing 'rm $junk.s'.
 
-The .s, .n, .p properties
+The .n, .s, .p properties
 -------------------------
```
takluyver: It would be good to have a summary explaining that these provide convenient ways to use the text in different formats.

docs/source/interactive/shell.rst
```diff
@@ -186,13 +186,11 @@ First, capture output of "hg status"::
...
-property return one string where the lines are separated by a newline
-(i.e. the original output of the function). If the items in string
-list are file names, ``.p`` can be used to get a list of "path" objects
-for convenient file manipulation.
-
+* ``.n`` returns (original) string with lines separated by a newline
+* ``.s`` returns string with lines separated by single space (for
+  convenient passing to system commands)
+* ``.p`` if string contains file names, returns list of "path" objects
+  for convenient file manipulation
```
takluyver: this last one doesn't quite read right

docs/source/interactive/shell.rst
techtonik: Right. Added.

techtonik: What are all places where SList object is returned from? `returned by execution operator !` is one place, but are there other uses?

techtonik: I'd like to add reference to SList API here.

takluyver: That's the only one I know of; you can search the codebase to see if it's used elsewhere.

Good idea to link to the reference.

docs/source/interactive/shell.rst
techtonik: Reworded.

Carreau:
Looks good, we can refine later if needed. Thanks !

PR #11591, opened 2019-02-04
Make all_completions use new completion code

### LINKED ISSUE #11541: all_completions returns nothing (simple fix suggested)
The `Completer.all_completions` method, which emacs python.el and elpy mode use, is returning nothing defined in the current scope. For example:

```
In [87]: xxx = 2

In [91]: get_ipython().Completer.all_completions('x')
Out[91]: ['%xdel', '%xmode']

In [92]: get_ipython().Completer.all_completions('xx')
Out[92]: []
```
Internally it's using jedi but discarding jedi completions. Now, `all_completions` calls the deprecated `complete` instead of `completions`. A simple change fixes both the usage of a deprecated method and a bug:

```
In [127]: list(map(lambda c: c.text, get_ipython().Completer.completions('x', len('x'))))
Out[127]: ['xxx', '%xdel', '%xmode']
```

Why not redefining `all_completions` like that? At least the following is working for me inside an emacs inferior python session:

```
get_ipython().Completer.all_completions = lambda t: map(lambda c: c.text, get_ipython().Completer.completions(t, len(t)))
```

memeplex:
This fixes a bug affecting emacs users introduced by 7.2.

Fixes https://github.com/ipython/ipython/issues/11541

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1134,13 +1134,14 @@ class IPCompleter(Completer):
                 self.dict_key_matches,
             ]
 
-    def all_completions(self, text):
+    def all_completions(self, text) -> List[str]:
         """
         Wrapper around the completions method for the benefit of emacs.
         """
         prefix = text[:text.rfind(".") + 1]
-        return map(lambda c: prefix + c.text,
-                   self.completions(text, len(text)))
+        with provisionalcompleter():
+            return list(map(lambda c: prefix + c.text,
+                   self.completions(text, len(text))))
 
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, I added one commit : 

```
   Fix api compatibility.

    Wrap into the with provisionalCompleter() and return a list instead of
    an iterator just in case.

    Also add types hints.
```

Sounds good otherwise.

memeplex:
I think this can be marged.

Carreau:
Thanks !

PR #11600, opened 2019-02-08
Fix 11559 weird regexp writefile

ppmfloss:
Hi, This is a fix for 11559 Weird regexp ['*']. I have fixed the regexp and added 2 test cases to handle embedded quotes in filenames in magic writefile. I have run regression testing locally without any issue.


Carreau:
Thanks !

PR #11601, opened 2019-02-08
Fix #11573 add docs for running tests

### LINKED ISSUE #11573: testing ipython documentation
I tried to run the test suite manually as suggested to help sort out my PR https://github.com/ipython/ipython/pull/11572, but there is nowhere a word at how to do it, at least I couldn't find any.

Through digging I found I could get the tests running with:

```
cd ipython
pip install -e ".[test]"
mkdir results
cd results
iptest
```
Simply replicated `appveyor.yml`

To run groups of tests, e.g. core:

```
iptest core
```

but I can't figure out how to run an individual test, e.g. How do I run: `core.tests.test_iplib.doctest_tb_verbose` which is failing? If there is one of course - rerunning the whole group is quite inefficient, especially if several tests fail and I need to the see the logs for one of them.

Can this information on how to run tests please be added to CONTRIBUTING.md?

Also it looks like the doc here is outdated: https://github.com/ipython/ipython/blob/320d21bf56804541b27deb488871e488eb96929f/IPython/testing/iptest.py

> 2. With the regular nose syntax, like `iptest -vvs IPython`.  In this form
>   the script simply calls nose, but with special command line flags and
>   plugins loaded.

which doesn't work. `iptest: error: unrecognized arguments: -vvs`

Thanks.

ppmfloss:
Closes #11573

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,3 +72,24 @@ particularly for PRs that affect `IPython.parallel` or Windows.
 
 For more detailed information, see our [GitHub Workflow](https://github.com/ipython/ipython/wiki/Dev:-GitHub-workflow).
 
+## Running Tests
+
+All the tests can by run by using
+```shell
+python3 IPython/testing/iptest.py IPython/core/tests
+```
+
+All the tests for a single file (for example **test_alias.py**) can be run by using
+```shell
+python3 IPython/testing/iptest.py IPython/core/tests/test_alias.py
+```
+
+Only a single test (for example **test_alias_lifecycle**) within a single file can be run by using
+```shell
+python3 IPython/testing/iptest.py IPython/core/tests/test_alias:test_alias_lifecycle
+```
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -72,3 +72,24 @@ particularly for PRs that affect `IPython.parallel` or Windows.
 
 For more detailed information, see our [GitHub Workflow](https://github.com/ipython/ipython/wiki/Dev:-GitHub-workflow).
 
+## Running Tests
+
+All the tests can by run by using
+```shell
+python3 IPython/testing/iptest.py IPython/core/tests
```
Carreau: I believe you ned to use the iptest entry point that will call iptestcontroller.py otherwise some setup might not be done correctly on all systems. I have to check thoughwhat the exact difference is.

ppmfloss: On Debian the above command works as expected.

Issue #11612, opened 2019-02-20
PyQt5 event loop fails for PyQt >5.11 with No module named 'sip'


flutefreak7:
This is effectively a duplicate of a qtconsole issue: https://github.com/jupyter/qtconsole/issues/291

When I attempt to run a cell with `%gui qt` I get the following:

    Traceback (most recent call last):
      File "c:\python37\lib\site-packages\tornado\ioloop.py", line 758, in _run_callback
        ret = callback()
      File "c:\python37\lib\site-packages\tornado\stack_context.py", line 300, in null_wrapper
        return fn(*args, **kwargs)
      File "c:\python37\lib\site-packages\ipykernel\kernelbase.py", line 306, in advance_eventloop
        eventloop(self)
      File "c:\python37\lib\site-packages\ipykernel\eventloops.py", line 129, in loop_qt5
        return loop_qt4(kernel)
      File "c:\python37\lib\site-packages\ipykernel\eventloops.py", line 116, in loop_qt4
        kernel.app = get_app_qt4([" "])
      File "c:\python37\lib\site-packages\IPython\lib\guisupport.py", line 114, in get_app_qt4
        from IPython.external.qt_for_kernel import QtGui
      File "c:\python37\lib\site-packages\IPython\external\qt_for_kernel.py", line 95, in <module>
        QtCore, QtGui, QtSvg, QT_API = load_qt(api_opts)
      File "c:\python37\lib\site-packages\IPython\external\qt_loaders.py", line 309, in load_qt
        result = loaders[api]()
      File "c:\python37\lib\site-packages\IPython\external\qt_loaders.py", line 220, in import_pyqt5
        import sip
    ModuleNotFoundError: No module named 'sip'

As noted in the qtconsole issue, this is due to a backwards incompatibility: http://pyqt.sourceforge.net/Docs/PyQt5/incompatibilities.html#pyqt-v5-11

The qtconsole PR is here: https://github.com/jupyter/qtconsole/issues/291

flutefreak7:
This was my first PR ever! Woot!

QuLogic:
Shouldn't this be closed by the above PR?

ccordoba12:
Sure, it seems this was solved by PR #11613.

PR #11613, opened 2019-02-20
Fixed PyQt 5.11 backwards incompatibility causing sip import failure

flutefreak7:
See issue: #11612 

This is the same fix as implemented in https://github.com/jupyter/qtconsole/issues/319.

### DIFF
--- a/IPython/external/qt_loaders.py
+++ b/IPython/external/qt_loaders.py
@@ -220,10 +220,6 @@ def import_pyqt5():
 
     from PyQt5 import QtCore, QtSvg, QtWidgets, QtGui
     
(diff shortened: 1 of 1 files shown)


tacaswell:
Could the `sip` import simply be deleted?  It is not returned from the function and you can not change the API version on pyqt5.

flutefreak7:
That's a good point... it should probably just be removed since it doesn't get used and is likely a leftover.

Removing it works on my end, so I'll do that for the PR.

tacaswell (approved):

Carreau:
Sounds good. Thanks !

Issue #11588, opened 2019-02-02
autoreload magic doesn't reload objects


stas00:
**update**: the problem appears to be with any object, unrelated to pickle. pickle was just exposing it in a forceful way. Skip to the latest diagnostic https://github.com/ipython/ipython/issues/11588#issuecomment-464949800

when autoreload reloads a python module, it appears to not be fully reloading the already instantiated objects, and pickle fails to pickle them with:

`Can't pickle <class 'mytestclass.Test'>: it's not the same object as mytestclass.Test`

Here is a notebook to reproduce, must be run in 3 parts (or 3 cells):

```
# cell1
%reload_ext autoreload
%autoreload 2

import pickle

file = "mytestclass.py"

text = """
class Test():
    def __init__(self):
        self.ok = 1
"""
with open("mytestclass.py","w") as f: f.write(text)

from mytestclass import *
test = Test()

# works
p = pickle.dumps(test)
```
```
# cell2
# touch forces autoreloader to reload the library file
!echo touch $file
!touch $file
```
```
# cell3
# now it fails
p = pickle.dumps(test)
```

The error is:

```
In [6]: p = pickle.dumps(test)
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
<ipython-input-6-740c326b650e> in <module>()
----> 1 p = pickle.dumps(test)

PicklingError: Can't pickle <class 'mytestclass.Test'>: it's not the same object as mytestclass.Test

```

Please note that it should be run in 3 parts - if it's run in one go the reload effect won't be activated. Can be run in jupyter notebook or directly in ipython shell.

I traced why pickle fails and expanded on it in this issue https://github.com/fastai/fastai/issues/1493 - tldr; it compares memory addresses of the object's class and the class in `sys.modules` - if their memory addresses don't match it refuses to pickle.

I think there are other issues with the incomplete reload, (i.e. object methods disappearing after reload), but I'm yet to make reproducible cases for those. I believe pickle is just exposing a deeper problem.

I tried the git master ipython and went all the way back to ipython-6.0.0 - the problem is the same.

And yes, I read the caveats section, that does say that reload isn't 100% perfect.

Thank you.

Carreau:
Thanks, I think it would make sens to try to fix this behavior and have autoreload not break pickle. I'm unsure how easy it would be.

stas00:
So it looks like autoreload doesn't quite update the existing objects to the reloaded classes, here is for example how one could hack it into working:

```
[...]

PR #11644, opened 2019-03-11
allow type of already existing class instances to be updated

Naikless:
This is my first attempt in solving the problematic discussed in https://github.com/ipython/ipython/issues/11588. Although `%autoreload` updates class attributes for reloaded modules "on the fly", thus providing these updated attributes also to previously instantiated objects of that class, these old objects remain of type "old class definition", meaning that they can't be pickled. 

(This is also my first open source contribution in general, so be gentle with me :) )

As you can see, this relies on identifying the execution frame from the stack that holds all user defined variables. My current search conditions were inspired by the definition of `%who_ls`, but even though it worked for manual tests (creating a module with a class, creating an instance of that class, change the definition inside the module, compare type of already existing instance with new ones), I couldn't get it to work in the FakeShell environment of the tests in test_autoreload. So maybe there is a more general/fail proof possibility to find the user namespace?

EDIT: The tests fail mainly because `FakeShell` doesn't have a `user_ns`. Adding an empty namespace lets the tests succeed, but the functionality is still lost.

### DIFF
--- a/IPython/extensions/autoreload.py
+++ b/IPython/extensions/autoreload.py
@@ -323,7 +323,6 @@ def update_instances(old, new, objects=None, visited={}):
 def update_class(old, new):
     """Replace stuff in the __dict__ of a class, and upgrade
     method code objects, and add new methods, if any"""
-    print('old is', id(old))
     for key in list(old.__dict__.keys()):
         old_obj = getattr(old, key)
         try:
--- a/IPython/extensions/autoreload.py
+++ b/IPython/extensions/autoreload.py
@@ -268,6 +268,60 @@ def update_function(old, new):
             pass
 
 
+def _find_instances(old_type):
+    """Try to find all instances of a class that need updating.
+
+    Classic graph exploration, we want to avoid re-visiting object multiple times.
+    """
+    # find ipython workspace stack frame, this is just to bootstrap where we
+    # find the object that need updating.
+    frame = next(frame_nfo.frame for frame_nfo in inspect.stack()
+                 if 'trigger' in frame_nfo.function)
+    # build generator for non-private variable values from workspace
+    shell = frame.f_locals['self'].shell
+    user_ns = shell.user_ns
+    user_ns_hidden = shell.user_ns_hidden
+    nonmatching = object()
+    objects = ( value for key, value in user_ns.items()
+            if not key.startswith('_')
+            and (value is not user_ns_hidden.get(key, nonmatching))
+            and not inspect.ismodule(value))
+
+    # note:  in the following we do use dict as object might not be hashable.
+    # list of objects we found that will need an update.
+    to_update = {}
+
+    # list of object we have not recursed into yet
+    open_set = {}
+
+    # list of object we have visited already
+    closed_set = {}
+
+    open_set.update({id(o):o for o in objects})
+
+    it = 0
+    while len(open_set) > 0:
+        it += 1
+        if it > 100_000:
+            raise ValueError('infinite')
+        (current_id,current) =  next(iter(open_set.items()))
+        if type(current) is old_type:
+            to_update[current_id] = current
+        if hasattr(current, '__dict__') and not (inspect.isfunction(current)
+                                                 or inspect.ismethod(current)):
+            potential_new = {id(o):o for o in current.__dict__.values() if id(o) not in closed_set.keys()}
+            open_set.update(potential_new)
+        # if object is a container, search it
+        if hasattr(current, 'items') or (hasattr(current, '__contains__')
+                                         and not isinstance(current, str)):
+            potential_new = (value for key, value in current.items()
+                       if not str(key).startswith('_')
+                       and not inspect.ismodule(value) and not id(value) in closed_set.keys())
+            open_set.update(potential_new)
+        del open_set[id(current)]
+        closed_set[id(current)] = current
+    return to_update.values()
+
 def update_instances(old, new, objects=None):
     """Iterate through objects recursively, searching for instances of old and
     replace their __class__ reference with new. If no objects are given, start 
@@ -319,6 +373,7 @@ def update_instances(old, new, objects=None):
 def update_class(old, new):
     """Replace stuff in the __dict__ of a class, and upgrade
     method code objects, and add new methods, if any"""
+    print('old is', old)
     for key in list(old.__dict__.keys()):
         old_obj = getattr(old, key)
         try:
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks for your contribution, and congrats, I know it is a sometime scary step ! 
You did (almost) everything right; you'll just see people ask you to not send the PR from `master` of your fork, but use a feature-branch. Though that's a detail and it does not matter for now. 

I'll try to come back to review this as soon as possible, though I'm quite busy this week, so I can't promise. 

From a quick read that looks great, I just need to understand it, and find out how to fix the tests. 

Thanks !

Carreau:
Starting to look into this.

IPython/extensions/autoreload.py
```diff
@@ -267,6 +268,54 @@ def update_function(old, new):
...
+
+            # if object is instance of other class, look for nested instances
+            if hasattr(obj, '__dict__') and not (inspect.isfunction(obj)
+                                                 or inspect.ismethod(obj)):
+                update_instances(old, new, obj.__dict__)
+
+            # if object is a container, search it
+            if hasattr(obj, 'items') or (hasattr(obj, '__contains__')
+                                         and not isinstance(obj, str)):
+                update_instances(old, new, obj)
```
Carreau: So I get an infinite recursion somewhere here. I'm going to attempt a refactor with a 2 step: search and update.

Carreau:
Would you mind having a look at this new version ? I proposed an alternative with less recursion, which appear to work on my machine. 

Test and docs are still needed though, and code need to be cleaned up.

Carreau:
Hey ! Basic test seem to pass. Sounds like a good start. 

@daharn how familiar are you with git/github ? Will you manage to get my changes on your machine, and can take care of (some-of) the cleaning documenting ? Or do you need some extra guidance ?

daharn:
Thanks for looking into this! 
I am fairly familiar with git although I haven't used github much. Getting your changes shouldn't be a problem. I will let you know once I got around to look into them.

Carreau:
Thanks ! Much appreciated.

IPython/extensions/autoreload.py
```diff
@@ -268,6 +268,60 @@ def update_function(old, new):
...
+        if type(current) is old_type:
+            to_update[current_id] = current
+        if hasattr(current, '__dict__') and not (inspect.isfunction(current)
+                                                 or inspect.ismethod(current)):
+            potential_new = {id(o):o for o in current.__dict__.values() if id(o) not in closed_set.keys()}
+            open_set.update(potential_new)
+        # if object is a container, search it
+        if hasattr(current, 'items') or (hasattr(current, '__contains__')
+                                         and not isinstance(current, str)):
+            potential_new = (value for key, value in current.items()
```
daharn: Just started looking into this: This line won't work for sequences, only for dicts.

daharn:
Next attempt. I kind of stuck to my original recursion but tried to include your suggested checks to avoid revisiting of already browsed objects.  

This passes the tests and does a good job on updating all instance references in the user namespace (a notable exception are instances yielded from a generator. I don't know whether it would even be possible to update them).

Let me know what you think and if you have an idea on how to add some tests for this feature in the autoreload tests.

IPython/extensions/autoreload.py
daharn: I have implemented some measures to avoid infinite recursion in my latest commit.

Carreau:
Thanks for this, I'll try to look at it soon.

Carreau:
I'm still planning to have a look. Apologies for the delay

daharn:
No worries, we are all doing this in our spare time I guess.

Carreau:
I wrote and push test for this functionality. 

You can run `cd ; iptest IPython.extensions.tests.test_autoreload:TestAutoreload.test_reload_class_type; cd -` to test just the behavior that the `__class__` has been updated.

Carreau:
Well, took little over 3 month; so thanks for your patience; but that looks god to me; I removed one more debug statement and will wait for tests to pass then merge. I need to remember to update the what's new to give an example of the new possibility.

daharn:
Thanks for including this. Keep up the great work!

Carreau:
Can you check #11849 for your use case ? I'm hoping this will still work properly and potentially be faster.

daharn:
I will as soon as I am able to.

If this iterating through workspace turns out to be unsuited in general, a new (and arguably more elegant approach) would be to maintain a list of all objects for any custom module and add each object to it, when it is created. 

Maybe this could be done by messing with the definition of __new()__...?

Carreau:
I think there might be a better way. 

Usually `obj.__class__` point to the class of the object, to technically all instances of `Foo` are in `gc.get_referers(Foo)`, so we _should_ be able to just check in `gc.get_referers(OldClass)`. I"d like to avoid messing with `__new__` as much as I can.

wmmxk:
The %load_ext autoreload magic is very handy when I work in a dual development mode: Jupyter+IDE.
I need to add "%load_ext autoreload" at the beginning of each jupyter notebook. Is there a way to set this configuration permanently? I know there is "~/.ipython/profile_default" folder, so I think it is possible but I am not sure.

Carreau:
See [the documentation, and this specific option](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-InteractiveShellApp.extensions)

daharn:
> I think there might be a better way.
> 
> Usually `obj.__class__` point to the class of the object, to technically all instances of `Foo` are in `gc.get_referers(Foo)`, so we _should_ be able to just check in `gc.get_referers(OldClass)`. I"d like to avoid messing with `__new__` as much as I can.

I have tinkered with `gc.get_referers()` a little and this seems to be by far the cleanest solution. Any reason against implementing it this way? Apart from routinely evoking a function of which the documentation says _"Avoid using get_referrers() for any purpose other than debugging."_?

Carreau:
> I have tinkered with `gc.get_referers()` a little and this seems to be by far the cleanest solution. Any reason against implementing it this way?

Not particularly; I was just conservative in case `get_referrers` did not get all objects; but if it works I don't see any reason not to use this one. Mostly I don't want to break your use case; but if that works for you I'll likely update this to use `gc.get_referers(...)`

daharn:
I just tried and it seems to work like charm, at least for small test cases. You can have a look at https://github.com/ipython/ipython/pull/11876. Would be good if you could test it in an environment where the old implementation took long.

daharn:
I gave it a try and it works quite well. Don't know if `gc.get_referrers()` has a performance advantage compared to `gc.get_objects()`, although the amount of objects to actively crawl is far less. Maybe you can test this in one of the more workspace demanding environments: https://github.com/ipython/ipython/pull/11876

PR #11650, opened 2019-03-15
Add 'normalize' argument to Audio to make normalization optional

### LINKED ISSUE #8608: Volume normalization in IPython.display.Audio should be optional

matangover:
Fixes #8608. Based on #11161 with changes as discussed (see [comment](https://github.com/ipython/ipython/pull/11161#issuecomment-470323709)).
It might be easier to review each commit separately in this PR.

While working on the fix, I encountered and fixed the following small bugs:
* Audio from list of samples, without numpy installed, would fail with ugly exception on Python 3.
* When rate wasn't specified but audio data was given, error was not helpful.
* Error when data was not found said 'image' instead of 'audio'.

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -97,7 +97,7 @@ class Audio(DisplayObject):
         super(Audio, self).__init__(data=data, url=url, filename=filename)
 
         if self.data is not None and not isinstance(self.data, bytes):
-            self.data = self._make_wav(data,rate)
+            self.data = Audio._make_wav(data, rate)
 
     def reload(self):
         """Reload the raw data from file or URL."""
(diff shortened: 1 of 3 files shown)


Carreau:
That looks good to me, thanks ! Apologies for the delay in reviewing, and welcome to the party !

matangover:
Great :) thanks!

PR #11663, opened 2019-03-23
Verbose error message when displaying a nonexistent image.

ndtallant:
The error is now a FileNotFoundError - the same as the Image class when the filename does not exist.


Carreau:
I'm not a huge fan, I believe we should not even call this function if we know it's a filename and not binary; but I see how the code does not make it easy to refactor. 

So sounds good; and we should put that on "technical debt for now". 

Thnaks !

PR #11671, opened 2019-04-03
Add a -q parameter to %rerun magic

MatthewWilkes:
This allows the rerun magic to be run in quiet mode, which suppresses
the `=== Executing: ===` and `=== Output: ===` headings as well as the
text of the cell being re-run. This means that cells that print can
optionally be re-run as part of a larger control structure like a loop
without IPython-specific noise.

### DIFF
--- a/IPython/core/magics/history.py
+++ b/IPython/core/magics/history.py
@@ -289,8 +289,12 @@ class HistoryMagics(Magics):
           current command.
 
           -g foo : Repeat the most recent line which contains foo
+
+          -q     : Quiet mode: do not add the code being executed or any headings
+          to the output, output only what the original cell would have.
+
         """
-        opts, args = self.parse_options(parameter_s, 'l:g:', mode='string')
+        opts, args = self.parse_options(parameter_s, 'l:g:q', mode='string')
         if "l" in opts:         # Last n lines
             n = int(opts['l'])
             hist = self.shell.history_manager.get_tail(n)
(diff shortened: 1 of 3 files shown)


IPython/core/magics/history.py
```diff
@@ -289,8 +289,10 @@ def rerun(self, parameter_s=''):
           current command.
 
           -g foo : Repeat the most recent line which contains foo
+
+          -q     : Quiet mode: do not echo the cell contents back
```
havocbane: I like this change, but could you make it clearer in the help docs that `quiet` means to suppress the history lines specifically? A user will still see any output from the command, right? "Quiet" usually implies that all regular output will be suppressed (i.e. non-errors) or maybe just lower verbosity.

IPython/core/tests/test_history.py
```diff
@@ -168,6 +169,28 @@ def test_magic_rerun():
...
+def test_magic_rerun_quiet():
+    """Simple test for %rerun -q (quiet, no echoing of output)"""
+    ip = get_ipython()
+    ip.run_cell("a = 10", store_history=True)
+    ip.run_cell("a += 1", store_history=True)
+    nt.assert_equal(ip.user_ns["a"], 11)
+    with capture_output() as cap:
+        ip.run_cell("%rerun -q", store_history=True)
+    nt.assert_equal(ip.user_ns["a"], 12)
+    nt.assert_not_in("a += 1", cap.stdout)
```
havocbane: Nice work on the tests!

IPython/core/magics/history.py
MatthewWilkes: Documentation tone can often be the hardest part of a contribution. I've reworded this, but let me know if it's still not quite right, or feel free to specify wording yourself if you'd like.

IPython/core/tests/test_history.py
MatthewWilkes: Thanks, nice work on the support functions like `capture_output`! ;)

MatthewWilkes:
I'm struggling to fix the sphinx build issue on Py3.7 here. I can't seem to build the docs locally, but I tried adding a newline to the options section as the output suggested to no effect.

PR #11672, opened 2019-04-04
Added support to list all available object types for the magic function "%psearch".

k-zen:
These commits add the functionality to list available object types for the **%psearch** magic. The reason for this modification is the need to know which objects are available for filtering. The documentation is also outdated and commands like:
`%psearch a* function` produce no results.

The correct command would be:
`%psearch a* builtinmethod`

Usage of new function:
`%psearch -l`, produces the following output in Python 3.6.8:

> asyncgenerator
> builtinfunction
> builtinmethod
> code
> coroutine
> frame
> function
> generator
> getsetdescriptor
> lambda
> mappingproxy
> memberdescriptor
> method
> module
> traceback

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ __pycache__
 .vscode
 .pytest_cache
 .python-version
+venv*/
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ __pycache__
 .pytest_cache
 .python-version
 venv*/
+.idea/
--- a/IPython/core/magics/namespace.py
+++ b/IPython/core/magics/namespace.py
@@ -173,6 +173,9 @@ class NamespaceMagics(Magics):
           'builtin', 'user', 'user_global','internal', 'alias', where
           'builtin' and 'user' are the search defaults.  Note that you should
           not use quotes when specifying namespaces.
+          
+          -l: List all available object types for object matching. This function
+          can be used without arguments.
 
           'Builtin' contains the python module builtin, 'user' contains all
           user data, 'alias' only contain the shell aliases and no python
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -27,3 +27,5 @@ __pycache__
 .vscode
 .pytest_cache
 .python-version
+venv*/
+.idea/
```
Carreau: Tip: you can put this in your system global gitignore.

Carreau:
That seem reasonable. Let me wrap my head around that as I haven't used psearch in a while. 

But otherwise that seem reasonable.

IPython/core/magics/namespace.py
```diff
@@ -200,6 +203,10 @@ def psearch(self, parameter_s=''):
         Show objects beginning with a single _::
 
           %psearch -a _*         list objects beginning with a single underscore
+          
+        List available objects::
+        
+          %psearch -l            list all available object types
         """
```
Carreau: Maybe dumb question, as you are the one having the use case; would it make more sens to actually add the list of `typestr2type` to the docstring ?

IPython/core/oinspect.py
```diff
@@ -962,7 +963,7 @@ def _source_contains_docstring(src, doc):
             return False
 
     def psearch(self,pattern,ns_table,ns_search=[],
-                ignore_case=False,show_all=False):
+                ignore_case=False,show_all=False, *, list_types=False):
```
Carreau: I made list_types kwarg only.

IPython/core/magics/namespace.py
k-zen: We could just put there, but then again it may change again in the future. I thought more practical to list the *built-ins*.

IPython/core/oinspect.py
k-zen: Looks good 👍

.gitignore
k-zen: Good idea @Carreau.

Carreau:
Thanks, apologies for the delay in review and merge.

PR #11679, opened 2019-04-08
Use ast.parse a  when breaking lines for the IPython Sphinx Directive

thequackdaddy:
Currently, IPython's sphinx directive doesn't really use the `ast.parse` to decide where to break lines, except for functions. This encourages the directive to do just that. 

This will likely break a bunch of people's docs, so I would suggest we do this right and give some warning. 

Additionally, I think currently bad syntax gets ignored by the code. This will now crash.

xref #11362

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -437,8 +437,6 @@ class EmbeddedSphinxShell(object):
         Process data block for INPUT token.
 
         """
-        import pdb
-        # pdb.set_trace()
         decorator, input, rest = data
         image_file = None
         image_directive = None
@@ -791,17 +789,18 @@ class EmbeddedSphinxShell(object):
         prompts as needed capturing stderr and stdout, then returns
         the content as a list as if it were ipython code
         """
-        import pdb
-        pdb.set_trace()
         output = []
+        fmtin = self.promptin
+        ct = 1
 
-        savefigs = [i for i in range(len(content))
-                    if content[i][:8] == '@savefig']
+        savefigs = [i for i, line in enumerate(content)
+                    if line[:8] == '@savefig']
+        comments = [i for i, line in enumerate(content)
+                    if line != '' and line[0] == '#']
         content_text = '\n'.join([c if c[:8] != '@savefig' else ''
                                   for c in content])
         mod = ast.parse(content_text)
         linestarts = [body.lineno - 1 for body in mod.body]
-        comments = [i for i in range(len(content)) if content[i][0] == '#']
 
         breaks = sorted(savefigs + linestarts + comments)
         for i, lineno in enumerate(breaks):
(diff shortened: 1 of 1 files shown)


thequackdaddy:
More details... 

Given an rst file like this...

```
.. test documentation master file, created by
   sphinx-quickstart on Sun Apr  7 18:18:27 2019.
   You can adapt this file completely to your liking, but it should at least
   contain the root `toctree` directive.

Welcome to test's documentation!
================================

.. toctree::
   :maxdepth: 2
   :caption: Contents:

.. ipython:: python

   def haha(x):
       return x

That was fun.

.. ipython:: python

   class myclass:
       def __init__(self, x):
         self.x = x

       def printme(self):
         return x

   x = 1

   # This is a comment
   def haha(x):
       return x

   @haha
   def haha2(x):
     return x


Indices and tables
==================

* :ref:`genindex`
* :ref:`modindex`
* :ref:`search`
```

I think you want this...

Screen Shot 2019-04-07 at 10 46 41 PM

That rst file crashes in master. However, if you trim this rst as such...

```
.. ipython:: python

   class myclass:
       def __init__(self, x):
         self.x = x

       def printme(self):
         return x
```

You get this, which is also wrong.

Screen Shot 2019-04-07 at 10 50 14 PM

Carreau:
Many thanks ! I'll review that as soon as I can. I'm not super familiar with the directive code, and having this code get some love is good !

IPython/sphinxext/ipython_directive.py
```diff
@@ -790,74 +790,56 @@ def process_pure_python(self, content):
...
-                continue
-
-            # handle decorators
-            if line_stripped.startswith('@'):
-                output.extend([line])
-                if 'savefig' in line:
-                    savefig = True # and need to clear figure
+        try:
+            savefigs = [i for i, line in enumerate(content)
+                        if line[:8] == '@savefig']
```
Carreau: I'm going to drop comment also for myself while reviewing and trying to understand what this code does. 

Is there anything that would prevent a `@savefig` to be indented ? Like in a loop or an if statement ?

IPython/sphinxext/ipython_directive.py
```diff
@@ -790,74 +790,56 @@ def process_pure_python(self, content):
...
+                                      for c in content])
+            mod = ast.parse(content_text)
+            linestarts = [body.lineno - 1 for body in mod.body]
+        except Exception:
+            savefigs = []
+            comments = []
+            linestarts = [0]
+
+        breaks = sorted(savefigs + linestarts + comments)
+        for i, lineno in enumerate(breaks):
```
Carreau: `enumerate(breaks[1:])` ?

IPython/sphinxext/ipython_directive.py
```diff
@@ -790,74 +790,56 @@ def process_pure_python(self, content):
...
+            if i == 0:
                 continue
+            output.append([c for c in content[breaks[i-1]:lineno]])
+
+        output.append([c for c in content[lineno:]])
```
Carreau: what if there was no breaks ? will lineno be undefined ?

thequackdaddy: Yeah I need to write some testcases for this.

thequackdaddy:
@Carreau So I'm having difficulty writing a test. I didn't really realize, but this entire module is not tested pretty much at all. 

The main challenge appears to be that I need to instantiate the `EmbeddedSphinxShell` and `IPythonDirective` in the module, but that hasn't been done. 

This (possibly) would be easy if you were using pytest because we could just load the sphinx module and all their infrastructure they've already built. 

Alternatively--and what I'm leaning towards--is to move some of the code that we want to test to module-level functions that don't need to be instantiated in a class, write tests against those, and then reference those functions in the class. 

Happy for other ideas. You likely can provide better guidance.

Carreau:

PR #11680, opened 2019-04-08
Update StoreMagic docstring on how to remove an alias

switowski:
I was playing a bit with the `%store` magic and I was surprised that I couldn't delete the aliases with the `%store -z`/`%store -d`. Then I started reading the code and realized that there is an `%unalias` method for that (which is logical, but not obvious when you are in the `%store` documentation), so I've update the documentation with a small info about that.
I hope it will help someone in the future to avoid this confusion.


Carreau:
Good Call !

PR #11685, opened 2019-04-11
docs: typo fix

switowski:
While reading the documentation, I found what seems to be a small duplication (the "function" / "work" means the same), so I removed it.


Carreau:
Thanks !

PR #11689, opened 2019-04-12
docs: update the aiohttp usage in autoawait.rst

### LINKED ISSUE #11688: Asynchronous in REPL documentation: aiohttp API has changed
The first 2 examples in the "Asynchronous in REPL" documentation are using the following code:
```
import aiohttp
result = aiohttp.get('https://api.github.com')
```

and if one is using version 2.0 of aiohttp, they won't work (the `aiohttp.get` method was removed, as explained here: https://github.com/aio-libs/aiohttp/issues/1593).

To make them work again with the least amount of boilerplate code, one can do this:
```
import aiohttp
session = aiohttp.ClientSession()
result = session.get('https://api.github.com')
```
(which will give a warning: `DeprecationWarning: The object should be created from async function` but it will work fine).

switowski:
Updates the documentation in the "Asynchronous in REPL" section, so
the code used there works with aiohttpd version>=2.0.0 (closes #11688).

### DIFF
--- a/docs/source/interactive/autoawait.rst
+++ b/docs/source/interactive/autoawait.rst
@@ -29,7 +29,8 @@ Python REPL::
     [GCC 4.2.1]
     Type "help", "copyright", "credits" or "license" for more information.
     >>> import aiohttp
-    >>> result = aiohttp.get('https://api.github.com')
+    >>> session = aiohttp.ClientSession()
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks ! 

Much appreciated update !

PR #11693, opened 2019-04-18
Lazy load unicode names

zzzz-qq:
Lazy load unicode names when they are used first time rather than when ipython startup.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -168,12 +168,6 @@ MATCHES_LIMIT = 500
 
 _deprecation_readline_sentinel = object()
 
-names = []
-for c in range(0,0x10FFFF + 1):
-    try:
-        names.append(unicodedata.name(chr(c)))
-    except ValueError:
-        pass
 
 class ProvisionalCompleterWarning(FutureWarning):
     """
@@ -997,6 +991,8 @@ def _make_signature(completion)-> str:
 
 class IPCompleter(Completer):
     """Extension of the completer class with IPython-specific features"""
+
+    _names = None
 
     @observe('greedy')
     def _greedy_changed(self, change):
@@ -2073,14 +2069,21 @@ class IPCompleter(Completer):
         return text, _matches, origins, completions
         
     def fwd_unicode_match(self, text:str) -> Tuple[str, list]:
-        # initial code based on latex_matches() method
+        if self._names is None:
+            self._names = []
+            for c in range(0,0x10FFFF + 1):
+                try:
+                    self._names.append(unicodedata.name(chr(c)))
+                except ValueError:
+                    pass
+
         slashpos = text.rfind('\\')
         # if text starts with slash
         if slashpos > -1:
             s = text[slashpos+1:]
-            candidates = [x for x in names if x.startswith(s)]
+            candidates = [x for x in self._names if x.startswith(s)]
             if candidates:
-                return s, [x for x in names if x.startswith(s)]
+                return s, candidates
             else:
                 return '', ()
         


Carreau:
@LucianaMarques that's your code, you might want to review it; 

But otherwise that's looks good to me.

LucianaMarques:
Hey there, sorry for the late reply. 

If tests are working and this will improve performance, I don't see why not accept it :) That's actually really nice improvement, thanks!

I would like though that you provide us more info on the profiling you did, how you did it, etc. I'm not used to profiling yet and I would certainly like to learn more about it. 

I am doing some work in the completer class, if you have some more suggestions I would love to hear it.

IPython/core/completer.py
```diff
@@ -998,6 +992,8 @@ def _make_signature(completion)-> str:
 class IPCompleter(Completer):
     """Extension of the completer class with IPython-specific features"""
 
+    _names = None
```
LucianaMarques: When is '_names' changes?

IPython/core/completer.py
```diff
@@ -2073,14 +2069,22 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
         return text, _matches, origins, completions
         
     def fwd_unicode_match(self, text:str) -> Tuple[str, list]:
+        if self._names is None:
```
LucianaMarques: Do you really need to make this check?

IPython/core/completer.py
```diff
@@ -2073,14 +2069,22 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
...
+                except ValueError:
+                    pass
+
         # initial code based on latex_matches() method
         slashpos = text.rfind('\\')
         # if text starts with slash
         if slashpos > -1:
             s = text[slashpos+1:]
-            candidates = [x for x in names if x.startswith(s)]
+            candidates = [x for x in self._names if x.startswith(s)]
```
LucianaMarques: If I'm not mistaken, we only use variables starting with underscore when that variable is not intended to be used (please someone correct me if needed). I don't think this is the case, so please consider changing this.

IPython/core/completer.py
```diff
@@ -2073,14 +2069,22 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
...
             if candidates:
-                return s, [x for x in names if x.startswith(s)]
+                return s, candidates
```
LucianaMarques: Nice, the code is much cleaner like this.

LucianaMarques (commented):
I just did a review on your pull request. Please take some time to consider my comments and let's discuss them :)

Also, I like to put descriptions in my pull requests, even if it's just something like "closes issue number #x". It's important to keep things documented in my opinion.

IPython/core/completer.py
Carreau: This is a tiny bit complex, but here we have *class variable*, in some case it is equivalent to `self._name`. In this particular case, that would be equivalent to 

```
def __init__(self, *args, **kwargs):
    self._names = None
    ...
````

File: IPython/core/completer.py (hunk 2 above)
Comment (Carreau) [reply IPython/core/completer.py]: Yes, so that it is recomputed only once. 
Above it is set to `None`, and the first time this is called `self._names` get computed.

File: IPython/core/completer.py (hunk 3 above)
Comment (Carreau) [reply IPython/core/completer.py]: It depends on context, when set as an attribute on an object is to mean that the attribute (or method) should be considered as private. It is meant for use only by the object author, and not by the object users (or at least at their own risk).

**Comment by Carreau** [comment]
> Hey there, sorry for the late reply.

No need to apologize, we are all busy. 
Those were really good comment and questions. 

> I would like though that you provide us more info on the profiling you did, how you did it, etc. I'm not used to profiling yet and I would certainly like to learn more about it.

I don't think much profiling is necessary. It won't be "faster" in all cases, it will mostly delay the computation of the unicodes mapping until actually requested. 

This mean that the startup of IPython will be a bit faster, but the first press on `<tab>` that needs completion will be a bit slower.

File: IPython/core/completer.py (hunk 3 above)
Comment (LucianaMarques) [reply IPython/core/completer.py]: We keep private variables with the underscore then? Is this stated in any PEP or best practices documment? I guess we could add some tool on it to the CI tests if they are not already there :) but thanks for the explanation

Comment (Carreau) [reply IPython/core/completer.py]: Yes it is in [pep8 ](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles): 

> _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose names start with an underscore.

It is a convention. Nothing enforces it.

**Comment by LucianaMarques** [comment]
I have one more choice l code comment but i'm on my phone now, please wait before merging just a little bit.

**Comment by Carreau** [comment]
> I have one more code comment but i'm on my phone now, please wait before merging just a little bit.

Don't worry plenty of time we'll wait for your approval.

File: IPython/core/completer.py [hunk 5]
```diff
@@ -2073,14 +2069,22 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
...
     def fwd_unicode_match(self, text:str) -> Tuple[str, list]:
+        if self._names is None:
+            self._names = []
+            for c in range(0,0x10FFFF + 1):
+                try:
+                    self._names.append(unicodedata.name(chr(c)))
+                except ValueError:
+                    pass
+
         # initial code based on latex_matches() method
```
LucianaMarques: Could you please exclude this comment? I added it while working on my last PR to keep track what I had to do, but it was intended to be erased. It's trash now.

IPython/core/completer.py
LucianaMarques: Nice! I guess we could include some code style checks, something to discuss in a thread :)

LucianaMarques:
> Don't worry plenty of time we'll wait for your approval.

Thanks! Just to sum up, only exclude the comment I mentioned above and add some description to the PR, and in my opinion it's ready to be merged.

zzzz-qq:
> > Hey there, sorry for the late reply.
> 
> No need to apologize, we are all busy.
> Those were really good comment and questions.
> 
> > I would like though that you provide us more info on the profiling you did, how you did it, etc. I'm not used to profiling yet and I would certainly like to learn more about it.
> 
> I don't think much profiling is necessary. It won't be "faster" in all cases, it will mostly delay the computation of the unicodes mapping until actually requested.
> 
> This mean that the startup of IPython will be a bit faster, but the first press on `<tab>` that needs completion will be a bit slower.

@Carreau  Really appreaciate your patience and kindness.

Carreau:
> Really appreciate your patience and kindness.

Thanks much appreciated; I wish I had more time to mentor and guide others. 
Thanks to you for taking the time to send a pull-request, and allow us to discuss it; I know it can be frustrating to wait. 

I just released 7.5 yesterday so this did not make the cut, but I'm hopping to get that in 7.6 next month.

LucianaMarques:
@Carreau are we wainting for 7.5 to merge this PR? I don't have merge permission, so I was just wandering why you didn't merge this yet.

Carreau:
Thanks; I have some time to look at ongoing PRs, merging this. 

Thanks both !

PR #11698, opened 2019-04-23
Fix applying @needs_global_scope to cell magics.

### LINKED ISSUE #11659: %%time magic prevents assignment
The `%%time` cell magic is preventing me from binding names. I found this in notebook code on Travis (py 3.6) and reproduced it locally (py 3.7) in notebook and in `ipython` at the terminal. It doesn't seem to apply to all cell magics:

```
Python 3.7.2 | packaged by conda-forge | (default, Mar 19 2019, 20:46:22)
IPython 7.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %%time
   ...: foo = "bar"
   ...:
   ...:
CPU times: user 3 µs, sys: 0 ns, total: 3 µs
Wall time: 6.91 µs

In [2]: foo
<ipython-input-2-f1d2d2f924e9> in <module>
----> 1 foo

NameError: name 'foo' is not defined

In [3]: %%prun
   ...: baz = "qux"
   ...:
   ...:
         3 function calls in 0.000 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 <string>:2(<module>)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}

In [4]: baz
Out[4]: 'qux'
```

Confirmed that 7.3.0 worked fine locally.

Carreau:
There was some side effect in some magic to allow passing
needs_local_scope. We know pass the global_scope to cell magice when
using @needs_local_scope, but at the same time this means that we can't
rely on the existance of local_ns from within the magic to know wheter
we are cell-called or line called.

Fixes #11659

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2345,7 +2345,7 @@ class InteractiveShell(SingletonConfigurable):
                 magic_arg_s = self.var_expand(line, stack_depth)
             kwargs = {}
             if getattr(fn, "needs_local_scope", False):
-                kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
+                kwargs['local_ns'] = self.user_ns
 
             with self.builtin_trap:
                 args = (magic_arg_s, cell)
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1127,8 +1127,8 @@ python-profiler package from non-free.""")
         ns = {}
         glob = self.shell.user_ns
         # handles global vars with same name as local vars. We store them in conflict_globs.
-        if local_ns is not None:
-            conflict_globs = {}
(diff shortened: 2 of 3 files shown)


dwhswenson:
I can verify that this fixes the original problem I reported in #11659 (at least, on my local machine -- I'll wait until 7.5.0 comes out to test on Travis, but that shouldn't make any difference). In addition, I see no unexpected side effects in the rest of our test suite (our system tests are in notebooks). Tested in Py 3.6.

Carreau:
Thanks @dwhswenson and @minrk

PR #11715, opened 2019-05-09
[Outdated] Add prompt_toolkit 3.0 support

gpotter2:
**This shouldn't be merged before the official release of [`prompt_toolkit` 3.0](https://github.com/prompt-toolkit/python-prompt-toolkit/commit/c7851c09ffbfa7d06458408abee77ab6f548852d)**. See the *[changelog](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/CHANGELOG)* for more infos

This will most likely be done by jonathanslenders once it is ready, but I allowed myself to start the work (at least make IPython usable). There's probably more potential (cool asyncio..). 

I hope that's not an issue if this PR waits.

## Details

<details>
<summary>
So far, this makes IPython runnable but triggers warnings on closing (I don't think that's on IPython's side)
</summary>

```

Task was destroyed but it is pending!
task: <Task pending coro=<History._start_loading() running at /home/gpotter/.pyenv/versions/3.7.2/lib/python3.7/site-packages/prompt_toolkit-3.0.0-py3.7.egg/prompt_toolkit/history.py:39>>
/home/gpotter/.pyenv/versions/3.7.2/lib/python3.7/asyncio/base_events.py:609: RuntimeWarning: coroutine 'History._start_loading' was never awaited
  self._ready.clear()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
Task was destroyed but it is pending!
task: <Task pending coro=<History._start_loading() running at /home/gpotter/.pyenv/versions/3.7.2/lib/python3.7/site-packages/prompt_toolkit-3.0.0-py3.7.egg/prompt_toolkit/history.py:39>>
Task was destroyed but it is pending!
task: <Task pending coro=<History._start_loading() running at /home/gpotter/.pyenv/versions/3.7.2/lib/python3.7/site-packages/prompt_toolkit-3.0.0-py3.7.egg/prompt_toolkit/history.py:39>>
```

*(It's actually a single issue that makes multiple warnings, I suspect it's a bug (ptk v3 is in early stages))*

</details>

<details>
<summary> 
 `patch_stdout` is said to have been updated for a newer format, but this info doesn't match the code behind it.
</summary>

From the changelog:
> `patch_stdout` now requires an `Application` as input.

In reality (as of writing): [commit](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/c7851c09ffbfa7d06458408abee77ab6f548852d/prompt_toolkit/patch_stdout.py#L35-L46)

```python
def patch_stdout(raw: bool = False) -> Generator[None, None, None]:
    """
    Replace `sys.stdout` by an :class:`_StdoutProxy` instance.
    Writing to this proxy will make sure that the text appears above the
    prompt, and that it doesn't destroy the output from the renderer.  If no
    application is curring, the behaviour should be identical to writing to
    `sys.stdout` directly.
    :param raw: (`bool`) When True, vt100 terminal escape sequences are not
                removed/escaped.
    """
```

</details>

## Status

**Important note:** `prompt_toolkit 3.0` requires `Python 3.6+`. ~~It's however totally doable to support both V2 and V3. However I'm not sure how IPython deals with versioning, therefore any directions would be appreciated.~~ **Therefore this supports both v2 and v3**

I'll update the PR to keep it working, if other changes are made, until it either gets discussed, or a better PR replaces it (in which case feel free to close this)

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -15,8 +15,8 @@ from traitlets import (
     Any, validate
 )
 
+import prompt_toolkit
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
-from prompt_toolkit.eventloop.inputhook import set_eventloop_with_inputhook
 from prompt_toolkit.filters import (HasFocus, Condition, IsDone)
 from prompt_toolkit.formatted_text import PygmentsTokens
 from prompt_toolkit.history import InMemoryHistory
@@ -39,6 +39,15 @@ from .ptutils import IPythonPTCompleter, IPythonPTLexer
 from .shortcuts import create_ipython_shortcuts
 
 DISPLAY_BANNER_DEPRECATED = object()
+
+try:
+    PROMPT_TOOLKIT_VERSION = int(prompt_toolkit.__version__[0])
+    if PROMPT_TOOLKIT_VERSION >= 3:
+        from prompt_toolkit.eventloop.inputhook import set_eventloop_with_inputhook
+except ValueError:
+    warn("Unknown prompt_toolkit version !")
+    # Let's assume v2 for now
+    PROMPT_TOOLKIT_VERSION = 2
 
 
 class _NoStyle(Style): pass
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks for working on this; 

> However I'm not sure how IPython deals with versioning, therefore any directions would be appreciated.

Usually explicit check fro `sys.version_info`  and or packages version , and switch codepath. 

If this is big enough of a change, we can think o f dropping support for some things an in that case bump the major version number to 8.

_if_ we can make it work with both prompt toolkit 2 and 3 then great, we can make a 8.0 when we drop support for prompt toolkit 2.

gpotter2:
Thanks for the feedback, I've updated it. As of today this is (should? I haven't tested every IPython function) support both V2 and V3.

I will try to follow the evolutions of V3 for it to keep working.

gpotter2:
Closed by https://github.com/ipython/ipython/pull/11726

Issue #11707, opened 2019-05-03
%%time cell magic suppresses cell output


angertdevsingh:
The following code does not display the value of `x` when run.
```
%%time
x = 10
x = x*2
x
```

Expected output includes the execution time and the value of `x`.

Using the `%time` line magic seems to work, even for multi-line code.

Additional info:
ipython_version: 7.5.0
os_name: nt
platform: Windows-10-10.0.17134-SP0
sys_platform': win32
sys_version: 3.7.3 [MSC v.1915 64 bit (AMD64)]

Carreau:
I believe this is on purpose, `%timeit` may run the code several time; and does some modifications to the actual statement you run; so returning the value of the "Last" statement might not be what you expect.

angertdevsingh:
> I believe this is on purpose, `%timeit` may run the code several time; and does some modifications to the actual statement you run; so returning the value of the "Last" statement might not be what you expect.

`%time` and `%%time` only run the code once though.

`In [1]: x = 5`

1. Using `%timeit`, `y` is not assigned a value

```
In [2]: %timeit y = x*2
# 32.6 ns ± 0.341 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
```
In [3]: y
# NameError: name 'y' is not defined
```

2. Using `%time` however, a value is assigned to `y`.

```
In [4]: %time y = x*2
# Wall time: 0 ns
```
```
In [5]: y
Out[5]: 10
```

3. Similarly, `%time` outputs a value in `Out[ ]` when the statement is not assigned to `y`.

```
In [6]: %time x*2
# Wall time: 0 ns
Out[6]: 10
```

4. However, `%%time`'s behaviour is inconsistent with `%time`.

- It does not output a value as in 3.:
```
In [7]: %%time
        z = x*2
        z
# Wall time: 0 ns
```
- But it does assign a value to `z` as in 2.:
```
In [8]: z
Out[8]: 10
```

Is there a reason why the behaviour of `%%time` is not consistent with `%time`?

Pragnya02:
@Carreau  , @angertdevsingh  I am planning on working on this. This will be my very first-issue I will be contributing to.

To understand better , %%time should display the output for the last line in the cell, instead of suppressing it, like it currently does?

angertdevsingh:
@Pragnya02 I think it should, yes.

Carreau:
> `%time` and `%%time` only run the code once though.

[...]

PR #11716, opened 2019-05-09
Fix magic time

Pragnya02:
In case of %%time with multi-lines in cell , eval the last line and display in output instead of suppressing it.

Refer Issue : https://github.com/ipython/ipython/issues/11707

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -49,6 +49,14 @@ from warnings import warn
 from logging import error
 from io import StringIO
 
+if sys.version_info > (3,8):
+    from ast import Module
+else :
+    # mock the new API, ignore second argument
+    # see https://github.com/ipython/ipython/issues/11590
+    from ast import Module as OriginalModule
+    Module = lambda nodelist, type_ignores: OriginalModule(nodelist)
+
 
 #-----------------------------------------------------------------------------
 # Magic implementation classes
@@ -1276,7 +1284,7 @@ python-profiler package from non-free.""")
                                                 , '<timed eval>'
                                                 , 'eval')
                 expr_ast=expr_ast.body[:-1]
-                expr_ast = ast.Module(expr_ast)
+                expr_ast = Module(expr_ast, [])
 
         t0 = clock()
         code = self.shell.compile(expr_ast, source, mode)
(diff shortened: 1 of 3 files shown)


Carreau:
Looks good to me, 

I'll try to see if I can come up with test to make sure the behavior does not regress.

IPython/core/magics/execution.py
```diff
@@ -1285,15 +1294,21 @@ def time(self,line='', cell=None, local_ns=None):
...
                 exec(code, glob, local_ns)
+                out=None
+                # multi-line %%time case
+                if expr_val and isinstance(expr_val, ast.Expr):                           
+                    out = eval(code_val, glob, local_ns)  
             except:
-                self.shell.showtraceback()
-                return
+                self.shell.showtraceback()               
+                return
```
gpotter2: If I'm not mistaken this only adds extra unneeded spaces.

Pragnya02: Did you mean showtraceback() ? That would be for the entire try block (from line 1301 to 1305)

gpotter2: What I meant was that your diff is full of spacing changes :smile:
B95A112B-B9A9-426A-BBE7-4396C0179C04

Pragnya02: Oops, thank you ! Fixed that now, should be fine.

Carreau:
Thanks for your patience, 

I rebased and squashed a few commits to not have the whitespaces changes in history, plus added a test. 

Once test passes I'm happy to merge this. 

Thanks.

IPython/core/magics/execution.py
```diff
@@ -1261,13 +1261,23 @@ def time(self,line='', cell=None, local_ns=None):
...
             mode = 'exec'
             source = '<timed exec>'
+            # multi-line %%time case
+            if len(expr_ast.body) > 1 :
+                expr_val=expr_ast.body[-1]
+                code_val = self.shell.compile(ast.Expression(expr_val.value)
+                                                , '<timed eval>'
+                                                , 'eval')
+                expr_ast=expr_ast.body[:-1]
+                expr_ast = ast.Module(expr_ast)
```
Carreau: That apparently is not compatible python 3.8+; I'd love to not break on a new Python.

Carreau: You need to do something similar to https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L110-L116

Carreau: I've done it.

Carreau:
Thanks @Pragnya02 !

PR #11718, opened 2019-05-10
Remove leading empty lines

### LINKED ISSUE #11594: Problem with indentation in the `run_cell` method
Hi,

When using the InteractiveShell method `run_cell`, there is a problem when the string to execute begins with a line that contains only 4 spaces. It then does not handle properly  the indentations of the subsequent loops.
image

Tested on JupyterNotebook, JupyterLab and the Ipython interactive shell in the terminal.

Config:
- jupyter --version 4.4.0
- ipython --version 7.0.1
- python3 --version Python 3.6.7
- Ubuntu 18.04.1 LTS

rgson:
When running a cell, the leading indentation on the first line is
removed from all of the cell's lines. This causes indentation errors
on subsequent lines in case the first line consists entirely of
whitespace matching a valid level of indentation that is greater than
the least indented line.

This change avoids the issue by first removing any leading,
whitespace-only lines.

Closes #11594

### DIFF
--- a/IPython/core/tests/test_inputtransformer2_line.py
+++ b/IPython/core/tests/test_inputtransformer2_line.py
@@ -86,3 +86,31 @@ def test_leading_indent():
     for sample, expected in [INDENT_SPACES, INDENT_TABS]:
         nt.assert_equal(ipt2.leading_indent(sample.splitlines(keepends=True)),
                         expected.splitlines(keepends=True))
+
+LEADING_EMPTY_LINES = ("""\
+    \t
+
+if True:
+    a = 3
+
+b = 4
+""", """\
+if True:
+    a = 3
+
+b = 4
+""")
+
+ONLY_EMPTY_LINES = ("""\
+    \t
+
+""", """\
+    \t
+
+""")
+
+def leading_empty_lines():
+    for sample, expected in [LEADING_EMPTY_LINES, ONLY_EMPTY_LINES]:
+        nt.assert_equal(
+                ipt2.leading_empty_lines(sample.splitlines(keepends=True)),
+                expected.splitlines(keepends=True))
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks looks good to me. 
I'm still going to ask thoughts of @takluyver who wrote most of this.

IPython/core/tests/test_inputtransformer2_line.py
```diff
@@ -86,3 +86,31 @@ def test_leading_indent():
...
+
+ONLY_EMPTY_LINES = ("""\
+    \t
+
+""", """\
+    \t
+
+""")
+
+def leading_empty_lines():
```
takluyver: I haven't looked into IPython's test system in a long time, but I think tests need to be named like `test_foo` to be found.

takluyver:
Thanks, other than the point about testing, this looks OK to me.

IPython/core/tests/test_inputtransformer2_line.py
Carreau: Fixed. Thanks !

ghazpar:
I understand the interest of being able to paste code into a notebook cell and run it without having to fix the indentation, but there are also some bad downsides to this kind of hack:
1. A Python beginner might think that such indentation is valid syntax.
2. If I copy such cell content that runs fine in a notebook and than paste it somewhere else, I will get a syntax error, which can again be confusing to a beginner, and frustrating to a non-beginner.

I don't think its a good idea that a notebook cell be allowed to run Python code with invalid indentation.  

If there is really a need to tolerate superfluous indentation when pasting, the indentation should be corrected automatically during the paste process so that if the cell runs without error, it's because the cell contains valid Python syntax. 

And such automatic correction behavior should ideally be optional.

Carreau:
Happy to make it optional, but someone would have to send patches to IPython/IPykernel to have the relevant configuration hooks.

Issue #11719, opened 2019-05-11
issue with Python-3.8.0a4


stonebig:
a bit like for cloupickle https://github.com/cloudpipe/cloudpickle/issues/266, I get this error when trying a basic function 

````
import asyncio 

await asyncio.sleep(0)
````

````
`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
C:\WinP\bd38\bu\WPy64-3800a4\python-3.8.0a4.amd64\lib\site-packages\IPython\core\interactiveshell.py in removed_co_newlocals(function)
    139     CO_NEWLOCALS = 0x0002
    140     code = function.__code__
--> 141     new_code = CodeType(
    142         code.co_argcount,
    143         code.co_kwonlyargcount,

TypeError: an integer is required (got type bytes)`
````

![image](https://user-images.githubusercontent.com/4312421/57568423-9b339780-73e7-11e9-8625-faf8f5b11b5b.png)

stonebig:
https://bugs.python.org/issue36886

It's apparently PEP 570 effect

stonebig:
There may be code to change because of PEP 570, which added a parameter to the function that is used in Types module https://github.com/python/cpython/pull/12701/files#diff-cb0b9d6312c0d67f6d4aa1966766ceddL27

stonebig:
you have to do a change like https://github.com/cloudpipe/cloudpickle/pull/267/files#diff-d2a3618afedd4e124c532151eedbae09

Carreau:
I'm trying to get top-level await as a native CPython thing see https://github.com/python/cpython/pull/13148 and https://github.com/ipython/ipython/pull/11713, 
If that's goes in I guess 11720 will not be necessary – so apologies if it sits there a bit.

MrMino:
This issue is no longer reproducible; I'm assuming it has been fixed by #11713.

PR #11713, opened 2019-05-07
Async exec in Python 3.8

Carreau:
Description will come later – this require a custom Python which have extra flags for compile.
See https://bugs.python.org/issue34616

PR #11720, opened 2019-05-11
Python-3.8 PEP570 positional only argument

stonebig:
to solve https://github.com/ipython/ipython/issues/11719


Carreau:
I'm going to merge this – having the test-not passing on 3.8 annoys me for some other work – though I may revert it if we don't need to call `removed_co_newlocals` in 3.8 final, hope that is ok.

Carreau:
Hi there, You seem to run IPython on Python 3.8; do you mind trying #11713 that tried to make use of the new 3.8b1 (not out yet) top-level await ? 

Thanks !

PR #11726, opened 2019-05-14
Support prompt_toolkit 3.0 as well

jonathanslenders:
These are the required changes to support prompt_toolkit 3.0, once released.

The two biggest changes in prompt_toolkit 3.0 are:
- We have >95% of the code base covered with strict type annotations.
- It uses the asyncio event loop internally, rather then a custom built loop. This is also interesting for IPython, because asyncio tasks created in the repl keep running.

Because of the asyncio event loop, the way input hooks work is slightly different. This works now by creating a custom Selector object, and passing that to an asyncio SelectorLoop.

We can easily remain compatible with both prompt_toolkit 2 & 3, because the rest of the API remains mostly untouched.

Some further testing is required:
- Make sure that the input hooks are actually working correctly for IPython.
- Make sure it runs fine on Windows as well (I did some basic testing).
- Make sure it runs fine on Windows, using Python 3.8. - Python 3.8 will get the ProactorEventLoop by default on Windows, I believe there are still some issues with the prompt_toolkit/ProactorEventLoop/inputhooks combination.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -25,6 +25,7 @@ from prompt_toolkit.patch_stdout import patch_stdout
 from prompt_toolkit.shortcuts import PromptSession, CompleteStyle, print_formatted_text
 from prompt_toolkit.styles import DynamicStyle, merge_styles
 from prompt_toolkit.styles.pygments import style_from_pygments_cls, style_from_pygments_dict
+from prompt_toolkit import __version__ as ptk_version
 
 from pygments.styles import get_style_by_name
 from pygments.style import Style
@@ -38,6 +39,7 @@ from .ptutils import IPythonPTCompleter, IPythonPTLexer
 from .shortcuts import create_ipython_shortcuts
 
 DISPLAY_BANNER_DEPRECATED = object()
+PTK3 = ptk_version.startswith('3.')
 
 
 class _NoStyle(Style): pass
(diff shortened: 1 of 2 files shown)


Carreau:
> * It uses the asyncio event loop internally, rather then a custom built loop. This is also interesting for IPython, because asyncio tasks created in the repl keep running.

Interesting; autoawait also support non-asyncio, eventloop and also let user call `asyncio.run()`;  I'm guessing that if we run ptk with the asyncio eventloop then, this prevent users from using `run()`.

jonathanslenders:
No, it shouldn't prevent that. ptk's ``PromptSession().prompt()`` will run asyncio's event loop while waiting for input. The loop stops running when the input is accepted, which means we are back in synchronous mode and can run any other loop. We will have a ``PromptSession().prompt_async()`` which requires a running event loop and is awaitable. The normal prompt call exposes a synchronous API.

jonathanslenders:
@Carreau: I just released prompt_toolkit 3.0. I guess it's time to play again with this PR.

One change we probably have to do is make sure in setup.py to fall back to prompt_toolkit 2.0, if we're on Python 3.3 - 3.5.

Carreau:
> One change we probably have to do is make sure in setup.py to fall back to prompt_toolkit 2.0, if we're on Python 3.3 - 3.5.

No need to, we're not compatible 3.5 anymore :-)

Carreau:
This seem to work as is for me tests are passing on my machine, so beyond the small one line conflict I'm happy to get this as-is for now.

Carreau:
(rebased).

Carreau:
Ok, let's get al least that in, and refine with subsequents PR if needed.

PR #11728, opened 2019-05-15
BUG: Fix savefig path in Windows

bashtage:
Replace os.path which uses platform-specific separator with Path.as_posix
which always uses forward slash


bashtage:
When using @savefig on windows, the path uses windows separators, which Sphinx treats as escapes `\`. I'm not 100% positive this is the correct solution, but it fixed my issue building `statsmodels` docs on Windows.

Carreau:
Thanks for your patience, 

 trusting you on this one I don't have a windows machine to try one.

PR #11734, opened 2019-05-16
Autoformat with black

Carreau:
Prototype, 

see https://twitter.com/Mbussonn/status/1128841958711889920

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -87,10 +87,23 @@ def create_ipython_shortcuts(shell):
 
 
 def newline_or_execute_outer(shell):
+
+    import black
     def newline_or_execute(event):
         """When the user presses return, insert a newline or execute the code."""
         b = event.current_buffer
         d = b.document
+        def try_reformat():
+            try:
+                tbc = b.delete_before_cursor(len(d.text[:d.cursor_position]))
+                fmt= black.format_str(tbc, mode=black.FileMode())
+                if not tbc.endswith('\n') and fmt.endswith('\n'):
+                    fmt = fmt[:-1]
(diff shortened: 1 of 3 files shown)


stonebig:
I replaced shortcuts.py per this one, but nothing happens...
Black is there... did I miss something ?
image

stonebig:
ah, .. it works only on ipython "DOS" console ?

IPython/terminal/shortcuts.py
```diff
@@ -87,10 +87,23 @@ def create_ipython_shortcuts(shell):
 
 
 def newline_or_execute_outer(shell):
+
+    import black
     def newline_or_execute(event):
         """When the user presses return, insert a newline or execute the code."""
         b = event.current_buffer
         d = b.document
+        def try_reformat():
```
LucianaMarques: I think it's a little bit confusing to define a function inside of another function. Could you please tell me whether there's a specific reason for this?

LucianaMarques (commented):
I'm not getting where you are checking whether black is installed, could you please point it out for me? :) Thanks!

IPython/terminal/shortcuts.py
Carreau: For both of your questions the answer is : "there is no reason beyond pure lazyness", I was prototyping and looking at "can this be done"; you are right that a proper implementation should move the import to the top of the file and define the `try_reformat()` function outside, passing in the `b` buffer.

LucianaMarques: Ok, fair enough :)

How do you check whether black is activated?

Carreau:
I've pushed some update, including some documentation; and will try to get that in a a preview for 7.10 at end of month.

PR #11737, opened 2019-05-18
Add keywords completion + docs / Questions about completion API

gpotter2:
**Edit: I've added a bunch of questions below :/ sorry if it's not really the right place (PR)**
This PR:
- adds doc for `_ipython_key_completions_` (it was only mentioned)
- implements `_ipython_kw_completions_` in a similar way (and add some docs)
- (off-topic): makes `./make.cmd` work on a Windows setup where both Python 2 & 3 are installed.

image

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1125,6 +1125,7 @@ class IPCompleter(Completer):
             return [
                 self.file_matches,
                 self.magic_matches,
+                self.python_func_kw_matches,  # partially disabled
                 self.dict_key_matches,
             ]
         else:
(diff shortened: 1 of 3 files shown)


gpotter2:
I'm not very aware of IPython's developement, what does
```text
Like complete but can also returns raw jedi completions as well as the
origin of the completion text. This could (and should) be made much
cleaner but that will be simpler once we drop the old (and stateful)
:any:`complete` API.


With current provisional API, cursor_pos act both (depending on the
caller) as the offset in the ``text`` or ``line_buffer``, or as the
``column`` when passing multiline strings this could/should be renamed
but would add extra noise.
```

means ? Is all of this going to be dropped in favor of `jedi` ?

drunkwcodes:
This is a huge topic. I hope this API is configurable.

The keyword completion is a good work but I would rather disable it and use `?` to look up the usage and type them manually.

It would conflict with other function names and have the same problem of jedi completion you mentioned.
The API needs to sort out completers after a objective evaluation in different senarios.

https://github.com/ipython/ipython/blob/380db0d7241372f9c12d5d5aaaa8e7e4c70575d5/IPython/core/completer.py#L2010

LucianaMarques:
I am doing a very slow refactoring of this file because indeed I find it confusing. But code completion can be quite complex, so I guess it's kind of expected that this code would be equally complex. 

I will try to answer your doubts but since I'm not the original writer it would be nice to have other people commenting on it.

gpotter2:
closed for inactivity

Issue #11602, opened 2019-02-11
Some matplotlib backends on MacOS significantly slow down jupyter


yluthu:
On my MacOS Mojave system, enabling some matplotlib backends will significantly slow down the execution of subsequent cells (the slowdown starts to occur about half a minute after setting the backend). See the screenshot below for an example (notice the 10.0s execution time of the third cell):

![image](https://user-images.githubusercontent.com/3724995/52561258-d4cf2380-2e35-11e9-9585-4ea3abdcf42b.png)

In particular, I notice that the 10.0s consists of three "stages":
1. After I tell jupyter to run the cell, the kernel stays idle for a few seconds;
2. The kernel becomes busy, and the output immediately appear;
3. The kernel stays busy for another few seconds after the output appears.

I've tested with different matplotlib backends, and the issue happens when using some of the backends. Here is a summary:

- Backends that make this issue happen:
    - `osx`
    - `qt`
- Backends that do NOT make this issue happen:
    - `inline`
    - `agg`
    - `tk`
    - `nbagg`
    - `notebook`

I've also tested with different ipython/jupyter clients, and the issue happens when using some of the clients. Here is a summary:

- Clients where this issue happens:
    - `jupyter notebook`
    - `jupyter console`
    - `jupyter qtconsole`
    - `jupyter lab`
    - Nteract app
- Clients where this issue does NOT happen:
    - `ipython` console

When tested in `jupyter console`, an additional warning is cast when executing the third line in the screenshot above:

```
/usr/local/lib/python3.7/site-packages/jupyter_console/ptshell.py:661: UserWarning: The kernel did not respond to an is_complete_request. Setting `use_kernel_is_complete` to False.
  warn('The kernel did not respond to an is_complete_request. '
```

I have tried to dig into the source code of `IPython.core.pylabtools.activate_matplotlib`, and here is a minimal (to my best) example:

```
$ jupyter console
Jupyter console 6.0.0

Python 3.7.2 (default, Jan 13 2019, 12:50:01)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.



In [1]: import matplotlib

In [2]: matplotlib.rcParams['backend'] = 'MacOSX'

In [3]: import matplotlib.pyplot

In [4]: import time

In [5]: time.sleep(35)

In [6]: 'test'    # this line will take a lot of time
[...]

PR #11740, opened 2019-05-19
FIX: make sure the QSocketNotifier is disabled

tacaswell:
This fixes issues that Matplotlib was seeing with Qt5 + IPython with
PT on OSX (see https://github.com/matplotlib/matplotlib/issues/13066)

This should be back-ported as far as possible.

Attn @efiring

### DIFF
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -7,6 +7,7 @@ from IPython.external.qt_for_kernel import QtCore, QtGui
 _appref = None
 _already_warned = False
 
+
 def inputhook(context):
     global _appref
     app = QtCore.QCoreApplication.instance()
@@ -41,9 +42,12 @@ def inputhook(context):
         # loop when there is input ready to read.
         notifier = QtCore.QSocketNotifier(context.fileno(),
                                           QtCore.QSocketNotifier.Read)
-        # connect the callback we care about before we turn it on
-        notifier.activated.connect(event_loop.exit)
-        notifier.setEnabled(True)
-        # only start the event loop we are not already flipped
-        if not context.input_is_ready():
-            event_loop.exec_()
+        try:
+            # connect the callback we care about before we turn it on
+            notifier.activated.connect(event_loop.exit)
+            notifier.setEnabled(True)
+            # only start the event loop we are not already flipped
+            if not context.input_is_ready():
+                event_loop.exec_()
+        finally:
+            notifier.setEnabled(False)


efiring:
Thank you, this fixes the flakiness I was seeing.

Carreau:
> This should be back-ported as far as possible.

Like 5.x ? for Python 2 users ?

kmtac:
There might be something else going on ...

Even with this fix, there still seems to be some trouble with the qt event loop, at least on my computer. After calling the "%gui qt" magic, ipython eventually slows down dramatically. "Eventually" varies, from a minute or so to half an hour. Dramatically means time.sleep(0.1) takes 10.1 seconds. The 10.1

I'm using the developer version of ipython (cloned yesterday), but with qt.py from the fix reported here.

Minimal example:
% ipython
Python 3.7.3 (default, Mar 27 2019, 16:54:48)
IPython 7.6.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: %gui qt
In [2]: import time
In [3]: x=0.1; tic=time.time();time.sleep(x);toc=time.time();print(f"Expected {x} sec, got {toc-tic:.2f} seconds.")
Expected 0.1 sec, got 0.10 seconds.
In [4]: x=0.1; tic=time.time();time.sleep(x);toc=time.time();print(f"Expected {x} sec, got {toc-tic:.2f} seconds.")
Expected 0.1 sec, got 10.10 seconds.

This problem has been around for at least a year, but less than two years. I first noticed it when using pymeasure (which uses time.sleep).

I'm using the anaconda python distribution, see version info below.
conda version : 4.6.14
conda-build version : 3.17.8
python version : 3.7.3.final.0

I put my comments here because it seems to be the same problem. Let me know if you'd like me to file a separate bug report or to do other tests.

Carreau:
I can reproduce; I'm not quite sure why. 

Should I still merge thins ?

There is not many version of IPython I can backport that to. It only applies to the 7.x/master branch.

efiring:
I can also reproduce it.  I think this PR is an improvement, but obviously there is still a bug somewhere.  In a sequence of 6 trials, I got 0.1 s on the first one, then 10.1 s on all but the 4th, which returned 1.32 s!  The 10.1 looks like a 10-s timeout somewhere--but what accounts for the 1.32?

Carreau:
Ok, so I'm going to merge this and make a release soon. 

I'm tempted to think that there might be a lock + timeout; of ~ 10 seconds. The 1.32 might actually be just 0.32 of overhead.

efiring:
There's definitely a 10-s timeout (effectively--it's clearly waiting for 10 s in most cases).  I find that sometimes I get several executions without the timeout before it starts; but once it starts, I haven't seen it go back.  It doesn't depend on the length of the sleep; even 0.001 s is enough.  Inserting 2 sleeps in sequence yields a delay of 20 s.

@anntzer I think you are very familiar with Qt, aren't you?  If so, perhaps you can help sort this out.

@Carreau What kind of thing is the `context.fileno()` pointing to?  I find it hard to trace back and figure out what `context` is here, and where it is created.

anntzer:
I don't have a mac readily available and I have never seen the original issue on linux :/

efiring:
Some more observations (all on OSX 10.12.6):
- Using the logging module to send debugging messages to a file, I find that the 10-s delay is occurring outside the inputhook function.
- The base number of threads is 3 without `%gui qt` and 6 with it, or 7 with qt and logging, but activity can boost the number of threads to over 30.
- When waiting for the 10 s to elapse, there are 2 additional threads (so 9 total when logging is used).

- Switching the method in the inputhook to the Windows-specific polling with a timer, I can still get a 10-s delay, and worse, I can lock up the process.

Regarding the context object: it appears to be an `InputHookContext` from prompt-toolkit, and the file object appears to be the read end of a pipe.  It looks like a thread is launched to write to that pipe, but at this point my head is spinning out of control and I have to stop looking at this.

tacaswell:
This may be related to the appnap issue: https://github.com/ipython/ipython/issues/11602

PR #11748, opened 2019-05-24
Accept matplotlib backends in uppercase

OriolAbril:
Call to `str.lower()` to accept matplotlib backends written in uppercase, i.e. "Qt5" or "GTK3"


Carreau:
Seem easy enough. Thanks.

PR #11752, opened 2019-05-27
Fixes Job Id increments. #11603

### LINKED ISSUE #11603: BackgroundJobManager wrong job.num's are assigned after flush call
Hello, there is a problem with how `job.num` is calculated.

### What I do:
I try to run the following code
```python
from IPython.lib import backgroundjobs as bg

def forever():
    while True:
        pass
    return 'Done'

def once():
    return 'Done'

jm = bg.BackgroundJobManager()
jm.new(forever)
jm.new(once)
jm.new(once)
jm.new(forever)
jm.flush()
jm.new(forever)
jm.new(once)
jm.new(once)
jm.new(once)
jm.flush()
```
### What is wrong
I got an exception when calling `jm.flush()` second time
```
KeyError                                  Traceback (most recent call last)
<ipython-input-35-a03db7172b42> in <module>
     19 jm.new(once)
     20 jm.new(once)
---> 21 jm.flush()

~/.virtualenvs/ipython3/lib/python3.7/site-packages/IPython/lib/backgroundjobs.py in flush(self)
    320         alljobs = self.all
    321         for job in self.completed+self.dead:
--> 322             del(alljobs[job.num])
    323 
    324         # Now flush these lists completely

KeyError: 4
```

### My guess
It is connected to the way job.num is calculated here: https://github.com/ipython/ipython/blob/master/IPython/lib/backgroundjobs.py#L190
```python
# when we call flush first time in the example above len(self.all) becomes 2 
job.num = len(self.all)+1 if self.all else 0 
self.running.append(job)
 # and now we count from 2 despite that we have running job with id=4 already
self.all[job.num] = job 
```

Also, there is second bug in the calculation:
we could never create a job with `id=1` as it will be 0 if `self.all` is empty or 1 + 1 if there is one job in `self.all`

### How to fix
My suggestion is to create new instance field `self._current_job_id = 0`, and increment it on every new job creation.
With that, we'll get a counter that will be tied to BackgroundJobManager instance only and not to the current jobs count

If this issue will be confirmed I probably will create a fix myself.

gooverdian:
job.num is now calculated independently of current job count and
increments throughout all jobs

### DIFF
--- a/IPython/lib/backgroundjobs.py
+++ b/IPython/lib/backgroundjobs.py
@@ -86,6 +86,7 @@ class BackgroundJobManager(object):
         self._s_running   = BackgroundJobBase.stat_running_c
         self._s_completed = BackgroundJobBase.stat_completed_c
         self._s_dead      = BackgroundJobBase.stat_dead_c
+        self._current_job_id = 0
 
     @property
     def running(self):
@@ -187,7 +188,8 @@ class BackgroundJobManager(object):
 
         if kwargs.get('daemon', False):
             job.daemon = True
-        job.num = len(self.all)+1 if self.all else 0
+        job.num = self._current_job_id
+        self._current_job_id += 1
         self.running.append(job)
         self.all[job.num] = job
         debug('Starting job # %s in a separate thread.' % job.num)


Carreau:
Seem reasonable to me; thanks !

PR #11760, opened 2019-05-30
Some preparation for pytest runner

Carreau:
Trying to migrate at least part-of our test suite to pytest. 

issues are : 
 1) globals
 2) MultipleInstanceError

For 2 we want get_ipython() to return InteractiveShell instances if they already exists; but so far it is trying to create a TerminalInteractiveShell

I'm not quite sure if I should have:

```
def get_ipython():
    return TerminalInteractiveShell.instance()
```

or 

```
def get_ipython():
    return InteractiveShell.instance()
```

The first one seem convenient; but actually we really want to test the core functionality... 

I'm guessing this might be of interest to @takluyver who I hope had fun in Paris :-) 

--- 

Much of the work in this PR was extracted and merged as part of many smaller PRs. 
This now just finishes up the work:
```
Also run test with Pytest.

This now run many of the tests using pytest.
Many of previous pull-requests brought pytest compatibility; this now
actually would run a passing test suite.

This does not mean we can drop nose; in particular we still have plugins
fro doctests using iptest; which pytest does not run;
and we are still limited to pytest 3 (pytest 4 does not support
yield-test anymore)

There is thus still much work to do
```


Carreau:
At least now the pytest test suit starts...

```
==== 104 failed, 662 passed, 12 skipped, 17 xfailed, 147 warnings, 14 error in 10.94 seconds === 
```

With the `--duration` flag I was also able to find a couple of super-slow test and reduce the test-suite by about 10-15 seconds (out of a minute)

Still pass with iptest -j6 (6 parallel workers) on my local machine in 12sec.

Carreau:
2 remaining failing test  ! (need pytest 3.0 as 4.0 does not support yield-tests)

PR #11764, opened 2019-05-31
Support os.PathLike in Image

alok:
`os.path.splitext` was all that was needed. Any further processing is already handled with `os`'s support for `os.PathLike`.


Carreau:
Well, thanks, was pretty straitforward !

Issue #11749, opened 2019-05-25
cleanup >3.5 conditional.


Carreau:
We still have some code to handle pre-3.5 Python; 
It is not supported anymore so can be removed.

gdhameeja:
If open, I'd like to work on this.

kd2718:
Does this still need to be done? Is there an example of cleanup code you can show to point in the right direction?

kd2718:
@Carreau looks like [this](https://github.com/ipython/ipython/blob/a9a72474c49d9c0a90ace5ee332ddd9d490b1a98/IPython/core/interactiveshell.py#L2261) is the kind of code you want cleaned up?

Carreau:
Sorry I missed these comments; Yes that was the kind of things to simplify. This one has been done. There are a few other; I've opened #11954 and #11955 for example. 

Thanks.

PR #11768, opened 2019-06-01
Upgrade to python 3.5+ syntax, add autoformat

alok:
Addresses #11749. If autoformat (done with `black`) is not desired, I can remove
that.

If it *is* desired, then I still need to add it to Travis to enforce that people
use it. One issue is that it requires 3.6+ to run.


LucianaMarques:
Hi @alok ! My latest PR changed the optestcontroller.py file, so you probably need to rebase your local repository.

Carreau:
Thanks ! I'll try to have a look when I can !

> I can remove that.

Well we'd love to enforce black, but we don't want to modify all the files at once as it would break some things like `git blame`; see among other [this discussion](https://github.com/python/black/issues/830) and [this thread on twitter](https://twitter.com/F_Vaggi/status/1133425252992557056); if you don't mind removing/rebasing that would be great.

alok:
@Carreau I rebased. Though if you wouldn't mind updating the tests, that would be nice since I have no idea what the mini-language for test case formatting is.

Carreau:
> since I have no idea what the mini-language for test case formatting is.

Not quite sure what mini language you are referring to. 
Also in #11749 I was referring to actuall checking of Python version like [here](https://github.com/ipython/ipython/blob/a9a72474c49d9c0a90ace5ee332ddd9d490b1a98/IPython/core/interactiveshell.py#L2261), [here](https://github.com/ipython/ipython/blob/a9a72474c49d9c0a90ace5ee332ddd9d490b1a98/IPython/core/interactiveshell.py#L2261) ,

Ss this touch ~150 files, I might end up integrating it in multiple step. Did you use a script or changed all by hand ?

alok:
Nevermind, it was an issue that came up with `black`.

I just ran `pyupgrade --py3-plus **.py`. That was it.

drunkwcodes:
This PR may really break python2.7 compatibility in `master` such as string prefix u deletion.
Would you like to add a warning for python2 users pinning the version to 5?

And it is better to remain [PEP461](https://www.python.org/dev/peps/pep-0461/) string formatting unchanged for the target python version is 3.5.

alok:
@drunkwcodes I'm not going to do any of that. 2.7 support has been dropped since 6.0, which came out 2 years ago.

drunkwcodes:
@alok Fine. But [PEP461 formatting?](https://www.python.org/dev/peps/pep-0461/)

alok:
What is that

drunkwcodes:
@alok That is the proof of unnecessity about the change of string formatting.
Reverting those changes and `u` prefix deletion makes this PR less possible to be rejected.

PR #11769, opened 2019-06-01
Allow passing pathlib Paths without keyword arg

alok:
Don't need to use `filename=` now. Hacky but easier than the alternative of
`os.PathLike` (which doesn't exist in 3.5).

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -14,6 +14,7 @@ import sys
 import warnings
 from copy import deepcopy
 from os.path import splitext
+from pathlib import Path, PurePath
 
 from IPython.utils.py3compat import cast_unicode
 from IPython.testing.skipdoctest import skip_doctest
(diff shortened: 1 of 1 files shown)


IPython/core/display.py
```diff
@@ -594,6 +595,9 @@ def __init__(self, data=None, url=None, filename=None, metadata=None):
         metadata : dict
             Dict of metadata associated to be the object when displayed
         """
+        if isinstance(data, (Path, PurePath)):
```
LucianaMarques: Hi there. 

Just a suggestion: you are repeating 4 times the same code here. Why don't create a separate function and use it instead? I believe this way code will be cleaner.

Carreau: That would not shorten the code that much, so I'm not sure it would be worth it for 4 instances; 

For a reader it might also interrupt the flow and have them check what this hypothetical function would do.

I believe in this case inlining the code is good enough.

PR #11776, opened 2019-06-06
Use CodeType.replace with Python 3.8

blueyed:
Via https://bugs.python.org/issue37032.


blueyed:
Followup to https://github.com/ipython/ipython/commit/248128dfaabb33e922b1e36a298fd7ec0c730069.
/cc @stonebig

Carreau:
Good call ! Now that Python 3.8b1 is out that seen reasonable.

PR #11779, opened 2019-06-11
Fix state leakage and speedup in test-suite

Carreau:
This is mostly visible when running the test via pytest that
collect-then-run vs nose that runs as it collects.

While I was at it I've improved some tests to be faster by changing the
maximum recursion limit.


Carreau:
@takluyver this touch recursion detection code, which I believe you wrote; 

@minrk and @takluyver 

Note that this uses `setup_module` instead of `setUp` which is understood by both nose and pytest; so I'm slowly switching to this form. 

This is a subset of #11760 which attempt to run the test suite with pytest.

minrk (approved):

PR #11704, opened 2019-04-28
Simplify _completer() constr. in core/completer.py

LucianaMarques:
Exclude the text field from the constructor and made modifications
accordingly to this new requirement.

This is part of #11622, still much more to do. I excluded the text field from the _complete() function and modified the class' constructor accordingly.

PR #11804, opened 2019-06-27
Massive completer cleanup - Jedi

gpotter2:
This PR makes `jedi` compulsory, and removes the older API.

**Of course: this change is super breaking. Maybe for 8.0.0 ?**

This should have been done long ago: the current state of `completer.py` is atrocious, with tons of "provisional" stuff, waiting since 6.0.0 to be removed

- **Enforce jedi**: remove all functions already implemented in jedi, massive cleanup
- Drop `_complete()` API
- add a `priority`, and `alone` parameter to the output of completers, to be able to sort results by priority when necessary, or mark a completer as "standalone" when necessary: this allows to reuse the same code for all the completers (magic, unicode...)

Here's a yolo patch to apply on `ipykernel` to try this out in some broader environment:

```diff
diff --git a/ipykernel/ipkernel.py b/ipykernel/ipkernel.py
index 2cc8eab..3b5a55e 100644
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -24,15 +24,6 @@ try:
 except ImportError:
     _asyncio_runner = None
 
-try:
-    from IPython.core.completer import (
-        rectify_completions as _rectify_completions,
-        provisionalcompleter as _provisionalcompleter,
-    )
-    _use_experimental_60_completion = True
-except ImportError:
-    _use_experimental_60_completion = False
-
 try:
     import debugpy
     from .debugger import Debugger
@@ -403,46 +394,18 @@ class IPythonKernel(KernelBase):
         return reply_content
 
     def do_complete(self, code, cursor_pos):
-        if _use_experimental_60_completion and self.use_experimental_completions:
-            return self._experimental_do_complete(code, cursor_pos)
-
-        # FIXME: IPython completers currently assume single line,
-        # but completion messages give multi-line context
-        # For now, extract line from cell, based on cursor_pos:
         if cursor_pos is None:
             cursor_pos = len(code)
-        line, offset = line_at_cursor(code, cursor_pos)
-        line_cursor = cursor_pos - offset
+        completions = list(self.shell.Completer.completions(code, cursor_pos))
 
-        txt, matches = self.shell.complete('', line, line_cursor)
-        return {'matches' : matches,
-                'cursor_end' : cursor_pos,
-                'cursor_start' : cursor_pos - len(txt),
-                'metadata' : {},
-                'status' : 'ok'}
-
-    async def do_debug_request(self, msg):
-        if _is_debugpy_available:
-            return await self.debugger.process_request(msg)
-
-    def _experimental_do_complete(self, code, cursor_pos):
-        """
-        Experimental completions from IPython, using Jedi.
-        """
-        if cursor_pos is None:
-            cursor_pos = len(code)
-        with _provisionalcompleter():
-            raw_completions = self.shell.Completer.completions(code, cursor_pos)
-            completions = list(_rectify_completions(code, raw_completions))
-
-            comps = []
-            for comp in completions:
-                comps.append(dict(
-                            start=comp.start,
-                            end=comp.end,
-                            text=comp.text,
-                            type=comp.type,
-                ))
+        comps = []
+        for comp in completions:
+            comps.append(dict(
+                        start=comp.start,
+                        end=comp.end,
+                        text=comp.text,
+                        type=comp.type,
+            ))
 
         if completions:
             s = completions[0].start
@@ -459,6 +422,10 @@ class IPythonKernel(KernelBase):
                 'metadata': {_EXPERIMENTAL_KEY_NAME: comps},
                 'status': 'ok'}
 
+    async def do_debug_request(self, msg):
+        if _is_debugpy_available:
+            return await self.debugger.process_request(msg)
+
     def do_inspect(self, code, cursor_pos, detail_level=0):
         name = token_at_cursor(code, cursor_pos)
 
diff --git a/setup.py b/setup.py
index 1cb41b4..586a861 100644
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup_args = dict(
         'importlib-metadata<5;python_version<"3.8.0"',
         'argcomplete>=1.12.3;python_version<"3.8.0"',
         'debugpy>=1.0.0,<2.0',
-        'ipython>=7.23.1,<8.0',
+        'ipython>=7.23.1',
         'traitlets>=4.1.0,<6.0',
         'jupyter_client<8.0',
         'tornado>=4.2,<7.0',
```

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1,9 +1,5 @@
 """Completion for IPython.
 
-This module started as fork of the rlcompleter module in the Python standard
-library.  The original enhancements made to rlcompleter have been sent
-upstream and were accepted as of Python 2.3,
-
 This module now support a wide variety of completion mechanism both available
 for normal classic Python code, as well as completer for IPython specific
 Syntax like magics.
@@ -76,7 +72,6 @@ You will find that the following are experimental:
     - :any:`provisionalcompleter`
     - :any:`IPCompleter.completions`
     - :any:`Completion`
-    - :any:`rectify_completions`
 
 .. note::
 
@@ -123,6 +118,8 @@ import time
 import unicodedata
 import uuid
 import warnings
+
+from collections import defaultdict
 from contextlib import contextmanager
 from importlib import import_module
 from types import SimpleNamespace
@@ -144,14 +141,12 @@ import __main__
 # skip module docstests
 skip_doctest = True
 
-try:
-    import jedi
-    jedi.settings.case_insensitive_completion = False
-    import jedi.api.helpers
-    import jedi.api.classes
-    JEDI_INSTALLED = True
-except ImportError:
-    JEDI_INSTALLED = False
+import jedi
+
+jedi.settings.case_insensitive_completion = False
+import jedi.api.helpers
+import jedi.api.classes
+
 #-----------------------------------------------------------------------------
 # Globals
 #-----------------------------------------------------------------------------
@@ -329,7 +324,7 @@ def completions_sorting_key(word):
             word = word[1:]
             prio2 = 1
 
-    return prio1, word, prio2
+    return (prio1, prio2), word
 
 
 class _FakeJediCompletion:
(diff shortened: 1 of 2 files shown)


gpotter2:
@LucianaMarques Hi ! Feel free to have a look at this. I think it's a better approach to rewrite the whole thing and enforce `jedi`, than perform small cleanups (https://github.com/ipython/ipython/pull/11704). You all know IPython better than I do, therefore any advices will be greatly appreciated 😄

ccordoba12:
> This PR makes jedi compulsory

This is really not good. Jedi makes completions lag a lot when introspecting large dataframes (for instance), and that's why we have its completions deactivated by default in Spyder right now.

@Carreau, please be careful with this one.

gpotter2:
Thanks for your note. I didn't know this.

I've asked what was the roadmap over https://github.com/ipython/ipython/pull/11737#issuecomment-493683869, but didn't get any answer (and I understand that it's time consuming to manage such a big project).

I however still think that this file needs a massive rework. If jedi is the issue, couldn't we either ask them to fix the issue (if possible) or drop it ?

ccordoba12:
> I however still think that this file needs a massive rework

Agreed.

>  If jedi is the issue, couldn't we either ask them to fix the issue (if possible) or drop it ?

Please test your changes with a large dataframe (say 1 million rows and 10 columns) to see if that's still the case. If it is, then an issue should be opened in the Jedi repo and you should try to solve it, so this PR could be merged in the near future.

Carreau:
Thanks for looking into this ! I would love to actually find some funding and get proper work done on the completions API; potentially upstream in Jedi. The Jedi maintainer is doing its best, but it will take time to get all the kinks out.

Yes the issues that @ccordoba pointed out are real and for now it would be really nice to still have a Jedi flag. 

This will likely need to be tested with ipykernel/notebook and various frontends; and I think that such a change would be good for an IPython 8.0 – I'm also trying to get the test-suite moved to Pytest;

Sorry I know I'm slow to respond; and I want to figure out how to be less of a bottle neck.

LucianaMarques:
Hey guys!

This kind of work was actually what I was trying to do, but I think we need to talk a little bit more about design. Also, the completer class is called in lots of other files, so we need to have this in mind. 

Is it ok if we work together on this? I have already done some homework on it, I just hadn't had the time to open new PRs because of my university workload.

LucianaMarques:
@gpotter2 but awsome work! :D

I guess we can start by writing some tests. Again, can we work together on this? I'm available for calls or chatting on discourse/hangouts/skype, etc

IPython/core/completer.py
```diff
@@ -124,6 +123,7 @@
 import string
 import warnings
 
+from collections import defaultdict
```
LucianaMarques: Could you help me and tell me what is this that you are importing?

IPython/core/completer.py
```diff
@@ -142,14 +142,11 @@
...
 
-try:
-    import jedi
-    jedi.settings.case_insensitive_completion = False
-    import jedi.api.helpers
-    import jedi.api.classes
-    JEDI_INSTALLED = True
-except ImportError:
-    JEDI_INSTALLED = False
+import jedi
```
LucianaMarques: Yup, this is dangerous. Perhaps you can revert to the old way temporarily?

LucianaMarques (commented):
I like the fact that there are less lines here, but this file is still too big in my opinion. We could improve this with some design refactoring, but I don't the best way to do this is in a single PR. I believe the best way to do it is to break what we need to do in smaller steps and do it in weekly sprints. We could write together a project proposition for the community, I think it's best to have more people taking a look at it. This could break code very easily.

gpotter2:
If we want to keep non-jedi compatibility, the whole PR is flawed.
I appreciate your attention :-) but it will probably be a waste of time to start reviewing this after what has been said above, before we all agree on what to do.

I'll perform some tests with very big buffers when I find the time, and will return to this

anntzer:
FWIW another case where jedi is less than optimal is https://github.com/anntzer/ipython-autoimport, which allows one to type e.g. `foo.b<TAB>` and get an autoimport of foo followed by autocompletion from the `foo` namespace (see also https://github.com/anntzer/ipython-autoimport#limitations).

Perhaps this could be toggled from the jedi side if it allows escape hatches to eval the being-completed code, though.

IPython/core/completer.py
gpotter2: `defaultdict` is a modified version of `dict` available in the default `collections` package.
It allows to specify a default value that is assigned to each key that doesn't exist.

```
a = defaultdict(lambda: True)
assert a["hello"] == True
```

IPython/core/completer.py
gpotter2: It's not that dangerous, since Jedi is an official IPython dependency (https://github.com/ipython/ipython/blob/5073312daa51b62aceb06626cf265d2da1836c7a/setup.py#L188)

For this PR I've opted for the "go fast break things" method. It'll require lots of testing on various frontends, as Carreau said above, but at least it will clean this up quickly.

gpotter2:
@ccordoba12 Hello & sorry for bothering.
The current code (I'll update the tests later) is working quite fine on IPython.

Could you please point me to where/how Spyder calls IPython completion ? Which method does it use among the few available ? I think I might have broken the integration

Thanks !

LucianaMarques:
@gpotter2 you are doing a fine job here! I still have some design ideas, though, and i would like to work together with you instead of waiting for this PR to be accepted, since it may still take a while. Please send me an e-mail if you don't mind working together :) lucianadacostamarques@gmail.com

gpotter2:
@LucianaMarques Thanks ! I'll be happy to work with you. However I believe that keeping most of the talk public (except if required), is more effective. How about creating a separate issue to talk about that ?

LucianaMarques:
Will do it, sorry for the delay!

gpotter2:
closed for inactivity

gpotter2:
I brought this PR back from the dead as 8.0.0 is near, happy to hear some feedback...
I saw that Spyder was moving away from IPython's completion, maybe now is a good time for breaking change?

MrMino:
This will not be possible any time soon and has bitrot considerably since posting.

I'd rather do the cleanup by abstracting the completion engine away completely, providing a way of plugging in your own, and moving the old machinery to a separate package. In that arrangement, IPython stays with Jedi by default, and Spyder can use the legacy completion system on its own, until Jedi can figure out how to be more performant or customizable.

PR #11806, opened 2019-06-28
Fixing duplicated definition of magic aliases after reset

### LINKED ISSUE #11646: Duplicated definition of built-in magic aliases after %reset
When running `%reset -f` in Jupyter notebook or Spyder, the following message will be printed.
```
ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.
ERROR:root:Invalid alias: The name more can't be aliased because it is another magic command.
ERROR:root:Invalid alias: The name less can't be aliased because it is another magic command.
ERROR:root:Invalid alias: The name man can't be aliased because it is another magic command.
```
However, running `%reset -f` in IPython itself is just fine.

I started experiencing this after upgrading to IPython 7.3.0, and I believe that this behavior was from the change #11528. Maybe those aliases should only be redefined when running in the IPython shell, not in Jupyter or Spyder as a kernel.

juanis2112:
Fixes #11646 by avoiding the double definition of the magic aliases.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1435,9 +1435,9 @@ class InteractiveShell(SingletonConfigurable):
         drop_keys.discard('__name__')
         for k in drop_keys:
             del ns[k]
-        
+
         self.user_ns_hidden.clear()
-        
+
         # Restore the user namespaces to minimal usability
         self.init_user_ns()
 
@@ -1450,9 +1450,7 @@ class InteractiveShell(SingletonConfigurable):
         # GUI or web frontend
         if os.name == 'posix':
             for cmd in ('clear', 'more', 'less', 'man'):
-                try:
-                    name = self.magics_manager.magics['line'][cmd]
-                except KeyError:
+                if cmd not in self.magics_manager.magics['line']:
                     self.alias_manager.soft_define_alias(cmd, cmd)
 
         # Flush the private list of module references kept for script
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1450,7 +1450,10 @@ class InteractiveShell(SingletonConfigurable):
         # GUI or web frontend
         if os.name == 'posix':
             for cmd in ('clear', 'more', 'less', 'man'):
-                self.alias_manager.soft_define_alias(cmd, cmd)
+                try:
+                    name = self.magics_manager.magics['line'][cmd]
+                except KeyError:
+                    self.alias_manager.soft_define_alias(cmd, cmd)
 
         # Flush the private list of module references kept for script
         # execution protection
--- a/IPython/core/tests/test_magic.py
+++ b/IPython/core/tests/test_magic.py
@@ -371,6 +371,31 @@ def test_reset_in_length():
     _ip.run_cell("reset -f in")
     nt.assert_equal(len(_ip.user_ns['In']), _ip.displayhook.prompt_count+1)
 
+class TestResetErrors(TestCase):
+
+    def test_reset_redefine(self):
+
+        @magics_class
+        class KernelMagics(Magics):
+              @line_magic
+              def less(self, shell): pass
+
+        _ip.register_magics(KernelMagics)
+
+        with self.assertLogs() as cm:
+            # hack, we want to just capture logs, but assertLogs fails if not
+            # logs get produce.
+            # so log one things we ignore.
+            import logging as log_mod
+            log = log_mod.getLogger()
+            log.info('Nothing')
+            # end hack.
+            _ip.run_cell("reset -f")
+
+        assert len(cm.output) == 1
+        for out in cm.output:
+            assert "Invalid alias" not in out
+
 def test_tb_syntaxerror():
     """test %tb after a SyntaxError"""
     ip = get_ipython()


Carreau:
Hi @juanis2112 !

Thanks for your PR ! I'll try to review it now.

IPython/core/interactiveshell.py
```diff
@@ -1450,7 +1450,10 @@ def reset(self, new_session=True):
         # GUI or web frontend
         if os.name == 'posix':
             for cmd in ('clear', 'more', 'less', 'man'):
-                self.alias_manager.soft_define_alias(cmd, cmd)
+                try:
+                    name = self.magics_manager.magics['line'][cmd]
+                except KeyError:
+                    self.alias_manager.soft_define_alias(cmd, cmd)
```
Carreau: Do you have a reasoning to use `try:...Except KeyError` instead of checking:

```
if cmd not in self.magics_manager.magics['line']:
    self.alias_manager.soft_define_alias(cmd, cmd)
```

I would tend to have a preference for the `if ...` version , as `try/except` could swallow bugs in `soft_define_alias`, but I could easily be convinced.

Carreau:
That is great ! I'll see if I can add a test to make sure there is no later regressions ! Unless you want to add a test; I'm happy to guide you in writing one if you desire.

ccordoba12:
Thanks a lot for this @juanis2112!

@Carreau, @juanis2112 is working with me in Spyder and I asked her to fix #11646 because it's really annoying for Spyder users. Could you guide her to write a regression test for this? Thanks!

IPython/core/interactiveshell.py
ccordoba12: I agree with @Carreau here: I think it's better to use the `if`version and not the `try/except` one here.

Carreau:
> Could you guide her to write a regression test for this? Thanks!

Of course I will. 

Side question; it seem GH allow us to give only triage permission to repositories (close issues and manage labels); would that be of interest to you @ccordoba12 ?

juanis2112:
@Carreau  I'll be happy to help with the test, with a bit of guidance as @ccordoba12 suggested.

ccordoba12:
> would that be of interest to you ?

Sorry, but no. I do the same job for almost all repos in the Spyder organization, so I really don't have time to help you with that.

Carreau:
> Sorry, but no. I do the same job for almost all repos in the Spyder organization, so I really don't have time to help you with that.

It was morre incase there was spyder specific issues here so that you could properly tag them; and more giving more rights that requesting more from you.

> I'll be happy to help with the test

Thanks, I'll try to lay out what I believe we should try to do.

1) the issue appear only when magics are already registered with the same name, which mean i out test we will wan to [register a fake magic](https://github.com/ipython/ipython/blob/master/IPython/core/tests/test_magic.py#L97-L99).

2) We also likely want to use [`AssertNotPrints`](https://github.com/ipython/ipython/blob/master/IPython/core/tests/test_magic.py#L425-L428) to make sure it does not prints said mesage.

3) We can likely put a test in `IPython/core/tests/test_magic.py`.

Carreau:
Ok, so I looked into it a bit more carefully and it is slightly more complicated to write test, so here is what I came up with:

```diff
modified: IPython/core/tests/test_magic.py
@ test_magic.py:374 @ def test_reset_in_length():
     _ip.run_cell("reset -f in")
     nt.assert_equal(len(_ip.user_ns['In']), _ip.displayhook.prompt_count+1)

+class TestResertFErrors(TestCase):

+    def test_reset_redefine(self):

+        @magics_class
+        class KernelMagics(Magics):
+              @line_magic
+              def less(self, shell): pass

+        _ip.register_magics(KernelMagics)

+        with self.assertLogs() as cm:
+            # hack, we want to just capture logs, but assertLogs fails if not
+            # logs get produce.
+            # so log one things we ignore.
+            import logging as log_mod
+            log = log_mod.getLogger()
+            log.info('Nothing')
+            # end hack.
+            _ip.run_cell("reset -f")

+        assert len(cm.output) == 1
+        for out in cm.output:
+            assert "Invalid alias" not in out

 def test_tb_syntaxerror():
     """test %tb after a SyntaxError"""
     ip = get_ipython()
```

You can use it as-is, or try to come up with a better way, but it's tricky enough that I can't request you to write an annoying test case like that. 

If you just want to change your code from `try/except` to `if cmd not in: self.magics_manager.magics['line']:`, I can take care of adding the test, and we can get that in for 7.7.

Carreau:
@juanis2112 You've done some great contribution, as I wan to release 7.7. soon; I've pushed the test on your branch; and also switched from Try-except to `if ... in`. I'm going to merge to make sure this contribution get into the next release. 

Thanks Again and looking forward to your next contribution.

ccordoba12:
@Carreau, thanks a lot for helping @juanis2112 to finish this PR!! I instructed her to work on other Spyder issues in the meantime and we totally forgot about this!

ccordoba12:
> It was morre incase there was spyder specific issues here so that you could properly tag them; and more giving more rights that requesting more from you.

In that case I accept your offer. I didn't know about the new Github functionality you mentioned above, and I think it'd be really useful in this case.

Carreau:
> I instructed her to work on other Spyder issues in the meantime and we totally forgot about this!

No problem, the hard part of the work was down, and it was only annoying details to finish. 



> In that case I accept your offer. I didn't know about the new Github functionality you mentioned above, and I think it'd be really useful in this case.

I'm not sure the feature is available on all orgs yet, hence why you might not have seen it.Ok, I'll invite you to the right team for that; let me know if you know any body else that would like to be added to triage.

PR #11807, opened 2019-06-28
[WIP] Crush all pngs to reduce sdist size.

Carreau:
Using "Magic" I also found some PNGs that MIGHT be unused.

```
$ diff -U0 <(find . -name '*.png'|xargs -L1 basename |sort ) <(find . -name '*.png' | xargs -L1 basename  | xargs -L1  rg | rev| cut -f1 -d'/'| rev| sort | uniq) | grep '^-'
--- /dev/fd/63	2019-06-28 16:52:14.000000000 -0700
-2x2.png
-2x2.png
-FrontendKernel.png
-besselj.png
-colors_dark.png
-image1.png
-ipy_013_qtconsole_baboon.png
-ipython_logo.png
-logo.png
-running.png
```

Should be checked manually

We could also remove some of these images from the source and replace the image in the docs by "see older docs build" as those are mostly references in what's new, and old changelog.


Carreau:
This get he sdist from 5.1 to 4.6 MB.

Carreau:
Need to rebase and amend the commit message that refer to wheel.

Carreau:
Crushed more; at the same time I'm unsure this is necessary; the tar.gz download are a hundreds of the whl download, and switching to `xz` instead of `tar.gz` also gain an extra 300kb.

So maybe I should just ignore that.

Carreau:
closing I believe now it is not worth it.

PR #11808, opened 2019-06-29
Fix manpage path in FreeBSD

sunpoet:
FreeBSD stores manpages under ${PREFIX}/man (defaults to /usr/local/man)
rather than ${PREFIX}/share/man.


Carreau:
If that helps !

Thanks !

sunpoet:
Thanks!

Issue #11809, opened 2019-06-30
getting "AttributeError: 'For' object has no attribute 'value'" with ipython=7.6.0


fehiepsi:
Hi ipython dev, I am facing an issue with the latest version of ipython (7.6.0), python 3.6. The following repro script shows the issue in a Jupyter notebook cell:
```
%%time
x = 1
for i in range(3):
    x += 1
```
It used to work fine with version 7.5.0.

Carreau:
Do you mind trying #11812 ? If it fixes it; I'll try to make a release before Friday.

fehiepsi:
Thanks @Carreau, #11812 works great in my tests!

Carreau:
Ok, great !

I'll wait for CI t pass, travis seem to have some backlog; and release soon.

Carreau:
7.6.1 should be released now with the fix.

fehiepsi:
Thanks @Carreau !

PR #11812, opened 2019-07-02
Fix time magic, 

### LINKED ISSUE #11811: IPython 7.6 problem with %%time
In IPython 7.6, the `%%time` magic is messing with the workspace. Looks like the last variable is never assigned in the workspace. This is a pretty serious error.

IPython 7.5.0 worked fine, so I assume it's related to https://github.com/ipython/ipython/pull/11716/. Also works fine when there's exactly one statement inside the cell.

```
In [1]:
%%time
a = 1
b = 2
c = 3
```

Wall time: 0 ns

```
In [2]:
a, b, c
```

```
<ipython-input-2-6fd489e9d195> in <module>
----> 1 a, b, c

NameError: name 'c' is not defined
```

Carreau:
Time magic was eagerly assuming the last node in an ast tree was an Expr, which lead t crashes.

Fixes #11811 #11809

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1281,13 +1281,11 @@ python-profiler package from non-free.""")
             mode = 'exec'
             source = '<timed exec>'
             # multi-line %%time case
-            if len(expr_ast.body) > 1 :
-                expr_val=expr_ast.body[-1]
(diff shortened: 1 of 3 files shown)


kernc:
Tested, works. Thanks for promptness!

PR #11814, opened 2019-07-04
Fix %matplotlib

OriolAbril:
Hey, I did a PR some days ago (#11748) in order to allow `%matplotlib` to accept uppercase backend names. However, I have just realized that I totally overlooked the fact that `args.gui` may not be a `str`. I think it can only be none or a string, but I feel this solution is safer. I am really sorry because now `%matplotlib` with no backend does not work. 

@Carreau

### DIFF
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -37,7 +37,7 @@ magic_gui_arg = magic_arguments.argument(
 @magics_class
 class PylabMagics(Magics):
     """Magics related to matplotlib's pylab support"""
-    
+
     @skip_doctest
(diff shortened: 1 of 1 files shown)


OriolAbril:
Oh, vim removed all the spaces it deemed as unnecessary. Should I undo this change to leave the diff as one line?

Carreau:
> Oh, vim removed all the spaces it deemed as unnecessary. Should I undo this change to leave the diff as one line?

Well, usually I don't like it, but it happens often enough that removing them if fine and will prevent it from happening again.

Carreau:
Thanks !

PR #11832, opened 2019-07-22
no mention os setup.py install

Carreau:


Carreau:
Thanks @drunkwcodes for the 👍

PR #11840, opened 2019-07-30
Enable changing the font color for LaTeX rendering

### LINKED ISSUE #9451: LaTeX color should be white when background is dark
[Change the color of Latex output in IPython (Jupyter)](http://stackoverflow.com/q/35539492/125507) has an example from SymPy:

njbxw 1

here is an example of Pint using MikTex:

2016-05-09 19_34_24-spyder python 2 7

and using matplotlib (I think, since miktex is uninstalled):

matplotlib

[Control LaTeX expression color in ipython qtconsole](http://stackoverflow.com/a/18260452/125507) gives a workaround, but it only applies to SymPy, not other objects that define `_repr_latex_()`, and doesn't work for matplotlib backend.

Ideally the Latex would just be printed in the same color as the plain text.

oscargus:
An issue that shows up in quite a lot of places is that SymPy rendering doesn't consider the background color.

IPython: https://github.com/ipython/ipython/issues/9451 
Spyder: https://github.com/spyder-ide/spyder/issues/3798
SymPy: https://github.com/sympy/sympy/issues/10655

This PR makes it possible to send along a dvips color name (default is black), which will make it a bit easier for all these tools to do something clever.

One could considering adding an "auto" default option as well, see https://github.com/sympy/sympy/pull/17244, but it wasn't clear if it was a good idea and/or possible to get the colors variable from lib.latextools (I imagine it is sometimes used standalone?). This would effectively (or at least in a statistical sense) close https://github.com/ipython/ipython/issues/9451 

If it makes sense, I can easily add a `scale` keyword argument as well to deal with high dpi screens. SymPy just got one: https://github.com/sympy/sympy/pull/17298

### DIFF
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -22,7 +22,7 @@ class LaTeXTool(SingletonConfigurable):
     """An object to store configuration of the LaTeX tool."""
     def _config_default(self):
         return get_config()
-    
+
     backends = List(
         Unicode(), ["matplotlib", "dvipng"],
         help="Preferred backend to draw LaTeX math equations. "
@@ -55,7 +55,7 @@ class LaTeXTool(SingletonConfigurable):
         ).tag(config=True)
 
 
-def latex_to_png(s, encode=False, backend=None, wrap=False):
+def latex_to_png(s, encode=False, backend=None, wrap=False, color='Black'):
     """Render a LaTeX string to PNG.
 
     Parameters
@@ -68,6 +68,8 @@ def latex_to_png(s, encode=False, backend=None, wrap=False):
         Backend for producing PNG data.
     wrap : bool
         If true, Automatically wrap `s` as a LaTeX equation.
+    color : string
+        Foreground color name among dvipsnames.
 
     None is returned when the backend cannot be used.
 
@@ -84,13 +86,13 @@ def latex_to_png(s, encode=False, backend=None, wrap=False):
         f = latex_to_png_dvipng
     else:
         raise ValueError('No such backend {0}'.format(backend))
-    bin_data = f(s, wrap)
+    bin_data = f(s, wrap, color)
     if encode and bin_data:
         bin_data = encodebytes(bin_data)
     return bin_data
 
 
-def latex_to_png_mpl(s, wrap):
+def latex_to_png_mpl(s, wrap, color='Black'):
     try:
         from matplotlib import mathtext
         from pyparsing import ParseFatalException
@@ -105,13 +107,13 @@ def latex_to_png_mpl(s, wrap):
     try:
         mt = mathtext.MathTextParser('bitmap')
         f = BytesIO()
-        mt.to_png(f, s, fontsize=12)
+        mt.to_png(f, s, fontsize=12, color=color)
         return f.getvalue()
     except (ValueError, RuntimeError, ParseFatalException):
         return None
 
 
-def latex_to_png_dvipng(s, wrap):
+def latex_to_png_dvipng(s, wrap, color='Black'):
     try:
         find_cmd('latex')
         find_cmd('dvipng')
@@ -133,8 +135,8 @@ def latex_to_png_dvipng(s, wrap):
 
             subprocess.check_call(
                 ["dvipng", "-T", "tight", "-x", "1500", "-z", "9",
-                 "-bg", "transparent", "-o", outfile, dvifile], cwd=workdir,
-                stdout=devnull, stderr=devnull)
+                 "-bg", "transparent", "-o", outfile, dvifile, "-fg", color],
+                 cwd=workdir, stdout=devnull, stderr=devnull)
 
         with open(outfile, "rb") as f:
             return f.read()
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -10,6 +10,7 @@ import tempfile
 import shutil
 import subprocess
 from base64 import encodebytes
+from textwrap import wrap as splitstring
 
 from IPython.utils.process import find_cmd, FindCmdError
 from traitlets.config import get_config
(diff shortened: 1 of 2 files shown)


oscargus:
Oh, it wasn't really clear what to test. But at least some exercise is done to see that the results are consistent and that dvips name can be used. (I have manually confirmed if for matplotlib.)

oscargus:
Clearly, there is no latex installed in the test environment...

Ahh, I saw the markers now... Will add those.

Carreau:
Hi ! Thanks for the t=pull-request; 

We should have a better story to skip test with latex. Worst cas, just comment the test; this look strait forward enough.

oscargus:
I've just added a PR in SymPy that includes a color representation converter: https://github.com/sympy/sympy/pull/17319

Maybe it makes sense to use that here as well? If not, one problem facing a user is that colors are (except for the 68 dvips colors) specified differently depending on which backend is used. And as one aspect is that one shouldn't have to worry about that when calling `latex_to_png`, it seems like the option otherwise is to provide two keyword arguments, one for color in case dvipng is used, one for color in case matplotlib is used. (I haven't really gotten into how to possibly use other backends.)

Also, for e.g. qtconsole/Spyder it is probably easier to provide colors in hex rgb than looking up the closest dvips color name. So unless you have some insights why I shouldn't do it, I aim to include the color conversion here as well (unfortunately the color conversion in Matplotlib doesn't play well with the LaTeX special color format, so another option is of course to change that so it can handle numerical LaTeX colors...).

oscargus:
Of course, I can add a subset of the conversions here. Just enough to get from hex rgb to something that LaTeX/dvipng can handle. In that case one can either specify a named LaTeX color or as hex rgb. Should probably be enough for most people. (SymPy users can still specify using any of the numerical LaTeX color formats.)

oscargus:
Hmm, appears as if I made some small changes anyway.

It is now possible to provide hex RGB colors as well (which are converted if dvi png is used). I also added a scale optional keyword to (yes!) scale the output. Finally, I ead the dvipng manual: "-x
num This option is deprecated; it should not be used.  It is much better to select the
output resolution directly with the -D option." So I used the -D option.

Carreau:
Thanks for your patience, i've been quite busy. 

This looks good to me, so I'm happy to get that in. It also works quite well with a current in progress feature that allow to display inline images in terminal that support them: 

Screen Shot 2019-08-07 at 6 28 38 PM

ccordoba12:
@Carreau, thanks a lot for taking a look at this one! We need this in Spyder to correctly adjust Sympy equation colors in our consoles.

When do you plan to release a new IPython version that includes this fix?

oscargus:
@Carreau No problem! It's been an issue for years, so a week or two is no big deal. :-) Do you think it may be possible/wanted to detect the IPython `colors` variable from here? Running SymPy from a plain ipython shell will probably not benefit from it (but then I am not certain at all how things are interconnected). Although, e.g., qtconsole will as they can set it based on the background color knowledge.

@ccordoba12 It is worth noting that running `init_printing` in SymPy (which Spyder does for a SymPy shell) will have SymPy override this, as it will send a PNG which qtconsole will show instead of rendering its own. However, it can be sorted by providing the color to `init_printing`. Still, in a plain console, before possibly executing `init_printing`, it will definitely help.

ccordoba12:
> However, it can be sorted by providing the color to init_printing

Yes, that's exactly what we need to do! However, we need to communicate the kernel if the console background color is dark or light (and if there are changes to it), and then re-execute `init_printing` accordingly.

That's our next task after spyder-ide/spyder#9343 is merged, which will make that process very simple.

IPython/lib/latextools.py
```diff
@@ -10,6 +10,7 @@
 import shutil
 import subprocess
 from base64 import encodebytes
+from textwrap import wrap as splitstring
```
ccordoba12: This is kind of non-standard, so I'd prefer if you use simply `wrap` here instead of using an alias.

Carreau:
> When do you plan to release a new IPython version that includes this fix?

I do a minor release every last Friday of the month (when I can), this is likely to make the cut to this month. 

> Do you think it may be possible/wanted to detect the IPython `colors` variable from here

Let's do that in a further iteration. I'm planning on adding a `text/latex` handler to IPython itself that could then decide on the color and that would alleviate sympy's needs to set the color or generate PNG, and then the frontend can just pickup the math in display in whatever color they wish.

IPython/lib/latextools.py
Carreau: The issue is there is a conflicting keyword named `wrap`, where this is used. I've pushed a commit that use the fully qualified name.

Carreau:
Strange, I've push but github don't show it:

```
$ git push
Enumerating objects: 9, done.
Counting objects: 100% (9/9), done.
Delta compression using up to 8 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 529 bytes | 529.00 KiB/s, done.
Total 5 (delta 4), reused 0 (delta 0)
remote: Resolving deltas: 100% (4/4), completed with 4 local objects.
To ssh://github.com/oscargus/ipython.git
   978f4da6a..e5dc064ee  enablelatexfontcolor -> enablelatexfontcolor
```

Carreau:
I'm going to close and reopen see if it sees the new commits.

Carreau:
Ok, it worked. CI need to re-run though.

ccordoba12 (commented):
Thanks @oscargus !

IPython/lib/latextools.py
oscargus: Thanks @Carreau ! Didn't really take a step back and figured out a better solution, but was just trying to get it to work. Clearly, using a fully qualified name is the way to go.

PR #11848, opened 2019-08-11
Provide hooks for arbitrary mimetypes handling.

Carreau:
This should allow at some point inline images, an math in the terminal.

Screen Shot 2019-11-17 at 10 48 45
 
Screen Shot 2019-11-17 at 10 49 45

### DIFF
--- a/docs/source/config/shell_mimerenderer.rst
+++ b/docs/source/config/shell_mimerenderer.rst
@@ -56,5 +56,5 @@ to png, before displaying it inline for iterm2 on macOS ::
         register_mimerenderer(ipython, 'text/latex', mathcat)
 
 This example only work for iterm2 on macOS and skip error handling for brevity.
-One could also invoke an external viewer with ``subporcess.run()`` and a
+One could also invoke an external viewer with ``subprocess.run()`` and a
 temporary file, which is left as an exercise.
--- a/docs/source/config/shell_mimerenderer.rst
+++ b/docs/source/config/shell_mimerenderer.rst
@@ -1,0 +1,60 @@
+
+.. _shell_mimerenderer:
+
+
+Mime Renderer Extensions
+========================
+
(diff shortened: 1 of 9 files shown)


Carreau:
@asmeurer I know you are a Terminal IPython user with many extensions. 
Thoughts ?

docs/source/config/shell_mimerenderer.rst
```diff
@@ -0,0 +1,60 @@
...
+        ipython.display_formatter.formatters[mime].enabled = True
+        ipython.mime_renderers[mime] = handler
+
+    def load_ipython_extension(ipython):
+        register_mimerenderer(ipython, 'image/png', imcat)
+        register_mimerenderer(ipython, 'image/jpeg', imcat)
+        register_mimerenderer(ipython, 'text/latex', mathcat)
+
+This example only work for iterm2 on macOS and skip error handling for brevity.
+One could also invoke an external viewer with ``subporcess.run()`` and a
```
asmeurer: This is spelled wrong.

asmeurer:
I don't use IPython, but if this supports iTerm2 I think it's great. My only comment would be to make the images be full retina resolution.

docs/source/config/shell_mimerenderer.rst
Carreau: Oh thanks.

Carreau:
> I don't use IPython, but if this supports iTerm2 I think it's great. My only comment would be to make the images be full retina resolution.

Sorry I thought you had some custom prompt extensions at some points. I might be confusing with something else. Apologies for the wrong ping and thanks for spotting the typo.

asmeurer:
I did have some custom stuff at one point but I gave up trying to finagle IPython at some point and just wrote my own thing https://github.com/asmeurer/mypython.

Carreau:
> https://github.com/asmeurer/mypython.

Ah thanks, I'll try to remember this then.

PR #11850, opened 2019-08-12
optionally enable case-insensitive tab completion in %cd, for macOS users

palxex:
Currently the tab completion in %cd relies on glob.glob(), which strangely case sensible on macOS, while case-insensitive was actually the default setting of macOS.
This PR trying to address it, while keeps original behavior as default, simple add 'c.TerminalInteractiveShell.completion_sensitive = False' to config file enables case-insensible completion in %cd command.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -123,6 +123,7 @@ import sys
 import unicodedata
 import string
 import warnings
+import fnmatch
 
 from contextlib import contextmanager
 from importlib import import_module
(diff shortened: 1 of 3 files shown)


palxex:
Made several changes, now the configure line changed to 'c.InteractiveShell.completion_sensitive = False', and the completion should works for all command, instead of just %cd. 
The initial commit only works for relative path, later commit also fixed it.

Carreau:
Thanks I'll have a look. 

Note that jedi recently introduce filename completion; so we may switch to jedi's implementation soon. I haven't touched completion in a while, so not quite sure how that would affect `cd` and other path  completion.

palxex:
Glad to hear that, will test whether affected once jedi way done; if jedy way works, then this patch is not needed, the central of this patch is a hacked globbing function:)

Carreau:
Apologies, it has been a few years, and I have difficulties following up with all the IPython PRs, this now conflict and is quite old; so I'm going to close to try to decrease the number of opened PRs.

PR #11852, opened 2019-08-13
Fix example usage of BaseFormatter.for_type_by_name.

danielballan:
I propose a small documentation fix for this excellent feature of
IPython.

The current example doesn't work because the type module and type name
are lumped into one string. They must be given as separate strings.

See docstring:

```
Signature: BaseFormatter.for_type_by_name(self, type_module, type_name, func=None)
```

For extra clarity, I have added a submodule to this made-up example
module so that the pattern is clear.


Carreau:
Seem straitforward. 

THanks !

PR #11855, opened 2019-08-14
Allow custom completions to define their own ordering.

tonyfast:
The completer by default orders all of the matches.  

Screen Shot 2019-08-14 at 5 24 18 PM

An author may desire to present an unordered set of completions.  This pull requests modifies the completer to return custom ordered completions, sorted results may be explicitly defined by a completer function.


Carreau:
Thanks Tony, 

Apologies if this took a lot of time to review. That looks good. 

<3

PR #11864, opened 2019-08-22
Silence pager stderr

timokau:
This is run as a test to determine if a pager is available. If there is
no pager, there is a fallback. Therefore, I think it is a bug that in
the no-pager case ipython will print

 sh: less not found

To stderr.

I encountered this when running the sagemath testsuite on a machine without less and got surprised by the error messages.

### DIFF
--- a/IPython/core/page.py
+++ b/IPython/core/page.py
@@ -15,9 +15,11 @@ rid of that dependency, we could move it there.
 
 
 import os
+import io
 import re
 import sys
 import tempfile
+import subprocess
 
 from io import UnsupportedOperation
 
@@ -208,9 +210,13 @@ def pager_page(strng, start=0, screen_lines=0, pager_cmd=None):
         else:
             try:
                 retval = None
-                # if I use popen4, things hang. No idea why.
-                #pager,shell_out = os.popen4(pager_cmd)
-                pager = os.popen(pager_cmd + ' 2>/dev/null', 'w')
+                # Emulate os.popen, but redirect stderr
+                proc = subprocess.Popen(pager_cmd,
+                                shell=True,
+                                stdin=subprocess.PIPE,
+                                stderr=subprocess.DEVNULL
+                                )
+                pager = os._wrap_close(io.TextIOWrapper(proc.stdin), proc)
                 try:
                     pager_encoding = pager.encoding or sys.stdout.encoding
                     pager.write(strng)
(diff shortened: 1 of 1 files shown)


IPython/core/page.py
```diff
@@ -210,7 +210,7 @@ def pager_page(strng, start=0, screen_lines=0, pager_cmd=None):
                 retval = None
                 # if I use popen4, things hang. No idea why.
                 #pager,shell_out = os.popen4(pager_cmd)
-                pager = os.popen(pager_cmd, 'w')
+                pager = os.popen(pager_cmd + ' 2>/dev/null', 'w')
```
Carreau: Hum, wondering if we should use `subprocess.Popen` . Potentially `subprocess.run` now that this is available and properly discard stderr. I'm worried about string munging... but I may just be lazy and merge that as is later.

timokau: I don't have enough experience with passing stdin to `subprocess.Popen` to be sure to cover all the edge-cases are covered in the same way they are covered now.

Carreau: > I don't have enough experience with passing stdin to `subprocess.Popen` to be sure to cover all the edge-cases are covered in the same way they are covered now.

Then don't worry I'll push a commit on this branch if I decide `subprocess.Popen` is better.

Maybe not this week though.

timokau: Friendly ping :)

Carreau: Yes sorry Aug/Sep/Oct were crazy with my team size at work decreasing by 33%... Slowly getting out of the tarpit.

Carreau:
I (finally) pushed some changes; would you mind having a look and let me know if that fixes your case ?

IPython/core/page.py
```diff
@@ -208,9 +210,13 @@ def pager_page(strng, start=0, screen_lines=0, pager_cmd=None):
...
-                # if I use popen4, things hang. No idea why.
-                #pager,shell_out = os.popen4(pager_cmd)
-                pager = os.popen(pager_cmd, 'w')
+                # Emulate os.popen, but redirect stderr
+                proc = subprocess.Popen(pager_cmd,
+                                shell=True,
+                                stdin=subprocess.PIPE,
+                                stderr=subprocess.DEVNULL
+                                )
+                pager = os._wrap_close(io.TextIOWrapper(proc.stdin), proc)
```
timokau: Should we really use a function that is not part of the public API? What is the intention here?

timokau:
I can't reproduce the issue reliably, but since stderr is suppressed it should fix my issue.

IPython/core/page.py
Carreau: In this case I believe this is ok, as this is exactly what `os.popen` call. There is not much reason for it to change.

PR #11876, opened 2019-09-04
switching update_instance() to gc.get_referrers() 

Naikless:
The resulting code is almost laughably simple and should also be way more efficient than the previous attempt.


daharn:
Nevermind, I overlooked that you already implemented this!

Carreau:
Thanks for working on this ! If that woks for you let's get that in.

esvhd:
Thanks for fixing this folks. When do you expect 7.9 to hit the road?

PR #11877, opened 2019-09-05
Prevent caching outputs in _, __, ___ when cache_size isn't positive

### LINKED ISSUE #3452: Memory leak even when cache_size = 0 and history_length = 0 or history_length = 1
i'm using git master ipython, python 2.7.5, arch linux 64-bit

using the following code as a starting point

``` python
from pandas import DataFrame, concat
from numpy.random import randint, randn
from string import ascii_letters as letters
ncols = 16
nrows = 1e7
letters = list(letters)
df = DataFrame(randn(nrows, ncols), columns=letters[:ncols])
col = DataFrame(randint(1, size=nrows), columns=[letters[ncols]])
big = concat([df, col], axis=1)
```

if you now repeatedly evaluate `big.values` (by hand not in a loop) then memory keeps on growing (open `htop` or `top` to watch it in action). same is true when i've set the output cache to zero and the history length to 0 or 1.

furthermore this _doesn't_ happen in vanilla python which has just `_` for history and if i e.g., evaluate `big.values` _n_ times then i need to execute _n_ other statements e.g,. `x = 1`  _n_ times to reclaim the memory.

see issue at pydata/pandas#3629 for a long discussion about this. am i missing some feature/quirk of the history or caching system?

mohammad7t:
Fixes https://github.com/ipython/ipython/issues/3452 as suggested by @takluyver on https://github.com/ipython/ipython/issues/3452#issuecomment-19781095

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -198,7 +198,7 @@ class DisplayHook(Configurable):
         """Update user_ns with various things like _, __, _1, etc."""
 
         # Avoid recursive reference when displaying _oh/Out
-        if result is not self.shell.user_ns['_oh']:
+        if self.cache_size and result is not self.shell.user_ns['_oh']:
             if len(self.shell.user_ns['_oh']) >= self.cache_size and self.do_full_cache:
                 self.cull_cache()
 


Carreau:
Seem reasonable to me. Thanks for your patience.

PR #11905, opened 2019-10-05
IPython directive: correctly clear cout

### LINKED ISSUE #11904: BUG: null bytes included in captured output of IP.run_cell (ipython directive)
We are having a problem in the pandas and geopandas docs that use the IPython directive with certain code blocks not being styled as such. See https://github.com/pandas-dev/pandas/issues/28796. 
And as an example, see https://geopandas.readthedocs.io/en/v0.6.0/missing_empty.html (some code blocks are greenish background with code highlighting, some code blocks are just verbatim, while all are using the ipython directive).

I have been trying to debug it, and I pinpointed it to the fact that the generated output by IPython includes "null bytes" (`\x00`) at the beginning of the line for those code blocks that give the problem.

To give a concrete example (from the linked geopandas docs above), if you have an ipython directive like

```
.. ipython:: python

    s.union(Polygon([(0, 0), (0, 1), (1, 1), (1, 0)]))
```

this gets translated by IPython to a plain `code-block` (which is then processed by sphinx and converted to html)

```
.. code-block:: ipython

   In [5]: s.union(Polygon([(0, 0), (0, 1), (1, 1), (1, 0)]))
   Out[5]: 
   0    POLYGON ((0.00000 1.00000, 1.00000 1.00000, 1....
   1                                                 None
   2    POLYGON ((0.00000 0.00000, 0.00000 1.00000, 1....
   dtype: geometry
```

For the code blocks that sphinx does not handle well, like the one above, there seem to be null characters included. So if we look at the actual strings instead of printing the joined output (showing the repr, so we see the null characters), we see:

```
['.. code-block:: ipython', '', '   In [5]: s.union(Polygon([(0, 0), (1, 1), 
(1, 0)]))', '   \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00Out[5]: ', '   0    POLYGON ((0.00000 1.00000, 1.00000 1.00
000, 1....', '   1                                                 None', 
'   2    POLYGON ((0.00000 0.00000, 0.00000 1.00000, 1....', '   dtype: 
geometry', '   ']
```

So for certain cells, there are a bunch of `\x00` included just before the `Out []`, and for others cells this is not the case.

I am not able to understand why it happens sometimes and sometimes not (or why it would happen at all). 
The output you above is from capturing the stdout that IPython generated when doing `IP.run_cell(..)`:

https://github.com/ipython/ipython/blob/4a2240fb440c497e9e195d9ff43d96973b7ae33f/IPython/sphinxext/ipython_directive.py#L396-L404

and the captured stdout is then read:

https://github.com/ipython/ipython/blob/4a2240fb440c497e9e195d9ff43d96973b7ae33f/IPython/sphinxext/ipython_directive.py#L504-L506

It is this `processed_output` that contains the `\x00` characters. Adding a `processed_output = processed_output.strip('\x00')` "solves" the issue (sphinx renders the blocks nice again, so it is definitely due to this).

I am not sure why sphinx has a problem with those `\x00` characters, but it also seems to me that IPython should not be generating those in the output.

jorisvandenbossche:
Closes https://github.com/ipython/ipython/issues/11904

The StringIO `cout` object was not correctly cleared (while there was actually a helper method already defined to do that), which caused those null bytes to be introduced, see explanation here https://stackoverflow.com/questions/39109068/why-does-truncating-a-bytesio-mess-it-up

(it seems there are currently no real tests for the directive, so also didn't add one (not sure directly how to do that))

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -585,7 +585,7 @@ class EmbeddedSphinxShell(object):
                 if self.warning_is_error:
                     raise RuntimeError('Non Expected warning in `{}` line {}'.format(filename, lineno))
 
-        self.cout.truncate(0)
+        self.clear_cout()
         return (ret, input_lines, processed_output,
                 is_doctest, decorator, image_file, image_directive)
 


jorisvandenbossche:
The travis failure was one build that had a connection problem during pip install

jorisvandenbossche:
Any review / merge possible for this PR? 
I can confirm this fixes the bug for the pandas docs

Carreau:
Apologies, just starting to work on backlog. 
Looks reasonable. SHould be release in 7.10 at end of month.

jorisvandenbossche:
Thanks for merging!

PR #11907, opened 2019-10-10
Easier empty, updatable display

vidartf:
Given the following code:
```python
[1]: from IPython.display import display
[2]: handle = display(display_id=True)
[3]: handle.update('world')
```

Currently, `[2]` produces no output (no `display_data` message sent), but still gives a handle that silently fails to update.

This PR changes `[2]` to be equivalent to:
```python
handle = display({}, raw=True, display_id=True)
```
Note that this only happens when the output list is empty *and* a `display_id` is given (either string or `True`).

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -296,6 +296,13 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
     if transient:
         kwargs['transient'] = transient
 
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, that looks reasonable to me.

Carreau:
Apologies for the delay in reviewing.

Issue #11909, opened 2019-10-11
Restore terminal title after exiting IPython


pasenor:
IPython sets the terminal title to `IPython: {cwd}`, which is cool, but after exiting the shell, on xterm the original title is not restored. It can be fixed by [saving and then restoring the original title](https://stackoverflow.com/a/20048874/2166825):

    sys.stdout.write('\033[22;0t')  # Save title on stack
    sys.stdout.write('\033[23;0t')  # Restore title from stack

MrMino:
Fixed within #11910.

Issue #10856, opened 2017-10-10
Terminal title is not reset after exiting IPython


cbrnr:
By default, IPython changes the terminal title to `IPython: {cwd}`. However, this change is not reset to the original terminal title after exiting IPython (I'm on macOS High Sierra).

Here's the output of `IPython.sys_info()`:

```
{'commit_hash': 'd86648c5d',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/clemens/anaconda/lib/python3.6/site-packages/IPython',
 'ipython_version': '6.1.0',
 'os_name': 'posix',
 'platform': 'Darwin-17.0.0-x86_64-i386-64bit',
 'sys_executable': '/Users/clemens/anaconda/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '3.6.3 |Anaconda, Inc.| (default, Oct  6 2017, 12:04:38) \n'
                '[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]'}
```

takluyver:
I don't believe we have a good way to reset the previous terminal title, so it's up to your shell to set it again after IPython exits.

I found some info [here](https://mail.python.org/pipermail/python-list/2013-October/657961.html) about retrieving or storing terminal titles, but it sounds like most terminal emulators don't support it by default. I tried the code suggested on Gnome terminal, but no joy.

cbrnr:
Hm, so you're saying there's no way to retrieve the current terminal title? I thought that because IPython sets the title it would also be easy to retrieve the current title (sorry, I didn't have time to poke around the source code to find the relevant lines). I know that you can disable this behavior in the profile, but given that the behavior is somewhat suboptimal would it make sense to change the default value of `c.TerminalInteractiveShell.term_title` to `False`?

takluyver:
> Hm, so you're saying there's no way to retrieve the current terminal title?

My brief investigation suggests that there is no way which works reliably across different terminal emulators, yes.

> given that the behavior is somewhat suboptimal would it make sense to change the default value of c.TerminalInteractiveShell.term_title to False?

Well, suboptimal is a matter of opinion. I'm fine with it. ;-)

cbrnr:
> Well, suboptimal is a matter of opinion. I'm fine with it. ;-)

[...]

PR #11910, opened 2019-10-11
restore terminal title on exit (xterm)

gfrlv:
issue #11909 

On xterm, there is a terminal "stack" that allows to save and then restore the current terminal title. It allows to reset title to the original value on exit in case it was changed.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -8,7 +8,7 @@ from warnings import warn
 from IPython.core.interactiveshell import InteractiveShell, InteractiveShellABC
 from IPython.utils import io
 from IPython.utils.py3compat import input
-from IPython.utils.terminal import toggle_set_term_title, set_term_title
+from IPython.utils.terminal import toggle_set_term_title, set_term_title, restore_term_title
 from IPython.utils.process import abbrev_cwd
 from traitlets import (
     Bool, Unicode, Dict, Integer, observe, Instance, Type, default, Enum, Union,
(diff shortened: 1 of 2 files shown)


pasenor:
Checked the master branch on windows 7, the title is restored on exit, so it seems that the problem shows on xterm only.

Carreau:
I don't have any objections; 

I remember some discussion about that some time ago; and I believe this may garbled the output on some other kind of terminal. 

Note that I believe the logic might be perfect in all the case as set_term_title might be called a second time if something change the internal title string. In which case another title is pushed; but only one will be pop'ed at exit. 

I believe in the long run we might want to have `with title()` context manager to make sure we do things properly.

pasenor:
> I remember some discussion about that some time ago; and I believe this may garbled the output on some other kind of terminal.

Do you have some pointers to that discussion? If restoring the title messes things up, chances are setting the title does too, perhaps we can do something special for such systems.

> I believe in the long run we might want to have with title()

Seems sensible, I'll look how to best do it. Is this a theoretical concern, or are there cases where we set the title multiple times?

Carreau:
> Do you have any pointers to that discussion? 
No, just in my memory, I found https://github.com/ipython/ipython/issues/10856 as well, but I coudl swear there is more.

> Seems sensible, I'll look how to best do it.

That will probably be a lot of refactor; so loooong term. 

> Or are there cases where we set the title multiple times?

With current code base I believe the title can be set multiple times. But with the proper context manager we should be able to count how many time it was changed from within the context manager and pop the appropriate number of time. 

It may even be a feature we want in [prompt_toolkit](https://python-prompt-toolkit.readthedocs.io/en/master/)

Issue #10055, opened 2016-11-15
gui wx incompatibilities with new wxpython 'phoenix'


karlacio:
Hi,

running a program that uses wxpython gui with iphyton:
`ipython --gui wx --i my_program.py`

gives an error like:

> File ".../site-packages/IPython/terminal/pt_inputhooks/wx.py", line 76, in inputhook_wx2
>     assert wx.Thread_IsMain()
> AttributeError: 'module' object has no attribute 'Thread_IsMain'
> 

On wxpython 'phoenix' seems that the method `Thread_IsMain` as been replaced by `IsMainThread`.

Replacing it on wx.py it seems to work.

I'm running

> Python 2.7.10 
> IPython 5.1.0
> wx.__version__ 3.0.3.dev2553+47f1fd2
> on OSX 10.11.6

takluyver:
Oh, wx is going again? I'd recommend making a new event loop integration for it. It looks like it doesn't have a clearly distinct version number, so we can't call it e.g. `wx3`. Maybe `wxphoenix`?

jaredl7:
I'm using IPython 6.2.1 and I had the same issue as @karlacio.

Using the example from here: https://github.com/ipython/ipython/blob/master/examples/IPython%20Kernel/gui/gui-wx.py

and, applying the fix that @karlacio recommended (this is line 118 in `inputhook_wx3` in IPython 6.2.1) - i.e.:

**`assert wx.Thread_IsMain()`** -> **`assert wx.IsMainThread()`**

one obtains a new error: 

> AttributeError: module 'wx' has no attribute 'Timer'

which appears to be solved by updating 

**`app.ProcessIdle()`** to **`evtloop.ProcessIdle()`** 

on **`<Python version>/site-packages/IPython/lib/inputhookwx.py`** and in **`<Python version>/site-packages/IPython/terminal/pt_inputhooks/wx.py`**

Having referenced this updated sample from wxPython: https://github.com/wxWidgets/wxPython/blob/4b20d984aaaf4a5d7dfcdc1490c0b42489d452f7/samples/mainloop/mainloop.py

Applying these fixes solves the issue, and one is able to run the example cited without issue.

takluyver:
Do you want to make a new inputhook for wxpython 4 with these changes? I.e. add `IPython.terminal.pt_inputhooks.wx4`. We don't use the `IPython.lib.inputhook*` ones any more.

[...]

PR #11911, opened 2019-10-11
Input hook for wxpython phoenix

pauldmccarthy:
Addresses #10055.

Currently working with linux/gtk3. I'll get it working on macOS next week, at which point I may clean up/refactor the code to reduce duplication.

### DIFF
--- a/IPython/terminal/pt_inputhooks/wx.py
+++ b/IPython/terminal/pt_inputhooks/wx.py
@@ -175,13 +175,6 @@ def inputhook_wxphoenix(context):
     # Wx uses milliseconds
     poll_interval = 100
 
-    # We have to create a dummy wx.Frame, otherwise wx.App.MainLoop will know
-    # that it has nothing to do, and will return immediately.
-    frame = getattr(inputhook_wxphoenix, '_frame', None)
-    if frame is None:
-        inputhook_wxphoenix._frame = frame = wx.Frame(None)
-        frame.Show(False)
-
     # Use a wx.Timer to periodically check whether input is ready - as soon as
     # it is, we exit the main loop
     def poll(ev):
@@ -198,6 +191,9 @@ def inputhook_wxphoenix(context):
     if not callable(signal.getsignal(signal.SIGINT)):
         signal.signal(signal.SIGINT, signal.default_int_handler)
 
+    # The SetExitOnFrameDelete call allows us to run the wx mainloop without
+    # having a frame open.
+    app.SetExitOnFrameDelete(False)
     app.MainLoop()
 
 
--- a/IPython/terminal/pt_inputhooks/wx.py
+++ b/IPython/terminal/pt_inputhooks/wx.py
@@ -154,7 +154,9 @@ def inputhook_wx3(context):
 def inputhook_wxphoenix(context):
     """Run the wx event loop until the user provides more input.
 
-    This function uses the same approach to that used in
+    This input hook is suitable for use with wxPython >= 4 (a.k.a. Phoenix).
+
+    It uses the same approach to that used in
     ipykernel.eventloops.loop_wx. The wx.MainLoop is executed, and a wx.Timer
     is used to periodically poll the context for input. As soon as input is
     ready, the wx.MainLoop is stopped.
--- a/IPython/terminal/pt_inputhooks/wx.py
+++ b/IPython/terminal/pt_inputhooks/wx.py
@@ -155,7 +155,9 @@ def inputhook_wxphoenix(context):
     """Run the wx event loop until the user provides more input.
 
     This function uses the same approach to that used in
-    ipykernel.eventloops.loop_wx.
+    ipykernel.eventloops.loop_wx. The wx.MainLoop is executed, and a wx.Timer
+    is used to periodically poll the context for input. As soon as input is
+    ready, the wx.MainLoop is stopped.
     """
 
     app = wx.GetApp()
--- a/IPython/terminal/pt_inputhooks/wx.py
+++ b/IPython/terminal/pt_inputhooks/wx.py
@@ -9,8 +9,12 @@ import wx
 
 
 def ignore_keyboardinterrupts(func):
-    """Decorator which causes KeyboardInterrupt exceptions to be
-    ignored during execution of the decorated function.
+    """Decorator which causes KeyboardInterrupt exceptions to be ignored during
+    execution of the decorated function.
+
+    This is used by the inputhook functions to handle the event where the user
+    presses CTRL+C while IPython is idle, and the inputhook loop is running. In
+    this case, we want to ignore interrupts.
     """
     def wrapper(*args, **kwargs):
         try:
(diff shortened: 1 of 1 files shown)


pauldmccarthy:
After experimenting a bit more with both IPython and Jupyter notebook, I discovered that the approach used in [ipykernel.eventloops.wx_loop](https://github.com/ipython/ipykernel/blob/master/ipykernel/eventloops.py#L152) works perfectly well on both Linux and macOS. I don't have access to a Windows machine unfortunately.

I suspect that that this approach will probably work with older versions of wxPython too, although I have not tested this suspicion.

So now, when `wxpython >= 4` is detected, the new `inputhook_wxphoenix` is set to the default input hook.

I've also cleaned up the existing code just a little bit, factoring out the `try` / `except KeyboardInterrupt` logic to a decorator.

Comments?

pauldmccarthy:
The travis failure looks unrelated - it looks like the version of `outcome` is expecting a more recent version of `attrs` - the `eq` argument to `attr.ib` was added in `attrs` 19.2.0, but the CI environment has 18.2.0 installed.

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -90,58 +103,110 @@ def inputhook_wx3(context):
...
+            self.Bind(wx.EVT_TIMER, self.on_timer)
+            self.func = func
+
+        def on_timer(self, event):
+            self.func()
+
+    frame = TimerFrame(wake)
+    frame.Show(False)
+
+    # The import of wx on Linux sets the handler for signal.SIGINT
```
tacaswell: This should maybe go outside the input hook?  Not sure we want to do this every time.

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -90,58 +103,110 @@ def inputhook_wx3(context):
...
+    # to 0.  This is a bug in wx or gtk.  We fix by just setting it
+    # back to the Python default.
+    if not callable(signal.getsignal(signal.SIGINT)):
+        signal.signal(signal.SIGINT, signal.default_int_handler)
+
+    app.MainLoop()
```
tacaswell: maybe check if the context is ready before launching the app?

IPython/terminal/pt_inputhooks/wx.py
tacaswell: although I see it is done in the input hook above so maybe we do want to do it every time.

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -90,58 +103,110 @@ def inputhook_wx3(context):
...
+            app.ExitMainLoop()
+
+    # We have to put the wx.Timer in a wx.Frame for it to fire properly.
+    # We make the Frame hidden when we create it in the main app below.
+    class TimerFrame(wx.Frame):
+        def __init__(self, func):
+            wx.Frame.__init__(self, None, -1)
+            self.timer = wx.Timer(self)
+            self.timer.Start(poll_interval)
+            self.Bind(wx.EVT_TIMER, self.on_timer)
```
tacaswell: This may be a daft question, but how does `self.Bind` know to watch the timer we defined above not some other timer?

tacaswell:
The failure is that trio requires a very new version of `attrs` and pip does not correctly catch the pinning.

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -90,58 +103,110 @@ def inputhook_wx3(context):
...
+    # input is ready, the wx main loop is stopped.
+    def wake():
+        if context.input_is_ready():
+            app.ExitMainLoop()
+
+    # We have to put the wx.Timer in a wx.Frame for it to fire properly.
+    # We make the Frame hidden when we create it in the main app below.
+    class TimerFrame(wx.Frame):
+        def __init__(self, func):
+            wx.Frame.__init__(self, None, -1)
```
tacaswell: can use `super()` here ipython no longer supports py2

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -90,58 +103,110 @@ def inputhook_wx3(context):
...
+            self.timer = wx.Timer(self)
+            self.timer.Start(poll_interval)
+            self.Bind(wx.EVT_TIMER, self.on_timer)
+            self.func = func
+
+        def on_timer(self, event):
+            self.func()
+
+    frame = TimerFrame(wake)
```
tacaswell: This creates a new Wx object everytime the input hook is called, where as the ipykernal code looks like it creates the frame once and then just keeps reusing it.

tacaswell:
I think the principle is sound, but I'm a bit concerned about the amount of work that is being done on every pass through the input hook.

pauldmccarthy:
Hey @tacaswell, all fair points - I did little more than copy-paste the code from [ipykernel.eventloops](https://github.com/ipython/ipykernel/blob/master/ipykernel/eventloops.py#L152). I'll have another play and see if I can make things a bit cleaner.

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: I don't think we actually need to associated the timer with the frame - a regular wx.Frame works just fine.

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: I think it's because the Frame was passed as the owner to the Timer (i.e. `wx.Timer(self)`). The new implementation is a bit clearer.

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: No longer defining a custom sub-class

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: You're right - updated the code to create and persist a single frame.

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: Done at beginning of function - I've found out that it is actually essential to perform an initial check, otherwise the prompt becomes unresponsive. The original implementation was implicitly doing this - I'm not sure how it worked, but the timer was firing pretty much immediately, which caused it to behave nicely and not hog the CPU.

pauldmccarthy:
I've refactored the code to make it quite a bit cleaner. 
 - No longer defining a custom sub-class - just a regular `wx.Frame` and `wx.Timer` work fine.
 - A single `wx.Frame` is created on the first call, and re-used on subsequent calls.
 - An initial call to `context.input_is_ready` is performed

Tested on both Ubuntu 18.04 and macOS 10.14, and works well for me.

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -90,58 +107,115 @@ def inputhook_wx3(context):
...
+
+    # Wx uses milliseconds
+    poll_interval = 100
+
+    # We have to create a dummy wx.Frame, otherwise wx.App.MainLoop will know
+    # that it has nothing to do, and will return immediately.
+    frame = getattr(inputhook_wxphoenix, '_frame', None)
+    if frame is None:
+        inputhook_wxphoenix._frame = frame = wx.Frame(None)
+        frame.Show(False)
```
RobinD42: You may be able to avoid the need for a frame if you use `app.SetExitOnFrameDelete(False)`

IPython/terminal/pt_inputhooks/wx.py
```diff
@@ -8,30 +8,45 @@
...
+
+        # Make a temporary event loop and process system events until
+        # there are no more waiting, then allow idle events (which
+        # will also deal with pending or posted wx events.)
+        evtloop = wx.EventLoop()
+        ea = wx.EventLoopActivator(evtloop)
+        while evtloop.Pending():
+            evtloop.Dispatch()
+        app.ProcessIdle()
+        del ea
```
RobinD42: `wx.EventLoopActivator` can be used as a context manager

IPython/terminal/pt_inputhooks/wx.py
tacaswell: :+1:

IPython/terminal/pt_inputhooks/wx.py
tacaswell: I think Matplotlib sets that on the app so that if you do call `app.MainLoop()` to block the process and then close all of the plots you get the prompt back.

That said, if there are no open windows at all, and we are waiting for user input is it OK that this short-circuits and immediately returns?

tacaswell:
Trust @RobinD42 over anything I say ;)

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: @RobinD42 yep, that seems to work fine, thanks for the tip!

@tacaswell It doesn't seem to be a problem; I'll do some more testing over the next few days (away from work, so have limited time right now).

IPython/terminal/pt_inputhooks/wx.py
pauldmccarthy: I didn't touch the existing hooks; I suspect that the strategy used in the new `inputhook_wxphoenix` will probably work for wxpython 3.x, but haven't set up an environment to check that.

pauldmccarthy:
Tested the current implementation on both Linux and macOS, and it seems to be working fine:

```python
# 1. Initialise/import:
%gui wx
import wx

# 2. Create an app and a frame
app = wx.App()
frame = wx.Frame(None)
frame.Show()

# 3. Manually close the frame

# 4. Clear the app/frame refs
app = None
frame = None

# 5. Repeat from step 2
```

I'd be happy to take a look at seeing if the new hook works with wxpython 3.x, but this would have to wait until next week when I'm back home.

Carreau:
Wow, thanks, it seem that now things get automatically reviewed when I'm busy and don't look at the PR queue for several weeks :-P. 

all look good to me, so I'll merge; it will be released at end of Month; and we can iterate if needed. 

THanks !

PR #11923, opened 2019-10-22
Update trio and curio dependencies eagerly.

Carreau:
Otherwise we get an old version of attr in travis CI; which fails at runtime.


Carreau:
self-merging to fix master.

PR #11930, opened 2019-10-27
Fix restoring more than 2 variables and support storing multiple variables %storemagic

gokceneraslan:
Currently `%store -r a b c` doesn't restore `c` because argument splitting is not performed correctly (see `args = argsl.split(None,1)` below). This is now fixed. `%store a b c` was not supported before, now it is working well. I also extended the tests to cover all scenarios.


Carreau:
Thanks, that looks reasonable to me !

Issue #11496, opened 2018-11-15
Prompts.in_prompt_tokens gets called several times per second


reuben:
While customizing my prompt I noticed that prompt_toolkit was calling `Prompt.in_prompt_tokens` several times per second which was slowing down my shell quite a bit. Here's the relevant code:
https://github.com/ipython/ipython/blob/844ad8b98fec387399062ef373f202d891ed708d/IPython/terminal/interactiveshell.py#L368-L393

Which gets passed into prompt_toolkit here: https://github.com/ipython/ipython/blob/844ad8b98fec387399062ef373f202d891ed708d/IPython/terminal/interactiveshell.py#L281-L292

prompt_toolkit supports both a callable and a static message, so I tried applying the following change:

```diff
diff --git a/IPython/terminal/interactiveshell.py b/IPython/terminal/interactiveshell.py
index e9f164943..f60dd3887 100644
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -365,14 +365,11 @@ def _extra_prompt_options(self):
         """
         Return the current layout option for the current Terminal InteractiveShell
         """
-        def get_message():
-            return PygmentsTokens(self.prompts.in_prompt_tokens())
-
         return {
                 'complete_in_thread': False,
                 'lexer':IPythonPTLexer(),
                 'reserve_space_for_menu':self.space_for_menu,
-                'message': get_message,
+                'message': PygmentsTokens(self.prompts.in_prompt_tokens()),
                 'prompt_continuation': (
                     lambda width, lineno, is_soft_wrap:
                         PygmentsTokens(self.prompts.continuation_prompt_tokens(width))),
```

As far as I could tell, it did not affect any prompt functionality, and all the tests pass with this change. But there's probably some prompt functionality I don't know about that requires the refreshing.

prompt_toolkit also has a `refresh_interval` option but passing a higher refresh interval doesn't seem to reduce the frequency of the prompt being invalidated.

Does the change above seem reasonable?

Carreau:
Yes, I believe they are reasonable. Technically someone could define a in_prompt_tokens` that change through time, but it's rare. So I think we can do your suggested change.

Carreau:
Do you like to send a PR?

a-ungurianu:
This has been resolved by https://github.com/ipython/ipython/pull/11933 so I think this should be closed

ivanov:
[...]

PR #11933, opened 2019-11-01
static message instead of function call to prevent shell slowdown

gerrit8143:
#11496

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -377,12 +377,24 @@ class TerminalInteractiveShell(InteractiveShell):
         """
         Return the current layout option for the current Terminal InteractiveShell
         """
+        def get_message():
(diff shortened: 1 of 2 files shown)


Carreau:
Hum, unlike what I said in #11496 his actually breaks the prompt when using vi mode for example. 
Vi mode (`$ ipython --TerminalInteractiveShell.editing_mode=vi`) shows in the prompt wether we are in insert or normal mode. 

So we'll have to a be a bit smarter as to wether we compute this once, or regularly.

Carreau:
I've push some changes on your brnach that shoudl fix the high number of call with the default mode, but keep the changing prompt when using vi keybinding. Let me know what you think.

gerrit8143:
looks good to me, so let's do it that way.

Carreau:
> looks good to me, so let's do it that way.

Thanks, I'll update the whats new note when I have time and then merge. 
Should be released by end of month.

PR #11935, opened 2019-11-05
Add more windows ci

bollwyvl:
This adds python 3.7 and 3.8 excursions on appveyor, using `choco` to get 3.8 for the time being.

We were seeing some stuff with `$HOME` handling, but otherwise was apparently mostly working.

Happy to tone it down to just x64, i guess, but we need to get some better understanding of some of the issues.

ref https://github.com/conda-forge/ipython-feedstock/pull/82

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,9 +36,9 @@ install:
       Set-PSDebug -Trace 1
       if ($Env:PYTHON_VERSION.StartsWith("3.8")) {
         if ($Env:PYTHON_ARCH -eq "64") {
-          iex "choco install python3 --version=3.8 --no-progress --params /InstallDir:$Env:PYTHON_HOME"
+          iex "choco install python3 --version=3.8 --no-progress --params /InstallDir:$Env:PYTHON"
         } else {
-          iex "choco install python3 --version=3.8 --no-progress --params /InstallDir:$Env:PYTHON_HOME --x86"
+          iex "choco install python3 --version=3.8 --no-progress --params /InstallDir:$Env:PYTHON --x86"
         }
       }
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,15 +32,6 @@ init:
   - "ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH%"
 
 install:
-  - ps: |
-      Set-PSDebug -Trace 1
-      if ($Env:PYTHON_VERSION.StartsWith("3.8")) {
-        if ($Env:PYTHON_ARCH -eq "64") {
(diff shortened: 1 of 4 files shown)


bollwyvl:
Also happy to start an azure-based approach if we're concerned about appveyor performance/queue times.

bollwyvl:
From https://docs.python.org/3.8/library/os.path.html#os.path.expanduser

> Changed in version 3.8: No longer uses HOME on Windows.

So is it better to:
- maintain backwards compatibility (somehow) and leave the test
- skip the test, and change the docs

Carreau:
I'm tempted to follow what Python does skip the test and update the docs.

bollwyvl:
Thanks @Carreau. Pushed with that. Here's hoping!

bollwyvl:
Ooof, gotta be importing...

bollwyvl:
OSX fail probably due to https://github.com/minrk/homebrew-python-frameworks/issues/1

bollwyvl:
Cool, windows win!

Carreau:
Didn't knew @minrk  was (also) maintaining this.
Let's see if we can ping him on this; or we disable 3.7 on os x.

minrk:
Fixed the homebrew tap, restarted the test. I forgot that repo was used anywhere but my laptop!

appveyor.yml
```diff
@@ -12,10 +12,35 @@ environment:
...
+     PYTHON_VERSION: "3.8.x"
+     PYTHON_ARCH: "64"
+
 init:
   - "ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH%"
 
 install:
+  - ps: |
+      Set-PSDebug -Trace 1
+      if ($Env:PYTHON_VERSION.StartsWith("3.8")) {
```
minrk: Appveyor has 3.8 as of [a few days ago](https://github.com/appveyor/ci/issues/3142#issuecomment-552049896), I think, as long as we use an image >= vs 2015

minrk (approved):

bollwyvl:
Huzzah!

bollwyvl:
Great, back to happy. @minrk @Carreau thanks for your help!

appveyor.yml
bollwyvl: removed, worked the first time!

minrk:
woo! thanks @bollwyvl!

bollwyvl:
I did note that the pyzmq install takes a very long time, accounting for the bulk of the difference in time between the excursions... is there more wheelsmithing needed upstream?

bollwyvl:
I see there are py38 wheels, hm... https://pypi.org/project/pyzmq/#files

minrk:
Working on the py38 pyzmq wheels. I've uploaded several, but not Windows yet since that's automated on CI and I have to make a release before new wheels get uploaded.

bollwyvl:
Hey, but it's working! No worries!

PR #11937, opened 2019-11-11
Adopt NEP 29, Drop Python 3.5 and Numpy <1.14

Carreau:
This lays out the foundation of adopting NEP29 in IPython,
And drop support for older versions of Python in the metadata, setup.py
and __init__.

This does not yet remove any codepath from within IPython.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ addons:
 
 python:
     - 3.6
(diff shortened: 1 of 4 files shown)


willingc (approved):
LGTM. Thanks @Carreau.

PR #11938, opened 2019-11-12
Change absolute imports to relative imports 

### LINKED ISSUE #11932: Should we remove self-references?
I noticed that when I tried to run iPython's interactive shell as a local dependency

``` from . import ipython ```

A lot of the files in the program have self-references that expect ipython to be a global package. For example in interactiveShell.py...

```from IPython.core.events import EventManager, available_events```

But if we replace this with, 

```from ..core.events import EventManager, available_events```

or

```from .events import EventManager, available_events```

the self-reference is eliminated, and we can import the library and it works without iPython being defined in the global namespace.

srinivasreddy:
Change absolute imports to relative imports to facilitate processes embedding kernel or debugger

Fixes #11932

### DIFF
--- a/IPython/config.py
+++ b/IPython/config.py
@@ -7,7 +7,7 @@ Shim to maintain backwards compatibility with old IPython.config imports.
 import sys
 from warnings import warn
 
-from IPython.utils.shimmodule import ShimModule, ShimWarning
+from .utils.shimmodule import ShimModule, ShimWarning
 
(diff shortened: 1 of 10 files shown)


Carreau:
Thanks !

srinivasreddy:
Does it make sense to do it for test files as well? I do not think so. what do you think?

Carreau:
Test files are less important I believe; But let's not spend too much time on it if it's not necessary.

Issue #11955, opened 2019-11-21
Simplification of logic since Python 3.5 removal.


Carreau:
Since #11953 was merged a couple of extra simplification can be done:

See in particular  this comments. 

 - https://github.com/ipython/ipython/pull/11953/files#r348243309

The value `DICT_IS_ORDERED` in `IPython/lib/pretty.py` is always `True`; any code that reply on it can be simplified; and the value should be documented for future removal.

This is a good issue for a first time contributor.

whiskeyjack-dev:
I'll take a crack at this.

jamescooke:
It looks like this is done and should be closed.

mijiturka:
Going by CONTRIBUTING.md it's now possible for everyone to close an issue:
@meeseeksdev close

PR #11956, opened 2019-11-21
Remove unchanging boolean DICT_IS_ORDERED and dependent unreachable code

petevdp:
This PR fixes issue #11955 by removing `DICT_IS_ORDERED` and removing the one if-statement that it relies on.

Since this is such a small change I'm not sure if this needs much explanation, but if there's any documentation that should be included I can do that as well.


Carreau:
Well, that was fast ! Thanks !

I think this is strait forward enough that it does not need much explanation.

petevdp:
You're welcome!

PR #11961, opened 2019-11-24
Update the what's new.

Carreau:


Carreau:
@augustogoulart I did some rough updates of the what's new. you expressed some interest in helping.

I might do the release on Wednesday as Thus/Fri are holidays  here and I might not have my computer w/ me. Or maybe I'll do the release later.

augustogoulart:
```
IPython 7.10 is the first double digit minor release in the  last decade, and
first since the release of IPython 1.0, previous double digit minor release was
in August 2009.
```

@Carreau that’s fantastic!

It seems that the release notes are in very good shape already. Are there any low hanging fruits I could take care of to help you release this tomorrow?

Carreau:
> It seems that the release notes are in very good shape already. Are there any low hanging fruits I could take care of to help you release this tomorrow?

Sorry for the late reply, fighting a cold.  I don't think there is much, I'll do the release, and then we can start to work on the release for next month :-)

PR #11965, opened 2019-11-25
[WIP] Speedup startup by beeing lazy.

Carreau:
A significant chunk of the startup time is spent loading pygments
lexers.

This make it lazy which slightly improve the hot and cold startup time.

time `ipython -c exit()` cold goes from ~0.7 to ~0.65 seconds on my work
laptop I expect much more on old or slower machines.

### DIFF
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -136,18 +136,41 @@ class IPythonPTLexer(Lexer):
     """
     def __init__(self):
         l = pygments_lexers
-        self.python_lexer = PygmentsLexer(l.Python3Lexer)
-        self.shell_lexer = PygmentsLexer(l.BashLexer)
+        self._python_lexer = None
+        self._shell_lexer = None
 
-        self.magic_lexers = {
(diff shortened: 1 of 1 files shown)


Carreau:
I realized that on my machine about 1/2 the startup time comes from the setuptools's installed ipython's script that loads pkg_resources,  and  then loads entry points. 

@takluyver you are our installer expert; any idea ?

takluyver:
Since a couple of years ago, pygments should lazy-load `pkg_resources` only if you try to list all lexers/formatters/styles/filters:
https://bitbucket.org/birkenfeld/pygments-main/pull-requests/690/delay-loading-pkg_resources-until-it-is/diff

If we can avoid triggering that kind of listing, we shouldn't need an extra lazy-loading layer on top of it.

As for the script: if you install with pip (and not the `-e` flag), you should get a script that doesn't use `pkg_resources`. pip creates the simpler scripts with distlib, rather than setuptools. For this reason, I tend to create my own developer installs by doing `pip install .`, then manually replacing the installed folder in `site-packages` with a symlink. Or I use Flit, of course. ;-)

PR #11973, opened 2019-11-27
Run the prompt in a separate asyncio loop.

jonathanslenders:
This requires a new prompt_toolkit 3.0.1, it does fix a little bug.

This prevents crashes from situations where the user starts scheduling new
coroutines that don't terminate. Like

    import asyncio
    async def func():
        while True:
            pass
    asyncio.ensure_future(func())

This would freeze the event loop and can't even be cancelled by pressing
control-c, because prompt_toolkit will never gain control from the event loop
to process key bindings.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -1,5 +1,6 @@
 """IPython terminal interface using prompt_toolkit"""
 
+import asyncio
 import os
 import sys
 import warnings
@@ -309,6 +310,7 @@ class TerminalInteractiveShell(InteractiveShell):
 
         editing_mode = getattr(EditingMode, self.editing_mode.upper())
 
+        self.pt_loop = asyncio.new_event_loop()
         self.pt_app = PromptSession(
                             editing_mode=editing_mode,
                             key_bindings=key_bindings,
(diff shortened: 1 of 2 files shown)


Carreau:
This + prompt toolkit master fix #11974

jonathanslenders:
prompt_toolkit 3.0.1 has been released. Feel free to merge it.

jonathanslenders:
By the way, sharing the loop between UI and ``eval()`` is something that we could support optionally if we want. This makes it possible to have coroutines running in the background and printing information above the prompt. - While waiting for input. (Similar to what we have for threads.)

It should definitely not be the default though. There is too much risk of people scheduling coroutines that don't yield control back to the UI, and no there is really no good way to interrupt "background" coroutines. (Coroutines are supposed to be cancelled only at ``await`` points, and using signals to cancel them anywhere can break prompt_toolkit internals in many ways.)

Carreau:
Thanks. I'll be away from my computer for a few days. I'll merge once I'm
back and do a point release.

mrzzheng:
Hi, looks like the fix doesn't work for me: Python 3.7.5 + IPython 7.10.0 + PTK 3.0.1. Importing matplotlib.pyplot still causes inputs echo to disappear from prompt. Is it because my Python is not 3.8?

jonathanslenders:
@mrzzheng: What operating system are you using?

mrzzheng:
> @mrzzheng: What operating system are you using?

macOS 10.15

jonathanslenders:
@mrzzheng : I can reproduce the issue on my Ubuntu machine, by simply doing ``%gui tk``. I'm looking into it.

jonathanslenders:
@mrzzheng : I think I fixed the issue. It was on prompt_toolkit's side.
It's fixed in this commit, which has been pushed to master: https://github.com/prompt-toolkit/python-prompt-toolkit/commit/7f0541b4d2cb5c5ba28db54bbf221a9388715569

Would you mind trying again prompt_toolkit's master and this branch together to see whether it's fixed now?

mrzzheng:
@jonathanslenders confirmed that the issue is fixed combining with PTK3's change, thanks a lot!

jonathanslenders:
prompt_toolkit 3.0.2 has been released with this fix. I think if we can merge this and push a new IPython release, it should be fine.

Carreau:
Thanks @jonathanslenders for the fast response time and @mrzzheng for the help in finding these issues.

PR #11981, opened 2019-11-30
print the version on crash

Carreau:
Also so minimal automation of release.


Carreau:
(mistake a forced push on this branch, that this was the same as #11982)

PR #11997, opened 2019-12-01
Fix #11984 Removed unused py3compat import

### LINKED ISSUE #11984: Remove py3compact in test_autocall.py ?
Here is an easy issue if you are a first time contributor:

The file `IPython/core/tests/test_autocall.py` have the following line:
```
```

Though py3compat seem to be not unused. 

Is that correct ?
If so the import can be removed.

chemsseddine:


Carreau:
Thanks !

PR #12001, opened 2019-12-02
Don't nest asyncio input hooks.

jonathanslenders:
This fixes the "event loop already running" error.

For prompt_toolkit 2.0: this will still nest the event loops, but that's fine,
because we have the asyncio event loop running an prompt_toolkit's custom loop.
They don't interfere with each other.

For prompt_toolkit 3.0: we only use one asyncio event loop when "%gui asyncio"
is used. The terminal prompt will now use the active asyncio event loop.

### DIFF
--- a/IPython/terminal/pt_inputhooks/asyncio.py
+++ b/IPython/terminal/pt_inputhooks/asyncio.py
@@ -28,6 +28,10 @@ prompt_toolkit`s `patch_stdout`)::
 
 """
 import asyncio
+from prompt_toolkit import __version__ as ptk_version
+
+PTK3 = ptk_version.startswith('3.')
+
 
 # Keep reference to the original asyncio loop, because getting the event loop
 # within the input hook would return the other loop.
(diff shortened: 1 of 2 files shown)


IPython/terminal/pt_inputhooks/asyncio.py
```diff
@@ -28,13 +28,30 @@
 
 """
 import asyncio
+from prompt_toolkit import __version__ as ptk_version
+
+PTK3 = ptk_version.startswith('3.')
```
minrk: This will result in incorrect logic when pt4 comes out. Let's parse it out correctly:

```suggestion
PTK3 = int(ptk_version.split('.', 1)[0]) >= 3
```

Carreau: My guess is PTK4 will also have a different API; so wether we go through the PTK2 or PTK3 branch it will fail.

It's also the same check we have in other locations... But up to @jonathanslenders

Carreau:
Should that go in a 7.11 or a 7.10.2 ?

IPython/terminal/pt_inputhooks/asyncio.py
jonathanslenders: I don't really have any opinion. In setup.py, we explicitly mention to not install PTK4, and at the point where we upgrade, it's easy enough to fix this to whatever makes sense.

jonathanslenders:
@Carreau : I consider it a bugfix, not sure in what version it should be merged.

Carreau:
Ok, let's get that in, I'll try to do a patch release soon.

PR #11979, opened 2019-11-30
Add prompt toolkit 3 compatibility to ipdb

Carreau:
closes #11962
closes #11975

PR #12003, opened 2019-12-04
bugfix for 7.10.1: revert inputhook param location

JBKahn:
In https://github.com/ipython/ipython/pull/11979, a fix was made for `ipdb` however, the location of the param lookup was changed by accident? it is certainly breaking it for me.


Carreau:
Thanks, I need to get more sleep I guess.

JBKahn:
Happens to the best of us, thanks for getting back to me pretty fast!

JBKahn:
@Carreau when is the next scheduled release?

Carreau:
Worse case End of month I do a 7.11, best case as soon as I have time, and figure out if PRs like 12001 need to go in a 7.10.2 or not.

PR #12007, opened 2019-12-04
Fix the right inputhook

Carreau:
Interactive shell and debugger have similar code. The fix was applied to
the wrong one, not sure how test were passing.


Carreau:
See #12003 and cc @JBKahn maybe we'll get it right ?

Carreau:
Note: this is not triggered by test as they are not in a real tty.

JBKahn:
Wow, what a day. That's confusing.

PR #12028, opened 2019-12-11
Use separate asyncio event loop for user input in debugger.

jonathanslenders:
This also fixes issues where ipdb is used in threads other then the main thread, and when it's used within an asyncio application itself.
I didn't test extensively, but think it's fine.


Carreau:
Thanks a lot, I should have thought about that. 

I think at some point I need to refactor those two areas to use helper functions to avoid fixing things twice.

PR #12054, opened 2019-12-21
Make stack_data optional, defer to old_ultratb if not present

alexmojaki:


alexmojaki:
Oops, this was meant to be a PR into my PR, as in: https://github.com/alexmojaki/ipython/pull/1

I'll delete this if you prefer.

Anyway, you can see the idea either way. `ultratb.py` now only contains:

```python
try:
    import stack_data
except ImportError:
    from .ultratb_old import *
else:
    from .ultratb_new import *
```

while `ultratb_old.py` is the same as `ultratb.py` on master and `ultratb_new.py` is the new version as per my PR.

This is in response to @Carreau saying:

> I'm happy to have an opt-in flag in IPython to use those if installed and see how it feels as a user for a couple of releases; and then switch to default once deemed good enough.

and I'm trying to figure out the best way to implement that.

I thought this would be pretty neat as it makes the switch seamless and keeps all imports of ultratb the same, but it completely destroys the git history and I don't see a way around that. I can't seem to get git to recognise any of this as a rename:

```
$ git mv ultratb.py ultratb_new.py
$ git status
On branch stack-data-optional
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        renamed:    ultratb.py -> ultratb_new.py

$ touch ultratb.py     
$ git status
On branch stack-data-optional
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        renamed:    ultratb.py -> ultratb_new.py

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        ultratb.py

$ git add ultratb.py 
$ git status        
On branch stack-data-optional
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   ultratb.py
        new file:   ultratb_new.py

```

So maybe if we want to preserve history and git blame the three files should be:

- ultratb: as it is now in #11886 
- ultratb_old
- ultratb_dispatch - which everything else must import.

Any thoughts?

PR #12055, opened 2019-12-21
VI-mode in prompt

jonathf:
From the code it is apparent that if vi-mode is enabled, it should
display in the prompt. Both in insert and navigation mode:

    [ins] In [1]:
    [nav] In [1]:

However, as a compare is done between an Enum instance and a string, it will always return False, irrespectively of the value.

The solution is either to access the Enum's `.value` attribute, or to
compare it to the declared Enum value directly. The latter is likely
more readable.

I have also added support for `InputMode.INSERT` and `InputMode.NAVIGATION` which the shell provides when I install from source.

### DIFF
--- a/IPython/terminal/prompts.py
+++ b/IPython/terminal/prompts.py
@@ -7,6 +7,7 @@ from IPython.core.displayhook import DisplayHook
 
 from prompt_toolkit.formatted_text import fragment_list_width, PygmentsTokens
 from prompt_toolkit.shortcuts import print_formatted_text
+from prompt_toolkit.enums import EditingMode
 
 
 class Prompts(object):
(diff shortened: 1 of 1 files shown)


jonathf:
Not sure what is going on with Travis. But looks like other PRs are getting the same installation error as here.

Carreau:
Don't worry about the tests; it's a change of travis/homebrew where the python-brew-cask seem to have the wrong checksum and refuses to install

Carreau:
Ha, and my guess is the issue only appears with Prompt toolkit 3 ?

jonathf:
Yes, that is correct. The `InputMode.` prefix appears with `prompt-toolkit` version 3. Changing it to version 2, and it reverts to the `vi-` prefix.

Carreau:
Thanks ! I rebased to fix the test which were an issue with master. Merging. 

I'll try with both prompt_toolkit 2 and 3 before merging.

Carreau:
Looks good ! Thanks !

Issue #12012, opened 2019-12-05
IPython/utils/py3compat.py unused `get_closure` function. 


Carreau:
The function `get_closure` seem to be unused; 

we should :  check if it is used, if not remove the function .

joshz123:
Should the function be removed? Is there any update on this issue?

mijiturka:
This has made it into master with PR #12060
@meeseeksdev close

PR #12060, opened 2019-12-25
Removed IPython/utils/py3compat.py unused get_closure function. #12012

kousikmitra:
`get_closure` function since it is not being used anywhere.


Carreau:
Rebased to restart the tests as they've been fix on master.

PR #12061, opened 2019-12-26
Remove 'annotate' function from IPython/utils/py3compat.py

pratyaypandey:
As it has been seen that the "annotate" function of the py3compat.py file in IPython/utils is a relic of Python 2 compatibility - which is no longer required as of the EOL of Python 2 in 2020 - I have removed the function and made sure that there were no uses of the function anywhere, which turned out to be the case.

### DIFF
--- a/IPython/utils/py3compat.py
+++ b/IPython/utils/py3compat.py
@@ -190,20 +190,6 @@ PY2 = not PY3
 PYPY = platform.python_implementation() == "PyPy"
 
 
-def annotate(**kwargs):
-    """Python 3 compatible function annotation for Python 2."""
-    if not kwargs:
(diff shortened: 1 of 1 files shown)


Carreau:
Apologies, but someone sent a similar PR a few days before you (#12052), so I went with the first one; and hence now there are conflicts. 

Thanks for spending the time to do this; I'm going to close but I'm looking forward to your next contribution !

BitLegion:
No problem! Thanks for letting me know.

By the way, I wanted to help contribute more to IPython and later on, to
other open sources. However, being new to this concept, I'm not entirely
sure where to find errors or things to work on and how to test them
subsequently. I would really appreciate it if you could take some time to
tell me.

Thanks,

Pratyay Pandey

PR #12073, opened 2020-01-02
Added arm64 jobs in Travis-CI

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 jobs in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,17 +65,37 @@ after_success:
 
 matrix:
   include:
-    - python: "3.7"
+    - arch: amd64
+      python: "3.7"
       dist: xenial
       sudo: true
-    - python: "3.8-dev"
+    - arch: arm64
+      python: "3.7"
       dist: xenial
       sudo: true
-    - python: "3.7-dev"
+    - arch: amd64
+      python: "3.8-dev"
       dist: xenial
       sudo: true
-    - python: "nightly"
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -223,7 +223,7 @@ def is_stable(extra):
 htmlhelp_basename = 'ipythondoc'
 
 intersphinx_mapping = {'python': ('https://docs.python.org/3/', None),
-                       'rpy2': ('https://rpy2.readthedocs.io/en/version_2.8.x/', None),
+                       'rpy2': ('https://rpy2.readthedocs.io/en/latest/', None),
```
rhenwood-arm: You should explain in the commit message why you have updated this line, as it is not obvious why it relates to adding an Arm64 Travis-ci job.

Carreau: That has been fixed in master; I'll rebase so that it does not appear in the PR

Carreau:
Oh, that's a lot of new entry in the test-matrix. Do you think they are all necessary ? It will likely quite increase the time to run all the tests. We also do not have compiled code; so is ti really worth it ; or should we maybe test only on things that would potentially break on CPython ie the -dev or 3.8 and nightly ?

Carreau:
Merging for now; but if that make CI too slow; I might trim a few.

PR #12074, opened 2020-01-04
Exit non-zero when ipython is given a file path to run that doesn't exist

jpotter:
Simple PR to update the exit code for ipython when given a file that doesn't exist from 0 ("success") to 2 (matches python's behavior for this condition).

Currently, running ipython with a file argument that does not exist causes a zero exit code. This can cause problems in automation, i.e. bash scripts that have `-e` set and have an invalid file path.

```
$ ./venv/bin/ipython no-such-file.py
[TerminalIPythonApp] WARNING | File not found: 'foo.py'
$ echo $?
0
```

This PR changes the behavior to:
```
$ ./venv/bin/ipython no-such-file.py
[TerminalIPythonApp] WARNING | File 'no-such-file.py' doesn't exist
$ echo $?
2
```

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -404,6 +404,9 @@ class InteractiveShellApp(Configurable):
             fname = self.file_to_run
             if os.path.isdir(fname):
                 fname = os.path.join(fname, "__main__.py")
+            if not os.path.exists(fname):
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, that seem reasonable,

Issue #12010, opened 2019-12-05
IPython/utils/py3compat.py unused _print_statement_sub.


Carreau:
function `_print_statement_sub` seem unused in code base. 

One should check if this is actually the case, and remove it if necessary.

mijiturka:
This has made it into master with PR #12076
@meeseeksdev close
(Sorry if I'm not supposed to close issues as a brand-new user but going by CONTRIBUTING.md)

Carreau:
> (Sorry if I'm not supposed to close issues as a brand-new user but going by CONTRIBUTING.md)

No that's good ! thanks !

PR #12076, opened 2020-01-07
Remove unused _print_statement_sub function in py3compat #12010

takuya1981:
I checked that the method is not referenced anywhere.


Carreau:
Thanks !

PR #12085, opened 2020-01-12
MNT: ensure we can show un-pickled figures

tacaswell:
For Matplotlib < 3.2 inline figures maybe re-hydrated with the canvas attribute as None.  If that is true, then `print_figure` will fail.

There is a PR into Matplotlib (https://github.com/matplotlib/matplotlib/pull/16189) to fix this going forward, but opening this as a belt-and-suspenders approach.

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -123,14 +123,18 @@ def print_figure(fig, fmt='png', bbox_inches='tight', **kwargs):
     }
     # **kwargs get higher priority
     kw.update(kwargs)
-    
+
     bytes_io = BytesIO()
+    if fig.canvas is None:
+        from matplotlib.backend_bases import FigureCanvasBase
+        FigureCanvasBase(fig)
+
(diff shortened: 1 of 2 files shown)


Carreau:
That only fails on ARM (and nightly), which are broken anyway. So will merge. 

I'm going to assume this should go in a release pretty soon right ? So I should likely backport to a 7.x branch if master goes toward a 8.x ?

tacaswell:
Goal is mpl3.2 within a few weeks which will have the fix on the mpl side so up to you if you want to merge and/or back port this.  This is also an extremely (as in forever) long standing bug that users have to go a little bit out of their way to hit (pickle an inline figure in a different cell than you created and displayed it).

Carreau:
Ok, thanks. I'm going to merge this before actually forking to a 8.0 then, so there is no need to backport; there is about 3 weeks to end of month; so I might end up doing  a 7.12 before making a 7.x.

(Merging even if CI red as it's only failing on ARM which is removed from master)

PR #12086, opened 2020-01-12
MNT: update pytest.ini to work with pytest > 3

tacaswell:


Carreau:
(Merging even if CI red as it's only failing on ARM which is removed from master)

Thanks !

tacaswell:
I was annoyed at having to downgrade pytest to test the other PR ;)

Issue #11816, opened 2019-07-07
Road to 8.0.


Carreau:
This is a meta issue to track the road to IPython 8.0
It is there to track the potential large changes and discuss them.

I'll try to keep this top comment edited to reflect the status of the thread.

Carreau:
Suggestions

1) use black on all the things and have a bot/ci that autoformat pull-requests. IE not _fail_ on improperly formatted code to not annoy user. The main question is: do we run black at once on the full codebase ?

2) Drop Nose, test using Pytest. It's still a bit annoying as we are so far stuck on pytest 3, and moving to pytest 4 or 5 will be a significant effort. Some test still only run w/ nose (doctests) and need to be migrated.

3) Completer refactor... this can be a lot of work and is a change of API.

gpotter2:
I'd like to have "Resolve the issues with the completer" on the roadmap.
see https://github.com/ipython/ipython/pull/11804

Carreau:
Yes, that would be a good idea as well.

terrdavis:
Suggestions:
* Remove [deprecated](https://github.com/ipython/ipython/blob/4842416da03f3cc04897bd8731e851c2f0c0e734/IPython/core/debugger.py#L247) [color_scheme](https://github.com/ipython/ipython/blob/4842416da03f3cc04897bd8731e851c2f0c0e734/IPython/core/debugger.py#L221) argument in `core.debugger`.
* Remove other deprecated items (esp. those deprecated before ipython5).
* Remove python 2 specific imports. I found `reprlib`, `winreg`, and `tkinter` with a search for `py 2`.

terrdavis:
> Suggestions
>
>1. use black on all the things and have a bot/ci that autoformat pull-requests. IE not fail on improperly formatted code to not annoy user. The main question is: do we run black at once on the full codebase ?
>...

[Yes](https://twitter.com/llanga/status/1163767833706323970)
It appears that it's possible to specify commits to be ignored by git blame by pointing `blame.ignoreRevsFile` to a file with the commit hashes corresponding to formatting commits.

Carreau:
> * Remove python 2 specific imports.

That does not need to wait for 8.0 if they are 2.x only.


> It appears that it's possible to specify commits to be ignored by git blame by pointing `blame.ignoreRevsFile` to a file with the commit hashes corresponding to formatting commits.

[...]

PR #12091, opened 2020-01-21
Support git blame ignore revs

terrdavis:
In response to this [comment](https://github.com/ipython/ipython/issues/11816#issuecomment-554671200).

Please note:
- neither Github nor GitLab supports this feature.
- the IntelliJ plugin [GitToolBox](https://github.com/zielu/GitToolBox) supports this feature, so perhaps other client-side git tools do too?

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,8 @@
+# When making commits that are strictly formatting/style changes, add the
+# commit hash here, so git blame can ignore the change.
+# See docs for more details:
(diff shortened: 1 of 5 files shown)


Carreau:
Thanks ! That is great. 

Even if it's not (yet) supported by github/gitlab, when we do such commits we can link to this documentation explicitly, and see if some clarifications are needed. 

I'll need to fix MANIFEST.in to `exclude .git-blame-ignore-revs` or test are failling; i'll try to do that later unless you beat me to it.

Carreau:
Thanks !

PR #12092, opened 2020-01-21
Remove Python 2 workarounds

terrdavis:
I don't know if docstring changes belong in a different PR.
I'd be happy to split this one up if needed.
This also includes one or two things related to python<3.5.

One of the docstring changes I made included a note about using a context manager. Could I replace the begin/end_group pairs with a context manager in this PR?
```python
    def begin_group(self, indent=0, open=''):	
        """
        Begin a group.  If you want support for python < 2.5 which doesn't has	        Begin a group.
        the with statement this is the preferred way:	
            p.begin_group(1, '{')	
            ...	
            p.end_group(1, '}')	
        The python 2.5 expression would be this:	
            with p.group(1, '{', '}'):	
                ...	
        The first parameter specifies the indentation for the next line (usually	        The first parameter specifies the indentation for the next line (usually
        the width of the opening text), the second the opening text.  All	        the width of the opening text), the second the opening text.  All
        parameters are optional.
```

I also found this test, with a TODO related to python 3.6+ support. I'm not sure what should be removed.
```python
def test_instantiation_FileLink():
    """FileLink: Test class can be instantiated"""
    fl = display.FileLink('example.txt')
    # TODO: remove if when only Python >= 3.6 is supported
    fl = display.FileLink(pathlib.PurePath('example.txt'))
```

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -8,13 +8,7 @@ import atexit
 import datetime
 import os
 import re
-try:
-    import sqlite3
-except ImportError:
-    try:
-        from pysqlite2 import dbapi2 as sqlite3
-    except ImportError:
-        sqlite3 = None
+import sqlite3
 import threading
 
 from traitlets.config.configurable import LoggingConfigurable
@@ -49,26 +43,8 @@ class DummyDB(object):
         pass
 
 
-@decorator
-def needs_sqlite(f, self, *a, **kw):
-    """Decorator: return an empty list in the absence of sqlite."""
-    if sqlite3 is None or not self.enabled:
-        return []
-    else:
-        return f(self, *a, **kw)
-
-
-if sqlite3 is not None:
-    DatabaseError = sqlite3.DatabaseError
-    OperationalError = sqlite3.OperationalError
-else:
-    @undoc
-    class DatabaseError(Exception):
-        "Dummy exception when sqlite could not be imported. Should never occur."
-    
-    @undoc
-    class OperationalError(Exception):
-        "Dummy exception when sqlite could not be imported. Should never occur."
+DatabaseError = sqlite3.DatabaseError
+OperationalError = sqlite3.OperationalError
 
 # use 16kB as threshold for whether a corrupt history db should be saved
 # that should be at least 100 entries or so
@@ -302,7 +278,6 @@ class HistoryAccessor(HistoryAccessorBase):
             return ((ses, lin, (inp, out)) for ses, lin, inp, out in cur)
         return cur
 
-    @needs_sqlite
     @catch_corrupt_db
     def get_session_info(self, session):
         """Get info about a session.
@@ -558,7 +533,6 @@ class HistoryManager(HistoryAccessor):
         profile_dir = self.shell.profile_dir.location
         return os.path.join(profile_dir, 'history.sqlite')
     
-    @needs_sqlite
     def new_session(self, conn=None):
         """Get a new session number."""
         if conn is None:
@@ -769,7 +743,6 @@ class HistoryManager(HistoryAccessor):
                 conn.execute("INSERT INTO output_history VALUES (?, ?, ?)",
                                 (self.session_number,)+line)
 
-    @needs_sqlite
     def writeout_cache(self, conn=None):
         """Write any entries in the cache to the database."""
         if conn is None:
@@ -818,7 +791,6 @@ class HistorySavingThread(threading.Thread):
         self.enabled = history_manager.enabled
         atexit.register(self.stop)
 
-    @needs_sqlite
     def run(self):
         # We need a separate db connection per thread:
         try:
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -43,7 +43,7 @@ class DummyDB(object):
 
 
 @decorator
-def needs_sqlite(f, self, *a, **kw):
+def only_when_enabled(f, self, *a, **kw):
     """Decorator: return an empty list in the absence of sqlite."""
     if not self.enabled:
         return []
@@ -277,7 +277,7 @@ class HistoryAccessor(HistoryAccessorBase):
             return ((ses, lin, (inp, out)) for ses, lin, inp, out in cur)
         return cur
 
-    @needs_sqlite
+    @only_when_enabled
     @catch_corrupt_db
     def get_session_info(self, session):
         """Get info about a session.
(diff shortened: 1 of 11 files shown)


IPython/core/history.py
```diff
@@ -8,13 +8,7 @@
...
 import os
 import re
-try:
-    import sqlite3
-except ImportError:
-    try:
-        from pysqlite2 import dbapi2 as sqlite3
-    except ImportError:
-        sqlite3 = None
+import sqlite3
```
Carreau: I think that not quite right; this make sqlite3 mandatory, before it would set slite3 to None, and still work.

IPython/core/magics/execution.py
```diff
@@ -19,16 +19,8 @@
...
-    import cProfile as profile
-    import pstats
-except ImportError:
-    # profile isn't bundled by default in Debian for license reasons
-    try:
-        import profile, pstats
-    except ImportError:
-        profile = pstats = None
+import cProfile as profile
+import pstats
```
Carreau: same here `# profile isn't bundled by default in Debian for license reasons` so this code is necessary for debian.

Carreau:
Thanks for doing this ! 

It looks like the test failures are genuine; but in principle from what I have read I do not have any objections to the code change. 

I guess it would be easier in smaller chunks to figure out what is failing.

In general if you want to change some things to be context managers; I do not have objections.

IPython/core/history.py
terrdavis: sqlite3 was [added in python 2.5](https://docs.python.org/2/library/sqlite3.html), so I assumed this import would never fail.

IPython/core/magics/execution.py
terrdavis: I saw that, but I assumed that the cProfile import couldn't fail since it was added in 2.5. Is the cProfile import ever expected to fail?

terrdavis:
You're welcome!  

I'm confused though, do I need revert the changes that you commented on?  

Can you also clarify what's meant by the comment in this test?
```python
def test_instantiation_FileLink():
    """FileLink: Test class can be instantiated"""
    fl = display.FileLink('example.txt')
    # TODO: remove if when only Python >= 3.6 is supported
    fl = display.FileLink(pathlib.PurePath('example.txt'))
```

Otherwise, I'll split this up into more pull requests and add the context manager I asked about!

Carreau:
> do I need revert the changes that you commented on

Maybe not revert, but the changes are going too far, for examplein 

```diff
-try:
-    import sqlite3
-except ImportError:
-    try:
-        from pysqlite2 import dbapi2 as sqlite3
-   except ImportError:
-        sqlite3 = None
+ import sqlite3
```
It probably should be more like: 

```diff
try:
    import sqlite3
except ImportError:
-    try:
-         from pysqlite2 import dbapi2 as sqlite3
-     except ImportError:
        sqlite3 = None
import sqlite3
```

Where we only remove the Py3 specific versions and keep the `@needs_sqlite` which is specific for debian.

Carreau:
```
# TODO: remove if when only Python >= 3.6 is supported
fl = display.FileLink(pathlib.PurePath('example.txt'))
```

I hav to check, but that likely mean that one of these two lines is  unnecessary now that we're Python 3.6+ only.

terrdavis:
Apparently, cProfile is [included in Debian](https://wiki.debian.org/Python#Deviations_from_upstream):
> The standard profile and pstats modules are not included, due to restrictive redistribution clauses in their license that don’t meet the DFSG. **cProfile and hotshot are however included.**

Is there any other case where cProfile isn't available?

terrdavis:
@Carreau, can you respond to my last message? As far as I can tell, cProfile and sqlite3 are included in Debian’s python 3.

Carreau:
Sorry, I'll have a look. It might have been an old policy in Debian, in which case we can revisit it, and I'm ok to have those import unconditionally present.

IPython/core/magics/execution.py
Carreau: As far as I can tell debian used to build python w/o some modules;  it seem to not be the case anymore. But yes `import profile` used to be able to fail.

IPython/core/history.py
Carreau: That seem to have been added in 2012; (#2718), so my guess is some people have-had custom builds that do not have sqlite3  for whatever reason. 

It's been 6 years so let's see if this is still necessary. If we have complaints we can re-add.

IPython/core/history.py
```diff
@@ -49,26 +43,8 @@ def __exit__(self, *args, **kwargs):
...
-else:
-    @undoc
-    class DatabaseError(Exception):
-        "Dummy exception when sqlite could not be imported. Should never occur."
-    
-    @undoc
-    class OperationalError(Exception):
-        "Dummy exception when sqlite could not be imported. Should never occur."
+DatabaseError = sqlite3.DatabaseError
+OperationalError = sqlite3.OperationalError
```
Carreau: This create actual test failures;  The decorator has the following statement:

```
if sqlite3 is None or not self.enabled:
```

Note the `or not self.enabled`; So this decorator is not only about sqlite being available.

Carreau:
I've pushed a commit that fix a merge conflict.

terrdavis:
I meant to finish off these changes on a new branch (i.e. remove-py2-workarounds-2of2) but my local branch still had this as its remote. I rebased on to master in the meantime. Let me know if you'd prefer that I close this PR and move it to a new one.

Carreau:
> I meant to finish off these changes on a new branch (i.e. remove-py2-workarounds-2of2) but my local branch still had this as its remote. I rebased on to master in the meantime. Let me know if you'd prefer that I close this PR and move it to a new one.

Sorry I've been super busy recently. 

I think something got messed up in the rebase. I'll have a look tomorrow but it should be fixable. If not we can do another PR but that should not be necessary.

Carreau:
Note to self, current head is e507997163fc97cf46fc3005331ba5d9f1aeb992

Carreau:
Ok, let's try this rebase – I kept thinking about it so had to do it tonight – 13 commits instead of 54.
And it looks good, let's let CI do its thing, and if it pass i'll merge tomorrow.

Carreau:
... Or later tonight depending as to whether I'm sleeping or not.

Carreau:
Woooo

PR #12095, opened 2020-01-22
[5.x] Use the proper python2 pygments lexer

### LINKED ISSUE #12094: IPythonConsoleLexer fails with python2 and pygments 2.5
As of pygments 2.5, `PythonLexer` [now](https://pygments.org/docs/changelog/) is an alias for `Python3Lexer`. iPython still uses `PythonLexer` for the python2 case:

https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/lib/lexers.py#L79

It should explicitly use `Python2Lexer` instead.

timokau:
Since pygments 2.5 `PythonLexer` refers to python3 [1]. We have to be
explicit if we want to use python2.

[1] https://pygments.org/docs/changelog/

Fixes #12094.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -45,11 +45,15 @@ from IPython.utils.signatures import signature
 from IPython.utils.colorable import Colorable
 
 from pygments import highlight
-from pygments.lexers import PythonLexer
+try:
+    # PythonLexer was renamed to Python2Lexer in pygments 2.5
+    from pygments.lexers import Python2Lexer
+except ImportError:
+    from pygments.lexers import PythonLexer as Python2Lexer
 from pygments.formatters import HtmlFormatter
 
 def pylight(code):
-    return highlight(code, PythonLexer(), HtmlFormatter(noclasses=True))
+    return highlight(code, Python2Lexer(), HtmlFormatter(noclasses=True))
 
 # builtin docstrings to ignore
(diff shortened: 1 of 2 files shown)


timokau:
One travis failure looks like pygments wasn't updated to 2.5 yet on that instance. I've amended the commit to be backwards-compatible.

The other failures seem unrelated.

timokau:
WIP: I need to re-test if this actually fixes my problem.

timokau:
Turns out in my case it was actually sphinx that caused the issue, not ipython. Still, this should be the right thing to do. Good to go in my view.

Carreau:
I'll try to get that in and maybe do a source release at least on PyPI; but no promises; 5.x have not seen a fix or a release in over 18 month.

timokau:
That's okay, python2 is EOL after all. I mostly fixed this for myself while debugging a sage issue. I already applied this patch on top of my distro (https://github.com/NixOS/nixpkgs/pull/78348), so if you don't want to spend too much time on this that's fine by me.

Carreau:
CI Failure is unrelated; and we are not actively hostile against Python2 users. So I don't see any reasons not to merge. There are also a couple of other fixes that would make sens in a 5.9

Carreau:
Should be released on PyPI both at source and wheels.

PR #12097, opened 2020-01-24
Trio loop

mehaase:
This PR adds the foundation for running a long-lived Trio loop on the main thread. Most of the implementation is in [a PR for `ipykernel`](https://github.com/ipython/ipykernel/pull/479). All this does is add a setter for the shell's trio runner, which `ipykernel` uses to set a trio runner that interacts with a long-lived loop.


Carreau:
That looks reasonable to me; I'll try to get that in 7.12 this afternoon at some point.

PR #12099, opened 2020-01-27
Start multi-line __repr__s on their own line

### LINKED ISSUE #12098: Pretty printing of lists of objects with multiline reprs is suboptimal
Consider
```python
def _rep(first, mid, n):
    return first + mid*(n-1) + first*(n>0)
class Rectangle:
    def __init__(self, rows, cols):
        self.rows = rows
        self.cols = cols
    def __repr__(self):
        return '\n'.join(_rep([_rep('+', '-', self.cols)],
                              [_rep('|', '#', self.cols)], self.rows))
```
In Ipython, I get:
```
In [78]: [Rectangle(3, 4), Rectangle(2, 2)]
Out[78]:
[+---+
 |###|
 |###|
 +---+, +-+
 |#|
 +-+]
```
however, I'd expect either of
```
Out[78]:
[+---+
 |###|
 |###|
 +---+, +-+
        |#|
        +-+]
```
or
```

Out[78]:
[+---+
 |###|
 |###|
 +---+,
 +-+
 |#|
 +-+]
```
xref numpy/numpy#11270, which I'd prefer wasn't necessary to solve this for numpy

eric-wieser:
Fixes gh-12098, by producing the output in the second code-block on that page:
```
In [115]: [Rectangle(3, 4), Rectangle(6, 2), 0, 1]
Out[115]:
[+---+
 |###|
 |###|
 +---+,
 +-+
 |#|
 |#|
 |#|
 |#|
 |#|
 +-+, 0, 1]
```

Related to gh-3851

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -206,7 +206,7 @@ class PrettyPrinter(_PrettyPrinterBase):
             self.output_width = x.output(self.output, self.output_width)
             self.buffer_width -= x.width
 
-    def _break_outer_groups(self, force=):
+    def _break_outer_groups(self):
         while self.max_width < self.output_width + self.buffer_width:
             group = self.group_queue.deq()
             if not group:
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -196,19 +196,22 @@ class PrettyPrinter(_PrettyPrinterBase):
         self.group_queue = GroupQueue(root_group)
         self.indentation = 0
 
-    def _break_outer_groups(self):
+    def _break_one_group(self, group):
+        while group.breakables:
+            x = self.buffer.popleft()
+            self.output_width = x.output(self.output, self.output_width)
+            self.buffer_width -= x.width
+        while self.buffer and isinstance(self.buffer[0], Text):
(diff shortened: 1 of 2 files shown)


IPython/lib/pretty.py
```diff
@@ -196,19 +196,22 @@ def __init__(self, output, max_width=79, newline='\n', max_seq_length=MAX_SEQ_LE
...
+        while group.breakables:
+            x = self.buffer.popleft()
+            self.output_width = x.output(self.output, self.output_width)
+            self.buffer_width -= x.width
+        while self.buffer and isinstance(self.buffer[0], Text):
+            x = self.buffer.popleft()
+            self.output_width = x.output(self.output, self.output_width)
+            self.buffer_width -= x.width
+
+    def _break_outer_groups(self, force=):
```
Carreau: You appear to have a syntax error here, and `force` seem unused. Did you forgot some commits ?

eric-wieser:
```suggestion
    def _break_outer_groups(self):
```

Nope, just clumsy in-browser editing

Carreau:
Thanks !

PR #12110, opened 2020-02-01
Remove py2 workarounds 1/2

terrdavis:
Breaking up #12092, per [discussion](https://github.com/ipython/ipython/pull/12092#issuecomment-580899602).

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -101,7 +101,7 @@ import time
 import tokenize
 import traceback
 
-generate_tokens = tokenize.tokenize
+from tokenize import generate_tokens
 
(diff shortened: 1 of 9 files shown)


IPython/core/ultratb.py
```diff
@@ -101,10 +101,7 @@
 import tokenize
 import traceback
 
-try:  # Python 2
-    generate_tokens = tokenize.generate_tokens
-except AttributeError:  # Python 3
-    generate_tokens = tokenize.tokenize
+generate_tokens = tokenize.tokenize
```
Carreau: The comments appear to be wrong, `generate_tokens = tokenize.generate_tokens` work on Python 3; and this is likely the cause for the  test failures.

PR #12113, opened 2020-02-01
Speedup _make_wav

jonashaag:
When numpy is installed, ~30x speedup; otherwise, ~5x speedup


Carreau:
Looks great ! Thanks.

Should be released at end of month !

PR #12114, opened 2020-02-01
Attempt global reformatting of 1 file; and add ignored revisions.

Carreau:
This is more interesting to look at on a per commit basis.

It runs some auto-reformatting; and then add the commits hash to a list of hashes to ignore.
It is not (yet) suported by github/gitlab, but works great locally where blame is not screwed up.

The goal would be to always automatically reformat the source code.


Carreau:
might want to redo , but update pyupgrade first and run it with --py36-plus flag.

Issue #12090, opened 2020-01-20
In interactive shell, ExecutionResult.result = None when silent = True


adammunich:
In interactive shell, ExecutionResult.result = None, always, when silent = True in the run_cell() method. Yet, ExecutionResult.success is true, and no error_before_exec, or, error_in_exec is populated. 

Is this normal behavior? It would seem run_cell isn't all that useful with silent=True.

TomNicholas:
I'm seeing the same problem, here's a minimum example:
```python
from IPython.core.interactiveshell import InteractiveShell

In [19]: InteractiveShell().run_cell(raw_cell="x=5 \nx") 
    ...:                                                                                                                                                                                      
Out[1]: 5
Out[19]: 
<ExecutionResult object at 7fd41a854210, execution_count=None error_before_exec=None error_in_exec=None info=<ExecutionInfo object at 7fd41a854e10, raw_cell="x=5 
x" store_history=False silent=False shell_futures=True> result=5>

In [20]: InteractiveShell().run_cell(raw_cell="x=5 \nx", silent=True) 
    ...:                                                                                                                                                                                      
Out[20]: 
<ExecutionResult object at 7fd41a858d90, execution_count=None error_before_exec=None error_in_exec=None info=<ExecutionInfo object at 7fd41a858610, raw_cell="x=5 
x" store_history=False silent=True shell_futures=True> result=None>
                                      x")
```

PR #12120, opened 2020-02-10
[WIP] New --return-result argument to %%capture cell magic

### LINKED ISSUE #3039: Exporting a single jupyter cell output
I checked the current issues and could not find a similar request:

We would very much like to be able to export the output from a highlighted cell to PDF or PNG or anything, really. Apparently this is possible in Mathematica, but not yet in Jupyter Notebook. 

The background comes from this: We're able to style Pandas Dataframes to show them just the way we need for our certification reports. However, after that, there is no "pyplot.savefig"-like function to store the output thereof in a format that would allow us to include it in other documents. 

Googling for solutions to such issues brings up a host of stackOverflow semi-answers that offer only a very convoluted way to recreate / rebuild the table somewhere else (using plotly or matplotlib tables). That's a big waste of the effort put into the pandas styler. 

Since Jupyter already allows you to export the entire notebook, is it possible to implement this on a per-cell basis as well, preferably only the cell output?

### LINKED ISSUE #8015: %%capture cell magic does not suppress automatic output
Is this on purpose? If yes, how can I suppress this kind of output?
```Python
In [1]: x = 42

In [2]: %%capture
   ...: x
Out[2]: 42
```

TomNicholas:
Intended to close #8015 by making the suppression of the captured result optional. 

I want this so that I can capture cell output and use it within the same cell, while still displaying it to the screen. I eventually want to wrap that non-suppressing capture magic within a new `%%save_output` cell magic in order to close [this jupyter issue](https://github.com/jupyter/notebook/issues/3039).

However, it's not working yet - executing this cell in jupyterlab
```python
%%capture --return-result out

x = 5
x
```
returns
```
<ExecutionResult object at 7f72d0a49f90, execution_count=None error_before_exec=None error_in_exec=None info=<ExecutionInfo object at 7f72d0a499d0, raw_cell="
x = 5
x
" store_history=False silent=False shell_futures=True> result=None>
```
which makes sense, except that the `ExecutionResult` object has attribute `result=None`. I would have expected it to have `result=5`. Obviously I can't return the result if it's always being set to `None`.

Confusingly this does work in the interactive ipython shell, just not in the notebook:
```python
In [1]: from IPython.core.interactiveshell import InteractiveShell
In [2]: InteractiveShell().run_cell(raw_cell="x=5 \nx")                                                                                                                                                                        
Out[1]: 
<ExecutionResult object at 7fd41a854210, execution_count=None error_before_exec=None error_in_exec=None info=<ExecutionInfo object at 7fd41a854e10, raw_cell="x=5 
x" store_history=False silent=False shell_futures=True> result=5>
```

This problem might be related to #12090 ?

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1443,6 +1443,9 @@ python-profiler package from non-free.""")
     @magic_arguments.argument('--no-display', action="store_true",
         help="""Don't capture IPython's rich display."""
     )
+    @magic_arguments.argument('--return-result', action="store_true",
+        help="""Return the captured result instead of suppressing it."""
+    )
     @cell_magic
     def capture(self, line, cell):
         """run the cell, capturing stdout, stderr, and IPython's rich display() calls."""
@@ -1451,9 +1454,13 @@ python-profiler package from non-free.""")
         err = not args.no_stderr
         disp = not args.no_display
         with capture_output(out, err, disp) as io:
-            self.shell.run_cell(cell)
+            result = self.shell.run_cell(cell)
         if args.output:
             self.shell.user_ns[args.output] = io
+        if args.return_result:
+            return result
+        else:
+            return None
 
 def parse_breakpoint(text, current_file):
     '''Returns (file, line) for file:line and (current_file, line) for line'''


Carreau:
it may be that %%capture is redefined in ipykernel ? The execution path is a tiny bit different in both because of wether we are actually attached to a terminal or not.

rceballos98:
Is this ready for merging? This feature would be extremely useful.

bpartridge:
I would love this as well, but I think there's a lot more to do. Specifically for the notebook to meet the goals of #8015 I believe you'd need to patch, if the argument is provided, all the hooks set up in this function to be "taps" that also pass through their output to the existing hooks. https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/utils/capture.py#L136

Also just IMO, but would `--show-output` be a better name than `--return-result`?

ijoseph:
Yes, this would be extraordinarily useful. Basically now, we have to choose whether to see either:
1.  intermediate progress from a long-running command immediately, or 
2. the ability to see the output if our connection to the UI gets broken. 

The issue with 1. is that of course, if the output gets broken, we see nothing. The issue with 2. is that we can't see intermediate cell output which is extremely helpful in verifying whether something is running, and sometimes helpful in deciding to halt execution entirely.

Carreau:

PR #12123, opened 2020-02-12
Change version number to reflect that we're working toward a 8.x

Carreau:


Carreau:
Failure is travis specific, bypassing CI and merging. Will fix in subsequent PR.

Issue #11530, opened 2018-12-14
Tab completions for paths broken again in 7.2.0


ivirshup:
Tab completion for paths within strings seems to be broken again (previously noted many places, like #10961, #10996, [notebook issue #3333](https://github.com/jupyter/notebook/issues/3333)). Here's what I get in `v7.2.0`:

![screen shot 2018-12-14 at 12 39 17 pm](https://user-images.githubusercontent.com/8238804/49978242-b667bb80-ff9d-11e8-9691-6a6bb598015c.png)

Compared to the behavior in `v7.1.1`:

![screen shot 2018-12-14 at 12 38 20 pm](https://user-images.githubusercontent.com/8238804/49978239-b23b9e00-ff9d-11e8-80f6-fe5ba2524d9a.png)

Probably related to #11503, as disabling Jedi (setting `c.IPCompleter.use_jedi = False` in `ipython_config.py`) gives the correct behavior.

<details>

<summary>System info</summary>

```
{'commit_hash': '523ed2fe5',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/usr/local/lib/python3.6/site-packages/IPython',
 'ipython_version': '7.2.0',
 'os_name': 'posix',
 'platform': 'Darwin-17.7.0-x86_64-i386-64bit',
 'sys_executable': '/usr/local/opt/python/bin/python3.6',
 'sys_platform': 'darwin',
 'sys_version': '3.6.5 (default, Jun 17 2018, 12:13:06) \n'
                '[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)]'}
```

</details>

mimakaev:
I have the same issue, ipython 7.2.0.

jvanheugten:
Downgrading to ipython==7.1.1 seemed to work as a fix for me.

vityas:
Could someone share if enabling `jedi` help anybody's workflow, and if it did, I would like to know how.

It certainly made my life much harder: no multi level completion, no folder completion, nor can i use ustocompletion before running the cell (`df = pd.DataFrame(); df.val<TAB>`fails)

rainwoodman:
The bug certainly has made my life much harder as well.

auto-completion is probably one of the most important feature of an interactive shell. Is there a way jedi and path auto-completion can coexist?
[...]

PR #12128, opened 2020-02-13
Improves detection of whether tab-completion is in a string and suppresses Jedi

ecoon:
Refs #10926 and #11530

Jedi results swamp file_matches and dict_key_matches in
tab-completion, which is a real nuisance.  The logic in the jedi
completor tried to catch cases where it was "in a string", but that
logic only looked at the last child in the current token, which was a little fragile,
breaking in lots of cases such as:

    './<tab>
    "mypath/<tab>

etc.

This seems a bit more robust in that it searchs for the first character in
the current parser tree and checks if its value starts with ' or ".

This detection of "in a string" then turns off jedi and returns some
sanity to the set of matches.

Note I have developed this with jedi 0.13.2 -- not sure if it is different logic in older versions.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1371,18 +1371,18 @@ class IPCompleter(Completer):
         try_jedi = True
 
         try:
-            # should we check the type of the node is Error ?
-            try:
-                # jedi < 0.11
-                from jedi.parser.tree import ErrorLeaf
-            except ImportError:
-                # jedi >= 0.11
-                from parso.tree import ErrorLeaf
-
-            next_to_last_tree = interpreter._get_module().tree_node.children[-2]
+            # find the first token in the current tree -- if it is a ' or " then we are in a string
             completing_string = False
-            if isinstance(next_to_last_tree, ErrorLeaf):
-                completing_string = next_to_last_tree.value.lstrip()[0] in {'"', "'"}
+            try:
+                first_child = next(c for c in interpreter._get_module().tree_node.children if hasattr(c, 'value'))
+            except StopIteration:
+                pass
+            else:
+                # note the value may be ', ", or it may also be ''' or """, or
+                # in some cases, """what/you/typed..., but all of these are
+                # strings.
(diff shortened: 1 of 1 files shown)


Carreau:
That would be nice ! Could it even be something that could be upstreamed in Jedi, so we coudl teach jedi to say "I'm in a string" ? or even complete strings ?

I want o merge a PR that touch many files first and then I'm likely to get that merge. Likely tomorrow. Thanks for your patience and contributions !

Carreau:
Side note your commit is attached to your Oak Ridge mail, but github is not aware that you have an Oak Ridge email, sothe contribution will not appear to your profile. 

Do you ant to :
 1) merge as is
 2) edit AUthors
 3) go into github setting and teach it about your oakridge email ?

ecoon:
Thanks, I updated my github profile.

ecoon:
> That would be nice ! Could it even be something that could be upstreamed in Jedi, so we coudl teach jedi to say "I'm in a string" ? or even complete strings ?

I believe so.  The code checking "I'm in a string" is entirely based on jedi's parsing.  I'm not sure what is there in jedi to complete strings -- most of the time useful string completions would be either filenames or keys in a container, so I'm not sure what is/could be added to jedi with regard to that.

Also, if I get time, I'll look at the implementation for "keys in a dict."  Ideally this would be anything that implements keys(), or maybe even anything that implements __contains__, not just dict -- I've noticed that other file objects such as hdf5/netcdf4 files, which implement the keys() method and mimic a dictionary to work within their file structure don't work with tab completion.

Carreau:
> I believe so. The code checking "I'm in a string" is entirely based on jedi's parsing. I'm not sure what is there in jedi to complete strings -- most of the time useful string completions would be either filenames or keys in a container, so I'm not sure what is/could be added to jedi with regard to that.

I'm mostly wondering if we could add completions plugin to jedi instead of hacking around jedi. 

For single-line string I agree, but if we are in a docstring, jedi might be able to help by completing also to the parameter of given functions or similar.

Carreau:
> Thanks, I updated my github profile.

Good, merging.

PR #12130, opened 2020-02-14
Ensure set_custom_completer is working. Fixes #11272

### LINKED ISSUE #11272: set_custom_completer seems to be a no-op
In [interactive_shell.py](https://github.com/ipython/ipython/blob/a2685d78f2403f84e6cb915ae11a4f6033ccc5f6/IPython/core/interactiveshell.py#L2034), `set_custom_completer` does

    self.Completer.matchers.insert(pos,newcomp)

Looking at the source code of IPCompleter, though, [matchers](https://github.com/ipython/ipython/blob/eed56ba34aa3856b5726d60eedfe60a17e1c26c5/IPython/core/completer.py#L1114) seems to be a property returning an ephemeral list, so the insert is a no-op.

To reproduce:

    import IPython

    def foo(*args, **kwargs):
        pass

    print(len(IPython.get_ipython().Completer.matchers))
    IPython.get_ipython().set_custom_completer(foo, 0)
    print(len(IPython.get_ipython().Completer.matchers))

This should print the same number twice (in my version, 5, 5)

ngoldbaum:
See #11272 for details. This implements [the suggestion](https://github.com/ipython/ipython/issues/11272#issuecomment-430590871) from @pganssle that @minrk endorsed there.

This is my first substantive IPython contribution so please let me know if I'm doing something wrong or non-idiomatic somewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ matrix:
     - arch: arm64
       python: "3.7"
       dist: xenial
-      env: ARM64=True
+      env: ARM64=True IPYTHON_TESTING_TIMEOUT_SCALE=2
       sudo: true
     - arch: amd64
       python: "3.8"
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -626,6 +626,8 @@ class Completer(Configurable):
         else:
             self.global_namespace = global_namespace
 
+        self.custom_matchers = []
+
         super(Completer, self).__init__(**kwargs)
 
     def complete(self, text, state):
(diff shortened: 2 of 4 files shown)


ngoldbaum:
I'm pretty sure the ARM64 failure is unrelated?

ngoldbaum:
OK, increasing the timeout does fix it, but at lest in that case it makes the ARM build go from 12 minutes to 18 minutes.

I'm happy to split off the build timeout change into a new PR or revert it if it doesn't belong here, whatever the maintainers prefer.

Carreau:
Yep that's perfect ! Thanks ! Sorry it took me that long to find time to review. 
I've set the milestone to 7.13 so our bot should automatically backport on merge.

Thanks !

PR #12131, opened 2020-02-16
Use Pygments instead of PyColorize to syntax highlight tracebacks

alexmojaki:
To start with I've just stuck in `TerminalFormatter()`. @Carreau How do I insert the correct formatter and style based on user configuration?

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -97,13 +97,13 @@ import time
 import traceback
 
 import stack_data
+from pygments.formatters.terminal import TerminalFormatter
 
 # IPython's own modules
 from IPython import get_ipython
 from IPython.core import debugger
 from IPython.core.display_trap import DisplayTrap
 from IPython.core.excolors import exception_colors
-from IPython.utils import PyColorize
 from IPython.utils import path as util_path
 from IPython.utils import py3compat
 from IPython.utils.terminal import get_terminal_size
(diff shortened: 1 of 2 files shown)


IPython/core/ultratb.py
```diff
@@ -152,23 +150,18 @@ def _format_traceback_lines(lines, Colors, lvals, _line_format):
...
-            line = new_line
-
+        line = stack_line.render(pygmented=True).rstrip('\n') + '\n'
         lineno = stack_line.lineno
         if stack_line.is_current:
             # This is the line with the error
             pad = numbers_width - len(str(lineno))
             num = '%s%s' % (debugger.make_arrow(pad), str(lineno))
-            line = '%s%s%s %s%s' % (Colors.linenoEm, num,
-                                    Colors.line, line, Colors.Normal)
```
alexmojaki: `Colors.line` (typically red) was meant to highlight the current line, but it was completely overridden by PyColorize highlighting every token. With pygments the syntax highlighting overrides `Colors.line` inconsistently, so we remove it. No one should notice a difference unless there's a way to have colors excluding syntax highlighting.

Carreau:
`TBTools` is and instance of `Colorable`, so it should have  a `default_style` that is hooked into the command light. 

To be honest it is a real mess; and `PyColorize.py` will also have some utils; There have been 2 or 3 ways of handling colors through the years; so if in a first step themes are not handled I think it's fine.

alexmojaki:
OK, so I've just looked at the config enough to know when there are no colors. Seems like this should be a mergeable PR now.

Carreau:
Thanks again ! That is awesome !

PR #12132, opened 2020-02-16
Use default shortcuts with terminal debugger

blueyed:
This uses `create_ipython_shortcuts` also for `TerminalPdb`, which then
allows for using e.g. `F2` there also etc.

`create_ipython_shortcuts` is used in
`TerminalInteractiveShell.init_prompt_toolkit_cli` otherwise only, also
passing them to `PromptSession` then only.
Therefore this could be done there (in `PromptSession`) by default maybe
even.


Carreau:
That looks reasonable ; thanks !

PR #12135, opened 2020-02-20
add configuration controlling whether the CWD gets added to sys.path

ngoldbaum:
As of IPython 7.2 (see #11502), the CWD comes after standard library paths in sys.path by default. This fixes local files shadowing the standard library, a common footgun (e.g. having `email.py` in the CWD). However in some cases it's desirable for reproducibility, [security](https://bugs.python.org/issue16202) and [usability](https://bugs.python.org/issue946373) purposes for all imports to either come from installed packages or the standard library.

This adds a new configuration option, `ignore_cwd`, that tells IPython not to include the current directory in sys.path.

I'm happy to bikeshed on the name, `ignore_cwd` was the best I could come up with but I'm not totally happy with it. It wasn't clear to me how to test this so I haven't included any new tests. Tips in that direction are very much appreciated.

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -60,6 +60,10 @@ addflag('color-info', 'InteractiveShell.color_info',
     colours.""",
     "Disable using colors for info related things."
 )
+addflag('ignore-cwd', 'InteractiveShellApp.ignore_cwd',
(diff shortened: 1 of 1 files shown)


Carreau:
Perfect !

PR #12137, opened 2020-02-21
Fix inability to interrupt processes on Windows

### LINKED ISSUE #3400: Kernel/Interrupt Kernel does not terminate stuck subprocesses in the notebook
When a subprocess is run from the notebook, if it gets stuck the kernel will get locked waiting for it. Selecting Kernel/Interrupt from the menu does not terminate the subprocess, but rather leaves the kernel in an unstable, "partially locked" state, where other cells do not execute. The only resolution is to restart the kernel.

This occurred for me on Windows - I do not know if it also happens on Unix.

To demonstrate, start a notebook and enter !python in a cell. The process will lock as it is waiting for interactive input. As there is no way to provide that input, the kernel must be restarted to continue.

itamarst:
Fixes #3400 

(or at least fixes the Windows subprocesses variant; there's a bunch of stuff in there that is probably separate tickets—at least pdb was moved out into its own).

It appears that in addition to wait() being uninterruptible (https://bugs.python.org/issue28168), so is read()ing from the subprocess.

Reading is therefore pushed off to threads, and `wait()` replaced with the equivalent `poll()`-based loop. If the user interrupts the process, it is killed.

This is demonstrated by a unit test that previously failed, and now passes. Still remaining for me to do is actually test this fix on Windows from Jupyter; I'm waiting for a VM to unpack. If a reviewer wants to try this out themselves, by all means do so :)

### DIFF
--- a/IPython/utils/_process_win32.py
+++ b/IPython/utils/_process_win32.py
@@ -21,7 +21,7 @@ import ctypes
 
 from ctypes import c_int, POINTER
 from ctypes.wintypes import LPCWSTR, HLOCAL
-from subprocess import STDOUT
+from subprocess import STDOUT, TimeoutExpired
 
 # our own imports
 from ._process_common import read_no_interrupt, process_handler, arg_split as py_arg_split
@@ -100,8 +100,14 @@ def _system_body(p):
         line = line.decode(enc, 'replace')
         print(line, file=sys.stderr)
 
-    # Wait to finish for returncode
-    return p.wait()
+    # Wait to finish for returncode. Unfortunately, Python has a bug where
+    # wait() isn't interruptible (https://bugs.python.org/issue28168) so poll in
+    # a loop instead of just doing `return p.wait()`.
+    while True:
+        try:
+            return p.wait(0.01)
+        except TimeoutExpired:
+            pass
 
 
 def system(cmd):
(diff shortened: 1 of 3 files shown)


itamarst:
Alright! Looks like I now have a process interrupting test that simulates how ipykernel handles interrupts, and it demonstrates that `system()`-run proceses:

1. Exit quickly on Linux in response to the interrupt.
2. Ignore the interrupt on Windows.

Next I will write a similar test for `getoutput()`, and then I will try to see if my hypothetical fix works.

itamarst:
It appears the subprocess becomes uninterruptible while reading, rather than during `wait()`.

itamarst:
I am not sure why Travis is failing, the failures seem unrelated.

itamarst:
Investigating the Travis failure.

Carreau:
Well you seem to know what you are doing and have a windows machine. I can't test, but the code seem reasonable. Thanks !

amrita112:
Should this be working in IPython version 7.15.0? I couldn't get it to work. 
I ran the following in a jupyter notebook cell:
input()
and then tried to interrupt the kernel, but couldn't.

itamarst:
@amrita112 you also need the newest IPykernel.

PR #12141, opened 2020-02-25
Run ipdb in separate thread (required when using within asyncio applications).

jonathanslenders:


Carreau:
Merging base on comments from other issue.

Thanks @jonathanslenders

Issue #12151, opened 2020-02-27
Sort out #12150 colors based on configuration.


Carreau:

alexmojaki:
Add to 8.0 milestone?

PR #12165, opened 2020-02-29
Add configurability to executing nodes.

Carreau:
A bit hacky but should help with #12151

### DIFF
--- a/IPython/core/excolors.py
+++ b/IPython/core/excolors.py
@@ -69,6 +69,7 @@ def exception_colors():
         line = C.NoColor,
         caret = C.NoColor,
         Normal = C.NoColor
+        ExecutingNode = '',
         ))
 
     # make some schemes as instances so we can copy them for modification easily
@@ -97,7 +98,7 @@ def exception_colors():
         line = C.Yellow,
         caret = C.White,
         Normal = C.Normal,
-        ExecutingNode = 'bg:#00005f'
+        ExecutingNode = 'bg:#00005f',
         ))
 
     # For light backgrounds, swap dark/light colors
@@ -127,7 +128,7 @@ def exception_colors():
         line = C.Red,
         caret = C.Normal,
         Normal = C.Normal,
-        ExecutingNode = 'bg:#005f00'
+        ExecutingNode = 'bg:ansicyan',
         ))
 
     ex_colors.add_scheme(ColorScheme(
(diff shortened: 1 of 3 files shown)


Carreau:
cc @alexmojaki

IPython/core/ultratb.py
```diff
@@ -716,7 +716,7 @@ def get_records(self, etb, number_of_lines_of_context, tb_offset):
         before = context - after
         if self.has_colors:
             style = get_style_by_name('default')
-            style = stack_data.style_with_executing_node(style, 'bg:#00005f')
+            style = stack_data.style_with_executing_node(style, self.color_scheme_table.active_colors.ExecutingNode)
```
alexmojaki:
```suggestion
            style = stack_data.style_with_executing_node(style, self.Colors.ExecutingNode)
```

IPython/core/excolors.py
```diff
@@ -96,7 +96,8 @@ def exception_colors():
         excName = C.LightRed,
         line = C.Yellow,
         caret = C.White,
-        Normal = C.Normal
+        Normal = C.Normal,
+        ExecutingNode = 'bg:#00005f'
```
alexmojaki:
```suggestion
        ExecutingNode = 'bg:#00005f',
```

IPython/core/excolors.py
```diff
@@ -126,6 +127,7 @@ def exception_colors():
         line = C.Red,
         caret = C.Normal,
         Normal = C.Normal,
+        ExecutingNode = 'bg:#005f00'
```
alexmojaki:
```suggestion
        ExecutingNode = 'bg:#005f00',
```

Keep diffs minimal :)

IPython/core/excolors.py
```diff
@@ -154,6 +156,7 @@ def exception_colors():
         line = C.Red,
         caret = C.Normal,
         Normal = C.Normal,
+        ExecutingNode = 'bg:#dddddd'
```
alexmojaki:
```suggestion
        ExecutingNode = 'bg:#dddddd',
```

alexmojaki (changes requested):

alexmojaki:
Seems to me that it would be easier to first have pygments styles sorted for each option and then we could see what the executing colors look like behind syntax highlighted code.

Carreau:
> Seems to me that it would be easier to first have pygments styles sorted for each option and then we could see what the executing colors look like behind syntax highlighted code.

Yes the whole color think is a mess. I was just hoping to use this as a workaround and slowly progress toward a clean solution.

IPython/core/excolors.py
```diff
@@ -69,6 +69,7 @@ def exception_colors():
         line = C.NoColor,
         caret = C.NoColor,
         Normal = C.NoColor
```
Carreau, line 71:
```suggestion
        Normal = C.NoColor,
```

PR #12168, opened 2020-03-02
Make IPython.core.debugger interruptible by KeyboardInterrupt

### LINKED ISSUE #10516: Hang after running pdb in a cell, Kernel/Interrupt doesn't help
If you debug a cell via pdb, then don't quit it before rerunning the cell, you can't get back to your IPython environment.  Reproduce it via [Test case for pdb hang in Jupyter notebook](https://gist.github.com/nealmcb/1f0a14d63d539940e8dfccc29d6d69ae). Here's a screenshot:

pdb_resists_kernel_interrupt ipynb

For the record and search engines, to hang your notebook:
* Run the following cell
* Do not "quit" via the pdb prompt
* Then run it again.

```
def test():
    import pdb; pdb.set_trace()  # XXX BREAKPOINT
    return 0

test()
```

Note that Kernel/Interupt doesn't help. You can save the notebook, but to run cells again, you need to restart the kernel, at which point "All variables will be lost"

This happens to me not infrequently if I am searching thru the rest of the notebook for hints on how to debug the cell, then forget to quit before changing the cell and running it again.  Yes, I can be more careful, but others have the same experience.  Thanks to @wmayner for a convenient test case described at #3400.

I'm running Jupyter notebook 4.3.1 on Python 3.4.3 (default, Nov 17 2016, 01:08:31) [GCC 4.8.4]
Kernel Information: Python 3.4.3 (default, Nov 17 2016, 01:08:31)

Input from @takluyver at #10499 
> pdb runs in the same process, and we do know when it's expecting input. I think it needs someone to do the work so that when the HTML input box is removed in the frontend, a message is sent back to the kernel which will raise EOFError.
> That may need a new message type, or we could do it as a metadata field on the existing stdin reply message.

itamarst:
Together with a number of other PRs, this is a step towards addressing #10516 

There are three issues:

1. `IPython.core.debugger` would swallow KeyboardInterrupts and not let them get raised. This fixes that issue in a new class, `InterruptiblePdb`, so as not to change existing behavior.
2. `pdb` had the same issue, and by default all debuggers in Jupyter kernel usage need to be interruptible by KeyboardInterrupt to fix this bug. https://github.com/ipython/ipykernel/pull/490 overrides pdb and the default debugger to use the new `InterruptiblePdb`.
3. Finally, ipykernel's mechanism for interrupting subprocesses doesn't raise KeyboardInterrupt on Windows when waiting for `input()`. Fixed by https://github.com/ipython/ipykernel/pull/498

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -290,7 +290,7 @@ class Pdb(OldPdb):
         try:
             OldPdb.interaction(self, frame, traceback)
         except KeyboardInterrupt:
-            self.stdout.write('\n' + self.shell.get_exception_only())
+            raise
 
     def new_do_up(self, arg):
         OldPdb.do_up(self, arg)
@@ -627,6 +627,19 @@ class Pdb(OldPdb):
 
     do_w = do_where
 
+    def _cmdloop(self):
+        # Variant that doesn't catch KeyboardInterrupts.
+        while True:
+            try:
+                # keyboard interrupts allow for an easy way to cancel
+                # the current command, so allow them during interactive input
+                self.allow_kbdint = True
+                self.cmdloop()
+                self.allow_kbdint = False
+                break
+            except KeyboardInterrupt:
+                raise
+
 
 def set_trace(frame=None):
     """
@@ -635,3 +648,10 @@ def set_trace(frame=None):
     If frame is not specified, debugging starts from caller's frame.
     """
     Pdb().set_trace(frame or sys._getframe().f_back)
+
+
+# Override built-in Pdb, since that version doesn't allow interrupting:
+import pdb
+pdb.set_trace = set_trace
+pdb.Pdb = Pdb
+del pdb
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -290,7 +290,7 @@ class Pdb(OldPdb):
         try:
             OldPdb.interaction(self, frame, traceback)
         except KeyboardInterrupt:
-            raise
+            self.stdout.write('\n' + self.shell.get_exception_only())
 
     def new_do_up(self, arg):
         OldPdb.do_up(self, arg)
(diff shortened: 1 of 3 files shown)


Carreau:
Thanks, marking as 7.14.

There seem to be some unexpected interaction as IPython tests are failing, but they should not depend on PyZMQ. I'll have to dig into this.

itamarst:
I can take a look too, just, step 1 is convincing pyzmq maintainer to accept that PR (or convincing myself that PR is not necessary).

Carreau:
if that's @minrk then he can also merge the fixes here if necessary and likely do synchronized releases.

itamarst:
Ah, great, just need to convince him it's necessary :)

itamarst:
Looks like I was wrong and pyzmq is fine, so just down to these two PRs.

itamarst:
After further research, I have determined that the interrupt mechanism used by ipykernel on Windows doesn't interrupt pdb (or rather, I suspect, `input()`). Some research brought up this epic comment (https://stackoverflow.com/a/35792192) which explains how to make a signal handler for Windows that actually works, and I reproduced it in the test.

Unfortunately this will require some corresponding changes to ipykernel, and still has some... oddities. Like, I need to send the signal multiple times (perhaps there's a race condition? A flag that needs to be set?). Python 3.6 isn't working in AppVeyor so that suggests more issues. If I can just get it working on e.g. POSIX plus Python 3.8 Windows that will still be an improvement, so I will try to do that next.

itamarst:
I switched to a simpler test that doesn't involve subprocesses. I will next try to write a process-interrupting test in an ipykernel branch, which is a better place to do it since it is the one that actually implements that logic, and see if I can fix the Windows issue I encountered there.

itamarst:
@Carreau OK I think this is ... plausibly maybe even correct and working now :grin:

Carreau:
OK, let's try.

tonyfast:
Anyone have an idea as to while my `pdb.Pdb` class is being updated to `IPython.core.debugger.InterruptiblePdb`? I can't figure out where in the debugger phases this class is being forced to change. It is causing issues with running `doctest` in a shell session.

image

itamarst:
@tonyfast It happens in `ipykernel`.

tonyfast:
Is there an option or flag to not do this? 

I don't want to use pdb and it is disrupting a testing pattern in the stdlib module which is harshing my buzz.

tonyfast:
Could this be something that the `InteractiveShell` handles with `traitlets` when `shell.pdb is True`?

PR #12171, opened 2020-03-05
support using '%run -m' for modules importable via an import hook

### LINKED ISSUE #11736: %run does not work with import hooks
Hi,

I ran into an issue where the use of `imp` vs `__import__` matters. `imp` is bypassing some things that import statements do, such as calling custom [import hooks](https://docs.python.org/3/reference/import.html#import-hooks). Is there are reason `imp` is being used? If not, can this be updated to use `__import__`?

As a note, this comes up when using `%run -m foo`.

Thanks!

ngoldbaum:
Fixes #11736 

Current `%run -m` can't be used with modules that are loadable via an import hook. This makes that possible. I've also added a test to confirm that this works as expected, using the test script provided by @mlucool in the issue.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -692,6 +692,11 @@ class ExecutionMagics(Magics):
                 warn('For Windows, use double quotes to wrap a filename: %run "mypath\\myfile.py"')
             error(msg)
             return
+        except TypeError:
+            if fpath in sys.meta_path:
+                filename = ""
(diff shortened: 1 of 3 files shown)


Carreau:
Test are due to a regression in mpl, [fixed there](https://github.com/matplotlib/matplotlib/pull/16709)

Carreau:
Thanks ! I rebased and force push to fix CI.

PR #12177, opened 2020-03-09
Newer matplotlib versions appear not to support file-like object

### LINKED ISSUE #12176: New matplotlib (3.2.0) breaks the test suite
Trying to reproduce matplotlib-y errors in #12168, I did `pip install --upgrade matplotlib`, and now I'm getting the same errors, e.g.:

```python
ERROR: Test color settings for latex_to_png.
  File "/home/itamarst/Devel/ipython/venv/lib64/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/itamarst/Devel/ipython/IPython/testing/decorators.py", line 212, in skipper_func
    return f(*args, **kwargs)
  File "/home/itamarst/Devel/ipython/IPython/lib/tests/test_latextools.py", line 144, in test_latex_to_png_color
    default_value = latextools.latex_to_png(latex_string, wrap=False)
  File "/home/itamarst/Devel/ipython/IPython/lib/latextools.py", line 104, in latex_to_png
    bin_data = f(s, wrap, color, scale)
  File "/home/itamarst/Devel/ipython/IPython/lib/latextools.py", line 126, in latex_to_png_mpl
    mt.to_png(f, s, fontsize=12, dpi=dpi, color=color)
  File "/home/itamarst/Devel/ipython/venv/lib64/python3.7/site-packages/matplotlib/mathtext.py", line 3436, in to_png
    with open(filename, "wb") as file:
TypeError: expected str, bytes or os.PathLike object, not _io.BytesIO
```

itamarst:
Fixes #12176.

### DIFF
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -4,7 +4,7 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from io import BytesIO, open
+from io import open
 import os
 import tempfile
 import shutil
@@ -121,10 +121,10 @@ def latex_to_png_mpl(s, wrap, color='Black', scale=1.0):
 
     try:
         mt = mathtext.MathTextParser('bitmap')
-        f = BytesIO()
+        f = tempfile.NamedTemporaryFile("wb+")
         dpi = 120*scale
-        mt.to_png(f, s, fontsize=12, dpi=dpi, color=color)
-        return f.getvalue()
+        mt.to_png(f.name, s, fontsize=12, dpi=dpi, color=color)
+        return f.read()
     except (ValueError, RuntimeError, ParseFatalException):
         return None
 


itamarst:
I see from comment from @Carreau that this was fixed upstream in matplotlib, so may or may not be worth merging this.

tacaswell (changes requested):
I think it is better to exclude the broken Matplotlib and get this fixed upstream (we expect to have a 3.2.1 soon).

Carreau:
Yeah, I prefer to wait for matplotlib to make a new release as this is already fixed in their 3.2.x branch. I will also likey not do an IPython release this month for personal reasons, so next release is going to be end of april. 

We can though do as @tacaswell suggest and pin matplotlib, as well as hotfix conda forge.

itamarst:
OK, will close.

PR #12183, opened 2020-03-10
Fixed objs in the display.py docstrings.

### LINKED ISSUE #12182: display() docstring provides wrong typehints
``` python
from IPython.display import display, Markdown
# line below generates warning "Expected type 'tuple', got 'Markdown' instead"
display(Markdown('test'))
```
This issue is due to the docstring for the method display it is:

```python
def display(*objs, include=None, exclude=None, metadata=None, transient=None, display_id=None, **kwargs):
    """ 
   ... Removed some docstring ... 
   Parameters
    ----------
    objs : tuple of objects
        The Python objects to display.
```

rather than 

```python
def display(*objs, include=None, exclude=None, metadata=None, transient=None, display_id=None, **kwargs):
    """ 
   ... Removed some docstring ... 
   Parameters
    ----------
    *objs : object
        The Python objects to display.
```

While [numpydoc](https://numpydoc.readthedocs.io/) is silent on the issue the [napoleon example of numpy style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) confirms that the latter is indeed the better option. Additionally the latter variant works with IntelliSense while the former does not.

Luttik:
Improves the docstring for `display.py` while also resolving warnings that are generated by IntelliSense due to the [docstring issue](https://github.com/ipython/ipython/issues/12182).

I changed the docstring for star args (i.e. `*objs`) from `objs : tuple of objects` to `*objs : object`. While the [numpydoc documentation](https://numpydoc.readthedocs.io/en/latest/) is silent on how to handle star args other documentation like [napoleon example of NumPy style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) is in line with the changes that I suggest. 

Therefore, this change improves readability (due to being in line with expected practices) as well as providing the proper IntelliSense suggestions.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -57,7 +57,7 @@ def _display_mimetype(mimetype, objs, raw=False, metadata=None):
     ----------
     mimetype : str
         The mimetype to be published (e.g. 'image/png')
-    objs : tuple of objects
+    *objs : object
         The Python objects to display, or if raw=True raw text data to
         display.
     raw : bool
@@ -139,7 +139,7 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
 
     Parameters
     ----------
-    objs : tuple of objects
+    *objs : object
         The Python objects to display.
     raw : bool, optional
         Are the objects to be displayed already mimetype-keyed dicts of raw display data,
@@ -398,7 +398,7 @@ def display_pretty(*objs, **kwargs):
 
     Parameters
     ----------
-    objs : tuple of objects
+    *objs : object
         The Python objects to display, or if raw=True raw text data to
         display.
     raw : bool
@@ -418,7 +418,7 @@ def display_html(*objs, **kwargs):
 
     Parameters
     ----------
-    objs : tuple of objects
+    *objs : object
         The Python objects to display, or if raw=True raw HTML data to
         display.
     raw : bool
@@ -435,7 +435,7 @@ def display_markdown(*objs, **kwargs):
 
     Parameters
     ----------
-    objs : tuple of objects
+    *objs : object
         The Python objects to display, or if raw=True raw markdown data to
         display.
     raw : bool
@@ -453,7 +453,7 @@ def display_svg(*objs, **kwargs):
 
     Parameters
     ----------
-    objs : tuple of objects
+    *objs : object
         The Python objects to display, or if raw=True raw svg data to
         display.
     raw : bool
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks !

CI failure is likely unrelated, and is waiting for another matplotlib patch release. I'll check if there are any other issues and merge when I can.

Luttik:
@Carreau I would be willing to fix more docstring issues like this. (maybe even start an effort to add type hints everywhere, I believe everything below 3.6 is now no longer supported so typing (available from 3.5) should work now). Should I do one big pull request for that or do you have a better suggestion (that is somewhat kind towards my time aswell).

Carreau:
> Should I do one big pull request for that or do you have a better suggestion (that is somewhat kind towards my time as well).

I would prefer 1 file at a time, large PRs tend to create conflicts down the road. 
If you know of any tools that could lint (or even better auto-fix) docstring like this, that would be great. Even a PR for a single function is fine. My time is limited as well, so the smallest the chunks the more chance I have to merge it. 

And yes I believe we are 36+ only now, so we can also go with type annotation. I'm not sure how much that is prefered/working with napolean/numpydoc.

Luttik:
@Carreau Could this change also be merged into a release branch (notably 7.x). It has been a year since this was merged but is still not in a release branch.

This is a non-breaking change that improves interpretability and fixes false-positives in linters... No reason to keep it out IMHO.

MrMino:
@Carreau looks like it docstrings-only stuff - if you have no objections I'm going to add this to 7.x on Monday, ok?

MrMino:
@MeeseeksDev backport to 7.x

MrMino:
@Luttik I just backported this to `7.x`.

Luttik:
Thank you!

Carreau:
> Edit: 😄 not sure what I did wrong here, gonna do it manually then

Not sure. Might not have been you:

> requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))

@meeseeksdev say hi

Carreau:
> pr_origin_org_repo = pr["head"]["repo"]["full_name"]
> TypeError: 'NoneType' object is not subscriptable

Something fishy...

MrMino:
@Carreau Maybe it's because the source repository (the fork) isn't there anymore?

PR #12193, opened 2020-03-17
Add ಠ_ಠ as a \hmm latex completion.

Carreau:
See https://twitter.com/jakevdp/status/1239979963920445441 for background.

-- 

hmm

### DIFF
--- a/IPython/core/latex_symbols.py
+++ b/IPython/core/latex_symbols.py
@@ -12,6 +12,7 @@
 latex_symbols = {
 
     "\\euler" : "ℯ",
+    "\\ohm" : "Ω",
     "\\^a" : "ᵃ",
     "\\^b" : "ᵇ",
     "\\^c" : "ᶜ",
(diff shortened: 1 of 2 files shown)


Carreau:
Note that  the `ಠ` can be obtain with `\KANNADA LETTER TTHA<tab>` already, as many other valid letters.

djkalmyk:
merge it, it's beautiful!

Carreau:
I need to document it, test it, and it may end up its own completer provide instead of hacking the latex completer.

IPython/core/latex_symbols.py
```diff
@@ -1295,6 +1296,7 @@
     "\\ttnine" : "𝟿",
     "\\underbar" : "̲",
     "\\underleftrightarrow" : "͍",
+    "\hmm" : "ಠ_ಠ",
```
ferrine, line 1299: Need to be escaped
```suggestion
    "\\hmm" : "ಠ_ಠ",
```

ferrine: Is it?

Carreau: Yes thanks !

PR #12207, opened 2020-03-26
Bump jedi to at least 0.16.0 and fix deprecated function usage

### LINKED ISSUE #12102: Deprecation warning from Jedi
xref https://github.com/pandas-dev/pandas/pull/31323 and https://github.com/pandas-dev/pandas/issues/31342

It looks like there is a deprecation warning in Ipython once starting to use the latest jedi version (0.16.0)

aiguofer:
This should fix #12102

I ran `iptest IPython.core.tests.test_completer` and it didn't seem to give any errors. Not sure if we want to add any new tests here, or if that is good enough.

According to jedi, `Multiple signatures are typical if you use Python stubs with @overload.` I made my change to account for all signatures, but not sure how this plays out IRL since I've never seen a function with multiple signatures.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -988,7 +988,8 @@ def _make_signature(completion)-> str:
 
     """
 
-    return '(%s)'% ', '.join([f for f in (_formatparamchildren(p) for p in completion.params) if f])
+    return '(%s)'% ', '.join([f for f in (_formatparamchildren(p) for signature in completion.get_signatures()
+                                          for p in signature.defined_names()) if f])
 
 class IPCompleter(Completer):
     """Extension of the completer class with IPython-specific features"""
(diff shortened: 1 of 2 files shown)


Carreau:
That sounds good ! Thanks !

stonebig:
would forcing support of jedi>=0.15.2 have been sufficient ? we're a bit stuck during the Spyder transition, as some fixes will only be in Jedi-0.17.1

Carreau:
@meeseeksbot backport to 7.x

Carreau:
> would forcing support of jedi>=0.15.2 have been sufficient ? we're a bit stuck during the Spyder transition, as some fixes will only be in Jedi-0.17.1

Sorry I missed this question; I'll see in the backport if I can pin to only 0.15.2, but it's less than a month difference, so it won't change much I think. And so far this is only planned to be released in 8.0 which is not before fall I think.

Carreau:
@meeseeksdev backport to 7.x

Carreau:
> I'll see in the backport if I can pin to only 0.15.2,

and no .complete() is only 0.16.0+

PR #12208, opened 2020-03-27
support for unicode identifiers in `?`, `??`

mwageringel:
This rewrites some of the regular expressions that are used to match Python identifiers, so that they are unicode compatible. In Python 3, identifiers can contain unicode characters as long as the first character is not numeric.

Examples for the changes:

• inputtransformer:

```
In [1]: π = 3.14
In [2]: π.is_integer?
Object `is_integer` not found.
```

----------
• namespace:

```
π.is_integ*?
```
or
```
In [1]: %psearch π.is_integ
Python identifiers can only contain ascii characters.
```

----------
• prefilter:

```
%autocall 1
φ = float
get_ipython().prefilter("φ 3")  # should be 'φ(3)', but returns 'φ 3'
```

----------
• completerlib:

If there is a file e.g. named `π.py` in the current directory, then

```
import IPython
IPython.core.completerlib.module_list('.')  # should contain module 'π'
```

### DIFF
--- a/IPython/core/completerlib.py
+++ b/IPython/core/completerlib.py
@@ -52,7 +52,7 @@ TIMEOUT_STORAGE = 2
 TIMEOUT_GIVEUP = 20
 
 # Regular expression for the python import statement
-import_re = re.compile(r'(?P<name>[a-zA-Z_][a-zA-Z0-9_]*?)'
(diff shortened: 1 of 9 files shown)


Carreau:
Oh ! Nice catch ! Thanks !

PR #12212, opened 2020-03-30
Add option to change HTML attributes in display.Video

mancellin:
I want to display video in a Jupyter notebook, and I would like to customize the behaviour of the video by changing some of the attributes of the HTML5 element (autoplay, loop, muted, ... see e.g. [1]) 

[1] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

This PR allows this by adding an option to the Video constructor:

    Video("foo.mp4", html_attributes="controls loop")

Another solution would have been to add boolean options for each properties

    Video("foo.mp4", loop=True, autoplay=False)

but that might be too cumbersome to maintain for such a specialized feature.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1348,17 +1348,20 @@ class Video(DisplayObject):
             If not supplied, defaults to the height of the video.
         html_attributes : str
             Attributes for the HTML `<video>` block.
-            Default: "controls" to get video controls.
-            Other examples: "controls muted" for muted video with controls,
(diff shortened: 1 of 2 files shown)


ivanov (approved):
This looks great, @mancellin! I like how you've already added the new functionality to the examples AND spruced up the formatting of the doc string for the rendered API docs ! :cake:

Carreau:
Thanks ! That looks great, and thanks @ivanov for the first review. 

Merging, that should be out by end of month !

PR #12219, opened 2020-04-03
Catch ValueError

impact27:
Typing "w = 1" in the debugger crashes the debugger

example:
```
In[1]: %debug print()
ipdb> w = 1
```

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -620,7 +620,11 @@ class Pdb(OldPdb):
         Take a number as argument as an (optional) number of context line to
         print"""
         if arg:
-            context = int(arg)
+            try:
+                context = int(arg)
+            except ValueError:
+                self.stdout.write('\n' + self.shell.get_exception_only())
+                return
             self.print_stack_trace(context)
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, I've rebased on master which should fix the CI tests.

IPython/core/debugger.py
```diff
@@ -620,7 +620,11 @@ def do_where(self, arg):
         Take a number as argument as an (optional) number of context line to
         print"""
         if arg:
-            context = int(arg)
+            try:
+                context = int(arg)
+            except ValueError:
+                self.stdout.write('\n' + self.shell.get_exception_only())
```
eric-wieser, line 626: Should this be `self.error(err)`, looking at lines above?

eric-wieser: Defined as https://github.com/python/cpython/blob/c606624af8d4cb3b4a052fb263bb983b3f87585b/Lib/pdb.py#L459

impact27: I changed it

Issue #2137, opened 2012-07-14
Add a utility method to pretty printer for common Class(a, b, c=d) style output


njsmith:
It'd be nice if common boring _repr_pretty_ methods could just be written like:

```
def _repr_pretty_(self, p, cycle):
    assert not cycle
    p.pretty_constructor(self, [self.a, self.b], [("c", self.c)])
```

Something like this (extracted from the patsy source): https://gist.github.com/3113186

eric-wieser:
Now that 3.6 is the minimum supported version, I wonder if we can simplify that spelling to
```python
p.pretty_invocation(type(self).__qualname__, self.a, self.b, c=self.c)
```

PR #12230, opened 2020-04-10
Add pretty-printing for types.SimpleNamespace

eric-wieser:
Eg:
```python
In[1]: import sys

In[2]: sys.implementation
namespace(name='cpython',
          cache_tag='cpython-38',
          version=sys.version_info(major=3, minor=8, micro=0, releaselevel='final', serial=0),
          hexversion=50856176)
```

If this looks good, in a future patch I might look at supporting:
* Namedtuples
* `sys.version_info`-like objects created with `PyStructSequence_InitType2`

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -648,6 +648,22 @@ def _re_pattern_pprint(obj, p, cycle):
     p.text(')')
 
 
+def _types_simplenamespace_pprint(obj, p, cycle):
+    """The pprint function for types.SimpleNamespace."""
+    name = 'namespace'
+    with p.group(len(name) + 1, name + '(', ')'):
+        if cycle:
+            p.text('...')
+        else:
+            for idx, (attr, value) in enumerate(obj.__dict__.items()):
+                if idx:
+                    p.text(',')
+                    p.breakable()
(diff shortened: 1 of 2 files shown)


IPython/lib/pretty.py
```diff
@@ -648,6 +648,22 @@ def _re_pattern_pprint(obj, p, cycle):
     p.text(')')
 
 
+def _types_simplenamespace_pprint(obj, p, cycle):
+    """The pprint function for regular expression patterns."""
```
eric-wieser, line 652:
```suggestion
    """The pprint function for types.SimpleNamespace."""
```

eric-wieser:
Ci failures look unrelated

Carreau:
Thanks that looks good, I've maked at to be backported on 7.x, and have rebased on master to fix the CI failures that appear to be due to the new version fo sphinx.

eric-wieser:
#2137 might be worth looking at again as a follow up, since it basically matches this implementation.

Carreau:
> #2137 might be worth looking at again as a follow up, since it basically matches this implementation.

Probably. That sound like a nice project for someone trying to get started. I've tagged it appropriately. Let's get that in for now.

Issue #12220, opened 2020-04-04
bug in core.display.DisplayObject.reload when filename is None and url is not None


arabidopsis:
Both SVG and JSON subclasses of DisplayObject have
`@data.setter` methods that transform the string data into SVG or JSON but
the `DisplayObject.reload` method both sets and reads `self.data` when downloading
a url  (and probably fails silently ... there is a catchall try block).

## Fix:
change line 659 of display.py from:
`self.data = response.read()`
to:
`data = response.read()`

and remove lines 668 and 669
```
 if encoding:
       self.data = self.data.decode(encoding, 'replace')
```
and replace with:
```
if encoding:
      self.data = data.decode(encoding, 'replace')
else:
      self.data = data
```

Carreau:
Do you want to send a Pull Request ?

arabidopsis:
Sure:  https://github.com/ipython/ipython/pull/12232

PR #12232, opened 2020-04-12
ensure SVG(url=...) is set correctly

arabidopsis:
I removed the try/except block in DisplayObject too since,  if you misspecify a filename you get a
FileNotFoundException I think the user should get a URLError or similar if they misspecify
a URL. I was quite mystified when the SVG(url...) seemed to "work" but nothing was displayed
in the notebook.

> 

> Also added gzip decompression for .svgz files

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -752,6 +752,11 @@ class Latex(TextDisplayObject):
 
 
 class SVG(DisplayObject):
+    """Embed an SVG into the display.
+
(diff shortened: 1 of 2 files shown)


arabidopsis:
Fixes https://github.com/ipython/ipython/issues/12220

Carreau:
Thanks ! That looks good to me, CI failing is unrelated and I rebased on master to fix it (hopefully).

PR #12235, opened 2020-04-14
Fix compatibility with the latest release of Sphinx

frenzymadness:
`l_` from sphinx.locale has been deprecated for a long time.
`_` is the new name for the same function but it seems that the
imports there are useless.

https://github.com/sphinx-doc/sphinx/commit/8d653a406dc0dc6c2632176ab4757ca15474b10f

Let's see what CI thinks about this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,11 +69,6 @@ matrix:
       python: "3.7"
       dist: xenial
       sudo: true
-    - arch: arm64
-      python: "3.7"
-      dist: xenial
-      env: ARM64=True IPYTHON_TESTING_TIMEOUT_SCALE=2
-      sudo: true
     - arch: amd64
       python: "3.8"
       dist: xenial
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -121,6 +121,29 @@ numpydoc_show_class_members = False # Otherwise Sphinx emits thousands of warnin
 numpydoc_class_members_toctree = False
 warning_is_error = True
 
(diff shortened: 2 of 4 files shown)


frenzymadness:
I am not sure but it seems that the CI problem isn't related to this change. Am I right?

Carreau:
> I am not sure but it seems that the CI problem isn't related to this change. Am I right?

It seem to be due to the fact that Sphinx 3 now seem to treat duplicate items as error. 
As we do autogenerate our API documentation, it sees duplicated between subclasses, and building the docs ends up being an error.

Carreau:
I've pushed commit that _should_ fix the CI build, woudl you mind having a look to make sure that make sens and I did not miss anything ? 

Thanks !

frenzymadness:
All the changes you have made seem reasonable to me and CI is happy now. It might be good to check that there is no reference to `MyKernel` left in the documentation. Also, it might make sense to comment out arm architecture from Travis config so it can be simply uncommented in the future.

Carreau:
> It might be good to check that there is no reference to `MyKernel` left in the documentation

There is another MyKernel somewhere else, and sphinx was complaining about duplicates. 

> Also, it might make sense to comment out arm architecture from Travis config so it can be simply uncommented in the future.

It's easy enough to revert commits, and ARM has been troublesome enough. 

Thanks, merging.

Carreau:
@meeseeksdev backport to 7.x

Issue #12223, opened 2020-04-07
Move display classes (JavaScript, HTML...) under IPython.lib.display


martinRenou:
Following our discussion @Carreau.

In xeus-python, we monkey-patch the `IPython.core.display` module in order to plug the base display functions to the xeus display logic.

Our problem is that there are lots of classes in this module that we don't actually need to monkey patch:
`'DisplayObject', 'TextDisplayObject', 'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'ProgressBar', 'JSON', 'GeoJSON', 'Javascript', 'Image', 'set_matplotlib_formats',
'set_matplotlib_close' and 'Video'`. 

We only need to monkey-patch `'display', 'display_pretty', 'display_html', 'display_markdown', 'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json', 'display_javascript', 'display_pdf', 'clear_output', 'publish_display_data' and 'update_display'`

I am wondering if we could move all the classes that we don't need to monkey-patch from `IPython.core.display` to `IPython.lib.display`.

I can come up with a PR myself, I just need the green light from you. I guess this would be a backward-incompatible change (if people are importing from `IPython.core.display`, but not if they import from `IPython.display`).

Carreau:
I think that would be fine, we can change API in 8.0. 

we could also do it in 7.x with aliases the IPython.core.display that import from lib. 


Just to understand you do replace IPython.core.display completely so you want it to be as minimal as possible right ?

martinRenou:
> Just to understand you do replace IPython.core.display completely so you want it to be as minimal as possible right ?

Exactly. The only approach we found was to monkey-patch this module (this one and the Comms implementation for ipywidgets support). Simply because lots of Notebooks/libraries import the display function.

Carreau:
Marked as easy and help wanted. I have no objections, we can re expose for now  in core with a module `__getattribute__` or `__getattr__` to raise a deprecation warning.

martinRenou:
> we could also do it in 7.x with aliases the IPython.core.display that import from lib

I'm afraid this won't be possible (at least I suppose). Because `lib` already imports `core.display`.

ivanov:
Hi @martinRenou  and @Carreau ! Drive-by comment here, but I'm worried about how much stuff is being moved in #12239 that will unnecessarily break imports for folks unless imports continue to work from both .core.display and .lib.display under the proposed changes.

[...]

PR #12239, opened 2020-04-16
Refactor display modules

martinRenou:
This moves core display functions (for publishing output to the front-end) in a separate file.

Related issue #12223

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -17,15 +17,37 @@ from pathlib import Path, PurePath
 
 from IPython.utils.py3compat import cast_unicode
 from IPython.testing.skipdoctest import skip_doctest
-from .display_functions import display, clear_output, publish_display_data, update_display, DisplayHandle
+from . import display_functions
 
-__all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
+
+__all__ = ['display_pretty', 'display_html', 'display_markdown',
            'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json',
            'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
            'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'ProgressBar', 'JSON',
-           'GeoJSON', 'Javascript', 'Image', 'clear_output', 'set_matplotlib_formats',
-           'set_matplotlib_close', 'publish_display_data', 'update_display', 'DisplayHandle',
+           'GeoJSON', 'Javascript', 'Image', 'set_matplotlib_formats',
+           'set_matplotlib_close',
            'Video']
+
+_deprecated_names = ["display", "clear_output", "publish_display_data", "update_display", "DisplayHandle"]
+
+__all__ = __all__ + _deprecated_names
+
+
+# ----- warn to import from IPython.display -----
+
+from warnings import warn
+
+
+def __getattr__(name):
+    if name in _deprecated_names:
+        warn(f"Importing {name} from IPython.core.display is deprecated since IPython 7.14, please import from IPython display", DeprecationWarning, stacklevel=2)
+        return getattr(display_functions, name)
+
+    if name in globals().keys():
+        return globals()[name]
+    else:
+        raise AttributeError(f"module {__name__} has no attribute {name}")
+
 
 #-----------------------------------------------------------------------------
 # utility functions
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -5,12 +5,11 @@
 # Distributed under the terms of the Modified BSD License.
 
 
-from binascii import b2a_hex, b2a_base64, hexlify
+from binascii import b2a_base64, hexlify
 import json
 import mimetypes
 import os
 import struct
-import sys
 import warnings
 from copy import deepcopy
 from os.path import splitext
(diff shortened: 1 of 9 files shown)


martinRenou:
Not sure I get why the CI is failing here. It looks unrelated but I'm not sure.

martinRenou:
This PR will make quite a difference in xeus-python!
diff

SylvainCorlay:
:+1: this is great!

SylvainCorlay:
@Carreau would you be OK with this? It should not have any impact functionally in IPython, but would simplify the story quite a bit on the xeus side.

Carreau:
> Not sure I get why the CI is failing here. It looks unrelated but I'm not sure.

Not your fault, new version of sphinx.

> would you be OK with this? It should not have any impact functionally in IPython, but would simplify the story quite a bit on the xeus side.

Yes, let me play with Module __getattr__, and merge a similar but backward compatible with warnings changes in 7.x. If I merge this now I'm going to forget.

Carreau:
(also rebasing on master should fix CI)

Carreau:
Thoughts about https://github.com/martinRenou/ipython/pull/1 ?

martinRenou:
Thanks for your PR @Carreau. So do you prefer this approach or do you think it's nicer to move everything that is not "core" in lib?

I commented on the second approach in #12223.

Carreau:
That is good for me if it's a clearer distinction of what was moved.

Do you want to move more objects later like DisplayObject and Co ? We can do that in subsequent PRs if needs be.

And do you feel like the `__getattr__` dance is safe enough to backport on 7.x ?

martinRenou:
> Do you want to move more objects later like DisplayObject and Co ?

It should be just fine where it is.

> And do you feel like the __getattr__ dance is safe enough to backport on 7.x ?

I would say so. And it is definitely cleaner than what I had.

IPython/core/display.py
```diff
@@ -5,27 +5,49 @@
...
+
+
+# ----- warn to import from IPython.display -----
+
+from warnings import warn
+
+
+def __getattr__(name):
+    if name in _deprecated_names:
+        warn(f"Importing {name} from IPython.core.display is deprecated since IPython 7.14, please import from IPython display", DeprecationWarning, stacklevel=2)
```
ivanov, line 43: I don't have a strong opinion on this, but is it even worth deprecating the location of these, if the monkey-patching that xeus-python will do will just take care of everything transparently and we can keep everyone happy without having to have them change their code?

IPython/core/display.py
```diff
@@ -5,27 +5,49 @@
...
+    else:
+        raise AttributeError(f"module {__name__} has no attribute {name}")
 
-__all__ = ['display', 'display_pretty', 'display_html', 'display_markdown',
-'display_svg', 'display_png', 'display_jpeg', 'display_latex', 'display_json',
-'display_javascript', 'display_pdf', 'DisplayObject', 'TextDisplayObject',
-'Pretty', 'HTML', 'Markdown', 'Math', 'Latex', 'SVG', 'ProgressBar', 'JSON',
-'GeoJSON', 'Javascript', 'Image', 'clear_output', 'set_matplotlib_formats',
-'set_matplotlib_close', 'publish_display_data', 'update_display', 'DisplayHandle',
-'Video']
```
ivanov, line 28: I'd be fine with keeping this `__all__` as it was an not leaking to anyone who doesn't care too much that some of these got moved into `display_functions`. How will we get around not being able to notify someone doing `from IPython.core.display import *`?

IPython/display.py
```diff
@@ -1,16 +1,44 @@
...
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 # Imports
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 
-from IPython.core.display import *
```
ivanov, line 15: I think it should be fine to leave this one. The point of this approach was to keep old code like this from breaking.

IPython/testing/iptest.py
```diff
@@ -57,6 +57,7 @@
 warnings.filterwarnings('error', message='.*disable_gui.*', category=DeprecationWarning, module='.*')
 
 warnings.filterwarnings('error', message='.*ExceptionColors global is deprecated.*', category=DeprecationWarning, module='.*')
+warnings.filterwarnings('error', message='.*IPython.core.display.*', category=DeprecationWarning, module='.*')
```
ivanov, line 60: and then we don't have to add more filter warnings :)

ivanov (approved):
Thanks, @martinRenou - this looks great. I think it'd be even better without deprecating anything. What do you think?

IPython/core/display.py
Carreau: We may still remove them from here at some point, I would prefer from people to not import from core anyway.

IPython/core/display.py
Carreau: the issue with keeping this all is that `import *` now import deprecated aliases (which IPython.display was doing), so was basically silencing warnings.

Carreau: I do see your point about other doing `import *`, but it should be discourage anyway. I'll think about it.

IPython/display.py
Carreau: The problem is is was triggering the deprecation warning at import time, which then would be silenced for users, as the default is `warning:once`, hence why I'm explicit.

IPython/testing/iptest.py
Carreau: I want us to avoid internal imports,as if we do import a single time then we silence the warnings.

Carreau:
Thanks @ivanov for the review. most of the changes you commented on are to make sure warnings get triggers by users. I get your point for `import*` but I also don't want trigger the warnings for people that might `import *`  and not use deprecated function. 

Xeus may do monkey patching that works, but if we can have them not do it, it's even better, and they might not be the only alternative implementation.

martinRenou:
> Xeus may do monkey patching that works, but if we can have them not do it, it's even better, and they might not be the only alternative implementation.

It might be difficult, as of today, to find another solution than monkey-patching. Lots of Notebooks out there contain things like `from IPython.display import display, Audio`. That is the single reason why we are monkey-patching, to add support for this kind of Notebooks.

Carreau:
What if we can detect if we are running in xeus and then import Video and Co conditionally from you or some 3rd party packages ? I do have https://pypi.org/project/disp/, and https://pypi.org/project/display/ seem to be kinda abandonned.

martinRenou:
That was actually one idea we had. Which was to export `Video`, `Audio`.. from `core.display` and everything from `lib.display` into a separate display package that IPython and Xeus uses.

Carreau:
Ok, i'm going to merge that with master, which will be released with 8.0 I'm not sure when. IF you want it in IPython 7.x, let me know and we can backport to 7.x.

PR #12244, opened 2020-04-18
Add helper objects to help implement _repr_pretty_

### LINKED ISSUE #2137: Add a utility method to pretty printer for common Class(a, b, c=d) style output

eric-wieser:
Closes #2137, but using a slightly different approach. This converts internal `_repr_pretty_`s to use this tool as much as possible.

cc @njsmith.

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -34,6 +34,22 @@ pretty printer passed::
         def _repr_pretty_(self, p, cycle):
             ...
 
+Here's an example for a class with a simple constructor::
+
+    class MySimpleObject:
+
+        def __init__(self, a, b, *, c=None):
+            self.a = a
+            self.b = b
+            self.c = c
+
+        def _repr_pretty_(self, p, cycle):
+            ctor = CallExpression.factory(self.__class__.__name__)
+            if self.c is None:
+                p.pretty(ctor(a, b))
+            else:
+                p.pretty(ctor(a, b, c=c))
+
 Here is an example implementation of a `_repr_pretty_` method for a list
 subclass::
 
@@ -93,7 +109,7 @@ from IPython.utils.decorators import undoc
 from IPython.utils.py3compat import PYPY
 
 __all__ = ['pretty', 'pprint', 'PrettyPrinter', 'RepresentationPrinter',
-    'for_type', 'for_type_by_name']
+    'for_type', 'for_type_by_name', 'RawText', 'RawStringLiteral', 'CallExpression']
 
 
 MAX_SEQ_LENGTH = 1000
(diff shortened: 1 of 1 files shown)


IPython/lib/pretty.py
```diff
@@ -500,6 +500,55 @@ def remove(self, group):
             pass
 
 
+class CallExpression:
+    """ Object which emits a line-wrapped call expression in the form `__name(*args, **kwargs)` """
+    def __init__(__self, __name, *args, **kwargs):
+        # dunders are to avoid clashes with kwargs, as python's name manging
+        # will kick in.
+        self.name = __name
```
eric-wieser, line 508:
```suggestion
        self = __self
        self.name = __name
```

IPython/lib/pretty.py
```diff
@@ -500,6 +500,56 @@ def remove(self, group):
...
+            started = True
+
+        prefix = self.name + "("
+        with p.group(len(prefix), prefix, ")"):
+            for arg in self.args:
+                new_item()
+                p.pretty(arg)
+            for arg_name, arg in self.kwargs:
+                new_item()
+                arg_prefix = '{}='.format(attr)
```
eric-wieser, line 532:
```suggestion
                arg_prefix = arg_name + "="
```

IPython/lib/pretty.py
```diff
@@ -500,6 +516,75 @@ def remove(self, group):
...
+        self = __self
+        self.name = __name
+        self.args = args
+        self.kwargs = kwargs
+
+    @classmethod
+    def factory(cls, name):
+        def inner(*args, **kwargs):
+            return cls(name, *args, **kwargs)
+        return inner
```
Carreau, line 546: You need to be doing this only to curry when `name` right ? Would you consider always currying on `name` ? Like moving `self.args = args` and `self.kwargs = kwargs` into for example a `__call__` ?

IPython/lib/pretty.py
```diff
@@ -623,45 +708,38 @@ def _super_pprint(obj, p, cycle):
...
-    p.text(')')
+
+
+def _re_pattern_pprint(obj, p, cycle):
+    """The pprint function for regular expression patterns."""
+    re_compile = CallExpression.factory('re.compile')
+    if obj.flags:
+        p.pretty(re_compile(RawStringLiteral(obj.pattern), _ReFlags(obj.flags)))
+    else:
+        p.pretty(re_compile(RawStringLiteral(obj.pattern)))
```
Carreau, line 733: Or just wondering if the factory is _that_ needed to avoid a small duplication of code ?

```suggestion
def _re_pattern_pprint(obj, p, cycle):
    """The pprint function for regular expression patterns."""
    if obj.flags:
        p.pretty(CallExpression('re.compile', RawStringLiteral(obj.pattern), _ReFlags(obj.flags)))
    else:
        p.pretty(CallExpression('re.compile', RawStringLiteral(obj.pattern)))
```

?

Or am I missunderstanding ?

Carreau:
That sounds good to me, I'm happy to iterate in IPython master if you wish. 

I'm just wondering if the factory is _that_ necessary actually

eric-wieser:
The factory is a gimmick to make the syntax look more like the repr.

Happy to remove it if you don't like it.

IPython/lib/pretty.py
eric-wieser: That's a actually what I started with, but thought I'd play with the syntactic sugar

IPython/lib/pretty.py
eric-wieser: I don't really like the idea of the `__call__()` mutating the object, but if there's a functionally-pure way to require the currying (make the class name private?), I'm not opposed to that.

eric-wieser: One simple option would be:
```python
class _NamedCallExpression
    ... # as before

def CallExpression(name):
    # contents of `def factory`
```

Carreau:
> The factory is a gimmick to make the syntax look more like the repr.
> 
> Happy to remove it if you don't like it.

No that's fine. I'll mark this as 8.0 so we have some time to play with it and change it if we are unhappy with the naming/API

PR #12255, opened 2020-04-21
fix getting signatures in jedi 0.17

Carreau:


Carreau:
@davidhalter does that looks reasonable to you ? Or is there a cleaner way ?

davidhalter:
Looks reasonable, but I would handle the case of `get_signatures() == []` as well. This makes it easier in the future, so you don't have to depend on type (which was a bad idea in the first place - my fault).

IPython/core/completer.py
```diff
@@ -9,7 +9,7 @@
 Syntax like magics.
 
 Latex and Unicode completion
-============================
+====================:========
```
davidhalter, line 12: Probably not intentional :)

Otherwise, LGTM.

Carreau: Doh.. force-pushed.

Carreau:
actually no need to backport, 7.x still use `completions()` and not `complete()`

PR #12259, opened 2020-04-23
36plus cleanup

Carreau:
Remove all usage of u_format (no op), and remove function
u_format was a compatibility layer between Py2 and 3 , now useless.

Then, 

Apply pygrade --36plus on IPython/core/tests/test_inputtransformer.py.

Previous commit remove the u_format compatiblity function, use the
occasion to update the file to a newer python 3.6 syntax.

This also use yield-from


ivanov (approved):
```
u've got to be kidding me, good work!'
```

Carreau:
A clean codebase for a tranquil mind.

ivanov:
Thank you @Carreau !

PR #12261, opened 2020-04-23
Do not exit if self.interact is True

impact27:
If self.file_to_run is not a valid file, the kernel closes, but if the file is unreadable, it doesn't.
To be consistant with the other uses of exit, exit should only be called if `self.interact` is false.
See the two other occurences of exit above and below.


Carreau:
Looks reasonable to me, thanks !

PR #12271, opened 2020-05-01
wahtsnew 5.10

Carreau:


Carreau:
(test won't pass as travis list too old pythons)

PR #12277, opened 2020-05-02
start using .git-blame-ignore-revs

ivanov:
We merged a pyupgrade commit back in #12261, but I didn't realize @Carreau added machinery to help local `git blame` to ignore such stylistic change commits. So we might as well start doing that now. Thanks Matthias!


Carreau:
Thanks !

ivanov:
Oh, and since this came from #12091, I think @terrdavis deserves some credit, too!

terrdavis:
Hooray! I'm so happy to see this being used!

PR #12280, opened 2020-05-04
Unify API between completers.

### LINKED ISSUE #12279: completer latex_matches no respecting API
It return matched text even if it return no completions.

Carreau:
Forward latex seem to be the only one that return a text match  when no
completions is provided.

In particular this leads to custom logins in other places.

Should close #12279


ivanov (commented):
:+1:  I saw this but didn't have a chance to test it

Carreau:
Sorry it was annoying me for the rest of the work on the completer, so I merged it. Happy to get review on the other stuff though.

PR #12284, opened 2020-05-04
Try to elide long completion  based on user input.

Carreau:
If what the user has typed is already in the completion and the
completion is really long, try to elide what the user has already typed
from the displayed text of the completion.

Keep the first 3 and last 3 of what is already present.

This will behave weirdly if all the completion have a common prefix as I
believe prompt toolkit will insert the common prefix (or do we?). I'll
have to check how to consider the common prefix as being typed.


Carreau:
Example of usage for completing unicode:

Screen Shot 2020-05-04 at 16 35 39

PR #12293, opened 2020-05-07
try to track Resource leaks on windows

Carreau:


Carreau:
That seem to be on purpose for subprocesses, I'm not quite sure I want to touch this as it might have unexpected effects.

PR #12296, opened 2020-05-07
Delete 109-112 for removing external test

### LINKED ISSUE #12281: test_embed_svg_url needs internet connection
Hello.

[This part](https://github.com/ipython/ipython/blob/master/IPython/core/tests/test_display.py#L109) of the test_embed_svg_url needs internet connection to download the SVG image from Wiki. But some build systems have internet connection disabled on purpose so this will fail.

It would be nice to have some decorator to skip tests like this when no internet connection is available. What do you think about it?

Inception95:
#12281 

Delete the 109~112 lines in core/tests/test_display.py to remove the external test.


Carreau:
Thanks !

PR #12298, opened 2020-05-08
ENH: Add configuration for profiling completions.

ssanderson:
Adds two new configurable attributes to IPCompleter to enable profiling of the
completion subsystem.

- IPCompleter.profile_completions, a boolean indicating whether we should
  profile completions. Default is False.
- IPCompleter.profiler_output_dir, a string indicating the directory where
  profiling data should be written. Default is '.completion_profiles'.

Example usage: 
completions

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -121,6 +121,7 @@ import string
 import sys
 import time
 import unicodedata
+import uuid
 import warnings
 from contextlib import contextmanager
 from importlib import import_module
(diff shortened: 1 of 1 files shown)


Carreau:
No objection with that. Congratulation on your new position, I myself just started at Quansight last Friday,   and I have been pushing on completions (#12282). 

I'll have a look but as this may conflict with #12282, i'll have to see whcih one is easier to rebase.

ssanderson:
Thanks, and congrats to you as well! If you're comparing for rebasing I'd probably look at https://github.com/ipython/ipython/pull/12299 rather than this one. I opened them as separate PRs since they're more or less independent changes, but that PR currently depends on this one.

PR #12299, opened 2020-05-08
PERF: Make unicode completions lazier.

ssanderson:
This commit renames `IPCompleter._names` to ``_unicode_names`` and uses
traitlets default machinery to defer its initialization until first access.

Previously, initialization was deferred until the first call to
fwd_unicode_names, but that was still called on the first completion attempt,
which caused about a .5 second lag on the first completion attempt on
my (relatively powerful) machine.

This is based off of https://github.com/ipython/ipython/pull/12298. With profiling enabled at least, I'm seeing this as taking about half a second on the first completion attempt on each process on my machine. That's probably a bit of an overstatement, since the expensive function in question is doing a big loop in pure python, which is worst-case scenario for cProfile overhead, but we're still making over a milllion calls to unicodedata.name, which adds up whether you're profiling or not.

Example profile output:

```
(IPython) [~/projects/IPython]@(make-unicode-completions-lazier:1485049ddc)$ python -m pstats .completion_profiles/bf0a29af-7ac8-4a1d-b42d-be7eb88ea325
Welcome to the profile statistics browser.
.completion_profiles/bf0a29af-7ac8-4a1d-b42d-be7eb88ea325% sort cumtime
.completion_profiles/bf0a29af-7ac8-4a1d-b42d-be7eb88ea325% stats 20
Fri May  8 12:08:36 2020    .completion_profiles/bf0a29af-7ac8-4a1d-b42d-be7eb88ea325

         2588133 function calls (2579958 primitive calls) in 0.680 seconds

   Ordered by: cumulative time
   List reduced from 748 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      116    0.000    0.000    0.681    0.006 /home/ssanderson/projects/IPython/IPython/core/completer.py:1864(_completions)
        1    0.000    0.000    0.675    0.675 /home/ssanderson/projects/IPython/IPython/core/completer.py:1994(_complete)
        1    0.403    0.403    0.594    0.594 /home/ssanderson/projects/IPython/IPython/core/completer.py:2107(fwd_unicode_match)
  1114112    0.118    0.000    0.118    0.000 {built-in method unicodedata.name}
        1    0.000    0.000    0.079    0.079 /home/ssanderson/projects/IPython/IPython/core/completer.py:1356(_jedi_matches)
        1    0.000    0.000    0.072    0.072 /home/ssanderson/.virtualenvs/IPython/lib/python3.8/site-packages/jedi/api/__init__.py:829(__init__)
        1    0.000    0.000    0.071    0.071 /home/ssanderson/.virtualenvs/IPython/lib/python3.8/site-packages/jedi/api/__init__.py:124(__init__)
  1114112    0.067    0.000    0.067    0.000 {built-in method builtins.chr}
```

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1059,21 +1059,6 @@ class IPCompleter(Completer):
         help="Template for path at which to output profile data for completions."
     ).tag(config=True)
 
-    _unicode_names = ListTrait(
-        Unicode(),
-        help="Lazily-initialized list of unicode code point names.",
-    )
-
-    @default('_unicode_names')
-    def _unicode_names_default_value(self):
-        names = []
-        for c in range(0,0x10FFFF + 1):
-            try:
-                names.append(unicodedata.name(chr(c)))
-            except ValueError:
-                pass
-        return names
-
     @observe('limit_to__all__')
     def _limit_to_all_changed(self, change):
         warnings.warn('`IPython.core.IPCompleter.limit_to__all__` configuration '
@@ -1150,6 +1135,12 @@ class IPCompleter(Completer):
 
         # This is set externally by InteractiveShell
         self.custom_completers = None
+
+        # This is a list of names of unicode characters that can be completed
+        # into their corresponding unicode value. The list is large, so we
+        # laziliy initialize it on first use. Consuming code should access this
+        # attribute through the `@unicode_names` property.
+        self._unicode_names = None
 
     @property
     def matchers(self):
@@ -2127,7 +2118,7 @@ class IPCompleter(Completer):
             # initialize it, so we don't want to initialize it unless we're
             # actually going to use it.
             s = text[slashpos+1:]
-            candidates = [x for x in self._unicode_names if x.startswith(s)]
+            candidates = [x for x in self.unicode_names if x.startswith(s)]
             if candidates:
                 return s, candidates
             else:
@@ -2136,3 +2127,20 @@ class IPCompleter(Completer):
         # if text does not start with slash
         else:
             return u'', ()
+
+    @property
+    def unicode_names(self) -> List[str]:
+        """List of names of unicode code points that can be completed.
+
+        The list is lazily initialized on first access.
+        """
+        if self._unicode_names is None:
+            names = []
+            for c in range(0,0x10FFFF + 1):
+                try:
+                    names.append(unicodedata.name(chr(c)))
+                except ValueError:
+                    pass
+            self._unicode_names = names
+
+        return self._unicode_names
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -136,7 +136,7 @@ from IPython.utils import generics
 from IPython.utils.dir2 import dir2, get_real_method
 from IPython.utils.path import ensure_dir_exists
 from IPython.utils.process import arg_split
-from traitlets import Bool, Enum, Int, Unicode, observe
+from traitlets import Bool, Enum, Int, List as ListTrait, Unicode, default, observe
 from traitlets.config.configurable import Configurable
 
 import __main__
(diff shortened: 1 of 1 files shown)


ssanderson:
cc @Carreau for thoughts on this.

Carreau:
So we could break the loop early with a timer or if the number of completions reaches a certain limit, and advance the iterator more on next request. 

See around line 1850 in _completions (search for _timeout). 

My other though are to: 

1) Precompute the ranges, there is probably gaps we can not test, and then we can avoid the exception checking. 

2) the checking loop could be rewritten as a tree (see #12282) 

I'm also looking into making this a non bound method, so we could throw a @numba jit or have a precompiled version.

ssanderson:
@Carreau it sounds like your suggestions are for speeding up the computation of the unicode names themselves? That might be independently worthwhile, but I think the changes here make the cost of unicode completions a non-issue for most users. With these changes, we only materialize the unicode name list if the string to be completed contains `\`, which I think is sufficiently rare in normal usage that a half second delay on first isn't a big deal.

If we do want to try to speed up the unicode completions themselves, I think your suggestion of skipping known gaps or otherwise precomputing the list makes sense. I'm not convinced that switching to a tree-based algorithm would improve the user experience of these completions.

> I'm also looking into making this a non bound method, so we could throw a @numba jit or have a precompiled version.

I'd be nervous about the packaging implications of this. As far as I know, IPython doesn't currently ship any extension modules. I'd expect that adding them would make packaging IPython significantly harder.

Carreau:
> With these changes, we only materialize the unicode name list if the string to be completed contains \, which I think is sufficiently rare in normal usage that a half second delay on first isn't a big deal.

It's likely easier to move the names computation just after the `if slashpos > -1` if you are simply trying to make it lazy, and not involve traitlets. I though you were involving traitlets because you were trying something more complicated with configuration involved. 

> If we do want to try to speed up the unicode completions themselves, I think your suggestion of skipping known gaps or otherwise precomputing the list makes sense. 

That get it down on my machine from 600ms to 80 ms. There are just 664 individual ranges of valid characters. 

Le longest gap is ('0x2fa1d', '0xe0001'), about 64% of the total range. 

If we do that we need something in the test suite that warns us if the gap change like when unicodedata is updated.

> I'd be nervous about the packaging implications of this. 

That would be optional with fallback on slow path.

ssanderson:
> It's likely easier to move the names computation just after the if slashpos > -1 if you are simply trying to make it lazy, and not involve traitlets. I though you were involving traitlets because you were trying something more complicated with configuration involved.

I don't have a strong preference here, so I'll defer to you , but I personally find the traitlets-based version a bit cleaner. It ensures that the value gets lazily initialized no matter how it's accessed, and it separates out the logic of initializing the data structure from the logic of consuming it. The main downside is that it makes the first attribute access of `_unicode_names` surprisingly expensive to a naive reader, but I think that also becomes less of an issue if we speed it up in the ways you suggest. I also do think it's worth it to defer the computation even if we speed it up, since I think improving the speed of the first completion is a significant UX improvement. In my personal workflow, for example, I often start and stop IPython many times while developing, which makes the first-completion slowness more prominent.

Carreau:
> I don't have a strong preference here, so I'll defer to you , but I personally find the traitlets-based version a bit cleaner. It ensures that the value gets lazily initialized no matter how it's accessed, and it separates out the logic of initializing the data structure from the logic of consuming it.

Fair enough, I'm still unsure traitlets are still necessary though and would just use `@property` which will be more readable to someone foreign. I wouldn't even have remembered `default_value` was lazy in traitlets.

ssanderson:
Pushed an update to use more conventional machinery for lazily initializing the name list. I also fixed the docstring to use an example that actually works as a completion. The docs were using "greek small letter alpha", but the names are all upper case, so that example didn't actually work.

Carreau:
ok, let's get that in.

Carreau:
Thanks !

Issue #12291, opened 2020-05-07
%run doesn't error on failure


MSeal:
The topic came up in https://github.com/jupyter/nbclient/issues/48 that the %run magic only displays errors rather than raising them to halt execution. A similar change to make bash magics raise on error was merged a while back: https://github.com/ipython/ipython/issues/11198. Would it be reasonable to apply the same execution error logic for the other magics to raise on error instead of just log and return?

Here's the code for the run magic that would need to be touched:
https://github.com/ipython/ipython/blob/0395aa5c2f8978e1d4ebb58e2efe61cdf6a33d49/IPython/core/magics/execution.py#L505

palewire:
I'm a +1 on this idea and I'd be happy to try to help implement it. I could imagine it might introduce some issues with backward compatibility, but it seems like the correct way for iPython to abide by the Zen's admonition that ["Errors should never pass silently."](https://www.python.org/dev/peps/pep-0020/)

Carreau:
Thanks, I'm pretty sure there is an open issue for this, but I can't find it. 

And yes, I agree it should properly return the error instead of displaying it by default.

fwetdb:
The [previous PR](https://github.com/ipython/ipython/pull/12301) implemented raising errors during %run for notebook files. What remains is to raise errors during %run for Python files.

If there is still willingness to review and merge such a PR, then I am happy to provide a PR.
What do you think?

Disclosure: I am working on this issue as part of my work at Databricks because customers did complain about this inconsistent behaviour.

jasongrout-db:
I am happy to review the fix, but since I also work at Databricks, I'll defer to someone else to merge.

PR #12301, opened 2020-05-09
Exceptions raised when executing notebooks via the %run magic command

palewire:
A first step toward fixing #12291. I only addressed the execution of ipynb files because I lacked the confidence to edit the complex shell operations below.

It's unclear to me if you would want the ultimate solution to be this narrowly tailored, or if you'd prefer other elements of this function to change behavior as well. If you let me know what you'd like, I can attempt to implement it.

I can verify that this modest patch did in fact work in my notebook testing.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -701,7 +701,7 @@ class ExecutionMagics(Magics):
         if filename.lower().endswith(('.ipy', '.ipynb')):
             with preserve_keys(self.shell.user_ns, '__file__'):
                 self.shell.user_ns['__file__'] = filename
-                self.shell.safe_execfile_ipy(filename)
+                self.shell.safe_execfile_ipy(filename, raise_exceptions=True)
             return
 
         # Control the response to exit() calls made by the script being run
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -672,8 +672,8 @@ class ExecutionMagics(Magics):
             modulename = opts["m"][0]
             modpath = find_mod(modulename)
             if modpath is None:
-                warn('%r is not a valid modulename on sys.path'%modulename)
-                return
+                msg = '%r is not a valid modulename on sys.path'%modulename
+                raise Exception(msg)
             arg_lst = [modpath] + arg_lst
         try:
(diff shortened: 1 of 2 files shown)


choldgraf:
This seems good to me! Do you think we could add a test to make sure that this will properly fail when run on a notebook w/ a `%run` that doesn't work? (though I am not super familiar with ipython's testing infra...)

MSeal (changes requested):
I think you need to change all the places with a `return` above the line you edited to instead raise the given exception (or create one). Otherwise typos or missing files will error silently and not raise as well.

MSeal:
> This seems good to me! Do you think we could add a test to make sure that this will properly fail when run on a notebook w/ a `%run` that doesn't work? (though I am not super familiar with ipython's testing infra...maybe @MSeal has a better idea than I?)

Doesn't have to be notebook specific. A test could be to call the `run` function with an execution target that will always fail and make sure it raises an exception. To be thorough, have a test for each failure path to ensure if raises an exception.

palewire:
@MSeal I've added some additional exceptions. Is that what you were after?

MSeal:
@palewire Yep. I think you missed one more: https://github.com/ipython/ipython/blob/6bde6703d7b924dfb308515e70ebd926c92fa3b2/IPython/core/magics/execution.py#L676

palewire:
Okay @MSeal, I've added that one yet too. Anything else I can do?

MSeal:
Probably to make this mergable, we need some updated or new tests in https://github.com/ipython/ipython/blob/master/IPython/core/tests/test_run.py.

Thanks for contributing and getting this fixed btw :)

MSeal:
@Carreau Can comment if there's anything else since he's the maintainer here

palewire:
I've pushed a fresh commit with some unit tests. Take a look and let me know what else I can do.

choldgraf:
Nice! It looks good to me!

MSeal (approved):
LGTM

palewire:
Thanks, guys. @Carreau, let me know if there's anything else I can do.

Carreau:
Thanks, looks good, sorry this ended up hidden with all the open PRs.

PR #12305, opened 2020-05-11
Fix ipython#11508: check if line_buffer is None

TheHubbit:
There are situations where def _complete() is called and text, full_text and line_buffer arguments are all None. This leads to `text = self.splitter.split_line(line_buffer, cursor_pos)` being called with a None argument, causing an exception. The added check will default to an empty text, thereby avoiding the exception.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -2021,8 +2021,8 @@ class IPCompleter(Completer):
         # if text is either None or an empty string, rely on the line buffer
         if (not line_buffer) and full_text:
             line_buffer = full_text.split('\n')[cursor_line]
-        if not text:
-            text = self.splitter.split_line(line_buffer, cursor_pos)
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks !

Do you have a reproducible case where this is the case? 

Also it looks like the commit have been done with a email address you have not registered with github, if I merge you won't get the credits, let me know if it's ok, of if you wish to update that first.

TheHubbit:
Thank you for pointing out the email issue! I've added it in settings now. 
I get this error after ending a session when I have invoked a particular C library using ctypes before. However, it is not a standard package. So no straight forward way to repro this. But callstack matched pretty much exactly what was in defect 11508.

Carreau:
ok, let's merge then, and if really we need to reproduce we'll figure it out.

PR #12317, opened 2020-05-18
Update unicode ranges to fix test on nightly.

Carreau:
The number of valid unicode symbols on nightly has increased, so update
the ranges we skip.


Carreau:
Travis build succeeded, but github UI still show pending. 

Merging.

PR #12342, opened 2020-05-26
Fixed bug on windows where relative image path has wrong slashes

blakejgriffin:
On Windows, the `source_dir` is not posix and thus the call to `os.path.relpath` returns something like
'_build\\html\\_static'. This causes the image to not load as sphinx expects a posix path. Using `pathlib.Path` for all of the manipulation and calling `as_posix()` at the end ensures the value of `outfile` is always a posix string.


Carreau:
Thanks ! Your commit does not seem to be associated with your GitHub email, I just want to make sure you are aware in case you want to fix it before I merge.

blakejgriffin:
No Problem! Thanks for the catch, I changed the email on the commit.

Carreau:
Thanks !

PR #12343, opened 2020-05-27
rename deprected magic method -> run_line_magic

btel:


Carreau:
Ha, you just beat me to the fix.

PR #12359, opened 2020-06-01
Implement understanding on __tracebackhide__

Carreau:
`__tracebackhide__` is an attribute that when set to a true value mean the
frame need to be skipped/hidden.

Implement this in both ultratb and un ipdb so that by default frames are
hidden, add switches to toggle the behavior.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -280,26 +280,31 @@ class Pdb(OldPdb):
 
         # Set the prompt - the default prompt is '(Pdb)'
         self.prompt = prompt
+        self.skip_hidden = True
 
     def set_colors(self, scheme):
         """Shorthand access to the color table scheme selector method."""
         self.color_scheme_table.set_active_scheme(scheme)
         self.parser.style = scheme
 
+
+    def hidden_frames(self, stack):
+        """
+        Given an index in the stack return wether it should be skipped.
+
+        This is used in up/down and where to skip frames.
+        """
+        ip_hide = [s[0].f_locals.get("__tracebackhide__", False) for s in stack]
+        ip_start = [i for i, s in enumerate(ip_hide) if s == "__ipython_bottom__"]
+        if ip_start:
+            ip_hide = [h if i > ip_start[0] else True for (i, h) in enumerate(ip_hide)]
+        return ip_hide
+
     def interaction(self, frame, traceback):
         try:
             OldPdb.interaction(self, frame, traceback)
         except KeyboardInterrupt:
-            self.stdout.write('\n' + self.shell.get_exception_only())
-
-    def new_do_up(self, arg):
-        OldPdb.do_up(self, arg)
-    do_u = do_up = decorate_fn_with_doc(new_do_up, OldPdb.do_up)
-
-    def new_do_down(self, arg):
-        OldPdb.do_down(self, arg)
-
-    do_d = do_down = decorate_fn_with_doc(new_do_down, OldPdb.do_down)
+            self.stdout.write("\n" + self.shell.get_exception_only())
 
     def new_do_frame(self, arg):
         OldPdb.do_frame(self, arg)
@@ -320,6 +325,8 @@ class Pdb(OldPdb):
         return self.do_quit(arg)
 
     def print_stack_trace(self, context=None):
+        Colors = self.color_scheme_table.active_colors
+        ColorsNormal = Colors.Normal
         if context is None:
             context = self.context
         try:
@@ -329,8 +336,21 @@ class Pdb(OldPdb):
         except (TypeError, ValueError) as e:
                 raise ValueError("Context must be a positive integer") from e
         try:
-            for frame_lineno in self.stack:
+            skipped = 0
+            for hidden, frame_lineno in zip(self.hidden_frames(self.stack), self.stack):
+                if hidden and self.skip_hidden:
+                    skipped += 1
+                    continue
+                if skipped:
+                    print(
+                        f"{Colors.excName}    [... skipping {skipped} hidden frame(s)]{ColorsNormal}\n"
+                    )
+                    skipped = 0
                 self.print_stack_entry(frame_lineno, context=context)
+            if skipped:
+                print(
+                    f"{Colors.excName}    [... skipping {skipped} hidden frame(s)]{ColorsNormal}\n"
+                )
         except KeyboardInterrupt:
             pass
 
@@ -485,6 +505,16 @@ class Pdb(OldPdb):
         except KeyboardInterrupt:
             pass
 
+    def do_skip_hidden(self, arg):
+        """
+        Change whether or not we should skip frames with the
+        __tracebackhide__ attribute.
+        """
+        if arg.strip().lower() in ("true", "yes"):
+            self.skip_hidden = True
+        elif arg.strip().lower() in ("false", "no"):
+            self.skip_hidden = False
+
     def do_list(self, arg):
         """Print lines of code from the current stack frame
         """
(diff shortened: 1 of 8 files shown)


ccordoba12:
Pinging @impact27 about this one because he could be interested in this new functionality.

Carreau:
Thanks @ccordoba12.

IPython/core/interactiveshell.py
```diff
@@ -2211,11 +2211,20 @@ def complete(self, text, line=None, cursor_pos=None):
...
 
-    def set_custom_completer(self, completer, pos=0):
+    def set_custom_completer(self, completer, pos=0) -> None:
         """Adds a new custom completer function.
 
         The position argument (defaults to 0) is the index in the completers
-        list where you want the completer to be inserted."""
+        list where you want the completer to be inserted.
+
+        `completer` shoudl have the following signature::
```
impact27, line 2220: should

IPython/core/interactiveshell.py
```diff
@@ -3310,6 +3319,9 @@ async def run_code(self, code_obj, result=None, *, async_=False):
         False : successful execution.
         True : an error occurred.
         """
+        # special value to say tha anything above is IPython and should be
```
impact27, line 3322: that

impact27:
I had a quick look. Do you know if the code can still stop at a hidden frame? If I press `next` and that the next step goes in ipython would it show? If that is the case you can look into `stop_here`. (https://docs.python.org/3/library/bdb.html#bdb.Bdb.stop_here) you can use this to skip frames.

impact27:
In Spyder we do something similar. One thing that would be nice would be to have a function that can be overloaded where for each frame we could decide to either:
- skip it (It would still be in the trace-back but it is not a frame that should be debugged)
- skip and hide it (For the IPython machinery)
- print it

What I am thinking is to replace the calls to: `frame.f_locals.get("__tracebackhide__", False)` by calls to `self.should_hide(frame)` and `self.should_skip(frame)`.
```

File: IPython/core/debugger.py [hunk 3]
```diff
@@ -329,8 +332,25 @@ def print_stack_trace(self, context=None):
         except (TypeError, ValueError):
                 raise ValueError("Context must be a positive integer")
         try:
-            for frame_lineno in self.stack:
+            ip_hide = [s[0].f_locals.get("__tracebackhide__", False) for s in self.stack]
+            ip_start = [i for i,s in enumerate(ip_hide) if s == '__ipython_bottom__']
```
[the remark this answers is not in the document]
Carreau: This is a special value in IPython, otherwise it is impossible to hide all the stacks we want. Typically: 

```
ipdb> w
  /Users/bussonniermatthias/miniconda3/bin/ipython(11)<module>()
      8 if __name__ == '__main__':
      9     sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
     10     sys.exit(
---> 11         load_entry_point('ipython', 'console_scripts', 'ipython')()
     12     )
```

I use this special value to say "everything above, ignore". Otherwise yes it is boolean.

IPython/core/interactiveshell.py
Carreau:
```suggestion
        `completer` should have the following signature::
```

IPython/core/interactiveshell.py
Carreau:
```suggestion
        # special value to say that anything above is IPython and should be
```

Carreau:
Ah, thanks I forgot about some of those Bdd utilities. 

> Do you know if the code can still stop at a hidden frame? 

Yes, you need to set skip_hidden to false. 
Right now "skip" and "hide" are synonyms, I have no objections of separating them, but I'm really wondering if this is something that should be differents. I think it would be quite confusing to skip non hidden frames of ot step into hidden ones. 

using a `self.should_hide(frame)` method, would that be ok to reuse `stop_here()` for the same purpose ?

impact27:
> Right now "skip" and "hide" are synonyms, I have no objections of separating them, but I'm really wondering if this is something that should be differents. I think it would be quite confusing to skip non hidden frames of ot step into hidden ones.

When debugging some code, I don't necessarily want to stop at every library along the way. But if I print a traceback, it can still be useful to know how the function ended up being called. This is why I thought that having a way of marking frames to skip but not hide could be interesting. 

 - stepping into hidden frames doesn't make sense and so hidden should imply skip
 - skipping non hidden frames can be useful to know that the frames are there but I don't necessarily want to debug them. In Spyder we don't stop in skipped frames but you can still access them by up and down.

Thinking about it, I don't feel easy hiding any frame while going up and down as it could create confusion. I think adding a ceiling at `__ipython_bottom__` is reasonable and all the ipython machinery can be hidden. What use case would there be in hiding a frame in the middle of the stack?

impact27:
> using a `self.should_hide(frame)` method, would that be ok to reuse `stop_here()` for the same purpose ?

I think the code would be more readable with `should hide()`, event if the implementation only calls `stop_here()`.

IPython/core/debugger.py
```diff
@@ -280,27 +280,29 @@ def __init__(self, color_scheme=None, completekey=None,
...
 
+    def stop_here(self, frame):
+        """
+        Stop here is usually used only to know wether commands should stop on this frame,
+        we are also going to use it to hide frames in where, up and down for example.
+
+        This will also imply that many other commands will now skip hidden tracebacks.
+        """
+
+        return super().stop_here(frame) and frame.f_locals.get("__tracebackhide__", False)
```
impact27, line 298: and not?

IPython/core/debugger.py
```diff
@@ -329,8 +333,25 @@ def print_stack_trace(self, context=None):
...
+            ip_hide = [s[0].f_locals.get("__tracebackhide__", False) for s in self.stack]
+            ip_start = [i for i,s in enumerate(ip_hide) if s == '__ipython_bottom__']
+            if ip_start:
+                ip_start = ip_start[0]
+            else:
+                ip_start = 0
+            
+            skipped = 0
+            for i,frame_lineno in enumerate(self.stack):
+                hidden = self.stop_here(frame_lineno[0]) or (i < ip_start)
```
impact27, line 345: `stop_here` is `True` if the debugger should stop and therefore is not hidden?

Carreau:
> What use case would there be in hiding a frame in the middle of the stack?

Decorators likely, or async framework like trio, maybe things like dask. There can be a non trivial amount of things which is non relevant.

IPython/core/debugger.py
```diff
@@ -280,27 +280,35 @@ def __init__(self, color_scheme=None, completekey=None,
...
 
+    def stop_here(self, frame):
+        """
+        Stop here is usually used only to know wether commands should stop on this frame,
+        we are also going to use it to hide frames in where, up and down for example.
+
+        This will also imply that many other commands will now skip hidden tracebacks.
+        """
+
+        return super().stop_here(frame) and not frame.f_locals.get("__tracebackhide__", False)
```
impact27, line 298: Should stop_here() be:
return super().stop_here(frame) and not self.should_skip(frame)

I don't think that if `super().stop_here(frame)` is false it means the frame should be hidden.

then `should_skip` would be:
`return frame.f_locals.get("__tracebackhide__", False)`

impact27: and:
```
def should_hide(self, frame):
    return self.should_skip(frame)
```

Carreau: That's fair, updated to reflect your suggestion.

Carreau:
@meeseeksdev suggestions

IPython/core/debugger.py
```diff
@@ -631,6 +668,86 @@ def do_where(self, arg):
...
+        except ValueError:
+            self.error('Invalid frame count (%s)' % arg)
+            return
+        skipped = 0
+        if count < 0:
+            _newframe = 0
+        else:
+            _newindex = self.curindex
+            counter = 0
+            for i in range(self.curindex-1, -1, -1):
```
impact27, line 694: Same here:
```
stack = self.stack[self.curindex-1:-1:-1]
ip_hide = self.hidden_frames(stack)
for frame, hidden in zip(stack, ip_hide):
    if self.skip_hidden and hidden:
        [...]
```
and below for `do_down`.

impact27 (commented):
I had a new look, and I now think that you need to have the stack to decide if the frame should be hidden or not. I would suggest replacing `should_hide` by a function that tells you which frames to hide from a stack.

Carreau:
OK, I need to stop working on this PR when it's late at night and wait to be properly awake. Sorry I'll get back to this.

Carreau:
Yeah I'm not happy with taking only a single frame, but `self.stack` is sometime empty, and I only have a single frame. SO you can't use should_skip/should_hide from within `stop_here`. Also I know I use indices in up/down, it's because the `_select_frame()` method takes an index.

Carreau:
@meeseeksdev suggestions

IPython/core/debugger.py
```diff
@@ -329,8 +336,21 @@ def print_stack_trace(self, context=None):
         except (TypeError, ValueError):
                 raise ValueError("Context must be a positive integer")
         try:
-            for frame_lineno in self.stack:
+            skipped = 0
+            for hidden, frame_lineno in zip(self.hidden_frames(), self.stack):
```
impact27, line 340: Reading that I find it not immediately obvious why `self.hidden_frames()` and `self.stack` should have the same size. Would it be more readable if self.hidden_frames() took a stack in argument? so you would have:
`for hidden, frame_lineno in zip(self.hidden_frames(self.stack), self.stack):`

impact27:
> Yeah I'm not happy with taking only a single frame, but `self.stack` is sometime empty, and I only have a single frame. SO you can't use should_skip/should_hide from within `stop_here`. Also I know I use indices in up/down, it's because the `_select_frame()` method takes an index.

I think the code looks reasonable now. I don't think `stop_here` would be called above `__ipython_bottom__` anyway

IPython/core/debugger.py
```diff
@@ -631,6 +661,102 @@ def do_where(self, arg):
...
+            hidden_frames = self.hidden_frames()
+            for i in range(self.curindex - 1, -1, -1):
+                frame = self.stack[i][0]
+                if hidden_frames[i] and self.skip_hidden:
+                    skipped += 1
+                    continue
+                counter += 1
+                if counter >= count:
+                    break
+
```
impact27, line 707: So what happens if all frames are skipped and counter = 0? Would the new frame be the top frame?

Carreau: good point.

IPython/core/debugger.py
Carreau: sure, I don't think that changes much. But done.

Carreau:
@meeseedev suggestions

Carreau:
Ok, let's get that in, and iterate if necessary,

PR #12360, opened 2020-06-02
use $SHELL in system_piped

minrk:
we already use this in [our call to subprocess.call](https://github.com/ipython/ipython/blob/c24e8af4bf631b51ddec077d13022aba00dcf30e/IPython/core/interactiveshell.py#L2489-L2492) in system_raw used in the terminal.

This makes kernel behavior (system_piped) better match terminal IPython (system_raw) on posix systems.

For an example that used to be different:

    !wc -l <(pwd)

with `SHELL=bash`

reported on discourse: https://discourse.jupyter.org/t/ipython-process-subsitution-and/4645

### DIFF
--- a/IPython/utils/_process_posix.py
+++ b/IPython/utils/_process_posix.py
@@ -59,11 +59,12 @@ class ProcessHandler(object):
 
     @property
     def sh(self):
-        if self._sh is None:        
-            self._sh = pexpect.which('sh')
+        if self._sh is None:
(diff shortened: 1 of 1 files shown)


minrk:
Bonus: this is my first PR to IPython in a very long time!

Carreau:
Thanks ! And welcome back ! I miss you here !

PR #12377, opened 2020-06-09
Fix exception causes in debugger.py

cool-RR:
I recently went over [Matplotlib](https://github.com/matplotlib/matplotlib/pull/16706), [Pandas](https://github.com/pandas-dev/pandas/pull/32322) and [NumPy](https://github.com/numpy/numpy/pull/15731), fixing a small mistake in the way that Python 3's exception chaining is used. If you're interested, I can do it here too. I've done it on just one file right now. 

The mistake is this: In some parts of the code, an exception is being caught and replaced with a more user-friendly error. In these cases the syntax `raise new_error from old_error` needs to be used.

Python 3's exception chaining means it shows not only the traceback of the current exception, but that of the original exception (and possibly more.) This is regardless of `raise from`. The usage of `raise from` tells Python to put a more accurate message between the tracebacks. Instead of this: 

    During handling of the above exception, another exception occurred:

You'll get this: 

    The above exception was the direct cause of the following exception:

The first is inaccurate, because it signifies a bug in the exception-handling code itself, which is a separate situation than wrapping an exception.

Let me know what you think!

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -220,8 +220,8 @@ class Pdb(OldPdb):
             self.context = int(context)
             if self.context <= 0:
                 raise ValueError("Context must be a positive integer")
-        except (TypeError, ValueError):
-                raise ValueError("Context must be a positive integer")
+        except (TypeError, ValueError) as e:
+                raise ValueError("Context must be a positive integer") from e
 
         # `kwargs` ensures full compatibility with stdlib's `pdb.Pdb`.
         OldPdb.__init__(self, completekey, stdin, stdout, **kwargs)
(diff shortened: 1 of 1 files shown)


Carreau:
Oh, yeah, I think that would be great in some locations. 
There might be places where we can have a more informative message and do `from None` as well to have a cleaner error message. 

(Test are passing but User interface is not updating, I'll merge anyway.)

Carreau:
(Also I'm pretty sure I saw a thread on Python-idea about that. Started by you maybe ? )

cool-RR:
Awesome, I'll work on the rest. Yes, I do think I discussed it on python-ideas a while back :)

I personally hate `from None`, so I'll just work on the places where we agree that there shouldn't be a `from None`. I'll assign you on the PR when it's ready.

PR #12383, opened 2020-06-10
Fix exception causes all over the codebase

cool-RR:


Carreau:
Thanks.

cool-RR:
I just made a blog post about this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

Carreau:
Thanks. Maybe you could also add a patch to py-upgrade or some similar project to also have an automatic fix.

PR #12395, opened 2020-06-17
[Feature] tuple completion

cphyc:
## Rationale
The motivation of this PR is to be able to have tab completion for dict-like access of objects that accept tuples or multiple arguments as keys.

This feature would be useful to make better completion suggestions in the yt project (see e.g. https://github.com/yt-project/yt/pull/2637), which relies heavily on dict-like access where the keys are two keywords (for example `ds['gas','density']`).

## Example
```python
class Foo(dict):
    def __init__(self):
        self['foo', 'bar'] = 1
        self['foo', 'baz'] = 2
        self['foo', 'foobar'] = 3
        self['foo', 'barfoo'] = 4
        self['oof', 'oof'] = 5

    def _ipython_key_completions_(self, *args): 
        return self.keys()

f = Foo()
f['foo', 'ba<TAB>  # shows bar, baz, barfoo
f['<TAB>           # shows foo, oof
```
image

If the extra prefix (the `foo` in the above example) is not a str, then no completion should happen.

## TODO:
- [x] write tests

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -780,14 +780,20 @@ def match_dict_keys(keys: List[Union[str, bytes, Tuple[Union[str, bytes]]]], pre
     """
     prefix_tuple = extra_prefix if extra_prefix else ()
     Nprefix = len(prefix_tuple)
-    def filter_by_prefix_tuple(key):
+    def filter_prefix_tuple(key):
+        # Reject too short keys
         if len(key) <= Nprefix:
             return False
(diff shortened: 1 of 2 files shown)


cphyc:
I think that's ready to merge on my side, as soon as anyone has time to review it!

Carreau:
Oh thanks that is great ! I'll have a look.

Carreau:
Thanks !THat looks great !

PR #12407, opened 2020-06-27
Allow passing extra options to TerminalPdb.pt_init

kmaork:
Currently, subclasses of `TerminalPdb` (like in https://github.com/kmaork/madbg/blob/master/madbg/debugger.py#L55) will have to duplicate the entire `pt_init` method in order to add additional arguments for the prompt session (which accepts about 40 of them).

An example for the need to pass arguments to the prompt session would be a subclass that wants to write the prompts to a different output stream than the default. After this change it would just have to override the call to `pt_init` with `pt_init(dict(input=input, output=output))`

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -26,12 +26,21 @@ PTK3 = ptk_version.startswith('3.')
 class TerminalPdb(Pdb):
     """Standalone IPython debugger."""
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, pt_session_options=None, **kwargs):
         Pdb.__init__(self, *args, **kwargs)
         self._ptcomp = None
-        self.pt_init()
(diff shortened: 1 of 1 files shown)


Carreau:
Sweet ! Nice project ! I've marked that at 8.0, but it can easily be backported to 7.x later on if you wish.

kmaork:
Thank you!! How should I go about backporting it?

Carreau:
> Thank you!! How should I go about backporting it?

Once you're happy with your project, just ask and depending on how far I an in the 8.0 release I can cherry-pick the commit on 7.x, I do a 7.x release every last friday of the month.

kmaork:
Awesome, thanks!

kmaork:
Hi @Carreau ! I'm pretty happy with my project, so a backport would be awesome :)

Carreau:
@meeseeksdev backport to 7.x

PR #12417, opened 2020-07-01
Test Python 3.6

SylvainCorlay:


tacaswell:
This collides with #12410.  

We are now 42mo past the initial release of py36 so it is time to start dropping support for it.

Carreau:
Yeah I was planning on removing 3.6, I'll see if I can test by running `sed -i bkp s/7/6/g setup.py` on travis if there are compelling reasons to test on 3.6, but the metadata of 7.17 will mark IPython as incompatible with 3.6.

SylvainCorlay:
I am all for dropoing support.

The motivation is to understand why xeus-python is failing with 3.6 since its version 0.8.1. The reason seems to be related to IPython magics.

Carreau:
Rebased to kick the test. Modified #12410 to sed `setup.py`, so CI should be fine.
There is no reason for things to fail as it is testing 3.6 on MacOS.

PR #12435, opened 2020-07-12
More docstring autoformatting

Carreau:


IPython/core/completer.py
```diff
@@ -435,12 +433,9 @@ def _deduplicate_completions(text: str, completions: _IC)-> _IC:
     Yields
     ------
     `Completions` objects
-
-
```
Carreau, line 439: this appear to be wrong

IPython/core/completer.py
```diff
@@ -475,13 +470,10 @@ def rectify_completions(text: str, completions: _IC, *, _debug=False)->_IC:
         text that should be completed.
     completions: Iterator[Completion]
         iterator over the completions to rectify
-
-
     :any:`jedi.api.classes.Completion` s returned by Jedi may not have the same start and end, though
     the Jupyter Protocol requires them to behave like so. This will readjust
     the completion to have the same ``start`` and ``end`` by padding both
     extremities with surrounding text.
-
```
Carreau, line 484: those are wrong as well.

PR #12440, opened 2020-07-14
Avoid calling the input transformer twice in the execution stack.

Carreau:
This should allow to register input transformers that have side effects,
at least in Notebooks. Right now there are sill multiple calls which
may happen in the terminal on execution requests.


Carreau:
@meeseeksdev backport to 7.x

PR #12463, opened 2020-07-23
ipywidget -> ipywidgets typo.

Carreau:


Carreau:
Let's not backport then.

PR #12499, opened 2020-08-20
misc fixes and make  inspector configurable via %config

Carreau:


IPython/core/interactiveshell.py
```diff
@@ -837,10 +837,15 @@ def init_builtins(self):
...
-                                            self.colors,
-                                            self.object_info_string_level)
+        self.inspector = oinspect.Inspector(
+            oinspect.InspectColors,
+            PyColorize.ANSICodeColors,
+            self.colors,
+            self.object_info_string_level,
+            parent=self,
+        )
+        if self.inspector not in self.configurables:
```
Carreau, line 847:
```suggestion
        if self.inspector not in getattr(self, 'configurables', []):
```

PR #12538, opened 2020-09-03
Pin traitlets to a Python3.5 compatible version

t-ionut:
ipython 7.x guarantees compatibility with Python<3.6 in its README.
`traitlets>=5` uses the `f''` string formatting which only exists in 3.6,
so traitlets 5.x should be avoided on this branch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -190,7 +190,7 @@ install_requires = [
     'jedi>=0.10',
     'decorator',
     'pickleshare',
-    'traitlets>=4.2',
+    'traitlets<5',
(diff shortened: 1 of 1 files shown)


Carreau:
No, the readme says:

```
IPython 7.17+ requires Python version 3.7 and above.
IPython 7.10+ requires Python version 3.6 and above.
IPython 7.0 requires Python version 3.5 and above.
```

And [traitlets 5.0 mark itself at 3.7+ only](https://github.com/ipython/traitlets/blob/efc76b86154250e51d03b1180eba1010de7eb347/setup.py#L90), If you get traitlets 5.0 installed on Python 3.6 then your pip or whatever you use to install it is broken.

PR #12543, opened 2020-09-04
allow specifying multiple extensions on the command-line with traitlets 5

### LINKED ISSUE #614: Can't pass multiple ipython extensions via command line arguments
Since traitlets upgrade to 5.0.0 when I try to launch ipython with these command line options:
```
ipython --ext django_extensions.management.notebook_extension --ext my.custom.extension
```
It fails with error:
```
[TerminalIPythonApp] CRITICAL | Bad config encountered during initialization: Invalid argument: 'TerminalIPythonApp.extra_extension=DeferredConfigList(['django_extensions.management.notebook_extension', 'my.custom.extension
`'])'
```

It works fine on version 4.3.3. I'm not sure if that's a regression or new expected behavior. As you can see here https://django-extensions.readthedocs.io/en/latest/shell_plus.html#ipython-notebook `django-extensions` library expected it to work when you specify two `--ext` options.

### LINKED ISSUE #12541: Support multiple extension passed using --ext command line option
traitlets 5.0 now can accept multiple values for single command line option. It makes sense to allow adding more than one extension using `--ext` option.

See related thread here: https://github.com/ipython/traitlets/issues/614

minrk:
In traitlets 4, only one extension could be loaded with `--ext`, so:

    ipython --ext Cython --ext autoreload

would only actually load `autoreload`, the last occurrence, silently ignoring Cython. With traitlets 5, this silent failure to activate two extensions has turned into an error (see  ipython/traitlets#614).

But this syntax Just Works™ for List traits with traitlets 5, so this PR adds a new `extra_extensions` plural trait that behaves the same as the existing one, while adding support for specifying multiple extensions if the alias is given multiple times.

For backwards-compatibility, instead of updating in-place, add a new `extra_extensions` list and point the `--ext` alias to the new list on traitlets 5 and leave it on the existing `extra_extension` on traitlets 4.

closes #12541
closes ipython/traitlets#614

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -19,8 +19,10 @@ from IPython.core.application import SYSTEM_CONFIG_DIRS, ENV_CONFIG_DIRS
 from IPython.core import pylabtools
 from IPython.utils.contexts import preserve_keys
 from IPython.utils.path import filefind
+import traitlets
 from traitlets import (
     Unicode, Instance, List, Bool, CaselessStrEnum, observe,
+    DottedObjectName,
 )
 from IPython.terminal import pt_inputhooks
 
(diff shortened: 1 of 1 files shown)


IPython/core/shellapp.py
```diff
@@ -7,6 +7,7 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from distutils.version import LooseVersion as V
```
Carreau, line 10: distutils is in the process of being removed from Python, let's not introduce new dependencies on it.

minrk: Arg, I just saw that. Is there another stdlib version-parsing implementation?

minrk: nvm, we have `traitlets.version_info`

IPython/core/shellapp.py
```diff
@@ -118,8 +125,27 @@ class InteractiveShellApp(Configurable):
...
+        DEPRECATED. Dotted module name of a single extra IPython extension to load.
+
+        Only one extension can be added this way.
+
+        Only used with traitlets < 5.0, plural extra_extensions list is used in traitlets 5.
+        """,
+    ).tag(config=True)
+
+    extra_extensions = List(
+        DottedObjectName(),
```
Carreau, line 141: Has, I forgot we could do that.

PR #12545, opened 2020-09-07
more useful feedback on style checker

minrk:
follow-up to #12502 

- give copy-pasteable command to fix formatting with `darker`
- use $TRAVIS_COMMIT_RANGE to limit to PR, instead of hardcoded ref
- only run on pull requests
- run style check before tests, for quicker feedback

For example feedback, see [this run](https://travis-ci.org/github/ipython/ipython/jobs/724854948#L471-L473):

```
    darker -r c4780a39d4d176bd73880658b1c9e8fca62acaf0
```


Carreau:
I'm unsure about `$TRAVIS_COMMIT_RANGE`, with a single commit this ensure that at some point int the future darker/black will be equivalent.

You can see that as a sum of delta accumulating floating point errors.

But thanks !

Issue #12515, opened 2020-08-27
Pathlib, Pathlib everywhere.


Carreau:
There are many places where we could make use of Pathlib. 

Look for any places that uses `with open(...)` and ask yourself:

- is the argument a string ? 
- would it make sens to make it a `Path()`, 
- how far upstream in the code can I make it a Path.

Don't try to bite more than you can chew (or more than I can review), try to fix 1 place at a time.

Carreau:
for example see https://github.com/ipython/ipython/pull/12512

Inception95:
Hi @Carreau, if no one is working on /core/page.py for using Pathlib, can I take this part?

Carreau:
> Hi @Carreau, if no one is working on /core/page.py for using Pathlib, can I take this part?

Sure, please. Thanks.

barhenkro:
Hi, I would like to work on core/interactiveshell.py and add there more use of Pathlib.
Can I help with that?

OdinTech3:
Hey @Carreau can I take on testing/iptestcontroller.py and core/magics/execution.py ?

arindam9999:
Hi, I would like to work on ipython/IPython/core/completer.py  and change string to Path.
Can I help with that?

MrMino:
Might also be of use: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

There are many places other than `with open()` that might use pathlib treatment:

```diff
-    conda_history = os.path.join(sys.prefix, 'conda-meta', 'history')
-    return os.path.exists(conda_history)
+    return Path(sys.prefix, 'conda-meta', 'history').exists()
```

OdinTech3:
That's a good point.

MrMino:
> Might also be of use: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module
> 
> There are many places other than `with open()` that might use pathlib treatment:
> 
> ```diff
> -    conda_history = os.path.join(sys.prefix, 'conda-meta', 'history')
> -    return os.path.exists(conda_history)
> +    return Path(sys.prefix, 'conda-meta', 'history').exists()
> ```


Although some of them might be less trivial and more error prone, especially when they're interpreted as text. For example, when returning a `PosixPath` object breaks code in the scope above because it is being `' '.join()`-ed with other shell arguments.

sumagnadas:
[...]

PR #11172, opened 2018-06-07
check that virtualenv dir exists before using it

cedric05:
closes #11100

Issue #11100, opened 2018-04-17
ipython startup error in terminal 


gkvallis:
I did a clean install of anaconda python on a Mac, and option 6.2.1 is (among things) installed. I type python (gives me python 3.6.5) on a terminal, and I get the expected response  of an ipython terminal. I exit and type conda update ipython, and it goes to 6.3.1. Now when I type ipython at the terminal it gives me a string of errors, for example:
```
 Traceback (most recent call last):
  File "/Users/gkv/anaconda3/bin/ipython", line 11, in <module>
    sys.exit(start_ipython())
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/IPython/__init__.py", line 125, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-112>", line 2, in initialize
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/IPython/terminal/ipapp.py", line 317, in initialize
    self.init_shell()
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/IPython/terminal/ipapp.py", line 333, in init_shell
    ipython_dir=self.ipython_dir, user_ns=self.user_ns)
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/traitlets/config/configurable.py", line 412, in instance
    inst = cls(*args, **kwargs)
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/IPython/terminal/interactiveshell.py", line 430, in __init__
    super(TerminalInteractiveShell, self).__init__(*args, **kwargs)
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 483, in __init__
    self.init_virtualenv()
  File "/Users/gkv/anaconda3/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 740, in init_virtualenv
    if p_exe_up2 and os.path.samefile(p_exe_up2, p_venv):
  File "/Users/gkv/anaconda3/lib/python3.6/genericpath.py", line 97, in samefile
    s2 = os.stat(f2)
[...]

PR #12548, opened 2020-09-09
Introduce pathlib on init_virtualenv.

GabrielSimonetto:
Motivated by #12515

I'm still looking for a way to test the code on a windows VM, and will try to introduce automated tests later

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -903,12 +903,12 @@ class InteractiveShell(SingletonConfigurable):
         if 'VIRTUAL_ENV' not in os.environ:
             # Not in a virtualenv
             return
-        elif os.environ['VIRTUAL_ENV'] == '':
+        elif os.environ["VIRTUAL_ENV"] == "":
             warn("Virtual env path set to '', please check if this is intended.")
             return
 
         p = Path(sys.executable)
-        p_venv = Path(os.environ['VIRTUAL_ENV'])
+        p_venv = Path(os.environ["VIRTUAL_ENV"])
 
         # fallback venv detection:
         # stdlib venv may symlink sys.executable, so we can't use realpath.
@@ -920,9 +920,9 @@ class InteractiveShell(SingletonConfigurable):
             paths.append(p.resolve())
 
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
-        if str(p_venv).startswith('\\cygdrive'):
+        if str(p_venv).startswith("\\cygdrive"):
             p_venv = Path(str(p_venv)[11:])
-        elif len(str(p_venv)) >= 2 and str(p_venv)[1] == ':':
+        elif len(str(p_venv)) >= 2 and str(p_venv)[1] == ":":
             p_venv = Path(str(p_venv)[2:])
 
         if any(os.fspath(p_venv) in os.fspath(p) for p in paths):
@@ -932,15 +932,12 @@ class InteractiveShell(SingletonConfigurable):
         warn("Attempting to work in a virtualenv. If you encounter problems, please "
              "install IPython inside the virtualenv.")
         if sys.platform == "win32":
-            virtual_env = Path(os.environ['VIRTUAL_ENV']).joinpath(
-                'Lib',
-                'site-packages'
+            virtual_env = Path(os.environ["VIRTUAL_ENV"]).joinpath(
+                "Lib", "site-packages"
             )
         else:
-            virtual_env = Path(os.environ['VIRTUAL_ENV']).joinpath(
-                'lib',
-                "python{}.{}".format(*sys.version_info[:2]),
-                'site-packages'
+            virtual_env = Path(os.environ["VIRTUAL_ENV"]).joinpath(
+                "lib", "python{}.{}".format(*sys.version_info[:2]), "site-packages"
             )
         
         import site
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -28,6 +28,7 @@ import subprocess
 import warnings
 from io import open as io_open
 
+from pathlib import Path
 from pickleshare import PickleShareDB
 
 from traitlets.config.configurable import SingletonConfigurable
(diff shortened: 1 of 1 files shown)


IPython/core/interactiveshell.py
```diff
@@ -902,42 +903,45 @@ def init_virtualenv(self):
...
             return
-        
-        p = os.path.normcase(sys.executable)
-        p_venv = os.path.normcase(os.environ['VIRTUAL_ENV'])
-
-        # executable path should end like /bin/python or \\scripts\\python.exe
-        p_exe_up2 = os.path.dirname(os.path.dirname(p))
-        if p_exe_up2 and os.path.exists(p_venv) and os.path.samefile(p_exe_up2, p_venv):
-            # Our exe is inside the virtualenv, don't need to do anything.
+        elif os.environ['VIRTUAL_ENV'] == '':
```
GabrielSimonetto, line 906: I'm changing this because of this issue https://github.com/ipython/ipython/issues/11100
which was resolved this way https://github.com/ipython/ipython/pull/11172

The code I'm using here was suggested on the issue, and it seems better to me,
specially now that we are using Pathlib, since Path('') actualy means the path in which you currently are

So, as is, we would get a True when we should get a False
In [4]: Path('').exists()
Out[4]: True

If the warning is a bit excessive I can remove it and merge with the upwards if statement

IPython/core/interactiveshell.py
```diff
@@ -902,42 +903,45 @@ def init_virtualenv(self):
...
+
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
         if p_venv.startswith('\\cygdrive'):
             p_venv = p_venv[11:]
         elif len(p_venv) >= 2 and p_venv[1] == ':':
             p_venv = p_venv[2:]
 
-        if any(p_venv in p for p in paths):
-            # Running properly in the virtualenv, don't need to do anything
+        if any(os.fspath(p_venv) in os.fspath(p) for p in paths):
```
GabrielSimonetto, line 928: Found no pathlibian way to do this.

Can change to `str(p_venv)` if it's better, we can even forget running pathlib on this section, since the previous code seemed a little better to me

IPython/core/interactiveshell.py
```diff
@@ -902,42 +903,45 @@ def init_virtualenv(self):
...
         paths = [p]
-        while os.path.islink(p):
-            p = os.path.normcase(os.path.join(os.path.dirname(p), os.readlink(p)))
-            paths.append(p)
-        
+        while p.is_symlink():
+            p = Path(os.readlink(p))
+            paths.append(p.resolve())
+
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
```
GabrielSimonetto, line 922: Didn't change this code because I couldn't test it on my linux, currently looking for some VM thing

IPython/core/interactiveshell.py
```diff
@@ -902,42 +903,45 @@ def init_virtualenv(self):
         if 'VIRTUAL_ENV' not in os.environ:
             # Not in a virtualenv
             return
-        
-        p = os.path.normcase(sys.executable)
-        p_venv = os.path.normcase(os.environ['VIRTUAL_ENV'])
-
-        # executable path should end like /bin/python or \\scripts\\python.exe
-        p_exe_up2 = os.path.dirname(os.path.dirname(p))
-        if p_exe_up2 and os.path.exists(p_venv) and os.path.samefile(p_exe_up2, p_venv):
```
GabrielSimonetto, line 911: This condicional is already met on line 928, I have changed it's comments to reflect this change

IPython/core/interactiveshell.py
```diff
@@ -902,42 +903,45 @@ def init_virtualenv(self):
...
         # stdlib venv may symlink sys.executable, so we can't use realpath.
         # but others can symlink *to* the venv Python, so we can't just use sys.executable.
         # So we just check every item in the symlink tree (generally <= 3)
         paths = [p]
-        while os.path.islink(p):
-            p = os.path.normcase(os.path.join(os.path.dirname(p), os.readlink(p)))
-            paths.append(p)
-        
+        while p.is_symlink():
+            p = Path(os.readlink(p))
```
GabrielSimonetto, line 919: According to [this link](https://stackoverflow.com/questions/41460434/getting-the-target-of-a-symbolic-link-with-pathlib) 
        
I believe it's not possible to replicate this on pathlib
If someone sees a better solution please let me know.

GabrielSimonetto (commented):
Left some reasonings behind my decisions

IPython/core/interactiveshell.py
Carreau: :+1: we are 3.7+ so `p.resolve()` might work, but we can check this later.

Carreau:
Thanks that's looks great. If there are breakage we can still fix later and make sure there is a test case in the windows CI.

IPython/core/interactiveshell.py
```diff
@@ -902,42 +903,45 @@ def init_virtualenv(self):
...
+            paths.append(p.resolve())
+
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
         if p_venv.startswith('\\cygdrive'):
             p_venv = p_venv[11:]
         elif len(p_venv) >= 2 and p_venv[1] == ':':
             p_venv = p_venv[2:]
```
Carreau, line 926:
```suggestion
        if str(p_venv).startswith('\\cygdrive'):
            p_venv = Path(str(p_venv)[11:])
        elif len(str(p_venv)) >= 2 and str(p_venv)[1] == ':':
            p_venv = Path(str(p_venv)[2:])
```

Let's try this to fix the test failures.

IPython/core/interactiveshell.py
GabrielSimonetto: `p.resolve()` seems to follow symlinks all the way. Here we would like to follow them one by one to check later if any of them are inside the scope for `any(os.fspath(p_venv) in os.fspath(p) for p in paths)`

MrMino:
@Carreau can I backport this to `7.x`? Need this for #13140.

Carreau:
> @Carreau can I backport this to 7.x? I need this for #13140.

Sure, you are a core dev, you don't need my authorisation :-) 

@meeseeksdev backport to 7.x

MrMino:
@Carreau I know, but I was worried that there are reasons behind not doing so that I'm not seeing. Thanks :)!

Carreau:
> but I was worried that there are reasons behind not doing so that I'm not seeing

I think It's also because i have less time than before and try to backport only what is critical and that was close to a year ago and I thoughts 8.0 was not that far down the line.

PR #12557, opened 2020-09-22
Use pathlib in page.py

Inception95:
#12515 Change the tmpname to tmppath, and find in https://docs.python.org/3/library/pathlib.html about equivalent of os.remove() and Path.unlink().


Carreau:
Thanks !

PR #12559, opened 2020-09-23
use pathlib in `utils.openpy`

rushabh-v:
related to: https://github.com/ipython/ipython/issues/12515

### DIFF
--- a/IPython/utils/openpy.py
+++ b/IPython/utils/openpy.py
@@ -7,6 +7,7 @@ Much of the code is taken from the tokenize module in Python 3.2.
 
 import io
 from io import TextIOWrapper, BytesIO
+from pathlib import Path
 import re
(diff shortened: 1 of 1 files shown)


IPython/utils/openpy.py
```diff
@@ -72,7 +73,8 @@ def read_py_file(filename, skip_encoding_cookie=True):
     -------
     A unicode string containing the contents of the file.
     """
-    with open(filename) as f:   # the open function defined in this module.
+    filepath = Path(filename)
+    with filepath.open() as f:   # the open function defined in this module.
```
Carreau, line 77:
```suggestion
    with open(filepath) as f:   # the open function defined in this module.
```

Slightly tricky as `open` here is not the builtin `open`; hence why test were failing.

Carreau:
Thanks !

PR #12560, opened 2020-09-24
Small typo fix in pylab deprecation message

alchzh:


Carreau:
Oops, thanks, nice catch.

PR #12562, opened 2020-09-27
use pahtlib in setupegg.py

Lightyagami1:
related to: #12515

### DIFF
--- a/setupegg.py
+++ b/setupegg.py
@@ -3,5 +3,7 @@
 
 # Import setuptools and call the actual setup
 import setuptools
-with open('setup.py', 'rb') as f:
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks ! 

The mail you used in the commit is not the same as the one registered with github, you will not get credit for the commit like that.  Is that intended  and do you want help fixing it ? 

Can you also run autoformatting on the code: 

```
$ pip install darker
$ darker -r f5176ad0445842d99985f103e378da9ddfd53020 .
``` 

Much appreciated.

Carreau:
Thanks, will close for lack of responses. 

Feel free to ask if you wish to reopen.

PR #12569, opened 2020-09-30
Improve Docs

BarnakGhosh:


README.rst
```diff
@@ -132,15 +132,15 @@ project that you might want to use:
 - `ptpython and ptipython <https://pypi.org/project/ptpython/>`
 - `xonsh <https://xon.sh/>`
 
-Ignoring commits with git blame.ignoreRevsFile
+Ignoring commits with the git blame.ignoreRevsFile
 ==============================================
```
Carreau, line 136: That is technically incorrect, `blame.ignoreRevsFile` is an option, thus it should be treated like a proper name and not a common name. Plus the RST syntax is now invalid with the underline.

README.rst
```diff
@@ -132,15 +132,15 @@ project that you might want to use:
...
 
 As of git 2.23, it is possible to make formatting changes without breaking
 ``git blame``. See the `git documentation
 <https://git-scm.com/docs/git-config#Documentation/git-config.txt-blameignoreRevsFile>`_
 for more details.
 
-To use this feature you must:
```
Carreau, line 143: That is incorrect english, you "must install" and "must configure", not "must do install" and "must do configure".

PR #12573, opened 2020-09-30
Update __init__.py

Gari1309:


Carreau:
Thanks for your contributions. 

What you changed is an [encoding comment](https://www.python.org/dev/peps/pep-0263/). They have a well defined structure, modifying the way you did would make this line incorrect. As we are python 3 only now you could simply remove the full line as an alternative.

PR #12574, opened 2020-10-01
Fix broken link in documentation

farisachugthai:
It's on an orphaned page so not the biggest deal but I wanted to start with a few small contributions to the project first. I'm also thinking about adding a handful of updates to the sphinx ipython directive documentation if that would be of any interest?


Carreau:
Thanks, this is great !  Welcome and enjoy hacktoberfest !

farisachugthai:
Thank you!

PR #12576, opened 2020-10-01
Updated comments 

shashank06-sudo:


Carreau:
Hi, 

Due to the heavy number of spam we get because of hacktoberfest, we need to take drastic measure and will close PRs and ban users which contributions look like spam.

Apologies if this was not the case, and if you have been blocked inadvertently reach out via another channel like twitter and we'll unblock you.

PR #12580, opened 2020-10-02
DOC: Define ipy directive as such in docs.

farisachugthai:
By utilizing the directive `rst:directive::`, one can now link to the
documentation on the sphinx directive using an inline reference as well
as now being able to link to any of the options.

Put this early in the documentation as it's useful to quickly summarize
the options and decorators available to the user immediately.

The currently provided examples are phenomenal but having a thorough
reference and all of the options in one place and having them be easy to
create a hyperlink to is also very important.


Carreau:
Thanks ! Looks great !

PR #12582, opened 2020-10-02
Improvement and corrections of text

Krishnapal-rajput:


ccordoba12:
This is not a meaningful contribution. It looks instead as spam due to Hacktoberfest.

PR #12584, opened 2020-10-03
Remove Release Notes Workflow

augustogoulart:
The `release-notes.ym` workflow did not take oof and the base GitHub action
for generating the notes doesn't seem to be maintained anymore.

Removing these files until we figure out a better alternative.


Carreau:
Thanks !

PR #12585, opened 2020-10-03
256 bit monokai theme for prompt

zacharesmer:
A better version of this would allow for a theme completely configurable in the config file, but I couldn't figure out how to do that. This works like the Linux, LightBG, etc. themes, so it affects ipdb, unlike the pygments themes.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -268,14 +268,13 @@ class Pdb(OldPdb):
         cst['LightBG'].colors.breakpoint_enabled = C.LightRed
         cst['LightBG'].colors.breakpoint_disabled = C.Red
 
-        cst['Neutral'].colors.prompt = C.Blue
-        cst['Neutral'].colors.breakpoint_enabled = C.LightRed
-        cst['Neutral'].colors.breakpoint_disabled = C.Red
-        
-        cst['Monokai'].colors.prompt = C.MonokaiBlue
-        cst['Monokai'].colors.breakpoint_enabled = C.MonokaiMagenta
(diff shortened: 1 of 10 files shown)


Carreau:
Thanks !

I want to re-dive into this as I'm pretty sure there are way to define external theme, and I don't want to duplicate al style defined in Pygments.

zazezizozu:
I dug around some more and figured out how to make the colors that aren't affected by `highlighting_style_overrides` configurable. I don't believe this was possible previously. I changed this theme's name to `User`, then made it possible to edit the colors for that theme in the config file. 

Unfortunately, I discovered that traitlets doesn't easily turn a nested tuple into a string, so the config doesn't generate or get read properly. When I change the traitlets configurable file to wrap input in a tuple, it passes all the tests though. I'd love to hear if there's a clever way to do that from the input side; everything I tried so far didn't work.

[The code is here](https://github.com/zazezizozu/ipython/tree/usertheme), and for reference the change in `<traitlets path>/traitlets/config/configurable.py` is:
`  -  lines.append(indent('Current: %r' % getattr(inst, trait.name), 4))` 
`  + lines.append(indent('Current: %r' % (getattr(inst, trait.name),), 4))`

zazezizozu:
Turns out the tuple solution was to configure it as a list and cast it to a tuple later, so there is now a working version of the configurable color theme [here](https://github.com/zazezizozu/ipython/tree/testbranch).

zazezizozu:
I figured out the tuple problem, and a working version of the configurable theme is [here](https://github.com/zazezizozu/ipython/tree/testbranch).

Carreau:

PR #12586, opened 2020-10-03
Automatic matching of brackets and quotes

mskar:
This PR 
1. implements insertion and deletion of matched brackets and quotes
2. makes changes to [IPython/terminal/shortcuts.py](https://github.com/ipython/ipython/pull/12586/files#diff-e9a59ec7c6360d84aafa289edca1c50b)
3. shamelessly borrows code from [randy3k/radian](https://github.com/randy3k/radian/blob/a57f0b50ae8a7f4f152d6b4d3b9d63c365eec788/radian/key_bindings.py#L171)

Example:
1. Start ipython
2. Press `(`, `[`, `{`, `'`, or `"` to insert a pair of round/square/curly brackets or single/double quotes with the cursor inside
3. Press `backspace` or ctrl h to delete the inserted bracket or quote pair

To install to version of `ipython` in this pull request, run
`python -m pip install git+https://github.com/mskar/ipython.git@auto_match`

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -188,6 +188,15 @@ class TerminalInteractiveShell(InteractiveShell):
         allow_none=True
     ).tag(config=True)
 
+    auto_match = Bool(
+        False,
+        help="""
+        Automatically add/delete closing bracket or quote when opening bracket or quote is entered/deleted.
+        Brackets: (), [], {}
+        Quotes: '', \"\"
+        """,
+    ).tag(config=True)
+
     mouse_support = Bool(False,
         help="Enable mouse support in the prompt\n(Note: prevents selecting text with the mouse)"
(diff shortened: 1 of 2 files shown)


Carreau:
Interesting, I'm wondering if we should make that opt-in; it might screw up people using IPython as a subprocess.

mskar:
Good idea, @Carreau, I will add an option called `auto_match` to the settings.

mskar:
@Carreau, I implemented the change to make this feature opt-in.
The feature only works if `c.InteractiveShell.auto_match = True` is added to `~/.ipython/profile_default/ipython_config.py`. In other words, the feature is disabled by default.
Please let me know if any other changes need to be made.

Carreau:
Sorry that got lost in the second pages of PRs in GitHub, rebased and fixed, sorry for the delay.

PR #12587, opened 2020-10-03
[Docs] Improve installing docs if on zsh

hongshaoyang:
- zsh uses square brackets for globbing (see https://stackoverflow.com/a/30539963).
- installation would be confusing for zsh users who will get a 'no matches found' error when running `pip install .[test]`


Carreau:
Thanks.

Issue #12575, opened 2020-10-01
Feature Request: Include parentheses when tab completing functions


mskar:
Tab completion in IPython shows functions with parentheses:
![image](https://user-images.githubusercontent.com/12806339/94769300-b5289e00-037f-11eb-82d8-5e37cea96131.png)
but when I tab complete a function the parentheses are not included:
![image](https://user-images.githubusercontent.com/12806339/94769325-c2de2380-037f-11eb-8164-e3aff31b6f96.png)
I would like tab completion of functions to also add parentheses and put my cursor inside:
![image](https://user-images.githubusercontent.com/12806339/94769372-e012f200-037f-11eb-8e63-9193813cd6ab.png)
This is the default behavior in VSCode (below) and PyCharm:
![image](https://user-images.githubusercontent.com/12806339/94769506-341dd680-0380-11eb-92b6-06d03ffe94b3.png)
![image](https://user-images.githubusercontent.com/12806339/94769524-40099880-0380-11eb-9331-4aa3eaf9286e.png)
It can be enabled in the [Python language server protocol (LSP)](https://code.visualstudio.com/docs/python/editing#_enable-intellisense-for-custom-package-locations).

I think this feature would be widely appreciated by users and including should be the default behavior for tab completion of functions. Some users may prefer the current behavior, so it could be controlled via a setting in `ipython_config.py`.

mskar:
I implemented this feature in my `ipython_config.py`:

<details><summary>Click here to see the code:</summary>

```python
from IPython import get_ipython
from prompt_toolkit import filters
from prompt_toolkit.enums import DEFAULT_BUFFER
from prompt_toolkit.completion import CompleteEvent
from prompt_toolkit.application.current import get_app

ip = get_ipython()
registry = ip.pt_app.key_bindings
handle = registry.add
focused_insert = filters.has_focus(DEFAULT_BUFFER) & filters.vi_insert_mode
focused_insert_and_completion = focused_insert & filters.has_completions


@handle("tab", filter=focused_insert)
@handle("c-space", filter=focused_insert)
def _(event):
    b = event.current_buffer
    if b.completer is None:
        return
    complete_event = CompleteEvent(completion_requested=True)
    completions = list(b.completer.get_completions(b.document, complete_event))
    if len(completions) == 1: # only one possible completion
        completion = completions[0]
    elif not b.complete_state: # no completion menu
        b.start_completion(insert_common_part=True)
        completion = None
[...]

PR #12589, opened 2020-10-04
Provide option for tab to apply completion

mskar:
In #12575, I requested a feature that adds parentheses after functions completed with `Tab` or `Enter`. 
This pull request implements the feature I requested (function completion with added parentheses), but also strives to make autocompletion work like in the VSCode Python extension and PyCharm. In VSCode and PyCharm, `Tab` applies completion instead of cycling through completion options, like in bash or zsh.

This feature is enabled by default, but it can be disabled by adding 
```python
c.InteractiveShell.tab_apply_completion = False
```
to `ipython_config.py`. 

For more information, please take a look at [this comment](https://github.com/ipython/ipython/issues/12575#issuecomment-702484214) in #12575.

To install IPython with the `tab_apply_completion` feature, run 
`python -m pip install git+https://github.com/mskar/ipython.git@tab_apply_completion`
## TLDR: 
previous behavior: 
- `Tab` cycles completion
- function completion does not add parentheses 

proposed behavior: 
- `Tab` applies completion
- function completion adds parentheses

## Changes:
Changed: when triggering completion with `Tab`, insert text that all options have in common
Changed: use `Tab` to accept only option or accept selected
Changed: do not move through options with not `Tab` and `Shift-Tab`
Unchanged: use `Enter` to accept if completion menu is showing
Unchanged: move through options with `c-n` and `c-p`
Unchanged: cancel completion with `c-g`

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -142,6 +142,32 @@ class TerminalInteractiveShell(InteractiveShell):
         help="Add shortcuts from 'emacs' insert mode to 'vi' insert mode.",
     ).tag(config=True)
 
+    auto_complete_function_parentheses = Bool(True,
+    help="Add parentheses when autocompleting functions.",
+    ).tag(config=True)
+
+    auto_complete_selected_option_on_tab = Bool(True,
+        help="""When the completion menu is showing
+        and a completion option is selected,
+        apply the currently selected completion option with tab.""",
+    ).tag(config=True)
+
+    auto_complete_top_option_on_enter = Bool(True,
+        help="""When the completion menu is showing,
+        but no completion options are selected,
+        apply the first completion menu option with enter.""",
+    ).tag(config=True)
+
+    auto_complete_top_option_on_tab = Bool(True,
+        help="""When the completion menu is showing
+        and no completion options are selected,
+        apply the first completion menu option with tab.""",
+    ).tag(config=True)
+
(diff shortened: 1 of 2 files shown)


mskar:
I split the original feature into 5 independent features.
Each feature has its own setting (disabled by default) to allow for granular control of completion behavior:

1. `auto_complete_function_parentheses`:
  - Tab: Unchanged unless combined with one of the settings that end in `_option_on_tab`.
  - Enter: Accepts selected completion and add parentheses to functions
2. `auto_complete_selected_option_on_tab`:
  - Tab: Accepts the selected completion option if an option is selected
  - Enter: Unchanged
3. `auto_complete_top_option_on_enter`:
  - Tab: Unchanged
  - Enter: Accepts the first completion option if the completion menu is showing
4. `auto_complete_top_option_on_tab`:
  - Tab: Accepts the first completion option if the completion menu is showing and no option is selected
  - Enter: Unchanged
5. `auto_complete_only_option_on_tab`:
  - Tab: Accepts completion if there is only one completion option or starts completion if there are multiple options
  - Enter: Unchanged

These features can be combined.
Enabling all of the features makes completion behave like in PyCharm and VSCode.

Is 5 settings too many?
Would it be a good idea to add another setting that combined 2-5 (`auto_complete_[selected,top,only]_option_on_[tab,enter]`)?

Please let me know if any of the settings should be renamed.

Carreau:

PR #12591, opened 2020-10-04
use pathlib on test_deepreload.py

dswij:
Using pathlib.Path() for better readability.

See #12515


Carreau:
Awesome  ! Much apprecitated.

Issue #11919, opened 2019-10-17
Broken link to "Edit on Github" in ipython documentation


btel:
The API documentation on readthedocs contains a link "Edit on github" in the header. Clicking on the link brings us to a 404 github page:

For example, 

Documentation page:  https://ipython.readthedocs.io/en/latest/api/generated/IPython.conftest.html

Github link: https://github.com/ipython/ipython/blob/master/docs/source/api/generated/IPython.conftest.rst

btel:
probably related to: https://github.com/readthedocs/sphinx_rtd_theme/pull/393

augustogoulart:
@meeseeksdev tag good first issue

augustogoulart:
@meeseeksdev tag "good first issue"

augustogoulart:
@meeseeksdev tag good first issue

omsitapara23:
HI,
I would like to work on this issue. I am a first timer so can someone please guide me what exactly needs to be done and how can i approach?

Thanks,
Om.

Nusart:
Hi, @btel I would like to work on this. Can you please assign?  Also since I am doing this for the first time, it would be really helpful if you can guide me.

btel:
Hi @Nusart, since I am not a maintainer of the project, I can not assign you. I did a quick check to understand where the problem comes from, but I couldn't make much progress. I guess you would need to dig into the innerworkings of sphinx and readthedocs template (https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/breadcrumbs.html).

A good first step would be to generate the docs locally and check whether this error appears as well.

Joe-Sin7h:
Hello! This is my first time contributing to an open source project , I would like to work on this issue .
can i get assigned ?

MrMino:
Not sure if I'm looking at the right thing, but the links in the docs this issue mentions are still broken.

23kcarlson:
.gitignore contains "docs/source/api/generated" so the pages aren't a part of the repository.
The pages seem to be generated based off of comments in the code, so should the edit button link back to that?

MrMino:
@23kcarlson that would be the best option, but I'm not sure it is feasible.

PR #12598, opened 2020-10-06
Fixing issue #11919 Broken link to "Edit on Github" in ipython documentation. 

ghostcodekc:
Files were blocked in .gitignore which causes them not to be built or put into github.


ghostcodekc:
I think this is correct, however, I don't think the correct solution would be to allow editing since these files are auto generated.

MrMino:
See #12871

MrMino:
I'm going to close this because I believe that adding generated files to the repository verbatim is not the right way to go about this.

It's an issue with the machinery that generates the files, not with the lack of the files in the repo.

Issue #12180, opened 2020-03-10
display.Video function seems broken when passed `filename` argument


fperez:
Calling (assuming the file `animation.m4v` exists locally) `Video("animation.m4v")` works just fine:

![image](https://user-images.githubusercontent.com/57394/76286463-7f935a80-625f-11ea-8a93-8d5e34ee1504.png)

but if I try to pass either explicitly the `filename` argument or the `data` one, as per the docs, I get errors. This for `filename`:

```
Video(filename="animation.m4v")

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-78db781f0939> in <module>
----> 1 Video(filename="animation.m4v")

~/local/conda/lib/python3.7/site-packages/IPython/core/display.py in __init__(self, data, url, filename, embed, mimetype, width, height)
   1362             url = data
   1363             data = None
-> 1364         elif os.path.exists(data):
   1365             filename = data
   1366             data = None

~/local/conda/lib/python3.7/genericpath.py in exists(path)
     17     """Test whether a path exists.  Returns False for broken symbolic links"""
     18     try:
---> 19         os.stat(path)
     20     except OSError:
     21         return False

TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

and this for `data` (granted, I'm not sure if I should be doing some unicode handling here, but the docstring says bytes OR strings are OK, so I assumed this was fine:

```
Video(data=open("animation.m4v", "rb").read())

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-1959c93b1077> in <module>
----> 1 Video(data=open("animation.m4v", "rb").read())

~/local/conda/lib/python3.7/site-packages/IPython/core/display.py in __init__(self, data, url, filename, embed, mimetype, width, height)
   1362             url = data
   1363             data = None
-> 1364         elif os.path.exists(data):
   1365             filename = data
   1366             data = None

~/local/conda/lib/python3.7/genericpath.py in exists(path)
     17     """Test whether a path exists.  Returns False for broken symbolic links"""
     18     try:
---> 19         os.stat(path)
     20     except OSError:
     21         return False

ValueError: stat: embedded null character in path
```

@Carreau, happy to try to fix this so I have a bit of code in IPython after so many years :)

Carreau:
> @Carreau, happy to try to fix this so I have a bit of code in IPython after so many years :)

[...]

PR #12609, opened 2020-10-09
Fix to #12180 (display.Video failing with data=None)

scarlehoff:
I've noticed a bug when reading Video by filename (checking the issues I've seen it was already reported here: #12180) 

**The bug:**
While this would work perfectly
```python
from IPython.core import display
im = display.Image(filename="path_to_some_image.jpg")
```

I noticed this would fail:
```python
from IPython.core import display
vi = display.Video(filename="path_to_some_video.avi")
```

with `TypeError` due to the `os.path.exists(data)` check, which I think it can be inconsistent. 

**The fix:**
I've just added a check to skip it if there is no `data`.

(I've noticed only afterwards that there was already someone assigned to it though!)

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1086,7 +1086,7 @@ class Video(DisplayObject):
         if url is None and isinstance(data, str) and data.startswith(('http:', 'https:')):
             url = data
             data = None
-        elif os.path.exists(data):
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks !

PR #12610, opened 2020-10-09
Use pathlib in display.py

scarlehoff:
Updated `display.py` classes to use pathlib when possible.

While looking at a different bug I noticed that the classes in `display.py` would mostly accept also a `Path` and also that there was an issue ( #12515) about changing to pathlib whenever possible. Don't know whether there is a specific reason why it wouldn't be used in this file though. Hope it is useful!

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -351,7 +351,7 @@ class DisplayObject(object):
     def reload(self):
         """Reload the raw data from file or URL."""
         if self.filename is not None:
-            with open(self.filename, self._read_flags) as f:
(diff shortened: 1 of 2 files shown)


scarlehoff:
Also, I was wondering whether it would make sense to add a setter such that any `self.filename = 'some_string'` would result in a `Path` (so that if someone wants to extend this function don't have to remember that `self.filename` is a path) but felt too much of a change to do without asking.

scarlehoff:
Closing this as I haven't got time to look at the error and people seems quite interested in the pathlib change (due to hacktoberfest I guess) so better leave it free.

PR #12612, opened 2020-10-09
Use pathlib in packaging.py, and gen_latex_symbols.py

jpavlav:
* Added pathlib in a few places where it makes sense to 
Related to #12515

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -11,6 +11,7 @@
 import re
 import shlex
 import sys
+from Pathlib import Path
 
 from pathlib import Path
 from IPython.core.magic import Magics, magics_class, line_magic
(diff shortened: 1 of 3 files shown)


jpavlav:
Anything I need to do to trigger a new build?

jpavlav:
Need to run darker on it. Plus, just realized I may have messed up the spirit of the `temp_pyfile()` function. Will check if any of its callers actually close the open file handle.

jpavlav:
@Carreau Hey there! Sorry for all of the noise on this one. Had some issues with my local darker not catching some things that were caught on the Travis runs.

Carreau:
No worries, thanks for everything, there are conflict right now, if you are inclined to rebase that would be appreciated, otherwise i'll do it later, and if darker is too strict we can disable it.

PR #12614, opened 2020-10-10
Use pathlib in autogen_magics.py, autogen_config.py, and test_magic.py

joyceerhl:
Change 'with open' to use Pathlib instead.
See #12515.


Carreau:
Perfect thanks, you.

PR #12615, opened 2020-10-11
Pathlib in core.magics

MrMino:
See #12515.

This introduces usage of pathlib, including in places where it could be used instead of `os.path`, in:
- `IPython.core.magics.execution`
- `IPython.core.magics.packaging`

Don't be scared by the amount of edited lines. By the boy-scout rule I've fixed some issues with code style that my linter was yapping about: redundant or missing whitespace / lines, wrong indentation, unnecessary usage of `\`. These changes shouldn't alter any behavior. In `IPython.core.magics.execution` it turned out to be _a lot_ of hunks.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -40,6 +40,7 @@ from IPython.utils.timing import clock, clock2
 from warnings import warn
 from logging import error
 from io import StringIO
+from pathlib import Path
 
 if sys.version_info > (3,8):
     from ast import Module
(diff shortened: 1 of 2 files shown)


MrMino:
Turns out I have to move the code style changes out, because `darken`ing them is a whole other big can of worms. They now live [here](https://github.com/MrMino/ipython/tree/core_magics_style). Not sure if I should PR partial style changes.

Also, I decided not to add the following change, as it didn't visibly improve anything:

```diff
-            args = shellglob(map(os.path.expanduser,  arg_lst[1:]))
+            args = shellglob(map(lambda p: str(Path(p).expanduser()), arg_lst[1:]))
```

I wonder if `IPython.utils.path.shellglob` still adds value, given it was implemented back in 2012.

Carreau:
Don't worry f the code formatting is too-agressive, we can disable it. 

This looks great ! Thanks !

PR #12620, opened 2020-10-12
mktempfile returns Path object

agusgs:
Related with #12515 

I changed the mktempfile function in IPython/core/interactiveshell.py to return a Path object instead of a string.
Now I'm changing the callers of that function to handle that.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3597,10 +3597,10 @@ class InteractiveShell(SingletonConfigurable):
           - data(None): if data is given, it gets written out to the temp file
             immediately, and the file is closed again."""
 
-        dir_path = Path(tempfile.mkdtemp(prefix=prefix))
-        self.tempdirs.append(dir_path)
+        temp_directory = Path(tempfile.mkdtemp(prefix=prefix))
+        self.tempdirs.append(temp_directory)
 
-        handle, filename = tempfile.mkstemp(".py", prefix, dir=str(dir_path))
+        handle, filename = tempfile.mkstemp(".py", prefix, dir=temp_directory)
         os.close(handle)  # On Windows, there can only be one open handle on a file
 
         file_path = Path(filename)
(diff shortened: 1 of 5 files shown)


IPython/core/interactiveshell.py
```diff
@@ -3596,18 +3596,18 @@ def mktempfile(self, data=None, prefix='ipython_edit_'):
...
-        handle, filename = tempfile.mkstemp(".py", prefix, dir=str(dir_path))
-        os.close(handle)  # On Windows, there can only be one open handle on a file
+        temp_directory = Path(tempfile.mkdtemp(prefix=prefix))
+        self.tempdirs.append(temp_directory)
 
+        handle, filename = tempfile.mkstemp('.py', prefix, dir=temp_directory)
         file_path = Path(filename)
-        self.tempfiles.append(file_path)
+        os.close(handle)  # On Windows, there can only be one open handle on a file
+        self.tempfiles.append(file)
```
Carreau, line 3605:
```suggestion
        self.tempfiles.append(file_path)
```

Maybe ?

agusgs: Yes, I messed up when rebasing master :)

IPython/core/tests/test_interactiveshell.py
```diff
@@ -561,6 +561,20 @@ def test_exit_code_signal_csh(self):
             else:
                 del os.environ['SHELL']
 
+    def test_removes_temp_files_and_directories(self):
+        # First we ensure that we have created at least a tempfile
+        ip.mktempfile()
+
+        self.system("exit 0")
+        ip.atexit_operations()
+
```
agusgs, line 570: @Carreau I'm not sure how to test this case...
I want to be sure that every time a shell is closed we are removing the temp dirs/files.
Any thoughts?

Carreau: Apologies, I received many PR during hacktoberfest, I'll have a look.

agusgs: No problem :+1:

Carreau:

PR #12628, opened 2020-10-14
Change os.path functions to pathlib.Path functions in gh_api.py

hexxdump:
Signed-off-by: Hussaina Begum Nandyala <hexxdump@gmail.com>


Carreau:
Thanks !

PR #12629, opened 2020-10-15
Use pathlib in autogen_config

kugi83:
Completely replaced os with Path, like asked in #12515 
#12614 already changed the `with open()` to Path, and I went a bit deeper.

The new line 114 behaves a little bit differently than the older, the output file will have a newline after `app.document_config_options()`

### DIFF
--- a/docs/autogen_config.py
+++ b/docs/autogen_config.py
@@ -9,12 +9,12 @@ from traitlets import Undefined
 from collections import defaultdict
 
 here = Path(__file__).parent
-options = here / 'source' / 'config' / 'options'
-generated = options / 'config-generated.txt'
+options = here / "source" / "config" / "options"
+generated = options / "config-generated.txt"
 
 
 def indent(text, n):
-    return textwrap.indent(text, n*' ')
+    return textwrap.indent(text, n * " ")
 
 
 def interesting_default_value(dv):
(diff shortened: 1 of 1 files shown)


ChiefsBestPal:
**Linter effective**

kugiyasan:
@Carreau What is happening? I'm very new to PRs, so I don't understand what is going on with that force-push. A friend told me to pull the changes into my fork, but what should I do now?

Carreau:
> @Carreau What is happening? I'm very new to PRs, so I don't understand what is going on with that force-push. A friend told me to pull the changes into my fork, but what should I do now?

Sorry, I've been busy trying to review all the PRs, do not worry about the force-push, it's something that github warns about that it use to not to before. 

Usually when "pushing" you can only add new commits, when force-pushing you (and in the above case me) replaced an already existing commit. that happens for example when you rewrite git history if someone for example commited a password and you want to remove it.

docs/autogen_config.py
```diff
@@ -96,32 +98,39 @@ def reverse_aliases(app):
...
+    text.append(app.document_config_options())
+
+    text.extend(
+        class_config_rst_doc(c, trait_aliases) for c in app._classes_inc_parents()
+    )
 
-        for c in app._classes_inc_parents():
-            f.write(class_config_rst_doc(c, trait_aliases))
-            f.write('\n')
+    filename.write_text("\n".join(text))
```
Carreau, line 120: I'm not quite sure all those are correct, as `write` likely does not happend \n, but '\n'.join() does, so this will not be strictly identical.

kugiyasan: Alright, can you revert back that change?

Carreau: yes, I can do that.

PR #12633, opened 2020-10-16
Use pathlib in docs/autogen_shortcuts.py

SnirBroshi:
Converted `os.path` usage in `docs/autogen_shortcuts.py` to `pathlib`, and also improved the surrounding code a little.

As per #12515


Carreau:
Thanks

PR #12635, opened 2020-10-16
Use pathlib instead of os.path on test_module_paths.py

dswij:
Replacing os.path with pathlib for readability

see #12515


Carreau:
Thanks !

Issue #12595, opened 2020-10-05
%%bash magic shows no intermediate output


the21st:
see https://github.com/jupyter/notebook/issues/2023#issuecomment-272726680 for original issue in jupyter

Setup:

```
In [1]: %%bash
   ...: echo '1'
   ...: sleep 3
   ...: echo '2'
   ...: 
```

Expected behavior: output prints 1, than nothing for 3 seconds, then prints 2.
Actual behavior: output prints nothing for 3 seconds, than prints 1 and 2 at the same time.

jetpacktuxedo:
This isn't specific to bash, it appears to affect anything using `%%script` magic. For example, you can also replicate the issue using something like this:
```python
In [1]: %%script python
   ...: from time import sleep
   ...: for i in [1, 2, 3]:
   ...:   print(i)
   ...:   sleep(i)
   ...:
```

This appears to be related to the way `subprocess.popen` is being used here: https://github.com/ipython/ipython/blob/master/IPython/core/magics/script.py#L211. That `p.communicate` call blocks until the entire call has completed and then returns something like 
```python
(b'1\n2\n3\n', b'')
```
before the output is handled and returned.

jetpacktuxedo:
This should be resolved in the `master` branch now, so it should show up in the `8.X` series of releases :)

MrMino:
While this was resolved, there's still a similar issue with `%%bash --out` in #13499.

PR #12639, opened 2020-10-18
Line-buffered output for %%magic scripts

jetpacktuxedo:
This was done to address #12595

The following will now behave more like what you would expect:
```bash
In [1]: %%bash
   ...: for i in 1 2 3; do
   ...:   echo $i
   ...:   sleep $1
   ...: done
```

This doesn't actually _fully_ address it, as it is still line buffered. The following will still exhibit behavior more like before:
```bash
In [1]: %%bash
   ...: for i in 1 2 3; do
   ...:   printf $i
   ...:   sleep $1
   ...: done
```

I also seem to have some test failures right now related to the event loop being closed unexpectedly. Hopefully I can get those sorted shortly, but I wanted to go ahead and open this up to try to get some feedback on whether or not this is even worth pursuing.

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -8,7 +8,7 @@ import os
 import sys
 import signal
 import time
-from subprocess import Popen, PIPE, CalledProcessError
+import asyncio
 import atexit
 
 from IPython.core import magic_arguments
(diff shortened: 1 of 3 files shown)


IPython/core/magics/script.py
```diff
@@ -8,9 +8,15 @@
...
 import time
-from subprocess import Popen, PIPE, CalledProcessError
+import asyncio
 import atexit
 
+try:
+    from asyncio import create_task
+except ImportError:
+    # 3.6 had this function in a different module, this can go away when 3.6 is dropped
+    from AbstractEventLoop import create_task
```
jetpacktuxedo, line 18: 3.6 isn't supported beyond IPython 7.16 so I guess I can drop all of this? I'll have test failures in travis, though...

jetpacktuxedo:
Alright, I think this is totally ready now. The only travis failure is 3.6 which is no longer supported on this branch (should probably be dropped from travis...).

IPython/core/magics/script.py
```diff
@@ -229,20 +262,9 @@ def shebang(self, line, cell):
...
-            sys.stdout.write(out)
-            sys.stdout.flush()
-        if args.err:
-            self.shell.user_ns[args.err] = err
-        else:
-            sys.stderr.write(err)
-            sys.stderr.flush()
         if args.raise_error and p.returncode!=0:
-            raise CalledProcessError(p.returncode, cell, output=out, stderr=err)
+            print(p.returncode)
```
Carreau, line 266: missing debug ?

Carreau:
> Alright, I think this is totally ready now. The only travis failure is 3.6 which is no longer supported on this branch (should probably be dropped from travis...).

@SylvainCorlay I think I kept 3.6 testing for you and Xeus, do you think we can drop 3.6 now on master (7.x will still test on 3.6).

SylvainCorlay:
That would make sense to drop 3.6.

SylvainCorlay:
@johanmabille do you recall how to fix this MT/MD issue with gtest?

JohanMabille:
I think we replaced the flags with something like

```
set(CompilerFlags
        CMAKE_CXX_FLAGS
        CMAKE_CXX_FLAGS_DEBUG
        CMAKE_CXX_FLAGS_RELEASE
        CMAKE_C_FLAGS
        CMAKE_C_FLAGS_DEBUG
        CMAKE_C_FLAGS_RELEASE
        )
foreach(CompilerFlag ${CompilerFlags})
  string(REPLACE "/MD" "/MT" ${CompilerFlag} "${${CompilerFlag}}")
endforeach()
```

IPython/core/magics/script.py
jetpacktuxedo: Whoops, that might be something I flipped when I was debugging and forgot to out back. Lemme take a look.

SylvainCorlay:
Woops, mixed things up with another thread!

Carreau:
Thanks, no worry for the off-topic, I've hidden the messages.

Carreau:
I pushed a commit that remove testing on 3.6

jetpacktuxedo:
I rebased, but travis still seems to be running tests on 3.6 🤔

Carreau:
> I pushed a commit that remove testing on 3.6

ha, you seem to have overwritten the commit I pushed on this branch, I'll redo it.

Carreau:
I just pushed on your branch (this PR) 958d1d1 that removes testing on 3.6

jetpacktuxedo:
> > I pushed a commit that remove testing on 3.6
> ha, you seem to have overwritten the commit I pushed on this branch, I'll redo it.

Oh, yep, I probably did when I rebased 😅

Carreau:
let's try.

PR #12641, opened 2020-10-19
Use Pathlib in test_application.py and test_completerlib.py

### LINKED ISSUE #12515: Pathlib, Pathlib everywhere.

kir0ul:
Refactored `test_application.py` to use Pathlib.

Fix #12515.

### DIFF
--- a/IPython/core/tests/test_application.py
+++ b/IPython/core/tests/test_application.py
@@ -6,7 +6,6 @@ import tempfile
 from pathlib import Path
 
 import nose.tools as nt
-
 from traitlets import Unicode
 
 from IPython.core.application import BaseIPythonApplication
--- a/IPython/core/tests/test_application.py
+++ b/IPython/core/tests/test_application.py
@@ -36,9 +36,8 @@ def test_unicode_ipdir():
     ipdir = Path(tempfile.mkdtemp(suffix=u"€"))
     
     # Create the config file, so it tries to load it.
-    with ipdir.joinpath("ipython_config.py").open("w") as f:
-        pass
-    
+    ipdir.joinpath("ipython_config.py").touch()
+
     old_ipdir1 = os.environ.pop("IPYTHONDIR", None)
     old_ipdir2 = os.environ.pop("IPYTHON_DIR", None)
     os.environ["IPYTHONDIR"] = str(ipdir)
--- a/IPython/core/tests/test_application.py
+++ b/IPython/core/tests/test_application.py
@@ -3,6 +3,7 @@
 
 import os
 import tempfile
+from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


IPython/core/tests/test_completerlib.py
```diff
@@ -77,21 +77,22 @@ def test_completion_more_args(self):
...
         print(repr(event.line))
         match = set(magic_run_completer(None, event))
         # We specifically use replace here rather than normpath, because
         # at one point there were duplicates 'adir' and 'adir/', and normpath
         # would hide the failure for that.
-        self.assertEqual(match, {join(self.BASETESTDIR, f).replace('\\','/')
-                            for f in (u'a.py', u'aao.py', u'aao.txt', u'adir/')})
+        self.assertEqual(match, {str(self.BASETESTDIR.joinpath(f)) if Path(f).suffix
+                                 else str(self.BASETESTDIR.joinpath(f)) + os.sep
+                            for f in (u"a.py", u"aao.py", u"aao.txt", u"adir/")})
```
kir0ul, line 88: Not sure it's the best way to do this, it works but seems more complicated than before... :thinking:

kir0ul: Indeed it doesn't work on Windows... :sweat:

IPython/core/tests/test_application.py
```diff
@@ -33,15 +33,15 @@ def test_unicode_cwd():
 @dec.onlyif_unicode_paths
 def test_unicode_ipdir():
     """Check that IPython starts with non-ascii characters in the IP dir."""
-    ipdir = tempfile.mkdtemp(suffix=u"€")
+    ipdir = Path(tempfile.mkdtemp(suffix=u"€"))
     
     # Create the config file, so it tries to load it.
-    with open(os.path.join(ipdir, 'ipython_config.py'), "w") as f:
+    with ipdir.joinpath("ipython_config.py").open("w") as f:
         pass
```
Carreau, line 40: might be able to replace with something like https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

```diff
(ipdir / "ipython_config.py").touch()
```

kir0ul: Good catch!

Carreau:
Many thanks for working on this a few years ago; I was never able to keep up with IPython PRs, and this now conflicts, so I'm going to close.

PR #12642, opened 2020-10-19
use pathlib in test_extension, test_logger, test_application, test_oinspect, and test_profile

Creonalia:
replaces path strings with pathlib.Path objects as per [this issue](https://github.com/ipython/ipython/issues/12515)

### DIFF
--- a/IPython/core/tests/test_application.py
+++ b/IPython/core/tests/test_application.py
@@ -37,12 +37,12 @@ def test_unicode_ipdir():
     ipdir = tempfile.mkdtemp(suffix=u"€")
     
     # Create the config file, so it tries to load it.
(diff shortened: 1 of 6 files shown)


Creonalia:
I think the tests failed because I missed a string conversion; I'll fix that.

Carreau:
Thanks, you've done many files and this currently unfortunatly conflict with some tings that have already been merged in master, there are thus conflicts. Would you be able to solve them, or do you need me the fix those for you ?

Creonalia:
Conflicts should be fixed now, looks like it was just some different syntax.

Carreau:
Thanks for your work I was not able to keep up with PRs, I'm going to close as this now conflict, and I'm trying to reduce the numbers of PRs.

PR #12644, opened 2020-10-19
Update history.py to use pathlib

deep-jkl:
Tests were returning exeption in atexit_operations
AttributeError: 'str' object has no attribute 'unlink'

This commit embraces usage of pathlib in history.py and related
tests, fixing aforementioned problem.

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -6,7 +6,7 @@
 
 import atexit
 import datetime
-import os
+from pathlib import Path
 import re
 import sqlite3
 import threading
@@ -16,8 +16,8 @@ from decorator import decorator
 from IPython.utils.decorators import undoc
 from IPython.paths import locate_profile
 from traitlets import (
-    Any, Bool, Dict, Instance, Integer, List, Unicode, TraitError,
-    default, observe,
+    Any, Bool, Dict, Instance, Integer, List, Unicode, Union, TraitError,
+    default, observe
 )
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


IPython/core/history.py
```diff
@@ -73,32 +73,33 @@ def catch_corrupt_db(f, self, *a, **kw):
...
+                ext = self.hist_file.suffix
+                size = self.hist_file.stat().st_size
                 if size >= _SAVE_DB_SIZE:
                     # if there's significant content, avoid clobbering
                     now = datetime.datetime.now().isoformat().replace(':', '.')
                     newpath = base + '-corrupt-' + now + ext
                     # don't clobber previous corrupt backups
                     for i in range(100):
-                        if not os.path.isfile(newpath):
+                        if not Path(newpath).exists():
```
deep-jkl, line 87: I am not sure that `isfile` was an error or if it was inteded, I changed the logic to check existing file.

IPython/core/tests/test_history.py
```diff
@@ -29,8 +29,9 @@ def test_proper_default_encoding():
 def test_history():
     ip = get_ipython()
     with TemporaryDirectory() as tmpdir:
+        tmp_path = Path(tmpdir)
```
deep-jkl, line 32: This can be changed once Pathlib will be used in `TemporaryDirectory`

IPython/testing/tools.py
```diff
@@ -10,6 +10,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import os
+from pathlib import Path
```
deep-jkl, line 13: There are plenty of places, where os can be replaced, I want to keep this PR short, therefore I ignore them.

deep-jkl:
I am not sure if I can fix CI, but if you can ping me, I'll do my best.

IPython/core/history.py
Carreau: I think `exists()` is a good choice, and is likely what was intended.

PR #12645, opened 2020-10-19
Fix mistake where str is inserted into a list of Path.

Carreau:


deep-jkl:
Hi, I've noticed the problem too, I've reimplemented history to use pathlib, see https://github.com/ipython/ipython/pull/12644.

PR #12648, opened 2020-10-19
use pathlib in core/display.py and lib/display.py

sumagnadas:
Use pathlib as a replacement to os.path in display.py.

See #12515 .

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -56,7 +56,7 @@ def _safe_exists(path):
     """Check path, but don't let exceptions raise"""
     try:
         filepath = Path(path)
-        return filepath.exists() and str(filepath) != "."
+        return filepath.exists() and str(path) != ""
     except Exception:
         return False
 
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1086,7 +1086,10 @@ class Video(DisplayObject):
         if url is None and isinstance(data, str) and data.startswith(('http:', 'https:')):
             url = data
             data = None
-        elif data is not None and Path(data.decode() if isinstance(data, bytes) else data).exists():
+        elif (
+            data is not None
+            and Path(data.decode() if isinstance(data, bytes) else data).exists()
+        ):
             filename = data
             data = None
 
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -55,7 +55,8 @@ def __getattr__(name):
 def _safe_exists(path):
     """Check path, but don't let exceptions raise"""
     try:
-        return Path(path).exists()
+        filepath = Path(path)
+        return filepath.exists() and str(filepath) != "."
     except Exception:
         return False
 
@@ -1013,7 +1014,7 @@ class Image(DisplayObject):
 
     def _find_ext(self, s):
         path = Path(s)
-        base, ext = path.name.strip(path.suffix), path.suffix
+        base, ext = path.stem, path.suffix
 
         if not ext:
             return base
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1086,7 +1086,7 @@ class Video(DisplayObject):
         if url is None and isinstance(data, str) and data.startswith(('http:', 'https:')):
             url = data
             data = None
-        elif data is not None and Path(data).exists():
+        elif data is not None and Path(data.decode()).exists():
             filename = data
             data = None
 
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1014,7 +1014,7 @@ class Image(DisplayObject):
 
     def _find_ext(self, s):
         path = Path(s)
-        base, ext = path.stem, path.suffix
+        base, ext = str(path.parent / path.stem), path.suffix
 
         if not ext:
             return base
@@ -1131,8 +1131,7 @@ class Video(DisplayObject):
             if not mimetype:
                 mimetype, _ = mimetypes.guess_type(self.filename)
 
-            with Path(self.filename).open("rb") as f:
-                video = f.read()
+            video = Path(self.filename).read_bytes()
         else:
             video = self.data
         if isinstance(video, str):
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1126,7 +1126,7 @@ class Video(DisplayObject):
             if not mimetype:
                 mimetype, _ = mimetypes.guess_type(self.filename)
 
-            with Path(self.filename).open('rb') as f:
+            with Path(self.filename).open("rb") as f:
                 video = f.read()
         else:
             video = self.data
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -347,7 +347,7 @@ class DisplayObject(object):
     def reload(self):
         """Reload the raw data from file or URL."""
         if self.filename is not None:
-            with Path(self.filename).open(self._read_flags) as f:
+            with open(self.filename, self._read_flags) as f:
                 self.data = f.read()
         elif self.url is not None:
             # Deferred import
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1086,7 +1086,7 @@ class Video(DisplayObject):
         if url is None and isinstance(data, str) and data.startswith(('http:', 'https:')):
             url = data
             data = None
-        elif data is not None and Path(data.decode()).exists():
+        elif data is not None and Path(data.decode() if isinstance(data, bytes) else data).exists():
             filename = data
             data = None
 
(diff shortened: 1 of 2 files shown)


ashwinvis:
@sumagnadas There are too many conversions from string `path` to pathlib `Path` in this PR. See if it is possible to set it at class level as `self.path = Path(path)`; if not do it only once within a method.

sumagnadas:
Will try to do that. But some questions come to my mind, are the two files
connected and is core/display.py the main file between these two?

ashwinvis:
Yes they are related. The public API [[docs]](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#module-IPython.display) can be found here:
https://github.com/ipython/ipython/blob/master/IPython/display.py

sumagnadas:
I am getting a problem where using `Path('').exists()` gives `True` and filename is changing from `None` to `''`. If you understood what I meant please tell me what should i do about it?

ashwinvis:
`Path('')` would always point to the current directory.

```python

In [2]: from pathlib import Path

In [3]: Path('')
Out[3]: PosixPath('.')

In [4]: Path('').resolve()
Out[4]: PosixPath('/home/avmo')

In [6]: Path.cwd()
Out[6]: PosixPath('/home/avmo')
```
Hope that helps.

sumagnadas:
Can you check the new changes and tell me if it is alright now or do i need
to change some more things?

IPython/core/display.py
```diff
@@ -1013,7 +1013,8 @@ def _repr_jpeg_(self):
             return self._data_and_metadata()
 
     def _find_ext(self, s):
-        base, ext = splitext(s)
+        path = Path(s)
+        base, ext = path.stem, path.suffix
```
ashwinvis, line 1017: This is not equivalent. `.stem` contains the filename alone, and strips the path to it.
```py
In [11]: a = Path.home() / "README.md"

In [17]: a
Out[17]: PosixPath('/home/avmo/README.md')

In [18]: splitext(a)
Out[18]: ('/home/avmo/README', '.md')

In [19]: a.parent / a.stem, a.suffix
Out[19]: (PosixPath('/home/avmo/README'), '.md')
```

IPython/core/display.py
```diff
@@ -1127,7 +1131,7 @@ def _repr_html_(self):
             if not mimetype:
                 mimetype, _ = mimetypes.guess_type(self.filename)
 
-            with open(self.filename, 'rb') as f:
+            with Path(self.filename).open("rb") as f:
                 video = f.read()
```
ashwinvis, line 1135:
```suggestion
            video = Path(self.filename).read_bytes()
```

IPython/lib/display.py
```diff
@@ -412,7 +412,7 @@ def _repr_html_(self):
     def __repr__(self):
         """return absolute path to file
         """
-        return abspath(self.path)
+        return str(Path(self.path))
```
ashwinvis, line 415:
```suggestion
        return str(Path(self.path).resolve())
```

IPython/lib/display.py
```diff
@@ -383,7 +383,7 @@ def __init__(self,
         result_html_suffix : str
             text to append at the end of link [default: '<br>']
         """
-        if isdir(path):
+        if Path(path).is_dir():
             raise ValueError("Cannot display a directory using FileLink. "
               "Use FileLinks to display '%s'." % path)
         self.path = fsdecode(path)
```
ashwinvis, line 389: Would it work if we initialize `self.path` as a `Path` object?

IPython/lib/display.py
```diff
@@ -520,10 +520,11 @@ def f(dirname, fnames, included_suffixes=None):
             # are going to be displayed
             display_fnames = []
             for fname in fnames:
-                if (isfile(join(dirname,fname)) and
-                       (included_suffixes is None or
-                        splitext(fname)[1] in included_suffixes)):
-                      display_fnames.append(fname)
+                fname = Path(fname)
+                if Path(dirname.joinpath(fname)).is_file() and (
```
ashwinvis, line 524:
```suggestion
                if (dirname / fname).is_file() and (
```

ashwinvis (changes requested):

sumagnadas:
Will do the suggested changes.

IPython/core/display.py
ashwinvis:
```suggestion
        base, ext = str(path.parent / path.stem), path.suffix
```

IPython/lib/display.py
sumagnadas: I will try to make it work if i can.

sumagnadas:
Don't know why but the test on MacOS is failing. Can you tell me some ways i can try to fix them?

ashwinvis:
Sorry, but [this beats me](https://travis-ci.org/github/ipython/ipython/jobs/738609718#L552). It seems that the temporary file generated by the `tempfile` stdlib gets a strange path in Travis macOS env.

IPython/lib/display.py
```diff
@@ -383,36 +383,41 @@ def __init__(self,
         result_html_suffix : str
             text to append at the end of link [default: '<br>']
         """
-        if isdir(path):
+        if Path(path).is_dir():
             raise ValueError("Cannot display a directory using FileLink. "
               "Use FileLinks to display '%s'." % path)
-        self.path = fsdecode(path)
+        self.path = Path(path)
```
ashwinvis, line 389: @sumagnadas could you try `self.path = Path(fsdecode(path)) instead?

sumagnadas: That can be done.

sumagnadas:
What should i do about the MacOS build?

ashwinvis (commented):
About the macOS build, you could try reverting the change based on the suggestion (https://github.com/ipython/ipython/pull/12648#discussion_r511489267). See if it helps.

This is another minor thing I noticed...

IPython/core/display.py
```diff
@@ -56,7 +55,8 @@ def __getattr__(name):
 def _safe_exists(path):
     """Check path, but don't let exceptions raise"""
     try:
-        return os.path.exists(path)
+        filepath = Path(path)
+        return filepath.exists() and str(filepath) != "."
```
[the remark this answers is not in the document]
sumagnadas: let me try.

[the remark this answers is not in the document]
sumagnadas: I added the change but it doesn't fix the build.
The problem is that the path gets the repr provides the prefix of `/private/`(i dont know where it gets that and I also dont know if the tempfile is actually in the `/private` direcctory) whereas the actual path to the tempfile isn't inside `/private`.

IPython/lib/display.py
```diff
@@ -383,36 +383,41 @@ def __init__(self,
         result_html_suffix : str
             text to append at the end of link [default: '<br>']
         """
-        if isdir(path):
+        if Path(path).is_dir():
             raise ValueError("Cannot display a directory using FileLink. "
               "Use FileLinks to display '%s'." % path)
-        self.path = fsdecode(path)
+        self.path = Path(fsdecode(path))
```
sumagnadas, line 389: This is where the problem starts. The path which is sent is in the form `/var/T/.........` whereas `self.path` gets or changes to the form `/private/var/T/......` in the build.

ashwinvis: That's my best guess too. Could you try reverting to the old version so we can confirm that?

sumagnadas: I am bringing back the `abspath` method for now. Will try to remove it later.

sumagnadas: The [MacOS build](https://travis-ci.com/github/ipython/ipython/jobs/415283806) actually didn't fail after reverting to the old version.

sumagnadas: `Path.absolute` works great in place of `abspath`.

sumagnadas: So i looked up a little and found out that `resolve()` not only normalizes the path but also resolves the symlinks, if any(hence the name) which causes the problem. So should i change the test, which fails because of the "resolve symlinks" part, to
```
expected = str(pathlib.Path(tf.name).resolve())
```
?

ashwinvis: Great detective work!
> But one problem of absolute() is that it doesn't normalize the path. So should i change the test, which fails because of the "resolve symlinks" part
I will leave that decision to the maintainers

sumagnadas: > So i looked up a little and found out that `resolve()` not only normalizes the path but also resolves the symlinks, if any(hence the name) which causes the problem. But one problem of `absolute()` is that it doesn't normalize the path. So should i change the test, which fails because of the "resolve symlinks" part, to
> ```
> expected = str(pathlib.Path(tf.name).resolve())
> ```
> ?

@Carreau should I do this?

sumagnadas:
Sorry for the delay. I was busy with my studies. 

I will change the test then. If one of the maintainers tell me to remove the changes made to the test, I will remove them

PR #12675, opened 2020-10-25
use pathlib in `setup.py`

baogianghoangvu:
Hi!

I took a jab at #12515, using pathlib in `setup.py`.

This simply includes changing an `os.remove` to `pathlib.Path.unlink` ~which conveniently allows skipping the `os.path.exists` check altogether~ Edit: `Path.unlink(missing_ok: bool)` is only available from Python 3.8.

There're unused aliases to `os.path.isfile` & `os.path.join`, which I remove.

Hope this doesn't take much of your time to review.

Thanks!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ requires utilities which are not available under Windows."""
 
 import os
 import sys
+from pathlib import Path
 
 # **Python version check**
 #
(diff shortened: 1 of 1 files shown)


baogianghoangvu:
It turns out the param `Path.unlink(missing_ok: bool)` was added a bit later (only since python3.8). https://github.com/python/cpython/commit/d9e006bcefe6fac859b1b5d741725b9a91991044.

I'll restore the "exists" check before unlinking to resolve this problem for python 3.6 and 3.7  builds.

Carreau:
Thanks !

PR #12683, opened 2020-10-28
FIX: Typing annotations

farisachugthai:
Running `mypy -p IPython.sphinxext` raised a few errors. Fix them and strip some trailing whitespace from a few places.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1362,9 +1362,9 @@ def match_dict_keys(
         # All checks passed!
         return True
 
-    filtered_key_is_final: Dict[
-        Union[str, bytes, int, float], _DictKeyState
-    ] = defaultdict(lambda: _DictKeyState.BASELINE)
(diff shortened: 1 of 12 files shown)


farisachugthai:
Sorry I got messed up by darker again. What is a good way of running and/or fixing this in advance? You won't know what the SHA for a commit is until after you do it, so running `darker -r <SHA>` like how it's run in the tests doesn't seem like it'd work.

Carreau:
You can either run just `darker` before comiting or, run `darker -r 60625f241f298b5039cb2debc365db38aa7bb522` always with this hash for IPython, Otherwise don't worry I'll run it.

farisachugthai:
Okay thank you! I'll make sure I do so because I've ran into this a few times and I don't wanna make extra work for you haha

Carreau:
It took w while but thanks, merging.

PR #12684, opened 2020-10-28
DOC: Remove sphinxext errors

farisachugthai:
The doc for the sphinx extension says the sentence "list directives
here" but the directives are now listed in full above.

In addition, the word "with" was repeated twice in a row but probably
went unnoticed since it happens at a line break.


Carreau:
Thanks !

PR #12686, opened 2020-10-28
Run mypy on more files.

Carreau:


farisachugthai:
Wouldn't it be arguably easier to do:

`mypy -p IPython.core.magics`

here and then add the single line of code

`mypy -p IPython.terminal`

once those last 2 files are fixed? It seems like that would allow for less copy pasting then manually listing files.

Carreau:
> once those last 2 files are fixed? It seems like that would allow for less copy pasting then manually listing files.

Yes likely; in the end we likely just want `mypy -p IPython`

PR #12690, opened 2020-10-31
validate ESC_PAREN ('/') is followed by a callable name and not empty

guybortnikov:
When running a single python expression with just a slash - `/`, a tuple is returned.

```python
In [1]: /
Out[1]: ()
```

This weird case happens because when a line starts with `/` the `inputtransformer` transforms it into a call of the first word after the `/` as the callable name and rest tokens after as arguments. 

This PR fixes that issue by validating that at least a callable name is given and it's not empty, if not a `SyntaxError` will be raised.

```python
In [1]: /
  File "<ipython-input-1-8f27084b6294>", line 1
    /
    ^
SyntaxError: invalid syntax


In [2]:
```

Validated that tests are passing.

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -395,7 +395,10 @@ def _tr_quote2(content):
 
 def _tr_paren(content):
     "Translate lines escaped with a slash: /"
-    name, _, args = content.partition(' ')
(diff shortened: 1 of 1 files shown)


guybortnikov:
@Carreau can you please look at this PR?

Issue #12107, opened 2020-01-31
Magics.parse_options parse problem


louismartin:
Hi,
I think there is a bug with the `parse_options` method in the sense that it modifies the input statement which causes a bug in my application.

## Repro
In jupyter notebook, when iterating over my dataframe, I tried to time the operation with the `timeit` magic:
```
> df = pd.DataFrame({'document': ['First doc.', 'Second doc.']})
> %timeit _ = [s.count('. ') for s in df['document'].tolist()]
KeyError: ' document'
```
Note the key that was transformed from `'document'` in my code to `' document'` at execution time.

This happens in the `parse_options` method https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/core/magics/execution.py#L1077 
```
        opts, stmt = self.parse_options(line,'n:r:tcp:qo',
                                        posix=False, strict=False)
```
When dropping into pdb, I get the following:
```
> print(line)
_ = [s.count('. ') for s in df['document'].tolist()]
> print(stmt)
_ = [s.count('. ') for s in df[' document'].tolist()]
```

and within `parse_options()`, I guess it happens when [splitting the line with `arg_split`](https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/core/magic.py#L627) and then [joining back with spaces](https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/core/magic.py#L652)
```
> argv = arg_split(arg_str, posix, strict)
> print(arg_str)
_ = [s.count('. ') for s in df['document'].tolist()]
> print(argv)
['_', '=', "[s.count('.", "') for s in df['", "document'].tolist()]"]
```
And then`args = ' '.join(args)` will join with an extra space.

Also changing my command from `_ = [s.count('. ') for s in df['document'].tolist()]` to `_ = [s for s in df['document'].tolist()]` removes the problem so there might be a problem with the `.count('. ')`.

Carreau:
Hum, thanks, that seem legit. This code is really nasty; and it might take some time to fix. 
Pull requests appreciated.

adityausathe:
#12708 
I have raised a PR to address this issue. @Carreau please review.

ivanov:
Hey there, I'm going through old issues and it seems to me that it makes sense to close this one.

[...]

Issue #12225, opened 2020-04-09
%load magic fails when calling a function with multiple keyword arguments and double-quotes


sedoris:
This is a very specific bug, apologies if it's been reported elsewhere and I just couldn't find it. 

To reproduce in a Jupyter notebook:

```
def example_fxn(first_name, last_name=""):
    return "Hello {} {}!".format(first_name, last_name)
```
This works: 
```
%load example_fxn(first_name='John', last_name='Smith')
```
This also works: 
```
%load example_fxn(first_name="John")
```
But this fails:
```
%load example_fxn(first_name="John", last_name="Smith")
```

Traceback (with personal details blocked out)
```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
~\AppData\Local\Continuum\miniconda3\envs\redacted\lib\site-packages\IPython\core\interactiveshell.py in find_user_code(self, target, raw, py_only, skip_encoding_cookie, search_ns)
   3649         try:                                              # User namespace
-> 3650             codeobj = eval(target, self.user_ns)
   3651         except Exception:

<string> in <module>

NameError: name 'Smith' is not defined

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-23-2728f334f028> in <module>
----> 1 get_ipython().run_line_magic('load', 'example_fxn(first_name="John", last_name="Smith")')

~\AppData\Local\Continuum\miniconda3\envs\redacted\lib\site-packages\IPython\core\interactiveshell.py in run_line_magic(self, magic_name, line, _stack_depth)
   2315                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2316             with self.builtin_trap:
-> 2317                 result = fn(*args, **kwargs)
   2318             return result
   2319 

<decorator-gen-47> in load(self, arg_s)

~\AppData\Local\Continuum\miniconda3\envs\redacted\lib\site-packages\IPython\core\magic.py in <lambda>(f, *a, **k)
    185     # but it's overkill for just that one bit of state.
    186     def magic_deco(arg):
--> 187         call = lambda f, *a, **k: f(*a, **k)
    188 
    189         if callable(arg):

~\AppData\Local\Continuum\miniconda3\envs\redacted\lib\site-packages\IPython\core\magics\code.py in load(self, arg_s)
    331         search_ns = 'n' in opts
    332 
--> 333         contents = self.shell.find_user_code(args, search_ns=search_ns)
    334 
[...]

PR #12708, opened 2020-11-21
Issue #12107 fix: additional spaces while parsing timeit-magic options

adityausathe:
The timeit magic needs to analyze its args to search for any of its supported options (-n/-r etc).
While doing so, the current implementation tokenizes whole args-string to look for the options. Though we should be only searching for the options in the prefix part of the args-string.
The full tokenization and splicing of the remainder string is changing the executable statement syntactically and/or semantically.
This fix preserves the original args-string and only removes the prefixed options, thus eliminating the problems associated with splicing.

### DIFF
--- a/IPython/core/magic.py
+++ b/IPython/core/magic.py
@@ -663,7 +663,7 @@ class Magics(Configurable):
             if preserve_non_opts:
                 args = remainder_arg_str.lstrip()
             else:
-                args = ' '.join(args)
+                args = " ".join(args)
 
         return opts,args
 
--- a/IPython/core/magic.py
+++ b/IPython/core/magic.py
@@ -618,6 +618,9 @@ class Magics(Configurable):
         posix = kw.get('posix', os.name == 'posix')
         strict = kw.get('strict', True)
 
+        preserve_non_opts = kw.get('preserve_non_opts', False)
+        remainder_arg_str = arg_str
+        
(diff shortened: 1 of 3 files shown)


adityausathe:
@Carreau I checked, this resolves both #12107 and #12225

Carreau:
Thanks ! 

Apologies for the delay in reviewing, let's get that in master to be tested by more people.

IPython/core/magic.py
```diff
@@ -629,10 +632,18 @@ def parse_options(self, arg_str, opt_str, *long_opts, **kw):
...
             except GetoptError as e:
-                raise UsageError('%s ( allowed: "%s" %s)' % (e.msg,opt_str,
-                                        " ".join(long_opts))) from e
-            for o,a in opts:
-                if o.startswith('--'):
+                raise UsageError(
+                    '%s ( allowed: "%s" %s)' % (e.msg, opt_str, " ".join(long_opts))
+                ) from e
+            for o, a in opts:
+                if mode is "string" and preserve_non_opts:
```
Carreau, line 639: `mode is "string"` emit a syntax warning, will need to be change to `mode == "string"` in a seubsequent PR

Carreau: see #12920.

adityausathe:
Glad to see this getting merged! Thanks

PR #12718, opened 2020-12-04
Mare reset -f more agressive and cull sys.modules;

Carreau:
This _should_ allow soft restart of sessions w/o killing the process


biblicabeebli:
Hi, coming here from this issue I reported.

https://github.com/ipython/ipython/issues/12753

Looks like this PR / a commit in this PR fixes an issue a couple people/projects are hitting that results in big pauses in, or outright crashes of, the ipython shell.

I can see that this code interacts with something about reloading the sys library, but it is well over my head.

Could I get some pointers or hypotheticals of how this can be triggered so that I can track it down in my project and squash it?

PR #12723, opened 2020-12-07
Add support for Jedi's fuzzy completion

### LINKED ISSUE #12673: Using jedi fuzzy completions
Jedi recently introduced fuzzy completions (so for example `art` can match
`abort`), and
[I'm trying to use it in IPython](https://github.com/davidhalter/jedi/issues/1687).
I was thinking about:

- Making fuzzy off by default
- Adding a configuration option to enable fuzzy completion
- Adding the code to IPython/core/completer.py

Is it OK if I send a PR with these changes?

Thanks!

infokiller:
closes #12673

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -136,6 +136,7 @@ from IPython.utils import generics
 from IPython.utils.dir2 import dir2, get_real_method
 from IPython.utils.path import ensure_dir_exists
 from IPython.utils.process import arg_split
+from IPython.utils.version import check_version
 from traitlets import Bool, Enum, Int, List as ListTrait, Unicode, default, observe
 from traitlets.config.configurable import Configurable
 
@@ -590,6 +591,10 @@ class Completer(Configurable):
         Set to 0 to stop computing types. Non-zero value lower than 100ms may hurt
         performance by preventing jedi to build its cache.
         """).tag(config=True)
+
+    jedi_fuzzy = Bool(
+        default_value=False,
+        help='Experimental: enable fuzzy completion in Jedi').tag(config=True)
 
     debug = Bool(default_value=False,
                  help='Enable debug for the Completer. Mostly print extra '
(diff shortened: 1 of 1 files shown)


ivanov:
For starters,  I'm just going to power-cycle this PR real quick since it predates the move away from Travis, so all the tests can run to completion.

ivanov:
Thanks again for taking the time to make this happen, @infokiller . I agree with your choice of making fuzzy off by default as the right call.


I think to get this in we would want to
- [ ] add tests (preferably testing that fuzzy matches work when enabled and don't when disabled)
- [ ] add documentation for this new configurable (to give it a signal boost as an option) - you can do this by adding an .rst file in the `docs/source/whatsnew/pr` directory. There's a README there with more instructions.
- [ ] ensure that either the minimal version of jedi that IPython depends on supports the new `get_completion_prefix_length` method you added, or add some checks and backwards-compatible workarounds to retain the old behavior.
- [ ] (optional) add docs for how to switch this dynamically during a session. That way users can start playing around with it right away and toggle it if it proves disruptive on specific variables and available namespace completion combinations.

infokiller:
Many thanks for your review and detailed instructions @ivanov
I will address your comments as soon as I have time to work on the PR (I believe it will take a few days) and will ping you back.

infokiller:
@ivanov just a quick question: Is checking the jedi version before calling `get_completion_prefix_length` ok? (like in https://github.com/ipython/ipython/pull/12723/files#diff-f10e4b6051cca1d8ce27936518d6908de43b5fd861902cd3923a4342e8d4fc5aR1462) Would it be better to check for the availability of the `get_completion_prefix_length` function instead?

Carreau:
Personally I have a slight preference to check for version; it is easier to remove code when we drop support for old version, otherwise a hasattr or getattr might stay forever even if not necessary.

ivanov:
> Personally I have a slight preference to check for version; it is easier to remove code when we drop support for old version, otherwise a hasattr or getattr might stay forever even if not necessary.

This makes sense to me., so I agree with that path forward.  In the unlikely event that jedi drops or changes the name of `get_completion_prefix_length` in the future, one would hope that they'd start by adding deprecation warning that would allow us to patch things up, if necessary.

The other minor thing that you may have spotted that changed since you first opened this is that `darker` is being used for formatting, so there's some minor changes that are needed for the top-most "Python package / build (3.8) (pull_request" check to pass.

We can safely ignore the Travis CI failure, and I think the Appveyor one is just intermittent, so hopefully that one clears up and passes when you have the chance to update this PR.

MrMino:
@infokiller If you need help with making the `darker` check pass - just ask :slightly_smiling_face:.

Carreau:

PR #12727, opened 2020-12-14
Update some docstring to conform to numpy docstring format

Carreau:


IPython/core/display.py
```diff
@@ -851,18 +851,23 @@ def __init__(self, data=None, url=None, filename=None, format=None,
...
+        a URL, or a filename from which to load image data.
+        The result is always embedding image data for inline images.
+
+        >>> Image('http://www.google.fr/images/srpr/logo3w.png')
+
+        >>> Image('/path/to/image.jpg')
+        
+        >>> Image(b'RAW_PNG_DATA...')
+
+        Specifying Image(url=...) does not embed the image data,
```
farisachugthai, line 866: Hey so not a contributor or anything so sorry for the drive-by comment, but wouldn't this fail if it were running as `python -m doctest display.py` or `sphinx-build -b doctest` in the docs folder?

I think you have to import the Image class here. I'll admit that that's odd because importing something into it's own docstring feels a bit pointless but I believe I've had doctests fail from this. I could be wrong so I'm sorry if that's the case.

Carreau: apologies for the delay in responding. 
Yest, That is a good question; so we should likely mark this as "check the syntax, but do not execute", and there should be some way to have doctest import the function with the correct nape in the namespace indeed. 

It's a difficult problem as you do not want the docstring to become too hard to read.

IPython/core/display.py
```diff
@@ -851,18 +851,23 @@ def __init__(self, data=None, url=None, filename=None, format=None,
...
-        # it only generates `<img>` tag with a link to the source.
-        # This will not work in the qtconsole or offline.
-        Image(url='http://www.google.fr/images/srpr/logo3w.png')
+
+        embedded image data, works in qtconsole and notebook
+        when passed positionally, the first arg can be any of raw image data,
+        a URL, or a filename from which to load image data.
+        The result is always embedding image data for inline images.
+
+        >>> Image('http://www.google.fr/images/srpr/logo3w.png')
```
Carreau, line 860:
```suggestion
        >>> Image('http://www.google.fr/images/srpr/logo3w.png')
        <IPython.core.display.Image object>
```

IPython/core/display.py
```diff
@@ -851,18 +851,23 @@ def __init__(self, data=None, url=None, filename=None, format=None,
...
+
+        >>> Image('/path/to/image.jpg')
```
Carreau, line 862:
```suggestion
        >>> Image('/path/to/image.jpg')
        <IPython.core.display.Image object>
```

IPython/core/display.py
```diff
@@ -851,18 +851,23 @@ def __init__(self, data=None, url=None, filename=None, format=None,
...
+        
+        >>> Image(b'RAW_PNG_DATA...')
```
Carreau, line 864:
```suggestion
        >>> Image(b'RAW_PNG_DATA...')
        <IPython.core.display.Image object>
```

IPython/core/display.py
Carreau: It actually seem to import `Image as Image` when doing the doctest; and it was actually failing as the output was not present.

Carreau:
Failure unrelated, merging. Thanks !

PR #12733, opened 2020-12-17
Feature autoreload 3

skalaydzhiyski:
The new option will allow users to autoload everything from a
module/script without having to reimport names. Also, if using
option 3, you will be able to autoreload/autoload global vars
of built-in types.

The way I personally use this feature is - I have two tmux screens open, one with vim and one with IPython. I code in vim and if I need to quickly experiment with something I jump to IPython and dig in using the interpreter which already has the full context and state of my script autoloaded.

For more information and a clearer example - please see the unit test I have added and the update of whatsnew.

P.S.
I am open for refactoring ideas for the check at extensions/autoload.py#400-407


skalaydzhiyski:
closing as duplicate of [https://github.com/ipython/ipython/pull/12734](url)

PR #12734, opened 2020-12-17
Feature add autoreload option 3

skalaydzhiyski:
This feature will allow us to synchonise a script/module with an open IPython session. 

Currently:
  With the option %autoreload 2 , we can only update the state of existing objects loaded from a module.
Now:
  We will be able to automatically load newly added objects to a module without us having to re-run our import statements.
  Also we can load and autoreload global vars of built-in types, which wasn't possible with previous versions of the autoreload extension.

Example Usage:
 
1. Make sure you load the extension with the new option 3
image

2. We have a **mod.py** (currently empty)
image
image

3. We create a bunch of objects inside **mod.py**
image

4. Then all of them automatically appear in the local context of our open session
image

5. We change the values of the loaded objects
image

6. And we can see they are updated automatically in the session, keeping the existing functionality:
image

For an example run, please see the unit test I have added as part of the PR.


MrMino:
Is this a duplicate of #12733? If so, could you please close one one of these PRs?

skalaydzhiyski:
> Is this a duplicate of #12733? If so, could you please close one one of these PRs?
Yes, it was a duplicate, thanks for that. Just deleted it.

Carreau:
Apologies for the delay in reviewing.

That looks good, let try to get that in master.

Thanks both.

PR #12736, opened 2020-12-21
Global options and local noflags for sphinx ipython directive

### LINKED ISSUE #12735: Global options for the IPython Directive
At the moment the IPython directive has a few flags and decorators that can only be set individually at every block.
This in some cases makes the rst files unnecessarily verbose and makes quick behavioral changes hard.

A nice way around this would be to have:
1. global options in the `conf` file such as `ipython_verbatim` or `ipython_okwarning`.
2. If (1) is done, have flags to invert the global behavior in individual blocks, such as `:noverbatim:` or `:nookwarning:`

Here are two example scenarios in which these features would be helpful:

1. The code starts issuing temporary future warnings (maybe due to a dependency changing API soon), which now need to be ignored, and which will need to stop being ignored a few weeks later after the dependency is updated. Now the only way to do so is to add the `:okwarning:` at every `ipython` directive block and later on remove them. Adding a single global `ipython_okwarning = True` would be me much more convenient.
2. Docs need to be repeatedly built while editing some rst files and there are several other rst files which take a while to execute their ipython code blocks. Being able to temporarily make everything `:verbatim:` (and maybe make the blocks being edited `:noverbatim:`) would make this process smoother.

So the full proposal would be:

- Add global alternatives for all the local option flags: `verbatim`, `okwarning`, `okexcept`, `doctest`, `suppress`
- Add inverse versions of the local flags to escape the global options: `noverbatim`, `nookwarning`, `noexcept`, `nodoctest`, `nosuppress`

csala:
Resolve #12735

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -884,13 +884,19 @@ class IPythonDirective(Directive):
     required_arguments = 0
     optional_arguments = 4 # python, suppress, verbatim, doctest
     final_argumuent_whitespace = True
-    option_spec = { 'python': directives.unchanged,
-                    'suppress' : directives.flag,
-                    'verbatim' : directives.flag,
-                    'doctest' : directives.flag,
-                    'okexcept': directives.flag,
-                    'okwarning': directives.flag
-                  }
+    option_spec = {
+        'python': directives.unchanged,
+        'suppress': directives.flag,
+        'verbatim': directives.flag,
+        'doctest': directives.flag,
+        'okexcept': directives.flag,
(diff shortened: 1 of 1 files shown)


csala:
Gave it a first shot here for initial review.
Feedback about whether the proposed approach is valid, as well as pointers to what other things need to be tweaked (documentation, change logs, etc) to call this ready will be appreciated.

csala:
@Carreau Mentioning you to draw your attention to this: Any chance I could get some feedback about it before the project deviates too far from the current state and this needs to be re-implemented?

Carreau:

PR #12739, opened 2020-12-25
BUG: The `%conda` magic doesn't work

farisachugthai:
Due to moving the repo away from os.path to pathlib, the wrong method is called on a Path object.

In addition, no checks are provided when trying to slice the args objects that the user provides. As a result, if they don't provide enough arguments (which can happen in normal use very frequently), running the magic  raises.

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -82,8 +82,9 @@ class PackagingMagics(Magics):
 
         conda = _get_conda_executable()
         args = shlex.split(line)
-        command = args[0]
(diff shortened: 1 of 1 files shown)


farisachugthai:
Just wanna note the only test that seems to have failed, did so because SQLite attempted to operate on a readonly database?

Carreau:
> Just wanna note the only test that seems to have failed, did so because SQLite attempted to operate on a readonly database?

that's unrelated, it's a race condition on travis. Thanks.

Issue #12745, opened 2020-12-28
NameError: name 'sys' is not defined


36rahu:
Got the below error when I use this code snipets,

```
In [1]: from elasticsearch import Elasticsearch
In [2]: es = Elasticsearch(hosts=[{"host": "host.docker.internal", "port": 9200}],
   ...:                     max_retries=30,
   ...:                     retry_on_timeout=True,
   ...:                     request_timeout=30)
In [3]: es.info()
```

Error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/IPython/terminal/ptutils.py", line 113, in get_completions
    yield from self._get_completions(body, offset, cursor_position, self.ipy_completer)
  File "/usr/local/lib/python3.8/site-packages/IPython/terminal/ptutils.py", line 129, in _get_completions
    for c in completions:
  File "/usr/local/lib/python3.8/site-packages/IPython/core/completer.py", line 438, in _deduplicate_completions
    completions = list(completions)
  File "/usr/local/lib/python3.8/site-packages/IPython/core/completer.py", line 1818, in completions
    for c in self._completions(text, offset, _timeout=self.jedi_compute_type_timeout/1000):
  File "/usr/local/lib/python3.8/site-packages/IPython/core/completer.py", line 1861, in _completions
    matched_text, matches, matches_origin, jedi_matches = self._complete(
  File "/usr/local/lib/python3.8/site-packages/IPython/core/completer.py", line 2029, in _complete
    completions = self._jedi_matches(
  File "/usr/local/lib/python3.8/site-packages/IPython/core/completer.py", line 1373, in _jedi_matches
    interpreter = jedi.Interpreter(
  File "/usr/local/lib/python3.8/site-packages/jedi/api/__init__.py", line 725, in __init__
    super().__init__(code, environment=environment,
TypeError: __init__() got an unexpected keyword argument 'column'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "manage.py", line 22, in <module>
    main()
  File "manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
[...]

PR #12746, opened 2020-12-28
Pin jedi dependency #12745

yurzo:
Jedi's Xmass release breaks tab completion

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -184,7 +184,7 @@ extras_require = dict(
 
 install_requires = [
     'setuptools>=18.5',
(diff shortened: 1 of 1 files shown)


jspricke:
I guess you want to target the 7.x branch with this as it is not an issue in master.

yurzo:
> I guess you want to target the 7.x branch with this as it is not an issue in master.

I think so too.

yurzo:
@jspricke: Should I do 7.10.x as well? I'm unfamiliar with this project's policies.

Issue #12740, opened 2020-12-25
Last jedi release (0.18.0) is incompatible with ipython (7.19 and 7.18 tested); reason - column arg was deprecated, and now removed


DaniilAnichin:
Relevant traceback reads as follows: 
```
  File "../venv/lib/python3.8/site-packages/IPython/core/completer.py", line 2029, in _complete
    completions = self._jedi_matches(
  File "../venv/lib/python3.8/site-packages/IPython/core/completer.py", line 1373, in _jedi_matches
    interpreter = jedi.Interpreter(
  File "../venv/lib/python3.8/site-packages/jedi/api/__init__.py", line 725, in __init__
    super().__init__(code, environment=environment,
TypeError: __init__() got an unexpected keyword argument 'column'
```

sys info: 
```
{'commit_hash': '62779a198',
 'commit_source': 'installation',
 'default_encoding': 'utf-8',
 'ipython_path': '../venv/lib/python3.8/site-packages/IPython',
 'ipython_version': '7.18.0',
 'os_name': 'posix',
 'platform': 'Linux-4.15.0-128-generic-x86_64-with-glibc2.17',
 'sys_executable': '../venv/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.8.5 (default, Jul 20 2020, 19:50:14) \n[GCC 5.4.0 20160609]'}
```
same reported in jedi repo too

benwu95:
https://github.com/ipython/ipython/commit/dcd9dc90aee7e4c5c52ce44c18e7518934790612

The code has been already updated, but `7.19.0` did not include this. :(

davidhalter:
As a temporary fix for anyone just trying to get things working again:

```
pip install jedi==0.17.2
```

----

It would be really nice if you could quickly release a 7.19.1.

Sorry for that. I did not realize that IPython with that fix was not released yet. I usually test IPython completions before doing a Jedi release, but not this time :/. It will probably also not happen in the future anymore, because I'm going to release Jedi 1.0 soon, so this is probably the last time for a long time that you have to deal with deprecations in Jedi.

Still wish you a happy Christmas!

davidhalter:
By the way, a `7.19.1` release with the dependency `jedi<0.18.0` would also suffice.

jkryanchou:
@davidhalter Thanks for your solution. it finally work.

SamuelDSR:
I had the same problem with ipython and thanks to the solution of @davidhalter , it works again. :+1:

[...]

PR #12751, opened 2020-12-31
Pin Jedi dependency.

yurzo:
Pins Jedi for the 7.x branch since master its fixed already.
Addresses #12740 and #12745

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -186,14 +186,14 @@ extras_require = dict(
 )
 
 install_requires = [
-    'setuptools>=18.5',
-    'jedi>=0.10',
-    'decorator',
-    'pickleshare',
-    'traitlets>=4.2',
-    'prompt_toolkit>=2.0.0,<3.1.0,!=3.0.0,!=3.0.1',
-    'pygments',
-    'backcall',
+    "setuptools>=18.5",
+    "jedi>=0.16,<0.18.0",
+    "decorator",
+    "pickleshare",
+    "traitlets>=4.2",
+    "prompt_toolkit>=2.0.0,<3.1.0,!=3.0.0,!=3.0.1",
+    "pygments",
+    "backcall",
 ]
 
 # Platform-specific dependencies:


NeilGirdhar:
What's the point of this?  Isn't the master ipython already compatible with the newest jedi?

yan12125:
> Isn't the master ipython already compatible with the newest jedi?

Yep, so the author of this pull request targets 7.x branch instead of master. Compatibility with Jedi 0.18 (dcd9dc90aee7e4c5c52ce44c18e7518934790612) is not backported to 7.x branch - the latter still uses a parameter removed from the latest Jedi API: https://github.com/ipython/ipython/blob/7.x/IPython/core/completer.py#L1374

NeilGirdhar:
Oh, got it, didn't notice the target branch!

bollwyvl:
@yurzo thank you! I know it's an annoying thing, but could you re-format the lines that just change the single quotes to double quotes? I'm sure we'll get around to `black`ening the codebase at some point, but until then, it's a lot easier to see changes when a one-line fix is a one-line change.

@Carreau this is starting to cause lots of ripples, and is getting patched in other places... we've pinned on conda-forge, but it can have other consequences. Can you take a look at getting this, or equivalent change in!

yurzo:
@bollwyvl I'd be happy to remove all the quotes.

My original commit was against master, mostly out of ignorance of this project's governance, and it was a minimal Jedi pin (no double quotes).
However, it failed the `darker` checks. I did not know darker but I'm a fan of black so I excitedly acknowledged darker's suggestions.
This morning, prompted by your request, i diffed master vs 7.x and I estimate that darker is enabled on both, so if I do remove the double quotes, checks will fail.

If per project governance I can make this PR as you ask and it can be merged despite failing checks, please let me know and I'll go push a new patch.

bollwyvl:
Yeah, I haven't been tracking to what the state of automation is... :heavy_check_mark: is good!

Carreau:
Sorry I've been a bit overwhelmed. I can't have a look this week end but will try next week.

bollwyvl:
Understandable, a lot of things are overwhelming right now. Take care of yourself, for sure!

Carreau:
Thanks ! I'll try to make a release soon. Will try to get some of the recent PRs in, and do release notes. This week is a bit packed.

h-vetinari:
Thanks! :)

May I just ask though - is there a reason why https://github.com/ipython/ipython/commit/dcd9dc90aee7e4c5c52ce44c18e7518934790612 shouldn't be backported to `7.x` (possibly with a version-guard to keep compatibility with old jedi-versions)? I ask because [8.0](https://github.com/ipython/ipython/issues/11816) still seems a while off.

davidhalter:
It's not a problem - Jedi 0.18.0 fixes only minor issues for IPython. The release gets rid of Python 2, which was its major goal (and which does not matter for IPython).

h-vetinari:
> It's not a problem - Jedi 0.18.0 fixes only minor issues for IPython. The release gets rid of Python 2, which was its major goal (and which does not matter for IPython).

Respectfully, I disagree. If Jedi 0.18.0 breaks ipython 7.x (the very reason for this PR) and 8.x is not happening in the next ~2 months, then - if possible without crazy complexity - IMO 7.x should stay compatible with the newest jedi.

SamuelMS:
> Thanks ! I'll try to make a release soon. Will try to get some of the recent PRs in, and do release notes. This week is a bit packed.

@Carreau Just adding my voice to the clamor here – this is a pretty frustrating issue, as all of our dependent tools (Jupyter, ipdb, Django's shell_plus extensions, etc) break both locally and in production environments. Can you please cut a release to 7.x that pins the version? It can be based on `7.9` with no other changes from `master` – that'd be plenty.

PR #12752, opened 2021-01-01
Make better prediction of the virtualenv site-packages version

bxsx:
This PR improves virtualenv integration.

Current implementation assumes that the Python used in virtualenv is in the same version as the Python that is being used to run IPython. This is definitely not truth in most cases. In fact, one of the purpose of using virtualenv is to run different Python's interpreter versions. So the assumption is in opposite of the one of principal goals of the virtualenv idea.

This PR predicts a proper path to the virtualenv's libraries by looking-up at `$VIRTUAL_ENV` environmental variable against `pythonX.Y` or `pyX.Y` substrings. The `$VIRTUAL_ENV` usually contains Python's version in the name, e.g. `project-name-py3.7`, `project-name-py3.8`. If predicted path doesn't exist, the previous functionality is proceed.

Example:
- Python/system - 3.9
- Python/virtualenv - 3.7

Results:

- Current implementation - predicted path: `$VIRTUAL_ENV/lib/python3.9/site-packages`
- PR - predicted path: `$VIRTUAL_ENV/lib/python3.7/site/packages`

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -940,12 +940,11 @@ class InteractiveShell(SingletonConfigurable):
             p_ver = sys.version_info[:2]
 
             # Predict version from py[thon]-x.x in the $VIRTUAL_ENV
-            re_m = re.search(r"\bpy(?:thon)?([23])\.(\d+)\b", os.environ["VIRTUAL_ENV"])
-            if re_m:
-                predicted_p_ver = [int(num) for num in re_m.groups()]
-                predicted_path = Path(str(virtual_env_path).format(*predicted_p_ver))
+            predicted_p_ver = re.search(r"\bpy(?:thon)?([23])\.(\d+)\b", os.environ["VIRTUAL_ENV"])
+            if predicted_p_ver:
+                predicted_path = Path(str(virtual_env_path).format(*predicted_p_ver.groups()))
                 if predicted_path.exists():
-                    p_ver = predicted_p_ver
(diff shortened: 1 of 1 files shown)


bxsx:
This is interesting fix at 62e6c27. This bug also existed in the current `master`. It was introduced via #12548. I don't understand why this passed before. I also think it's necessary to patch the Windows part:
```
diff --git a/IPython/core/interactiveshell.py b/IPython/core/interactiveshell.py
index 1770b4f67..ded20f975 100644
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -932,7 +932,7 @@ def init_virtualenv(self):
             return

         if sys.platform == "win32":
-            virtual_env = Path(os.environ["VIRTUAL_ENV"], "Lib", "site-packages")
+            virtual_env = str(Path(os.environ["VIRTUAL_ENV"], "Lib", "site-packages"))
         else:
             venv_path_prefix = Path(os.environ["VIRTUAL_ENV"], "lib")
             venv_path_suffix = Path("site-packages")
```

On the other hand, AppVeyor build passes. Strange. I don't have access to Windows to test it myself.

bxsx:
OK. The PR is ready to review. I can also add a commit proposed in the https://github.com/ipython/ipython/pull/12752#issuecomment-753385477

Let me know what do you think. Thanks.

Carreau:
Thanks, and apologies for the delay. This looks good to me and I'll trust you a s virtualenv user.

PR #12757, opened 2021-01-06
Don't allow Jedi to go to v1.18.0

dangreenisrael:
This addresses issue https://github.com/ipython/ipython/issues/12740, where autocomplete breaks with the latest version of Jedi (1.18.0)

Currently, if you do a fresh install of iPython it results in broken autocomplete. This fixes it by preventing Jedi from going to whatever the latest version is.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -184,7 +184,7 @@ extras_require = dict(
 
 install_requires = [
     'setuptools>=18.5',
-    'jedi>=0.16',
+    'jedi>=0.16,<0.18.0',
     'decorator',
     'pickleshare',
     'traitlets>=4.2',
--- a/setup.py
+++ b/setup.py
@@ -183,15 +183,15 @@ extras_require = dict(
 )
 
 install_requires = [
-    'setuptools>=18.5',
-    'jedi>=0.16,<0.18.0',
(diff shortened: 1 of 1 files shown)


dangreenisrael:
Note that there is a fix to problem caused by leaving this open ended, but this could come up again in the future is we just leave this open to the latest version - even if it is a major version change.

https://github.com/ipython/ipython/commit/dcd9dc90aee7e4c5c52ce44c18e7518934790612

yan12125:
FWIW: the same changes were proposed previously and that PR is closed without merging https://github.com/ipython/ipython/pull/12746

Carreau:
Thanks ! I'll try to do a realease soon.

dangreenisrael:
Hi, I'm unclear on if I need to do a backport still.

Carreau:
This is actually wrong, master is compatible with 0.18, 7.x is not.

dangreenisrael:
Thanks @Carreau for noticing this.  May I ask that going forward we pin versions of anything not controlled by a lock file. Allowing versions to change like can result in non deterministic deployments and for organizations that deploy multiple times a day that can be problematic. 

Thanks again

Issue #11615, opened 2019-02-21
ipython UnicodeEncodeError when providing unicode char to "os.read(0,1)"


arekkusu:
I recently started playing with IPython and found a case where it exists with an "UnicodeEncodeError" error.
I confirmed the issue exist on the latest version. See how to reproduce / error bellow.

Note that I do expect an error to be raised by  'os.read()',  however I don't think IPython itself should exit.

```
vagrant@debian-9:~$ ipython
Python 3.5.3 (default, Sep 27 2018, 17:25:39)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.3.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os; os.read(0,1)
a
Out[1]: b'a'

In [2]:

In [2]: import os; os.read(0,1)
ab
Out[2]: b'a'

In [3]: b
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-89e6c98d9288> in <module>
----> 1 b

NameError: name 'b' is not defined

In [4]: import os; os.read(0,1)
日
Out[4]: b'\xe6'

In [5]: ??
Traceback (most recent call last):
  File "/home/vagrant/.local/bin/ipython3", line 11, in <module>
    sys.exit(start_ipython())
  File "/home/vagrant/.local/lib/python3.5/site-packages/IPython/__init__.py", line 125, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/home/vagrant/.local/lib/python3.5/site-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/home/vagrant/.local/lib/python3.5/site-packages/IPython/terminal/ipapp.py", line 356, in start
    self.shell.mainloop()
  File "/home/vagrant/.local/lib/python3.5/site-packages/IPython/terminal/interactiveshell.py", line 498, in mainloop
    self.interact()
  File "/home/vagrant/.local/lib/python3.5/site-packages/IPython/terminal/interactiveshell.py", line 489, in interact
    self.run_cell(code, store_history=True)
  File "/home/vagrant/.local/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2843, in run_cell
    raw_cell, store_history, silent, shell_futures)
  File "/home/vagrant/.local/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2863, in _run_cell
    if self.should_run_async(raw_cell):
[...]

PR #12758, opened 2021-01-07
Fix issue #11615  handle unicode encode error when print stack trace

takuya1981:


takuya1981:
This PR fixed issue #11615

IPython/core/interactiveshell.py
```diff
@@ -2068,13 +2068,17 @@ def showtraceback(self, exc_tuple=None, filename=None, tb_offset=None,
...
 
         Subclasses may override this method to put the traceback on a different
         place, like a side channel.
         """
-        print(self.InteractiveTB.stb2text(stb))
+        val = self.InteractiveTB.stb2text(stb)
+        try:
+            print(val)
+        except UnicodeEncodeError:
+            print(val.encode("utf-8", "backslashreplace").decode())
```
MrMino, line 2081: What are the benefits of having `try...except` here rather than just `print(val.encode(...).decode()`?

Carreau: I believe this is question of efficiency. In most case `print()` will work, so there is no need to do the expensive `encode()`, then `.decode()` dance, that would in the end often be no-op.

Carreau:
I think that it's harmless enough that we can get that in. 
Thanks.

PR #12759, opened 2021-01-07
Tests compatibility with Python 3.10

frenzymadness:
Python 3.10.0a4 has some differences and we are already building Python RPM packages with it so I'm bringing here the changes I need to do to make the tests pass with the latest Python 3.10.0a4. Tests still work for me in Python 3.9.

More explained in individual commits.

### DIFF
--- a/IPython/lib/tests/test_pretty.py
+++ b/IPython/lib/tests/test_pretty.py
@@ -6,16 +6,18 @@
 
 
 from collections import Counter, defaultdict, deque, OrderedDict
+from importlib import import_module
 import os
 import types
 import string
+import sys
 import unittest
 
 import nose.tools as nt
 import pytest
 
 from IPython.lib import pretty
-from IPython.testing.decorators import skip_without, skip_iptest_but_not_pytest
+from IPython.testing.decorators import skip_iptest_but_not_pytest
 
 from io import StringIO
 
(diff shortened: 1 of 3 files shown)


frenzymadness:
Not sure if I have to fix the tests. On windows, there is an error unrelated to this change:
```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\tmpvkn1o7qu\\tmpf87s_f5ntest_hist.sqlite'
```
and linter complains about stuff I haven't changed but I can definitely fix if needed.

IPython/lib/tests/test_pretty.py
```diff
@@ -138,13 +140,17 @@ def test_sets(obj, expected_output):
...
-    import xxlimited
-    output = pretty.pretty(xxlimited.Null)
+    module_name = "xxlimited" if sys.version_info < (3, 10, 0) else "xxlimited_35"
+    try:
+        module = import_module(module_name)
+    except ImportError:
+        pytest.skip(f"Module {module_name} is not available.")
+
+    output = pretty.pretty(getattr(module, "Null"))
     nt.assert_equal(output, 'xxlimited.Null')
```
hroncok, line 154: Just a curious ask: This is **not** `'xxlimited_35.Null'` on xxlimited_35, right?

frenzymadness: Exactly.

frenzymadness:
The change of test_oinspect is no longer needed because the postponed evaluation of type annotations has been postponed. For the rest, I'm gonna use PR #13080

Issue #12778, opened 2021-01-22
Example of display.Image not properly formatted


coelhudo:
The example shown at https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.Image isn't properly formatted. 

Just add a code block to the example, so it can be rendered correctly.

Carreau:
Fixed by #12727 thanks !

PR #12779, opened 2021-01-22
Fix code formatting. Issue #12778

coelhudo:
Fixed example of the display.Image. Before, the example wasn't shown as a code snippet. Instead, everything was in a single line. Now, the code is properly formatted.

Tested by regenerating the documentation using the `make html` command.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -851,18 +851,23 @@ class Image(DisplayObject):
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, this is addressed in #12727, which will make the documentation clearer and also test the examples in the docs. This also remove the # comments to make it more readable.

Though your contribution is much appreciated !

I'm going to close in favor of #12727, but let me know if there is anything I misunderstood.

Issue #12781, opened 2021-01-22
7.x: Racy use of ultratb.getargs should be removed in Python 3


Jongy:
The 7.x branch of IPython has its own version of `inspect.getargs`, used during traceback extraction, called in the flow of `VerboseTB.structured_traceback`. Using the decorator `with_patch_inspect` we temporarily replace `inspect.findsource` and `inspect.getargs`, restoring them afterwards.

Replacement is not safe in the multithreaded case. Multiple threads concurrently running `IPython.core.ultratb.AutoFormattedTB(mode="Verbose").structured_traceback()` may race to the point `inspect.getargs` is incorrectly "restored" to `IPython.core.ultratb.getargs`.

I assume with Python 2 it was not much of an issue (because IPython's `getargs` was merely augmented with the unmeged patch from http://bugs.python.org/issue14611).
In Python 3, it is an issue: `getargs` might access `co_code` (now a byte array) as a string, yielding an exception like:
```
...
  File "/usr/local/lib/python3.9/dist-packages/IPython/core/ultratb.py", line 260, in getargs
    op = ord(co.co_code[step])
TypeError: ord() expected string of length 1, but int found
```

for e.g such a code object:
```
In [20]: x = (_ for _ in ())                                                                                                                                                                               

In [21]: inspect.getargs(x.gi_code) 
```

And anyway - as far as I understand it, the addition in `getargs` is irrelevant in Python 3, because [tuple parameter unpacking has been removed](https://www.python.org/dev/peps/pep-3113/).

We have an application running `AutoFormattedTB(mode="Verbose").structured_traceback()` from multiple threads, and this bug was observed a couple of times already. I have fixed it by replacing updating `IPython.core.ultratb.getargs = inspect.getargs` on startup.

Here's a small & contrived reproducer. After running this snippet on a clean Docker `python:3.9` with IPython 7.19, it takes 2~ seconds to trigger.
```
import threading
import IPython.core.ultratb
import time
import random
import inspect

def g(n):
  # generate some stack depth
  if n > 0:
    g(n - 1)
  raise Exception()

def f():
  while True:
    time.sleep(random.random() / 10)
    try:
      g(20)
    except:
      print(''.join(IPython.core.ultratb.AutoFormattedTB(mode="Verbose").structured_traceback()))

    if inspect.getargs is IPython.core.ultratb.getargs:
      print("done!")
      break


for _ in range(100):
  threading.Thread(target=f).start()

# wait a bit...

# replaced forever
[...]

PR #12790, opened 2021-01-29
Remove old ultratb.getargs

Jongy:
See #12781.

Since 7.x is Python 3 only, it's okay to remove that piece of code.


Jongy:
Thanks @Carreau :)

PR #12795, opened 2021-01-30
fix async test with new pytest-asyncio.

Carreau:
us trio instead of asyncio as we can't run nested loops.


Carreau:
ok, let's try it.

PR #12804, opened 2021-02-02
Add explicit argtype counting, required for osx-arm64.

erykoff:
As reported in #12803 ipython crashes when loading matplotlib (or any windowing plugin) when running on Apple Silicon (arm64) macs.  This PR fixes the crash by adding explicit argument counts when calling out to objc, which should have been there before, but are required for osx-arm64.  (See also https://github.com/minrk/appnope/pull/7 ).


Carreau:
Thanks for diving into this ! How's the M1 ?

erykoff:
Thanks for your quick response!  Sadly I don't have an M1, still on the DTK until I get all my essential code ported.

stnatter:
> Thanks for diving into this ! How's the M1 ?

Thanks for fixing this! The M1 is absolutely fabulous. Completely disruptive. I can't see how Intel or AMD are going to survive this. x86 has run its course.

PR #12823, opened 2021-02-19
Added clear kwarg to display()

dsblank:
display() is a very useful function in the ipython kernel (especially now that it is in scope by default). However, many times that I use it, I also need `clear_output()`, and `clear_output(wait=True)`. However, it is quite a pain to have to:

```python
from IPython.display import clear_output
```

In addition, many user don't know about `clear_output`, nor the wait option.

This PR adds `display(clear=True)` with `wait=True` by default. If you want to clear it now rather than later, you can:

```python
display(clear=True)
print() # forces the clear now!
```

### DIFF
--- a/IPython/core/display_functions.py
+++ b/IPython/core/display_functions.py
@@ -120,6 +120,9 @@ def display(*objs, include=None, exclude=None, metadata=None, transient=None, di
         Set an id for the display.
         This id can be used for updating this display area later via update_display.
         If given as `True`, generate a new `display_id`
+    clear : bool, optional
+        Should the output area be cleared before displaying anything? If True,
+        this will wait for additional output before clearing. [default: False]
     kwargs: additional keyword-args, optional
         Additional keyword-arguments are passed through to the display publisher.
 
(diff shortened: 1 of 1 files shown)


dsblank:
@martinRenou What do you think about this? Also, not sure why it failed... I think a formatting rule, but I couldn't get "dark" to re-format it.

Carreau:
I have no objections with that; if it's ok with the xeus folks to make sure we don't have diverging implementation.
I'll take care of the reformatting.

dsblank:
I must admit that I don't understand the consistency check with xeus. There are quite a few kernels that have "display" functions. Is there some project to try to keep some core functionality across kernels? If so, metakernel should be in the loop.

Anyway, if you point me to the xeus implementation, I can make a PR there too.

martinRenou:
Now that xeus-python depends on IPython there should not be any problem on our side :) It should work out of the box with this change.

dsblank:
@Carreau I guess this is ready then?

Carreau:
Rebased on master and applied autoformatting where linter was complaining.

Carreau:
Test are passing, merging. Shis should be automatically backported and will thus be release at end of month.

PR #12826, opened 2021-02-21
IPDB: "context" command

MrMino:
Makes it possible to set the number of lines of code that will be shown in the backtrace _from inside_ the debugger.


Carreau:
Thanks !

Issue #10057, opened 2016-11-16
Exceptions in PyQt event loop and ipython


SamuelDeleglise:
I am not sure if this is really a bug, but rather an issue that I am facing with ipython and option qt. My problem is the following:

I have a PyQt program that is displaying some widgets and buttons.

I want the program to run either as a standalone python instance, or inside an ipython environment.  In this case, I use the following magic command in Jupyter console (formerly I had to use --gui=qt when launching the ipython qtconsole) 

    %pylab qt

In order to have a program that works both ways, my main module has the following lines:
    

    APP = QtGui.Qapplication.instance() # retrieves the ipython qt application if any
    if APP is None:
        APP = QtGui.QApplication(["foo"]) # create one if standalone execution

    if __name__=='__main__':
        APP.exec_() # Launch the event loop here in standalone mode 

**Here is my problem:** exceptions generated by the event loop are very hard to detect by the user because they pop-out in the background console. I would like to catch any exception occuring in the event loop, and display a warning (for intance in the QMainWindow status bar to make the user aware that an exception occured).

I have tried several strategies, but there seems to be a conspiracy between PyQt's and Ipython's internal machineries to make this impossible:

 - Reimplement sys.excepthook (see http://stackoverflow.com/questions/18740884/preventing-pyqt-to-silence-exceptions-occurring-in-slots): not working because ipython keeps overwritting sys.excepthook
 - Detecting if IPython is running, and then using IPYTHON.set_custom_exc (http://stackoverflow.com/questions/15752437/opening-an-ipython-shell-on-any-uncatched-exception): Unforunately, qt event loop exceptions do not trigger the handler.
 - Overwrite QApplication.notify: bad luck, the native QApplication.notify function that I intend to call in the derived function is not throwing exceptions, nor does the return value (boolean) reflect the correct execution of the slots. The answer in this thread is interesting: http://stackoverflow.com/questions/10665655/how-to-log-uncatched-exceptions-of-a-qapplication, however, it seems that this strategy works in Qt c++, but the python wrapper of notify just prints the exceptions to the console instead of raising them.  
 - Overwriting the function that IPython will use as a monkey patch for sys.excepthook (as suggested in http://stackoverflow.com/questions/1261668/cannot-override-sys-excepthook/28758396#28758396) is not working either. In fact, Ipython with the qt option will prevent the qt event loop from even calling sys.excepthook.

It is a problem that keeps bothering me since a long time. Does anyone have a solution?

[...]

PR #12842, opened 2021-03-03
Reclaim sys.excepthook for interactive shell in %gui qt

tlambert03:
Currently, if someone uses the `%gui qt` magic, and an exception occurs in the event loop, it gets rendered with an unformatted traceback.  This appears to be because sys.excepthook is being restored back to `BaseIPythonApplication.excepthook` before the Qt event loop is run (see https://github.com/ipython/ipython/issues/10057#issuecomment-754209074).   To see this, run the following gui in an `ipython --gui qt` session:
```python
from qtpy.QtWidgets import QPushButton

def _raise():
    raise ValueError("clicked")

b = QPushButton("hi")
b.clicked.connect(_raise)
b.show()  # and click the button for traceback
```

Following [this tip](https://github.com/ipython/ipython/issues/10057#issuecomment-261494713), this PR schedules a callback to restore `sys.excepthook` back to the `InteractiveShell.excepthook` immediately after the event loop starts.

### DIFF
--- a/IPython/terminal/pt_inputhooks/qt.py
+++ b/IPython/terminal/pt_inputhooks/qt.py
@@ -19,17 +19,17 @@ def inputhook(context):
     global _appref
     app = QtCore.QCoreApplication.instance()
     if not app:
-        if sys.platform == "linux":
-            if not os.environ.get("DISPLAY") and not os.environ.get("WAYLAND_DISPLAY"):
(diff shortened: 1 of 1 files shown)


Carreau:
Marking it for end of march release.

tlambert03:
thanks!

PR #12844, opened 2021-03-04
Fix UnboundLocalError in IPDB when running 'context' command without argument

MrMino:
Minor issue, doesn't break the main functionality - throughout refactoring and fiddling with initial implementation I managed to bork that, sorry.


Carreau:
I should have seen it as well; fault is not on you. Thanks.

PR #12849, opened 2021-03-06
7.x timeit bug?

jhsxy2005:
```
In [1]: a=b=0

In [2]: %%timeit
    ...: a=1+b
    ...:
    ...:
63.9 ns ± 0.16 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: a
Out[60]: 0

# is ok

In [4]: %%timeit
    ...: a=1+b
    ...: b=1
UnboundLocalError: local variable 'b' referenced before assignment
# !!!

In [5]: %%timeit
    ...: b=0
    ...: a=1+b
    ...:
    ...:
60.1 ns ± 1.62 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
Is it a bug?Since we at least know that the timeit result does not affect the global, we can implicitly mean that the variable in the expression is a copy of the global variable in the statement, and this is true for every operation, but the meaning here is that we need to manually assign at least one b each time.
I think since In[2]:a=1+b implicitly calls global b, why not implicitly add b=0 every time?And In[2] takes longer to run than In[5].

它是一个bug吗？因为我们至少知道整体下来，timeit的计算结果并不影响全局，所以可以隐式的理解为语句中，表达式中的变量只是全局变量的副本，每一次运算都是如此，但是这里的意思是，每次至少需要手动赋值一个b。

我认为，既然In[2]:a=1+b，隐式的调用了全局的b，那为何不每次都隐式的加上b=0。而且In[2]的运行时间，要比In[5]长。


Carreau:
1) That's the same in pure python using the timeit module:

```
$ python
Python 3.8.5 | packaged by conda-forge | (default, Sep 16 2020, 17:43:11)
[Clang 10.0.1 ] on darwin
>>> from timeit import repeat
>>> repeat(stmt='a=1+b; b=1', repeat=2, number=10, globals={'b':0})
  File "/Users/bussonniermatthias/miniconda3/lib/python3.8/timeit.py", line 238, in repeat
    return Timer(stmt, setup, timer, globals).repeat(repeat, number)
  File "/Users/bussonniermatthias/miniconda3/lib/python3.8/timeit.py", line 205, in repeat
    t = self.timeit(number)
  File "/Users/bussonniermatthias/miniconda3/lib/python3.8/timeit.py", line 177, in timeit
    timing = self.inner(it, self.timer)
  File "<timeit-src>", line 6, in inner
UnboundLocalError: local variable 'b' referenced before assignment
>>>
```

`%timeit` is just a convenience syntax; if you think the behavior should change, then ask cpython. 

2) did you try to do a PR on purpose ? that seemed like an error, so I'm going to close. Thanks.

PR #12856, opened 2021-03-12
Add loading best practice to custom magic docs

### LINKED ISSUE #12838: Custom magic best practice: import vs. load_ext
Hello, I'm developing a Python package which implements a custom magic with the `@register_line_cell_magic` decorator.

Something I found unclear when reading the [documentation](https://ipython.readthedocs.io/en/stable/config/custommagics.html) is whether my magic should be loaded as an extension or not. The initial example on that page that uses the decorator registers the magic on module import, so then a user who runs `import my_module` would have the magic available. 

However, later examples that use the `Magics` class-based approach instead register the magic in `load_ipython_extension`, which to my understanding requires the user to use `%load_ext my_module` to make the magic available.

I'm not seeing discussion on which is the recommended way of registering magics, especially since it appears they lead to a different entrypoint for a user. Is there a standard or best practice recommended for which approach to use? 

One thing I noticed is that the import approach throws a `NameError` when the module is loaded outside of an IPython kernel, which is straightforward to work around with a try-except.

jayqi:
Adding to the documentation for custom magics on the recommended best practice of how to register the magic when distributing via a package, based on the discussion in #12838. 

Closes #12838.

### DIFF
--- a/docs/source/config/custommagics.rst
+++ b/docs/source/config/custommagics.rst
@@ -153,6 +153,14 @@ Here is a full example of a magic package. You can distribute magics using
 setuptools, distutils, or any other distribution tools like `flit
 <http://flit.readthedocs.io>`_ for pure Python packages.
 
+When distributing magics as part of a package, recommended best practice is to
+execute the registration inside the `load_ipython_extension` as demonstrated in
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks !

PR #12860, opened 2021-03-14
Bail on sunken brackets

### LINKED ISSUE #12859: Bug: extra closing ) does not trigger SyntaxError
Consider the following:
```
> ipython
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: print('hello'))
   ...:
   ...:
```
The extra `)` should have resulted in a `SyntaxError`; instead, more input is apparently expected. This problem is present in earlier versions of IPython as well.

MrMino:
Fixes #12859.

TL;DR - instead of:
```
In [1]: ]
   ...: 
```

Do:
```
In [1]: ]
  File "<ipython-input-1-de603c91038f>", line 1
    ]
    ^
SyntaxError: unmatched ']'
```

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -678,7 +678,11 @@ class TransformerManager:
 
         # Bail if we got one line and there are more closing parentheses than
         # the opening ones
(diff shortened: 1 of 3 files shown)


Carreau:
Ah ! Nice !

PR #12863, opened 2021-03-18
fix grammar/typos in docs/interactive

kramred:
Hello,

This PR just has a few quick (low-level) grammar and typo fixes for a doc page that I was reading.
Feel free to integrate/change whatever you see fit.

### DIFF
--- a/docs/source/interactive/python-ipython-diff.rst
+++ b/docs/source/interactive/python-ipython-diff.rst
@@ -3,20 +3,20 @@ Python vs IPython
 =================
 
 This document is meant to highlight the main differences between the Python
-language and what are the specific construct you can do only in IPython.
+language and what are the specific constructs you can do only in IPython.
 
-Unless expressed otherwise all of the construct you will see here will raise a
+Unless expressed otherwise all of the constructs you will see here will raise a
 ``SyntaxError`` if run in a pure Python shell, or if executing in a Python
 script.
 
-Each of these features are described more in detail in further part of the documentation.
+Each of these features is described more in detail in the further parts of the documentation.
 
 
 Quick overview:
 ===============
 
 
-All the following construct are valid IPython syntax:
+All the following constructs are valid IPython syntax:
 
 .. code-block:: ipython
 
@@ -58,8 +58,8 @@ All the following construct are valid IPython syntax:
        ...: print $months[0];
 
 
-Each of these construct is compiled by IPython into valid python code and will
-do most of the time what you expect it will do. Let see each of these example
+Each of these constructs is compiled by IPython into valid python code and will
+do most of the time what you expect it will do. Let see each of these examples
 in more detail.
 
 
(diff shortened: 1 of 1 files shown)


docs/source/interactive/python-ipython-diff.rst
```diff
@@ -58,8 +58,8 @@ All the following construct are valid IPython syntax:
        ...: print $months[0];
 
 
-Each of these construct is compiled by IPython into valid python code and will
-do most of the time what you expect it will do. Let see each of these example
+Each of these constructs is compiled by IPython into valid python code and will
+do most of the time what you expect it will do. Let see each of these examples
```
MrMino, line 62: "Let see" seems invalid too?

```suggestion
do most of the time what you expect it will do. Let's see each of these examples
```

docs/source/interactive/python-ipython-diff.rst
```diff
@@ -89,8 +89,8 @@ shortcut to get help. A question mark alone will bring up the IPython help:
     -------------
     ...
 
-A single question mark before, or after an object available in current
-namespace will show help relative to this object:
+A single question mark before or after an object available in the current
+name-space will show help relative to this object:
```
MrMino, line 93: `namespace` is a perfectly valid word in the context of computing, there's no need to change it.

https://en.wikipedia.org/wiki/Namespace#:~:text=In%20computing%2C%20a%20namespace%20is,they%20can%20be%20easily%20identified.

docs/source/interactive/python-ipython-diff.rst
```diff
@@ -126,8 +126,8 @@ and if possible display the python source code of this object.
 
 
 
-If you are looking for an object, the use of wildcards ``*`` in conjunction
```
MrMino, line 129: I think temoval of this comma makes the sentence nonsensical? What's wrong here?

docs/source/interactive/python-ipython-diff.rst
```diff
@@ -142,10 +142,10 @@ Shell Assignment
 ================
 
 
-When doing interactive computing it is common to need to access the underlying shell.
+When doing interactive computing it is a common need is accessing the underlying shell.
```
MrMino, line 145:
```suggestion
When doing interactive computing it is a common need to access the underlying shell.
```

docs/source/interactive/python-ipython-diff.rst
kramred: Good catch, thanks for the review. I'll update the PR with all/most of your suggestions included in a few minutes

MrMino (changes requested):
Thanks, nice @kramred! I'm not a native speaker, but I think I found a few things that could be improved further.

MrMino (approved):

Carreau:
Thanks @kramred  and @MrMino for the review; merging as you approved !

PR #12864, opened 2021-03-19
Allow specifying alt text for IPython.display.Image.

blois:
For accessibility reasons users should be encouraged to add meaningful alt text to images. Currently there is no recommended way to do this using the standard Image object. Adding this field is a step towards having a common way for notebook renderers to support alt text and towards encouraging libraries creating images to add meaningful alt text.

See https://www.w3.org/standards/webdesign/accessibility#examples

Currently JLab just does https://github.com/jupyterlab/jupyterlab/blob/master/packages/rendermime/src/renderers.ts#L674.

I'd like to see the ecosystem get to the point where titles on graphs displayed as images automatically become alt text for those images.


bollwyvl:
This is brilliant, and we should do it at the _syntax_ level so that people can adopt it easily. 

However, `alt` is a subset of a much larger ecosystem that spans a number of concerns, including internationalization, intent, etc. in addition to assistive technology awareness.

Ideally, we'd encourage consistency at the metadata level by adding "blessed" terms to the nbformat JSON schema so that clients aren't playing guessing games. The best of worlds would be to identify an existing standard, whether in JSON schema or whatever: my pick for this would be the [web annotation data model](https://www.w3.org/TR/annotation-model/) which has mappings to [w3c annotation-aria](https://w3c.github.io/annotation-aria/#annotation-model-mappings) which can be accessibly embedded in [html](https://www.w3.org/TR/annotation-html/#example-using-blockquote-and-q-tags).

Practically, this would mean replacing:

```yaml
"metadata": {"image/png": {"alt": "a picture of an elephant"}}
```

to...

```yaml
"metadata": {"image/png": {
    "annotations": [
      # some more verbose JSON that other tools can read and reason about
    ]
  }
}
```

For some related things around making complex things describable, the [p5](https://p5js.org/) folk have [done](https://github.com/processing/p5.accessibility) and [written about](https://medium.com/processing-foundation/making-p5-js-accessible-e2ce366e05a0) some sterling work.

:bellhop_bell: @tonyfast @isabela-pf @manfromjupyter

blois:
I honestly do not have much experience in this area and am happy to defer to others for a more formal specification if there is interest in that direction.

I think this is particularly important because while many people have been doing great work to make the tools accessible for the authors of these notebooks this seems like an obvious gap for the readers of these notebooks, especially given the plethora of publishing avenues.

bollwyvl:
@blois really, thanks for starting this: I don't think anybody is _experienced_ at this point, as we're trying to do something relatively difficult across a large number of pieces.

> given the plethora of publishing avenues.

This is _definitely_ the motivation for getting the metadata right. With the right metadata, we can enrich:
- the library experience: as you've started, every kernel completion request can encourage adding this information
- the authoring UI: Lab can have offer tools for annotating stuff, post-hoc, if you don't control the underlying code
- "raw" generated HTML: nbconvert and nbviewer can show annotations
- "soft" publication: sphinx, jupyter-book, etc. can present and make this information searchable
- "hard" publication: more advanced PDF pipelines _can_ make this work (not just browser screenshots)

Starting where you have is a really great place!

blois:
@bollwyvl If the desire is more than a plain string then it probably makes sense for this to be specified in nbformat first. I had started here since it appears to be the de facto specification for output metadata fields.

blois:
I'd prefer to avoid broadening the scope of this change to to encompass more accessibility features across more types of outputs as I feel that work is much larger, more speculative and lacks a clear owner. Adding an `alt` field to the metadata is a straightforward fix to the problem. 

If there can be a more generalized set of accessibility metadata added to nbformat and outputs then it should be entirely reasonable for those to supersede simpler `alt` fields similar to how outputs themselves contain multiple prioritized formats.

Please let me know if there are further changes I should make here, otherwise feel free to close this PR if an nbformat specification approach is preferred.

manfromjupyter:
I personally prefer keeping it as:

manfromjupyter:
I personally prefer keeping it as:
```
"metadata": {"image/png": {"alt": "a picture of an elephant"}}
```
The closer we can keep it to what that coding languages already exists (HTML, JQuery, etc) the better. Users shouldn't have to learn a new language or way of doing things just to support persons with disabilities/handicaps. `alt` is preferred and if that's already in the product, I say don't change it, but that's just me.

Carreau:
Sorry for the delay in reviewing / merging. 

I have no opposition to this, it's just unclear to me if you want to get that in as is, or build consensus first.
I'm happy to get this in master, Ipython 8.0 will not be released soon, so there is still time if this gets in to be refined.

Let me know and I'm happy to merge.

blois:
If someone is interested in picking up the work to formally spec accessibility for outputs then I think this PR should be dropped. I opened https://github.com/jupyter/accessibility/issues/37 and attended the regular meeting to discuss, I do not believe anyone currently has the time to pick up the larger work.

I don't think there's any immediate urgency for this issue and am fine waiting to see if we can build traction elsewhere. 

The one thing is that once some variant of this lands is that it may be good to try to get matplotlib images to automatically fill in the alt text. I believe the primary work would be somewhere in https://github.com/ipython/ipython/blob/d737c325060f8d059f381078a3f2f73a087b2d8d/IPython/core/pylabtools.py#L209  and it may be nice to try to get this into the same release. I'll start looking into this next step shortly.

blois:
At this point I believe that the approach proposed here is the best path forward. I'd like to see this merged so we can move on to encouraging users and libraries to include accessible alt text on images.

minrk:
I agree that landing this as-is is the best start. It doesn't preclude more detailed abstract approaches, and we shouldn't have to wait for those. This is simple, straightforward, and solves a real issue without introducing any conflict with other ideas.

minrk (approved):

PR #12884, opened 2021-04-02
Strip vi-mode style prompts from the input

### LINKED ISSUE #12841: paste into ipython stop working with vi mode
this works well:

```
[ins] In [46]: In [41]: s = 'Abc Def Ghi'
[ins] In [47]:
```

this does not work.
```
[ins] In [44]: [ins] In [41]: s = 'Abc Def Ghi'                       
          ...:                 
          ...:                                                        
  File "<ipython-input-44-0af6a6a3f6c4>", line 1
    [ins] In [41]: s = 'Abc Def Ghi'
          ^
SyntaxError: invalid syntax
```

is it possible to configure ipython to also accept transcript under vi mode?

MrMino:
This strips `"[ins] "` and `"[nav] "` prefix of the prompt from the commands.

See [this debugex snippet](https://www.debuggex.com/r/4trT0ugmRh0W8VoO) for a visualization of what the regex does now.

To do:
- [x] Single-line paste stripping
- [x] Multiple-line paste stripping
- [x] Tests for both
- ~~Docs? Should this also land in that one sphinxext that does this for doctest?~~

Fixes #12841

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -89,7 +89,28 @@ classic_prompt = PromptStripper(
     initial_re=re.compile(r'^>>>( |$)')
 )
 
-ipython_prompt = PromptStripper(re.compile(r'^(In \[\d+\]: |\s*\.{3,}: ?)'))
+ipython_prompt = PromptStripper(re.compile(
+    r'''
+    ^(                         # Match from the beginning of a line, either:
+
+                               # 1. First-line prompt:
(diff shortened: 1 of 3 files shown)


Carreau:
Rebased and reformatted.

MrMino:
Thanks @Carreau!

Carreau:
I'm happy to get that in as-is, if you wish.

MrMino:
Weird. I thought that multiline pastes still don't work (and are passed through different code), but I've just tested it in all the ways I can think of and it works.

I added tests, so if that looks good - let's get that in.

PR #12888, opened 2021-04-04
Add a testcase for Image() raising FileNotFoundError on a bad filename

### LINKED ISSUE #11535: TypeError when no image found
Hi there!
I get the following error when executing display(Image('file.jpg')) on Jupyter with no image file available.
Shouldn't it error with a meaningful message?

```
c:\users\alex\appdata\local\programs\python\python36\lib\site-packages\IPython\core\formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

c:\users\alex\appdata\local\programs\python\python36\lib\site-packages\IPython\core\display.py in _repr_mimebundle_(self, include, exclude)
   1242         if self.embed:
   1243             mimetype = self._mimetype
-> 1244             data, metadata = self._data_and_metadata(always_both=True)
   1245             if metadata:
   1246                 metadata = {mimetype: metadata}

c:\users\alex\appdata\local\programs\python\python36\lib\site-packages\IPython\core\display.py in _data_and_metadata(self, always_both)
   1251     def _data_and_metadata(self, always_both=False):
   1252         """shortcut for returning metadata with shape information, if defined"""
-> 1253         b64_data = b2a_base64(self.data).decode('ascii')
   1254         md = {}
   1255         if self.metadata:

TypeError: a bytes-like object is required, not 'str'

c:\users\alex\appdata\local\programs\python\python36\lib\site-packages\IPython\core\formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

c:\users\alex\appdata\local\programs\python\python36\lib\site-packages\IPython\core\display.py in _repr_png_(self)
   1268     def _repr_png_(self):
   1269         if self.embed and self.format == self._FMT_PNG:
-> 1270             return self._data_and_metadata()
   1271 
   1272     def _repr_jpeg_(self):

c:\users\alex\appdata\local\programs\python\python36\lib\site-packages\IPython\core\display.py in _data_and_metadata(self, always_both)
   1251     def _data_and_metadata(self, always_both=False):
   1252         """shortcut for returning metadata with shape information, if defined"""
-> 1253         b64_data = b2a_base64(self.data).decode('ascii')
   1254         md = {}
   1255         if self.metadata:

TypeError: a bytes-like object is required, not 'str'
```

MrMino:
Fixes #11535 (i.e. I think that adding this test would be the last thing needed to close that issue).

Proper exception was added in #11663. This tests for if it's actually there, so that whenever the refactor mentioned in that PR comes, we don't accidentally get back to the state from before it.

### DIFF
--- a/IPython/core/tests/test_display.py
+++ b/IPython/core/tests/test_display.py
@@ -460,4 +460,4 @@ def test_display_handle():
 
 @nt.raises(FileNotFoundError)
 def test_image_bad_filename_raises_proper_exception():
-    display.Image("/this/file/does/not/exist/")
+    display.Image("/this/file/does/not/exist/")._repr_png_()
--- a/IPython/core/tests/test_display.py
+++ b/IPython/core/tests/test_display.py
@@ -457,3 +457,7 @@ def test_display_handle():
         'update': True,
     })
 
+
+@nt.raises(FileNotFoundError)
+def test_image_bad_filename_raises_proper_exception():
+    display.Image('/this/file/does/not/exist/')
--- a/IPython/core/tests/test_display.py
+++ b/IPython/core/tests/test_display.py
@@ -460,4 +460,4 @@ def test_display_handle():
 
 @nt.raises(FileNotFoundError)
 def test_image_bad_filename_raises_proper_exception():
-    display.Image('/this/file/does/not/exist/')
+    display.Image("/this/file/does/not/exist/")


Carreau:
Thanks ! I tried to fix the tests.

1) I ran the formatter 
2) the data is actually computed when the repr is called; so I explicitly called it. 
3) Maybe we should make it eager at creation time. It's weird to not have `Image(...)` on its own fail.

MrMino:
> the data is actually computed when the repr is called; so I explicitly called it.

Oh... I got bamboozled by the repr firing up in the console and showing the TB :smile:...

> Maybe we should make it eager at creation time. It's weird to not have Image(...) on its own fail.

Yup. Is there a good reason for loading the data only after the formatter fires up the repr? The `__init__` already checks existence via `_safe_exists(data)`, it just looks like it is confused as to how to interpret this argument.

Carreau:
> Yup. Is there a good reason for loading the data only after the formatter fires up the repr? The `__init__` already checks existence via `_safe_exists(data)`, it just looks like the implementation is "confused" as to how to interpret this argument.

No, I think we can load at `__init__`. Let get that in, and I'll open a new issue.

PR #12898, opened 2021-04-09
Cleaner shutdown

Carreau:
Reset recreate a new history; so ther is no point in calling it after
closing history session.

Unset  class variable to be sure they are not reused.


Carreau:
Move some of the testing to in-memory; this also will make the test faster and fix the sqlite write only issues.

Issue #12692, opened 2020-11-01
#12639 seem to create random errors


Carreau:
@jetpacktuxedo your PR seem to randomly create errors like this:

https://travis-ci.com/github/ipython/ipython/jobs/423309288:

```
        if args.raise_error and p.returncode!=0:
>           raise CalledProcessError(p.returncode, cell)
E           subprocess.CalledProcessError: <unprintable CalledProcessError object>
```

As you wrote it maybe you will find this issue faster than I.

jetpacktuxedo:
That's interesting... I'm not sure why that would be unprintable. I think `p.returncode` should be either `None` (if the process hasn't completed yet) or an integer return code, which would leave `cell` to make it unprintable, but I'm pretty sure that's just an encoded binary string.

It's possible that I introduced some sort of timing issue in the tests where it is trying to evaluate the success of the test before the test script call is done executing?

Carreau:
It's possible, it's possible it's also a but in asyncio on a specific python version. I haven't really had the time to investigate; 
And I think the test is not expecting a non-zero return value; so it's likely possible something else is wrong.

takluyver:
I think this occurs when `p.returncode` is None.

The `<unprintable ... object>` message appears when `str(x)` fails in the traceback module. Looking at the code of `CalledProcessError.__str__`, it is clearly assuming that it will always get an integer, and the string formatting with throw an error with None.

On Github actions, this seems to fail consistently rather than occasionally, so I'll figure out some workaround.

takluyver:
The workaround/fix I've got now is in two parts:

- Wait for the process to finish in `_stream_communicate()`. The stdout & stderr pipes being closed doesn't automatically mean that the child process has terminated, or that we've processed its return code. The `.communicate()` methods in subprocess and asyncio both do this.
- If we reach the point of raising the error and returncode is still None, use -9 (i.e. SIGKILL) instead. I think this should only be possible when the process has been killed but we haven't got the return code yet.

I'm not sure about the logic in the KeyboardInterrupt handler, though - it's manipulating an asyncio process object, but outside the asyncio event loop, calling `time.sleep()` to wait between SIGINT/SIGTERM/SIGKILL. I'm not convinced `p.returncode` will actually get set even if the process exits correctly.

Carreau:
Closing as this does not seem to be an issue anymore.

PR #12900, opened 2021-04-10
Move to Github Actions

MrMino:
#12694

To do:
- [ ] nightly?
- [x] docs
- [x] `check-manifest`
- [x] `fixup_whats_new_pr.py`
- [x] OSx
- [x] ?deploy? stuff
- [x] `iptest`
- [x] `pytest`


MrMino:
Aaand it failed on 

```
ERROR: IPython.core.tests.test_magic.test_script_out_err
...
subprocess.CalledProcessError: <unprintable CalledProcessError object>
```

Looks like the knownfail from earlier?

takluyver:
The failure you mentioned seems to be #12692; I think I've fixed it (roughly).

takluyver:
I think this is ready-ish now.

The one part definitely missing compared to the Travis setup is checking on Python 'nightly'. This is an allowed failure in Travis, but Github actions has no equivalent of that, and I've never really been convinced that allowed failures are useful anyway - my experience is that they're ignored.

There's some complication in the Travis config for OSX that I haven't copied (installing a Python framework with Homebrew). I'm guessing that that isn't necessary on GHA - it seems to work without it.

For now, I've put the creation of Github releases in a separate workflow from the tests. This keeps things simple, but it will make the release even if the tests fail, which may or may not be what you want. Hopefully by the point a tag is pushed, you're confident in it anyway, and if there's a rare random failure you don't necessarily want it to interrupt the automation. But obviously that can be rearranged if you prefer. [Here are the docs for the gh-release action](https://github.com/marketplace/actions/gh-release) which I used.

takluyver:
(PS: I'll try to keep an eye out for comments, but anyone should feel free to take the branch over - especially in time zones where you can carry on while I'm asleep :-)

minrk:
You're definitely right that the framework Python isn't needed because GHA's setup-Python works on mac, unlike Travis.

I also don't think we need nightly Python, at least not before merging.

minrk (approved):

Carreau:
Thanks, all, let's get that in, and try if that backports cleanly on stable.

PR #12902, opened 2021-04-10
Make %precision work for numpy.float64 type

petercorke:
Many NumPy functions return a `float64` type whose display does not obey `%precision`, for example:

```
In [4]: precision 2
Out[4]: '%.2f'

In [5]: np.linalg.norm([1,2])
Out[5]: 2.23606797749979
```

IPython already sets precision for NumPy **arrays** using `set_printoptions()` but this does not affect the format for scalars.

This simple change sets the formatter for `float64` to be the same as for `float` so that any changes to precision will take effect for `float64`.

### DIFF
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -636,7 +636,6 @@ class PlainTextFormatter(BaseFormatter):
 
         This parameter can be set via the '%precision' magic.
         """
-
         new = change['new']
         if '%' in new:
             # got explicit format string
@@ -678,7 +677,9 @@ class PlainTextFormatter(BaseFormatter):
     def _type_printers_default(self):
         d = pretty._type_pprinters.copy()
         d[float] = lambda obj,p,cycle: p.text(self.float_format%obj)
+        # if NumPy is used, set precision for its float64 type
         if 'numpy' in sys.modules:
+            import numpy
             d[numpy.float64] = lambda obj,p,cycle: p.text(self.float_format%obj)
         return d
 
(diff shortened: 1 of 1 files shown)


MrMino (changes requested):
I haven't investigated all of the implications of this approach yet, but this could use a test & appropriate mention in the `%precision` docs.

Also, merging this will require adding a "What's new" fragment to the pr directory.

See [CONTRIBUTING.md#Pull Requests](https://github.com/ipython/ipython/blob/master/docs/source/whatsnew/pr/README.md) and [`docs/source/whatsnew/pr/README.md`](https://github.com/ipython/ipython/blob/master/docs/source/whatsnew/pr/README.md)

petercorke:
Just [this section](https://ipython.readthedocs.io/en/stable/interactive/magics.html?highlight=precision)?  
Testing should be with formatters or with magics?

MrMino:
> Just this section?

Yes, [here](https://ipython.readthedocs.io/en/stable/interactive/magics.html?highlight=%25precision#magic-precision) to be precise (the highlight in your link didn't work for me).

> Testing should be with formatters or with magics?

I would do both, but at least having a doctest-style test that checks whether the precision has actually changed after using `%precision` should be sufficient.

Thanks!

Carreau:
Let's get that in and I'll open an issue about the test and what's new.
Especially doctest might be tricky as we are trying to drop support for nose (incompatible 3.10) and pytest does not currently run doctests.

petercorke:
Thanks, I was going to do this tomorrow...  Do you still want me to do the docktest things?  I know what it is but have never used it before.

MrMino:
@petercorke It's always appreciated if you try. If you run into issues you can always communicate them through #12917 and that way we'll at least know where the problems are.

Issue #12913, opened 2021-04-15
Make ipdb interact run an embedded IPython shell instead of code.interact()


iwanb:
The `interact` in (i)pdb runs `code.interact()`:
https://docs.python.org/3/library/pdb.html#pdbcommand-interact

It would be great if it would instead run an IPython shell with current frame's namespace instead.

```
Python 3.6.13 (default, Feb 16 2021, 00:00:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %debug on                                                                                                                                                       
NOTE: Enter 'c' at the ipdb>  prompt to continue execution.
> <string>(1)<module>()

ipdb> interact                                                                                                                                                          
*interactive*
In : 
```

AFAIK something like this should do, but I'm not sure this handles all cases:
```
from IPython.terminal.embed import InteractiveShellEmbed

class _CustomTerminalPdb(TerminalPdb):
    def do_interact(self, arg):
        ipshell = InteractiveShellEmbed(banner1="*interactive*", exit_msg="now exiting InteractiveConsole...")
        global_ns = self.curframe.f_globals
        ipshell(module=sys.modules.get(global_ns['__name__'], None), local_ns=self.curframe_locals)
```

MrMino:
Nice, thank you. Could you please put it into a PR? We'll see how it works afterwards.

PR #12915, opened 2021-04-16
Start embedded IPython shell for ipdb interact, #12913

iwanb:
Gets into a recursive embedded IPython shell when you type "interact" inside the debugger, instead of `code.interact()`.

I'm not sure that's the right way to implement it but it works. It might be better to reuse `self.shell`.

```
ipython
Python 3.9.3 (default, Apr  8 2021, 23:35:02) 

In [1]: %debug on
> <string>(1)<module>()

ipdb> interact
*interactive*
In [1]: %debug on
NameError: name 'on' is not defined

In [2]: %debug
ERROR:root:No traceback has been produced, nothing to debug.

In [3]: %pdb on
Automatic pdb calling has been turned ON

In [4]: raise Exception()

<ipython-input-4-32c500991385> in <module>
----> 1 raise Exception()

Exception: 
> <ipython-input-4-32c500991385>(1)<module>()
----> 1 raise Exception()

ipdb> 
ipdb> interact
*interactive*
In [1]:                                                                         
now exiting InteractiveConsole...
ipdb>                                                                           


In [5]:                                                                         
now exiting InteractiveConsole...
ipdb>                                                                           


In [2]:                                                                         
```

@MrMino

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -136,7 +136,7 @@ class TerminalPdb(Pdb):
         ipshell = embed.InteractiveShellEmbed(
             config=self.shell.config,
             banner1="*interactive*",
-            exit_msg="now exiting InteractiveConsole...",
+            exit_msg="*exiting interactive console...*",
         )
         global_ns = self.curframe.f_globals
         ipshell(
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -7,6 +7,7 @@ from IPython.core.debugger import Pdb
 from IPython.core.completer import IPCompleter
 from .ptutils import IPythonPTCompleter
 from .shortcuts import create_ipython_shortcuts
+from . import embed
 
 from pygments.token import Token
 from prompt_toolkit.shortcuts.prompt import PromptSession
@@ -130,6 +131,18 @@ class TerminalPdb(Pdb):
             self.postloop()
         except Exception:
             raise
+
+    def do_interact(self, arg):
+        ipshell = embed.InteractiveShellEmbed(
+            config=self.shell.config,
+            banner1="*interactive*",
+            exit_msg="now exiting InteractiveConsole...",
+        )
+        global_ns = self.curframe.f_globals
+        ipshell(
+            module=sys.modules.get(global_ns["__name__"], None),
+            local_ns=self.curframe_locals,
+        )
 
 
 def set_trace(frame=None):
--- a/docs/source/whatsnew/pr/ipdb-interact.rst
+++ b/docs/source/whatsnew/pr/ipdb-interact.rst
@@ -1,0 +1,4 @@
+IPython shell for ipdb interact
+-------------------------------
+
+The ipdb ``interact`` starts an IPython shell instead of Python's built-in ``code.interact()``.


MrMino:
You need to run `darker` on your changes for CI to pass. Here's how:

```
# Run this in the repository root
# Install darker
pip install darker
# Run darker on the repository root (the ./), only modify lines that differ from the given revision (-r)
# In your case, the revision to use is the current master
# You can also get this hash by checking the logs of the failed CI job
darker -r 60625f241f298b5039cb2debc365db38aa7bb522 .

# Commit & push the changes
```

Make sure you are only "darkening" your own changes.

---

Also, you need to add a "what's new entry": see [`ipython/docs/source/whatsnew/pr/`](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr).

MrMino (changes requested):
You need to run `darker` on your changes for CI to pass. Here's how:

```
# Run this in the repository root
# Install darker
pip install darker
# Run darker on the repository root (the ./), only modify lines that differ from the given revision (-r)
# In your case, the revision to use is the current master
# You can also get this hash by checking the logs of the failed CI job
darker -r 60625f241f298b5039cb2debc365db38aa7bb522 .

# Commit & push the changes
```

Make sure you are "darkening" only your own changes.

---

Also, you need to add a "what's new entry": see [`ipython/docs/source/whatsnew/pr/`](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr).

iwanb:
Ok thank you, all green now.

I missed the pr doc entry in `CONTRIBUTING.rst` but I don't see about CI and style checks, it might be handy to add some pointers (to be honest I only skimmed quickly through it though).

IPython/terminal/debugger.py
```diff
@@ -7,6 +7,7 @@
 from IPython.core.completer import IPCompleter
 from .ptutils import IPythonPTCompleter
 from .shortcuts import create_ipython_shortcuts
+from . import embed
```
MrMino, line 10: For consistency:

```suggestion
from .terminal.embed import embed
```

MrMino (approved):
I don't know if there are any unwanted consequences of using `embed.InteractiveShellEmbed`, other than that - LGTM

Tested with:
```py
In [2]: def debugme():  
   ...:     a = 123     
   ...:     breakpoint()
   ...:                 
```

Seems to isolate the scopes nicely.

IPython/terminal/debugger.py
iwanb: That's not the same import, I can change it to one of those if that's prefered:
```
from IPython.terminal import embed
from IPython import embed
import IPython.terminal.embed as embed
```

I can't import only the class I need, which I'd normally do, I think because it creates a circular import from ``IPython/__init__.py``.

iwanb: Actually the circular import is embed -> interactiveshell -> debugger -> ember.

MrMino: Oh, wait, you're right, brainfart on my part.

MrMino: @iwanb Brainfart on my part - please disregard :).

IPython/terminal/debugger.py
```diff
@@ -131,6 +132,18 @@ def in_thread():
         except Exception:
             raise
 
+    def do_interact(self, arg):
+        ipshell = embed.InteractiveShellEmbed(
+            config=self.shell.config,
+            banner1="*interactive*",
+            exit_msg="now exiting InteractiveConsole...",
```
MrMino, line 139: Maybe this one small change:

```suggestion
            exit_msg="*exiting interaractive...*",
```

Carreau:
Thanks !

iwanb:
Thank you!

PR #12921, opened 2021-04-17
Fix docstring syntax to be properly parsed by numpydoc.

Carreau:


MrMino (approved):
LGTM. Also, this might be useful: https://github.com/PyCQA/pylint/blob/master/pylint/extensions/docparams.rst

Carreau:
Thanks for the link. I'm trying to focus on autofixing for now, I don't want to hurdle contributors with yet another thing that can fail in CI.

PR #12932, opened 2021-04-30
remove use of deprecated pipes module

minrk:
pipes.quote is actually an alias to shlex.quote, so use its true name instead

pipes is (probably) deprecated by [PEP 594](https://www.python.org/dev/peps/pep-0594), to be removed possibly in 3.10.

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -213,7 +213,7 @@ class ScriptMagics(Magics):
                     *cmd,
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
-                    stdin=asyncio.subprocess.PIPE
+                    stdin=asyncio.subprocess.PIPE,
                 )
             )
         except OSError as e:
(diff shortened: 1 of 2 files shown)


MrMino (approved):

MrMino:
Weird. Why is `darker` linting `IPython/core/magics/script.py`?

```
Run darker -r 60625f241f298b5039cb2debc365db38aa7bb522 --check --diff . || (
--- IPython/core/magics/script.py
+++ IPython/core/magics/script.py
@@ -213,7 +213,7 @@
                     *cmd,
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
-                    stdin=asyncio.subprocess.PIPE
+                    stdin=asyncio.subprocess.PIPE,
                 )
             )
         except OSError as e:
Error: Process completed with exit code 1.
```

Carreau:
LIkely because it lints all the differences since commit 60625f241f298b5039cb2debc365db38aa7bb522, and the recently new version of black had a few changes.

PR #12937, opened 2021-04-30
Drop support for numpy 1.16

Carreau:
We said we would do that some time ago and havent.

Likely for 7.24


ivanov (approved):
👍 seems like a low-risk change, given that this is for the ipython[test] only

Carreau:
Thanks !

PR #12938, opened 2021-04-30
Remove release trigger.

Carreau:
We don't do github release and the action is broken anyway


ivanov (approved):
I love the smell of yaml going :fire: :fire: :fire:

PR #12952, opened 2021-05-06
Ensure %recall does not show error after argument eval

### LINKED ISSUE #12946: %recall mistakenly prints "Couldn't evaluate or find in history"
## Steps to reproduce

* Execute `a = 1`
* Execute `%recall a`

## Expected result
"1" written in the input prompt, no errors

## Actual result
"1" written in the input prompt, but the following message is printed:
```
Couldn't evaluate or find in history: a
```

## System info
Manjaro Linux, Python 3.9.4, IPython 7.22.0 (reproduces on IPython 8.0.0.dev)

MrMino:
Fixes #12946

### DIFF
--- a/IPython/core/magics/history.py
+++ b/IPython/core/magics/history.py
@@ -282,6 +282,7 @@ class HistoryMagics(Magics):
                 return
         else:
             self.shell.set_next_input(cmd.rstrip())
+            return
         print("Couldn't evaluate or find in history:", arg)
 
     @line_magic


MrMino:
Am I reading this correctly? Has `darker` flipped over on this code?

```
Run darker -r 60625f241f298b5039cb2debc365db38aa7bb522 --check --diff . || (
  File "/opt/hostedtoolcache/Python/3.8.10/x64/bin/darker", line 5, in <module>
    from darker.__main__ import main
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/darker/__main__.py", line 16, in <module>
    from darker.import_sorting import apply_isort, isort
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/darker/import_sorting.py", line 22, in <module>
    isort_code = getattr(isort, "code")
AttributeError: 'NoneType' object has no attribute 'code'
```

IPython/core/magics/history.py
```diff
@@ -282,6 +282,7 @@ def recall(self, arg):
                 return
         else:
             self.shell.set_next_input(cmd.rstrip())
+            return
         print("Couldn't evaluate or find in history:", arg)
```
ivanov, line 286: I believe that this new return on 285 makes the print on the next line unreachable. With this patch, either you return as part of `except`, or as part of the `else`. So the next question is - under which circumstances _should_ we be printing the "Couldn't evaluate or find n history" line?

ivanov:
The believe that the `darker` error is [caused by isort not being installed](https://github.com/akaihola/darker/pull/132)

ivanov:
> The believe that the `darker` error is [caused by isort not being installed](https://github.com/akaihola/darker/pull/132)

now fixed on master, so you won't get any trouble from darker once you rebase.
Happy hacking! :bowtie:

IPython/core/magics/history.py
MrMino: The `print` is still reachable. Look closely at what L279 does - it skips the L282 `return`, and after exiting the loop you land on L286. I could've improved this whole thing to streamline it and make it readable, but I'm still not sure what is the meaning of the condition on this `if ... : continue`.

ivanov: ah, I see, you're right.

PR #12940, opened 2021-05-01
IPython.core.pylabtools: fix configure_inline_support import

ahesford:
The deprecated function `IPython.core.pylabtools.configure_inline_support` attempts to import `configure_inline_support_orig` from the module `matplotlib_inline.backend_inline`, but the actual function name is `configure_inline_support`. Correct the import.

PR #12953, opened 2021-05-07
fix darker version 1.2.3 linting

ivanov:
it looks like the current version of darker doesn't work properly if
isort isn't installed - see https://github.com/akaihola/darker/pull/132
so let's just install isort to fix it.

found via https://github.com/ipython/ipython/pull/12952#issuecomment-833855862 and https://github.com/ipython/ipython/pull/12940#issuecomment-832757006

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


ivanov:
sure enough - import formatting wasn't fully black compliant while darker stopped working, so had to include an extra patch for that here.

Carreau:
<3

PR #12809, opened 2021-02-04
Allow passing a custom CachingCompiler to the interactive shell

martinRenou:
cc. @johanmabille @SylvainCorlay @jtpio

This permits customization of the filename computation prior to caching/compiling. This is needed if we want to have JupyterLab debugger support in the future, as the debugger protocol relies on a cell "filename" in order to track breakpoints added to non-executed cells.

In the JupyterLab debugger implementation, this filename is a computed hash of the raw cell code, which must be the same as the filename used when executing the code in the kernel in order for debugpy to stop on breakpoints.

As a proof of concept, I pushed [a PR to xeus-python which relies on IPython](https://github.com/jupyter-xeus/xeus-python/pull/400) for the code execution/completion. The xeus-python PR builds against this one, and allows starting a debugging session!

ipython_debug

**Note:** My PR contains multiple trailing spaces removal, review from this link to remove noise: https://github.com/ipython/ipython/pull/12809/files?diff=split&w=1

PR #12955, opened 2021-05-07
Changing the compiler that caches to the same one that works with AST

Sammyalhashe:
I just noticed that there is a chance that `run_async_cell` would use a different compiler for caching and visiting the AST. Just felt weird to me.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3122,7 +3122,7 @@ class InteractiveShell(SingletonConfigurable):
         _run_async = False
 
         with self.builtin_trap:
-            cell_name = self.compile.cache(
+            cell_name = compiler.cache(
                 cell, self.execution_count, raw_code=raw_cell
             )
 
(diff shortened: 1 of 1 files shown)


ivanov:
Looks like this was last modified by @martinRenou in #12809 - so perhaps you can shed some light on this, Martin? Before that, this code was from @takluyver 8 years ago, so perhaps Thomas wants to weigh in.

martinRenou:
Yeah, I also noticed this when making #12809 but I decided to not change it because I honestly didn't understand properly the difference between `self.compile` and `compiler` (see https://github.com/ipython/ipython/pull/12809/files#diff-f0df2d6942c603793b48dc8ba2b3032e7ebedcf05a2d65883c374b5b60ef2eb0R3112).

Sammyalhashe:
@Carreau Thanks for the heads up about the email. And I see, so this is pretty much a remnant of history. Thanks for the explanation! You can merge if you want.

PR #12969, opened 2021-05-10
Implement more predicates to skip frames in debugger.

Carreau:
This refactors the code to have distinct predicates, give them
individual flags; and add the default predicates that read-only files
should be hidden.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -33,6 +33,7 @@ import linecache
 import sys
 import warnings
 import re
+import os
 
 from IPython import get_ipython
 from IPython.utils import PyColorize
@@ -281,6 +282,9 @@ class Pdb(OldPdb):
         self.prompt = prompt
         self.skip_hidden = True
 
+        # list of predicates we use to skip frames
+        self._predicates = {"tbhide": True, "readonly": True, "ipython_internal": True}
+
     def set_colors(self, scheme):
         """Shorthand access to the color table scheme selector method."""
         self.color_scheme_table.set_active_scheme(scheme)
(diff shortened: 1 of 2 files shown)


IPython/core/debugger.py
```diff
@@ -532,15 +551,62 @@ def print_list_lines(self, filename, first, last):
...
+
+        """
+        if not args.strip():
+            print("current predicates:")
+            for (p, v) in self._predicates.items():
+                print("   ", p, ":", v)
+            return
+        type_value = args.strip().split(" ")
+        if len(type_value) != 2:
+            print("Usage: skip_predicates <type> <value>")
```
MrMino, line 574: Nit: maybe this should mention possible `<type>`s, just to make it simpler in case someone makes a typo?

IPython/core/debugger.py
```diff
@@ -292,6 +296,26 @@ def set_trace(self, frame=None):
...
+
+        if self._predicates["readonly"]:
+            fname = frame.f_code.co_filename
+            # we need to check for file existence and interactively define
+            # function would otherwise appear as RO.
+            if os.path.isfile(fname) and not os.access(fname, os.W_OK):
+                return True
+
+        if self._predicates["tbhide"]:
+            if frame in (self.curframe, getattr(self, "initial_frame", None)):
```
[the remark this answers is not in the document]
Carreau: ha, I never used debug like so, only for post-mortem. Yep definitely something we can improve later.

IPython/core/debugger.py
Carreau: I've added the possible types, but if the user does a typo it will say:

```
ipydb> skip_predicates a bc
'a' not in {'readonly', 'ipython_internal', 'tbhide'}
```

So this is more if they use an incorrect syntax by putting only one value.

MrMino (approved):

PR #12460, opened 2020-07-23
Make stop_here actually stop

impact27:
The return was not False in case of `__tracebackhide__`

PR #12999, opened 2021-05-30
Allow to stop in frames when _wait_for_mainpyfile is defined

ccordoba12:
- This fixes the Spyder debugger after the release of 7.24.0. Otherwise, all frames are skipped and the debugger exits immediately for us.
- I did this PR against `7.x` because these lines are not present in `master`.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -765,8 +765,6 @@ class Pdb(OldPdb):
         hidden = False
         if self.skip_hidden:
             hidden = self._hidden_predicate(frame)
-            if self._wait_for_mainpyfile:
-                return False
         if hidden:
             Colors = self.color_scheme_table.active_colors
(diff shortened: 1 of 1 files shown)


Carreau:
Hum, I need to look at the PR that added those (#12460) it might have targeted 7.x instead of master,  and have a proper fix. It was added "only" 7 minor releases ago.

ccordoba12:
Ok, pinging @impact27 about this one then (he's a Spyder core dev as well).

impact27:
Looking at #12460 I dont think the following lines should have been added to 7.x:
```
if self._wait_for_mainpyfile:
    return False
```

Carreau:
Ok, merging.

ccordoba12:
Thanks @impact27 for your help and @Carreau for merging!

PR #13000, opened 2021-05-30
Allow to control if reporting hidden frames

ccordoba12:
- This allows downstream projects to decide if they want to show those messages to users.
- We'd really appreciate if this functionality could be included in 7.24.1, so we could use it in Spyder along with PR #12999.
- If this is accepted, I'll create another PR for master because the `stop_here` code is very different there and can't be easily backported.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -282,6 +282,7 @@ class Pdb(OldPdb):
         # Set the prompt - the default prompt is '(Pdb)'
         self.prompt = prompt
         self.skip_hidden = True
+        self.report_skipped = True
 
         # list of predicates we use to skip frames
         self._predicates = {"tbhide": True, "readonly": False, "ipython_internal": True}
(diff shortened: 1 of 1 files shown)


MrMino:
Do I understand correctly: this is different from `skip_predicates` functionality in that you want to skip the frames, but still report them?

ccordoba12:
Hey @MrMino, yes, this is different from `skip_predicates`. We want to skip some frames but *not* to report that they were skipped. Right now those messages are reported for every hidden frame, but I think that could be confusing to our users so we'd like to control it.

MrMino:
> We want to skip some frames but not to report that they were skipped

Ah, I got it backwards. Thanks.

Carreau:
+1 I'll try to cut a 7.24.1 soon with this and the other, likely this Friday.

ccordoba12:
Thanks a lot @Carreau!

Carreau:
@meeseeksdev backport to master

PR #13004, opened 2021-06-02
Forward port PR #13000: Allow to control if reporting hidden frames

Carreau:


ccordoba12:
Thanks for your help with this one @Carreau!

PR #13009, opened 2021-06-04
Fix formatting typo in `%time` docstring

wjandrea:
This paragraph shouldn't be formatted as code.

I have barely any experience with Sphinx, so please check my work before pulling.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1235,11 +1235,11 @@ class ExecutionMagics(Magics):
           CPU times: user 0.00 s, sys: 0.00 s, total: 0.00 s
           Wall time: 0.00
 
-          Note that the time needed by Python to compile the given expression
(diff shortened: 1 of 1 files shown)


MrMino:
Unfortunately, this does not fix it:

image

You might want to try to build the docs and see it for yourself. It's very simple to do that - take an environment (virtual or otherwise), and from the repository root perform the following commands:

```
pip install -r docs/requirements.txt
make -C docs
```

The generated docs will be in `docs/build/html/`. You can either do `xdg-open docs/build/html/index.html` for the index page or `xdg-open docs/build/html/interactive/magics.html` for the page in question.

MrMino:
A hint about Sphinx: in RST, these code blocks are designated by `::` - I don't remember how to break them up though.

MrMino (changes requested):
As mentioned above - this change does not fix the issue in the docs.

(I forgot to mark my comments as a review).

PR #13021, opened 2021-06-15
Don't access current_frame f_locals

### LINKED ISSUE #13020: accessing f_locals in debugger resets namespace
This is the pendant of https://github.com/deshaw/pyflyby/pull/144, see there for a longer explanation.

We do access `current_frame.f_locals` in a couple of place which reset the namespace changes.

Carreau:
This should fix #13020

Though it's a suttle enough bug that tests needs to be written.
And we can't rely on doctests as they are not ran.

### DIFF
--- a/IPython/core/tests/test_debugger.py
+++ b/IPython/core/tests/test_debugger.py
@@ -277,14 +277,14 @@ def test_xmode_skip():
 
     block = dedent(
         """
-def f():
-    __tracebackhide__ = True
-    g()
+    def f():
+        __tracebackhide__ = True
+        g()
 
-def g():
-    raise ValueError
+    def g():
+        raise ValueError
 
-f()
+    f()
     """
     )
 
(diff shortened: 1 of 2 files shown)


MrMino:
Is this related to https://bugs.python.org/issue5215?

I wonder why I never noticed this.

Carreau:
This is probably related, but Pdb in cpython use self.curframe_locals as well, so it's likely on purpose.

Carreau:
Test added.

IPython/core/tests/test_debugger.py
```diff
@@ -321,3 +321,69 @@ def g():
     child.expect("ipdb>")
 
     child.close()
+
+
+def test_where_erase_value():
+    """Test that `where` does nto access f_locals and erase values."""
```
MrMino, line 327:
```suggestion
    """Test that `where` does not access f_locals and erase values."""
```

MrMino:
For anyone interested, and for the future me - it's the call to `PyFrame_FastToLocalsWithError` in `frame_getlocals`, i.e. the C implementation of `frame.f_locals` "property", that refreshes it.

This has few other funny quirks that might bite us in the future, if pdb fixes any and we manage to shadow the patch in ipdb. There's a draft PEP for fixing the `f_locals` semantics that mentions them in the references: [PEP-558](https://www.python.org/dev/peps/pep-0558/), [related paragraph](https://www.python.org/dev/peps/pep-0558/#resolving-the-issues-with-tracing-mode-behaviour).

PR #13027, opened 2021-06-24
allow for spaces in magic arguments, fixes #12729

### LINKED ISSUE #12729: Unable to pass argument with space to cell magic
Cython defines this argument for the `%%cython` margin:

```
    @magic_arguments.argument(
        '--link-args', action='append', default=[],
        help="Extra flags to pass to linker via the `extra_link_args` "
             "Extension flag (can be specified  multiple times)."
    )
```

But trying to pass `%%cython --link-args="-L/Users/carlos/Desktop/Test LGBM API"` I get:

```
UsageError: unrecognized arguments: LGBM API"
```

If I mimic what I believe this should be doing, I get:

```
In [15]: l = shlex.shlex('%%cython --link-args="-L/Users/carlos/Desktop/Test LGBM API"', posix=False)

In [16]: a.whitespace_split=True

In [17]: a.commenters=''

In [18]: list(l)
Out[18]:
['%',
 '%',
 'cython',
 '-',
 '-',
 'link',
 '-',
 'args',
 '=',
 '"-L/Users/carlos/Desktop/Test LGBM API"']
```

which seems fine.

My system is MacOS 11.0.1. I'm using python 3.8.6 with ipython 7.19.0.

ivanov:
- [x] a failing test
- [x] new code
- [x] change to tests that break with the new code, due to backwards incompatibility
- [x] documentation of backwards incompatible change

### DIFF
--- a/IPython/core/tests/test_magic_arguments.py
+++ b/IPython/core/tests/test_magic_arguments.py
@@ -116,3 +116,16 @@ def test_magic_arguments():
     assert_equal(real_name(foo), 'foo')
     assert_equal(foo(None, ''), argparse.Namespace(foo=None))
     assert hasattr(foo, 'has_arguments')
+
+
+@magic_arguments()
+@argument('filepath', type=str, nargs=1, help='An argument with possible spaces')
+def magic_one_arg(args):
+    return parse_argstring(magic_foo1, args)
+
+def test_spaces_and_quotes():
+    "Test for GH-12729"
+    assert_equal(parse_argstring(magic_one_arg, "\"this is one arg\""),
+        argparse.Namespace(filepath=["this is one arg"]))
+    assert_equal(parse_argstring(magic_one_arg, "this\ is\ also\ one\ arg"),
+        argparse.Namespace(filepath=["this is also one arg"]))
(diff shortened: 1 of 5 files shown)


ivanov:
Windows :boom:

bollwyvl:
> Windows :boom:

every dang time, man

IPython/core/tests/test_magic_arguments.py
```diff
@@ -116,3 +116,21 @@ def test_magic_arguments():
     assert_equal(real_name(foo), 'foo')
     assert_equal(foo(None, ''), argparse.Namespace(foo=None))
     assert hasattr(foo, 'has_arguments')
+
+
+@magic_arguments()
+@argument("filepath", type=str, nargs=1, help="An argument with possible spaces")
+def magic_one_arg(args):
+    return parse_argstring(magic_foo1, args)
```
MrMino, line 124: I think you meant..?

```suggestion
    return parse_argstring(magic_one_arg, args)
```

MrMino:
> I have a feeling that posix=True might break under Windows 

I didn't realize `strict=False` makes it fallback on the non-win32 implementation, I seriously thought this would be an ignored flag on win32 :smile:.

Does it actually work? I have no way of checking how the call stack looks like, I don't have any win32 on hand :stuck_out_tongue:.

IPython/core/tests/test_magic_arguments.py
ivanov: good catch, of course! that was my copy-pasta!

Carreau:

Issue #13039, opened 2021-06-30
Licensing issue with Stackoverflow's CC-BY-SA License


uptickmetachu:
Hi, thank you for this amazing tool!

We have recently been FOSS audited and this package was flagged with potential licensing issues. 

Code copied/distributed and/or adapted from [Stackoverflow requires](https://stackoverflow.com/help/licensing):
1. attribution and
2. **licensing under the same license as the original ([CCA-BY-SA-3.0](https://creativecommons.org/licenses/by-sa/3.0/) or [CCA-BY-SA-4.0](https://creativecommons.org/licenses/by-sa/4.0/)).** 
This repositories BSD-3 License is not compatible with CCA-BY-SA.

The code that is causing issue:
https://github.com/ipython/ipython/blob/a2c2a2dcec36224c7dca561efcf9ed9e5c514c3c/IPython/utils/text.py#L408

Consider changing the implementation of the function such that the attribution isn't required.

Carreau:
> 2\. licensing under the same license as the original ([CCA-BY-SA-3.0](https://creativecommons.org/licenses/by-sa/3.0/) or [CCA-BY-SA-4.0](https://creativecommons.org/licenses/by-sa/4.0/)).

Original post if pre 2011 so will be CC-BY-SA-2.5 (also you have extra A in CCA).
The effect is still the same though IIUC.

MrMino:
TBH I don't find this code necessary. If I understand it correctly, the part required is finding the common _prefix_, not the common substring, so this is unnecessarily quadratical.

I'll refactor it and post a PR today.

Carreau:
Actually this is used in only one place, and to find the longest common prefix that only contain whitespace and `>`. So it can be replaced by something much faster and simpler.

Carreau:
Ha, simultaneous comments :-)

uptickmetachu:
Thank you for your quick replies! Sorry about the typo in the original post.

MrMino:
Fixed in #13041.

MrMino:
@uptickmetachu FYI 7.26 which fixed this issue has been released today.

uptickmetachu:
Thank you very much much!

PR #13041, opened 2021-06-30
Refactor `IPython.utils.text.strip_email_quotes`

MrMino:
Also removes long_substr, which fixes the licesing issue mentioned
in #13039.

There's also a small nuance here in that the previous subroutine didn't remove all whitespace characters past the first `>` character - this one does, so it _will_ strip prefixes like `> \t\t\t >`.

### DIFF
--- a/IPython/utils/text.py
+++ b/IPython/utils/text.py
@@ -10,6 +10,7 @@ Inheritance diagram:
 
 import os
 import re
+import string
 import sys
 import textwrap
 from string import Formatter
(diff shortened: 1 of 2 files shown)


IPython/utils/text.py
```diff
@@ -447,27 +432,30 @@ def strip_email_quotes(text):
...
-        prefix = re.match(r'^(\s*>[ >]*)', line)
-        if prefix:
-            matches.add(prefix.group(1))
+    strip_len = 0
+
+    for characters in zip(*lines):
+        # Check if all characters in this position are the same
+        if len(set(characters)) > 1:
+            break
+        prefix_char = characters[0][0]
```
Carreau, line 448:
```suggestion
        prefix_char = characters[0]
```
I guess ? 

If characters is `('a', 'e', 'i', 'o', 'u')`, then just a single `[0]` is needed. The second 0, is just because you index the first char of a string, which is the identity, as we have a tuple of strings of len 1.

unless I misunderstood.

Carreau:
Nice implementation, I would have done more complicated. 
One minor suggestion.

IPython/utils/text.py
MrMino: Yes, you're right, I made a thinko here. Thanks!

MrMino:
@Carreau I've backported this by pushing to origin instead of my fork by mistake, I'm on a PC where I didn't have `no_push` set 😕. Sorry for breaking the established process. If I need to roll it back (maybe if meeseeks needs it?), just ping me.

Carreau:
@MrMino, no that's ok, those things happen. We just try to keep things in PR for visibility.

Personally I clone origin over https to be sure to not be able to push, and my fork over ssh.

PR #13043, opened 2021-07-02
Improve error tracebacks so that it shows the cell prompts

martinRenou:
This PR improves the error traceback so that it shows the prompt `In [x]` in front of code coming from cells in Jupyter Notebook and IPython, instead of the filename hash (used for compilation).

The result is the following:

traceback

Versus the old version:

vs

### DIFF
--- a/IPython/core/compilerop.py
+++ b/IPython/core/compilerop.py
@@ -92,6 +92,10 @@ class CachingCompiler(codeop.Compile):
         # (otherwise we'd lose our tracebacks).
         linecache.checkcache = check_linecache_ipython
 
+        # Caching a dictionary { filename: execution_count } for nicely
+        # rendered tracebacks. The filename corresponds to the filename
+        # argument used for the builtins.compile function.
+        self._filename_map = {}
 
     def ast_parse(self, source, filename='<unknown>', symbol='exec'):
         """Parse code to an AST with the current compiler flags active.
@@ -153,6 +157,10 @@ class CachingCompiler(codeop.Compile):
             raw_code = transformed_code
 
         name = self.get_code_name(raw_code, transformed_code, number)
+
+        # Save the execution count
+        self._filename_map[name] = number
+
         entry = (
             len(transformed_code),
             time.time(),
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -160,7 +160,7 @@ def _format_traceback_lines(lines, Colors, has_colors, lvals):
         else:
             num = '%*s' % (numbers_width, lineno)
             start_color = Colors.lineno
-        
+
         line = '%s%s%s %s' % (start_color, num, Colors.Normal, line)
 
         res.append(line)
@@ -300,7 +300,7 @@ class ListTB(TBTools):
 
     Calling requires 3 arguments: (etype, evalue, elist)
     as would be obtained by::
-    
+
       etype, evalue, tb = sys.exc_info()
       if tb:
         elist = traceback.extract_tb(tb)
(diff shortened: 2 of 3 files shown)


Carreau:
Seem reasonable to me.

IPython/core/compilerop.py
```diff
@@ -92,6 +92,9 @@ def __init__(self):
         # (otherwise we'd lose our tracebacks).
         linecache.checkcache = check_linecache_ipython
 
+        # Caching a dictionary { filename: execution_count } for nicely
+        # rendered tracebacks
+        self.filename_map = {}
```
MrMino, line 97: IIUC this isn't "execution count", it's the cell number. To me `execution_count` suggests, that this dictionary is used to remember how many times code with a given filename was run/compiled/etc. - it's a bit misleading.

Similarly, it's not _really_ the filename that is being remembered here as a key - it's a synthetic identifier that IPython gives its code blocks in the `code_name` func above. I think it would be worth it to try to try and signify it in the name of this variable, for posterity.

Maybe:

```suggestion
        # Caching dictionary { filename: cell_id } for nicely
        # rendered tracebacks
        self.co_filename_to_cell_id = {}
```

`co_filename` is the name of the attribute that the code objects use to keep this data.

MrMino: You could also get rid of this dictionary altogether, by creating a `code_name_to_cell_id` func (next to `code_name`), that would parse out the `number` out of the:

```
    return '<ipython-input-{0}-{1}>'.format(number, hash_digest[:12])
```

Then add a test for the `code_name_to_cell_id(code_name(code, number)) == number` assumption.
That way you could omit the `get_ipython` call, and therefore make it all a bit less coupled.

MrMino (commented):
Apart from the one suggestion above - looks cool.

MrMino:
@martinRenou Is it ok with you if I modify your PR to fit my suggestion? This is one of those good looking nice-to-haves that I cannot resist pushing through, but I'd really like to leave `CachingCompiler` stateless, as it was before.

IPython/core/compilerop.py
martinRenou: Your suggestion is taking the assumption that the filename is using the format `<ipython-input-{0}-{1}>` which is not the case in ipykernel, nor in xeus-python as they are not using the same `code_name` function. It's only the case when using the ipython shell directly.

MrMino: :/ I honestly wasn't aware this has impact on these projects.

That's a shame. It would be really unfortunate to make `CachingCompiler` responsible for remembering cell IDs. It just seems like a code smell to me.

I guess there isn't any better way currently to achieve this. I might go on a little crusade afterwards to put this dict somewhere else, just to ensure the stateless class remains stateless, but that's outside the scope of this PR.

BTW. do you happen to know where `xeus-python` depends on this?

MrMino, line 97: Let's just take steps to ensure this dict won't get used somewhere else. Also, `filename_map` is a bit misleading here -  these aren't filenames after all ;).

```suggestion
        self._co_filename_map = {}
```

martinRenou: ipykernel and xeus-python provide their own `CachingCompiler` class to the shell, allowing them to use their own hash for the filename.

This is needed for the compatibility with the JupyterLab debugger. When stopping at a breakpoint, the filename used for AST compilation must refer to an actual file saved in a temporary folder for the debugger to retrieve the source.

> That's a shame. It would be really unfortunate to make CachingCompiler responsible for remembering cell IDs. It just seems like a code smell to me.

Why do you think it's a code smell? And why do you think it's important to have the `CachingCompiler` stateless?

I guess we could store the `{ filename: cell_id } ` mapping in the shell. But the caching compiler being already responsible for choosing the filename given the cell id, I find it to be the appropriate place.

martinRenou: Filename here refers to the `filename` argument passed to the [compile function](https://docs.python.org/3/library/functions.html#compile) used for AST compilation.

I agree, let's make it more private.

martinRenou: Also, the fact that the `CachingCompiler` actually caches data during the session does not sound crazy to me. It already manipulates the Python standard's `linecache` module which is also used for error tracebacks if I remember correctly.

MrMino: > Why do you think it's a code smell? And why do you think it's important to have the CachingCompiler stateless?

I guess it's 90% a matter of taste, but looking through the history of this code, the `Caching` was supposed to mean `manipulates the linecache`, not `holds data related to Code objects`. It just looks like this class is inching towards another responsibility, and it already has one that is quite difficult and technical. I would've liked it to stay stateless solely because its an interface to what looks like a hack, and knowing that it is stateless makes this hack that little bit easier to understand when reading the code.

This is in no way a deal breaker, but rather just me ensuring there's really no better way to do this. Seems like currently there isn't. If I'll find other places in the codebase where this kind of "find info about code cells" functionality is attached, I'll probably want to make a separate place for it.

MrMino: If I saw this, it would be easier to connect the dots if it told me that it's related to `co_filename`, but it's probably just me having done one too many frame hacks in python.

Up to you :).

MrMino:
Oh - one more thing - it would be nice to have a short [what's new entry](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr) for this.

psychemedia:
This looks like a really handy update to the error report and makes it easier to find the code cell(s) related with an issue. I note that the [`friendly-traceback`](https://friendly-traceback.github.io/docs/jupyter.html) also employs such a trick, along with a range of user friendly error explanations:

image

MrMino:
@martinRenou would you mind me finishing this up?

martinRenou:
Will do. I'm on vacation for the next two weeks and won't be close to a proper keyboard, but I'll take your review into account when I'm back.

MrMino:
Ok, thanks! In case you'd like to leave something to me I'm happy to finish it of :).

martinRenou:
Done :)

MrMino:
@martinRenou there was a minor issue in your what's-new entry RST. I fixed it.

In RST code blocks are designated by `::`, so unless it's there, the `-----` in your code snippets will be seen by RST parser as a section header.

MrMino (approved):

MrMino:
All good! Thanks @martinRenou!

martinRenou:
Thank you!

Carreau:
I'll have to add an option to disable this, this screws up doctest of some projects, as now you start to get `In [xxx]` in outputs, and that messes up the normalizers/things that try to split expected outputs.

martinRenou:
Ok 😕 thanks!

Note that there were changes related to this and SyntaxErrors there as well: https://github.com/ipython/ipython/pull/13088

Carreau:
I yes, good point, thanks. 

I may just prepend with "Interactive input", so that it can be distinguished. It just seem problematic when it's at beginning of the line.

PR #13048, opened 2021-07-08
Add `.mypy_cache` to .gitignore

MrMino:


Carreau:
Thanks, if you often want to exclude a file,  you also may want to look into [`.gitgnore_global`](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

PR #12854, opened 2021-03-09
Hist rangepattern Issue 12451

DanielGoldfarb:
This enhancement is for https://github.com/ipython/ipython/issues/12451

I have implemented the ability to specify a range of sessions/lines simultaneously with a `-g` glob pattern, for example:

`hist ~4/1-~0/10 -g <pattern>`  to glob for `<pattern>` only from line 1 of four sessions ago, until the present session line 10.

PR #13049, opened 2021-07-08
Accept empty history ranges

MrMino:
Makes it possible to use the following magics with empty history ranges:
 - `%save`
 - `%load`
 - `%pastebin`
 - `%pycat`

When an empty range is given, these magics will fetch the whole history up to the point of the magic call (so the call itself is not included).

This enables a number of handy shortcut usages (see [whats-new entry](https://github.com/ipython/ipython/pull/13049/files#diff-85ef0b54c47109325d8764a709cd935b196567cbeb61f48aae8fe709e658d14bR16-R19)).

It is implemented by adding two simple modalities to existing code:
 - [`IPython.core.history.extract_hist_ranges`](https://github.com/ipython/ipython/pull/13049/files#diff-1dfe2aeed55927e6e7dfafed248585e56699f46feb90d7174c0e51f7843e102eR865-R868) will now return a range corresponding to "current session, whole history" (`[(0, 1, None)]`), if given an empty string as `ranges_str`.
 - [`IPython.core.interactiveshell.InteractiveShell.extract_input_lines`](https://github.com/ipython/ipython/pull/13049/files#diff-f0df2d6942c603793b48dc8ba2b3032e7ebedcf05a2d65883c374b5b60ef2eb0L3718-R3730), when given an empty string in `range_str`, uses the above to fetch the whole history, and then remove the last line (the one with the magic call).
 
These two methods exist in call stacks of other magics (see the flowchart below), but the changes have no impact on them. When given empty hist range or no arguments:
 - `%macro` lists existing macros.
 - `%edit` treats it as "open editor inside an empty temp file".
 - `%history` retains the functionality it had before, i.e. lists the whole history (_including_ the magic call).
 - `%recall` treats it as "take output of the last command".
 - `%rerun` treats it as "rerun last command".

The "take the whole history" `if...else` branch in `%history` has been refactored, since with these changes the `else` part is no longer needed.

image

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -445,8 +445,11 @@ class HistoryAccessor(HistoryAccessorBase):
         Parameters
         ----------
         rangestr : str
-          A string specifying ranges, e.g. "5 ~2/1-4". See
-          :func:`magic_history` for full details.
+          A string specifying ranges, e.g. "5 ~2/1-4". If empty string is used,
+          this will return everything from current session's history.
+
+          See the documentation of :func:`%history` for the full details.
+
         raw, output : bool
           As :meth:`get_range`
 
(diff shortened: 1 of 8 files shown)


Carreau:
That's really good work ! Thanks and sorry for not having a lok earlier, it's my last few days in europe and I'll be more reactive starting next month !

Thanks also for all the triaging and merging !

Carreau:
I believe this rely on #12854, which is not on the 7.x branch, so I'm going to tag 8.0....

MrMino:
I missed that. Thanks!

PR #13050, opened 2021-07-10
Make Ipython.utils.timing work with jupyterlite

dsblank:
JupyterLite has a resource, but not getrusage(). This make the code a little more general.

### DIFF
--- a/IPython/utils/timing.py
+++ b/IPython/utils/timing.py
@@ -23,6 +23,11 @@ import time
 # If possible (Unix), use the resource module instead of time.clock()
 try:
     import resource
+except ImportError:
+    resource = None
+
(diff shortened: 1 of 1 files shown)


MrMino:
I'm not sure if I understand why this needs to be so - could you please elaborate the description? I haven't used jupyterlite yet, much less read through its code.

dsblank:
Sure! The code was previously written so that it would check to see if it could import `resource` and decide to use the fallbacks. jupyterlite has a `resource` module, but it doesn't have a `getrusage` method. This extends the previous behavior to also work in these types of environments.

Thanks for the review!

Carreau:
Thanks both ! Merging this and releasing soon.

PR #13051, opened 2021-07-11
Preserve changes to frame locals when going up/down in the stack frame

ccordoba12:
## Description

This complements the excellent improvements on PR #13021 by preserving frame locals when going up/down in the stack. Below I show gifs of the previous and new behaviors using Spyder:

**Before**

Old

**After**

New

## Implementation

This was accomplished by saving a copy of all frame locals before the debugger starts and using them to set `curframe_locals` after every command run in the debugger. That prevents that accesses to `f_locals` (which happens in up/down) reset the locals set by the user.

## Notes

* I draw inspiration for this from the patch present in https://bugs.python.org/issue9633, although it's very different.
* There's [PEP 558](https://www.python.org/dev/peps/pep-0558/) that should solve this problem the right way, but currently is marked for Python 3.11 and [an implementation](https://github.com/python/cpython/pull/3640) for it has been four years in the making.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -304,6 +304,9 @@ class Pdb(OldPdb):
         # list of predicates we use to skip frames
         self._predicates = self.default_predicates
 
+        # To save frame locals so we can restore them on up/down
+        self.frame_locals = []
+
     def set_colors(self, scheme):
         """Shorthand access to the color table scheme selector method."""
         self.color_scheme_table.set_active_scheme(scheme)
@@ -351,6 +354,17 @@ class Pdb(OldPdb):
             ip_hide = [h if i > ip_start[0] else True for (i, h) in enumerate(ip_hide)]
         return ip_hide
 
+    def preloop(self):
+        """
+        Save a copy of all frame locals so changes to them are not lost
+        when going up/down in the stack frame.
+        """
+        self.frame_locals = [
+            stack_entry[0].f_locals.copy() for stack_entry in self.stack
+        ]
+        self.curframe_locals = self.frame_locals[self.curindex]
+        super().preloop()
+
     def interaction(self, frame, traceback):
         try:
             OldPdb.interaction(self, frame, traceback)
@@ -368,6 +382,11 @@ class Pdb(OldPdb):
         line = super().precmd(line)
 
         return line
+
+    def postcmd(self, stop, line):
+        """Set current frame locals using the ones we saved in preloop."""
+        self.curframe_locals = self.frame_locals[self.curindex]
+        return super().postcmd(stop, line)
 
     def new_do_frame(self, arg):
         OldPdb.do_frame(self, arg)


MrMino (changes requested):
Unfortunately, this causes a regression - it makes it impossible to change the values of the variables in the local scope. The changes are only visible to the debugger, but not to the debugged code.

On `master` branch:
```
Python 3.8.6 (default, Jun 23 2021, 23:50:42) 

In [1]: def print_my_var():
   ...:     my_var = '==== unchanged ===='
   ...:     print(my_var)
   ...: 

In [2]: %debug print_my_var()
> <string>(1)<module>()

ipdb> s
--Call--
> <ipython-input-1-5aa28ff21139>(1)print_my_var()
----> 1 def print_my_var():
      2     my_var = '==== unchanged ===='
      3     print(my_var)
      4 

ipdb> n
> <ipython-input-1-5aa28ff21139>(2)print_my_var()
      1 def print_my_var():
----> 2     my_var = '==== unchanged ===='
      3     print(my_var)
      4 

ipdb> n
> <ipython-input-1-5aa28ff21139>(3)print_my_var()
      1 def print_my_var():
      2     my_var = '==== unchanged ===='
----> 3     print(my_var)
      4 

ipdb> my_var = "==== changed ===="
ipdb> n
==== changed ====
--Return--
None
> <ipython-input-1-5aa28ff21139>(3)print_my_var()
      1 def print_my_var():
      2     my_var = '==== unchanged ===='
----> 3     print(my_var)
      4 
```

With the changes presented in this PR:

```
Python 3.8.6 (default, Jun 23 2021, 23:50:42) 

In [1]: def print_my_var():
   ...:     my_var = '==== unchanged ===='
   ...:     print(my_var)
   ...: 

In [2]: %debug print_my_var()
> <string>(1)<module>()

ipdb> s
--Call--
> <ipython-input-1-5aa28ff21139>(1)print_my_var()
----> 1 def print_my_var():
      2     my_var = '==== unchanged ===='
      3     print(my_var)
      4 

ipdb> n
> <ipython-input-1-5aa28ff21139>(2)print_my_var()
      1 def print_my_var():
----> 2     my_var = '==== unchanged ===='
      3     print(my_var)
      4 

ipdb> n
> <ipython-input-1-5aa28ff21139>(3)print_my_var()
      1 def print_my_var():
      2     my_var = '==== unchanged ===='
----> 3     print(my_var)
      4 

ipdb> my_var = "==== changed ===="
ipdb> n
==== unchanged ====
--Return--
None
> <ipython-input-1-5aa28ff21139>(3)print_my_var()
      1 def print_my_var():
      2     my_var = '==== unchanged ===='
----> 3     print(my_var)
      4 
```

ccordoba12:
Hey @MrMino, thanks a lot for checking that! I'll try to address that problem when I have some free time.

IPython/core/debugger.py
```diff
@@ -351,6 +354,17 @@ def hidden_frames(self, stack):
             ip_hide = [h if i > ip_start[0] else True for (i, h) in enumerate(ip_hide)]
         return ip_hide
 
+    def preloop(self):
+        """
+        Save a copy of all frame locals so changes to them are not lost
+        when going up/down in the stack frame.
+        """
+        self.frame_locals = [
+            stack_entry[0].f_locals.copy() for stack_entry in self.stack
```
madbird1304, line 363: @ccordoba12 , is there any need to copy this dict? 
If there isn't, removing `copy()` could fix the problem mentioned by @MrMino 
```suggestion
            stack_entry[0].f_locals for stack_entry in self.stack
```

ccordoba12: Could you clarify? What's a `ChainMap`?

madbird1304: image

ccordoba12: Ok, thanks a lot for the suggestion! I'll take a look at it as soon as Io have some free time.

Carreau:

Issue #12896, opened 2021-04-08
Built-in magic commands do not escape spaces in path


carlosperate:
It does work with shell commands (e.g. `!pip` instead of `%pip`), but when using the magic commands spaces are not escaped.

In macOS Big Sur, a brand new virtualenv with spaces in its path:
```
$ pip list
Package          Version
---------------- -------
appnope          0.1.2  
backcall         0.2.0  
decorator        5.0.6  
ipykernel        5.5.3  
ipython          7.22.0 
ipython-genutils 0.2.0  
jedi             0.18.0 
jupyter-client   6.1.12 
jupyter-core     4.7.1  
parso            0.8.2  
pexpect          4.8.0  
pickleshare      0.7.5  
pip              19.2.3 
prompt-toolkit   3.0.18 
ptyprocess       0.7.0  
Pygments         2.8.1  
PyQt5            5.15.4 
PyQt5-Qt5        5.15.2 
PyQt5-sip        12.8.1 
python-dateutil  2.8.1  
pyzmq            22.0.3 
qtconsole        5.0.3  
QtPy             1.9.0  
setuptools       41.2.0 
six              1.15.0 
tornado          6.1    
traitlets        5.0.5  
wcwidth          0.2.5  

$ ipython
Python 3.8.2 (default, Jun  2 2020, 13:57:42) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: !pip --version
pip 19.2.3 from /Users/microbit-carlos/workspace/deleteme/really-delete/path with spaces/.venv/lib/python3.8/site-packages/pip (python 3.8)

In [2]: %pip --version
zsh:1: no such file or directory: /Users/microbit-carlos/workspace/deleteme/really-delete/path
Note: you may need to restart the kernel to use updated packages.

In [3]: pip --version
zsh:1: no such file or directory: /Users/microbit-carlos/workspace/deleteme/really-delete/path
Note: you may need to restart the kernel to use updated packages.

```

Closest open issue i've found is this one, but in that it looks like the spaces are escaped incorrectly? https://github.com/ipython/ipython/issues/11138

carlosperate:
It can also fail the same way if a path is not quoted when running shell command:

```
$ ipython
Python 3.8.2 (default, Jun  2 2020, 13:57:42) 
Type 'copyright', 'credits' or 'license' for more information
[...]

PR #13052, opened 2021-07-12
Fix path handling in `pip` line magic

madbird1304:
Fix for issue: https://github.com/ipython/ipython/issues/12896

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -78,7 +78,7 @@ class PackagingMagics(Magics):
           %pip install [pkgs]
         """
         python = _get_full_path(sys.executable)
-        args = ('-m', 'pip', *shlex.split(line))
+        args = ("-m", "pip", *shlex.split(line))
         command = _format_command(python, *args)
 
         self.shell.system(command)
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -10,16 +10,10 @@
 
 import re
 import shlex
-import subprocess
 import sys
 from pathlib import Path
 
 from IPython.core.magic import Magics, magics_class, line_magic
-
-
-def _format_command(*args):
-    """Use subprocess.list2cmdline to quote and escape executable path"""
-    return subprocess.list2cmdline(list(args))
 
 
 def _is_conda_environment():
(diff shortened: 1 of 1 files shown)


IPython/core/magics/packaging.py
```diff
@@ -10,12 +10,23 @@
...
 
 
+def _format_command(*args):
+    """Use subprocess.list2cmdline to quote and escape executable path"""
+    return subprocess.list2cmdline(list(args))
+
+
+def _get_full_path(path):
+    """Return a normalised path of the current Python executable"""
+    return Path(sys.executable).expanduser().absolute().resolve()
```
MrMino, line 27: This is unnecessary. Judging by [the docs](https://docs.python.org/3/library/sys.html#sys.executable), `sys.executable` is already an absolute path, so `expanduser().absolute().resolve()` will have no effect on it.

I don't have macOS (vide the issue), so I cannot check it. Could you please explain this, and maybe show the values of `sys.executable` this is meant to shield against?

madbird1304: You're right about `sys.executable`.
I used this function for testing purposes, with different paths, but now it seems redundant.
Beside of that, I forgot to handle the `path` argument.

IPython/core/magics/packaging.py
```diff
@@ -66,7 +72,11 @@ def pip(self, line):
         Usage:
           %pip install [pkgs]
         """
-        self.shell.system(' '.join([sys.executable, '-m', 'pip', line]))
+        python = sys.executable
+        args = shlex.split(line)
+        command = _format_command(python, "-m", "pip", *args)
```
MrMino, line 77: Why this requires `subprocess.list2cmdline`? `list2cmdline` is the windows-style argument splitting, I'm not sure if it is a good idea to mix it with `shlex.split`. Can you explain why a `' '.join` is not sufficient?

madbird1304: > Can you explain why a `' '.join` is not sufficient?

Using just `' '.join()` we may lose whitespaces in `sys.executable`.

This command
```sh
$ "/home/my user with spaces/.../python" "-m" "pip" "install 123 456" 
```
may become
```sh
$ "/home/my user with spaces/.../python -m pip install 123 456"
```

So, I guess, we have 2 options here:
- quote whole file path
- escape contained whitespaces

MrMino: I'd go with quoting the file path instead. The responsibility of splitting the command properly based on the OS / platform is on the internals of `shell.system`, so it would be better to leave it there.

madbird1304: Updated, please take a look

MrMino (approved):
I tested it and it works as intended. Thank you @madbird1304 for the fix!

Issue #13058, opened 2021-07-13
`%gui qt` support for PyQt6


outofculture:
`%gui qt` in an environment with PyQt6 complains that no qt implementations are available.

```
In [1]: %gui qt
ERROR:root:
    Could not load requested Qt binding. Please ensure that
    PyQt4 >= 4.7, PyQt5, PySide >= 1.0.3 or PySide2 is available,
    and only one is imported per session.

    Currently-imported Qt library:                              None
    PyQt4 available (requires QtCore, QtGui, QtSvg):            False
    PyQt5 available (requires QtCore, QtGui, QtSvg, QtWidgets): False
    PySide >= 1.0.3 installed:                                  False
    PySide2 installed:                                          False
    Tried to load:                                              ['pyqtdefault', 'pyside', 'pyqt5', 'pyside2']
```

I made a stab at implementing this locally, but I can say that just treating PyQt6 like PyQt5 is insufficient:

```
In [1]: %gui qt

In [2]: Traceback (most recent call last):
  File "/home/martin/.miniconda3/envs/qt6/bin/ipython", line 33, in <module>
    sys.exit(load_entry_point('ipython==8.0.0.dev0', 'console_scripts', 'ipython')())
  File "/home/martin/.miniconda3/envs/qt6/lib/python3.9/site-packages/ipython-8.0.0.dev0-py3.9.egg/IPython/__init__.py", line 124, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/home/martin/.miniconda3/envs/qt6/lib/python3.9/site-packages/traitlets/config/application.py", line 845, in launch_instance
    app.start()
  File "/home/martin/.miniconda3/envs/qt6/lib/python3.9/site-packages/ipython-8.0.0.dev0-py3.9.egg/IPython/terminal/ipapp.py", line 356, in start
    self.shell.mainloop()
  File "/home/martin/.miniconda3/envs/qt6/lib/python3.9/site-packages/ipython-8.0.0.dev0-py3.9.egg/IPython/terminal/interactiveshell.py", line 604, in mainloop
    self.interact()
  File "/home/martin/.miniconda3/envs/qt6/lib/python3.9/site-packages/ipython-8.0.0.dev0-py3.9.egg/IPython/terminal/interactiveshell.py", line 587, in interact
    code = self.prompt_for_code()
[...]

PR #13060, opened 2021-07-14
failing attempt at using Qt6 for %gui

outofculture:
From my efforts on #13058, though it's not to the point of working yet.

- [ ] HiDPI ( research where Qt6 keeps this option )
- [ ] exit behavior
- [ ] maintain backwards compatibility
- [ ] test?
- [ ] style?
- [ ] documentation?
- [ ] QA?

Oh, and if it helps, the code I'm using to make a gui is:

```python
import pyqtgraph as pg
pg.plot([0, 1, 1])
```

That should open a simple interactive line plot ( it requires pyqtgraph be installed, though ).


QuLogic:
Is some, or all, of this replaced by #13085?

PR #13090, opened 2021-08-15
Fix: `async with` doesn't allow newlines (ipython/ipython/issues/12975)

madbird1304:


Carreau:
Thanks a lot for working on this !

I'm wondering if the regexp is the best way to check for async await. I believe the re are flags for compile commands, the should allow to detect that, where compile fails, but compiles +allow async works. I'll have to redive into exactly how that works.

madbird1304:
What about reproducing `ast.parse` behavior?
This code doesn't compiles
```python
In [9]: code = '''\
   ...: async with x:
   ...:     pass
   ...: '''
   ...: compile(code, '<console>', 'exec')

  File "/home/madbird/.cache/pypoetry/virtualenvs/ipython-WHLgF_BT-py3.9/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3469, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  File "<ipython-input-9-cb1edaa8f943>", line 5, in <module>
    compile(code, '<console>', 'exec')

  File "<console>", line 1
SyntaxError: 'async with' outside async function
```
but it can be easily parsed by `ast.parse()`
```python
In [10]: import ast
    ...: ast.parse(code)
Out[10]: <ast.Module at 0x7fdf51b12e50>
```
`parse()`, in its turn, just calls `compile()` with `ast.PyCF_ONLY_AST`
```python
In [11]: ast.parse??
Signature:
ast.parse(
    source,
    filename='<unknown>',
    mode='exec',
    *,
    type_comments=False,
    feature_version=None,
)
Source:
def parse(source, filename='<unknown>', mode='exec', *,
          type_comments=False, feature_version=None):
    """
    Parse the source into an AST node.
    Equivalent to compile(source, filename, mode, PyCF_ONLY_AST).
    Pass type_comments=True to get back type comments where the syntax allows.
    """
    flags = PyCF_ONLY_AST
    if type_comments:
        flags |= PyCF_TYPE_COMMENTS
    if isinstance(feature_version, tuple):
        major, minor = feature_version  # Should be a 2-tuple.
        assert major == 3
        feature_version = minor
    elif feature_version is None:
        feature_version = -1
    # Else it should be an int giving the minor version for 3.x.
    return compile(source, filename, mode, flags,
                   _feature_version=feature_version)
File:      /usr/lib/python3.9/ast.py
Type:      function
```
So that, I guess, we found needed flags: we could use`ast.PyCF_ONLY_AST` or `ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`.
The first option seems to be better as long as we don't actually need compiled code object for any purpose, so we could avoid actual compilation.
```python
In [12]: compile(code, '<console>', 'exec', flags=ast.PyCF_ONLY_AST | ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
Out[12]: <ast.Module at 0x7fdf43428dc0>
```
@Carreau , what do you think?

Carreau:
It's quite possibly the right way, I was the one who send PR for Python to get PyCF_ALLOW_TOP_LEVEL_AWAIT so we might have to do that conditionally on the Python version (I belive it works well only on 3.8+)

IPython/core/async_helpers.py also have helpers for maybe async code.

madbird1304:
@Carreau , I've changed the PR accordingly. Could you please review it again? Thank you in advance!

Carreau:
let's try. thanks !

PR #13092, opened 2021-08-15
Deadheading `IPython.utils.py3compat`

MrMino:
During review of another PR I noticed, that there's a bunch of code in `IPython.utils.py3compat` that looks unused / unnecessary.

The module itself is deprecated. I removed most of its usage, then I realized that simply replacing `decode(s)` with `s.decode(errors="replace")` might not be the way to go, given that `DEFAULT_ENCODING` of `decode` isn't necessarily `"utf-8"`.

I'm still looking into it. I'd be really pleased if we'd be able to get rid of this module in 8.0. The way it uses `IPython.utils.encoding` to come up with the default encoding is magical and surprising, and its difficult to wrap one's head around what happens there.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -31,7 +31,6 @@ from IPython.lib.pretty import pretty
 from IPython.testing.skipdoctest import skip_doctest
 from IPython.utils import PyColorize
 from IPython.utils import openpy
-from IPython.utils import py3compat
 from IPython.utils.dir2 import safe_hasattr
 from IPython.utils.path import compress_user
 from IPython.utils.text import indent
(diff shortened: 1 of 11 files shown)


Carreau:
I think most of the magic things were due to Python2, so it maybe that we don't need the `"replace"` anymore.

I think I'll be ok if we try to do a bit less magic in 8.0 and if things break for some people then we can re-add the missing functionalities.

dbivolaru:
Was searching for encoding related items and found this draft.
It seems in `python 3.10` there is a new option to check for ambiguous encodings which I am using for other reasons.

```
PYTHONWARNDEFAULTENCODING=1 python -W error -c 'from IPython.core import ultratb'
```
yields an EncodingWarning in utils/io.py:87 at `devnull = open(os.devnull, 'w')`

Environment: `python3-ipython-7.26.0-2.fc35` `python3-3.10.1-2.fc35`

I just muted the warnings coming from ipython, however I thought it might be useful for this draft.

Carreau:

PR #13093, opened 2021-08-15
Fix #13084

### LINKED ISSUE #13084: Windows %pip magic broken for 7.26 release due to shlex.quote pathing issue
Hi Team,

OS: Windows 10 Pro (build 19041.1110)
Python: 3.9
IPython: 7.26

I found working with PyCharm and demonstrating use of Notebooks on Windows that `%pip install` stopped working this week in new environments and complaining about pathing, the error:

```
The filename, directory name, or volume label syntax is incorrect.
```

The use of `%conda` seemed to work find for conda environments, it was just pip that was failing.

When a new project is created in PyCharm is pulls the latest version of `ipython` for use with Jupyter. Downgrading to 7.25 the functionality of `%pip install` is restored again.

Looking at the release and related pull requests I found this pathing issues for Mac.

https://github.com/ipython/ipython/pull/13052

The path to my virtual env when testing was: `C:\projects\Demos\pycharmNotebook\venv\`

I believe this is an issue with using single quotes for commands in Windows which causes problems and has differences to that of Unix/Mac/Linux systems.

For example, if I have the following notebook running 7.26
```
#%%
import shlex
import sys
import os
with_quote = shlex.quote(sys.executable)
without_quote = sys.executable
full_call_with_quote = " ".join([with_quote,"-m","pip","install servicestack"])
full_call_without_quote = ' '.join([without_quote,'-m','pip','install servicestack'])
print(full_call_without_quote)
print(full_call_with_quote)
os.system(full_call_without_quote)
os.system(full_call_with_quote)
#%%
```

The output of the print statements is:
```
C:\projects\Demos\pycharmNotebook\venv\Scripts\python.exe -m pip install servicestack
'C:\projects\Demos\pycharmNotebook\venv\Scripts\python.exe' -m pip install servicestack
```
And the output from the `os.system` calls is successful for the first, but fails with `The filename, directory name, or volume label syntax is incorrect.` for the second which is the same implementation changed in 7.26. 

Unfortunately it seems the fix for Mac paths with spaces has broken all paths on Windows.

I believe double quotes `"` works correctly on both but I don't have a Mac to easily accessible to test on. Windows WSL running Ubuntu 20.04 works with double quotes, but again I'm not 100% sure on the implications for out of the ordinary pathing.

To anyone experiencing this issue on Windows, downgrade to 7.25 resolves the issue for me.

madbird1304:
Fix for [issue](https://github.com/ipython/ipython/issues/13084) caused by using shlex to quote executable path.
Sorry, [I've introduced this bug](https://github.com/ipython/ipython/pull/13052).
To fix this, let's replace shlex.quote with `subprocess.list2cmd()`. It
- uses `"` to quote path,
- performs quoting only if necessary

@MrMino , could you take a look please?
image

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -12,7 +12,7 @@ import re
 import shlex
 import sys
 from pathlib import Path
-from multiprocessing import list2cmdline
+from subprocess import list2cmdline
 from IPython.core.magic import Magics, magics_class, line_magic
 
 
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -69,7 +69,7 @@ class PackagingMagics(Magics):
         python = sys.executable
         if " " in python:
             if sys.platform == "win32":
-                python = "\"" + python + "\""
+                python = '"' + python + '"'
             else:
                 python = shlex.quote(python)
 
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -12,7 +12,7 @@ import re
 import shlex
 import sys
 from pathlib import Path
-from subprocess import list2cmdline
+from multiprocessing import list2cmdline
 from IPython.core.magic import Magics, magics_class, line_magic
 
 
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -66,11 +66,12 @@ class PackagingMagics(Magics):
         Usage:
           %pip install [pkgs]
         """
+        python = sys.executable
         if " " in sys.executable:
             if sys.platform == "win32":
-                python = "\"" + sys.executable + "\""
+                python = "\"" + python + "\""
             else:
-                python = shlex.quote(sys.executable)
+                python = shlex.quote(python)
 
         self.shell.system(" ".join([python, "-m", "pip", line]))
 
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -67,11 +67,10 @@ class PackagingMagics(Magics):
           %pip install [pkgs]
         """
         python = sys.executable
-        if " " in python:
-            if sys.platform == "win32":
-                python = '"' + python + '"'
-            else:
-                python = shlex.quote(python)
+        if sys.platform == "win32":
+            python = '"' + python + '"'
+        else:
+            python = shlex.quote(python)
 
         self.shell.system(" ".join([python, "-m", "pip", line]))
 
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -12,7 +12,7 @@ import re
 import shlex
 import sys
 from pathlib import Path
-from multiprocessing import list2cmdline
+
 from IPython.core.magic import Magics, magics_class, line_magic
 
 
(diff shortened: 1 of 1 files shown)


MrMino (changes requested):
Ugh... I didn't wanted `list2cmdline` cause it's MSC specific [but it turns out that `shlex.quote` also isn't "platform agnostic"](https://github.com/python/cpython/commit/f9a8386e44a695551a1e54e709969e90e9b96bc4). It didn't occur to me that single quotes would have different meaning to double quotes - I should've tested it on Windows.

One other argument against using `list2cmdline`: [it isn't really "public"](https://bugs.python.org/issue11827) in the same sense other methods of `subprocess` are. It's an implementation detail, not a "public API".

Furthermore, I feel like introducing MSC shell semantics on all other platforms will introduce subtle bugs and edge cases. A simple `.replace("'", "\"")` also isn't enough, since virtualenv paths can contain `"` on UNIX-es.

At this point I'm questioning whether this should even use `shell.system`. I should probably look into creating a `argv`-style API for shell commands in magics. But let's fix this first. Any changes to the `shell.system` API would probably need to go into 8.0, and we need to get the fix into 7.27.

I don't see any other option than using something like `if sys.platform == "win32": "\"" + s + "\""; else: shlex.quote(...)`.

madbird1304:
@MrMino , you said:
> At this point I'm questioning whether this should even use shell.system.

Then let's just get rid of it?
image

MrMino:
We cannot get rid of or substantially change `shell.system` until at least 8.0, and even then I'd rather not add more to an already bulky major release. The fix has to go into the next 7.x.

Coming up with something that is a valid replacement for it isn't trivial either.

Unless there's another option, I'd rather bodge it in with a `sys.platform` branch.

IPython/core/magics/packaging.py
```diff
@@ -66,7 +66,13 @@ def pip(self, line):
         Usage:
           %pip install [pkgs]
         """
-        python = shlex.quote(sys.executable)
+        python = sys.executable
+        if " " in python:
+            if sys.platform == "win32":
+                python = "\"" + python + "\""
+            else:
+                python = shlex.quote(python)
```
MrMino, line 74: Branching over `" " in python` isn't necessary here.

```suggestion
        if sys.platform == "win32":
            python = "\"" + python + "\""
        else:
            python = shlex.quote(python)
```

MrMino (commented):
Exactly what I had in mind :). Thanks!

If you ~add a whats-new entry and~ fix the linter errors, I'll merge.

Edit: @madbird1304 since I'm on vacation next week, and I don't want to postpone it till the next release, I went ahead and fixed the lint for you. Hope you don't mind.

IPython/core/magics/packaging.py
```diff
@@ -66,7 +66,13 @@ def pip(self, line):
         Usage:
           %pip install [pkgs]
         """
-        python = shlex.quote(sys.executable)
+        python = sys.executable
+        if " " in python:
+            if sys.platform == "win32":
+                python = '"' + python + '"'
+            else:
+                python = shlex.quote(python)
```
MrMino, line 74: Branching over `" " in python` isn't necessary here.

```suggestion
        if sys.platform == "win32":
            python = "\"" + python + "\""
        else:
            python = shlex.quote(python)
```

MrMino, line 74: Branching over `" " in python` isn't necessary here.

```suggestion
        if sys.platform == "win32":
            python = '"' + python + '"'
        else:
            python = shlex.quote(python)
```

MrMino (approved):

MrMino:
Thanks @madbird1304!

madbird1304:
Sorry for late reply. I was busy last week, so thanks for finishing that without me.

PR #13094, opened 2021-08-17
Fix virtual environment user warning for lower case pathes

marcus-wirtz-snkeos:
In my environment the pathes "p" and "p_venv" from lines 913 and 914 of `IPython/core/interactiveshell.py` are:
> p: "c:/users/marcus.wirtz/venvs/dlapptrainer/scripts/python.exe"
> p_venv: "C:/Users/marcus.wirtz/VEnvs/dlapptrainer"

thus, even though ipython is installed within the virtual env and all libraries are available, ipython will give me the UserWarning "Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv." Here a screenshot:

image

I don't exactly understand why the pathes are in one case lower-case and in the other one mixed case - Windows path's are always messy... However, I guess given the path chaos in Windows it might be good to only compare the ".lower()" versions of the pathes.

Best,
Marcus

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -928,7 +928,12 @@ class InteractiveShell(SingletonConfigurable):
         elif len(str(p_venv)) >= 2 and str(p_venv)[1] == ":":
             p_venv = Path(str(p_venv)[2:])
 
-        if any(os.fspath(p_venv).lower() in os.fspath(p).lower() for p in paths):
+        if sys.platform == "win32":
+            # In Windows there might be a mixture of lower-case and mixed-case pathes
+            p_venv = Path(str(p_venv).lower())
+            paths = [Path(str(p).lower()) for p in paths]
+
+        if any(os.fspath(p_venv) in os.fspath(p) for p in paths):
             # Our exe is inside or has access to the virtualenv, don't need to do anything.
             return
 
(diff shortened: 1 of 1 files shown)


MRCWirtz:
Now that I think about again, maybe we want to do this "lower-case check" only in the `if sys.platform == "win32":` block a few lines below. Just o be sure to not introduce a misbehavior for UNIX-like systems... :-)

IPython/core/interactiveshell.py
```diff
@@ -928,6 +928,11 @@ def init_virtualenv(self):
         elif len(str(p_venv)) >= 2 and str(p_venv)[1] == ":":
             p_venv = Path(str(p_venv)[2:])
 
+        if sys.platform == "win32":
+            # In Windows there might be a mixture of lower-case and mixed-case pathes
```
MrMino, line 932:
```suggestion
            # In Windows there might be a mixture of lower-case and mixed-case paths
```

MrMino:
@MRCWirtz Yes, this should only be specific to platforms which have case-insensitive paths in their default filesystems (Windows).

Ideally it should be limited to _filesystems_ that have case-insensitive paths (FWIW it's possible to mount FAT32 under Linux, but I don't remember how the paths behave). I don't know how complex is that to check. Probably an overkill.

In any case, `sys.platform=="win32"` seems like a good heuristic.

MrMino:
LGTM, thanks @MRCWirtz

MrMino:
Squish squash...

MrMino:
After #13146 the changes are no longer needed, and so...

%meeseeksdev backport to 7.x

Let's see if it works this time.

PR #13095, opened 2021-08-18
Remove some nose

Carreau:


IPython/core/tests/test_formatters.py
```diff
@@ -347,8 +351,8 @@ def __getattr__(self, key):
     call_hat = CallableMagicHat()
     with capture_output() as captured:
         result = f(call_hat)
-    
-    nt.assert_equal(result, None)
+
+    assert result == None
```
MrMino, line 355: Nit, not sure if valid for tests / the style: flake8 complains with `E711: comparison to None should be 'if cond is None:' ` on such lines.

```suggestion
    assert result is None
```

MrMino (approved):
LGTM

MrMino:
Do we want to refactor all the others before 8.0? I'd like to volunteer for that next week (after I get back from my vacation).

Also, wouldn't it be better to get this into 7.x? Won't git complain about conflicts otherwise, if someone edits these hunks in other PRs for 7.x?

Carreau:
> I'd like to volunteer for that next week (after I get back from my vacation).

I have to catch a plane in a few hours and will be offline until Wednesday. 

I'm trying to get some funding to actually hire someone to do most of the removal of nose, It's a bit more complicated as we have some dependencies with ipykernel.

I'll open an issue next week or maybe at the airport to add a bit more details, and organise.

PR #13097, opened 2021-08-18
Protect against failure to show local data.

Carreau:
See https://github.com/napari/napari/pull/3201 where formatting
traceback was failing and lead to silenced errors.

With this change we  instead get the rest of the traceback but without
the local variables, which is still better.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -634,8 +634,16 @@ class VerboseTB(TBTools):
         lvals = ''
         lvals_list = []
         if self.include_vars:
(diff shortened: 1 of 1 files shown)


alexmojaki:
Hi @Carreau, just happened to come across this now. Did you find a way to reproduce the issue? Something I should fix in stack_data?

Carreau:
I can try to reproduce in an older version of Napari and narrow down or put instruction, Let me open another issue to remember to do this.

Carreau:
I've opened #13141 to remember to find some time to do that.

PR #13107, opened 2021-08-25
Change setupbase.py to use pathlib instead of os.

david8381:
Co-authored-by: Zachary Blackwood <zblackwo@gmail.com>
Co-authored-by: Nathan Lis <Nathan.L.2018@gmail.com>

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -21,45 +21,32 @@ from logging import log
 from setuptools import Command
 from setuptools.command.build_py import build_py
 
-# TODO: Replacement for this?
+# TODO: Replacement for this? 
 from distutils.command.build_scripts import build_scripts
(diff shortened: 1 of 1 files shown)


Carreau (changes requested):
Thanks a lot for your first contributions, 

I'm in general for simplification and update like this, unfortunately in this case it seem to create problems and IPython can't be build as some part of setuptools requires str and not path, so I think part of what you've done has to either be reverted, or wrapped in str(), to convert back into strings before returning from functions.

blackary:
Are there any additional changes that should happen for this to be accepted?

wxman22:
@Carreau Did our latest updates address your requested changes?

Carreau:

Issue #13068, opened 2021-07-27
Documentation on profiles is unclear


MNRK01:
I wish to report a bug or at least an inconsistency in the documentation for how IPython finds a customized startup code in a profile_default directory that is in an alternate directory.

In the [IPython documentation on configuration and profiles](https://ipython.org/ipython-doc/3/config/intro.html#profiles), it is stated that one can have an alternate profile in the **current working directory**.

My question: what should the structure of the directory for this alternate profile be? Is it **./.ipython/profile_default/** or **./profile_default/**? I expected that it would be the former given the standard directory structure for IPython configuration. However, the following code indicates that it is the latter since IPython looks for a directory called profile_default under the current working directory:

https://github.com/ipython/ipython/blob/b3355a943823cfb24c1202bebbeb751e718ff172/IPython/core/profiledir.py#L202-L209

Maybe the documentation could be clearer on the directory structure for custom configurations, but in reality it would be nice to retain **./.ipython/profile_default/** as the directory structure for the alternate configuration since that shows the intent of that directory structure as having something related to IPython vs. some generic profile_default directory that a less familiar user might not know about.

The fix could be simple.

```python
ipy_userconfig_dir = os.path.join(os.getcwd(), u'.ipython') if os.path.split(os.getcwd())[-1] != u'.ipython' else os.getcwd()
```

Add the above code after:

https://github.com/ipython/ipython/blob/b3355a943823cfb24c1202bebbeb751e718ff172/IPython/core/profiledir.py#L202

and then replace the call to [os.getcwd()](https://github.com/ipython/ipython/blob/b3355a943823cfb24c1202bebbeb751e718ff172/IPython/core/profiledir.py#L203) with ipy_userconfig_dir.

```python
paths = [ipy_userconfig_dir, ipython_dir]
```

I believe that this recommendation is the path of least surprise. I'm hoping that this change will be accepted.

Thanks for reading and all the good IPython work!

MrMino:
FWIW you're linking to the outdated docs. Here's the right link: [IPython documentation on configuration and profiles](https://ipython.readthedocs.io/en/stable/config/intro.html#profiles).

> My question: what should the structure of the directory for this alternate profile be?

I doubt that this is relevant to the user. IMO this should be handled by IPython, even if the profile is being created in CWD instead of a designated directory in user's home.

Having said that, the documentation is indeed unclear. I never used this feature, so I had to actually read the docs to get how it works. I've read it... and I'm confused.

[...]

PR #13123, opened 2021-09-07
Updated profile docs

nav1s:
This clears the documentation of profiles in IPython as discussed in #13068.

### DIFF
--- a/docs/source/config/intro.rst
+++ b/docs/source/config/intro.rst
@@ -166,14 +166,17 @@ Profiles
 
 IPython can use multiple profiles, with separate configuration and
 history. By default, if you don't specify a profile, IPython always runs
-in the ``default`` profile. To use a new profile::
+in the ``default`` profile. To use a profile::
 
-    ipython profile create foo   # create the profile foo
-    ipython --profile=foo        # start IPython using the new profile
+    ipython --profile=foo        # start IPython using foo if it exists or create it if it doesn't
 
 Profiles are typically stored in :ref:`ipythondir`, but you can also keep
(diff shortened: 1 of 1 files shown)


MrMino (commented):
Have you found out why the subcommand was here in the first place and how it's different without it? What is its relationship to the behavior of `--profile` :)?

I didn't mean to suggest it should be outright replaced with whatever I found to be working. The issue is about finding out the raison d'être of `ipython profile create`, and making it clear which command does what.

If we'll just substitute this with whatever _we_ feel could be the right way of working with profiles, then we're cutting out something that might actually be valuable, without having any insight into why this was here in the first place.

Flodur871:
Sorry for being brief before, I have dived into both the source code and the debug messages and from what I can tell the only difference between ```--profile``` and ```ipython profile create``` is that because the latter is an IPython command it doesn't start IPython interactively.

This is what the end of the output for ```ipython --debug profile create``` looks like:

```
[TerminalIPythonApp] WARNING | File 'profile' doesn't exist
[TerminalIPythonApp] Exiting application: ipython
```

This is the source for the warning:

https://github.com/ipython/ipython/blob/981302a49e7fdc18dccc1b67666380d87b3a2ad9/IPython/core/shellapp.py#L444-L457

And this is what  ```ipython profile create```  does:

https://github.com/ipython/ipython/blob/981302a49e7fdc18dccc1b67666380d87b3a2ad9/IPython/core/profileapp.py#L258-L286

Additionally from what I can see if one were to create a file called ```profile``` it would execute it when using the command which I'm not sure is the desirable outcome.

PR #13127, opened 2021-09-08
Backport PR #13094 on branch 7.x (Fix virtual environment user warning for lower case pathes)"

MrMino:
See #13094.

Slight changes were required since 7.x is not pathlibed, but it does not have impact on the overall behavior.


MrMino:
No longer needed. After merging #12548 an automatic backport can be made: #13149.

PR #13133, opened 2021-09-13
enable autoplay in embed youtube player

yuji96:
Added a new argument to autoplay YouTubeVideo in Chrome without muting it.
The official Chrome Developers blog on this topic is here:
https://developer.chrome.com/blog/autoplay/

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -263,13 +263,15 @@ class IFrame(object):
             src="{src}{params}"
             frameborder="0"
             allowfullscreen
+            {extra}
         ></iframe>
         """
 
-    def __init__(self, src, width, height, **kwargs):
+    def __init__(self, src, width, height, extra="", **kwargs):
         self.src = src
         self.width = width
         self.height = height
+        self.extra = extra
         self.params = kwargs
 
     def _repr_html_(self):
@@ -282,7 +284,9 @@ class IFrame(object):
         return self.iframe.format(src=self.src,
                                   width=self.width,
                                   height=self.height,
-                                  params=params)
+                                  params=params,
+                                  extra=self.extra)
+
 
 class YouTubeVideo(IFrame):
     """Class for embedding a YouTube Video in an IPython session, based on its video id.
@@ -310,11 +314,13 @@ class YouTubeVideo(IFrame):
     will be inserted in the document.
     """
 
-    def __init__(self, id, width=400, height=300, **kwargs):
+    def __init__(self, id, width=400, height=300, allow_autoplay=False, **kwargs):
         self.id=id
         src = "https://www.youtube.com/embed/{0}".format(id)
+        if allow_autoplay:
+            kwargs.update(autoplay=1, extra='allow="autoplay"')
         super(YouTubeVideo, self).__init__(src, width, height, **kwargs)
-    
+
     def _repr_jpeg_(self):
         # Deferred import
         from urllib.request import urlopen
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -8,7 +8,7 @@ from os import walk, sep, fsdecode
 
 from IPython.core.display import DisplayObject, TextDisplayObject
 
-from typing import Tuple
+from typing import Tuple, Iterable
 
 __all__ = ['Audio', 'IFrame', 'YouTubeVideo', 'VimeoVideo', 'ScribdDocument',
            'FileLink', 'FileLinks', 'Code']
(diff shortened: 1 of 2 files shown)


IPython/lib/display.py
```diff
@@ -263,13 +263,15 @@ class IFrame(object):
             src="{src}{params}"
             frameborder="0"
             allowfullscreen
+            {extra}
         ></iframe>
         """
 
-    def __init__(self, src, width, height, **kwargs):
+    def __init__(self, src, width, height, extra="", **kwargs):
```
MrMino, line 270: Minor thing: it would look better to use an `extras: List = None` here, and append the `.join()`-ed items to the iframe text.

Constraining this to only one extra field seems a bit too limitting in such generic class.

MrMino (changes requested):
LGTM for the most part. Thank you!

The following needs to happen in order for this PR to get merged:
 - We need to make the CI pass
     - Please reformat the code using `darker` (see message in the [logs](https://github.com/ipython/ipython/pull/13133/checks?check_run_id=3593776610)).
     - We'll need to look into why the docs build fails
 - Please make the argument `extras: List` rather than `extra = ""`, as noted in the comment.
 - A what's-new entry: see [the WN readme](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr)

IPython/lib/display.py
yuji96: Thank you for reviewing.

Because attributes are separated by spaces, `str` also allows to set multiple fields. I think different users will have different preferences, as shown below:
```python
# My assumption.
extra = '''
    example="1"
    example2="2"
'''

extra = [
    'example="1"',
    'example="2"'
]

extra = {
    "example1": "1",
    "example2": "2",
}
# There are some attributes without value, so dict seems to be hard to use.
```
Should `extra` support both str and list?

MrMino: To me it's more natural and Pythonic to use an ordered collection (note the "s" at the end of `extras`):

```
extras = ['example="1"', 'example="2"']
```

That way the fact that it can take attributes - plural - isn't hidden behind a singular name.

MrMino:
Looks to me that the issue with parallel docs should be firstly fixed on master. Thank you very much for including it, I didn't have the time to look into it, and you've fixed the situation for at least 3 other PRs :).

I'll move a0f7568 to a separate PR, just to make sure we have a clean history.

IPython/lib/display.py
```diff
@@ -310,11 +319,14 @@ class YouTubeVideo(IFrame):
     will be inserted in the document.
     """
 
-    def __init__(self, id, width=400, height=300, **kwargs):
+    def __init__(self, id, width=400, height=300, allow_autoplay=False, **kwargs):
         self.id=id
         src = "https://www.youtube.com/embed/{0}".format(id)
+        if allow_autoplay:
+            extras = list(kwargs.get("extras", [])) + ['allow="autoplay"']
+            kwargs.update(autoplay=1, extras=extras)
```
MrMino, line 327: Why `autoplay=1`?

yuji96: `autoplay` is a parameter required to get an iframe that can autoplay. ([IFrame Player API](https://developers.google.com/youtube/player_parameters#Parameters))
Is it a bit undesirable that IPython source code includes parameters defined by external services?

As shown below, my purpose is satisfied as long as there is an argument `extras`, so this part can be removed.

```python
YouTubeVideo("id", autoplay=1, extras=['allow="autoplay"'])
```

I think argument `extras`, which allows to add attributes to iframe tags from outside, is in demand.

MrMino: No, I meant to ask: why specifically `kwargs.update(autoplay=1)` here, it looks like an unused kwarg? `IFrame` class, which is the base class for `YouTubeVideo` doesn't use any `autoplay` kwarg, I thought `extras=['allow="autoplay"']` is all that was needed here.

yuji96: `kwargs` will finally be passed to `self.params` in the `IFrame` class and used as a query parameter for API. So, `kwargs` is a useful argument.

https://github.com/ipython/ipython/blob/deeb968055d1a2b1590bdc132ce3e3ea420da929/IPython/lib/display.py#L259-L268

MrMino: Oh, I missed that, sorry. Nevermind then.

MrMino (approved):

MrMino:
@yuji96 could you please prepare a [what's new entry](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr#readme)?

MrMino:
Thanks!

PR #13151, opened 2021-09-22
Make the virtualenv warning informative

MrMino:
This is something that will make more sense after I finish prettifying the warnings. Sending it as a draft so that I don't forget about it.

This is the warning that is probably the most often triggered one, and so it deserves to be a bit more revealing.
I'm wondering if I'll be able to fit an instruction to use `hash -d`/`rehash` somewhere in there, since it's a virtualenv gotcha that many people don't know about, we'll see.


Carreau:

PR #13140, opened 2021-09-17
Use pathlib parent relationships to compare virtualenv directories

MrMino:
Fixes #13124

PR #13153, opened 2021-09-22
Adapt to all sorts of drive names for cygwin

Mithil467:
Earlier implementation (from https://github.com/ipython/ipython/pull/13140) assumed the drive name to be `C` and ignored the possibility of it being different. Drive names could also be greater than 1 in length. This possibility should also be considered.

The old implementation of taking `p_venv[11:]` to get rid of "\cygdrive\c" won't work with drives having names greater than 1.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -924,7 +924,9 @@ class InteractiveShell(SingletonConfigurable):
         
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
         if str(p_venv).startswith("\\cygdrive"):
-            p_venv = "C:" / Path(str(p_venv)[11:])
+            p_venv = Path(str(p_venv))
+            drive_name = p_venv.parts[2]
+            p_venv = (drive_name + ":/") / Path(*p_venv.parts[3:])
 
         if any(p_venv == p.parents[1] for p in paths):
             # Our exe is inside or has access to the virtualenv, don't need to do anything.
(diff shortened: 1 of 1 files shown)


IPython/core/interactiveshell.py
```diff
@@ -924,7 +924,9 @@ def init_virtualenv(self):
         
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
         if str(p_venv).startswith("\\cygdrive"):
-            p_venv = "C:" / Path(str(p_venv)[11:])
+            p_venv = Path(str(p_venv))
```
MrMino, line 927: Do we still need to `Path`→`str`→`Path` here?

IPython/core/interactiveshell.py
```diff
@@ -924,7 +924,9 @@ def init_virtualenv(self):
         
         # In Cygwin paths like "c:\..." and '\cygdrive\c\...' are possible
         if str(p_venv).startswith("\\cygdrive"):
```
MrMino, line 926: I've just noticed this, but not sure if it's correct (haven't checked)

```suggestion
        if p_venv.parts[1] == cygdrive:
```

IPython/core/interactiveshell.py
Mithil467: Good catch! No, we don't need it. I'll remove it.

IPython/core/interactiveshell.py
Mithil467: This is neat! Just tested it works.

MrMino:
Thanks!

PR #13157, opened 2021-09-24
Add history file to debugger

Carreau:
This adds a configurable
`InteractiveShell.debugger_history_file=...` which default to
`~/.pdbhistory`, that store what is typed in ipdb;
this make it easy to persist across sessions.

Some of the logic is moved into the debugger itself so that existence
and creation of file is used only once Pdb is started.


minrk:
Great!

PR #13162, opened 2021-09-29
print_figure return base64 str instead of bytes

minrk:
sending bytes relies on serializer's `default` functionality, where we can just send the right thing here. This used to be ambiguous on Python 2 when `str is bytes`, but not anymore.

This should also avoid https://github.com/jupyter/jupyter_client/issues/702


minrk:
I made it backward-compatible by leaving the default value as False in case there are any other callers of `print_figure`, even though we call it with True 100% of the time ourselves.

PR #13168, opened 2021-10-01
Use path in setupbase.py

heksadecimal:
for #12515

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -17,6 +17,7 @@ import re
 import sys
 from glob import glob
 from logging import log
+from pathlib import Path
 
 from setuptools import Command
 from setuptools.command.build_py import build_py
@@ -38,6 +39,7 @@ pjoin = os.path.join
 repo_root = os.path.dirname(os.path.abspath(__file__))
 
 def execfile(fname, globs, locs=None):
+    fname = Path(fname)
     locs = locs or globs
     with open(fname) as f:
         exec(compile(f.read(), fname, "exec"), globs, locs)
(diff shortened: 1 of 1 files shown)


setupbase.py
```diff
@@ -262,7 +264,7 @@ def run(self):
             name, entrypt = script.split('=')
             name = name.strip()
             entrypt = entrypt.strip()
-            outfile = os.path.join(self.build_dir, name)
+            outfile = Path(os.path.join(self.build_dir, name))
```
Carreau, line 267: One of the goal of using Pathlib is to not have to use things like `os.path.join`, so a better way would be:
```suggestion
            outfile = Path(self.build_dir) / name
```

And even better woudl be to ensure that self.build_dir is already a path object, but that might be hard here.

setupbase.py
```diff
@@ -388,6 +390,7 @@ def _record_commit(self, base_dir):
             repo_commit = repo_commit.strip().decode("ascii")
 
             out_pth = pjoin(base_dir, pkg_dir, 'utils', '_sysinfo.py')
+            out_pth = Path(out_pth)
             if os.path.isfile(out_pth) and not repo_commit:
```
Carreau, line 394: Same here, we want to avoid pjoin, 


```suggestion
            out_pth = Path(base_dir)/  pkg_dir/ 'utils'/ '_sysinfo.py'
            if os.path.isfile(out_pth) and not repo_commit:
```

And then you can likely go further, is `out_path` is a `Path`, you can the use `out_path.is_file()`

Carreau (changes requested):
Thanks for your contribution ! This is a great start, though I believe you can go further.

One of the reason to use Pathlib is to simplify a lot of the code handling later, and avoid usage of methods like os.path.join and others.

I gave you some examples of how you could make use of Pathlib. There are others, you can try to find them if you wish to.

setupbase.py
heksadecimal: Thank you for the kind words. This was my first contribution and the experience so far is good :)
I didn't know we could use a `/` for joining paths. Guess we learn something new everyday. I will commit the requested changes right away.

Carreau:

Issue #13165, opened 2021-09-30
Invalid pathlib.Path prepended to sys.path inside virtualenv using outside IPython


reillysiemens:
## Software Versions Tested
<table>
  <tr>
    <td>🐍 <b>Python</b></td>
    <td>3.9.7, 3.10.0rc2</td>
  </tr>
  <tr>
    <td>✔️ <b>Passing IPython</b></td>
    <td>7.27.0</td>
  </tr>
  <tr>
    <td>❌ <b>Failing IPython</b></td>
    <td>7.28.0</td>
  </tr>
</table>

## Issue Description

With fresh installations of the latest Python and IPython I cannot use an IPython installed outside of a virtualenv to import modules inside that virtualenv.

I assume from the `UserWarning`
> Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.

that this workflow is supported on a best-effort basis, if at all. I understand if this isn't a high priority issue. That said, it is handy to not install IPython into each virtualenv I use. Historically it has worked quite well in this regard.

As far as I can determine, the cause of the issue is an unexpected `pathlib.Path` instance prepended to `sys.path`.

```python
In [1]: import sys

In [2]: sys.path[0]
Out[2]: PosixPath('/tmp/venv/lib/python3.9/site-packages')
```

Near as I can tell from [the documentation](https://docs.python.org/3/library/sys.html#sys.path) and cursory searching, `sys.path` should only contain `str` instances.

I don't know the IPython codebase well enough to say for sure, but from searching through issues and PRs it seems like this might have been an unanticipated result of #12548 and its 7.x backport #13146.

The issue does **not** seem to be reproducible in [IPython 7.27.0](https://github.com/ipython/ipython/releases/tag/7.27.0), so that would seem to support this claim as the aforementioned backport merged a few days ago prior to the release of [IPython 7.28.0](https://github.com/ipython/ipython/releases/tag/7.28.0).

If that is the case, maybe it would suffice to replace
https://github.com/ipython/ipython/blob/c31fed6f3227cccad94ab318e42058a4a677680f/IPython/core/interactiveshell.py#L946

with something like
```python
sys.path.insert(0, os.fspath(virtual_env))
```

## Steps to Reproduce
1. Install IPython outside a virtualenv.
   ```shell
   pip3.9 install ipython
   ```
2. Create a virtualenv.
   ```shell
   python3.9 -m venv venv
   ```
3. Source the virtualenv.
    ```shell
[...]

PR #13170, opened 2021-10-02
Convert WindowsPath to str in init_virtualenv.

deep-jkl:
Hi, I was investigating problems described in #13165 and I have found similar behavior in current master on windows. I have also prepared backport for 7.x, but it might be a greater issue.


Carreau:
Apologies for missing this, this looks straitforward.
And lets see if our bot can autobackport.

PR #13171, opened 2021-10-02
Manually backport whole init_virtualenv

deep-jkl:
I was wondering about #13165 and I believe that best way to solve it is to copy whole implementation from master, I have also included conversion of `Path` to `str` for "win32", which is proposed in #13170.


Carreau:
Thanks for backporting !

PR #13173, opened 2021-10-03
Use pathlib in apigen.py

distgr:
For #12515

### DIFF
--- a/docs/sphinxext/apigen.py
+++ b/docs/sphinxext/apigen.py
@@ -21,6 +21,7 @@ project."""
 # Stdlib imports
 import ast
 import inspect
+import pathlib
 import os
 import re
 from importlib import import_module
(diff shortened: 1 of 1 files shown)


Carreau:
Apologies for the delay in reviewing, the last few/next few weeks have been busy. Thanks for your patience.

docs/sphinxext/apigen.py
```diff
@@ -175,34 +176,34 @@ def _uri2path(self, uri):
...
-            path = os.path.join(path, '__init__.py')
+        elif pathlib.Path(pathlib.PurePath(path, '__init__.py')).exists():
+            path = pathlib.PurePath(path, '__init__.py')
         else:
             return None
         return path
 
     def _path2uri(self, dirpath):
         ''' Convert directory path to uri '''
         relpath = dirpath.replace(self.root_path, self.package_name)
```
GalBr, line 203: The build fails here because you pass a PurePath object to this function, but PurePaths don't have I/O operations (replace() is one of those). You want to use Path instead when accessing real files.

PR #13174, opened 2021-10-03
Fix cpaste documentation

### LINKED ISSUE #12930: %cpaste doc error: Magics ARE supported
`%cpaste?` claims "IPython statements (magics, shell escapes) are not supported (yet)."  But at least magics apparently are!

sgaist:
The current documentation was not accurate anymore.

This merge request does the following:

- Rewrote the statement about IPython magics not being supported
- Fix the example that was using the old Python 2 print style
- Add an example using a magic

Closes #12930


Carreau:
Thanks ! That's great !

PR #13181, opened 2021-10-08
create ipython_dir if not exists

haobibo:
Create the ipython_dir (e.g.: `~/.ipython`) if it does not exists.
Encountered this issue when using ipyparallel, which scans  `ipython_dir` directly: https://github.com/ipython/ipyparallel/blob/main/ipyparallel/util.py#L666


Carreau:
Thanks, let me think about whether we want to actually do that, or check in ipyparallel whether the folder exists.

Issue #11592, opened 2019-02-05
IPython installs a "tests" package inadvertently


arel:
Installing IPython appears to also install a top-level package `tests` from

```
ipython/IPython/extensions/tests
```

I believe this is unintentional and it has the effect of clobbering any local packages named `tests`. For example, say I have the following directory structure in a fresh Python 3.6 environment with only IPython installed:

```
tests/fixtures.py
```

If I import this module from Python, it works fine:

```
Python 3.6.3 |Anaconda, Inc.| (default, Oct 13 2017, 12:02:49) 
[GCC 7.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tests.fixtures  # works
```

If I try to import the same package from IPython, it fails.

```
Python 3.6.3 |Anaconda, Inc.| (default, Oct 13 2017, 12:02:49) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import tests.fixtures                                                                                                                   
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-e3a550607dd1> in <module>
----> 1 import tests.fixtures

ModuleNotFoundError: No module named 'tests.fixtures'
```

I also noticed that this only affects packages using Python's [Implicit Namespace Packages](https://www.python.org/dev/peps/pep-0420/): if an empty `__init__.py` file is added to the `tests/` directory, then IPython finds the `fixtures` module.

I believe the solution is to exclude `tests` from [`find_packages`](https://github.com/ipython/ipython/blob/d38770bbe722785f00860430228d9476e096a87f/setupbase.py#L90).

---

```
{'commit_hash': '523ed2fe5',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/home/ubuntu/.local/share/virtualenvs/ip-7UZ8tV0G/lib/python3.6/site-packages/IPython',
 'ipython_version': '7.2.0',
 'os_name': 'posix',
 'platform': 'Linux-4.4.0-1074-aws-x86_64-with-debian-stretch-sid',
 'sys_executable': '/home/ubuntu/.local/share/virtualenvs/ip-7UZ8tV0G/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.6.3 |Anaconda, Inc.| (default, Oct 13 2017, 12:02:49) \n'
[...]

PR #13188, opened 2021-10-15
Fix tests turning into module 

jbedorf:
The existence of the `__init__.py` file in the extensions/tests folder results in that the folder is added to the Python path when the extensions are loaded which breaks the tests of other projects.

Also see: #12892   and #11592


Carreau:
Yeah, it seem to be in IPython/__init__.py line 47, the code is from 2006, I think it can be removed and many of the built in extension removed as they are deprected.

PR #13198, opened 2021-10-18
match log message for profile by name and path

minrk:
we had debug-level for name, info-level for full path.

Make both debug-level for consistency


Carreau:
THanks !

PR #13204, opened 2021-10-20
Do not use mutables as default arguments

DimitriPapadopoulos:
Articles that discuss why mutable default arguments should be avoided:
* [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)
* [Using a mutable default value as an argument](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html)
* [Python Pitfall: Mutable Default Arguments](https://towardsdatascience.com/python-pitfall-mutable-default-arguments-9385e8265422)
* ...

Noticed in https://github.com/matplotlib/matplotlib/pull/21402 while fixing the matplotlib file `docs/sphinxext/github.py` which had been copied over from IPython (https://github.com/matplotlib/matplotlib/commit/820985ad65be943568512dc20ae3eca4a3a0f273).

### DIFF
--- a/docs/sphinxext/github.py
+++ b/docs/sphinxext/github.py
@@ -51,7 +51,7 @@ def make_link_node(rawtext, app, type, slug, options):
                            **options)
     return node
 
-def ghissue_role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+def ghissue_role(name, rawtext, text, lineno, inliner, options=None, content=None):
     """Link to a GitHub issue.
 
     Returns 2 part tuple containing list of nodes to insert into the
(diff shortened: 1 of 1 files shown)


MrMino:
@DimitriPapadopoulos I analyzed the flow of these mutables through the call stack - unless `options` dict contains mutables itself (I haven't checked what happens in that case), it isn't mutated in any way that would break by changing the signature this way.

And, honestly, if it did, I'd say it's a bug in docutils, or at least a very poor choice in API design.

Issue #13108, opened 2021-08-25
Numfocus Small Developper Grant to tackle removing nose and IPython 8.0


Carreau:
I open this issue as a central point to coordinate, track progress and for transparency.

I'm going to submit a numfocus Small Developpement Grant hoping to subcontract someone to tackle the most critical issues to get IPython 8.0 through the door.  The critical points are
  - [x] Remove any dependency and mention of nose. Indeed this is impeding packaging on linux, and on Python 3.10.
  - [ ] Making sure 8.0 (or at least the 8.x branch) is a future proof with respect to changes in Python 3.10/3.11. 
  - [ ] This would mean that 8.x would become the new stable branch and that 7.x would not get any more releases except critical ones. 
  - [ ] We might break a couple of APIs as this would be a major version.
  - [ ] This might need some changes on IPykernel as well as it imports some of the testing facilities of IPython that rely on nose.

Below is what was submitted (and accepted), it's done via a form, for simplicity so this attempt to reproduce the questions of the forms.

<details>
SUBMIT BY SEPT 3rd (Friday of next week).

**Name of Submitter:**
**Email:**
**Proposal Title:** IPython maintenance and future proofing
Two Sentence Summary: Work is required to ensure the compatibility of IPython and IPykernel with the upcoming Python 3.10 and 3.11 releases and migrate the test suite from nose to pytest.  In addition, this work, along with two years of accumulated bug fixes and enhancements, will be release as IPython 8.0 and next version of IPykernel

**Description of Proposal:This question is required. No more than 750 words (4,500 characters max).**

We propose three main tasks:
- Ensure that IPython is compatible and warning free with Python 3.9, 3.10, and 3.11
- Migrate the test suite from nose to pytest
- Complete the release process for IPython 8.0

Critical maintenance work is needed to ensure IPython and IPykernel are fully compatible and warning free with Python 3.9 and the upcoming Python 3.10 and 3.11 releases. Python 3.9 and future versions of Python have deprecated or moved several parts of the standard library that IPython and IPykernel rely on and work is needed to mitigate these changes.  This work will need to be done on the master branch that will become the new stable (8.x series), and relieve maintainer from tedious backporting of critical bugfixes by making the 8.x branch.  In addition, we will work toward dropping support for older versions of Python or other dependencies.

[...]

PR #13206, opened 2021-10-20
Nose cleanup first step

sgaist:
This merge request goes through the test files and makes a first pass cleanup to remove the direct use of nose in the test files themselves.

Each file has been cleaned up independently so that it can be reviewed in an easy fashion.

Part of #13108

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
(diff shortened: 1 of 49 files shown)


Carreau (changes requested):
A couple of comments, I think you forgot one assert. 

I'm also trying to see if  it's possible to rebase and and applying the reformat hook to each commit, so that it does not complain in the test suite. If that works i'll point you to the rebased branch and the reformat commit, and we can decide if we ant to squash them or not.

Feel free to also do many smaller PRs, that might allow me to quickly review them and merge when I'm in between two compilations or waiting on another test suite.

, to see if I can get the test to pass, and

Carreau:
> is it intentionally?

I don't quite remember, I believe that when we originally added codecov the bot was poting a new comment on each push/each time a new CI item would finish. I don't believe it's the case anymore so we can re-enable. I'll see if I can do 
that.

Sidenote, there is a conflict now.

Carreau (approved):

PR #13210, opened 2021-10-23
Pathlib in IPython/paths.py

deep-jkl:
I have removed usage of os.path in IPython/paths.py.

I have updated tests accordingly (primarily to not use os.path).

However, there is a change, former code utilized os.path.realpath, which is not equivalent to pathlib.Path.resolve(). The former converts path to absolute. I have addressed this matter in tests, but I am not sure that it won't affect the library in general.

### DIFF
--- a/IPython/core/tests/test_paths.py
+++ b/IPython/core/tests/test_paths.py
@@ -84,7 +84,7 @@ def test_get_ipython_dir_3():
         ), warnings.catch_warnings(record=True) as w:
             ipdir = paths.get_ipython_dir()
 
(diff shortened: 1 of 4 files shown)


Carreau:
Apologies for delay in reviewing, the last few weeks/next few ones have been busy. I'll do my best to review when possible.

Carreau:

deep-jkl:
Hi,

this is understandable, thank you for maintaining this project!
I hope I can bring up something more helpful and relevant.

Best wishes,
Jakub

Issue #7998, opened 2015-03-08
Traceback without line number


RyanCarey:
Hi,

I notice that IPython tends to give me a traceback that is less helpful than the Python REPL. Specifically, it's missing the line numbers.

MWE:

#test.py
import numpy as np
def test():
  a = np.zeros(3)
  a[:] = [1,2,3,4]        # error here

#Calling test.py from REPL
>>> exec(open('test.py').read())
>>> test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 4, in test
ValueError: cannot copy sequence with size 4 to array axis with dimension 3

#Calling test.py from IPython notebook
In [10]:
exec(open('test.py').read())
test()

ValueError                                Traceback (most recent call last)
<ipython-input-10-c7b4c34a48bc> in <module>()
      1 exec(open('test.py').read())
----> 2 test()
<string> in test()
ValueError: cannot copy sequence with size 4 to array axis with dimension 3

I tend to be missing the line numbers for basically all scripts (except for within IPython notebook itself). For bigger files, it's essential to have the line number, so I often have to alt-tab from IPython to the ordinary REPL and recreate all my IPython variables, then call the script again there to get the full error message. I'd much prefer if I could do >99% of my interactive computing using just IPython rather than having to retype code into the REPL. Any thoughts on how to patch this, or is it something more fundamental about the way IPython works?

takluyver:
Short answer: just use `%run test.py` - it's less typing and you get a better traceback.

Longer answer: the line numbers are displayed next to the lines in the traceback. But when it can't find the file to get those lines, they're not visible, so the line numbers aren't either. We could probably make it display the line number in that case, but if you just use %run, it should be able to find the file. If you prefer a regular Python function to a magic, you can use `IPython.utils.py3compat.execfile('test.py')`.

RyanCarey:
%run - excellent, it works as advertised.

Thanks takluyver.

takluyver:
No worries. Leaving this open since it does make sense to show line numbers in a case like that, but it's low priority.

PR #13221, opened 2021-10-26
Fixed and added Line Number into error messages

L0uisJ0shua:
I fixed it as requested. Attached is how it looks like now. Not sure if this is what you have in mind @Carreau ! Do let me know if you need me to work on it a bit more! I think I roughly get how the code there works.

Here is `test.py` which is from #7998
Screenshot 2021-10-27 at 00 24 11

Here is the output on my terminal:
Screenshot 2021-10-27 at 00 24 37


L0uisJ0shua:
#13237 might have a better fix to this!

Carreau:
Apologies for the delay in reviewing the last weeks, and next few ones are going to be busy. I'll do my best to review soon.

PR #13238, opened 2021-10-30
Simplify and modernize setup.py and packaging

J-M0:
Hi,

I noticed that IPython still uses a plain `setup.py` and several distutils imports to make its Python packages. Since distutils is deprecated and setuptools is moving away from directly invoking `setup.py`, I decided to try and take a stab at simplifying the `setup.py` script by removing old code and bringing it up to modern standards.

I've been working on adding support for [PEP 517](https://www.python.org/dev/peps/pep-0517/), [PEP 518](https://www.python.org/dev/peps/pep-0518/), and setuptools' [declarative config](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html).

I wanted to get some feedback before going further. Is this something the project would be interested in? Is the direction I'm going in one that you like?


Carreau:
Thanks a lot, I haven't looked at the code, this month is quite busy but this is definitely something I was hopping to do ! 
I'll look into it when I have a bit more time on my hands.

setup.py
```diff
@@ -190,83 +161,14 @@
...
-setup_args.update(setuptools_extra_args)
-
-
-
-def main():
-    setup(**setup_args)
-
-if __name__ == '__main__':
-    main()
+setup(**setup_args)
```
Kojoley, line 174: This is a footgun for pytest, is removing the guard necessary?

setup.cfg
```diff
@@ -1,5 +1,107 @@
...
+IPython.core.tests = *.png, *.jpg, daft_extension/*.py
+IPython.lib.tests = *.wav
+IPython.testing.plugin = *.txt
+
+[options.entry_points]
+console_scripts =
+    ipython = IPython:start_ipython
+    iptest = IPython.testing.iptestcontroller:main
+    ipython3 = IPython:start_ipython
+    iptest3 = IPython.testing.iptestcontroller:main
```
Kojoley, line 100: iptest was removed on trunk, could you please rebase?

J-M0: Will do.

setup.py
J-M0: I suppose it isn't necessary and I wasn't aware of any interaction with pytest. Why would pytest be testing `setup.py` though?

Kojoley: doctest plugin loads every python script to collect and test examples in documentation

setup.cfg
```diff
@@ -1,5 +1,107 @@
...
+[options]
+packages = find:
+python_requires = >=3.7
+zip_safe = False
+install_requires =
+    setuptools>=18.5
+    jedi>=0.16
+    decorator
+    pickleshare
+    traitlets>=4.2
```
Carreau, line 73: Not to self, bump 5+ after merge.

setup.cfg
```diff
@@ -1,5 +1,107 @@
...
+    stack_data
+    matplotlib-inline
+    pexpect>4.3; sys_platform != "win32"
+    appnope; sys_platform == "darwin"
+    colorama; sys_platform == "win32"
+
+[options.packages.find]
+exclude =
+    deathrow
+    quarantine
```
Carreau, line 86: Note to self those two can be removed after merge.

Carreau:
Rebased and applied some of the requested changes. I"ll try to get that in soon so we can try it ! Thanks !

J-M0:
You're welcome! Sorry I didn't get to making the requested changes, I suddenly got very busy right after they were asked for.

PR #13239, opened 2021-10-30
Pytest ipdoctest plugin

Kojoley:
The plugin is based on Pytest doctest.py.

I am very confident that no tests are missing in Pytest discovery (except one bogus and two phantom duplicates, the iptest vs pytest difference can be found here https://gist.github.com/Kojoley/00ccf0aa22097a06cf2bb8d0c7ff78e8/revisions).

The doctest `doctest_who` was changed to do a soft reset, because a hard reset somehow causes failures in other tests that do `run_cell(x, store_history=True); assert _ip.user_ns['_'] == x`.

### DIFF
--- a/IPython/testing/plugin/pytest_ipdoctest.py
+++ b/IPython/testing/plugin/pytest_ipdoctest.py
@@ -1,0 +1,724 @@
+"""Discover and run doctests in modules and test files."""
+import bdb
+import inspect
+import platform
+import sys
+import traceback
+import types
+import warnings
+from contextlib import contextmanager
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Generator
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Pattern
+from typing import Sequence
+from typing import Tuple
+from typing import Type
+from typing import TYPE_CHECKING
+from typing import Union
+
+import py.path
+
+import pytest
+from _pytest import outcomes
+from _pytest._code.code import ExceptionInfo
+from _pytest._code.code import ReprFileLocation
+from _pytest._code.code import TerminalRepr
+from _pytest._io import TerminalWriter
+from _pytest.compat import safe_getattr
+from _pytest.config import Config
+from _pytest.config.argparsing import Parser
+from _pytest.fixtures import FixtureRequest
+from _pytest.nodes import Collector
+from _pytest.outcomes import OutcomeException
+from _pytest.pathlib import import_path
+from _pytest.python_api import approx
+from _pytest.warning_types import PytestWarning
+
+if TYPE_CHECKING:
+    import doctest
+
+DOCTEST_REPORT_CHOICE_NONE = "none"
+DOCTEST_REPORT_CHOICE_CDIFF = "cdiff"
+DOCTEST_REPORT_CHOICE_NDIFF = "ndiff"
+DOCTEST_REPORT_CHOICE_UDIFF = "udiff"
+DOCTEST_REPORT_CHOICE_ONLY_FIRST_FAILURE = "only_first_failure"
+
+DOCTEST_REPORT_CHOICES = (
+    DOCTEST_REPORT_CHOICE_NONE,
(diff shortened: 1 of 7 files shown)


IPython/testing/plugin/pytest_ipdoctest.py
```diff
@@ -0,0 +1,789 @@
...
+        for reprlocation, lines in self.reprlocation_lines:
+            for line in lines:
+                tw.line(line)
+            reprlocation.toterminal(tw)
+
+
+class MultipleDoctestFailures(Exception):
+    def __init__(self, failures: Sequence["doctest.DocTestFailure"]) -> None:
+        super().__init__()
+        self.failures = failures
```
Carreau, line 175: note to self, reviewed until roughly here.

Kojoley: I can split the commit into two commits: copying a Pytest's doctest module and the modifications to make it ipdoctest-capable. That should make it easier to review and to sync with Pytest in the future if needed.

Kojoley: I went ahead and done that. I could also split off doctest->ipdoctest renaming and darker pleasing changes as separate commits if it will be helpful.

Kojoley:
I have identified the reason of strange failures when I don't disable those three doctests, it is because `exec` call adds its result into `buildins` module under `_` key. Why on earth it does that? The documentation keeps silence too. The iptest is unaffected by that because Nose doctest plugin wipes out `_` from `builtins` (`delattr(builtin_mod, self._result_var)`) while Pytest doctest does not.

```python
import builtins
assert '_' not in dir(builtins), 'before'
exec(compile('4\n', 'asd.py', "single", 0, True), {'__builtins__': {}})
assert '_' not in dir(builtins), 'after'
```
```
4
  File "...", line 4, in <module>
    assert '_' not in dir(builtins), 'after'
AssertionError: after
```

Kojoley:
I think the last test failure is uncovered longstanding bug, hard reset in doctests seems to screw things up in some way, and it was not observed in `iptest` because other regular tests happens to make hard reset too in the needed moment.

`iptest` runs tests in this order:
```
IPython/core/tests/test_logger.py::test_logstart_unicode ... ok
IPython/core/tests/test_magic.py::test_reset_out ... ok
IPython/core/tests/test_magic.py::IPython.core.tests.test_magic.doctest_who ... ok
```
If I force pytest to run them in that order -- everything is ok.
```
pytest -vv IPython/core/tests/test_logger.py::test_logstart_unicode IPython/core/tests/test_magic.py::test_reset_out IPython/core/tests/test_magic.py::IPython.core.tests.test_magic.doctest_who
```

I have workarounded the bug by making a soft reset instead in the doctest.

IPython/testing/plugin/pytest_ipdoctest.py
Carreau: Thanks, it's ok, and we can put new files in darker ignore file if that helps.

Carreau:
I'm not surprised there are test ordering issues, feel free to make them as "skip" and "to fix later".

> it is because `exec` call adds its result into `builtins` module under `_` key

There is some behavior that try to keep the last result in `_` for when users forgot to assign a expensive result. It should be done via a shadow namespace now, and I guess the assigning to builtins has not been removed.

Kojoley:
> I'm not surprised there are test ordering issues, feel free to make them as "skip" and "to fix later".
> 
> > it is because `exec` call adds its result into `builtins` module under `_` key
> 
> There is some behavior that try to keep the last result in `_` for when users forgot to assign a expensive result. It should be done via a shadow namespace now, and I guess the assigning to builtins has not been removed.

Nah, it is just CPython `doctest` temporary restores the original display hook `sys.displayhook = sys.__displayhook__` and then calls `exec(compile(..., "single", ...), ...)` which calls `sys.displayhook` with the result (`sys.__displayhook__(result)` is essentially `builtin._ = result; print(result, file=sys.stdout)`).

PR #13240, opened 2021-10-31
test_completer tests resurrection

Kojoley:
* Class in class tests never was a thing.
* Generating methods seems to not work even in nose, they report no coverage.
* Installed pandas in CI to run test_dataframe_key_completion


Kojoley:
FYI there is unconditionally disabled test_no_ascii_back_completion, but I have no idea what it should be.

Carreau:
> FYI there is unconditionally disabled test_no_ascii_back_completion, but I have no idea what it should be.

It's ok, I don't think it's the only one. I'll try to review the history to find why it was there.

PR #13242, opened 2021-11-01
Remove unused _find_cmd

Kojoley:
It was replaced with shutil/py3compat.which in f42c9edb152ac392caa39fe9e50373ef0ec42a55


Carreau:
+1

PR #13243, opened 2021-11-02
Revive doctest_tb_sysexit

Kojoley:
Output changed with `pure-eval>=0.2.0` which learned to parse binary operators and to show its arguments. Test changed to not use the binary operator '%' to pass on any version.

### DIFF
--- a/IPython/core/tests/test_iplib.py
+++ b/IPython/core/tests/test_iplib.py
@@ -43,15 +43,15 @@ def doctest_tb_plain():
 In [18]: xmode plain
 Exception reporting mode: Plain
 
-In [19]: run simpleerr.py
-Traceback (most recent call last):
-  ...line 32, in <module>
-    bar(mode)
-  ...line 16, in bar
-    div0()
-  ...line 8, in div0
-    x/y
-ZeroDivisionError: ...
+        In [19]: run simpleerr.py
+        Traceback (most recent call last):
+          ...line ..., in <module>
+            bar(mode)
+          ...line ..., in bar
+            div0()
+          ...line ..., in div0
+            x/y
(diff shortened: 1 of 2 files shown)


Kojoley:
I could not please darker here without touching a lot of unrelated tests.

Carreau:
> I could not please darker here without touching a lot of unrelated tests.

That is weird, I'll try. It may depend on the version of black, and if black changes its formatting.

Kojoley:
The thing is that adding a newline, as darker wants, changes line numbers which is imprinted in other tests.

Carreau:
Ah, I see what you mean. 

I've pushed a fix the the line number, I think it's ok to do that, I don't believe we are actually need to test that the line numbers are the right one.

And just after fixing them I realized that we could actually replace all lines numbers with `...`  that will match anything.

IPython/core/tests/test_iplib.py
```diff
@@ -118,91 +117,100 @@ def doctest_tb_verbose():
...
+    <BLANKLINE>
+    ...sysexit(stat, mode)
+         10 def sysexit(stat, mode):
+    ---> 11     raise SystemExit(stat, f"Mode = {mode}")
+    <BLANKLINE>
+    SystemExit: (2, 'Mode = exit')
+    """
+
+
+def doctest_tb_sysexit_verbose():
```
Kojoley, line 175: Is there anywhere else a test that ensures that you can call %tb multiple times with different verbosity?

Carreau: Yes, the one just before switch between plain and context. I separated this one just to make it a tiny bit easier to find the rights lines when fixing doctests.

PR #13247, opened 2021-11-04
Skip test_pylabtools when no matplotlib installed

Kojoley:
Currently Pytest will bail out the whole suit because of ImportError.


Carreau:
Thanks !

PR #13251, opened 2021-11-07
Split coverage into distinct library and tests stats

Kojoley:
Some projects exclude tests coverage completely because counting it with other code artificially increases total code coverage, but tests coverage monitoring is a good thing because it catches mistakenly not running tests. The split should eliminate situations with 'coverage decreases' status reports when tests code shrinks while the library code and its coverage remains the same.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,12 @@ coverage:
   status:
     patch: off
     project:
-      default:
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks ! I did not knew you could do that.

PR #13252, opened 2021-11-07
Remove iptest and other Nose-dependent code

Kojoley:
This removes/changes bunch of stuff that was not privatized but should not have been used by anything outside of IPython. I could try to find another way if you think it is too brutal or want to preserve Nose-ipdoctest plugin/the whole iptest-stuff.


Carreau:
Thanks, I think I want to get that in, but I want to just make a couple of checks in a few repos, 

mainly https://github.com/ipython/ipykernel and a few in https://github.com/jupyter as they used to be the same repos and were split at some point in the past, so they may still reference some of the testing tools here. 

If anything I'll probably merge that anyway, so that downstream repositories testing the master branch come here and complain, we can still restore the file if needs be.

Kojoley:
I do not see jupyter to be using this stuff https://github.com/search?q=org%3Ajupyter+IPython.testing&type=code, the ipykernel does https://github.com/ipython/ipykernel/search?q=IPython.testing but not the things I am removing here and since it also uses Pytest it should not even notice the changes, though I cannot ensure that because ipykernel downgrades ipython or gives the `ipykernel 6.5.0 requires ipython<8.0,>=7.23.1, but you have ipython 8.0.0.dev0 which is incompatible.`.

Carreau:
Thanks for checking, 

I've sent https://github.com/ipython/ipykernel/pull/796 to unpin IPython and remove some more dependencies of IPykernel on IPython testing decorators. I'll try to review and merge that soon. Thanks.

Kojoley:
Hmm, I see that for some reason `onlyif_unicode_paths` decorator after this changed started to skip tests.

Kojoley:
Turns out that `onlyif` passes a function to `skipif`, fixed that and added type assertion to `skipif`.

Kojoley:
It seems that I've uncovered some bugs in these decorator usages too. Everything should be good now.

Carreau:
Ok, let's get that in, thanks.

PR #13254, opened 2021-11-07
Suppress bunch of self-deprecation warnings

Kojoley:
And exclude these transition-modules from coverage report


Carreau:
Thanks !

PR #13257, opened 2021-11-08
TST: Remove unneeded requirement on pywin32

Kojoley:
The only routine that requires `pywin32` is `win32_clipboard_get` which is not tested.


Kojoley:
Uncovered a bug with this. Let me know if you would prefer to have the fix as a separate PR or reorder the commits.

Carreau:
No need to reorder commits, thanks.

PR #13260, opened 2021-11-09
Fix codecov.yml syntax

Kojoley:
It expects `paths` to be a list (though a string had been working previously) but for some reason Codecov will silently ignore mistakes and won't tell you that the file has incorrect syntax.


Kojoley:
Yep, now it works.

Carreau:
Thanks !

PR #13261, opened 2021-11-09
Update documentation dependencies

astrojuanlu:
Attempt to fix the CSS weirdness described in https://github.com/readthedocs/readthedocs.org/issues/8638.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,13 +1,17 @@
 name: ipython_docs
+channels:
+ - conda-forge
+ - defaults
 dependencies:
-- python=3.8
-- setuptools>=18.5
-- sphinx>=4.2
-- sphinx_rtd_theme>=1.0
-- numpy
-- testpath
-- matplotlib
-- pip:
+  - python=3.8
+  - setuptools>=18.5
+  - sphinx>=4.2
+  - sphinx_rtd_theme>=1.0
+  - numpy
+  - nose
+  - testpath
+  - matplotlib
+  - pip:
     - docrepr
     - prompt_toolkit
     - ipykernel
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -2,8 +2,8 @@ name: ipython_docs
 dependencies:
 - python=3.8
 - setuptools>=18.5
-- sphinx>=4.2.0
-- sphinx_rtd_theme>=1.0.0
+- sphinx>=4.2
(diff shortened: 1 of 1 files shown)


astrojuanlu:
@Carreau Would you like to [enable the pull request builder for IPython](https://docs.readthedocs.io/en/stable/pull-requests.html) so we can see the effect of this before merging?

astrojuanlu:
Also, I see that RTD is using `docs/environment.yml`:

https://github.com/ipython/ipython/blob/1a57d41f6b58b41a7f78468ce1044e8a090a9000/readthedocs.yml#L1-L2

But there's also `docs/requirements.txt`, used by the CI:

https://github.com/ipython/ipython/blob/1a57d41f6b58b41a7f78468ce1044e8a090a9000/.github/workflows/docs.yml#L19-L22

On one hand, we could make RTD use `docs/requirements.txt` and avoid the use of conda (unless there's a good reason for it), and on the other hand, the GH Actions workflow for building the docs would become redundant if the RTD PR Builder is enabled.

Carreau:
> @Carreau Would you like to [enable the pull request builder for IPython](https://docs.readthedocs.io/en/stable/pull-requests.html) so we can see the effect of this before merging?

Done.

Carreau:
> On one hand, we could make RTD use `docs/requirements.txt` and avoid the use of conda (unless there's a good reason for it), and on the other hand, the GH Actions workflow for building the docs would become redundant if the RTD PR Builder is enabled.

Yeah, I need to cleanup a lot of the CI stuff.

astrojuanlu:
I think the PR builder is not running 🤔

Carreau:
I've set up the webhook again but I got a "Webhook activation failed. Make sure you have the necessary permissions.", red message on top.

Carreau:
I had to re connect my account. Even if it was connected, I guess the requested permissions are broader than before, and it was thus not working. Maybe something need to be added in https://docs.readthedocs.io/en/latest/connected-accounts.html

Carreau:
It's being triggered now.

astrojuanlu:
Build is passing and, looking at https://ipython--13261.org.readthedocs.build/en/13261/whatsnew/version7.html#ipython-7-29 vs https://ipython.readthedocs.io/en/latest/whatsnew/version7.html#ipython-7-29, the CSS got fixed 👍🏽

PR #13264, opened 2021-11-09
f_locals might not have get method

impact27:
See https://github.com/spyder-ide/spyder/issues/16780

pytorch uses a `EvalEnv` object as a locals namespace. This obgect doesn't have a `get` method. See https://github.com/pytorch/pytorch/blob/6831d8e379392da1340a28fdb3e7e1382176d1d4/torch/jit/annotations.py#L150

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -415,9 +415,10 @@ class Pdb(OldPdb):
         if self._predicates["tbhide"]:
             if frame in (self.curframe, getattr(self, "initial_frame", None)):
                 return False
-            else:
-                return self._get_frame_locals(frame).get("__tracebackhide__", False)
-
+            frame_locals = self._get_frame_locals(frame)
+            if "__tracebackhide__" not in frame_locals:
+                return False
+            return frame_locals["__tracebackhide__"]
         return False
 
     def hidden_frames(self, stack):


Kojoley:
That seems to be a bug in Pytorch.

According to [PEP-667](https://www.python.org/dev/peps/pep-0667/#python):

> `frame.f_locals` will return a view object on the frame that implements the `collections.abc.Mapping` interface.

And [`collections.abc.Mapping`](https://docs.python.org/3/library/collections.abc.html#:~:text=keys%2C%20items%2C%20values%2C-,get,-%2C%20__eq__%2C%20and%20__ne__) interface lists `get`:

> ABC | Inherits from | Abstract Methods | Mixin Methods
> -- | -- | -- | --
> Mapping | Collection | `__getitem__`, `__iter__`, `__len__` | `__contains__`, `keys`, `items`, `values`, `get`, `__eq__`, and `__ne__`

ccordoba12:
> According to PEP-667

Sorry @Kojoley but that PEP is still in draft status. Since there's still no standard about namespaces in Python yet, I think we need to work around this bug both in IPython and Spyder (see spyder-ide/spyder-kernels#332) in the meantime.

Kojoley:
PEP-667 is indeed a draft atm.

[`PyEval_GetLocals`](https://docs.python.org/3.11/c-api/reflection.html#c.PyEval_GetLocals) promises to `Return a dictionary of the local variables in the current execution frame, or NULL if no frame is currently executing.` and since CPython directly returns `f_locals` https://github.com/python/cpython/blob/c5bfb88eb6f82111bb1603ae9d78d0476b552d66/Python/ceval.c#L5729-L5743 means that Pytorch breaks CPython API?

ccordoba12:
> and since CPython directly returns f_locals means that Pytorch breaks CPython API?

You're right, most probably that's the case. However, I'd like to know if we can be pragmatic here, in the sense that it's much easier for us to work around this problem in Pytorch and any other libraries that break CPython API like it, than chasing and submitting PRs for them (given that we're always understaffed and with a lot of things to do).

Also, @impact27's patch is pretty small and non-invasive. If that wouldn't be the case, I'd agree with you on fixing things upstream instead.

Carreau:
Let's get that in, and I'll open an issue to track maybe reverting/simplifying/adding a known failure test to re-simplify later.

ccordoba12:
Thanks a lot @Carreau!

PR #13269, opened 2021-11-12
Fix DeprecationWarning: There is no current event loop

kkirsche:
Fix:

```
/Users/redacted/.asdf/installs/python/3.10.0/lib/python3.10/site-packages/IPython/terminal/interactiveshell.py:464: DeprecationWarning: There is no current event loop
  old_loop = asyncio.get_event_loop()
```

Per https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop this method is deprecated and get_running_loop is preferred over this method moving forward. As IPython requires 3.7+ now, this can be updated

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -505,7 +505,7 @@ class TerminalInteractiveShell(InteractiveShell):
         # while/true inside which will freeze the prompt.
 
         try:
-            old_loop = asyncio.get_event_loop()
+            old_loop = asyncio.get_running_loop()
         except RuntimeError:
             # This happens when the user used `asyncio.run()`.
             old_loop = None


kkirsche:
Test failures appear to be unrelated to my changes. Please let me know if that is incorrect.

Example failure:
```
ERROR: IPython.core.tests.test_inputtransformer2.test_check_complete
  File "/opt/hostedtoolcache/Python/3.9.8/x64/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/ipython/ipython/IPython/core/tests/test_inputtransformer2.py", line 270, in test_check_complete
    assert cc("a \\ ") == ("invalid", None)  # Nothing allowed after backslash
  File "/home/runner/work/ipython/ipython/IPython/core/inputtransformer2.py", line 753, in check_complete
    res = compile_command(''.join(lines), symbol='exec')
  File "/opt/hostedtoolcache/Python/3.9.8/x64/lib/python3.9/codeop.py", line 178, in __call__
    return _maybe_compile(self.compiler, source, filename, symbol)
  File "/opt/hostedtoolcache/Python/3.9.8/x64/lib/python3.9/codeop.py", line 98, in _maybe_compile
    code2 = compiler(source + "\n\n", filename, symbol)
  File "/home/runner/work/ipython/ipython/IPython/core/inputtransformer2.py", line 786, in __call__
    return compile(*args, **kwds)
SystemError: Negative size passed to PyUnicode_New
```

Carreau:
I'm tempted to think this is the new Python 3.9.8 release. I'll try to narrow it down.

Carreau:
I guess it is better to backport that to 7.x to not get deprecation warnings there ? Though 7.x it's unlikely to ever support 3.11.

Kojoley:
This seems to be totally wrong, after this commit I get:
```
In [1]: import asyncio

In [2]: %gui asyncio
ERROR:root:There is no current event loop in thread 'MainThread'.
```

kkirsche:
Most likely related, but "totally wrong" implies this didn't fix the original issue which it did.

Kojoley:
Sure, but fixing the deprecation warning from `asyncio` by introducing an error seems totally wrong to me.

I suspect the problem is in this requirement on `asyncio.get_running_loop()`: `This function can only be called from a coroutine or a callback.` (https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop)

kkirsche:
Ok to say it more clearly, the way the note was written comes off as presumptively that this is intentional instead of understanding that sometimes bugs get introduced while fixing a different one. My point was simply it fixed one and introduced another. So the next step is making another improvement to fix this bug. 

As it was written, at least for me, it turns off contributing to solving the new issue because of the tone.

Sorry a bug impacted your work, that wasn't intentional.

minrk:
It seems really bizarre that they chose to deprecate `get_event_loop` but provided no replacement (`get_running_loop` is not equivalent unless already running within asyncio ).

I need to dig in a little bit, but if this function isn't called from within a coroutine already, the current behavior is to always set `old_loop = None` _even if there is a current loop_, which means _clearing_ the current event loop. But since `set` has been called, later calls to `get_event_loop` will not create it as needed.

Avoiding the deprecation looks like it's going to be more difficult than a one-line patch, because the _functionality_ provided by `get_event_loop` is deprecated with no replacement. We have lots of `get_event_loop` calls, and this was only one. I think we need to reimplement `get_event_loop` ourselves, to track persistent thread-local loops while they aren't running, before we can remove calls to `get_event_loop`.

kkirsche:
Sorry for the issue, but thank you for the feedback. Please revert as needed if not already done so.

minrk:
#13349 should fix it, following the pattern in #13289

Thanks for your patches! I added a test that would have caught the error, too.

PR #13270, opened 2021-11-12
Add Xfail test on 3.9.8

Carreau:
See https://bugs.python.org/issue45738 once fixed (or later if not fixed),
we can also catch system error only on the relevant Python versions and
that _should_ be ok, but I want to avoid doing it for now.

### DIFF
--- a/IPython/core/tests/test_inputtransformer2.py
+++ b/IPython/core/tests/test_inputtransformer2.py
@@ -5,11 +5,14 @@ more complex. See test_inputtransformer2_line for tests for line-based
 transformations.
 """
 import string
+import sys
+from textwrap import dedent
+
+import pytest
 
 from IPython.core import inputtransformer2 as ipt2
-from IPython.core.inputtransformer2 import make_tokens_by_line, _find_assign_op
-
-from textwrap import dedent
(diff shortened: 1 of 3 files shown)


IPython/core/tests/test_inputtransformer2.py
```diff
@@ -256,20 +259,43 @@ def __init__(self, s):
...
+    pytest.param("\\\r\n", "incomplete", 0),
+    pytest.param("a = '''\n   hi", "incomplete", 3),
+    pytest.param("def a():\n x=1\n global x", "invalid", None),
+    pytest.param(
+        "a \\ ",
+        "invalid",
+        None,
+        marks=pytest.mark.xfail(
+            reason="Bug in python 3.9.8 – bpo 45738",
+            condition=sys.version_info[:3] == (3, 9, 8),
```
Kojoley, line 278: Maybe also add `raises=SystemError`?

Carreau: +1, also added strict=True.

Carreau:
Ok, I"m not going to wait for appveyor. Also I've activated rolling builds to cancel previously  submitted jobs when new commits are pushed.

PR #13278, opened 2021-11-13
Use pathlib in historyapp.py

GalBr:
Replace os with pathlib Path in historyapp.py

Related to #12515


Carreau:
And congrats for a second successful contribution !

PR #13279, opened 2021-11-13
Port `deepreload` to `importlib`

Kojoley:
`imp` is deprecated since Python 3.3 and scheduled for removal in Python 3.12

The code is severely undertested, I added several, and also fixed a bug that `modules_reloading` was not cleaned-up on an import failure.


Carreau:
Thanks, and yes that some 20ish+ years old code, circa Python 2.2 Thanks !

PR #13290, opened 2021-11-16
Use sphinxify (if available) in object_inspect_mime path

fasiha:
Background: if we use `%pinfo` line magic (or equivalently `?`; and similarly for `pinfo2` and `??`), we will call `interactiveshell`'s `_inspect` method which automatically uses the docrepr module (if available and activated) to beautifully format docstrings.

However, if we use the lower-level `object_inspect_mime` public method in `interactiveshell`, docrepr is not used.

(Sidenote: practically speaking, this function is called from JupyterLab's Context Help, which creates an "inspect_request" message, which in turns calls `object_inspect_mime`.)

This commit enables `object_inspect_mime` to use the same functionality as `_inspect`: see https://github.com/ipython/ipython/blob/66843b1ea573772309a6b29a69b2afd4641ae0d1/IPython/core/interactiveshell.py#L1763-L1769

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1813,8 +1813,9 @@ class InteractiveShell(SingletonConfigurable):
         with self.builtin_trap:
             info = self._object_find(oname)
             if info.found:
+                docformat = sphinxify if self.sphinxify_docstring else None
                 return self.inspector._get_info(info.obj, oname, info=info,
-                            detail_level=detail_level
+                            detail_level=detail_level, formatter=docformat
                 )
             else:
                 raise KeyError(oname)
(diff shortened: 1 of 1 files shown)


jasongrout (approved):
Thanks!

jasongrout:
> (Sidenote: practically speaking, this function is called from JupyterLab's Context Help, which creates an "inspect_request" message, which in turns calls `object_inspect_mime`.)

FYI, to fill in the connection here, JLab sends an `inspect_request` message, which ipykernel receives and then ipykernel calls `object_inspect_mime`: https://github.com/ipython/ipykernel/blob/b74ab3767df94abc1404f7ee7892f6e9a8e32317/ipykernel/ipkernel.py#L470

The upshot here is that even the classic notebook (which sends the same message) should now start getting formatted replies if docrepr is available. (For example, if you *execute* `myfunction?` in classic notebook with docrepr installed, it sends back formatted text, but if you `shift-tab` on `myfunction`, then expand the tooltip with the caret icon, you'll see that you do *not* get back formatted text - this change should fix this inconsistency there too)

Carreau:
Thanks, that seem reasonable.

PR #13296, opened 2021-11-18
Enable `check_make_token_by_line_never_ends_empty` test

Kojoley:
It was added in 28a8ea639a7a37bfb03ee87d397ea5e69dcc537e but missed `test_` prefix in the name.

Note: There is a type mismatch in calls to `make_tokens_by_line` but that does not lead to an error.

### DIFF
--- a/IPython/core/tests/test_inputtransformer2.py
+++ b/IPython/core/tests/test_inputtransformer2.py
@@ -143,7 +143,8 @@ def null_cleanup_transformer(lines):
     """
     return []
 
-def check_make_token_by_line_never_ends_empty():
+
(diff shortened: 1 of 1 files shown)


Kojoley:
Meh,
```
  /home/runner/work/ipython/ipython/IPython/core/inputtransformer2.py:512: UserWarning: `make_tokens_by_line` received a list of lines which do not have lineending markers ('\n', '\r', '\r\n', '\x0b', '\x0c'), behavior will be unspecified
```

Not sure whether that warning was expected or not.

Carreau:
I think that's fine and I't likely due to really old compatibility with readline.

PR #13298, opened 2021-11-18
test_path: revive match tests

Kojoley:
These tests had not been running on Nose and I did not spot that previously.


Carreau:
the only remaining test is RTD, which is about docs, which this PR does not touch. Merging.

PR #13306, opened 2021-11-19
Preserve function/method identity in magic decorators

Kojoley:
`decorator(call, func)` does `func.__doc__ = call.__doc__` and so on, where `func` is the original function and `call = lambda f, *a, **k: f(*a, **k)`. Because of that ipdoctest is not able to see the function original docstring, leading to missing tests. The resulting wrapper is simply calling the original function and thus is not needed.

Pytest finds additional 17 tests with this.


Carreau:
Awesome thanks !

Issue #12192, opened 2020-03-17
Infinite loop when using ipdb on ipython 7.13+


werdeil:
I use pytest with appium and when I use ipdb.set_trace() I'm stuck in a loop:

```
Exception in thread Thread-21249:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/vverdeil/workspace/venv/testenv/lib/python3.7/site-packages/IPython/terminal/debugger.py", line 102, in in_thread
    line = self.pt_app.prompt()
  File "/Users/vverdeil/workspace/venv/testenv/lib/python3.7/site-packages/prompt_toolkit/shortcuts/prompt.py", line 986, in prompt
    return self.app.run()
  File "/Users/vverdeil/workspace/venv/testenv/lib/python3.7/site-packages/prompt_toolkit/application/application.py", line 788, in run
    return get_event_loop().run_until_complete(self.run_async(pre_run=pre_run))
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/events.py", line 644, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'Thread-21249'.
```

As you can see I reached thread 21249 before managing to kill the process.

With the same code it works as expected on ipython 7.12, so I guess is it linked to https://github.com/ipython/ipython/pull/12141/

Am I the only one having this bug?

zogzog:
I just discovered that same exact thing today.

srggrs:
I can confirm that this happen to me as well:
```
python=3.6.10
ipython=7.13.0
ipdb=0.12.3
```

igandhi:
I believe this is an issue in `7.13`. I was able to resolve this by downgrading to iPython `7.10.2`

werdeil:
Any news on this? I tried python 7.17.0 and I still have the same issue...

werdeil:
Tried with python 7.19.0 and I still have the issue, any news?

werdeil:
[...]

Issue #12267, opened 2020-04-25
Calling SIGINT twice on suspended ipython results in exception


PythonCoderAS:
The most surprising thing is that replicating in bash and zsh results in two different ways to get to the same exception. 

Ipython has to be suspended first, using Ctrl+Z. 

# Bash

```
bash-3.2$ ipython
Python 3.7.7 (default, Mar 10 2020, 15:43:03) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.


[1]+  Stopped                 ipython
bash-3.2$ kill -2 %1

[1]+  Stopped                 ipython
bash-3.2$ kill -2 %1

[1]+  Stopped                 ipython
bash-3.2$ fg
ipython
In [1]:                                                                                                                 
KeyboardInterrupt escaped interact()


Traceback (most recent call last):
  File "/usr/local/bin/ipython", line 8, in <module>
    sys.exit(start_ipython())
  File "/usr/local/lib/python3.7/site-packages/IPython/__init__.py", line 126, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/traitlets/config/application.py", line 664, in launch_instance
    app.start()
  File "/usr/local/lib/python3.7/site-packages/IPython/terminal/ipapp.py", line 356, in start
    self.shell.mainloop()
  File "/usr/local/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 558, in mainloop
    self.interact()
  File "/usr/local/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 541, in interact
    code = self.prompt_for_code()
  File "/usr/local/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 469, in prompt_for_code
    **self._extra_prompt_options())
  File "/usr/local/lib/python3.7/site-packages/prompt_toolkit/shortcuts/prompt.py", line 994, in prompt
    return self.app.run(set_exception_handler=set_exception_handler)
  File "/usr/local/lib/python3.7/site-packages/prompt_toolkit/application/application.py", line 812, in run
    self.run_async(pre_run=pre_run, set_exception_handler=set_exception_handler)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
[...]

PR #13311, opened 2021-11-21
Use ThreadPoolExecutor for debugger cmdloop, solving infinite thread-creation and traceback loop

kmaork:
In `TerminalPdb.cmdloop`, the prompt is run on a separate thread. This is done using `threading.Thread.start` + `threading.Thread.join` in a loop. When `self.pt_app.prompt` fails with an exception other than `EOFError` or `KeyboardInterrupt`, it causes an infinite loop that creates threads which fail immediately and print their traceback.

On my linux it is recreatable using the following script:
```python
from concurrent.futures import ProcessPoolExecutor
import ipdb

ipdb.set_trace()
ProcessPoolExecutor(1).submit(ipdb.set_trace).result()
```

Instead of using the low-level `threading.Thread`, I propose using `ThreadPoolExecutor`, which can handily retrieve return values from the thread, and reraise exceptions raised in it. It simplifies the existing code and prevents the infinite loop issue.

This also solves an issue with the debugger which I ran into using `madbg` - https://github.com/kmaork/madbg/issues/28, in which sending a `SIGINT` to the debugger while it is in the background also caused this infinite traceback loop (maybe an issue similar to https://github.com/ipython/ipython/issues/12267), so a backport would be awesome 😄

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -1,11 +1,8 @@
 import asyncio
 import os
 import sys
-import threading
 
 from IPython.core.debugger import Pdb
-
-
 from IPython.core.completer import IPCompleter
 from .ptutils import IPythonPTCompleter
 from .shortcuts import create_ipython_shortcuts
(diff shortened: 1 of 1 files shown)


Carreau:
Oh ! Thanks ! That's looks great ! Let't try it.

kmaork:
Thank you! Can you please backport it to version 7?

Carreau:
> Thank you! Can you please backport it to version 7?

Let's try. 

@meeseeksdev backport to 7.x

Carreau:
Ok, i'll try to backport manually later.

kmaork:
Thank you @Carreau !!

kmaork:
I think this closes #12192

Issue #13316, opened 2021-11-23
Remove pre-3.8 codepath.


Carreau:
Once https://github.com/ipython/ipython/pull/13315 is merged the master branch will not support 3.7 anymore, this gives opportunities to remove all the pre-3.8 codepath.

If you like to contribute, find 1 block of code that is execute only before python 3.8 (look for sys.version_info in the code base), 
and simplify it. 

It may be more than just deleting code, as you might be then able to cleanup imports, or simplify more logics.

Carreau:
https://app.codecov.io/gh/ipython/ipython/compare/13315/changes might be a good way to see code that can potentially be removed.

GalBr:
@Carreau Should I continue working on this issue or should we save it for newer people?

Carreau:
> @Carreau Should I continue working on this issue or should we save it for newer people?

Up to you, depend how comfortable you are around the code base. It's a good way to get started maybe try to remove the conditional code and try to understand/simplify/improved nearby code ?

PR #13317, opened 2021-11-23
Remove pre-3.8 codepath in interactiveshell.py

GalBr:
Related to #13316 .
I'll probably do more relevant work in interactiveshell.py in this branch if that's ok. I just want to get feedback on what I do in the process, to make sure I do it correctly.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -117,12 +117,8 @@ else :
     from ast import Module as OriginalModule
     Module = lambda nodelist, type_ignores: OriginalModule(nodelist)
 
-if sys.version_info > (3,6):
-    _assign_nodes         = (ast.AugAssign, ast.AnnAssign, ast.Assign)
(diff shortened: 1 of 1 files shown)


GalBr:
`ast_asyncify` is only being called from a `sys.version_info < (3,8)` branch, but there's this if inside this function:
`if sys.version_info >= (3,8):
        return ast.parse(cell)`

This means we never use `_ast_asyncify`. How do we want to handle this? Remove this function completely or keep it and just remove the code inside it that handles the `sys.version_info < (3,8)` case?

GalBr:
`init_deprecation_warnings` is not doing anything because it checks for ` sys.version_info < (3,7)`.
Should it be removed completely or should it be changed to `< (3,8)`?

Carreau:
Yeas, I think that ast_asyncify can be removed, I sent upstream patches to CPython to make it unnecessary.

Carreau:
Same for anything that is uncalled anymore, we should be able to remove them.

PR #13318, opened 2021-11-23
Move most of Windows test runners to GitHub Actions

Kojoley:
Leave only 32-bit jobs on Appveyor, I don't know how to configure actions/setup-python to install 32-bit CPython.


Carreau:
Thanks !

PR #13324, opened 2021-11-24
Remove pre-3.8 codepath in interactiveshell.py

GalBr:
Related to #13316.
The last commit is pretty big. The main change is the removal of ast_asyncify() and the rest is the removal of affected code.
I checked it against the coverage changes seen in https://app.codecov.io/gh/ipython/ipython/compare/13315/changes for this file.


Carreau:
Thanks, that's great !

Hope some of the code here made sens.

PR #13330, opened 2021-11-26
deepreload: Use `sys.builtin_module_names` to exclude builtins

Kojoley:


Carreau:
Ah, interesting, I was not aware of this value.

PR #13343, opened 2021-11-26
Allow more control over what sections of the "info" messages are desired via a sections blocklist

fasiha:
Currently, IPython responds to (1) `pinfo` messages or to (2) ipykernel's "inspect" messages (which invoke IPython's `object_inspect_mime` function) by building a document containing a number of sections, for example, for `numpy.sin` we get 
- Call signature
- Type
- String form
- File
- Docstring
- Class docstring

<details>
<summary>(Code snippet to generate HTML, for example)</summary>

```py
from IPython.core.oinspect import Inspector
import numpy as np

a = Inspector()._get_info(np.sin)
with open('foo.html', 'w') as fid:
    fid.write(a['text/html'])
```

</details>

In order to support building richer documentation browsers, it would be helpful to customize which sections are desired in the output. For example, we might be interested in displaying *just* the docstring to beginners using Jupyter notebooks.

Of course IPython's clients are free to analyze the text/HTML and remove the sections they don't want, but it would be nice to specify which sections a client wants because it can
- reduce runtime (generating HTML documentation with [Docrepr](https://github.com/spyder-ide/docrepr) involves calling Sphinx),
- reduce bandwidth to transmit documentation that won't be displayed, and
- generally fits with our desire to let frontends for exactly the information they want of backends.

This PR makes a small change to the core private method, `Inspector._get_info`, allowing it to skip sections that callers don't want based on a new keyword argument. This method is invoked by both "info" and "inspect" messages, and so I've threaded the new argument up to some public methods.

With this in place, clients can customize which sections they *don't* want to see. A couple of open questions remain:

- are we interested in *further* customization of the output? This PR only adds a list of sections to *omit*. The inverse of that would be explicitly a list of sections you *want*. I chose to add just the former to gauge interest, I am happy to add the latter.
- Are there other users of Inspector's private `_get_info` that I'm missing?

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1804,7 +1804,7 @@ class InteractiveShell(SingletonConfigurable):
         """Get object info as formatted text"""
         return self.object_inspect_mime(oname, detail_level)['text/plain']
 
-    def object_inspect_mime(self, oname, detail_level=0):
+    def object_inspect_mime(self, oname, detail_level=0, omit_sections={}):
         """Get object info as a mimebundle of formatted representations.
 
         A mimebundle is a dictionary, keyed by mime-type.
@@ -1820,6 +1820,7 @@ class InteractiveShell(SingletonConfigurable):
                     info=info,
                     detail_level=detail_level,
                     formatter=docformat,
+                    omit_sections=omit_sections,
                 )
             else:
                 raise KeyError(oname)
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1804,7 +1804,7 @@ class InteractiveShell(SingletonConfigurable):
         """Get object info as formatted text"""
         return self.object_inspect_mime(oname, detail_level)['text/plain']
 
-    def object_inspect_mime(self, oname, detail_level=0, omit_sections={}):
+    def object_inspect_mime(self, oname, detail_level=0, omit_sections=()):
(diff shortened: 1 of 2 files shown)


fasiha:
FYI the failing test in "Downstream tests / test" ([link](https://github.com/ipython/ipython/runs/4337574894?check_suite_focus=true#step:7:24)) passes locally.

Carreau:
Thanks, 

That make a lot of sens. Should we avoid the mutable default argument, and use an empty tuple for the default value ?

fasiha:
> Should we avoid the mutable default argument, and use an empty tuple for the default value ?

Thanks for noting this, sorry for making this basic newbie mistake! Changing the default to be `()`, which is immutable—of course callers can provide any object that supports `in` including set 😄.

Carreau:
No problem, I often also have the weird impression that empty set are immutable, unlike their non-empty counterpart. 
I think it's a weird human brain thing, that `{}` looks like a literal and has no "variable" part :-)

fasiha:
For completeness, I should add that there is one more place where this could be leveraged in IPython: in the `pinfo` and `pinfo2` magics: https://github.com/ipython/ipython/blob/9c54f3ee28612e334f6f4b609cfc2950b1de62d2/IPython/core/magics/namespace.py#L58-L67

It's not clear to me how to incorporate the `omit_sections` keyword argument into these two calls to `self.shell._inspect` 🤔.

Carreau:
Also `{}` are dict, not empty set...

PR #13345, opened 2021-11-27
Don't add ipython unconditionally to sys.path

### LINKED ISSUE #13294: Problems due to adding IPython profile dir (~/.ipython) to sys.path
When starting IPython, the profile directory gets added to sys.path. This means any `import` statement will also look for modules in that folder and in some cases allow you to import packages that aren't installed.

For instance, after using the Cython magic, I can import cython even when cython is not installed in my environment.

This causes issues when using packages/modules that rely on ImportError exceptions to check for optional features, for instance. We have a package that uses `cython.compiled` to check if one is using a plain-code or compiled version of a package. Due to IPython's behaviour regarding the profile directory, we cannot rely on the ImportError anymore - because `import cython` will just work.

I guess IPython uses the addition to sys.path to discover extensions and such. But giving the interference with module discovery, I think different ways for this should be found.

Carreau:
Load ext will already do it with a context manager.

This should remove false positive when import modules, whithout breaking
functionalities.

Maybe closes #13294 (making this part of 8.0 release make sens).

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -185,14 +185,16 @@ class BaseIPythonApplication(Application):
                 get_ipython_package_dir(), u'config', u'profile', change['new']
         )
 
-    add_ipython_dir_to_sys_path = Bool(False,
+    add_ipython_dir_to_sys_path = Bool(
+        False,
         """Should the IPython profile directory be added to sys path ?
 
         This option was non-existing before IPython 8.0, and ipython_dir was added to
         sys path to allow import of extensions present there. This was historical
         baggage from when pip did not exist. This now default to false,
         but can be set to true for legacy reasons.
-        """).tag(config=True)
+        """,
+    ).tag(config=True)
 
     ipython_dir = Unicode(
         help="""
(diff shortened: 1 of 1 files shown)


IPython/core/application.py
```diff
@@ -185,6 +185,15 @@ def _profile_changed(self, change):
                 get_ipython_package_dir(), u'config', u'profile', change['new']
         )
 
+    add_ipython_dir_to_sys_path = Bool(False,
+        """Should the IPython profile directory be added to sys path ?
+
+        This option was non-existing before IPython 8.0, and ipython_dir was added to
+        sys path to allow import of extensions present there. This was historical
+        baggage from when pip did not exist. This now default to false,
+        but can bet to true for legacy reasons
```
minrk, line 194:
```suggestion
        but can be set to true for legacy reasons.
```

IPython/core/application.py
```diff
@@ -294,21 +303,22 @@ def _ipython_dir_changed(self, change):
...
-            path = os.path.join(new, d)
-            try:
-                ensure_dir_exists(path)
-            except OSError as e:
-                # this will not be EEXIST
-                self.log.error("couldn't create path %s: %s", path, e)
-        self.log.debug("IPYTHONDIR set to: %s" % new)
+        if self.add_ipython_dir_to_sys_path:
+            str_path = os.path.abspath(new)
+            #sys.path.append(str_path)
```
minrk, line 308: I'm not sure this is right - the append is unconditionally skipped, and the subdir creation presumably shouldn't be under this condition?

Carreau:
```suggestion
            sys.path.append(str_path)
```

Carreau: Yeah, like a typo just before committing. 

I"d like to remove this option in the future, so my hope was to never create the dir if it is not needed.
Nowadays people can install extensions with pip so nothing should go in that dir maybe ?

I need to try with the cython extension see how it behaves.

PR #13348, opened 2021-12-01
make sure to run async tests

minrk:
there are some `async def` tests, but they are skipped without `mark("asyncio")`

I think this reveals a real failure, that `%%bash` and other script magics cannot be run in asyncio (i.e. cannot be run in ipykernel).

The tests were added in #12639, but I don't think they've ever passed.

### DIFF
--- a/IPython/conftest.py
+++ b/IPython/conftest.py
@@ -1,15 +1,30 @@
-import types
-import sys
 import builtins
+import inspect
 import os
-import pytest
 import pathlib
 import shutil
+import sys
+import types
+
+import pytest
 
 # Must register before it gets imported
 pytest.register_assert_rewrite("IPython.testing.tools")
 
 from .testing import tools
+
+
+def pytest_collection_modifyitems(items):
+    """This function is automatically run by pytest passing all collected test
+    functions.
+
+    We use it to add asyncio marker to all async tests and assert we don't use
+    test functions that are async generators which wouldn't make sense.
+    """
+    for item in items:
+        if inspect.iscoroutinefunction(item.obj):
+            item.add_marker("asyncio")
+        assert not inspect.isasyncgenfunction(item.obj)
 
 
 def get_ipython():
@@ -32,7 +47,7 @@ def work_path():
     if path.exists():
         raise ValueError('IPython dir temporary path already exists ! Did previous test run exit successfully ?')
     path.mkdir()
-    yield 
+    yield
     shutil.rmtree(str(path.resolve()))
 
 
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -14,32 +14,85 @@ Python semantics.
 import ast
 import asyncio
 import inspect
+from functools import wraps
+
+_asyncio_event_loop = None
+
+
+def get_asyncio_loop():
+    """asyncio has deprecated get_event_loop
+
+    Replicate it here, with our desired semantics:
+
+    - always returns a valid, not-closed loop
+    - not thread-local like asyncio's,
+      because we only want one loop for IPython
+    - if called from inside a coroutine (e.g. in ipykernel),
+      return the running loop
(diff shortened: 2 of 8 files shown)


minrk:
There are various problems in the asyncio implementation of script magics. Running in the background doesn't work anymore because the call to `p.wait()` is a coroutine, but assumed to be blocking. These tests are currently skipped because of the use of unsupported `async def` tests.

I think the only way this can work is to put the asyncio loop in a background thread and make all calls blocking across threads via concurrent.futures.

minrk:
I'm not sure that the script magics can be done with asyncio as they are currently. At least not with `--bg`.

The main issues:

- the magic has certain requirements on the event loop, e.g. safe_child_watcher or ProactorEventLoop on Windows), which means it can't assume it will work on the current event loop
- in ipykernel, an asyncio loop will always be running already, which means asyncio calls cannot be blocking (cannot wait for any coroutine to finish on the main loop) and calls to `set_event_loop` will wreak havoc.
- putting an asyncio call in a background thread works for the _blocking_ case, but that means `stdout.read()` for instance is attached to a different loop, so backgrounded output can no longer be read:

```python
In [2] %%bash --bg --out stdout
  ...: echo 1

In [3]: await stdout.read()
RuntimeError: Task <Task pending name='Task-1711' coro=<InteractiveShell.run_cell_async() running at /Users/minrk/dev/ip/ipython/IPython/core/interactiveshell.py:3074> cb=[_run_until_complete_cb() at /Users/minrk/conda/lib/python3.9/asyncio/base_events.py:184]> got Future <Future pending> attached to a different loop
```


It would probably be easiest if we could assume that asyncio is _always_ running and make the magic itself and all its properties natively async. Is that possible? In what circumstances would that not work? %gui? trio/curio? Windows non-proactor?

It might also work to use nest_asyncio, which seems to make `run_until_complete` work inside the current loop. I'm not sure if the coroutines and such remain connected to the nested loop, however.

bollwyvl:
About the weirdest place I've run is in pyodide, and it is always async.
Indeed, cell magics might logically benefit from being assumed to be async,
but no doubt there are knock-on effects in some of those edge cases.

minrk:
asyncio stuff in IPython is definitely tricky. I think we can't (yet) assume asyncio is running, but we _definitely_ can't ever assume it's _not_ running, which the current script magics do, since it is always running in ipykernel.

I got it working (locally, at least) by putting the async process in a thread. `--bg` is still totally broken, but at least the blocking case works.

minrk:
@Carreau I think resolving this should probably block 8.0. I'll try to find some time this week.

Carreau:
Probably, but that will likely not impact the actual Ipython code.

minrk:
I think it does. The script magics don't work when run in asyncio anymore, which means they don't run in ipykernel. I think the fix probably belongs in IPython, or at least part of it.

minrk:
@Carreau I think I've come up with something that works. The ScriptMagics instance maintains a single long-running event loop in a thread, which replaces the BackgroundJobManager. Inter-thread communication is handled via `asyncio.run_coroutine_threadsafe`. The new implementation never interacts with the main event loop or policy, so it shouldn't be sensitive to what they are, and I think the end result is actually simpler than before.

The only bit I don't really like is the `AsyncIOProxy` wrappers which are needed to make an async object (the Process and its streams) awaitable from the main loop. These create proxy methods that call `asyncio.wrap_future(asyncio.run_coroutine_threadsafe(coro()))`. This does:

1. run a coroutine on the background asyncio loop
2. wrap the result in a `concurrent.futures.Future` for threadsafety
3. wrap _that_ in an `asyncio.Future` so it is awaitable in the current loop of the calling context, no matter what thread that is.

Carreau:
That looks good to me, i'll test more in depth later, but let's merge it so that at least people on the master branch get it.

PR #13349, opened 2021-12-01
get_running_loop is only valid in coroutines

minrk:
This fixes the problems introduced in #13269 by following the pattern in #13289. Using `get_running_loop` is only a valid way to get the event loop from inside a coroutine, which is not what's happening here.

I don't really understand why `asyncio.get_event_loop` is deprecated and `policy.get_event_loop` is not, but that is the case, at least for now.

await doesn't work in 7.30 because of this, so this should perhaps be in 7.30.1

### DIFF
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -12,17 +12,28 @@ Python semantics.
 
 
 import ast
+import asyncio
 import inspect
 
 
 class _AsyncIORunner:
+    def __init__(self):
+        self._loop = None
+
+    @property
+    def loop(self):
+        """Always returns a non-closed event loop"""
+        if self._loop is None or self._loop.is_closed():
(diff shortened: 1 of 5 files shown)


minrk:
This is running into various problems with invalid assumptions about asyncio in the script magics (#13348)

minrk:
In two commits because the first commit should be backported to 7.x, while the second is only for the 8.0 updates to script magics.

minrk:
FWIW, referring to [the Python patch deprecating get_event_loop](https://bugs.python.org/issue39529), it is an _accident_ that `policy.get_event_loop` is not also deprecated. So this and #13289 is only an extremely temporary measure.

We should move to our own handle on the event loop, and remove all calls to` get_event_loop` and `set_event_loop` (including the policy methods). That should be 8.0-only, I think.

PR #13357, opened 2021-12-01
Remove all the lib/inputhook logic and related files.

Carreau:
1) for terminal ipython hooks are in pt_inputhook,
2) for ipykernel the inputhook are in ipykernel itself.


Carreau:
@tacaswell you are a user of --simple-prompt with emacs, can you confirm it does not break anything ?

tacaswell:
It seems to not break anything new.  It seems to work as I expect for Qt5, Qt6, and tk (without using the magic).


---

With tk it seems that with both this branch and 7.30 

```
✔ 12:06:44 $ ipython --simple-prompt
Python 3.8.12+ (heads/3.8:c37a0d9c0a, Nov 29 2021, 09:42:48) 
IPython 7.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %matplotlib tk

In [2]: import matplotlib.pyplot as plt

In [3]: plt.plot(range(5))
Out[3]: [<matplotlib.lines.Line2D at 0x7f602f3d5a60>]

In [4]: plt.show()

In [5]: 

```

does not work (never get a window up), but

```
✔ 12:18:28 $ ipython --simple-prompt
Python 3.9.9 (main, Nov 20 2021, 21:30:06) 

In [1]: import matplotlib

In [2]: matplotlib.use('tkagg')

In [3]: import matplotlib.pyplot as plt

In [4]: plt.ion()
Out[4]: <matplotlib.pyplot._IonContext at 0x7fab7ceb6880>

In [5]: plt.plot(range(5))
Out[5]: [<matplotlib.lines.Line2D at 0x7fab5e9dfeb0>]

In [6]: 

```

works so something in the `%matplotlib` magic is breaking tk, but not this code!

tacaswell:
I understand the `%matplotlib tk` not working.  The issue is that as part of the magic IPython must be triggering the auto-backend selection logic which is picking Qt and then re-setting it to tk.  On import PyQt install the `PyOS_InputHook` function I assume that is interfering with tk's version of that somehow.

Because in `--simple-prompt` all of the input hook work is handled by CPython (presumably via readline), the wrong input hook breaks means no "live" tk, but doing `plt.show(block=True)` brings up the window as expected.

tacaswell (approved):

Carreau:
Thanks !

Carreau:
And my understanding is that most of that is unused anyway nowdays for all the `%matpltolib` stuff.

PR #13282, opened 2021-11-15
Python 3.10 fixes

Kojoley:
Supersedes #13080
Closes #13079
Closes #12412

PR #13371, opened 2021-12-04
Refactor inspect.getabsfile/getsource/getsourcelines + handle OSError + tests

Kojoley:
As noted in https://github.com/ipython/ipython/pull/13282#issuecomment-986017539 the fc57ec912eec9da4c33d8db5416fdb2dada706e7 did not fix all the places. 

`if hasattr(obj, '__class__'):` branches were removed because everything is an object and has a `__class__` in Python 3.


Kojoley:
I think fc57ec912eec9da4c33d8db5416fdb2dada706e7 + this worth backporting to 7.x branch.

Issue #13378, opened 2021-12-08
timeit zeros hard to count


mathandy:
This is a feature request -- I'll make a pull request if there's a desire.

It's bugged me for years that the output of the wonderful `timeit` magic function is often made difficult to use for quick comparisons due to the long strings of zeros in the loop count that are often not aligned when comparing multiple results.
Here's an example:
![image](https://user-images.githubusercontent.com/3175485/145135507-66e675ee-caf4-4096-8952-3a0bce8670b7.png)
Reading this, first you might think, "Wow a 10x speedup!"  and then count some zeros and realize it's really a 100x speedup.

Three potential solutions:
1. Used fixed widths for output values so results from multiple runs align
2. group the zeros with some delimiter as is often down outside the computing world (e.g. 100_000_000 -- CPython ignores underscores in numbers)
3. use scientific notation (e.g. 1e8)

To me, solution #2 or #3 makes the most sense.  It looks like this would only take a couple of characters to change:
![image](https://user-images.githubusercontent.com/3175485/145137264-c1fa108e-92f3-494a-956d-20bd500ef3c9.png)

mathandy:
Pull request #13379 fixes this issue.

PR #13379, opened 2021-12-08
Add comma groupings to timeit magic loop count

mathandy:
This is a pull request related to issue #13378.   I opted to use commas to group digits as the syntax `'{x:,}'.format(x=1000)` has been supported since at least 2.7.  That said, I'd think underscores might make more sense as CPython ignores underscores included in integers.

### Reiteration of the problem mentioned in #13378:
It's bugged me for years that the output of the wonderful `timeit` magic function is often made difficult to use for quick comparisons due to the long strings of zeros in the loop count that are often not aligned when comparing multiple results.
Here's an example:
image
Reading this, first you might think, "Wow a 10x speedup!"  and then count some zeros and realize it's really a 100x speedup.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -92,17 +92,15 @@ class TimeitResult(object):
                 pm = u'\xb1'
             except:
                 pass
-        return (
-            u"{mean} {pm} {std} per loop (mean {pm} std. dev. of {runs} run{run_plural}, {loops} loop{loop_plural} each)"
-                .format(
-                    pm = pm,
-                    runs = self.repeat,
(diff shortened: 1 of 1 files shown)


minrk (approved):

minrk:
This is great! I ran `darker` to satisfy the linter.

To be clear, though, the speedup really is 10-12x in this case, no matter how many loops it ran. The reported measurement is _per loop_, not the total for all loops, so the number of loops doesn't change the reported measurement. It measures N loops then reports `total time / N` so it's always the average value for _one_ run. The numbers are indeed always directly comparable.

PR #13397, opened 2021-12-10
enable formatting by default

Carreau:


luzfcb:
How to disable this feature permanently, globally?

adamchainz:
I've started a new issue for discussion, with PR: #13463

ayhanfuat:
One other option might be making it easier to disable formatting (especially when used in education). Maýbe something like `%disable_black`?

SimonBiggs:
I'd like to add in some encouragement. Thank you for all of your hard work @Carreau. We really appreciate IPython and everything you do to make it happen :slightly_smiling_face:. Thank you. I don't have a horse in this particular race... but... I thought it was important to say thank you, and counter some of the criticism that is going on around this.

Thanks @Carreau :slightly_smiling_face:.

PR #13401, opened 2021-12-11
Remove not working pip caching

Kojoley:
It [requires `requirements.txt`](https://github.com/actions/setup-python#caching-packages-dependencies) which does not exist, it also forces workers in forks to fail with `Error: getCacheEntry failed: Cache service responded with 500` message.


Kojoley:
The pip cache is actually working, and fails I observe seems to be a server issue https://github.com/actions/cache/issues/698

PR #13402, opened 2021-12-11
Disable coverage collection on PyPy

Kojoley:
Coverage collection overhead on PyPy is 330-420% (20-110% on CPython)


Carreau:
Ouch. Thanks!

PR #13403, opened 2021-12-13
Update version8.rst

MichaelTiemannOSC:
Fix typo (bog->big) and then reword grant information so that it reads more naturally.


Carreau:
thanks !

PR #13432, opened 2022-01-05
Fix high CPU load on macOS

culler:
With `gui tk` in effect IPython has very high CPU load in certain situations where Apple's window manager is in control of the Tk event loop.  These situations include when a menu is posted or when a file dialog is being displayed.  To demonstrate run:
`In [1]: gui tk`
`In [2]: import tkinter`
`In [3]: w = tkinter.Tk()`
Then press and hold the mouse button on the menubar at the top of the screen while top or Activity Monitor are visible.  You will see CPU load go to almost 200%.

This problem appeared with the change to  prompt_toolkit 3.0.  It has serious side effects because the high CPU load persists for quite a while after the file dialog is closed, and that interferes with processing the selected file.

This PR fixes the problem by using select to test whether the IPython input pipe is readable instead of using a Tcl filehandler.  The CPU load when using the version of the tk input hook in this PR is about 40% which is the same as for the Wish app.

### DIFF
--- a/IPython/terminal/pt_inputhooks/tk.py
+++ b/IPython/terminal/pt_inputhooks/tk.py
@@ -38,7 +38,9 @@ way we don't block the UI of for instance ``turtle`` and other Tk libraries.
 in readline. ``prompt-toolkit`` doesn't understand that input hook, but this
 will fix it for Tk.)
 """
+import sys
 import time
+import select
 
 import _tkinter
 import tkinter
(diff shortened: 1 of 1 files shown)


Carreau:
I've marked this as backport for 7.32 (which should be out at the end of January)

As there is an 8.0.0RC published, I might wait to merge that for just after 8.0 release (likely next week), and will likely release a 8.1 at the same time as 7.32 at the end of the month. Is that ok with you ?

culler:
Sure.  I am fine with whatever is more convenient for you.

prachetasolutions:
Remove any dependency and mention of nose. Indeed this is impeding packaging on linux, and on Python 3.10.

culler:
After more testing, I do not think this PR is ready to be merged.  While it solves one problem, it introduces a different one.  Namely, the Tk UI become sluggish.  I think the cause is that the code in this PR requires a python function call for each Tk event to be processed.  The existing code, if it worked correctly, would allow Tk events to be processed with no delays until the pipe became readable.  I am going to investigate whether this can be addressed at the Tk level.  I suggest delaying merging this PR until I have looked into that.

culler:
Here is a summary of what we have found by experimenting:

- With prompt_toolkit 2.0.10 the existing tk.py input hook works great.  The Tk GUI is responsive and stable and there is no high CPU usage.  While pressing the mouse button on the menubar the CPU usage is about 16%.  However, when using the patched tk.py in this PR, Tk is completely dead.  Activating a Tk window creates the spinning wheel of death.
- With prompt_tookit >= 3.0.0 the existing tk.py input hook makes everything unstable.  CPU usage when pressing the mouse button on the menubar is close to 200% and there are crashes.
- With prompt_toolkit >= 3.0.0 the patched tk.py input hook in this PR makes some improvement but not enough.  The CPU usage when pressing the mouse button on the menubar drops to about 40%.  The Tk GUI functions, but it is very slow and unresponsive.  It appears as though there is about a 100 millisecond delay after generating a mouse event before Tk responds.

Clearly this PR does not solve the problem.  The problem did not exist before prompt_toolkit was rewritten to use asyncio.  The asyncio module appears to somehow interfere with Tk's event loop.  We have not gotten to the bottom of this yet.  But this PR is not the answer.

PR #13436, opened 2022-01-10
Fix and test for "async with does not allow new lines".

### LINKED ISSUE #12975: `async with` doesn't allow newlines
When writing a `with` statement, pressing enter goes onto a new line, rather than immediately running the code:
```py
In [26]: with example:
    ...:     do_1()[enter]
    ...:     [<-pointer goes here]
```
However, when using `async with`, this does not apply to any lines except for the first after the colon:
```py
In [26]: async with example:
    ...:     do_1()[enter]
[code is run]
```

(Tested in latest PyPI: `IPython.__version__ == "7.23.1"`)

### LINKED ISSUE #13424: Multi-line edit executes too early with await
I was trying to type this example in IPython:
```
from prefect.client import OrionClient

async with OrionClient() as client:
    deployment = await client.read_deployment_by_name("Addition Machine/my-first-deployment")
    flow_run = await client.create_flow_run_from_deployment(deployment)
```
However, typing ENTER after the first line with an `await` statement stops the multi-line edit too soon and executes the code:
```
In [16]: async with OrionClient() as client:
    ...:     deployment = await client.read_deployment_by_name("Addition Machine/my-first-deployment")

In [17]:
```

I think this is a general bug with `await`, the same bug happens with any other example, e.g.
```
In [17]: with open():
    ...:     spam = await Thing()
```

@minrk  - related to #13348 ?

Carreau:
Use the opportunity to add a test, and parametrise a few other, plus set
the correct stacklevel.

Closes #12975

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -507,9 +507,12 @@ def make_tokens_by_line(lines:List[str]):
 
     #   reexported from token on 3.7+
     NEWLINE, NL = tokenize.NEWLINE, tokenize.NL  # type: ignore
-    tokens_by_line:List[List[Any]] = [[]]
-    if len(lines) > 1 and not lines[0].endswith(('\n', '\r', '\r\n', '\x0b', '\x0c')):
-        warnings.warn("`make_tokens_by_line` received a list of lines which do not have lineending markers ('\\n', '\\r', '\\r\\n', '\\x0b', '\\x0c'), behavior will be unspecified")
+    tokens_by_line: List[List[Any]] = [[]]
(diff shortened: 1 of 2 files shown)


naterush (approved):
@Carreau I'm new here, so not sure what the etiquette is on reviews. You mentioned elsewhere you're the main PR reviewer, so thought I'd try to help out on your PRs. If it's not helpful (or there's a more helpful way to be involved), please let me know.

I've verified these changes move the reporting of the warning to the call site of `make_tokens_by_line`, which seems to be the primary goal of this PR. LGTM ::))

Carreau:
Thanks, that is helpful review. 

Even a LGTM, and making sure I don't forget a print statement or anything else is welcome.

PR #13450, opened 2022-01-13
Show suggestions in error messages in Python 3.10

### LINKED ISSUE #13445: NameError/AttributeError suggestions in Python 3.10 error messages
Python 3.10 added suggestions for AttributeError and NameError in the error messages. It seems the suggestions are not stored in the exception object but calculated when Error is displayed. There is a note that that this won't work with IPython but it will be good to see if it's feasible. Opening an issue for discussion.

https://bugs.python.org/issue38530 
https://docs.python.org/3/whatsnew/3.10.html#attributeerrors 

Discussion on HN : https://news.ycombinator.com/item?id=29909799

tirkarthi:
Fixes #13445

### DIFF
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -11,6 +11,8 @@ from textwrap import dedent
 import traceback
 import unittest
 
+import pytest
+
 from IPython.core.ultratb import ColorTB, VerboseTB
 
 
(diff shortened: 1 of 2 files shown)


Carreau:
Hum , I think we might be able to get something better with 

```python
def structured_traceback(...):
    from difflib import get_close_matches
    get_close_matches(value.name, dir(value.obj)
```

And re-implemented the match logic (even if slightly different). That way we can properly highlight tokens.

aroberge:
I believe that the suggestion to use difflib would only work for AttributeError cases but not for NameError (where the misspelled name can be found in the local or global scope).

aroberge:
Furthermore, `difflib` gives different results from those given by cPython.
```python
>>> import math
>>> math.Pi
AttributeError: module 'math' has no attribute 'Pi'. Did you mean: 'pi'?
>>> from difflib import get_close_matches
>>> get_close_matches('Pi', dir(math))
[]
```

tirkarthi:
CPython has done several tweaks to get better results. IMO, it's better to use the suggestions from the compiler itself to avoid any discrepancy and to avoid reimplementing/maintaining the algorithm.

https://github.com/python/cpython/pull/25412
https://github.com/python/cpython/pull/25443
https://github.com/python/cpython/pull/25460
https://github.com/python/cpython/pull/25584
https://github.com/python/cpython/pull/25776

Carreau:

PR #13464, opened 2022-01-15
Revert "enable formatting by default"

### LINKED ISSUE #13463: Revert “Black by default”
Version 8.0 added automatic formatting with Black. This was added in #13397 without much discussion - at least there is none linked.

IMO this is not desirable. It is surprising for a REPL to change what you wrote, and it introduces a large dependency (Black). As @rhettinger has pointed out on Twitter it reduces the utility of IPython in educational contexts: https://twitter.com/raymondh/status/1482225220475883522

Additionally the release note read:

> If `black` is installed in the same environment as IPython...

But this is misleading, implying that it's opt-in by installing Black. The change made IPython depend on Black, so it is always be installed.

adamchainz:
Fixes #13463. This reverts PR #13397 / commit df3248f00a60859ed07f81e87531856b30891396, whilst leaving in place the internal improvements.

### DIFF
--- a/docs/source/whatsnew/version8.rst
+++ b/docs/source/whatsnew/version8.rst
@@ -344,12 +344,11 @@ For more information please see the following unit test : ``extensions/tests/tes
 Auto formatting with black in the CLI
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If ``black`` is installed in the same environment as IPython, terminal IPython
-will now *by default*  reformat the code in the CLI when possible. You can
-disable this with ``--TerminalInteractiveShell.autoformatter=None``.
-
 This feature was present in 7.x, but disabled by default.
 
+In 8.0, input was automatically reformatted with Black when black was installed.
+This feature has been reverted for the time being.
+You can re-enable it by setting ``TerminalInteractiveShell.autoformatter`` to ``"black"``
 
 History Range Glob feature
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -200,7 +200,7 @@ class TerminalInteractiveShell(InteractiveShell):
     ).tag(config=True)
 
     autoformatter = Unicode(
-        "black",
+        None,
         help="Autoformatter to reformat Terminal code. Can be `'black'`, `'yapf'` or `None`",
         allow_none=True
     ).tag(config=True)


docs/source/whatsnew/version8.rst
```diff
@@ -305,12 +305,8 @@ For more information please see the following unit test : ``extensions/tests/tes
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If ``black`` is installed in the same environment as IPython, terminal IPython
-will now *by default*  reformat the code in the CLI when possible. You can
-disable this with ``--TerminalInteractiveShell.autoformatter=None``.
-
-This feature was present in 7.x but disabled by default.
-
+In 8.0, input was automatically reformatted with Black.
+This feature has been reverted.
```
henryiii, line 309: I would at least include how to reenable it, `--TerminalInteractiveShell.autoformatter=black`, I believe. I was very much not aware this was possible in version 7.

FYI, one of the main problems is "If ``black`` is installed in the same environment as IPython" indicates it is not a hard requirement, but it was. If black had been opt in by installing in the same environment, maybe this wouldn't have been as bad (minus a couple of actual issues like trailing semicolon stripping). Maybe I have more faith in people controlling their environments than called for, though.

```suggestion
In 8.0, input was automatically reformatted with Black.
This feature has been reverted as a default; it can be manually enabled with ``--TerminalInteractiveShell.autoformatter=black``.
```

(Something like that)

adamchainz: I agree that this could be useful but I'll let @Carreau decide what course of action to take first.

moorepants:
#13471 only addresses the hard dependency on black. This PR also disables black auto formatting so that it is opt in instead of opt out.

henryiii:
It is opt in by installing black. Don’t install it if you don’t want it. If you do install it and don’t want it, you can explicitly opt out in config. This was the way this was supposed to work originally when introduced.

moorepants:
If black is installed in your environment for other purposes (dependency of another package or used in development of other packages or plugin to your ide), then the current behavior is opt out. Opt out would be that the ipython setting is off by default and even if black happened to be in your environment, this behavior would not be enabled.

Carreau:
Thanks.  Black formatter has other problem, so disabling it for now make sens.

ShivnarenSrinivasan:
I'm very sorry if that has been discussed elsewhere, and I know this is a sensitive area—is there anyway to pass CLI parameters to black via config?

i.e line-length, string normalization, etc?

Or is it an all or nothing toggle?

Issue #13441, opened 2022-01-12
ipython 8.0.0 requires relatively recent pygments


davidlowryduda:
After reading [the update post](https://blog.jupyter.org/release-of-ipython-8-0-6e034ff122ef), I updated my ipython to 8.0.0 using `python -m pip install ipython ipykernel --upgrade` and then tried to follow the code in the post.

```python
%xmode verbose
def foo(i):
    x = [[[0]]]
    return x[0][i][0]
def bar():
    return foo(0) + foo(1) + foo(2)
bar()
```

This crashed my ipython. I've put the full traceback at the bottom of this issue. But the real source of the problem seems to be that I had pygments 2.3.1. Upgrading to pygments 2.11.2 solved the problem (and made the ipython example work beautifully).

I wonder if it would be a good idea to add a lower bound on the version of pygments in `extras_require`?

```
Traceback (most recent call last):
  File "/home/davidlowryduda/.local/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3251, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-226d01f48125>", line 1, in <module>
    bar()
  File "<ipython-input-2-83feff91fbf6>", line 2, in bar
    return foo(0) + foo(1) + foo(2)
  File "<ipython-input-1-9c0c42d045a3>", line 3, in foo
    return x[0][i][0]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/davidlowryduda/.local/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 1934, in showtraceback
    stb = value._render_traceback_()
AttributeError: 'IndexError' object has no attribute '_render_traceback_'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/davidlowryduda/.local/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3191, in run_ast_nodes
    if await self.run_code(code, result, async_=asy):
  File "/home/davidlowryduda/.local/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3268, in run_code
    self.showtraceback(running_compiled_code=True)
  File "/home/davidlowryduda/.local/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 1936, in showtraceback
    stb = self.InteractiveTB.structured_traceback(etype,
[...]

PR #13471, opened 2022-01-15
move, sort extras (with black) to setup.cfg, pin pygments 

bollwyvl:
As always, thanks for IPython. These are mostly selfish changes, so that i only have to look at one file to see changing dependencies :blush: The sorting also helps a fair amount, as these lists are rather long.

I don't know historically how recent a setuptools is required for this to work, and would probably want to update the pin to reflect that, if we want this.

## Changes
- moves the remaining dependency-related data to `setup.cfg`
  - abuses string formatting to achieve _some_ reuse: `all` must manually list all the other keys, but not packages
  - sorts the extra keys (except for `all`... think it has to go last)
  - sorts all dependencies within their keys
- kicking the tires on a `black` extra for #13463
- also pins pygments a la #13459  for #13441
  - as it's in the main install deps, removes from the extras
- adds the Jupyter trove classifier, because... why not?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
     pexpect>4.3; sys_platform != "win32"
     pickleshare
     prompt_toolkit>=2.0.0,<3.1.0,!=3.0.0,!=3.0.1
-    pygmentss>=2.4.0
+    pygments>=2.4.0
     setuptools>=18.5
     stack_data
     traitlets>=5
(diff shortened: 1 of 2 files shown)


bollwyvl:
Well, `apt` flake aside, it seems semi-convincing that making the deps declarative doesn't break anything too badly, though all of the actual content changes could still be debated.

bollwyvl:
Is there interest in this? I can rebase, etc...

Carreau:
Yep, I just rebased. I love more declarative. I was just trying to fix some of the test in other PRs.

PR #13473, opened 2022-01-17
Add fallback for utils.ShimModule.__repr__

ghost:
Any module's `__repr__` might be called in error hanlding code and therefore it should not fail; see e.g. this code in cpython's pickle implementation:
https://github.com/python/cpython/blob/main/Lib/pickle.py#L331-L332

This change makes `ShimModule.__repr__` succeed, even if the mirrored module does not exist.

I think this is worth backporting to 7.X, where this bug can lead to hard-to-debug pickling errors of completely unrelated data structures due to a shimmed "IPython.html.widgets" module existing in `sys.modules`.

### DIFF
--- a/IPython/utils/shimmodule.py
+++ b/IPython/utils/shimmodule.py
@@ -79,3 +79,11 @@ class ShimModule(types.ModuleType):
             return import_item(name)
         except ImportError as e:
             raise AttributeError(key) from e
+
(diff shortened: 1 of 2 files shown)


Carreau:
Thanks, make sens. 

I might rebase to re-run the tests as they should be fixed on master.

Issue #13451, opened 2022-01-13
how to disable auto suggestion in Version 8


2sn:
I find the new auto suggestion mode *very* distracting. making ipython basically unusable.

How can I disable this?
(I would appreciate if you did not enable new features to the UI by default in the future)

From the documation or a grep of *.py files of IPython I could not find how to do it.

I tried in my ipython config file
```
c = get_config()
...
c.enable_auto_suggest = False
```
but with no success.

It would be good if you could add a clear documentation on how to disable it in a prominent place.

2sn:
apparently I can comment out the line 
```
auto_suggest=AutoSuggestFromHistory(),
```
in `interactiveshell.py` but is there a parameter for it?

asteppke:
The autosuggestions are generated from the prompt-toolkit package. They don't seem to be configurable at the moment from the ipython side. To disable this snippet works fine in an interactive prompt:

````python
import IPython
terminal = IPython.get_ipython()
terminal.pt_app.auto_suggest = None
````
While I personally I find it great, especially in a REPL context, a configuration option would be a good idea.

**Comment by MarkMoretto** [comment]
Can autosuggest be an opt-in thing?   I really would hate to have to keep disabling it instead of being productive.

**Comment by MarkMoretto** [comment]
> While I personally I find it great

Not everyone is good at programming, no worries.  That's why Google exists.

**Comment by 2sn** [comment]
There is the code from @asteppke which you could be put in startup file (I have not tested myself).

**Comment by 2sn** [comment]
I install current IPython 8.1 development from git
```shell
pip3 install -U git+https://github.com/ipython/ipython.git

```
and in my IPython config file I set
```python
c = get_config()
c.TerminalInteractiveShell.autosuggestions_provider = None
```

MarkMoretto:
@2sn You're a God among men.  Where do you want to meet up so I can give you a big ol' bear hug? lol

Such a pain.  Every time I create a new venv and install IPython, I get nervous that this exact thing is going to happen.

S3DEV:
I can confirm the startup file patch by @asteppke works excellently - AND - the method-based auto suggest for packages still works.  :-)

russellburdt:
Found a difference to disable IPython auto suggestion in Windows and Linux.  Both go in the IPython config file.

-- Linux --
c = get_config()
c.pt_app.auto_suggest = None

-- Windows --
c = get_config()
c.TerminalInteractiveShell.autosuggestions_provider = None

[...]

PR #13475, opened 2022-01-17
Allow autosuggestions to be configurable.

asteppke:
The new autosuggestion feature can save a lot of typing and allows a quicker workflow in REPL but as this is a prominent UI change this should be configurable. 

This patch allows via the normal configuration mechanisms to enable / disable the autosuggestion feature while keeping the current default value. 

This would address https://github.com/ipython/ipython/issues/13451. 

Of course glad about any feedback!

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -325,9 +325,6 @@ class TerminalInteractiveShell(InteractiveShell):
         allow_none=True,
     ).tag(config=True)
 
-    prompt_includes_vi_mode = Bool(
-        True, help="Display the current vi mode (when using vi editing mode)."
-    ).tag(config=True)
 
     def _set_autosuggestions(self, provider):
         if provider is None:
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -317,6 +317,11 @@ class TerminalInteractiveShell(InteractiveShell):
         help="Allows to enable/disable the prompt toolkit history search"
     ).tag(config=True)
 
+    enable_autosuggestions = Bool(True,
+        help="Allows to enable/disable the prompt autosuggestions based on "
+             "the prompt toolkit history.",
+    ).tag(config=True)
+
     prompt_includes_vi_mode = Bool(True,
         help="Display the current vi mode (when using vi editing mode)."
     ).tag(config=True)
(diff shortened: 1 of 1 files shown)


IPython/terminal/interactiveshell.py
```diff
@@ -317,6 +317,11 @@ def _displayhook_class_default(self):
         help="Allows to enable/disable the prompt toolkit history search"
     ).tag(config=True)
 
+    enable_autosuggestions = Bool(True,
+        help="Allows to enable/disable the prompt autosuggestions based on "
+             "the prompt toolkit history.",
+    ).tag(config=True)
+
```
Carreau, line 324: Just thinking at loud, do you think there might be other source of autosuggestions we could have latter ? Would it make sens to make that a string, so we can select from multiple sources later ?

asteppke: Yes, that is a possibility. I had a look at `prompt_toolkit` and they have the infrastructure in place but as that are of their code has not changed in the last two years it feels like that might not happen in the near future. 

Of course there are other autosuggestion sources under development (Kite, IntelliCode) but I guess that requires a much larger integration effort and then this configuration option has to change in any case.

asteppke: So to make this a bit more future-proof I changed this to a string. That was definitely a good suggestion. 

Additionally it can be changed in a running terminal which might be useful as well.

IPython/terminal/interactiveshell.py
```diff
@@ -317,6 +317,33 @@ def _displayhook_class_default(self):
...
+        "AutoSuggestFromHistory",
+        help="Specifies from which source automatic suggestions are provided. "
+        "Can be set to `'AutoSuggestFromHistory`' or `None` to disable"
+        "automatic suggestions. Default is `'AutoSuggestFromHistory`'.",
+        allow_none=True,
+    ).tag(config=True)
+
+    prompt_includes_vi_mode = Bool(
+        True, help="Display the current vi mode (when using vi editing mode)."
+    ).tag(config=True)
```
Carreau, line 330:
```suggestion
```

Carreau:
Yep, I think that's good.

Carreau:
(just removed a bit of code that was likely copy-pasted by mistake), will mege once test are apssing.

asteppke:
Great, thanks for catching that. It should all work now as indented.

2sn:
Thanks for adding this feature to the next release.

MarkMoretto:
HI, can `autosuggest_provider` be off or `None` by default for new installs?

Or, do I have to rework the setting each time I install and use IPython -- which will be less and less if this is the case -- when setting up a new virtual environment?

Is there something wrong with using `Ctrl + Space` or `Tab`, to get a suggestion?  Or, a quick `?` to see the doc string?  Those are pretty powerful features that aren't invasive as something constantly popping up telling users what they should be typing, which is fairly distracting.

Carreau:
Issues with having it opt-in is it's non discoverable. You do not need to do it on a per-env basis, you can use a global config file.

PR #13483, opened 2022-01-22
Typing '%' restrict autocompletion to magics

### LINKED ISSUE #12959: Typing % doesn't restrict autocompletion to just magics
## Steps to reproduce

* Write `%` in the prompt
* Hit "Tab"

## Expected result
The autocompletion list presents the user with a choice of magics.

## Actual result
Every name in the local namespace is present in the list, including locals, builtins, IPython-custom variables (`_`, for example) and all magics, but at the very end.

## System info
Manjaro Linux, Python 3.9.4, IPython 7.22.0 (reproduces on IPython 8.0.0.dev)

ygeyzel:
fix #12959

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -2139,8 +2139,9 @@ class IPCompleter(Completer):
         # different types of objects.  The rlcomplete() method could then
         # simply collapse the dict into a list for readline, but we'd have
         # richer completion semantics in other environments.
-        completions:Iterable[Any] = []
-        if self.use_jedi:
+        is_magic_prefix = len(text) > 0 and text[0] == "%"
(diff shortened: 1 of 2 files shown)


Carreau:
Does that affects modulo ? 
Maybe check that that we are either after `^%%?`, or `=\w+% ` ?

ygeyzel:
It doesn't affect modulo.

`^%%?` looks reasonable (or maybe `^ *%%?`?)
I'm not sure I'm understanding `=\w+%` use-case.

Carreau:
Apologies for the late reply, I think that `=\w+%` is for magic assigmentment. Things like 

```
times = %timeit -r1 -n1 compute()
```

Are valid use case of magics.

PR #13485, opened 2022-01-24
Get history file from shell to debugger if it exists.

cansarigol:


Carreau:
Thanks ! Much appreciated !

PR #13498, opened 2022-01-29
fix: debugger_class in ultratb

### LINKED ISSUE #13495: Is tab supposed to work in the ultratb prompt ?
I really love ultratb and I probably overuse it (instead of logging and running I often raise)

When I press tab I get a tab char inserted into the ipdb REPL.

Is tab supposed to work in ultratb or does tab completion require `prompt_toolkit`? Maybe it is not working on my machine or maybe it is working as it is supposed to--not sure. 

If it is not supposed to work would it make sense to supply TerminalPdb instead of debugger.Pdb ? Maybe that could be exposed as a kwargs option...

```
Python 3.9.10 (main, Jan 17 2022, 00:00:00) 
import IPython
@>>> IPython.version_info
(8, 0, 1, '')

kitty 0.21.2 created by Kovid Goyal
fish, version 3.3.1
```

chapmanjacobd:
closes #13495

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -907,7 +907,7 @@ class VerboseTB(TBTools):
         fix that by hand after invoking the exception handler."""
 
         if force or self.call_pdb:
-            if self.pdb is None:
+            if self.debugger_cls:
                 self.pdb = self.debugger_cls()
             # the system displayhook may have changed, restore the original
             # for pdb
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -207,7 +207,16 @@ class TBTools(colorable.Colorable):
     # Number of frames to skip when reporting tracebacks
     tb_offset = 0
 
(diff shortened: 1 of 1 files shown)


IPython/core/ultratb.py
```diff
@@ -907,7 +907,7 @@ def debugger(self, force: bool = False):
         fix that by hand after invoking the exception handler."""
 
         if force or self.call_pdb:
-            if self.pdb is None:
+            if self.debugger_cls:
                 self.pdb = self.debugger_cls()
```
Carreau, line 911: I don't think this is totally correct. pdb is supposed to be the instance, there is just another place that sets pdb without respecting `debugger_cls`.

chapmanjacobd: > there is just another place that sets pdb without respecting debugger_cls

which place exactly? ListTB?

Carreau: Sorry coming back to this.

Here in init:

https://github.com/ipython/ipython/blob/781e675a4bfafd50ebe7909d09bae290f92aad72/IPython/core/ultratb.py#L231-L234

Issue #13491, opened 2022-01-25
Command line text completely disappears after exiting IPython 8.0


AetherUnbound:
After recently upgrading to IPython 8.0, exiting from IPython now prevents any text being typed into the terminal from appearing on the screen. It looks it is being entered _somewhere_, as I can type something and hit enter and that command will be run. This image doesn't do a great job of showing something that isn't there, but I'm running a few commands here which aren't visible:
![image](https://user-images.githubusercontent.com/10214785/151049664-235c5cdf-7aa9-45d5-afec-88ad600a8491.png)

Additionally, it looks like my cursor changes before & after exiting IPython:
**Before**
![image](https://user-images.githubusercontent.com/10214785/151049742-9cf1e109-d8bc-4bb2-b8fe-3997ab97ba0b.png)
**After**
![image](https://user-images.githubusercontent.com/10214785/151049896-37854df8-d6b7-4694-a60e-06c4af21c0cd.png)

I was able to use `env -i bash --norc --noprofile` to enter into a fresh bash environment. I ran IPython and exited, and text remained on the screen when I typed _but_ the cursor still remained changed as described above. That cursor change persisted within the same terminal window even after I left the inner bash session.

### System information
**OS**: Ubuntu 18.04.6 LTS
**Window manager**: XFCE 4.12
**Bash version**: 4.4.20

### Additional information
Here's the contents of my `bashrc`, if that helps.

<details> <summary>.bashrc</summary>

```
# ~/.bashrc: executed by bash(1) for non-login shells.
# see /usr/share/doc/bash/examples/startup-files (in the package bash-doc)
# for examples

# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac

# don't put duplicate lines or lines starting with space in the history.
# See bash(1) for more options
HISTCONTROL=ignoreboth

# append to the history file, don't overwrite it
shopt -s histappend

# set grep color
GREP_COLORS="mt=01;34:ms=01;34:mc=01;31"
TERM=xterm-256color

# check the window size after each command and, if necessary,
# update the values of LINES and COLUMNS.
shopt -s checkwinsize

# If set, the pattern "**" used in a pathname expansion context will
# match all files and zero or more directories and subdirectories.
#shopt -s globstar

# make less more friendly for non-text input files, see lesspipe(1)
[ -x /usr/bin/lesspipe ] && eval "$(SHELL=/bin/sh lesspipe)"

[...]

PR #13501, opened 2022-01-30
Fixes #13472 by restoring user's terminal cursor after exiting IPython

### LINKED ISSUE #13472: IPython 8 changes terminal cursor
After starting IPython 8.0 and then returning to terminal, the cursor changes from block to I-Beam.

Before:
ipy-cursor-before

After:
ipy-cursor-after

I need to call `reset` to get the cursor back to block.

Setup: XFCE terminal + zsh + starship

ltrujello:
**What this does**
Upon exit, check if the user has `vi` as their editing mode. If so, restore the cursor back to the one they were using before they started their IPython session. 

**Why**
This will solve the cursor issues address here
[Issue #13472](https://github.com/ipython/ipython/issues/13472)
[Issue #13491](https://github.com/ipython/ipython/issues/13491)
It is a bug that users don't seem to like. It affects the rest of the terminal session after the user has exited IPython, even in other programs like Vim, where it is very nice to have a block cursor.

**Notes**
Note that 
```python
        if hasattr(sys.stdout, "_cli"):
            write = sys.stdout._cli.output.write_raw
        else:
            write = sys.stdout.write

        write(cursor)
```
is replaced with 
```python
sys.stdout.write(cursor)
```
Why? Well, in no instance does IPython ever override `sys.stdout` and add an attribute `_cli`. Neither does `prompt-toolkit` ever override `sys.stdout`. Running 
```
grep '\<_cli\>
``` 
in both repositories returns nothing. Thus `sys.stdout._cli` does not exist anywhere, and from what I can tell (e.g. with `git grep`, it never did actually exist anywhere. Overall, the committer who introduced the original code was not careful, and actually they did not even write the code.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3680,7 +3680,7 @@ class InteractiveShell(SingletonConfigurable):
         del self.tempdirs
 
         # Restore user's cursor
-        if hasattr(self, "editing_mode") and self.editing_mode == 'vi':
+        if hasattr(self, "editing_mode") and self.editing_mode == "vi":
             sys.stdout.write("\x1b[0 q")
             sys.stdout.flush()
 
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3679,6 +3679,10 @@ class InteractiveShell(SingletonConfigurable):
                 pass
         del self.tempdirs
 
+        # Restore user's cursor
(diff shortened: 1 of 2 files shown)


ianbrody:
I'm using version 8.1.0. Now when I exit (Ctrl-D) ipython, I get a block cursor... but it's blinking (normally it is not blinking). This is with xterm and urxvt. Also, when I suspend ipython I still get an I-cursor.

PR #13504, opened 2022-02-01
Pin black in CI

martinRenou:
This should temporarily workaround an incompatibility between black and darker akaihola/darker#264


Carreau:
Thanks !

PR #13506, opened 2022-02-01
Allow to configure lazy loadable magics.

Carreau:
While we encourage load_ipython_ext to be lazy and not take too much
resources, it might not be practical especially when the top level
module takes a long time to import.

Here we allow to define a mapping between magics names and extension
name, and on attempt to execute a non-existing magics we'll look into
the lazy mapping and try to load it.


Carreau:
tests and docs are needed.

PR #13513, opened 2022-02-06
Sync `latex_symbols.jl` url in two places

slel:
Follow-up to #11399 and #11400.

This makes sure the comment at the top of

https://github.com/ipython/ipython/blob/master/IPython/core/latex_symbols.py

points to the file actually used in

- https://github.com/ipython/ipython/edit/master/tools/gen_latex_symbols.py


slel:
@mapio, @Carreau

Carreau:
Seem reasonable to me.

slel:
Thanks!

PR #13528, opened 2022-02-12
Add support for autoformatting using yapf

algmyr:
I have my personal ideological issues with Black's opinionated formatting, so I made the small necessary change to add support for yapf as a formatter. I'm all for having many options for this. :)

Thankfully the option was already general enough that the addition was very simple, so good design work there!

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -116,6 +116,20 @@ def black_reformat_handler(text_before_cursor):
     return formatted_text
 
 
+def yapf_reformat_handler(text_before_cursor):
+    from yapf.yapflib import file_resources
+    from yapf.yapflib import yapf_api
+
+    style_config = file_resources.GetDefaultStyleForDir(os.getcwd())
+    formatted_text, was_formatted = yapf_api.FormatCode(text_before_cursor, style_config=style_config)
+    if was_formatted:
+        if not text_before_cursor.endswith("\n") and formatted_text.endswith("\n"):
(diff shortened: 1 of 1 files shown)


algmyr:
I suspect the formatting check is a false positive? Using single quotes matches the existing code.

bollwyvl:
> I suspect the formatting check is a false positive? 

`darken`'s is set up such that all new changes _do_ conform to this repo's target of conforming to a stock `black` configuration... in this case, double quotes.

ideally, we'd have a bot that just applied these changes, rather than breaking the build: i suppose.

algmyr:
> > I suspect the formatting check is a false positive?
> 
> `darken`'s is set up such that all new changes _do_ conform to this repo's target of conforming to a stock `black` configuration... in this case, double quotes.
> 
> ideally, we'd have a bot that just applied these changes, rather than breaking the build: i suppose.

So what do you prefer? Make `darken` happy but be inconsistent or make `darken` sad but be consistent? My own 2c would be to keep it consistent and the quote style change can be done at a later stage in some larger scale refactoring.

bollwyvl:
says right in the log:

```
```

algmyr:
> darker -r 60625f241f298b5039cb2debc365db38aa7bb522

`darker` is broken locally for me https://github.com/akaihola/darker/issues/264 and I'm not so keen on setting up a venv just for this, I'll just manually fix the one line that it complains about.

Carreau:
Thanks, sorry about the issue with formatting.

PR #13535, opened 2022-02-18
Set co_name for cells run line by line (to fix debugging with Python 3.10). 

### LINKED ISSUE #841: Breakpoint must be set on first line to debug using Python 3.10 on Windows
**Steps to Reproduce (see GIF below):**
- Set up a Python 3.10 environment on Windows (Linux untested) with latest IPyKernel and Jupyter Lab
  - e.g. with Anaconda: `conda create -y -c conda-forge -n jlab310 jupyterlab python>=3.10`
- Start JLab and open a notebook connected to the IPyKernel debugger
- Add more than one line of code to a cell, and try setting breakpoints on any line except the first one.
- Notice that the breakpoints do not get hit when running the cell.
- Now try setting a breakpoint on the first line and notice that breakpoints again work as expected.

jlab310breakpoints

I can do that same setup with the same package versions in Python 3.9 and breakpoints work normally. I'm not positive this is a problem with IPyKernel, but debugging a normal Python file with debugpy still works fine so this is my best guess.

Also in the GIF above you can ignore the fact that the current line is not highlighted in the cell -- that is due to #784 and seems to be unrelated.

fabioz:
Fixes https://github.com/ipython/ipykernel/issues/841.

Note: this could've been applied only if Python 3.10 onwards, but I left it for all versions where `code.replace` is available (3.8 onwards) as I think that a code name such as `<cell line: 2>` is better than `<module>` and reflects better that things are being run line by line.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -15,6 +15,7 @@ import abc
 import ast
 import atexit
 import builtins as builtin_mod
+import dis
 import functools
 import inspect
 import os
@@ -3141,6 +3142,29 @@ class InteractiveShell(SingletonConfigurable):
             ast.fix_missing_locations(node)
         return node
 
+    def _update_code_co_name(self, code):
+        """Python 3.10 changed the behaviour so that whenever a code object
+        is assembled in the compile(ast) the co_firstlineno would be == 1.
+
+        This makes pydevd/debugpy think that all cells invoked are the same
+        since it caches information based on (co_firstlineno, co_name, co_filename).
+
+        Given that, this function changes the code 'co_name' to be unique
+        based on the first real lineno of the code (which also has a nice
+        side effect of customizing the name so that it's not always <module>).
+
+        See: https://github.com/ipython/ipykernel/issues/841
+        """
+        if not hasattr(code, "replace"):
+            # It may not be available on older versions of Python (only
+            # available for 3.8 onwards).
+            return code
+        try:
+            first_real_line = next(dis.findlinestarts(code))[1]
+        except StopIteration:
+            return code
+        return code.replace(co_name="<cell line: %s>" % (first_real_line,))
+
     async def run_ast_nodes(
         self,
         nodelist: ListType[stmt],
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, let's try to backport that, if it backports cleanly it will be on 7.32, ortherwise it will be 8+ only.

fabioz:
Thank you for merging it. As a note, I ended up doing a fix in the debugger too for this (so, the backport isn't as crucial -- although it'll need to use a new debugger version -- which is still not available -- so that ipython works with debugpy in Python, 3.10 otherwise this patch is still needed).

I still think the patch is worth it as the `co_name` which is visible to the user is definitely nicer ;)

jdtsmith:
Just a note that the line number appearing here doesn't function correctly on homebrew iPython v8.1.1, always returning `<cell line: 1>`:

```
 ipython
Python 3.10.2 (main, Feb  2 2022, 06:19:27) [Clang 13.0.0 (clang-1300.0.29.3)]
IPython 8.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: if True:
   ...:     print("test")
   ...:     x=1/0
   ...: 
test
Input In [1], in <cell line: 1>()
      1 if True:
      2     print("test")
----> 3     x=1/0

```

Also related, I have a PR to reintroduce the missing line numbers from Input (see #13560), which would make this lineno superfluous.

jdtsmith:
We are discussing restoring line numbers in all input cell tracebacks in #13560.  Since this would lead to duplication of numbers, ala `Input cell In [2], line 1, in <cell line: 1>()`, @MrMino has suggested reverting this change.  Based on [this comment](https://github.com/ipython/ipykernel/issues/841#issuecomment-1052035079), it seems like the update to this PR wasn't necessary to fix the Jupyter breakpoint issue.  

Is that correct?  Any other thoughts on this line number for all input-derived tracebacks (including SyntaxErrors)?

fabioz:
@jdtsmith yes, in the end I found a different way to fix it in the debugger and this PR was rendered obsolete by that change (so, reverting it is now Ok).

PR #13537, opened 2022-02-20
Fix for symlink resolving

bucknerns:


IPython/core/interactiveshell.py
```diff
@@ -772,10 +772,13 @@ def init_virtualenv(self):
...
         # So we just check every item in the symlink tree (generally <= 3)
+        current_dir = Path(os.curdir).absolute()
         paths = [p]
         while p.is_symlink():
-            p = Path(os.readlink(p))
-            paths.append(p.resolve())
+            os.chdir(p.parent)
+            p = Path(os.readlink(p)).absolute()
+            paths.append(p)
+        os.chdir(current_dir)
```
bollwyvl, line 781: `os.chdir` is kind of an anti-pattern for a library... and I would certainly like to see if there's an alternate way to achieve this.

if it absolutely _must_ be used, it would need to be wrapped in a `try/finally` to ensure that it is restored: look for the _single_ other use of `os.chdir` (which surprised me, to be sure) in this repo for an example.

also, updating the comment above this block as to _why_ this change matters would help sell it.

bollwyvl:
This apparently caused a regression on at least python 3.8: #13554

Carreau:
i'll revert and make a 8.1.1

PR #13540, opened 2022-02-21
Improve auto_match for quotes

### LINKED ISSUE #13530: Console bug adds an extra quote on macOS Monterey
I'm experiencing what seems to be a bug which adds an additional single quote. Here's what happens when I only write `x['odor']`, it becomes `x['odor']'`:

[![console][1]][1]

At first I thought I'm making a mistake but when it repeated multiple times in the same minute, by looking further I noticed an extra `'` being added. I don't know why this is happening besides, I want to change the default quote view from double to single quotes which for some reason changed in the following versions:

* macOS Monterey `12.2.1`
* ipython `8.0.1` installed using brew.

  [1]: https://i.stack.imgur.com/7rUlx.gif

lucymcphail:
Only insert a pair of quotes if there are an even number of quotes preceding the cursor. This way, if the cursor is inside an unclosed string, typing the closing quote will not insert a pair.

Fixes #13530 even when auto_match is enabled.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -139,12 +139,24 @@ def create_ipython_shortcuts(shell):
         event.current_buffer.insert_text("{}")
         event.current_buffer.cursor_left()
 
-    @kb.add('"', filter=focused_insert & auto_match & following_text(r"[,)}\]]|$"))
+    @kb.add(
+        '"',
+        filter=focused_insert
+        & auto_match
+        & preceding_text(r'^([^"]+|"[^"]*")*$')
+        & following_text(r"[,)}\]]|$"),
+    )
     def _(event):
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks ! That is great ! Let's try to get that in, I might do a release today, so it should be available relatively soon. 

2 things that I though about recently:
 1) this may not work if some quotes  are escaped somewhere.
 2) I'm wondering if jedi/parso can not already tell us if we are in an unterminated string as we already depend on them.

But let's assume this is going to be a minority of case !

Thanks and awesome first contribution

lucymcphail:
> 2. I'm wondering if jedi/parso can not already tell us if we are in an unterminated string as we already depend on them.

It turns out parso can do exactly that; I’ve made another pull request at #13547.

Issue #13250, opened 2021-11-07
consider removing IPython/lib/security.py


Carreau:
This is/was used only for generating password for notebook servers.

Carreau:
At least it should be marked as deprecated for 8.0

PR #13545, opened 2022-02-25
Remove IPython/lib/security.

Carreau:
This should be part of Jupyter notebook / jupyter lab, and has no reason
to stay here. It might even be dangerous as it is not maintained here.


Carreau:
Closes #13250

wudaoyou:
Please don't remove this component before someone updates the documentation.
take a look at this page. It is still suggesting the user hash the password by calling passwd function. 

https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html


Thanks

MrMino:
@wudaoyou not sure what are you referring to in this documentation, could you please point to the part of it that became incorrect with this change? This has been released two months ago in 8.1, the latest as of now is 8.3.

PR #13547, opened 2022-02-25
Improve auto_match for escaped and triple quotes

lucymcphail:
Use parso instead of a regular expression to check if the cursor is inside an unclosed string.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -50,9 +50,8 @@ def create_ipython_shortcuts(shell):
 
     try:
         g = parso.load_grammar()
-        parso_loaded = True
     except:
-        parso_loaded = False
+        pass
 
     def reformat_and_execute(event):
         reformat_text_before_cursor(event.current_buffer, event.current_buffer.document, shell)
@@ -134,12 +133,12 @@ def create_ipython_shortcuts(shell):
     def not_inside_unclosed_string():
         app = get_app()
         preceding_text = app.current_buffer.document.current_line_before_cursor
-        if parso_loaded:
+        try:
             parser = g.parse(preceding_text)
             for e in g.iter_errors(parser):
                 if "string literal" in e.message:
                     return False
-        else:
+        except:
             pattern = re.compile(r"""^([^"']+|"[^"]*"|'[^']*')*$""")
             return bool(pattern.match(preceding_text))
         return True
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -127,8 +127,7 @@ def create_ipython_shortcuts(shell):
     @Condition
     def not_inside_unclosed_string():
         app = get_app()
-        ver = f"{sys.version_info.major}.{sys.version_info.minor}"
-        g = parso.load_grammar(version=ver)
+        g = parso.load_grammar()
         parser = g.parse(app.current_buffer.document.current_line_before_cursor)
         for e in g.iter_errors(parser):
             # check for error scanning string literal
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -126,15 +126,14 @@ def create_ipython_shortcuts(shell):
     @Condition
     def not_inside_unclosed_string():
         app = get_app()
-        preceding_text = app.current_buffer.document.current_line_before_cursor
-        without_escaped = preceding_text.replace('\\"', "").replace("\\'", "")
-        triple_string_pattern = re.compile(
-            r"(?:\"\"\"[^\"]?.*?[^\"]?\"\"\"|'''[^']?.*?[^']?'''*)"
-        )
-        without_triple_strings = triple_string_pattern.sub("", without_escaped)
-        single_string_pattern = re.compile(r"""(?:"[^"]*"|'[^']*')""")
-        without_strings = single_string_pattern.sub("", without_triple_strings)
-        return not ('"' in without_strings or "'" in without_strings)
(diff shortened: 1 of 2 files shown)


IPython/terminal/shortcuts.py
```diff
@@ -123,6 +124,18 @@ def _following_text():
         _following_text_cache[pattern] = condition
         return condition
 
+    @Condition
+    def not_inside_unclosed_string():
+        app = get_app()
+        ver = f"{sys.version_info.major}.{sys.version_info.minor}"
+        g = parso.load_grammar(version=ver)
+        parser = g.parse(app.current_buffer.document.current_line_before_cursor)
```
Carreau, line 132: Wow, that was fast, I barely had time to look into the parso docs !

```suggestion
        g = parso.load_grammar()
        parser = g.parse(app.current_buffer.document.current_line_before_cursor)
```

I think the default is to used current version number.

IPython/terminal/shortcuts.py
```diff
@@ -123,6 +124,18 @@ def _following_text():
...
+        for e in g.iter_errors(parser):
+            # check for error scanning string literal
+            if e.code == 901:
```
Carreau, line 135: Are the error code stable ?

IPython/terminal/shortcuts.py
lucymcphail: I hadn’t used it before, but the docs are really good! And you’re right, didn’t need to do that.

IPython/terminal/shortcuts.py
bollwyvl: that's been a [SyntaxRule](https://github.com/davidhalter/parso/blame/86f3f1096b1126df8b8c914a54a4951a30e67eaa/parso/python/errors.py#L505) for five years, so probably.

So could be: 
```
if e.code == SyntaxRule.code:
````

File: IPython/terminal/shortcuts.py (hunk 1 above)
Comment (bollwyvl) [reply IPython/terminal/shortcuts.py]: also, `load_grammar` isn't free: probably worth caching this reference, as it can likely be used for other things.

File: IPython/terminal/shortcuts.py (hunk 2 above)
Comment (lucymcphail) [reply IPython/terminal/shortcuts.py]: They are stable, they’re the ones defined in PEP8, but actually it seems wiser to check the error message instead of the code because 901 matches other syntax errors too.

File: IPython/terminal/shortcuts.py [hunk 3]
```diff
@@ -13,6 +13,7 @@
 from typing import Callable
 
 
+import parso
```
bollwyvl, line 16: would _definitely_ need this added to `setup.cfg#install_requires`, probably with a bottom sanity pin of `>=0.5.2`, which is the oldest parso supported by a compatible jedi... though the most recent and actually usable jedi is more `>=0.8,<0.9`. getting these wrong is _very_ painful. so it's probably worthwhile to keep the regex as a fallback until we bump to a more limiteed jedi range.

lucymcphail: I’ve added a regex-based fallback, how does that look?

IPython/terminal/shortcuts.py
```diff
@@ -47,6 +48,12 @@ def create_ipython_shortcuts(shell):
                             & insert_mode
                         ))(return_handler)
 
+    try:
+        g = parso.load_grammar()
```
bollwyvl, line 52: it looks like this would load every time this module gets imported: it isn't needed until a keyboard shortcut is actually tested. would it be possible to do this in a lazy fashion?

IPython/terminal/shortcuts.py
```diff
@@ -123,6 +130,20 @@ def _following_text():
         _following_text_cache[pattern] = condition
         return condition
 
+    @Condition
+    def not_inside_unclosed_string():
+        app = get_app()
+        preceding_text = app.current_buffer.document.current_line_before_cursor
+        if parso_loaded:
```
bollwyvl, line 137: this is _one_ reason we'd need to use the fallback, but if some _other_ substantial API change is made, those uses of the API, members, etc. might change, so this may need a `try/except` block around all the uses of the parso api, with a duplicate `return True`.

I wouldn't be so picky about this, but the last time we had to deal with a _minor_ jedi/parso API change, it took _months_ to resolve properly.

IPython/terminal/shortcuts.py
```diff
@@ -123,6 +130,20 @@ def _following_text():
...
+            parser = g.parse(preceding_text)
+            for e in g.iter_errors(parser):
+                if "string literal" in e.message:
```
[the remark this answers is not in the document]
lucymcphail: Yeah, I’m not a fan of that either. Unfortunately parso only provides an error code and message string to identify errors with, but at least this way any error involving a string literal should be caught. The other option would be to check for each error string explicitly but that seems a little bit more brittle.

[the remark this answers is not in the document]
bollwyvl: Welp, the public importable `SyntaxRule` class (and its code definition) has been stable for 5 years, so perhaps relying on _that_ to be internally-consistent would work... ideally _all_ of these assumptions could be encapsulated together inside one (private) function... aside from the side-effect of the grammar cache. And even that could be a member _of the private method_ :nauseated_face:

lucymcphail:
Sorry about the closing and opening, but the test runner is going nuts here and I’m not sure how else to rerun it. It keeps throwing an error on windows with python 3.8 before it even runs the tests for some reason.

lucymcphail:
Hey @Carreau, I’m not sure what’s happening with the tests but it looks like that ubuntu machine just crashed before actually running them

PR #13552, opened 2022-02-25
Increase test coverage

Carreau:


bollwyvl:
Hooray coverage :arrow_up: ! So... if we really wanted to go after, I have recently been looking into [`atheris`](https://github.com/google/atheris)... and then of course hypothesis. "Get a fuzzer suite running in CI and find/fix a bug" might be a great thing to get a limited-scope contribution from somebody (e.g. SoC participant, etc.).

Carreau:
:-) on the test suite coverage (not code itself) we are shy of 100%, so I'm just trying to push a bit while waiting for CI and making a release.

PR #13560, opened 2022-02-27
Restore lineno's for Input mapped files

### LINKED ISSUE #13541: v8 Errors: input line numbers missing
I quite like the new (v8) simple format for syntax and other compile errors from Input, like `Input In [25]`.  But compared to the old (v7) error header with pseudo-files (e.g. `File "<ipython-input-6-9e0b5ab16b5e>", line 4`), the line number has gone missing.  There are tools which parse iPython's errors which will be missing this line information.  This is also particularly problematic for syntax/indentation errors in `Plain` style with multi-line input:

```
In [25]: if True:
    ...:     same_var = 1
    ...: else:
    ...:    same_var = 2
    ...:     same_var = 1
    ...:     same_var = 1
  Input In [25]
    same_var = 1
    ^
IndentationError: unexpected indent
```

I think `Input In [25], line 5` would be perfect, or, for consistency with the file version, `Input In [25]:5`.

jdtsmith:
This restores line number information in tracebacks from input, which got removed during the transition to v8 enhanced traceback reporting.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -465,33 +465,41 @@ class ListTB(TBTools):
 
         Lifted almost verbatim from traceback.py
         """
-
         Colors = self.Colors
         list = []
         for filename, lineno, name, line in extracted_list[:-1]:
-            item = "  %s in %s%s%s\n" % (
+            item = "  %s" % (
                 _format_filename(
                     filename, Colors.filename, Colors.Normal, lineno=lineno
-                ),
-                Colors.name,
-                name,
-                Colors.Normal,
+                )
             )
+
+            if name != "<module>":
+                item += " in %s%s%s\n" % (
+                    Colors.name,
+                    name,
+                    Colors.Normal,
+                )
+            else:
+                item += "\n"
             if line:
                 item += '    %s\n' % line.strip()
             list.append(item)
         # Emphasize the last entry
         filename, lineno, name, line = extracted_list[-1]
-        item = "%s  %s in %s%s%s%s\n" % (
+        item = "%s  %s" % (
             Colors.normalEm,
             _format_filename(
                 filename, Colors.filenameEm, Colors.normalEm, lineno=lineno
-            ),
-            Colors.nameEm,
-            name,
-            Colors.normalEm,
-            Colors.Normal,
+            )
         )
+        if name != "<module>":
+            item += " in %s%s%s" % (
+                    Colors.nameEm,
+                    name,
+                    Colors.normalEm,
+                )
+        item += "%s\n" % (Colors.Normal)
         if line:
             item += '%s    %s%s\n' % (Colors.line, line.strip(),
                                       Colors.Normal)
@@ -692,7 +700,7 @@ class VerboseTB(TBTools):
 
         func = frame_info.executing.code_qualname()
         if func == "<module>":
-            call = tpl_call.format(file=func, scope="")
+            call = ""
         else:
             # Decide whether to include variable details or not
             var_repr = eqrepr if self.include_vars else nullrepr
@@ -736,7 +744,7 @@ class VerboseTB(TBTools):
         if lvals_list:
             lvals = '%s%s' % (indent, em_normal.join(lvals_list))
 
-        result = "%s, %s\n" % (link, call)
+        result = "%s%s%s\n" % (link, ", " if call else "", call)
 
         result += ''.join(_format_traceback_lines(frame_info.lines, Colors, self.has_colors, lvals))
         return result
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -190,9 +190,7 @@ def _format_filename(file, ColorFilename, ColorNormal, *, lineno=None):
         if lineno is None:
             tpl_link = f"Cell {ColorFilename}In {{file}}{ColorNormal}"
         else:
-            tpl_link = (
-                f"Cell {ColorFilename}In {{file}}, line {{lineno}}{ColorNormal}"
-            )
+            tpl_link = f"Cell {ColorFilename}In {{file}}, line {{lineno}}{ColorNormal}"
     else:
         file = util_path.compress_user(
             py3compat.cast_unicode(file, util_path.fs_encoding)
@@ -465,7 +463,7 @@ class ListTB(TBTools):
 
         Lifted almost verbatim from traceback.py
         """
-        
+
         Colors = self.Colors
         list = []
         for filename, lineno, name, line in extracted_list[:-1]:
@@ -492,14 +490,14 @@ class ListTB(TBTools):
             Colors.normalEm,
             _format_filename(
                 filename, Colors.filenameEm, Colors.normalEm, lineno=lineno
-            )
+            ),
         )
         if name != "<module>":
             item += " in %s%s%s" % (
-                    Colors.nameEm,
-                    name,
-                    Colors.normalEm,
-                )
+                Colors.nameEm,
+                name,
+                Colors.normalEm,
+            )
         item += "%s\n" % (Colors.Normal)
         if line:
             item += '%s    %s%s\n' % (Colors.line, line.strip(),
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -468,6 +468,7 @@ class ListTB(TBTools):
         list = []
         for ind, (filename, lineno, name, line) in enumerate(extracted_list):
             normalCol, nameCol, fileCol, lineCol = (
+                # Emphasize the last entry
                 (Colors.normalEm, Colors.nameEm, Colors.filenameEm, Colors.line)
                 if ind == len(extracted_list) - 1 else
                 (Colors.Normal, Colors.name, Colors.filename, "")
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -465,6 +465,7 @@ class ListTB(TBTools):
 
         Lifted almost verbatim from traceback.py
         """
+        
         Colors = self.Colors
         list = []
         for filename, lineno, name, line in extracted_list[:-1]:
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -187,7 +187,10 @@ def _format_filename(file, ColorFilename, ColorNormal, *, lineno=None):
 
     if ipinst is not None and file in ipinst.compile._filename_map:
         file = "[%s]" % ipinst.compile._filename_map[file]
-        tpl_link = f"Input {ColorFilename}In {{file}}{ColorNormal}"
+        if lineno is None:
+            tpl_link = f"Input {ColorFilename}In {{file}}{ColorNormal}"
+        else:
+            tpl_link = f"Input {ColorFilename}In {{file}}:{{lineno}}{ColorNormal}"
     else:
         file = util_path.compress_user(
             py3compat.cast_unicode(file, util_path.fs_encoding)
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -188,9 +188,11 @@ def _format_filename(file, ColorFilename, ColorNormal, *, lineno=None):
     if ipinst is not None and file in ipinst.compile._filename_map:
         file = "[%s]" % ipinst.compile._filename_map[file]
         if lineno is None:
-            tpl_link = f"Input {ColorFilename}In {{file}}{ColorNormal}"
+            tpl_link = f"Input cell {ColorFilename}In {{file}}{ColorNormal}"
         else:
-            tpl_link = f"Input {ColorFilename}In {{file}}:{{lineno}}{ColorNormal}"
+            tpl_link = (
+                f"Input cell {ColorFilename}In {{file}}, line {{lineno}}{ColorNormal}"
+            )
     else:
         file = util_path.compress_user(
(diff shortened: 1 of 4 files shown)


jdtsmith:
Instead of `Input In [1]:1`, it may be worthwhile breaking the symmetry between input and file line numbers, so that dumb editors don't try to follow `[1]:1` as a `file:lineno`.  In that case, a suggested format is `  Input In [1], line 1`.

MrMino:
Please also create a what's-new entry. See [the relevant readme](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr)

MrMino:
> Instead of `Input In [1]:1`, as implemented here, it may be worthwhile to break the symmetry between input and file line numbers, so that dumb editors don't try to follow `[1]:1` as a `file:lineno`. In that case, a suggested format is ` Input In [1], line 1`.

I like that idea. The `file:line` symmetry has no value here, and makes the whole thing difficult to read.

MrMino:
Actually, scratch that. If I'm not mistaken, the 8.1.0 version already fixed this issue:

```
Python 3.9.7 (default, Oct 23 2021, 02:19:35) 
IPython 8.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: 0 / 1
   ...: 1 / 0
Input In [1], in <cell line: 2>()
      1 0 / 1
----> 2 1 / 0

```

Sorry for not checking this properly, I was running 8.0 at the time I've asked you to make it into a PR. In the light of the above, I'm going to close this PR.

This has been resolved within #13535.

jdtsmith:
Thanks.  I updated to 3.1.1 and indeed line information is provided in the input module block position for runtime errors.  But it is _not_ provided for syntax and similar errors:

```
In [1]: if True:
   ...:     x=1 b
  Input In [1]
    x=1 b
        ^
SyntaxError: invalid syntax
```

I also noticed that the error location itself in the `<module>` bit is incorrect, at least in some cases (always mentioning `cell line: 1`):

```
In [2]: if True:
   ...:     x=1/0
   ...: 
Input In [2], in <cell line: 1>()
      1 if True:
----> 2     x=1/0

```

I think it makes more sense for line info to appear with the filename.  Here's how the last example looks with my patch in place (which points out the line mismatch!):

```
In [2]: if True:
   ...:     x=1/0
   ...: 
Input In [2]:2, in <cell line: 1>()
      1 if True:
----> 2     x=1/0

```

MrMino:
You're right.

I think it should have the same format in all cases though. Having different formats depending on error type seems wrong. I know that `SyntaxError` means that there isn't even a cell created, so maybe:

```
Input In [2], in <line: 1>
```

The `()` at the end, when `<xyz>` is used in the traceback bugs me too. It's never a callable, it's either a cell or a module.

BTW. I can't reproduce the line number issue you mentioned, mine shows `line: 2`, as it should 

```
Input In [1], in <cell line: 2>()
      1 if True:
----> 2     x=1/0

```

jdtsmith:
> I think it should have the same format in all cases though. Having different formats depending on error type seems wrong. I know that `SyntaxError` means that there isn't even a cell created, so maybe:
> 
> ```
> Input In [2], in <line: 1>
> ```

Well the `<>` part also tells you about the function:

```
In [6]: def foo():
   ...:     x = 1/0
   ...: 

In [7]: foo()
Input In [7], in <cell line: 1>()
----> 1 foo()

Input In [6], in foo()
      1 def foo():
----> 2     x = 1/0

```

So here I'd say `Input In [6], line 2, in foo()` would be the most straightforward.  I also find it strange to reference a "cell" if you are not using Jupyter.  And it's redundant with "Input In [123]".

> The `()` at the end, when `<xyz>` is used in the traceback bugs me too. It's never a callable, it's either a cell or a module.

Agreed.  I wasn't sure what that's for...

> 
> BTW. I can't reproduce the line number issue you mentioned, mine shows `line: 2`, as it should
> 
> ```
> Input In [1], in <cell line: 2>()
>       1 if True:
> ----> 2     x=1/0
> 
> ZeroDivisionError: division by zero
> ```

Hmm, for me I just pulled the latest from home-brew:

```
% ipython
Python 3.10.2 (main, Feb  2 2022, 06:19:27) [Clang 13.0.0 (clang-1300.0.29.3)]
IPython 8.1.1 -- An enhanced Interactive Python. Type '?' for help.
```

and it definitely always says `<cell line: 1>()` no matter how long the input is:

```
In [4]: if True:
   ...:     x=14
   ...:     x=13
   ...:     x=12
   ...:     x=11
   ...:     x=10
   ...:     x=9
   ...:     x/=0
   ...: 
Input In [4], in <cell line: 1>()
      6 x=10
      7 x=9
----> 8 x/=0
```

jdtsmith:
Any other thoughts on this?  I definitely favor a simpler format like `Input In [6], line 2, in foo()`, and would suggest then removing the redundant info from #13535.

MrMino:
@jdtsmith Sorry for the long wait. I think the code from #13535 would have to go, but I'm not sure. I might not be aware of all the consequences of that (Jupyter? CPython 3.10?), so it would be helpful if you included removal of that code in this PR.

Also, looking on the format from this PR, I'm not sold on using `:` to denote line numbers. Similarly, it seems wrong that the current code uses `()` in `<cell line: ...>()`. All in all the current format looks confusing to me, and adding `:line_no` doesn't help that.

To accept a change in the format, all of this would have to be taken into account.

My idea to make it clearer would be:
```
ZeroDivisionError                  
Input cell In [3], line 1
----> 1 zzz()

Input cell In [2], line 2 in zzz()
      1 def zzz():
----> 2     0/0
```

Where, preferably, the only highlighted (colored) parts are the ones that are executable (in this case `In [2]` and `zzz()`). The line number could be highlighted too, to make it easier to lock onto it visually.

jdtsmith:
I like your proposed format. Here's how it looks (with the `_update_code_co_name` bit from #13535 in for now):

image

BTW, this leaves alone the `file:lineno` format recently introduced to support more editors for TB's pointing to external files. 

I too am concerned about the `cell line: 123` stuff being important for Jupyter debugging, which I don't use.  If we simply revert that entire change, it goes back to:

image

Is that preferable?  BTW, it seems #13535 wasn't actually necessary to fix the breakpoint problem, based on [this comment](https://github.com/ipython/ipykernel/issues/841#issuecomment-1052035079). But that should be verified.

jdtsmith:
Looking more closely at #13535, this was definitely a "workaround" to embed the line number in the code name, to permit debugpy to correctly identify positions in the absence of unique line numbers.  But it seems this was fixed on the debugpy side.  And it's a bit confusing syntax as you note, and in iPython it doesn't seem to function — I've opened #13608 on that topic.

jdtsmith:
I received [confirmation](https://github.com/ipython/ipython/pull/13535#issuecomment-1080590958) that the `cell line` fix from #13535 is no longer necessary and can be reverted, so have done so here. I've also removed the redundant ", in <module>" bit for all xmodes when the traceback refers to an input line.  It now looks quite clean:

image

jdtsmith:
We may also consider dropping the initial `Input`, since it will always have `In` after it.

image

jdtsmith:
Let me know when you've had a chance to look this over, @MrMino.

jdtsmith:
I had neglected to add the `whatsnew` @MrMino; that's now done.  Anything else?

MrMino:
Not really, I just need to find the time to get back to this and focus. Maybe the checks need some investigation, I have an aversion to merging things with failing checks.

jdtsmith:
Thanks for your efforts, @MrMino.  I fixed the linter issue and also the ultratb-related tests (expecting `in <module>` still to occur).  The remaining checks are for 3.11dev and seem to be timeouts in expect-related debugger stuff, so unrelated.

jdtsmith:
@MrMino just a ping to see if anything else remains here.  Thanks for you efforts.

IPython/core/ultratb.py
```diff
@@ -464,29 +467,38 @@ def _format_list(self, extracted_list):
...
             _format_filename(
                 filename, Colors.filenameEm, Colors.normalEm, lineno=lineno
             ),
-            Colors.nameEm,
-            name,
-            Colors.normalEm,
-            Colors.Normal,
         )
+        if name != "<module>":
+            item += " in %s%s%s" % (
```
MrMino, line 496: Since we are changing this anyway, can we please get rid of `%`-formatting and use f-strings?

You can leave other lines in the module as is, but let's be good scouts and rice up the ones that you are changing.

docs/source/whatsnew/pr/restore-line-numbers.rst
```diff
@@ -0,0 +1,5 @@
+Restore line numbers for Input
+==================================
+
+Line number information in tracebacks from input are restored.
+Line numbers from input were removed during the transition to v8 enhanced traceback reporting.
```
MrMino, line 5: It would be super nice discoverability-wise to include an example of changed traceback here. Just so that the users see that _we care_ ‍😁.

MrMino:
I searched through Jupyter and Ipykernel source to find whether this is being parsed anywhere, and to my surprise - it is. And #13535 might have introduced a regression. `ipykernel.debugger` sometimes looks for `"<module>"` in the stack frames - I'm not sure if it is has any relation to the `code.replace` that was done there before, but it doesn't look good.

So yes, I think we should go forward with this.

Other than that, there are some minor issues that I'd like to address first (see comments).
Also, please double check that other places where `_format_filename` is being used aren't negatively affected by this change.

docs/source/whatsnew/pr/restore-line-numbers.rst
jdtsmith: Done.

jdtsmith:
> I searched through Jupyter and Ipykernel source to find whether this is being parsed anywhere, and to my surprise - it is. And #13535 might have introduced a regression. `ipykernel.debugger` sometimes looks for `"<module>"` in the stack frames - I'm not sure if it is has any relation to the `code.replace` that was done there before, but it doesn't look good.

That is interesting.  So I guess our having removed ", in <module>" we are no worse off than the situation after #13535.

jdtsmith:
> Also, please double check that other places where _format_filename is being used aren't negatively affected by this change.

Took a look at this and didn't see anything obvious.  Thanks for the review.

IPython/core/ultratb.py
```diff
@@ -731,7 +735,7 @@ def format_record(self, frame_info):
         if lvals_list:
             lvals = '%s%s' % (indent, em_normal.join(lvals_list))
 
-        result = "%s, %s\n" % (link, call)
+        result = "%s%s%s\n" % (link, ", " if call else "", call)
```
MrMino, line 738: Could you please redo this to an f-string too? Also the ones at L493, L485, L470?

jdtsmith: I simplified the logic to avoid repetition for the final case, and converted all in that function to f-strings.

IPython/core/ultratb.py
```diff
@@ -463,34 +466,24 @@ def _format_list(self, extracted_list):
...
-            item = "  %s in %s%s%s\n" % (
-                _format_filename(
-                    filename, Colors.filename, Colors.Normal, lineno=lineno
-                ),
-                Colors.name,
-                name,
-                Colors.Normal,
+        for ind, (filename, lineno, name, line) in enumerate(extracted_list):
+            normalCol, nameCol, fileCol, lineCol = (
+                (Colors.normalEm, Colors.nameEm, Colors.filenameEm, Colors.line)
```
MrMino, line 471:
```suggestion
                # Emphasize the last entry
                (Colors.normalEm, Colors.nameEm, Colors.filenameEm, Colors.line)
```

MrMino:
Please check out the failures in [Build docs](https://github.com/ipython/ipython/runs/6851840033?check_suite_focus=true) and [formatting](https://github.com/ipython/ipython/runs/6851839929?check_suite_focus=true) - they are relevant.

MrMino (changes requested):
Please check out the failures in [Build docs](https://github.com/ipython/ipython/runs/6851840033?check_suite_focus=true) and [formatting](https://github.com/ipython/ipython/runs/6851839929?check_suite_focus=true) - they are relevant.

jdtsmith:
Thanks, I don't know the doc build system well; is there an obvious change here?

```
Restore line numbers for Input
==================================

Line number information in tracebacks from input are restored.
Line numbers from input were removed during the transition to v8 enhanced traceback reporting.

So, instead of:

  Input In [3], in <cell line: 1>()
  ----> 1 myfunc(2)

  Input In [2], in myfunc(z)
        1 def myfunc(z):
  ----> 2     foo.boo(z-1)

  File ~/code/python/ipython/foo.py:3, in boo(x)
        2 def boo(x):
  ----> 3     return 1/(1-x)

The error traceback now looks like:

  Cell In [3], line 1
  ----> 1 myfunc(2)

  Cell In [2], line 2, in myfunc(z)
        1 def myfunc(z):
  ----> 2     foo.boo(z-1)

  File ~/code/python/ipython/foo.py:3, in boo(x)
        2 def boo(x):
  ----> 3     return 1/(1-x)

or, with xmode=Plain:

    Cell In [12], line 1
      myfunc(2)
    Cell In [6], line 2 in myfunc
      foo.boo(z-1)
    File ~/code/python/ipython/foo.py:3 in boo
      return 1/(1-x)
```

MrMino:
Looks good, thanks!

PR #13566, opened 2022-03-03
Adding and "End" key binding for applying an auto suggestion

### LINKED ISSUE #13492: Auto-complete hinted suggestions using the END key
Sorry if this is duplicated and has been implemented already, but I did search the issue tracker and did not find anything mentioning it.

The auto-complete hints in the new version are awesome, however I find myself trying to apply them by pressing the END key, just like I do in the fish shell, and IPython does not seem to have that keybind enabled by default.

Would it be possible to consider adding this to the default keybinds? Several other auto-completion systems do the same, so I don't think I am the only user that would be benefit from this change. I think it would make the defaults slightly better :blush: 

Thank you!

jyuv:
For [#13492](https://github.com/ipython/ipython/issues/13492)

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -278,7 +278,7 @@ def create_ipython_shortcuts(shell):
     @kb.add("end", filter=has_focus(DEFAULT_BUFFER) & ebivim)
     def _(event):
         _apply_autosuggest(event)
-      
+
     @kb.add("c-e", filter=focused_insert_vi & ebivim)
     def _(event):
         _apply_autosuggest(event)
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -267,14 +267,21 @@ def create_ipython_shortcuts(shell):
     focused_insert_vi = has_focus(DEFAULT_BUFFER) & vi_insert_mode
 
     # Needed for to accept autosuggestions in vi insert mode
-    @kb.add("c-e", filter=focused_insert_vi & ebivim)
-    def _(event):
+    def _apply_autosuggest(event):
         b = event.current_buffer
         suggestion = b.suggestion
         if suggestion:
             b.insert_text(suggestion.text)
         else:
             nc.end_of_line(event)
+
+    @kb.add("end", filter=has_focus(DEFAULT_BUFFER) & ebivim)
+    def _(event):
+        _apply_autosuggest(event)
+      
+    @kb.add("c-e", filter=focused_insert_vi & ebivim)
+    def _(event):
+        _apply_autosuggest(event)
 
     @kb.add("c-f", filter=focused_insert_vi)
     def _(event):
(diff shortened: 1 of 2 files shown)


IPython/terminal/shortcuts.py
```diff
@@ -267,15 +267,22 @@ def ebivim():
...
-    def _(event):
+    def _apply_autosuggest(event):
         b = event.current_buffer
         suggestion = b.suggestion
         if suggestion:
             b.insert_text(suggestion.text)
         else:
             nc.end_of_line(event)
 
+    @kb.add("end", filter=has_focus(DEFAULT_BUFFER) & ebivim)
```
MrMino, line 278: The `ebvim` here states, that if you disable `TerminalInteractiveShell.emacs_bindings_in_vi_insert_mode`, the `End` key stops working in non-vi input mode too, which I think is something we don't want.

The filter here should check if it's in non-vi input mode - if yes, then apply the binding unconditionally.
If we're using Vi bindings, apply it only when `ebvim` is `True`.

docs/source/whatsnew/pr/end-shortcut-accept-suggestion.rst
```diff
@@ -0,0 +1,5 @@
+Added shortcut for accepting auto suggestion
+============================================
+
+Added End key shortcut for accepting auto-suggestion
```
MrMino, line 4: Would be nice to mention that the binding works in Vi mode too, provided `TerminalInteractiveShell.emacs_bindings_in_vi_insert_mode` is set.

IPython/terminal/shortcuts.py
MrMino: See [prompt toolkit's builtin filters](https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/filters.html#built-in-filters) - I think you could use `~vi_mode` to get the right condition.

docs/source/whatsnew/pr/end-shortcut-accept-suggestion.rst
jyuv: you are right. I adapted accordingly

docs/source/whatsnew/pr/end-shortcut-accept-suggestion.rst
```diff
@@ -0,0 +1,7 @@
...
+This binding works in Vi mode too, provided
```
MrMino, line 5:
```suggestion
This binding works in Vi insert mode too, provided 
```

MrMino:
LGTM

Thanks @jyuv, and congratulations on your first contribution to IPython 🎉

FFY00:
Thank you for this!

digismack:
I believe this feature has introduce a bug which is driving me crazy. I can no longer reliably use the end key to jump to the end of the line.

IPython/terminal/shortcuts.py
krassowski: So this ended as `(ebivim | ~vi_insert_mode)`. It feels kind of wrong to me because in vi command mode it still evaluates to true (AFAIK). I wonder if it would be better as `(vi_insert_mode & ebivim) | emacs_insert_mode`. @MrMino do you happen to still remember if it was considered/would have any disadvantage?

PR #13567, opened 2022-03-03
[WIP] Use prompt_toolkit's built-in cursorshape support.

jonathanslenders:
prompt_toolkit now has built-in support for cursor shapes. This should take advantage of this feature.
This should also fix some cursorshape related issues, like resetting the cursor shape when IPython suspends.

The PR currently doesn't merge yet. I'll try to rebase and retest later this week.

Note that this requires at least prompt_toolkit 3.0.27, which is set as a minimal required version. Because of that requirement, and this issue that was recently filed: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1576 , I plan into looking into that as well and fix both together. (If we set 3.0.27 as minimal required version, that other issue has to be fixed.)

Related issue: https://github.com/ipython/ipython/issues/13472

### DIFF
--- a/IPython/terminal/debugger.py
+++ b/IPython/terminal/debugger.py
@@ -10,6 +10,8 @@ from . import embed
 
 from pathlib import Path
 from pygments.token import Token
+from prompt_toolkit.application.current import get_app
+from prompt_toolkit.cursor_shapes import ModalCursorShapeConfig
 from prompt_toolkit.shortcuts.prompt import PromptSession
 from prompt_toolkit.enums import EditingMode
 from prompt_toolkit.formatted_text import PygmentsTokens
@@ -73,7 +75,6 @@ class TerminalPdb(Pdb):
 
         options = dict(
             message=(lambda: PygmentsTokens(get_prompt_tokens())),
-            editing_mode=getattr(EditingMode, self.shell.editing_mode.upper()),
             key_bindings=create_ipython_shortcuts(self.shell),
             history=self.debugger_history,
             completer=self._ptcomp,
@@ -82,6 +83,7 @@ class TerminalPdb(Pdb):
             complete_style=self.shell.pt_complete_style,
             style=getattr(self.shell, "style", None),
             color_depth=self.shell.color_depth,
+            cursor=(ModalCursorShapeConfig() if self.shell.modal_cursor else None),
         )
 
         if not PTK3:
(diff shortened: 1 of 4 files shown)


jonathanslenders:
There is still a sigint-related issue here.

Running ``killall -SIGINT ipython`` in another terminal window is expected to either do nothing or cause the same behavior as pressing control-c. However, IPython crashes with the following error:

```
  File "/home/jonathan/.pyenv/versions/radkit/bin/ipython", line 33, in <module>
    sys.exit(load_entry_point('ipython', 'console_scripts', 'ipython')())
  File "/home/jonathan/git/ipython/IPython/__init__.py", line 124, in start_ipython
  File "/home/jonathan/.pyenv/versions/3.9.10/envs/radkit/lib/python3.9/site-packages/traitlets/config/application.py", line 846, in launch_instance
    app.start()
  File "/home/jonathan/git/ipython/IPython/terminal/ipapp.py", line 318, in start
    self.shell.mainloop()
  File "/home/jonathan/git/ipython/IPython/terminal/interactiveshell.py", line 690, in mainloop
    self.interact()
  File "/home/jonathan/git/ipython/IPython/terminal/interactiveshell.py", line 675, in interact
    code = self.prompt_for_code()
  File "/home/jonathan/git/ipython/IPython/terminal/interactiveshell.py", line 611, in prompt_for_code
    text = self.pt_app.prompt(
  File "/home/jonathan/git/python-prompt-toolkit/src/prompt_toolkit/shortcuts/prompt.py", line 1034, in prompt
    return self.app.run(
  File "/home/jonathan/git/python-prompt-toolkit/src/prompt_toolkit/application/application.py", line 937, in run
    return loop.run_until_complete(
  File "/home/jonathan/.pyenv/versions/3.9.10/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/jonathan/git/python-prompt-toolkit/src/prompt_toolkit/application/application.py", line 664, in run_async
    assert not self._is_running, "Application is already running."
AssertionError: Application is already running.

If you suspect this is an IPython 8.5.0.dev bug, please report it at:

to interactively debug it.

```

However, the above error does not happen if we pass ``handle_sigint=True`` (the prompt_toolkit default), which I expect is what happens right now with latest prompt_toolkit version. Most likely a bug in prompt_toolkit that's there for some time.

jonathanslenders:
Okay, I think I understand what's going on here regarding the signal issues.

If no signal handler is installed, and a ``SIGINT`` is delivered to IPython (externally, not by pressing control-c which is handled as a key binding), then Python throws a ``KeyboardInterrupt`` in the middle of the code.  This currently cannot be handled in a coroutine, which prevents any proper finalization to happen.

Take this code snippet for instance:
```python
import asyncio
async def run_forever():
    print('before')
    try:
        await asyncio.Future()
    finally:
        print('in finally block in coroutine')
try:
    asyncio.run(run_forever())
finally:
    print('in main finally block')
```

Send a ``SIGINT`` to that process. The finally block in the coroutine won't execute, the finally block outside of the coroutine will execute. This means that there's a lot of cleanup in the coroutine that won't happen. (See also: https://github.com/python/asyncio/issues/341 )

In case of prompt_toolkit, where the blocking ``prompt()`` call spawns an event loop and has everything happen async. This means that if no signal handler is installed, the prompt_toolkit ``Application`` ends up in an unusable state and much of the cleanup won't happen (restoring TTY properties, restoring the ``Application`` state.) I can try to recover from that, but having prompt_toolkit handle the signal is the cleaner option in any case. So, I'll revert this part of the pull request.

jonathanslenders:
While doing these changes, I discovered that IPython's built-in key bindings somewhat clash in Vi mode. Maybe this is why the ttimeoutline/timeoutlen configuration options are there (they were not passed correctly in the previous version of my patch here).

The problem is, I'm typing ``escape, b`` fast enough in insert mode, but expect to go into navigation mode and go back one word. IPython seems to do "backward_word", but doesn't end up in vagitation mode due to the custom binding in ``terminal/shortcuts.py``, line 325.

Anyway, I guess the changes here are fine to merge.

Carreau:

PR #13572, opened 2022-03-04
Start migrating somethings to pytest's monkeypatch

Carreau:
I think this may be some of things that have created troubles for @bollwyvl on conda-forge ?


bollwyvl:
Yeah, these look grand.

Issue #13573, opened 2022-03-04
support for recursive autoreloading


smason:
I'm using `autoreload` in a Jupyter notebook while developing code out in Python files.  This code is spread across a few different modules, hence I've been using something like:

```
%load_ext autoreload

%autoreload 1
%aimport foo.bar, foo.baz, foo.baz.qux
```

but I've got a few submodules, so I thought it might be nice to allow a "recursive" option to `aimport`, e.g. something like:

```
%aimport foo.*
```

looking at the code a bit I think I might need to use something other than `aimport` to register a pattern, as I don't care about actually doing the import at that point, I just want to be able to do:

```
from foo.baz import qux
```

later on and have `qux` autoreloaded, along with whatever else it pulls in under `foo`

happy to put a pull request together, just wondering if this sensible.  am tempted to use the `fnmatch` module for this matching in `ModuleReloader.check`, but open to suggestions

MrMino:
>  just wondering if this sensible. 

I think it depends on how you detect whether something is a submodule or not. And if the feature will be as simple as "substitute `A.*` for each absolute import path of A's submodules at the time of calling `%aimport`" - I'm down.

Anything more magical than this runs the risk of introducing regressions - autoreload is already a rather complex feature, any additions to it will need to stay on the simple side.

PR #13575, opened 2022-03-06
support for pattern based autoreloading

smason:
a minimal implementation of glob based autoreloading, as proposed in #13573

usage is something like:

```python
%autoreload 1
%autoreload foo.*

import foo.bar
```

and then modifications to `foo/bar.py` will autoreload.  not sure about overloading `autoreload` here, but it was easy to put in and should be easy to change to something else.  it also needs the mode to be set to `1` for this to work, but could be done automatically if people think it's a good idea

would be good to add a test, I've been using the attached script [test.txt](https://github.com/ipython/ipython/files/8193588/test.txt), for interactive tests

### DIFF
--- a/IPython/extensions/autoreload.py
+++ b/IPython/extensions/autoreload.py
@@ -145,10 +145,9 @@ class ModuleReloader:
         # Modules that failed to reload: {module: mtime-on-failed-reload, ...}
         self.failed = {}
         # Modules specially marked as autoreloadable.
-        self.modules = {}
-        self.module_pats = {}
+        self.modules = set()
         # Modules specially marked as not autoreloadable.
-        self.skip_modules = {}
+        self.skip_modules = set()
         # (module-name, name) -> weakref, for replacing old code objects
         self.old_objects = {}
         # Module modification timestamps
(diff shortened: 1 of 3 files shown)


smason:
@MrMino is this change in line with what you're expecting?  if it looks like it's going to be acceptable I can fill out more.  the main thing I'd want to do would be to include patterns when printing out the current state, via `%aimport`

MrMino:
The usecase certainly looks good.

But I'm not sure why it's necessary to keep the patterns in memory. Why not just populate them while parsing, i.e. expand the pattern at the time the user executes `%aimport`?

Recalculating them at the time of the reload would be error prone, especially in cases where modules are being renamed / moved.

smason:
> Recalculating them at the time of the reload would be error prone, especially in cases where modules are being renamed / moved.

Could you say what you mean by moves/renames being "error prone"?  The existing upstream code just keeps the module names in a list of strings, so renaming a file would just cause the renamed file not to be reloaded.

With my patterns, you could register "foo.*" and if you renamed `foo/bar.py` to `foo/baz.py` it would continue to autoreload `baz` if/when it was imported under the new name. Which is what I'd want it to do.  Maybe other other users have different use cases?

> But I'm not sure why it's necessary to keep the patterns in memory. Why not just populate them while parsing, i.e. expand the pattern at the time the user executes `%aimport`?

My pattern based approach does not rely on `%aimport` being used at all.  In my mind `aimport` does two things, 1. imports the module, and 2. registers this module for autoreloading.

When using patterns these get broken out into two steps, the magic just register a pattern for autoreloading, you then use standard Python language features for doing the actual import.  `import` in Python is much more flexible than the syntax supported by the autoreload `%aimport` magic.  For example, how would you do `from foo import bar`, or what if you just wanted some sub-module imported but didn't care about importing it directly.  These cases are easily supported when this functionality is separated.

At the very least, I think we need to "keep them in memory" so we can remind the user what is currently being autoreloaded, similar to the existing functionality when you call `%aimport` without an argument.

MrMino:
> Could you say what you mean by moves/renames being "error prone"?

I somehow got the wrong idea that you'd also import them as soon as the machinery gets the pattern. I now understand its not the case.

I'm ok with making it take new files into account, as long as it doesn't automatically import them.

With that said, there are other things stay in the way of pulling this:

- I'm not sold on holding the patterns in a separate dict. Why not use `fnmatch` over all available keys from `self.modules`? This would simplify the code. I don't think that the marginal performance improvement from `re.compile` is worth it, if there even is any - `fnmatch` uses `re` under the hood, `re` compiles and caches all patterns anyway. For almost all cases, you're only optimizing dict lookup `re` does under the hood, which IMO isn't worth the additional lines.
- It would be nice to make `%aimport` work with this as well. Are there any reasons not to do so?
- This needs tests. Preferably some that will also test renames. I don't know how cumbersome it is to test this stuff, but at the very least we need to have something that tests basic usage.
- [What's-new entry](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr).

smason:
> * I'm not sold on holding the patterns in a separate dict. [...] I don't think that the marginal performance improvement from `re.compile` is worth it

My thinking was that an IPython session with just matplotlib imported results in a thousand modules being loaded and checking these against N patterns on every line of input seems like a lot of unnecessary work.

I've had a look at [fnmatch](https://github.com/python/cpython/blob/3.10/Lib/fnmatch.py#L160) and it seems to indicate that users already combine multiple patterns together.  I've changed my code to do this, so it can compile a big regex over all patterns, so we only need to check this against our loaded modules.

> * It would be nice to make `%aimport` work with this as well. Are there any reasons not to do so?

had a rethink, and have changed things around, you can now register a pattern by doing:

```python
%aimport +foo.*
```

and it also supports negative/skip patterns as well now:

```python
%aimport -matplotlib.*
```

when using `+` or `-` it just registers the pattern and doesn't actually perform the import at that point.

> * This needs tests. Preferably some that will also test renames. I don't know how cumbersome it is to test this stuff, but at the very least we need to have something that tests basic usage.

yup, haven't looked into how I'd test an IPython extension yet.  wanted to make sure basic functionality was likely to be accepted first.  will have a look at how it's done and try and get something together over the next few days

> * [What's-new entry](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr).

done, not sure why the check is failing.  the `tools/update_whatsnew.py` says there's "no automatic update available".

smason:
> > * This needs tests. Preferably some that will also test renames. I don't know how cumbersome it is to test this stuff, but at the very least we need to have something that tests basic usage.
> 
> yup, haven't looked into how I'd test an IPython extension yet. wanted to make sure basic functionality was likely to be accepted first. will have a look at how it's done and try and get something together over the next few days

I've had a look at the test infrastructure and given it a bit of an update in 58d00f53178288a656405ec095aa31fe8c705ec9 to remove the annoying 1 second pauses on each file rewrite.  I've also added a test of basic pattern based usage.

I've been struggling to think of a good use case that would "test renames" in any meaningful way.  As far as the autoreloader is concerned it's just the same as the old module being deleted, and a new one appearing.  The new one would need to be imported before it'll ever be autoreloaded again, at which point the existing code will take over.

Or is this something that you think is worth explicitly testing?

Carreau:

PR #13592, opened 2022-03-14
Get history from sql.

### LINKED ISSUE #13585: %paste + history
From a user point of view, it seem like `%paste` should not store itself in IPython's history, but instead store the text that was pasted. It seems unlikely that a user will want to use history to get to paste again over get to the thing that was pasted.

ipython-paste

Carreau:
Fixes #13585

By getting history from sql we can get the transformed history.
This also skip storing history if `%paste` is used and `%paste` itself
will insert the pasted value in history which is more conveninent.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3042,9 +3042,8 @@ class InteractiveShell(SingletonConfigurable):
                 cell = raw_cell
 
         # Store raw and processed history
-        if store_history:
-            self.history_manager.store_inputs(self.execution_count,
-                                              cell, raw_cell)
+        if store_history and raw_cell.strip(" %") != "paste":
+            self.history_manager.store_inputs(self.execution_count, cell, raw_cell)
         if not silent:
             self.logger.log(cell, raw_cell)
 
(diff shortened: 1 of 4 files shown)


NathanDunfield:
This PR introduced an incompatibility with `prompt_toolkit 2.0`, see #13645.

monester:
Hello, thanks for this PR - it would definitely work in some cases, but what if I constantly want to copy from tutorial and paste to console. Is there a way to keep %paste in history instead of actual code pasted?

IPython/core/interactiveshell.py
```diff
@@ -3042,9 +3042,8 @@ def error_before_exec(value):
                 cell = raw_cell
```
Jesuis33, line 3043: [license.txt](https://github.com/ipython/ipython/files/8829434/license.txt)

IPython/terminal/interactiveshell.py
```diff
@@ -32,7 +30,7 @@
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
 from prompt_toolkit.filters import (HasFocus, Condition, IsDone)
 from prompt_toolkit.formatted_text import PygmentsTokens
-from prompt_toolkit.history import InMemoryHistory
+from prompt_toolkit.history import History
```
Jesuis33, line 33: fastly compute init --from=https://github.com/fastly/compute-starter-kit-javascript-default

PR #13603, opened 2022-03-25
Fix ctrl-\ behavior

### LINKED ISSUE #12461: CTRL backslash sends QUIT signal but freezes the terminal
A convenient way to quit Ipython and kill the debugged process is to press CTRL \ (backslash).
Once pressed, the two exit but the terminal freezes. I can type commands, press enter and see the result but I can't see the command I'm typing.
Is there any problem with STD OUT redirection after the signal is sent?
I tested on gnome-terminal and xterm with bash

kmaork:
This pull request fixes two issues with ctrl-\ when using IPython on linux:
- Currently, pressing ctrl-\ will make IPython exit without resetting the terminal configuration. The terminal is rendered unusable unless you do something like typing `stty sane` and hoping for the best
- IPython users cannot override the behavior of ctrl-\ using `signal.signal(signal.SIGQUIT, ...)` as they would in other terminal apps. This change simulates better the usual behavior of ctrl-\ in linux terminal apps.

Closes #12461

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -10,6 +10,7 @@ import warnings
 import signal
 import sys
 import re
+import os
 from typing import Callable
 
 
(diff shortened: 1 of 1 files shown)


MrMino:
Would be nice to have that explanation directly in the commit message for posterity. The code looks good, but I still need to reproduce the original issue (I'm using tmux which, TIL, hijacks ctrl+\ for "go to last pane")

kmaork:
@MrMino thanks, I updated the commit message :)

Carreau:
Seem reasonable.

PR #13612, opened 2022-03-29
Don't strip decorators before feeding into block_parser

### LINKED ISSUE #13531: Sphinx ipython directive greedily strips all decorators, including valid python ones!

TomNicholas:
This should actually fix #13531 @Carreau (in conjunction with #13532).

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -821,8 +821,11 @@ class EmbeddedSphinxShell(object):
                 output.append(line)
                 continue
 
-            # handle decorators
-            if line_stripped.startswith('@'):
+            # handle pseudo-decorators, whilst ensuring real python decorators are treated as input
+            if any(
+                line_stripped.startswith("@" + pseudo_decorator)
+                for pseudo_decorator in PSEUDO_DECORATORS
+            ):
                 output.extend([line])
                 if 'savefig' in line:
                     savefig = True # and need to clear figure
--- a/docs/source/whatsnew/pr/stripping-decorators-bug.rst
+++ b/docs/source/whatsnew/pr/stripping-decorators-bug.rst
@@ -1,0 +1,4 @@
+Stripping decorators bug
+========================
+
+Fixed bug which meant that ipython code blocks in restructured text documents executed with the ipython-sphinx extension skipped any lines of code containing python decorators.


TomNicholas:
Can this please be merged? The failure seems to appear on a bunch of other PRs so isn't caused by this PR's changes.

blsqr:
@TomNicholas @Carreau If I may jump in (as I'm also interested in this getting merged and released): Is there anything left to do here?

TomNicholas:
As far as I know there isn't anything to do - I don't think the test failure has anything to do with me as it also happens on `main`.

MrMino:
Please add a what's new entry that explains what changed from the PoV of the end user. See [documenting what's new](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr).

TomNicholas:
I've added a whatsnew, though now somehow some of the tests appear to be failing...

MrMino:
The failures are irrelevant. See #13695.

Comparing this to `master`, this does seem to fix the issue.

Thank you!

PR #13614, opened 2022-03-30
Try to delay non-local invalidation.

### LINKED ISSUE #12408: 'nonlocal' misclassified as incomplete code
If I enter, line-by-line, a nonlocal statement in an inner function, ipython throws a syntax error, e.g.:

```
In [1]: def a():
   ...:     def b():
   ...:         nonlocal c
  File "<ipython-input-1-dbf568c371f7>", line 3
    nonlocal c
    ^
SyntaxError: no binding for nonlocal 'c' found
```

The following code runs fine on python interpreter (and if execured as a file):

```python
def a():
    def b():
        nonlocal c
        return c + 1
    c = 1
    b()
    return c

print(a())
```

This was originally reported by Chris Angelico as a bug in IDLE -- https://bugs.python.org/issue19335

I can successfully reproduce it on IPython 7.16.1. Here is the output of `python -c "import IPython; print(IPython.sys_info())"`:

```python
{'commit_hash': '2486838d9',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/rahuljha/.virtualenvs/vulture/lib/python3.7/site-packages/IPython',
 'ipython_version': '7.16.1',
 'os_name': 'posix',
 'platform': 'Darwin-16.7.0-x86_64-i386-64bit',
 'sys_executable': '/Users/rahuljha/.virtualenvs/vulture/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '3.7.0b5 (v3.7.0b5:abb8802389, May 30 2018, 20:03:19) \n'
                '[Clang 6.0 (clang-600.0.57)]'}
```

Carreau:
If we see non-local, parse without using non-local lines
if the last line is non-empty.

This give users a change to define the variable

Closes #12408

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -752,11 +752,28 @@ class TransformerManager:
         # We'll use codeop.compile_command to check this with the real parser
         try:
             with warnings.catch_warnings():
-                warnings.simplefilter('error', SyntaxWarning)
-                res = compile_command(''.join(lines), symbol='exec')
-        except (SyntaxError, OverflowError, ValueError, TypeError,
-                MemoryError, SyntaxWarning):
-            return 'invalid', None
+                warnings.simplefilter("error", SyntaxWarning)
+                res = compile_command("".join(lines), symbol="exec")
+        except (
+            SyntaxError,
+            OverflowError,
+            ValueError,
+            TypeError,
+            MemoryError,
+            SyntaxWarning,
+        ):
+            code = "".join(lines)
+            if "nonlocal" in code and lines[-1]:
+                try:
+                    res = compile_command(
+                        "".join([l for l in lines if "nonlocal " not in l]),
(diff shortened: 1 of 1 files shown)


MrMino:
This still fails one level below:

```
In [1]: def a():
   ...:     def b():
   ...:         nonlocal c
   ...:         def d():
   ...:             nonlocal e
   ...:         e = 123
  Input In [1]
    nonlocal c
    ^
SyntaxError: no binding for nonlocal 'c' found
```

MrMino:
Perhaps we could special-case on the exception itself? I.e. if we get `SyntaxError`, check if it's a no-binding for nonlocal, if it is - return `"incomplete"`?

There are other `SyntaxError`s for nonlocals probably, but the only one I can think of now is the one which happens if you define a nonlocal on a module level. If we'll do the above, we'd need to keep it in mind.

PR #13619, opened 2022-04-03
Make terminal title better behaved

### LINKED ISSUE #11482: Terminal title should be disabled when using IPython in a pipeline
This came up recently here: https://github.com/sharkdp/bat/issues/403

Here's an example of how IPython prints control characters even when used in a pipeline:

```
/$ ipython -c "import os; print(os.isatty(1))"
True

/$ ipython -c "import os; print(os.isatty(1))" | gcat -A
^[]0;IPython: /^GFalse$
```

It would be good if IPython checked `isatty` (as shown above) to not add terminal title control characters if `ipython` is being used in a pipeline.

To be clear, this isn't just a convenience feature, as it can actually cause unexpected behavior when used in a pipeline. For example, one would expect that a `grep` for the letter `o` in the output of "False" above would return false:

```
/$ ipython -c "import os; print(os.isatty(1))" | grep o
01;31mon: /False
```

But it returns a match! Running the same command using stock Python correctly returns nothing (and 1 return code):

```
/$ python3 -c "import os; print(os.isatty(1))" | grep o
/$:1
```

nsmaciej:
* Do not `toggle_set_term_title` if the output is not a tty (Fixes #11482)
* Fix non-interactive IPython setting the term title but neglecting to restore it
* Only save the the term title the first time `set_term_title` is called. Before this `%cd` would also save the title causing the initial title to not be restored.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -405,14 +405,14 @@ class TerminalInteractiveShell(InteractiveShell):
     @observe('term_title')
     def init_term_title(self, change=None):
         # Enable or disable the terminal title.
-        if self.term_title:
+        if self.term_title and _is_tty:
             toggle_set_term_title(True)
             set_term_title(self.term_title_format.format(cwd=abbrev_cwd()))
         else:
             toggle_set_term_title(False)
 
     def restore_term_title(self):
-        if self.term_title:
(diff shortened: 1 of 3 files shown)


maciej-irl:
Only semi-related but #11909 can probably be closed, it's been fixed since 2019

PR #13621, opened 2022-04-04
display.Javascript should use `rel` in CSS link.

asford:
Seeing errors when using `display.Javascript` with `css` parameters.
The target css does not load in current chrome, due to the use of:

`<link ref="stylesheet" ...>` 

rather than:

`<link rel="stylesheet" ...>`

in the display call.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -625,6 +625,7 @@ class JSON(DisplayObject):
     def _repr_json_(self):
         return self._data_and_metadata()
 
+
 _css_t = """var link = document.createElement("link");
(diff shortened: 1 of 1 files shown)


IPython/core/display.py
```diff
@@ -614,7 +614,7 @@ def _repr_json_(self):
         return self._data_and_metadata()
```
Carreau, line 615:
```suggestion


```

Carreau: Formatter is unhappy about just a single empty line...

PR #13640, opened 2022-04-17
Fix non-printable chars and reproducibility

apteryks:


IPython/sphinxext/ipython_directive.py
```diff
@@ -28,7 +28,7 @@
 
    In [0]: import time
 
-   In [0]: %timeit time.sleep(0.05)
+   In [0]: %pdoc time
```
MrMino, line 31: That result is a bit too big. Let's try this instead:

```suggestion
   In [0]: %pdoc time.sleep
```

MrMino:
Thanks!

Issue #12378, opened 2020-06-10
Custom traceback not working as documented


dakoop:
It looks like the custom traceback documentation doesn't match the implementation. From the `set_custom_exc` documentation:

https://github.com/ipython/ipython/blob/a80908bd8360ff4d476111500b66ff11af4373ed/IPython/core/interactiveshell.py#L1853-L1861

yet the structured traceback is not output:

```
Python 3.8.3 (default, May 19 2020, 13:54:14) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: aa                                                                      
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-1-312382290f4f> in <module>
----> 1 aa

NameError: name 'aa' is not defined

In [2]: def dummy_handler(self, etype, value, tb, tb_offset=None): 
   ...:     return ['Test'] 
   ...:                                                                         

In [3]: get_ipython().set_custom_exc((NameError,), dummy_handler)               

In [4]: aa                                                                      

In [5]:                                                                         
```

Investigating, in both `run_cell_async` and `run_code`, nothing is done with the returned value.

https://github.com/ipython/ipython/blob/a80908bd8360ff4d476111500b66ff11af4373ed/IPython/core/interactiveshell.py#L3034-L3037

https://github.com/ipython/ipython/blob/a80908bd8360ff4d476111500b66ff11af4373ed/IPython/core/interactiveshell.py#L3340-L3344

`CustomTB` seems to call a wrapper which calls the handler. It seems like there needs to be a `_showtraceback` around the `CustomTB` call to output the structured traceback. However, it's not totally clear what was intended as the `dummy_handler` code in `set_custom_exc` suggests that the handler should perhaps be printing directly, but the `validate_stb` call goes to great effort to verify the return value of the handler is a list of strings:

https://github.com/ipython/ipython/blob/a80908bd8360ff4d476111500b66ff11af4373ed/IPython/core/interactiveshell.py#L1878-L1905

Carreau:
HUm, the documentation around that is likely super old, and the traceback code was recently refactor to rely on stack_data. I'll have to dive into this.

dakoop:
[...]

PR #13668, opened 2022-05-02
fix uncaught `BdbQuit` exceptions on ipdb `exit`

telamonian:
fixes:
  - jupyterlab/jupyterlab#12501

refs:
  - ipython/ipython#876
  - ipython/ipython#1273
  - ipython/ipython#4474
  - ipython/ipython#5306
  - ipython/ipython#9731
  - ipython/ipython#9942
  - ipython/ipython#9950
  - ipython/ipython#10006
  - ipython/ipython#12378

`BdbQuit` is now handled in the top-most scope of `InteractiveShell.run_code`. This ensures that `BdbQuit` is correctly handled but can still do its job of breaking out of all user code/loops/further breakpoint requests. Hopefully will work better than previous attempts, which put the `BdqQuit` handling in `Pdb.set_trace`.

In terms of ipdb integration with the jlab UI, it's a big improvement.

Before:
image

After:
image

Much less noisy

pinging @Carreau for a review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,5 +28,13 @@ __pycache__
 .pytest_cache
 .python-version
 venv*/
-.idea/
 .mypy_cache/
+
+# jetbrains ide stuff
+*.iml
+.idea/
+
+# vscode ide stuff
+*.code-workspace
+.history
+.vscode
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -101,7 +101,6 @@ All the changes since then are under the same license as IPython.
 #
 #*****************************************************************************
 
-import bdb
 import inspect
 import linecache
 import sys
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -14,6 +14,7 @@
 import abc
 import ast
 import atexit
+import bdb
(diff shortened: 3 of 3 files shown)


telamonian:
I also tested to make sure that the change in this PR is compatible with the work done in #9942 and #9950 to enable running `terminal.debugger` as main, eg:

```bash
python -m IPython.terminal.debugger testDebug.py
```

and it seems to be compatible. However, there were other strange problems relating to imports that did keep the `python -m <some debugger> ...` stuff from working. I was able to work around those issues by shuffling some imports from top level to under the `TerminalPdb` class scope. It's out of scope for this PR, tho, so I'll open a new issue about that

.gitignore
```diff
@@ -28,5 +28,13 @@ __pycache__
...
 .mypy_cache/
+
+# jetbrains ide stuff
+*.iml
+.idea/
+
+# vscode ide stuff
+*.code-workspace
+.history
+.vscode
```
Carreau, line 40: really all these should be in your gitignore global:

https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files

Carreau (approved):

PR #13673, opened 2022-05-13
Sort parameter completions first and in the signature order

### LINKED ISSUE #12344: Autocomplete list is clutered with buitin names
I often use interactive transient ipython sessions to interactively explore the API of libraries I do not remember by heart.

For instance take the following function:

```python
>>> def f(key_word_arg=None): 
...     pass
...
```

If you tab complete in IPython, you get the following suggestions, most of which are from the builtins module and are probably useless 99.999% of the time:

```
>>> f( 
  abs()                     Exception                  
  all()                     exec()                     
  any()                     exit                       
  ArithmeticError           f()                        
  ascii()                   False                      
  AssertionError            FileExistsError            
  AttributeError            FileNotFoundError          
  BaseException             filter()                   
  bin()                     float                      
  BlockingIOError           FloatingPointError         
  bool                      format()                   
  breakpoint()              frozenset                  
  BrokenPipeError           FutureWarning              
  BufferError               GeneratorExit              
  bytearray                 get_ipython()              
  bytes                     getattr()                  
  BytesWarning              globals()                  
  callable()                hasattr()                  
  ChildProcessError         hash()                     
  chr()                     help()                     
  classmethod               hex()                     >
  compile()                 id()                       
  complex                   ImportError                
  ConnectionAbortedError    ImportWarning              
  ConnectionError           In                         
  ConnectionRefusedError    IndentationError           
  ConnectionResetError      IndexError                 
  copyright()               input()                    
  credits()                 int                        
  delattr()                 InterruptedError           
  DeprecationWarning        IOError                    
  dict                      IsADirectoryError          
  dir()                     isinstance()               
  divmod()                  issubclass()               
  ellipsis                  iter()                     
  Ellipsis                  key_word_arg=              
  enumerate                 KeyboardInterrupt          
  EnvironmentError          KeyError                   
  EOFError                  len()                      
  eval()                    license()                                          
```

In particular I would like to see `key_word_arg=` stand out of the crowded list.

Would it be possible to either:

- put the kwargs at the top of the tab complete options of a callable?
- always put the builtins at the bottom of the suggestions, maybe even below the locals names?
- or even remove the builtins entirely from the suggestion list?

This is running:
- ipython 7.12.0
- jedi 0.16.0

joelostblom:
Fixes https://github.com/ipython/ipython/issues/12344

This is a suggestion to change the sort order of parameters to be first and in the order the order they appear in the signature.

Currently, the default behavior when using Jedi is to not sort parameters first, which is not that useful when completing on a function call:
image

This PR changes that so that the function parameters are shown first:
image

The previous screenshot is already the default behavior when *not* using Jedi. This PR also includes an improvement for that which is sorting the parameters in the order they are specified in the signature instead of alphabetically (I think this is the most intuitive but open to feedback):
image

There is unfortunately not a way to achieve this last step with Jedi currently, but I have opened https://github.com/davidhalter/jedi/pull/1855 to implement that.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1960,7 +1960,7 @@ class IPCompleter(Completer):
         jedi_matches = []
         jedi_param_matches = []
         for jedi_match in iter(all_jedi_matches):
-            if jedi_match.name.endswith('='):
+            if jedi_match.name.endswith("="):
                 jedi_param_matches.append(jedi_match)
             else:
                 jedi_matches.append(jedi_match)
@@ -2151,8 +2151,7 @@ class IPCompleter(Completer):
         if self.use_jedi and not is_magic_prefix:
             if not full_text:
                 full_text = line_buffer
-            jedi_completions = self._jedi_matches(
-                cursor_pos, cursor_line, full_text)
+            jedi_completions = self._jedi_matches(cursor_pos, cursor_line, full_text)
 
         if self.merge_completions:
             matches = []
@@ -2178,13 +2177,15 @@ class IPCompleter(Completer):
             t, c = m
             if t not in seen:
                 # Keep the parameter matches in the function signature order
-                if c == 'IPCompleter.python_func_kw_matches':
+                if c == "IPCompleter.python_func_kw_matches":
                     filtered_param_matches[m] = None
                 else:
                     filtered_matches.add(m)
                 seen.add(t)
 
-        _filtered_matches = list(filtered_param_matches) + sorted(list(filtered_matches), key=lambda x: completions_sorting_key(x[0]))
+        _filtered_matches = list(filtered_param_matches) + sorted(
+            list(filtered_matches), key=lambda x: completions_sorting_key(x[0])
+        )
 
         custom_res = [(m, 'custom') for m in self.dispatch_custom_completer(text) or []]
         
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -317,9 +317,6 @@ def completions_sorting_key(word):
     elif word.startswith('_'):
         prio1 = 1
 
-    if word.endswith('='):
-        prio1 = -1
-
     if word.startswith('%%'):
         # If there's another % in there, this is something else, so leave it alone
         if not "%" in word[2:]:
@@ -1551,8 +1548,7 @@ class IPCompleter(Completer):
                        v.kind in _keeps)
         except ValueError:
             pass
-
-        return list(set(ret))
+        return list(dict.fromkeys(ret))
 
     def python_func_kw_matches(self, text):
         """Match named parameters (kwargs) of the last open function"""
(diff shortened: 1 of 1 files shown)


joelostblom:
I fixed the failing tests relating to the changes in this PR. The single remaining failure seems unrelated to this PR.

---

I also discovered a new error; when completing a function the order of the parameters influence the completion. Take the triangular distribution:

image

If I type the first parameter and then pres tab, I get the expected completion:

image

However, if I start with one of the other parameters, only one additional parameter shows up and the first one from the function signature (low) is hidden:

image

It is similar if I type out two parameters, depending on their order, the third parameter might be hidden. Here, `mode` shows up as expected:

image

Here, `high` is hidden:

image

Although it is valid code:

image

This was present before this PR and only happens in Jedi mode, but was harder to discover previously due to the parameters not being ordered first. I have opened an issue for it in the Jedi repo https://github.com/davidhalter/jedi/issues/1856 and don't think there is anything we can do about it in this PR.

joelostblom:
@MrMino @Carreau Let me know if you require any additional info or more examples before reviewing this PR. I appreciate that there are a lot of PRs to review in the repo and this one might be further back in the queue; I just want to check in while the approach I took here is still fresh in my mind.

MrMino:
I don't think it does, although I haven't read it through completely. It's a question of when I or any other maintainer finds a chunk of free time large enough to review it properly. I can only accept if I have a reasonable degree of certainty, that I understand _all_ implications of this change.

That said, I think it's very useful and I do want to have this, even for the sake of my own convenience.

Don't worry about the downstream test failures, happens all the time. See #13695.

joelostblom:
Thanks for the reply! I am happy to hear you think this functionality would be useful so thanks for sharing that. I totally understand that it is difficult to find enough time to review all PRs that come in to avoid unintended regressions and side effects.

krassowski:
https://github.com/ipython/ipython/pull/13745 will allow to significantly simplify this PR and remove the special case by specifying `ordered=True` in `python_func_kw_matcher`.

This PR will still be needed to add special handling of jedi completions.

joelostblom:
Just a note that the suggestion I gave to enable param first sorting in Jedi was closed https://github.com/davidhalter/jedi/pull/1855#issuecomment-1311885108, so if we want to add that in the future, some new approach would need to be taken (this doesn't affect the content of this PR, just follow up PRs trying to add sorting with parameters first for jedi completions)

krassowski:
@joelostblom would you have time to reapply this PR using `ordered=True` suggestion?

joelostblom:
Thanks for the ping @krassowski . Unfortunately, I'm currently have too many things on my plate and since it was a while ago I worked on this PR last I would need to sit down for some time and refamiliarize myself. I am more than happy for someone else to take it over and continue what I started.

Carreau:

PR #13680, opened 2022-05-24
Use relative filename in latex and dvips commands of latex_to_png_dvipng

### LINKED ISSUE #13678: latex_to_png fails on windows
The `latex_to_png` command fails in the method `latex_to_png_dvipng`. The reason is the temporary file that is generated, contains a `~` in the path, which is not accepted by the latex command.

The temporary file generated is ` 'C:\\Users\\EENDEB~1\\AppData\\Local\\Temp\\tmpxefqozm3\\tmp.tex'`
The command that fails is:
```
            subprocess.check_call(
                ["latex", "-halt-on-error", tmpfile,
                cwd=workdir, stdout=devnull, stderr=devnull)
```
Locally I can fix this by changing the command to
```
            subprocess.check_call(
                ["latex", "-halt-on-error",  os.path.split(tmpfile)[-1]],
                cwd=workdir, stdout=devnull, stderr=devnull)
```
(only the last part is required, since the `cwd` is also set).

To make this work some other calls need to be modified as well

eendebakpt:
This prevents problems with temporary directories with special characters in them.

Fixes #13678

### DIFF
--- a/IPython/lib/latextools.py
+++ b/IPython/lib/latextools.py
@@ -146,11 +146,11 @@ def latex_to_png_dvipng(s, wrap, color='Black', scale=1.0):
         return None
     try:
         workdir = Path(tempfile.mkdtemp())
-        tmpfile = workdir.joinpath("tmp.tex")
-        dvifile = workdir.joinpath("tmp.dvi")
-        outfile = workdir.joinpath("tmp.png")
+        tmpfile = "tmp.tex"
+        dvifile = "tmp.dvi"
+        outfile = "tmp.png"
 
-        with tmpfile.open("w", encoding="utf8") as f:
+        with workdir.joinpath(tmpfile).open("w", encoding="utf8") as f:
             f.writelines(genelatex(s, wrap))
 
         with open(os.devnull, 'wb') as devnull:
@@ -181,7 +181,7 @@ def latex_to_png_dvipng(s, wrap, color='Black', scale=1.0):
                 stderr=devnull,
             )
 
-        with outfile.open("rb") as f:
+        with workdir.joinpath(outfile).open("rb") as f:
             return f.read()
     except subprocess.CalledProcessError:
         return None
--- a/docs/source/whatsnew/pr/latex_rendering_relative_files.rst
+++ b/docs/source/whatsnew/pr/latex_rendering_relative_files.rst
@@ -1,0 +1,6 @@
+Relative filenames in Latex rendering
+=====================================
+
+The input and output file arguments to `latex` and `dvipis` are files relative to the current working directory.
+This solves a problem where the current working directory contains characters that are not handled properly by `latex` and `dvips`.
+
--- a/docs/source/whatsnew/pr/latex_rendering_relative_files.rst
+++ b/docs/source/whatsnew/pr/latex_rendering_relative_files.rst
@@ -1,6 +1,7 @@
 Relative filenames in Latex rendering
 =====================================
 
-The input and output file arguments to `latex` and `dvipis` are files relative to the current working directory.
+The `latex_to_png_dvipng` command internally generates input and output file arguments to `latex` and `dvipis`. These arguments are now generated as relative files to the current working directory instead of absolute file paths.
 This solves a problem where the current working directory contains characters that are not handled properly by `latex` and `dvips`.
+There are no changes to the user API.
 


MrMino:
Similar to one of the earlier PR of yours, could you please add a what's-new entry that describes the change to users?

https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr

eendebakpt:
@MrMino Could you review again?

docs/source/whatsnew/pr/latex_rendering_relative_files.rst
```diff
@@ -0,0 +1,6 @@
+Relative filenames in Latex rendering
+=====================================
+
+The input and output file arguments to `latex` and `dvipis` are files relative to the current working directory.
```
MrMino, line 4: I had a hard time understanding this before I looked through the whole module. I need to nitpick this, sorry.

I don't think this explains the change to the user particularly well. The user doesn't necessarily see `latex` nor `dvipis` commands. For the user, the change is in terms of the API they call.

Could you please explain the change in terms of the API that changes, as opposed to its internals?

IPython/lib/latextools.py
```diff
@@ -146,11 +146,11 @@ def latex_to_png_dvipng(s, wrap, color='Black', scale=1.0):
         return None
     try:
         workdir = Path(tempfile.mkdtemp())
-        tmpfile = workdir.joinpath("tmp.tex")
-        dvifile = workdir.joinpath("tmp.dvi")
-        outfile = workdir.joinpath("tmp.png")
+        tmpfile = "tmp.tex"
+        dvifile = "tmp.dvi"
```
MrMino, line 150: Does this mean that this file is no longer "temporary"? I.e., it doesn't get removed by the `shutil.rmdir` in line 189?

MrMino: I think it's the case of missing `workdir.joinpath(...)` in line 175.

MrMino (commented):
Come to think of it, maybe there's actually a way of making `latex` accept the tilde?

Have a look at https://tex.stackexchange.com/questions/425954/tilde-in-a-image-path.

Have you tried to do a `str.replace("~", "\string~")` on that path when passing it to `latex`?

eendebakpt:
> Come to think of it, maybe there's actually a way of making `latex` accept the tilde?
> 
> Have a look at https://tex.stackexchange.com/questions/425954/tilde-in-a-image-path.
> 
> Have you tried to do a `str.replace("~", "\string~")` on that path when passing it to `latex`?

The link is about using a title inside a latex document, this PR is about a title in a command line argument to latex.  I tried to put the argument inside `""`, escape the ~ (with `\` or `\string`), but did not manage to get it working that way.

docs/source/whatsnew/pr/latex_rendering_relative_files.rst
eendebakpt: For the user there are no changes to the API at all, so we could also skip the whatsnew. 

I changed the text to make clear this is an internal change.

IPython/lib/latextools.py
eendebakpt: All files generated are still temporary. The `dvifile` from line 175 is generated by `dvips` in the `workdir` (it not, line 184 would fail) and is removed at line 189.

MrMino:
I understand, but from my experience with `latex`, IIRC it parses some of its arguments as if they are written _in latex_. Unfortunately I don't have the environment ready to check this, so that's why I'm asking about it. 

I'll check the code, and if there truly is no way to achieve that, we'll merge.

eendebakpt:
> I understand, but from my experience with `latex`, IIRC it parses some of its arguments as if they are written _in latex_. Unfortunately I don't have the environment ready to check this, so that's why I'm asking about it.
> 
> I'll check the code, and if there truly is no way to achieve that, we'll merge.

@MrMino Any update on this? On my system
```
In [78]: import tempfile; tempfile.mkdtemp()
Out[78]: 'C:\\Users\\EENDEB~1\\AppData\\Local\\Temp\\tmpnzyikr5c'
```
The `~` in the output should not only be parsed properly by latex, but somehow the name 'C:\\Users\\EENDEB~1\\AppData\\Local\\Temp\\tmpnzyikr5c' should also be expanded to `'C:\\Users\\EENDEBAKPT\\AppData\\Local\\Temp\\tmpnzyikr5c'.

Carreau:
Thanks, let's try.

PR #13683, opened 2022-05-28
Correct the debugger_cls() call to self.debugger_cls() call in TBTool

genelkim:
…s.__init__() so that it doesn't crash when the optional argument, debugger_cls, is unspecified.


MrMino:
Indeed, I don't think we wanted to `None()` there. Thanks!

Carreau:
@meeseeksdev backport to 7.x

Issue #13672, opened 2022-05-12
Documentation of Demo has python2 code examples


albertogomcas:
https://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.demo.html

The example of demo.py uses print statements instead of print function, which does not work in python 3.x

yashlad681:
Hey @albertogomcas mind if I work on this issue?

MrMino:
@yashlad681 go for it if

MrMino:
I don't think it was. As long as there's no PR sent - anyone is welcome to do so :).

PR #13689, opened 2022-06-02
Documentation uses python2 for a code example instead of python3

AxolotlTears:
https://github.com/ipython/ipython/issues/13672
Just a quick fix that converts `print ' '`  -->  `print(' ')`

Build:
image


MrMino:
Check failures are unrelated.

Thanks!

PR #13693, opened 2022-06-09
Fix: move quiet handling to interactiveshell

agoose77:
I don't know a huge amount about the IPython internals, so this might not be the right fix.

Currently, `DisplayHook` uses the shell history to determine whether the current cell should be quiet or not. This means that if the user executes the cell with `store_history=False`, the display hook will look to a _previous_ cell to determine whether to render the current cell's output.

This PR fixes this problem by explicitly computing and setting a new `is_quiet` attribute on `DisplayHook`. I would want to make this private, or at least, make it clear that only `InteractiveShell` should set this. I'd be keen to know what the most IPythonic approach would be here.

I also don't know if there are any wider repercussions of this, e.g. if the display hook is invoked outside of `run_cell_async`. I would naively assume that this is fine?

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -58,6 +58,8 @@ class DisplayHook(Configurable):
         to_user_ns = {'_':self._,'__':self.__,'___':self.___}
         self.shell.user_ns.update(to_user_ns)
 
+        self.is_quiet = False
+
     @property
     def prompt_count(self):
         return self.shell.execution_count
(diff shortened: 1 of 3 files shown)


agoose77:
This is a commandline program which demonstrates the issue. You must pass the path to a kernel connection file as the only argument;

```python
import argparse
from jupyter_client.blocking.client import BlockingKernelClient


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("conn")
    args = parser.parse_args()

    client = BlockingKernelClient()
    client.load_connection_file(args.conn)
    client.start_channels()
    client.wait_for_ready()

    print("The following won't display 'Hello world':")
    # Note the trailing `;` in the first call to `execute-interactive`
    client.execute_interactive("None;", store_history=True)
    client.execute_interactive("'Hello world'", store_history=False)

    print("The following will display 'Hello world':")
    client.execute_interactive("None", store_history=True)
    client.execute_interactive("'Hello world'", store_history=False)
```

agoose77:
@MrMino is there anything I can do here to move this along?

Carreau:

PR #13706, opened 2022-07-14
Add setting to disable venv warning

### LINKED ISSUE #13457: IPython-8 warns about virtualenv created by pyenv
Hi, is there a way to disable the virtualenv warning?

I use pyenv, and every time IPython-8.0 starts from a virtual environment that's based on pyenv's miniconda, it shows this warning:

```bash
# Activate virtualenv via 'pyenv activate my_project_env' or 'pyenv activate my_base_python_env/envs/my_project_env'.
...

# Then, start ipython (installed in that virtual env).
$ ipython
/Users/xxx/.pyenv/versions/rubix-mds/lib/python3.9/site-packages/IPython/core/interactiveshell.py:802: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
  warn(
Python 3.9.7 (default, Sep 16 2021, 08:50:36) 
IPython 8.0.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import sys
   ...: 
   ...: sys.executable
Out[1]: '/Users/xxx/.pyenv/versions/rubix-mds/bin/python'

In [2]: import os
   ...: 
   ...: os.environ["VIRTUAL_ENV"]
Out[2]: '/Users/xxx/.pyenv/versions/miniconda3-latest/envs/rubix-mds'
```

This only happens for virtualenv based on another pyenv's miniconda3-latest. No warning shown with virtualenv based on another pyenv's "python version".

```bash
$HOME
`-- versions
    |-- miniconda3-latest
    |   `-- envs
    |       `-- rubix-mds
    `-- rubix-mds -> $HOME/.pyenv/versions/miniconda3-latest/envs/rubix-mds
```

I'd like to know if there's a way to silent this particular warning?

Thank you.

cbrnr:
Fixes #13457.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -481,7 +481,7 @@ class InteractiveShell(SingletonConfigurable):
 
     warn_venv = Bool(
         True,
-        help='Warn if running in a virtual environment with no IPython installed (so IPython from the global environment is used).'
+        help="Warn if running in a virtual environment with no IPython installed (so IPython from the global environment is used)."
     ).tag(config=True)
 
     # TODO: this part of prompt management should be moved to the frontends.
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -479,6 +479,11 @@ class InteractiveShell(SingletonConfigurable):
         """
     ).tag(config=True)
 
+    warn_venv = Bool(
+        True,
+        help="Warn on startup if running inside a virtual environment."
+    ).tag(config=True)
+
     # TODO: this part of prompt management should be moved to the frontends.
     # Use custom TraitTypes that convert '0'->'' and '\\n'->'\n'
     separate_in = SeparateUnicode('\n').tag(config=True)
@@ -847,11 +852,11 @@ class InteractiveShell(SingletonConfigurable):
                     p_ver = re_m.groups()
 
             virtual_env = str(virtual_env_path).format(*p_ver)
-
-        warn(
-            "Attempting to work in a virtualenv. If you encounter problems, "
-            "please install IPython inside the virtualenv."
-        )
+        if self.warn_venv:
+            warn(
+                "Attempting to work in a virtualenv. If you encounter problems, "
+                "please install IPython inside the virtualenv."
+            )
         import site
         sys.path.insert(0, virtual_env)
         site.addsitedir(virtual_env)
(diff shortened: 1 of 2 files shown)


cbrnr:
To me it looks like the two failing tests are unrelated, but I might be wrong.

IPython/core/interactiveshell.py
```diff
@@ -479,6 +479,10 @@ def input_splitter(self):
         """
     ).tag(config=True)
 
+    warn_venv = Bool(
+        True, help="Warn on startup if running inside a virtual environment."
```
MrMino, line 483: That's not what happens. The warning is shown if virtualenv is active but IPython runs _outside_ of it.

cbrnr: That's what I meant – I agree the wording is ambiguous. What about:
"Warn if a virtual environment is active, and IPython is installed outside of it."

MrMino: Oh, I get it now. Could you please rephrase this help message so that it is clear?

This would confuse people who a re looking at it from a PoV of trying to run outside of a virtualenv.

MrMino (changes requested):
Please also add a [what's-new entry](https://github.com/ipython/ipython/tree/master/docs/source/whatsnew/pr) for this.

IPython/core/interactiveshell.py
cbrnr: Is this OK?

"Warn if running in a virtual environment which does not have IPython installed (so IPython from the global environment is used)."

cbrnr: I ended up including this text:

'Warn if running in a virtual environment with no IPython installed (so IPython from the global environment is used).'

I also added the what's new entry.

IPython/core/interactiveshell.py
```diff
@@ -479,6 +479,11 @@ def input_splitter(self):
         """
     ).tag(config=True)
 
+    warn_venv = Bool(
+        True,
+        help='Warn if running in a virtual environment with no IPython installed (so IPython from the global environment is used).'
```
MrMino, line 484: You need to fix the formatting otherwise the pipeline won't pass:

```suggestion
        help="Warn if running in a virtual environment with no IPython installed (so IPython from the global environment is used)."
```

cbrnr:
@MrMino all style issues fixed, the one failing test seems still unrelated.

MrMino (approved):

jamilraichouni:
Hi @MrMino , there seems to be an issue with this change.
See here, please: https://github.com/ipython/ipython/issues/13875

All best,
Jamil

PR #13707, opened 2022-07-16
Fix #13700, consistently use the original name in error message

### LINKED ISSUE #13700: Given inexistent file X, "%run X" reports that it's unable to open "X.py"
... which confuses users. See #13542.

suzaku:
If there's no file found for the given name with or without the ".py" suffix, the message should consistently use the original name to avoid confusion.


Carreau:
Thanks.

PR #13708, opened 2022-07-16
Fix #13654, improve performance of auto match for quotes

### LINKED ISSUE #13654: `auto_match` slow and hogs CPU in some cases
To reproduce in IPython 8.2, paste `function_with_long_nameeee('arg`, and add the closing quotes at the end.
The slowness increases when the prefix is longer (in this case dominated by the function name `function_with_long_nameeee`), so if you don't notice it, try increasing it).

## Steps to reproduce with docker

1. Build the image and start a container:

```sh
build_oci_img() {
  docker build "$@" - <<EOF
  FROM python:3.8-slim
  RUN pip install ipython==8.2
EOF
}

build_oci_img
docker run --rm --interactive --tty "$(build_oci_img -q)" ipython -i -c 'get_ipython().auto_match = True'

```

2. Paste `function_with_long_nameeee('arg` and then type another `'` (single quote)

suzaku:
As pointed out in #13654, auto matching of quotes may take a long time
if the prefix is long.
To be more precise, the longer the text before the first quote, the
slower it is.
This is all caused by the regex pattern used: `r'^([^"]+|"[^"]*")*$'`,
which I suspect is O(2^N) slow.

```python
In [1]: text = "function_with_long_nameeee('arg"

In [2]: import re

In [3]: pattern = re.compile(r"^([^']+|'[^']*')*$")

In [4]: %timeit pattern.match(text)
10.3 s ± 67.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit pattern.match("1'")
312 ns ± 0.775 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops
    each)

In [6]: %timeit pattern.match("12'")
462 ns ± 1.95 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops
    each)

In [7]: %timeit pattern.match("123'")
766 ns ± 6.32 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops
    each)

In [8]: %timeit pattern.match("1234'")
1.59 µs ± 20.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops
  each)
```

But the pattern we want here can actually be detected with a Python
implemention in O(N) time.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -113,7 +113,7 @@ def create_ipython_shortcuts(shell):
             c = buf[i]
             if c == quote:
                 paired = not paired
-            elif c == '\\':
+            elif c == "\\":
                 i += 1
             i += 1
         return paired
@@ -127,16 +127,19 @@ def create_ipython_shortcuts(shell):
             return _preceding_text_cache[pattern]
 
         if callable(pattern):
+
             def _preceding_text():
                 app = get_app()
                 before_cursor = app.current_buffer.document.current_line_before_cursor
                 return bool(pattern(before_cursor))
+
         else:
             m = re.compile(pattern)
 
             def _preceding_text():
                 app = get_app()
-                return bool(m.match(app.current_buffer.document.current_line_before_cursor))
+                before_cursor = app.current_buffer.document.current_line_before_cursor
+                return bool(m.match(before_cursor))
 
         condition = Condition(_preceding_text)
         _preceding_text_cache[pattern] = condition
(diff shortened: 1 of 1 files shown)


suzaku:
BTW, I wonder why we use `_preceding_text_cache`, `re.compile` already has its own cache, doesn't it?

Carreau:
Apologies for the delay, 

I think that conflicts with #13713 that I already merge, sorry about that. 
I'll see if I can port your improvement, unless you want to take a shot at it.

suzaku:
@Carreau I'll see if I can resolve the conflict.

suzaku:
@Carreau Conflict resolved.

suzaku:
Hi @Carreau , could you please take a look so that I don't have to resolve conflicts again?

Carreau (approved):

PR #13713, opened 2022-07-24
Improve auto_match for triple quotes and escaped quotes

### LINKED ISSUE #13654: `auto_match` slow and hogs CPU in some cases

lucymcphail:
Automatically insert matching triple quotes, and recognise triple-quoted strings and escaped quotes when matching.

Also fixes #13654.

### DIFF
--- a/IPython/terminal/shortcuts.py
+++ b/IPython/terminal/shortcuts.py
@@ -140,6 +140,18 @@ def create_ipython_shortcuts(shell):
         _following_text_cache[pattern] = condition
         return condition
 
+    @Condition
+    def not_inside_unclosed_string():
+        app = get_app()
+        s = app.current_buffer.document.text_before_cursor
+        # remove escaped quotes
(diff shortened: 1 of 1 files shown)


Carreau:
Let's try this !

Issue #13666, opened 2022-05-02
Error using HistoryAccessor.get_last_session_id


seanbreckenridge:
On the newest version of ipython:

```
$ pip list | rg -i ipython
ipython                       8.3.0
```

Trying to get the last session id from the history accessor for any ipython history db crashes for me:

```
>>> from IPython.core.history import HistoryAccessor
>>> hist = HistoryAccessor(hist_file="")  # use current database
>>> hist.get_last_session_id()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
  File "/home/sean/.local/lib/python3.10/site-packages/IPython/core/history.py", line 77, in catch_corrupt_db
    return f(self, *a, **kw)
  File "/home/sean/.local/lib/python3.10/site-packages/IPython/core/history.py", line 340, in get_last_session_id
    for record in self.get_tail(n=1, include_latest=True):
  File "/usr/lib/python3.10/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
  File "/home/sean/.local/lib/python3.10/site-packages/IPython/core/history.py", line 77, in catch_corrupt_db
    return f(self, *a, **kw)
  File "/home/sean/.local/lib/python3.10/site-packages/IPython/core/history.py", line 374, in get_tail
    (self.session_number, n),
AttributeError: 'HistoryAccessor' object has no attribute 'session_number'. Did you mean: 'section_names'?
```

If I downgrade to anything before 8.3.0, it works as expected:

```
>>> from IPython.core.history import HistoryAccessor
>>> hist = HistoryAccessor()
>>> hist.get_last_session_id()
2750
```

Also fails on [this CI](https://github.com/seanbreckenridge/HPI/runs/6251660329?check_suite_focus=true) which has some useful pytest context (running [this test](https://github.com/seanbreckenridge/HPI/blob/e17d101fe52d6acdaaeab9a6495a190ea61b5775/tests/test_ipython.py#L21-L28)), using a [custom ipython history database](https://github.com/seanbreckenridge/HPI/blob/master/tests/testdata/ipython.sqlite) instead of leaving `hist_file` empty

MrMino:
My `git bisect` tells me its from dc5bcc1c50892a5128fcf128af28887226144927, i.e. #13659.

@Carreau any ideas why this happened?

[...]

PR #13719, opened 2022-08-06
Fix HistoryAccessor.get_tail bug (#13666)

balval:
Current implementation of get_tail in HistoryAccessor assumes context
present only in subclass , so it's moved there and the old
implementation is restored.


Carreau:
Thanks.

PR #13725, opened 2022-08-11
keep autocomplete results stable in multi-thread access

richard-fung-db:
IPython can be used in a GUI where a notebook cell is running while autocomplete is getting invoked on another thread. The running cell can change the iterator of the namespace or global_namespace while the autocomplete invocation is enumerating over the dictionary of results. The resulting error message from autocomplete is `RuntimeError: dictionary size changed during iteration`.

This change proposes to get a copy of the keys() so that can be enumerated.

Example code that causes the issue:
```
import time

glbl = globals()
count = 0
time_limit = 50
start = time.time()
end = start + time_limit

while time.time() < end:
  glbl[f"global_var_{count}"] = None
  count += 1
```

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -677,15 +677,15 @@ class Completer(Configurable):
         n = len(text)
         for lst in [keyword.kwlist,
                     builtin_mod.__dict__.keys(),
-                    self.namespace.keys(),
-                    self.global_namespace.keys()]:
+                    list(self.namespace.keys()),
+                    list(self.global_namespace.keys())]:
             for word in lst:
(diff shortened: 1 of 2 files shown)


richard-fung-db:
You can fix this by running ``pip install black[jupyter]``

jasongrout:
As well as:

```
```

jasongrout (approved):
Thanks!

jasongrout:
@Carreau - could we add this to the 8.5 milestone, or a bugfix release if that would be coming earlier than 8.5?

Carreau:
Yeah, that seem reasonable, I'll merge that and make a release soon.

Issue #12411, opened 2020-06-30
Autoreload fails when reloading frozen data classes


eehusky:
I have autoreload turned on and when reloading a data class with the frozen attribute enabled i get the following error.

```
[autoreload of MotionPlanner.TrajectoryGenerator failed: Traceback (most recent call last):
  File "C:\Users\sparky\anaconda3\lib\site-packages\IPython\extensions\autoreload.py", line 245, in check
    superreload(m, reload, self.old_objects)
  File "C:\Users\sparky\anaconda3\lib\site-packages\IPython\extensions\autoreload.py", line 410, in superreload
    update_generic(old_obj, new_obj)
  File "C:\Users\sparky\anaconda3\lib\site-packages\IPython\extensions\autoreload.py", line 347, in update_generic
    update(a, b)
  File "C:\Users\sparky\anaconda3\lib\site-packages\IPython\extensions\autoreload.py", line 317, in update_class
    update_instances(old, new)
  File "C:\Users\sparky\anaconda3\lib\site-packages\IPython\extensions\autoreload.py", line 280, in update_instances
    ref.__class__ = new
  File "<string>", line 4, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field '__class__'
]
```

It appears that the module does in fact reload successfully if a re-run a cell (I am using IPython via JupyterLab)

```
{'commit_hash': '2922831ac',
 'commit_source': 'installation',
 'default_encoding': 'cp1252',
 'ipython_path': 'C:\\Users\\sparky\\anaconda3\\lib\\site-packages\\IPython',
 'ipython_version': '7.15.0',
 'os_name': 'nt',
 'platform': 'Windows-7-6.1.7601-SP1',
 'sys_executable': 'C:\\Users\\sparky\\anaconda3\\python.exe',
 'sys_platform': 'win32',
 'sys_version': '3.7.6 (default, Jan  8 2020, 20:23:39) [MSC v.1916 64 bit '
                '(AMD64)]'}
```

This is an example of the class definition.

```python
@dataclass(init=True, repr=True, eq=True, frozen=True, order=True)
class Constraints(object):
    q0: float
    q1: float
    v0: float
    v1: float
    j_max: float
    a_max: float
    v_max: float
    duration: float = None
```

It works eventually but figured it would be worth bringing it up.  I'm not 100% certain this is an IPython problem (could be Jupyter?).

ottumm:
[...]

Issue #12185, opened 2020-03-11
Unable to autoreload @dataclass(frozen=True)


Techcable:
I think you can bypass the FrozenInstanceError by directly invoking object.__setattr__

````
[autoreload of pitch_detection failed: Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/IPython/extensions/autoreload.py", line 245, in check
    superreload(m, reload, self.old_objects)
  File "/usr/lib/python3.8/site-packages/IPython/extensions/autoreload.py", line 410, in superreload
    update_generic(old_obj, new_obj)
  File "/usr/lib/python3.8/site-packages/IPython/extensions/autoreload.py", line 347, in update_generic
    update(a, b)
  File "/usr/lib/python3.8/site-packages/IPython/extensions/autoreload.py", line 317, in update_class
    update_instances(old, new)
  File "/usr/lib/python3.8/site-packages/IPython/extensions/autoreload.py", line 280, in update_instances
    ref.__class__ = new
  File "<string>", line 4, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field '__class__'
]
````

**Comment by Carreau** [comment]
Hummm... If you know what you are doing then we'll happily take a Pull Request.

**Comment by kvjrhall** [comment]
I have observed this in vscode, in particular.

Let us say that we have a frozen dataclass in some user-specified module `usermodule` (I haven't tested with any 3rd party packages) along with other code currently in development.

```
from dataclasses import dataclass

@dataclass
class some_frozen_class:
    a: int

def algorithm(a: int) -> int:
    return a + 2
```

In a notebook cell we have the following

```
from usermodule import algorithm

display(algorithm(2))
```

Next we modify algorithm (or any code in that python file)

```
from dataclasses import dataclass

@dataclass
class some_frozen_class:
    a: int

def algorithm(a: int) -> int:
    return a + 3   # a change
```

...and re-run the cell. I've been seeing two things: (1) the autoreload error as reported by @Techcable; and (2) the cell continues to run despite the error, with unpredictable behavior (i.e., code may or may not be changed, etc). Restarting the python interpreter results in the error disappearing and code running as expected for the first run.

PR #13732, opened 2022-08-16
Let %autoreload work on modules with frozen dataclasses

hhoppe:
See #12411 and #12185.

The problem is that a frozen `dataclasses.dataclass` overrides the `__setattr__()` method, so updating its `__class__` member requires going through the base `object` class.

A simple way to reproduce the problem and see that the fix works is as follows.

Create a file `./module/__init__.py` containing:
```python
__version__ = 1
import dataclasses

@dataclasses.dataclass(frozen=True)
class Class:
  a: int

instance = Class(1)
```

Then create a notebook containing two cells:
```python
%load_ext autoreload
%autoreload 2
import module

# %%
print(module.__version__)
```

Run the notebook; edit the `__init__.py` file to contain `__version__ = 2`, then rerun the second notebook cell (`print`).
Without the fix, an exception is raised.

### DIFF
--- a/IPython/extensions/autoreload.py
+++ b/IPython/extensions/autoreload.py
@@ -300,7 +300,7 @@ def update_instances(old, new):
 
     for ref in refs:
         if type(ref) is old:
-            ref.__class__ = new
+            object.__setattr__(ref, "__class__", new)
 
 
 def update_class(old, new):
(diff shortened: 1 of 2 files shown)


hhoppe:
@Carreau 
The linter is failing in a file which I did not modify.  What should I do?

Carreau:
I'll take care of fixing it.

hhoppe:
Wonderful; thanks!

hhoppe:
Sorry I was confused and thought that my PR had been merged in.  So re-opening it.

PR #13733, opened 2022-08-18
Broken link to "Edit on Github" in ipython documentation #11919

pratikbhadane24:


Carreau:
Thanks but this is incorrect. Those files should not be added to git, they are generated. 
Best course of action would be to disable view on github on those pages.

PR #13743, opened 2022-09-03
Added additional entrypoint script.

UltimateLobster:
Added a third entrypoint to use python's minor version as well.

This can help when testing out differences of python versions. One could easily open "ipython3.10" and test it's differences with "ipython3.8".


Carreau:
Seem reasonable. Thanks

Issue #13486, opened 2022-01-24
bgcolor highlighting of ast nodes in traceback may be unreadable


alexandru-dinu:
The commit https://github.com/ipython/ipython/commit/3026c205487897f6874b2ff580fe0be33e36e033 introduced a change with a hard-coded `bg:ansiyellow` color.

Unfortunately, for some terminal color schemes, white-on-yellow may be unreadable, for example:
![image](https://user-images.githubusercontent.com/14110183/150824435-dd65d1ff-7b3c-4b29-acb1-2a774546f820.png)

Changing to `bg:ansired` looks better:
![image](https://user-images.githubusercontent.com/14110183/150831775-1e4e174d-145e-4760-abcd-07ef698ce153.png)

Is it possible to make this option configurable?

Thank you!

Carreau:
Duplicate #13446

amerlyq:
Linked issue is locked, and as I understood it mostly focused on choosing new default color/style instead of exposing Token.ExecutingNode for style_overrides config option.

Mentioning that "bg:ansiyellow" is being only used as fallback in `stack_data.style_with_executing_node(style, "bg:ansiyellow")` is wrong. Because its value is used as override:
```
    class NewStyle(style):
        for_executing_node = True

        styles = {
            **style.styles,
            **{
                k.ExecutingNode: v + " " + modifier
                for k, v in style.styles.items()
            }
        }
```
Therefore we really don't have a way to modify it from user config -- only through modifiying sources itself.

kanbara:
I have spent several hours reading the source and issues trying to override these colours. It's pretty bad for dark backgrounds, and I finally have the normal colours sorted, but I would really really _really_ like the ability for the traceback to be customised-- especially the node colour. It's unusable on dark backgrounds and clashes with everything else.

DanielO:
Being able to customise it would be ideal as currently is illegible on a dark terminal unless you have a muted yellow. Unfortunately I don't know how it's all plumbed together so I am not sure how to make the change.

For iTerm2 at least using anisred  for the BG colour is quite legible for both light and dark (when editing the source).

If I change the line to:
                style = stack_data.style_with_executing_node(style, style.styles[pygments.token.Generic.Error])

(And add 'import pigments.token' up the top) then it uses the existing error style rather than hard coding it - perhaps that is a suitable solution.

0xquad:
Has there been plans to fix this issue?

[...]

Issue #13446, opened 2022-01-12
Should VerboseTB.get_records have hardcoded background colour?


zx64:
Hello, when upgrading to IPython 8, I noticed that the error highlighting feature of tracebacks has poor contrast with my terminal colour scheme (zenburn):
![bg ansiyellow](https://i.gyazo.com/edfe5ea05389d8799f924650b3c42760.png)

While tracking this down, I noticed that the background colour chosen is [hardcoded](https://github.com/ipython/ipython/blob/c0abea7a6dfe52c1f74c9d0387d4accadba7cc14/IPython/core/ultratb.py#L830) and I think I only started to notice this issue with 3026c20 as the previous value didn't result in any background colour being set.

If I change the hardcoded parameter to `"bg:ansired"`, the contrast is greatly improved for me, but I'm sure that's not a good choice for other users and themes:
![bg ansired](https://i.gyazo.com/461c22c752027bb0a88692519eb7e52a.png)

Being able to customise this would be a nice short term fix, but that might not be the best long term approach of requiring some users to set a config flag to get readable tracebacks.

Carreau:
Actually the line you point to is the fallback (and that should map to your terminal style so you might be able to tweek it in your terminal preference), I believe the color theme can be configured IPython/utils/PyColorize.py, but it's true that color for the highlighting might be missing.

It would be good to have a more generic way of updating styles in general.

caleboverman:
I'm having the same issue with the `solarized-dark` palette. I've confirmed that changing my terminal color mapping  alters the highlighting color.  It does seem less than ideal to have to remap the terminal yellow to a different color as this affects other places the color is used.

Definitely not a color expert but this feels like it might be a common issue with dark backgrounds which use a lighter font and are tough to read against a yellow background.

aidos:
Have just been down this rabbit-hole myself. I've been through the builtin IPython themes against all the standard item2 ones and they all have the same contrast issue so it feels that yellow background on white text would seem to be a bad default.

Would be interested to see what it looked like on the terminals where it was being developed for comparison.

Carreau:
Sure thing, pencil-light and pencil-dark (iterm 2, mac) . 
<img width="538" alt="Screen Shot 2022-01-23 at 19 59 35" src="https://user-images.githubusercontent.com/335567/150693693-f7d7dc05-58e4-4e14-a24b-e712d6ee47a2.png">
[...]

PR #13756, opened 2022-09-12
Quick hack to make tb highlighting configurable.

Carreau:
For example put

    from IPython.core.ultratb import VerboseTB
    VerboseTB._tb_highlight = "bg:#039dfc"

In your ipython_config.py

Mitigate #13446 and #13486

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -610,6 +610,8 @@ class VerboseTB(TBTools):
     traceback, to be used with alternate interpreters (because their own code
     would appear in the traceback)."""
 
+    _tb_highlight = "bg:ansiyellow"
+
     def __init__(
         self,
         color_scheme: str = "Linux",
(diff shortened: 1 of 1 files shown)


Carreau:
Note that this is only. A stop-gap as Python 3.11 is introducing traceback annotations, and also now annotate nodes (but underline them), so the code that does all this may need to be re-written. 

Python exception now also have exception groups, and annotations, so it might be quite a bit of work, and I'd love for someone to find the time to do so, it is jus likely not going to be me.

Carreau:
@meeseeksdev tag "help wanted"

simon-ging:
Note that the variable was renamed from `_tb_highlight` to `tb_highlight`, so the fix now must update the new variable, e.g.:

`from IPython.core import ultratb;ultratb.VerboseTB.tb_highlight = 'bg:ansired';`

ili3p:
The file is in:
` ~/.ipython/profile_default/startup/ipython_config.py`

Issue #13622, opened 2022-04-04
IPython 8.2.0 ERROR with %cpaste -q


ljmc-github:
I recently updated my workflow to use pyenv, which brought along an update to all my pip installed tools.

IPython was updated from 8.1.1 to 8.2.0.

Since then running a block with `%cpaste -q` which I use a lot causes an ERROR:

```
ERROR! Session/line number was not unique in database. History logging moved to new session 3
```

I compared with my old install and with a downgraded version, both work as expected: 

1. Old install using brew and pip
```
Python 3.10.2 (main, Feb  2 2022, 06:19:27) [Clang 13.0.0 (clang-1300.0.29.3)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %cpaste -q
i = 1
print(i)
--
1

In [2]:
```

2. New install in pyenv and pip
```
Python 3.9.11 (main, Apr  4 2022, 12:07:49)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %cpaste -q
j = 2
print(j)
--
2

ERROR! Session/line number was not unique in database. History logging moved to new session 3
In [3]:
```

3. Downgraded to 8.1.1
```
Python 3.9.11 (main, Apr  4 2022, 12:07:49)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %cpaste -q
i = 1
print(i)
--
1

In [2]:
```
 
There was a difference in the history.sqlite database, both 8.1.1 version has a single row for the `%cpaste -q`, the 8.2.0 was also showing the `j = 2` and `print(j)` on a second row, but I'm unable to find this output at this time. The overincrementing can be seen in the jump from prompt 1 straight to 3.

Carreau:
Yeah, I know what this is, there was a recent modification so that the content of `%paste` instead of `%paste` itself would be put in history. I probably forgot about cpaste.

Paul-Aime:
I don't know if that helps, but it also happens with `%paste`.

- `%paste` : overincrementing
- `%paste -q` : overincrementing + history database error
- `%paste -r`: no issue

```python
In [1]: %paste
print('hi')

## -- End pasted text --
hi

In [3]: %paste -q
hi

[...]

PR #13763, opened 2022-09-19
fixed cpaste session error in #13622

### LINKED ISSUE #13585: %paste + history

CosmosAtlas:
In the commit that fixed #13585 , the function that both `%cpaste` and `%paste` depends on is hard coded to `store_history`.

However, with `%cpaste` sharing the same code, it causes #13622 , which has also received some discussion at other places that relies on it such as:

https://github.com/jpalardy/vim-slime/issues/323

### DIFF
--- a/IPython/terminal/magics.py
+++ b/IPython/terminal/magics.py
@@ -41,7 +41,7 @@ class TerminalMagics(Magics):
     def __init__(self, shell):
         super(TerminalMagics, self).__init__(shell)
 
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, much appreciated.

Issue #11606, opened 2019-02-18
Generating Docs with Sphinx and IPython directive throws error if path has whitespace


MaxVanDeursen:
In the `pandas` repository, a issue was reported on the documentation generation (https://github.com/pandas-dev/pandas/issues/22836). When the path to the doc source files contains white spaces (e.g. `/home/max/contains space/source`), a long list of UsageErrors is thrown: 
```
UsageError: %bookmark: too many arguments
UsageError: %bookmark -d: Can't delete bookmark 'ipy_savedir'
```

Upon further inspection of this issue, I have been able to create a minimum example (https://github.com/pandas-dev/pandas/issues/22836#issuecomment-464770002) which throws the same error, using the generated config file by `sphinx-quickstart` along with the extension, found in Pastebin [here](https://pastebin.com/uRGDxh17), as well as the following `index.rst`:
```
.. ipython:: python

    answer = 42
```

Without the interactive section in the documentation, the error is not thrown, leading to believe that the `IPython.sphinxext.ipython_directive` is at fault.

**Statistics**
```
$ python -c "import IPython; print(IPython.sys_info())"
{'commit_hash': '523ed2fe5',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/home/max/.conda/envs/pandas-dev/lib/python3.7/site-packages/IPython',
 'ipython_version': '7.2.0',
 'os_name': 'posix',
 'platform': 'Linux-4.15.0-45-generic-x86_64-with-debian-buster-sid',
 'sys_executable': '/home/max/.conda/envs/pandas-dev/bin/python',
 'sys_platform': 'linux',
 'sys_version': '3.7.2 (default, Dec 29 2018, 06:19:36) \n[GCC 7.3.0]'}

$ sphinx-build --version
sphinx-build 1.8.4

$ python -c "import pandas pandas.__version__"
'0.25.0.dev0+128.g659e0cae6'
```

sjpfenninger:
I also came across this bug when using `@savefig` in a freshly created (`sphinx-quickstart`) Sphinx documentation, and it took me a while to diagnose that the problem is not related to the issues reported in #8733 but rather to the project path containing whitespace. I get the same errors:

```
UsageError: %bookmark: too many arguments
UsageError: %bookmark -d: Can't delete bookmark 'ipy_savedir'
```

[...]

PR #13765, opened 2022-09-28
Allow spaces in file paths when using sphinx directive Closes: #11606

elguitar:
A simple change and doing the PR in a hurry. I think that this resolves the problem discussed in #11606.

Please let me know if there's anything more I can do :)


jasongrout:
I confirmed that building pandas docs in a directory with a space did not work before this PR, but do work after this PR. Thanks!

jasongrout (approved):

jasongrout:
The mypy error is fixed over in #13820

PR #13768, opened 2022-10-01
raise an error when user tries to open a standard stream

### LINKED ISSUE #13718: Incorrect call to `open` function not handled correctly.
<!-- This is the repository for IPython command line, if you can try to make sure this question/bug/feature belong here and not on one of the Jupyter repositories. 

If it's a generic Python/Jupyter question, try other forums or discourse.jupyter.org.

If you are unsure, it's ok to post here, though, there are few maintainer so you might not get a fast response. 

-->

If you run this code:

```
def crash(file_name="test.txt"): 
     open(file_name)                                                                                                                                                                                                             

crash(True)  
```

ipython will crash with the following error:

```
  File "/home/yunoac/anaconda3/bin/ipython", line 11, in <module>
    sys.exit(start_ipython())
  File "/home/yunoac/anaconda3/lib/python3.7/site-packages/IPython/__init__.py", line 125, in start_ipython
  File "/home/yunoac/anaconda3/lib/python3.7/site-packages/traitlets/config/application.py", line 664, in launch_instance
    app.start()
  File "/home/yunoac/anaconda3/lib/python3.7/site-packages/IPython/terminal/ipapp.py", line 356, in start
    self.shell.mainloop()
  File "/home/yunoac/anaconda3/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 498, in mainloop
    self.interact()
  File "/home/yunoac/anaconda3/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 478, in interact
    print(self.separate_in, end='')
OSError: [Errno 9] Bad file descriptor

If you suspect this is an IPython bug, please report it at:

to interactively debug it.

Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
OSError: [Errno 9] Bad file descriptor
```

osherdp:
Fixes #13718

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1258,9 +1258,10 @@ class InteractiveShell(SingletonConfigurable):
         @functools.wraps(io_open)
         def modified_open(file, *args, **kwargs):
             if file in {0, 1, 2}:
-                raise ValueError(f"IPython won't let you open fd={file} by default "
-                "as it is likely to crash IPython. If you know what you are doing, "
-                "you can use builtins' open."
+                raise ValueError(
+                    f"IPython won't let you open fd={file} by default "
+                    "as it is likely to crash IPython. If you know what you are doing, "
+                    "you can use builtins' open."
                 )
 
             return io_open(file, *args, **kwargs)
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1255,6 +1255,15 @@ class InteractiveShell(SingletonConfigurable):
         if user_ns is None:
             user_ns = user_module.__dict__
 
+        @functools.wraps(io_open)
+        def modified_open(file, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


IPython/core/interactiveshell.py
```diff
@@ -1239,6 +1239,15 @@ def prepare_user_module(self, user_module=None, user_ns=None):
         if user_ns is None:
             user_ns = user_module.__dict__
 
+        @functools.wraps(io_open)
+        def modified_open(file, *args, **kwargs):
+            if file in {0, 1, 2}:
+                raise ValueError(f"IPython won't let you open fd={file} by default")
```
Carreau, line 1245:
```suggestion
                raise ValueError(f"IPython won't let you open fd={file} by default "
                "as it is likely to crash IPython. If you know what you are doing, "
                "you can use builtins' open."
                )
```

PR #13773, opened 2022-10-04
Stop monkeypatching `linecache`

o11c:
It turns out this has never been necessary if we just change the cache entry to forbid expiry in the first place.


o11c:
Ugh, I *cannot* get over `black`'s weird defaults.

To avoid noise I'll avoid fixing that in case there's something else that should be changed early.

PR #13774, opened 2022-10-04
autoreload verbosity

shaperilio:
Worked on three things:
1. More descriptive parameter names for `%autoreload`; `now`, `off`, `explicit`, `all`, `complete`. (This last one could probably use a better name, but I couldn't think of anything better based on the message in 1d3018a93e98ad55f41d4419f835b738de80e1b7)
2. New optional arguments for `%autoreload` allow displaying the names of modules that are reloaded. Use `--print` or `-p` to use `print` statements, or `--log` / `-l` to log at `INFO` level.
3. `%aimport` can parse whitelist/blacklist modules on the same line, e.g. `%aimport os, -math` now works.

`%autoreload` and will also now raise a `ValueError` if the parameter is invalid. I suppose a bit more verification could be done for input to `%aimport`....


shaperilio:
Can I get some feedback on this?

krassowski:
Thank you for working on it. Is there an issue where the proposed enhancements were previously discussed? If so, could you link it? If not, could you search for related issues to summarise any prior/related discussions?

IPython/extensions/autoreload.py
```diff
@@ -595,15 +618,41 @@ def aimport(self, parameter_s="", stream=None):
...
+                    self._reloader.mark_module_skipped(_module)
+                else:
+                    top_module, top_name = self._reloader.aimport_module(_module)
+
+                    # Inject module to user namespace
+                    self.shell.push({top_name: top_module})
+
+    @line_magic
+    def averbose(self, parameter_s=""):
+        r"""%averbose => Turn verbosity on/off for autoreloading.
```
krassowski, line 634: Personally I would prefer if instead of adding another magic, the `%autoreload` magic accepted `--verbose` argument. There is already a mechanism for parsing arguments in magics.

IPython/extensions/autoreload.py
```diff
@@ -595,15 +618,41 @@ def aimport(self, parameter_s="", stream=None):
...
 
-                # Inject module to user namespace
-                self.shell.push({top_name: top_module})
+        %averbose 0 or %averbose off
```
krassowski, line 636: I would avoid adding more magic numbers. off/on/log is fine.

IPython/extensions/autoreload.py
```diff
@@ -552,21 +573,22 @@ def autoreload(self, parameter_s=""):
           autoreloaded.
 
         """
-        if parameter_s == "":
+        if parameter_s == "" or parameter_s.lower() == "now":
```
krassowski, line 576: You could add a new variable like `parameter_s_lower = parameter_s.lower()` and only use that in tests to save quote a few calls. Even better, this could be solved with a dictionary but probably not in this PR.

krassowski (commented):
Some high-level comments.

shaperilio:
> Thank you for working on it. Is there an issue where the proposed enhancements were previously discussed? If so, could you link it? If not, could you search for related issues to summarise any prior/related discussions?

I did not find any relevant discussions when I did a search prior to working on this, though I may have missed something since I'm not active in the repo.

We use `%autoreload` often in one of my projects, and have been "bitten" by some of the quirks (i.e. not realizing we had changed a class and then getting "nonsense" errors after autoreload). So we modified the source so that autoreload shows which modules it's updating.

I found this very useful, so I decided to work on submitting the change upstream.

At that point I realized I could also change the arguments from simple ints to something more descriptive--I've never met anyone in person who knew what they were, I was always told "use 2" or whatever they were passing along from previous experience.

IPython/extensions/autoreload.py
shaperilio: Sounds great; I'll work on this.

IPython/extensions/autoreload.py
shaperilio: I'll get rid of `%averbose` as discussed above.

IPython/extensions/autoreload.py
shaperilio: Is this a style thing or an efficiency thing? Happy to make the change; I'm just curious.

IPython/extensions/autoreload.py
shaperilio: This is done; added `--print` and `--log` options for `%autoreload`.

shaperilio:
I made changes as you suggested. Let me know if there are any other improvements you'd like to see.

IPython/extensions/autoreload.py
```diff
@@ -552,21 +605,45 @@ def autoreload(self, parameter_s=""):
...
+
+        def pl(msg):
+            p(msg)
+            l(msg)
+
+        if args.print is False and args.log is False:
+            self._reloader._report = lambda msg: None
+        elif args.print is True:
+            if args.log is True:
+                self._reloader._report = lambda msg: pl(msg)
```
krassowski, line 625: Unless there is a reason for wrapping in lambdas I would suggest to just assign the function directly.

```suggestion
                self._reloader._report = pl
```

IPython/extensions/autoreload.py
```diff
@@ -552,21 +605,45 @@ def autoreload(self, parameter_s=""):
...
+            else:
+                self._reloader._report = lambda msg: p(msg)
```
krassowski, line 627:
```suggestion
                self._reloader._report = p
```

IPython/extensions/autoreload.py
```diff
@@ -552,21 +605,45 @@ def autoreload(self, parameter_s=""):
...
+        elif args.log is True:
+            self._reloader._report = lambda msg: l(msg)
```
krassowski, line 629:
```suggestion
            self._reloader._report = l
```

IPython/extensions/autoreload.py
```diff
@@ -552,21 +605,45 @@ def autoreload(self, parameter_s=""):
           autoreloaded.
 
         """
-        if parameter_s == "":
+        args = magic_arguments.parse_argstring(self.autoreload, line)
+        mode = args.mode.lower()
+
+        def p(msg):
+            print(msg)
```
krassowski, line 612: Worth making it one func def and one call on each reload less?

```suggestion
        p  = print
```

IPython/extensions/autoreload.py
```diff
@@ -552,21 +605,45 @@ def autoreload(self, parameter_s=""):
...
+
+        def l(msg):
+            logging.getLogger("autoreload").info(msg)
```
krassowski, line 615: Unless there is a case for hot-swapping the logger at execution.

```suggestion
        logger = logging.getLogger("autoreload")
        def l(msg):
            logger.info(msg)
```

IPython/extensions/autoreload.py
```diff
@@ -581,8 +658,9 @@ def aimport(self, parameter_s="", stream=None):
         %aimport foo, bar
         Import modules 'foo', 'bar' and mark them to be autoreloaded for %autoreload 1
 
-        %aimport -foo
-        Mark module 'foo' to not be autoreloaded for %autoreload 1
+        %aimport -foo, bar
+        Mark module 'foo' to not be autoreloaded for %autoreload 1, 2, or 3, and 'bar'
+        to be autoreloaded for 1.
```
krassowski, line 663: Feel welcome to replace the digits with verbose names here and in the paragraph above if you agree that this is a good idea.

docs/source/whatsnew/pr/autoreload-verbosity.rst
```diff
@@ -0,0 +1,21 @@
+Autoreload verbosity
+====================
+
+We introduce more descriptive names for the `%autoreload` parameter:
```
krassowski, line 4: Two suggestions:
- replace all single backticks with double backtics
- add a new line here so that the list renders nicely

rst has quite specific requirements.

```suggestion
We introduce more descriptive names for the ``%autoreload`` parameter:

```

krassowski (commented):
Tested locally works perfectly. Small suggestions to reduce number of function definitions and RST syntax.

docs/source/whatsnew/pr/autoreload-verbosity.rst
shaperilio: Sorry about that. I rendered the RST and think I got it right now.

IPython/extensions/autoreload.py
shaperilio: Done; thanks for the suggestion.

krassowski (approved):
Thanks!

shaperilio:
I'm not sure what to do about this failing doc build...

krassowski:
The docs build will be fixed by https://github.com/ipython/ipython/pull/13872. Thanks for highlighting it.

shaperilio:
@krassowski I don't know what the failed test is about, is there a way to rerun it?

krassowski:
Yes, you can close and then immediately reopen the pull request to kick GitHub actions.

shaperilio:
Cool, that worked. Everything is passing now.

krassowski:
@fperez what do you think about this one? This would also fix https://github.com/ipython/ipython/issues/13915 (as discussed in https://github.com/ipython/ipython/pull/13916).

shaperilio:
FYI I updated the original comment at the top of the PR to reflect the behavior after all your suggestions.

shaperilio:
@Carreau @fperez can we merge this one for 8.11 too?

fperez (approved):
Excellent, I very much like these improvements, many thanks! They'll come in handy in my current class.

Tagging @facusapienza21 and @isaacdsloan for awareness, we'll be using this when released.

fperez:
Done @shaperilio, many thanks for the great improvements and well documented/tested work. Much appreciated (and sorry for the slow response!).

PR #13785, opened 2022-10-13
Use `pathlib` instead of `os` in `IPython/{lib/demo,core/hooks}`

ChandanChainani:
Progress for #12515


ChandanChainani:
@Carreau please review the PR

Carreau:
Thanks for your PRs. Though there is 81 PRs in the list that are older than yours, and all the work here is volunteer. I'll do my best, but it is not likely to be fast.

drunkwcodes (approved):

Carreau:

PR #13788, opened 2022-10-16
Support square bracket indexing with literal with pinfo/?/pinfo2/??

### LINKED ISSUE #13780: Support for question mark (?) for objects part of list
**Consider this reproducer** 

```python
class B:
    
    @property
    def foo(self):
        """
        Documentation for Foo
        """
        return self._foo
    
    @foo.setter
    def foo(self, v):
        self._foo = v
        
    def some_method(self):
        """
        Documentation for some method
        """
        pass

class A:
    
    def __init__(self):
        self.b = [B(), B()]
        
a = A()

```
Now doing 

```python
a.b[0].some_method?
```
or
```python
a.b[1].foo?
```

Does not seem show docstring, instead it says `Object ``Foo`` not found` 

**Context**

We are developing a Python library (class) that holds objects in a list. The objects have methods and properties defined on them which does not seem to show help as per the above snippet. It would be really useful if users can quickly access documentation.

In general, the object on which property or method is requested may be present in `dict` e.g.  `a.b['key'].foo?` or it may be coming from a function `a().foo?` - in these cases as well we get the same error. Though if only simple list case works, it would suffice for us.

We are aware that this works - 

```python
temp = a.b[0]
temp.foo?
```
It would be great if it works without needing to put into a variable first.

**Relevant Code Sections**
https://github.com/ipython/ipython/blob/ae0dfcd9d4fb221f0278f07b022f9837b0e51e67/IPython/core/magics/basic.py#L298-L303
https://github.com/ipython/ipython/blob/57eaa12cb50c9a95213b9e155032e400b9424871/IPython/core/interactiveshell.py#L1532-L1544

Open for suggestions / workarounds!

Carreau:
We add limited support for evaluating `[xxx]` where xxx is a numeric literal. This still abit dangerous as __getindex__ can run arbitrary code.

We could support a wider range of constructs, but this would start to reach into complex and dangerous territory. Alredy this with lead to weird suff like
```
   >>> def foo(*arg, **kwargs):pass
   >>> c = [foo]
   >>> c[0]?
   Signature: c[0](*args, **kwargs)
```
Instead of
```
   >>> c[0]?
   Signature: foo(*args, **kwargs)
```
As the name of the object is used in the signature is the one if the main namespace.

This should fix https://github.com/ipython/ipython/issues/13780

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -147,7 +147,8 @@ dedent_re = re.compile(r'^\s+raise|^\s+return|^\s+pass')
 # Utilities
 #-----------------------------------------------------------------------------
 
-def is_integer_string(s:str):
+
+def is_integer_string(s: str):
     """
     Variant of "str.isnumeric()" that allow negative values and other ints.
     """
@@ -156,7 +157,8 @@ def is_integer_string(s:str):
         return True
     except ValueError:
         return False
-    raise ValueError('Unexpected error')
+    raise ValueError("Unexpected error")
+
 
 @undoc
 def softspace(file, newvalue):
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -213,14 +213,17 @@ class ExecutionInfo(object):
         raw_cell = (
             (self.raw_cell[:50] + "..") if len(self.raw_cell) > 50 else self.raw_cell
         )
-        return '<%s object at %x, raw_cell="%s" store_history=%s silent=%s shell_futures=%s cell_id=%s>' % (
-            name,
-            id(self),
-            raw_cell,
-            self.store_history,
-            self.silent,
-            self.shell_futures,
-            self.cell_id,
+        return (
+            '<%s object at %x, raw_cell="%s" store_history=%s silent=%s shell_futures=%s cell_id=%s>'
+            % (
+                name,
+                id(self),
+                raw_cell,
+                self.store_history,
+                self.silent,
+                self.shell_futures,
+                self.cell_id,
+            )
         )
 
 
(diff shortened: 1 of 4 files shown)


IPython/core/interactiveshell.py
```diff
@@ -1537,10 +1537,33 @@ def _ofind(self, oname, namespaces=None):
...
+        parts_ok = True
+        for p in raw_parts:
+            if p.endswith("]"):
+                var, *indices = p.split("[")
+                if not var.isidentifier():
+                    parts_ok = False
+                    break
+                parts.append(var)
+                for ind in indices:
+                    if ind[-1] != "]" and not ind[:-1].isnumeric():
```
krassowski, line 1551: This currently prohibits `my_list[-1]`, `my_list[10_000]` or `my_list[0b101]`. One solution would be to use tokenize. I can provide a snippet. Other parts should not require a change for the first two cases since `int('-1')`, `int('10_000')` should just work. The binary/hex specification seems less important but could be allowed with a simple change too.

krassowski: No, it actually works ok, this code is much more permissive than I understood from reading it. It will actually accept anything.

Carreau: Yeah, it will allow it but will rarely receive it, the regex that transform `?` into `pinfo` is already restricted to 0-9. You can still call %pinfo directly.

krassowski: Could the regex also allow a single `-` prefix then? I often check the last item in the array and being able to check `mylist[-1]?` would be useful.

Carreau: Sure.

PR #13790, opened 2022-10-17
Typo fix in reset method docstring in namespace.py

jcollins1983:
Correct unattended to unintended in reset method docstring.


Carreau:
THanks.

PR #13808, opened 2022-11-02
pass sys.argv through %run when invoking .ipynb and .ipy files

tonyfast:
this change moves the `safe_execfile_ipy` a litte later in the `%run` magic.
with this change it is now possible to pass command line arguments into in the `%run <file.ipynb|ipy>`.

part of the motivation for this change is to use the `%run` method to bootstrap notebooks that are running in `jupyterlite`.  we have to `await micropip` and currently `IPython` scripts are the only way to top-level `await` statements. :metal: passing the `sys.argv` around will make it possible to write more versatile startup code and blocks.


tonyfast:
seems like things are working after merging main, but i don't know where that docs error is coming from.

krassowski:
Docs will be fixed by https://github.com/ipython/ipython/pull/13872.

Carreau:

PR #13819, opened 2022-11-09
Filter out new jupyter_core deprecation warning

jasongrout:
Jupyter_core 5.0 has a new deprecation warning that shows by default and halts our CI tests.

See https://github.com/jupyter/jupyter_core/issues/310


jasongrout:
yep, looks like the [ipykernel PR](https://github.com/ipython/ipykernel/pull/1019) fixed the issue with the test here:

PR #13820, opened 2022-11-09
Fix mypy error by being more explicit and verbose

jasongrout:
This mypy error is affecting CI for all other PRs.


jasongrout:
The mypy test now passes!

Issue #13759, opened 2022-09-15
Make inspect calls more predictable and safe


jasongrout-db:
The Colab team has addressed some issues it found in the https://github.com/ipython/ipython/blob/main/IPython/core/oinspect.py module in an open-source module at https://github.com/googlecolab/colabtools/blob/main/google/colab/_inspector.py. These include safeguarding against calling `repr(obj)`.

CC @blois @craigcitro - have we tried upstreaming some of these safeguards back to IPython?

More specifically, quoting the colab _inspector.py file:

https://github.com/googlecolab/colabtools/blob/947fd913f8ba1bd834cbf6498462e162f8675ddf/google/colab/_inspector.py#L65-L73
```
  IPython.core.oinspect.getdoc wraps Python's inspect.getdoc to catch exceptions
  and allow for objects with a custom getdoc() method. However, there are two
  problems:
   * inspect.getdoc already catches any exceptions
   * it then calls get_encoding, which calls inspect.getfile, which may call
     repr(obj) (to use in an error string, which oinspect.getdoc throws away).
  We replace this with our own wrapper which still allows for custom getdoc()
  methods, but avoids calling inspect.getfile.
```
and https://github.com/googlecolab/colabtools/blob/947fd913f8ba1bd834cbf6498462e162f8675ddf/google/colab/_inspector.py#L182-L199
```
  Colab's UI is aggressive about inspecting objects, and we've discovered that
  in practice, too many objects can have a repr which is expensive to compute.
  To make this work, we specify a set of types for which we compute a repr:
   * builtin types which aren't Iterable are safe
   * Sized objects with a `.shape` tuple attribute (eg ndarrays and dataframes)
     get a summary with type name and shape
   * list, dict, set, frozenset, and tuple objects we format recursively, up to
     a fixed depth, and up to a fixed length
   * other Iterables get a summary with type name and len
   * all other objects get a summary with type name
  In all cases, we limit:
   * the number of elements we'll format in an iterable
   * the total depth we'll recur
   * the size of any single entry
  Any time we cross one of these thresholds, we use `...` to imply additional
  elements were elided, just as python does when printing circular objects.
```

craigcitro:
Hey @jasongrout-db!

We didn't try to upstream the whole thing, for two reasons:
1. The "aggressive about inspecting objects" seemed to be particular to the Colab UI.
[...]

PR #13833, opened 2022-11-17
Predictable and safe object inspection

jasongrout:
In https://github.com/ipython/ipython/issues/13759, we discussed how Colab's inspection routine guards against calling an arbitrary object's `__repr__()` method, since in practice that can take a lot of time and/or memory, depending on how the object was implemented.

This PR experiments with including an updated/modified version of the Colab implementation (licensed Apache 2.0), and making it easy to configure the interactive shell's inspector implementation.

Use the new inspector with 
```
ipython --InteractiveShell.inspector_class='IPython.core.oinspect_safe.SafeInspector'
```

CC @craigcitro


jasongrout:
I see a difference in the %pinfo output between the two inspectors:

```python
class BadRepr:
    def __repr__(self):
        import time
        time.sleep(10)
        return 'a very long string'*100
```

Default inspector: 

```
In [2]: %pinfo BadRepr
Init signature: BadRepr()
Docstring:      <no docstring>
Type:           type
Subclasses:     
```
SafeInspector:

```
In [2]: %pinfo BadRepr
Type:        type
String form: <class '__main__.BadRepr'>
Docstring:   <no docstring>
```

jasongrout:
Colab special-cases tensorflow objects:

```python
        if (
            isinstance(shape, tuple)
            or hasattr(shape, "__module__")
            and isinstance(shape.__module__, str)
            and "tensorflow." in shape.__module__
        ):
            return "{} with shape {}".format(type_name, shape)
```
Should the IPython implementation also specialcase tensorflow? If not, can we make it easy for a subclass to have its own special cases at this point?

jasongrout:
I tried to unify the safe inspect info() and oinspect info() implementations. They mostly align. There have been many updates to oinspect info(), which I tried to reason about in the context of the safe inspection. There are still a few notes about TODO items I'm not sure about, but by in large this brings safe inspection in line with what the oinspect info currently produces. It seems that there are only a few key places where we needed to change the upstream implementation, so I put stubs in for functions the safe implementation can override. Then deleted the safe_inspect info() code. This *greatly* simplifies things. I spent a lot of time reading the Python inspect module code to convince myself that it was safe to rely on standard library code for more things. I tried to clearly indicate reasons behind many of the changes in the commits in this PR.

PR #13834, opened 2022-11-18
Pin minimum prompt-toolkit to 3.0.11

### LINKED ISSUE #13631: ipython 8.2.0 mixing history from multiple ipython sessions
Hi, 

After updating to ipython 8.2.0 I observe a very annoying behaviour, when pressing 'UP_ARROW' in long running ipython session brings not the previous command from history, but a command from some other ipython session running concurrently. 

How to reproduce.
* Start ipython session 1
* Start ipython session 2
* Type x=1 in session 1
* Type y=1 in session 2
* Type x=2 in session 1
* Type y=2 in session 2
* Now in session 1 press 'UP_ARROW', and it will bring y=1 instead of x=2 

This is very inconvenient behaviour that completely breaks the workflow in long running ipython sessions. I assume it must be a regression.

Thank you.

(I'm using python 3.8 with the latest prompt-toolkit 3.0.29 on linux (ubuntu 18.04)

hristog:
Fixes #13631.

`3.0.11` is the first `prompt-tookit` version which fixes the bug related to scrolling through IPython command history.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ install_requires =
     matplotlib-inline
     pexpect>4.3; sys_platform != "win32"
     pickleshare
-    prompt_toolkit>3.0.1,<3.1.0
+    prompt_toolkit>=3.0.11,<3.1.0
     pygments>=2.4.0
     stack_data
     traitlets>=5


Carreau:
Thanks.

Issue #13837, opened 2022-11-20
%time magic displays output even when code ends in semicolon


fperez:
This was already mentioned in passing by @valrus #6130 opened by @minrk, but it was reported again by @norvig and it's worth tracking specifically for `%time`. It's easy to replicate,

```
In[1]: foo();
```

prints no output, because of the semicolon.

```
In[2]: %time foo();
```

prints timing info, and also the result of calling foo(), even though there is a semicolon.

The [code for time](https://github.com/ipython/ipython/blob/28f76f8a2f37700f8d74e91c4a54929dc367000f/IPython/core/magics/execution.py#L1197) isn't completely trivial, but basically we need to make at least some of our magics related to code execution uniformly use the same display logic. See also #10227.

PR #13841, opened 2022-11-25
%time magic displays output even when code ends in semicolon #13837

nfgferreira:
After the magic is evaluated and the result is calculated, the modification tests whether the evaluated magic was _time_ and whether semicolon is the final character. The result is killed if both things happen.

My choice would be to remove the _time_ test, so a semicolon would prevent the print of the output of any magic, but this is only a suggestion I keep open.

I did not write any automated test, but I can do that once (and if) the solution is accepted.

[#13837](https://github.com/ipython/ipython/issues/13837) points to [#10227](https://github.com/ipython/ipython/issues/10227) (_Cell magic result in printing the last evaluated line even if followed by semicolon_). There, somebody says that ';' may be a meaningful character because we could have a C++ expression, for instance. The IPython repository says the documentation for other languages is in Jupyter. I ran Jupyter on my browser with C++ and saw that a semicolon after the last statement prevents the output to be printed (a semicolon between 2 statements in a cell seems to be necessary, though). See attached file for simple examples. Therefore, it seems that the semicolon at the end in C++ already behaves the same way that in Python and is not required by the interpreter.

 
IPython_Cpp


IPython/core/interactiveshell.py
```diff
@@ -2362,6 +2362,14 @@ def run_line_magic(self, magic_name: str, line, _stack_depth=1):
                 kwargs['local_ns'] = self.get_local_scope(stack_depth)
             with self.builtin_trap:
                 result = fn(*args, **kwargs)
+
+            # The code below prevents output from being displayed
+            # when using magic %time.
+            # Output from '%time foo();', for instance, would never
+            # be displayed.
+            if magic_name == "time" and len(magic_arg_s) > 0 and magic_arg_s[-1] == ";":
+                return None
```
krassowski, line 2371: Comparing to the `quiet` implementation:

https://github.com/ipython/ipython/blob/28f76f8a2f37700f8d74e91c4a54929dc367000f/IPython/core/displayhook.py#L84-L103

this proposal does not seem to handle comments or flanking whitespaces; I know that comments in general are not handled well by magics right know (which is  a pity), but I think that there should be a test case for suffix space/tab etc to be a bit more permissive in interactive coding.

krassowski (commented):
> There, somebody says that ';' may be a meaningful character because we could have a C++ expression, for instance.

I understand that this concern applies to magics, not other kernels. For example, IPython provides a `%%javascript` magic:

Screenshot from 2022-11-27 15-47-13

as well as `%%perl`, `%%markdown`, `%%latex` etc. Other packages provide other magics (e.g. [rpy2](https://github.com/rpy2/rpy2) - `%R` and `%%R`, [rust-magic](https://github.com/axil/rust-magic) - `%rust` and `%%rust`). There are also magics for languages which require semicolons like C and we probably would not want to break them.

> My choice would be to remove the time test, so a semicolon would prevent the print of the output of any magic, but this is only a suggestion I keep open.

I see the appeal of making this work across magics. I would enjoy if `%R` magic which I use often could benefit from this behaviour, but as explain above, we would NOT want this to happen to all magic without giving magic authors a way out.

A solution that we could go for is to allow to opt-in, or opt-out using a special decorator, in the same way how [`@no_var_expand`](https://github.com/ipython/ipython/blob/28f76f8a2f37700f8d74e91c4a54929dc367000f/IPython/core/magic.py#L260-L276) allows magic authors to opt-out of variable expansion. This would be better than comparing against the magic name since users may create a custom magic with the same name as built-in `%time` (in opt-in scenario we would add this decorator to built-in time magic, in opt-out we would target this change against next major version).

What do you think?

IPython/core/interactiveshell.py
nfgferreira: Thanks for educating me. I am also not a big fan of comparing the magic name for a particular implementation directly in the code either. I simply took a look on a list of magics I found on the internet and thought that the code in the PR (excluding the magic name comparison) would be OK for every magic. I did not know magics are extensible. Since they are, the idea of creating a decorator makes sense.

As for your comment about comparing the last character against the code in `quiet()`: I am pretty sure that the semicolon at the end of the line is not detected by `quiet()` today for any implemented magic. Debugging the code, the command

`%time 1;  #comment `

creates the following statement that is parsed by `quiet()`:

`get_ipython().run_line_magic('time', '1;  #comment')`

As you see, ; is not the last token of the expression and therefore is not detected.

The fact that the semicolon never prevents the expression value from being calculated today when a magic is performed plus the fact that magics are extensible and can basically do anything, makes me think that using something like `quiet()` (which is based on Python syntax) is too restrictive. I guess that perhaps we can configure the sub-string as a parameter of the decorator.

Example:

The definition of the magic time would be:

```
    @skip_doctest
    @no_var_expand
    @needs_local_scope
    @line_cell_magic
    @prevent_output_printing('%%')  # <--- THIS WILL BE ADDED
    def time(self,line='', cell=None, local_ns=None):
    ...
```

`@prevent_output_printing` would remove the sub-string '%%' if it finds one at the end of the line, call the magic implementation (time in the example) and then force `None` as the return value if the sub-string was found.

A line like that:

`%time 1  # comment %%`

would not produce a '`1'` on the screen.

We could implement something like `quiet()` in the decorator too, but my concern is that this may be too restrictive.

What do you think? Please be aware that I am not a hard IPython user and I am completely sure you and other people in this forum have used the software much more than me.

nfgferreira: One more note: I am not sure I should convert this to draft even if it is not ready to be merged, since I am afraid if I draft it, not everybody will have access to this discussion. What should I do here, @krassowski? Thanks.

krassowski: Feel free to convert it to draft and back, it does not affect the visibility of discussion.

> magics are extensible and can basically do anything, makes me think that using something like quiet() (which is based on Python syntax) is too restrictive. I guess that perhaps we can configure the sub-string as a parameter of the decorator.

I don't think that we need to support arbitrary suffixes; I am not against this in principle, but this may introduce interaction with other parsing bits, so I would suggest to postpone it possibly for another PR.

nfgferreira: New attempt. Please take a look. I don't know why, but darker complained about the single quotes in _ipython_magic_no_var_expand, which I changed to double quotes. Thanks.

krassowski (approved):
Nice, thank you! The only change I would recommend is `output_can_be_disabled` → `output_can_be_silenced` but this is subjective.

nfgferreira:
No problem. I will do that. Please give me some hours.

nfgferreira:
Hmmm... It's been more than 1 month since this PR reached its current status, but it has not been merged and raised no comments either, so I suspect it is in a kind of limbo. Is there anything wrong with it? Thanks.

krassowski:
I personally think it should be good to merge. A gentle ping to @fperez who opened https://github.com/ipython/ipython/issues/13837.

fperez:
Thx for the ping @krassowski, I'd missed this. Thanks for the well tested code @nfgferreira and the detailed review/feedback for improvement, @krassowski.

My only request before merging would be to add a note to [the docs](https://github.com/ipython/ipython/blob/add5877a42ba8e3960bc92eb994c15955eacf254/docs/source/config/custommagics.rst) about the existence of this new decorator/capability for magic authors. Otherwise it's unlikely anyone will realize this is possible, as they'd have to find it in the code itself.  That will ensure this improvement has more benefit.

I'd be happy to merge after that, but feel free to do it @krassowski if you see it before I do.

Thanks again!

nfgferreira:
@fperez 's comments make sense. I updated the rst file and pulled the new version. After I tried to sync my PR to the current head, one of the tests failed with the following message:
`Error: fatal: unable to access 'https://github.com/ipython/ipython/': The requested URL returned error: 429`
I suspect this is s github glitch which I do not know how to workaround. Anyway, perhaps the best thing to do right now is just merge the branch. @krassowski , please let me know if there is anything I should do in my end. Thanks.

krassowski:
Thank you for the documentation updates, I have no further comments. Pinging @fperez and @Carreau as it is up to them now to accept (or not).

fperez:
Thanks for the doc update @nfgferreira, and sorry I missed this last week (swamped at work)! 

Merging now to avoid stalling you further - any other refinements can be done later if needed. 

Great work, thanks all!!

Issue #13815, opened 2022-11-06
ipythonMAJ.MIN entry points are not installed


bnavigator:
Contrary to what the release notes say, #13743 does not have such effect on a `python setup.py install` or `pip install`. The published wheel in PyPI does not have the entry points either, because it is minor-version agnostic:

`ipython-8.6.0-py3-none-any.whl::ipython-8.6.0.dist-info/entry_points.txt`
```
[console_scripts]
ipython = IPython:start_ipython
ipython3 = IPython:start_ipython
```

Simple pip install from PyPI:
```
[ben@skylab:~]% python3 -m venv ipenv                                                                                                                                                                             [0]
[ben@skylab:~]% . ipenv/bin/activate                                                                                                                                                                              [0]
(ipenv) [ben@skylab:~]% pip install ipython                                                                                                                                                                       [0]
Collecting ipython
  Downloading ipython-8.6.0-py3-none-any.whl (761 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 761.1/761.1 kB 6.8 MB/s eta 0:00:00
Collecting stack-data
  Downloading stack_data-0.6.0-py3-none-any.whl (24 kB)
Collecting traitlets>=5
  Downloading traitlets-5.5.0-py3-none-any.whl (107 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 107.4/107.4 kB 6.4 MB/s eta 0:00:00
Collecting matplotlib-inline
  Downloading matplotlib_inline-0.1.6-py3-none-any.whl (9.4 kB)
Collecting pygments>=2.4.0
  Downloading Pygments-2.13.0-py3-none-any.whl (1.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.1/1.1 MB 7.7 MB/s eta 0:00:00
Collecting backcall
  Using cached backcall-0.2.0-py2.py3-none-any.whl (11 kB)
Collecting decorator
  Downloading decorator-5.1.1-py3-none-any.whl (9.1 kB)
Collecting prompt-toolkit<3.1.0,>3.0.1
  Downloading prompt_toolkit-3.0.32-py3-none-any.whl (382 kB)
[...]

PR #13842, opened 2022-11-25
Reused the previously unused find_entry_points in setup.py

UltimateLobster:
In continuation to #13743. It seems that the function was unused and therefore the changes I've made did not cause the new entry point to be added. This fix should solve #13815.

Unfortunately, since setup.cfg does not support dynamic name for entrypoints (name based on sys.version_info), I had to bring that configuration back to setup.py and reuse the function.


Carreau:
Thanks.

PR #13851, opened 2022-12-02
docs: remove mention of `_ipython_display_` being ignored in the REPL

agoose77:


jasongrout (approved):
Thanks! I checked, and the ipython repl does indeed respect defining an _ipython_display_ method.

PR #13854, opened 2022-12-04
Fix codecov coverage uploads

### LINKED ISSUE #13853: Incorrect coverage reporting
Back in February the coverage was close to 73.89% library and 97.83% for tests. Currently the coverage for tests is not reported and for library sits at around 40%.

This started happening in April. e6432249582e05f438303ce73d082a0351bb383e has 73.84% library coverage but next commit, cf0cbaf5cfbd8464e5b0ea69da8d141671008e82 has only 41.20%. Both commits are unrelated (both fixing typos in documentation).

When working on #13852 I saw a coverage drop even when adding comprehensively tested functions. Many functions and classes which I do test extensively are not marked as covered at all.

I also see some of incorrect misses locally but these are different for some reason.

It is very likely due to some change in `pytest-cov` and `coverage`.

krassowski:
Fixes #13853. The `coverage.xml` from test was not being picked up by codecov uploader. It seems to be a common occurrence. Providing a full path to the file fixes the issue. The coverage is now back at 73.02%.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -77,4 +77,7 @@ jobs:
       run: |
         pytest --color=yes -raXxs ${{ startsWith(matrix.python-version, 'pypy') && ' ' || '--cov --cov-report=xml' }}
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v2
+      uses: codecov/codecov-action@v3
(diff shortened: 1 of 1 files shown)


jasongrout:
Thanks!

PR #13856, opened 2022-12-04
Remove unused pytest ignores (files have been deleted)

krassowski:
Minor cleanup, as these files no longer exist.


jasongrout (approved):
Thanks!

PR #13858, opened 2022-12-07
Add py.typed to setup.cfg

rhelmot:
This makes py.typed actually get copied to the site-packages folder, which is necessary to enable typechecking on non-editable installs!


Carreau:
Thanks.

PR #13863, opened 2022-12-09
Refactor inspect info reply logic

jasongrout:
This refactors the logic of the inspector `_get_info` function to make it easier to override and extend its behavior. This PR should not affect the overall function of `_get_info`.

We factor out the logic to append an info field to the inspect_reply mimebundle and the logic for what information is in the mimebundle into separate functions that can easily be overridden by an inspector subclass. This allows a subclass to easily:

* format information into yet another mimetype besides text/plain and text/html
* modify or add to the default information

without having to copy the default implementation.

We also refactor and simplify the logic around formatting text/plain and text/html mimetypes in the inspect reply.

We also fix an escaping bug with the default html formatter so that the default docstrings `<no docstring>` appears instead of being interpreted as an HTML tag.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -588,7 +588,7 @@ class Inspector(Colorable):
             bundle["text/plain"].append((title, formatted_field["text/plain"]))
             bundle["text/html"].append((title, formatted_field["text/html"]))
 
-    def _make_info_unformatted(self, info, formatter, detail_level, omit_sections):
+    def _make_info_unformatted(self, obj, info, formatter, detail_level, omit_sections):
         """Assemble the mimebundle as unformatted lists of information"""
         bundle = {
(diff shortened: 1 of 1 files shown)


krassowski:
I was recently looking at these very lines and thinking about making the mime bundle more extendable in https://github.com/python-lsp/docstring-to-markdown/issues/25#issuecomment-1334632223. I though it's worth a mention in case if you find that discussion useful.

jasongrout:
Nice, thanks for pointing to your other discussion there. Do you think these changes are enough to experiment with the ideas you proposed in that other comment? (I think probably they are?). Also https://github.com/ipython/ipython/pull/13865 is important to make it easy for a user to change the inspector from configuration or command line, so they can easily use a subclass that is distributed separately.

Issue #13859, opened 2022-12-07
Support for `PySide6` via `%gui`


shaperilio:
I was twiddling about examining how `%gui` works, and see that the code effectively forces you into `PyQt` instead:

In [get_inputhook_name_and_func](https://github.com/ipython/ipython/blob/main/IPython/terminal/pt_inputhooks/__init__.py#L53), if 'qt5' or 'qt6' are requested, an environment variable is set forcing you away from `PySide<x>`. I'm not yet sure what the background is on this `QT_API` environment variable, but it doesn't seem great to me to set it for the purpose of specifying which library to load (I always assume environment variables with non-specific names could be used by other things in the system, but maybe that's a crazy perspective).

This environment variable will then dictate which Qt package will get imported via [qt_for_kernel.py](https://github.com/ipython/ipython/blob/main/IPython/external/qt_for_kernel.py#L107). `PySide6` is definitely supported here, but effectively inaccessible via `%gui`.

(The only missing "support" for `PySide6` that I see so far is in the error message in [qt_loaders.py](https://github.com/ipython/ipython/blob/main/IPython/external/qt_loaders.py#L379).)

I'm happy to work on this (I have some [notes on my branch](https://github.com/shaperilio/ipython/tree/shaperilio/pyside6-gui)) but would like some guidance from more active members about how this `QT_API` environment variable should be handled.

shaperilio:
I see mention of `QT_API` in [matplotlib](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/backends/qt_compat.py#L7)

shaperilio:
Never mind; I verified `PySide6` is preferred even if `PyQt5` is installed.

shaperilio:
Actually, after working on this for `ipykernel`, I see it's an issue (print statements and updated import error message are from [my branch](https://github.com/shaperilio/ipython/tree/shaperilio/pyside6-gui))
```
In [3]: %gui pyside6
Someone called `enable_gui` with gui='pyside6'.
`get_inputhook_name_and_func` called with gui='pyside6'
ERROR:root:No event loop integration for 'pyside6'. Supported event loops are: qt, qt4, qt5, qt6, gtk, gtk2, gtk3, gtk4, tk, wx, pyglet, glut, osx, asyncio
```
```
In [4]: %gui qt6
Someone called `enable_gui` with gui='qt6'.
`get_inputhook_name_and_func` called with gui='qt6'
gui_mod='qt'
[...]

PR #13864, opened 2022-12-09
ENH: support for `PySide6` in `%gui`

shaperilio:
Addresses #13859

Changes here parallel the changes in [ipykernel](https://github.com/ipython/ipykernel/pull/1054), i.e. prefer `PyQt` over `PySide` and allow requesting explicit versions (e.g. 'qt5') with one difference (see below). 

I believe that, eventually, the Qt importing logic should be all here, and `ipykernel` should defer to`ipython`.  I chose not to do that at this time since it would mean the latest `ipykernel` would require the latest `IPython`; I'm happy to discuss further.

The only difference between `IPython` and `ipykernel`, after these two pull requests, is that, in `IPython`, it's not possible to explicitly request an event loop for Qt4. This is because an alias exists [here](https://github.com/ipython/ipython/blob/5409de68d87ddd073a35111aca0cb8360ff63ca8/IPython/terminal/pt_inputhooks/__init__.py#L5) which effectively makes "qt4" be "the latest Qt available". I did not remove the alias because I don't know the history behind it.


shaperilio:
I'm not sure what to do about the docs failing to build - it doesn't seem to be an issue in main?

shaperilio:
@Carreau what's this force push about?

Carreau:
> @Carreau what's this force push about?

Some failures were due to a problem on the main branch, I fixed the test on the main branch and rebase this one top of main, to get the test to pass.

shaperilio:
This is ready for review. The equivalent changes in ipykernel have already been merged.

shaperilio:
There's a problem here with `matplotlib`, related to commit 5c57ffec50e1e86013a236f922ea9a53baec427e. 

`matplotlib.get_backend()` will actually set a backend if none has been set... and then returns it. I think the relevant code is [here](https://github.com/matplotlib/matplotlib/blob/be7ea76c49f19ce48958de82a82b3951d9b7f0d2/lib/matplotlib/__init__.py#L740). 

I have an application where I require the `TkAgg` backend, so I do something like this:
```python
if matplotlib.get_backend() != 'TkAgg':
  _l.debug('IPython shell detected. Attempting to change '
           'matplotlib backend to `TkAgg`...')
  try:
    matplotlib.use('TkAgg')
  except ImportError as e:
    raise RuntimeError('Cannot run. Matplotlib backend cannot be changed to `TkAgg`:\n' 
                       f'   {str(e)}')
```

Sadly, the call to `get_backend` actually sets the backend to `QtAgg` on my system, which hooks in the Qt eventloop. The error I added in 5c57ffec50e1e86013a236f922ea9a53baec427e then prevents me from switching to `TkAgg`, but evidently the exception is getting absorbed somewhere in the call to `matplotlib.use`.

So I'm changing the error to a warning. My rationale was that I'd want to prevent people from unknowingly changing which event loop was tied in - you could stop some other application from functioning if you do so - but I suppose if you really wanted to prevent that, `%gui` would have to be changed to support multiple event loops being stacked in the hook.

IPython/terminal/pt_inputhooks/__init__.py
```diff
@@ -40,6 +39,80 @@ def __str__(self):
...
+        QT_API_PYSIDE2: "qt5",
+        QT_API_PYQT5: "qt5",
+        QT_API_PYSIDE6: "qt6",
+        QT_API_PYQT6: "qt6",
+    }
+    if loaded is not None and gui != "qt":
+        if qt_env2gui[loaded] != gui:
+            raise ImportError(
+                f"Cannot switch Qt versions for this session; must use {qt_env2gui[loaded]}."
+            )
```
ccordoba12, line 73: Why not change this `raise` by a `print` instead? I think it's not a good UX to show a long traceback for this, i.e. users don't need to receive that kind of info, only that IPython can't switch to another Qt version or binding.

IPython/terminal/pt_inputhooks/__init__.py
```diff
@@ -40,6 +39,80 @@ def __str__(self):
...
+                except ImportError:
+                    os.environ["QT_API"] = "pyqt6"
+        elif gui == "qt":
+            # Don't set QT_API; let IPython logic choose the version.
+            if "QT_API" in os.environ.keys():
+                del os.environ["QT_API"]
+        else:
+            raise ValueError(
+                f'Unrecognized Qt version: {gui}. Should be "qt5", "qt6", or "qt".'
+            )
```
ccordoba12, line 113: The same here. I don't think it's necessary to raise an error for this.

ccordoba12 (commented):
Thanks @shaperilio for your work on this! I left some comments for you, which are basically the things I changed in ipykernel to provide a better user experience, in my opinion.

shaperilio:
> Thanks @shaperilio for your work on this! I left some comments for you, which are basically the things I changed in ipykernel to provide a better user experience, in my opinion.

Thanks for these. I should have learned my lesson with the `matplotlib` issue. :)

shaperilio:
Any thoughts on when this may be merged, or if there's more feedback? I believe it's been released in ipykernel.

Carreau:
Sorry, this fell under the radar. Let's try it.

PR #13889, opened 2023-01-09
Use prompt_toolkit.application.create_app_session for debugger prompt

### LINKED ISSUE #12192: Infinite loop when using ipdb on ipython 7.13+

kmaork:
Running the debugger prompt in the default `prompt_toolkit` session causes issues when more than one prompt_toolkit app is running simultaneously. I'm encountering this while trying to debug multiple threads using [madbg](https://github.com/kmaork/madbg). The errors look exactly like those mentioned in #12192. This commit solves this by using the dedicated API from prompt_toolkit.

BTW, this PR combined with #13311 should be enough to close #12192.


Carreau:
Thanks, would it be worth it to also create on in `IPython/terminal/interactiveshell.py`?
And, does it make sens to keep the session around and reuse it in cmdloop?

kmaork:
> would it be worth it to also create on in `IPython/terminal/interactiveshell.py`?

If we want to allow multiple IPython instances to run concurrently in threads, then yes. When researching this bug, I remember finding some old issues with `IPython` being run in threads inside `scrapy`, but can find it now so maybe it was unrelated.

> does it make sens to keep the session around and reuse it in cmdloop?

It is not necessary functionality-wise. Performance-wise, looking at the [source code](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/src/prompt_toolkit/application/current.py#L143), creating the session is very lightweight, and applying it uses the "private" `_current_app_session`, so I wouldn't do something like that anyway.

kmaork:
Hey @Carreau, can you please rereview?

Carreau:
Apologies for the delay. Let's try it.

PR #13890, opened 2023-01-10
let's avoid walrus operator.

tanmoyio:
In this entire ipython this `:=` appears only once which is sufficient enough to create mypy issues for py<3.8. Why do we even need that, let's avoid this. 

Sincerely 
On behalf of entire Graphistry Team

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -184,10 +184,9 @@ def _format_filename(file, ColorFilename, ColorNormal, *, lineno=None):
         ColorScheme's normal coloring to be used.
     """
     ipinst = get_ipython()
-    if (
-        ipinst is not None
-        and (data := ipinst.compile.format_code_name(file)) is not None
-    ):
+    data = ipinst.compile.format_code_name(file)
+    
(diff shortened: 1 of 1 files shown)


Carreau:
Note that we don't officially support `<3.8` since late in 2021, so you may want to upgrade at some point anyway. And the package metadata should prevent non-compatible IPython to install on 3.7.

I'm ok with removing it to have IPython sort-of working on 3.7, but you will likely hit other issues at some point. 

Also note that as IPython follows [nep29](https://numpy.org/neps/nep-0029-deprecation_policy.html), we'll likely officially drop support for Python 3.8 in April.

ivanov:
Hey there, I'm going through old issues and it seems to me that it makes sense to close this one. IPython now supports Python >= 3.10, so though this is the only place where the walrus operator is used, there isn't a good reason to avoid it, and the comment that needed to be addressed before this could be merged went unaddressed for a year. I guess it wasn't "URGENT !!!" afterall. ;) 

Thanks everyone and happy hacking! :bowtie:

PR #13891, opened 2023-01-12
Handle non utf-8 characters during decoding for %%bash

garlandz-db:
Handle non utf-8 characters during decoding for %%bash

Before this would fail for outputs that include non utf-8 characters. For example
```
curl https://www.google.com
```

**Before**
Screen Shot 2023-01-11 at 3 55 26 PM

**After**
Screen Shot 2023-01-11 at 3 55 37 PM


jasongrout (approved):
Can confirm that this fixes the issue noted in the ticket, and the code looks good to me. Thanks!

Issue #13879, opened 2022-12-30
Proposal: Improved multiline editing


mlucool:
IPython supports multi-line edits when its internal logic determines that the initial input line is not a valid standalone statement (e.g. ends with a colon or has unclosed parentheses). Once inside a multi-line edit, the only way to execute it is to go to the end of the line end press enter twice. 

There are two features that could be independently useful:
1) Key binding (e.g. CTRL-ENTER) to execute multi-line input regardless of the location of the cursor. This is useful when one is editing a multi-line input and wants to execute it without navigating to the last line and pressing ENTER twice.
2) Key binding (e.g. SHIFT-ENTER) to force start a multi-line edit (even if the first line is a valid single-line statement).

What do you think?

krassowski:
There are existing key bindings for both of these actions. I agree that the default key-bindings are not intuitive here. I agree that the suggested ones seem to better align with the current expectation of the user.

> (1) Key binding (e.g. CTRL-ENTER) to execute multi-line input regardless of the location of the cursor. This is useful when one is editing a multi-line input and wants to execute it without navigating to the last line and pressing ENTER twice.

This is currently possible with <kbd>Alt</kbd> + <kbd>Enter</kbd> or <kbd>Esc</kbd> + <kbd>Enter</kbd>.

> (2) Key binding (e.g. SHIFT-ENTER) to force start a multi-line edit (even if the first line is a valid single-line statement).

This is currently possible with <kbd>Ctrl</kbd> + <kbd>o</kbd> (but possibly cursor placement could be improved).

Previously, there were technical limitations preventing detection of <kbd>Ctrl</kbd> + <kbd>Enter</kbd> https://github.com/ipython/ipython/issues/2493#issuecomment-264454032 but we should revisit this as it was many years ago.

mlucool:
Maybe we should document this feature better and allow for user customization?

FWIW, I think Ctrl+ are more natural and mirrors what other products do, but that may be considered a breaking change.

suewonjp:
> This is currently possible with Alt + Enter or Esc + Enter

AFAIK, [Alt + Enter is assigned to toggling full windows on Win32 system](https://support.microsoft.com/en-us/topic/alt-enter-switches-between-window-and-full-screen-d1b23daa-644d-3fbf-2f5e-c2c779584db5)

Can you consider choosing a different key combination as default?

Carreau:
[...]

PR #13894, opened 2023-01-16
MAINT: refactor in cleanup in view of #13879 (shortcut configuration).

Carreau:
We will likely need a data-mapping shortcut to filter and commands, this commands (functions), need to not be closures for delayed binding, and we want non-complex filter I'd like to avoid in the configuration having syntax like `(foo | bar) & baz`.


krassowski:
> this commands (functions), need to not be closures for delayed binding

did you see my proposed changes in #13888 which I think go in that direction?

Carreau:
No sorry, I missed it, will have a look it.

PR #13902, opened 2023-01-24
FIX: Update vi bindings to work with nav auttosuggestion.

### LINKED ISSUE #13901: Problems with the new auto suggestion keybindings in #13888
I'm using a git build of IPython and found two problems with autosuggestion related to #13888.

1. In Vi mode, when there is a suggestion displayed, `escape` will only discard the suggestion, but will not enter the Vi normal mode like it usually does. I find it more reasonable if both actions are performed.
2. If I type some words so a suggestion pops up, and then press `left` followed by `right`, there will be one spurious character from the suggestion inserted at the wrong position.

Example (using `|` to indicate cursor position)
```
("123" is the suggestion)
In [1]: a=|123

(after pressing left, the suggestion disappears)
In [1]: a|=

(after pressing right, an extra character 1 is inserted)
In [1]: a1|=
```

Carreau:
1) dismiss the autosuggest on cursor motion
 2) in vi mode, escape already dismiss autosuggestions

closes #13901

### DIFF
--- a/IPython/terminal/shortcuts/__init__.py
+++ b/IPython/terminal/shortcuts/__init__.py
@@ -363,9 +363,9 @@ def create_ipython_shortcuts(shell, for_all_platforms: bool = False) -> KeyBindi
     kb.add("c-right", filter=has_suggestion & has_focus(DEFAULT_BUFFER))(
         auto_suggest.accept_token
     )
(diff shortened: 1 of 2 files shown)


infmagic2047:
This works correctly now, thanks!

PR #13911, opened 2023-01-27
Fix formatting of usage examples for `set_env` magic

deepyaman:
See the difference between how `env` and `set_env` look in the [docs](http://ipython.readthedocs.io/en/stable/interactive/magics.html).

image

image


Carreau:
Thanks !

Issue #13915, opened 2023-01-30
Autoreload fails silently when a comment is on the same line.


vmerckle:
`%autoreload 2 # this line will automatically reload every imported library`

This line will fail silently. It does nothing at all and does not produce any error. Libs will not be auto-reloaded and no error ever appear.

See line 555 of https://github.com/ipython/ipython/edit/main/IPython/extensions/autoreload.py

PR #13916, opened 2023-01-30
Raise error on wrong autreload parameter

vmerckle:
https://github.com/ipython/ipython/issues/13915


krassowski:
This will be better solved by #13774

vmerckle:
I agree, do you need me to do something? I'm not sure how this works. But it's a very important fix, as it can cost hours because there is 0 mention of it on the web.

Carreau:
https://github.com/ipython/ipython/pull/13774 was merged and does this plus a few other things. So closing this one. 

Thanks for taking the time to send a PR, this will be released soon.

PR #13921, opened 2023-02-02
Optional prefix arg for dynamic key completions

### LINKED ISSUE #12420: Custom dict completion with _ipython_key_completions_ should allow access to prefix string
I would love to be able to use dictionary completion to implement a quick-search style feature. It looks as though currently I can implement the _ipython_key_completions_ method, but it does not receive any information about what has been typed after the opening square braces.

Could we have a way to receive the prefix string in the _ipython_key_completions_ method? Something like:
`my_object['str <tab>`
causes `_ipython_key_completions_(prefix: str)` to be triggered, where `prefix == 'str'`

I looked into implementing a custom completer, but couldn't work out how to associate the line context with MyClass without some AST magic. Is there a simple recipe to implement more general completion for dict keys?

TomNicholas:
Closes #12420 by adding an optional `prefix` argument to `_ipython_key_completions_`.

- [x] Closes #12420
- [x] Tests added
- [x] Documentation updated
- [x] User visible changes (including notable bug fixes) are documented in `whats-new.rst`
- [ ] ~~New functions/methods are listed in API~~


TomNicholas:
Q: Should I have done this using a custom completer instead? I couldn't find any documentation on how to write or register custom completers...

TomNicholas:
@meeseeksdev tag tab-completion

krassowski:
If `_ipython_key_completions_` were to be extended, I would think that it might be worth adding a full support for the Matcher API (https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.completer.html#matcher-api). What do you think? What you are proposing would already make the signature compatible with `MatcherAPIv1`, but instead of inspecting the signature we could require users to use a decorator that would mark it with `matcher_api_version = 1` attribute (not sure on this yet).

As for adding a custom completers, you can do this by appending to `IPCompleter.custom_matchers` list (but that does not resolve the relevant object for you).

TomNicholas:
Hi @krassowski -  thanks for the quick reply!

> What do you think?

I'm potentially open to working on generalising this, but this is my first PR to ipython's main codebase, and I would need a bit of guidance.

I've read through `IPython.core.completer.py`, but I'm still not sure exactly what you're suggesting. My understanding is that there is already a `dict_key_matcher`, which implements the Matcher API, and can get custom keys via `._get_keys()` - that's what my PR changes.

Are you suggesting we define a new matcher method on `IPCompleter`? How would that work exactly?

Carreau:
@krassowski You've worked more than me on the completion recently, so I'll follow your decision here.

IPython/core/completer.py
```diff
@@ -2486,12 +2486,22 @@ def python_func_kw_matches(self, text):
...
+
+        if key_completions_getter is not None:
+            # older versions of ipython assumed _ipython_key_completions_ took no arguments
+            if "prefix" in inspect.signature(key_completions_getter).parameters:
+                if prefix is not None:
+                    # strip leading apostrophe from key before passing to ._ipython_key_completions
+                    return key_completions_getter(prefix=prefix[1:])
+                else:
+                    return key_completions_getter(prefix=None)
+
```
krassowski, line 2502: What I was suggesting was that instead of passing a `prefix` string we would pass an entire [`CompletionContext`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.completer.html#IPython.core.completer.CompletionContext), something like:

```python
class ItemCompletionContext(CompletionContext):
    """Completion context for item sub-matchers (providing information about `__getitem__`)"""
    prefix: str
```

```suggestion
            if (
                 hasattr(key_completions_getter, 'matcher_api_version')
                 and key_completions_getter.matcher_api_version == 2
             ):
                context = ItemCompletionContext(
                    key_prefix=key_prefix,
                    full_text=context.full_text,
                    cursor_line=context.cursor_line,
                    cursor_position=context.cursor_position,
                    token=context.token,
               )
                return key_completions_getter(context)['completions']
```

krassowski: As for stripping leading apostrophe: what if the key is a number? Or hex-number? Or raw string? I think we do not want to modify the prefix.

Carreau:

PR #13923, opened 2023-02-03
Fix typo in version8.rst

dkav:


krassowski:
There is already a PR fixing typos in release notes: #13918.

PR #13932, opened 2023-02-07
Replace usage of os.devnull with subprocess.DEVNULL

### LINKED ISSUE #13906: replace usage of os.devnull with subprocess.DEVNULL
I believe os.devnull is used in a few place, where it can be replaced with subprocess.DEVNULL.

Yann-P:
Refactoring one file in the latex library

Closes https://github.com/ipython/ipython/issues/13906

# Code changes

Replaced one occurence of `open(os.devnull)` with using directly the `subprocess.DEVNULL` special value, removing a with/open block along the way.

# User-facing changes

None


Carreau:
The failure is unrelated, and due to a new version of the type checker. 
#13933 wil fix it.

Carreau:
Thanks !

PR #13940, opened 2023-02-12
capture_output does not respect trailing semicolon

nfgferreira:
This change tests whether the Python code inside the `%%capture` magic ends with a `';'`. If it does, the code no longer creates the object that is supposed to store the cell output. If such object already exists, it is deleted.
Please review and suggest changes if you think they can improve the solution. Thanks.


akhmerov:
I'm wondering whether the addition should go into `capture_output` instead. I think it would make sense, and would also enable easier programmatic usage.

nfgferreira:
I think it wouldn't work.  The decision of showing or not the output comes from the parsing of the Python code in the cell. `capture_output` is not aware of the cell contents and therefore does not see the semicolon.

Carreau:
That seam reasonable. 

I'll still think about it a bit, and mere later if no-one has objections.

Carreau:
Thanks !

PR #13941, opened 2023-02-12
Don't emit a trailng newline in base64-encoded data like 'image/png'

xl0:
This addresses the issue highlighted in

https://github.com/jupyter/jupyter_client/issues/930

Since, unlike the latest Jupyter, IPython aims to support python<3.6, I'm using an approach that won't break this compatibility.


xl0:
Looks like current ipython requires python >3.8. I've switched to just passing `newline=fallse` to `b2a_base64` lie is done in `jupyter_clinent`.

Carreau:
Seem reasonable. Let's try.

PR #13944, opened 2023-02-13
Quote jedi in type hint

bollwyvl:
This just quotes the used of `jedi` in a type hint, so that it can import without jedi being installed (or rather, removed, patched, or otherwise made incomplete).


Carreau:
Thanks !

PR #13945, opened 2023-02-13
add a little detail to custom repr method docs

minrk:
These docs are a bit sparse, so I added a couple details that might help implementers:

- explicitly state the return type, rather than only via example
- give some examples of PNG, HTML
- note that formatters shouldn't be sensitive to other formats

related to https://github.com/Textualize/rich/pull/2806 which didn't follow the spec, in part I think because the docs weren't explicit enough about exactly what a formatter should do.

### DIFF
--- a/docs/source/config/integrating.rst
+++ b/docs/source/config/integrating.rst
@@ -26,15 +26,17 @@ Rich display
 
 Custom methods
 ----------------------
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks !

PR #13951, opened 2023-02-24
don't evaluate descriptor

nadavmisgav:
after [checking with jedi API](https://github.com/davidhalter/jedi/pull/1918), I have added a change in the default settings of jedi to avoid evaluating `@property` methods while auto-completing

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -252,6 +252,13 @@ try:
     import jedi.api.helpers
     import jedi.api.classes
     JEDI_INSTALLED = True
+
+    # disable evaluating descriptors
+    from packaging import version
+    if version.parse(jedi.__version__) > version.parse("0.18.2"):
+        jedi.settings.instance_allow_descriptor_getattr = False
+    else:
+        jedi.Interpreter._allow_descriptor_getattr_default = False
 except ImportError:
     JEDI_INSTALLED = False
 
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -255,6 +255,7 @@ try:
 
     # disable evaluating descriptors
     from packaging import version
+
     if version.parse(jedi.__version__) > version.parse("0.18.2"):
         jedi.settings.instance_allow_descriptor_getattr = False
     else:
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,6 +41,7 @@ install_requires =
     pygments>=2.4.0
     stack_data
     traitlets>=5
+    packaging
 
 [options.extras_require]
 black =


krassowski:
Could you condition the setting on [evaluation level configuration](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-Completer.evaluation) of IPython completer?

While this might solve your problem, many users would see this as a downgrade in completion experience.

davidhalter:
It's pretty obvious that this is not what users want and we have had this discussion in the Jedi issue tracker a lot before, it's not what most users want. If a library really does a lot of computations in a property it's probably broken anyway.

For all the people that want to set this individually they should be able to have a workaround.

Please close this.

krassowski:
Thank you for chiming in Dave!

> For all the people that want to set this individually they should be able to have a workaround.

Yes, this is why I added `Completer.evaluation` setting which user can set to e.g. `Completer.evaluation = 'forbidden'` (with the default being `'limited'` which roughly follows the same trade-offs as made in Jedi).

I think that integrating jedi settings with IPython's `Completer.evaluation` config would be of value. As previously stated I do not support changing the default behaviour - I think we have a consensus on this. I think that this PR does not need closing but a rework (but this is just my opinion).

nadavmisgav:
> If a library really does a lot of computations in a property it's probably broken anyway.

Its not strictly related to heavy computation, this could also relate to logging or other side affects.  It is not "obvious" that auto completer would run arbitrary code, and `jedi` does not run the `@property` code when using the `Script.complete` by default.

take for example this,
```python
Python 3.8.10 (default, Jun 22 2022, 20:18:18)
IPython 8.10.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import IPython

In [2]: IPython.core.completer.Completer.evaluation = 'forbidden'

In [3]: class A:
   ...:     @property
   ...:     def foo(self):
   ...:         print("hello")
   ...:         return 4
   ...:

In [4]: f = A()

hello
In [5]: f.<tab>
```

davidhalter:
Don't use side effects in properties. Properties that do complex stuff are really an anti-pattern IMO.

This was really discussed a lot before: See for example https://github.com/davidhalter/jedi/issues/1299. I don't think you are going to convince anyone here without actually trying to convince the people that argued the exact opposite of your position. I was on your side a few years ago, but now think that the current behavior is better.

nadavmisgav:
I read https://github.com/davidhalter/jedi/issues/1299, not sure why you changed your mind from back then, here are two examples of packages who run code with side affect in `@property` methods,
1. https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L609
2. https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/tubes/ssh.py#L2001

the default behavior IMO should be **safe** completion, i.e. not running arbitrary code. maybe we can add completion based on type hints for properties.

davidhalter:
Tried to write something, but realized that it's pointless. I'm not having that argument again. Goodbye.

Carreau:
I don't disagree with any of the sides here, there are pros and cons. 
But those "evaluate" or "don't" evaluate discussion keep going back and forth. 

You can use your code snippet in the IPython configuration file as they are executable to set that option on Jedi ff needs be. 

I'm going to close this as "we are not going to change the defaults again", apologies if it forces you to tweak your configuration file. 

Thanks for sending The PRs, the code snippet will be useful for other that have this issue and need to set this up in their configuration file.

PR #13956, opened 2023-02-27
Fix typos in docs

zrottman:
Brand new to open source, sure hope this is helpful and that I'm doing this right.


Carreau:
Thanks, yes this is great. 

The test are failing but this is due to another bug I'm currently trying to fix. 

Contribution much appreciated!

PR #13957, opened 2023-03-01
Shaperilio/qtgui fixes

shaperilio:
I started using the released version of my `PySide6`-enabling changes and noted some problems. In this PR, I fix those, and also overall improve the feedback to the user when a GUI event loop is hooked in:

 - Report which event loop is running when using `%gui <some GUI>`; e.g. `%gui qt` will show `Installed qt6 event loop hook.`
 - Report when the event loop is disabled; i.e. `%gui` will show `GUI event loop hook disabled.` if an event loop hook was installed, or `No event loop hook running.` if nothing was installed.
 - Requesting a second event loop will give the message `Shell is already running a gui event loop for <some GUI>. Call with no arguments to disable current loop.`
 - Requesting a different version of Qt, i.e. `%gui qt6` followed by `%gui` followed by `%gui qt5` will show `Cannot switch Qt versions for this session; will use qt6.` followed by `Installed qt6 event loop hook.`

(Fixes / improves #13864)

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -915,8 +915,9 @@ class TerminalInteractiveShell(InteractiveShell):
     def enable_gui(self, gui=None):
         if self._inputhook is not None and gui is not None:
             print(
-                f"Shell was already running a gui event loop for {self.active_eventloop}; switching to {gui}."
+                f"Shell is already running a gui event loop for {self.active_eventloop}."
             )
+            return
         if gui and (gui not in {"inline", "webagg"}):
             # This hook runs with each cycle of the `prompt_toolkit`'s event loop.
             self.active_eventloop, self._inputhook = get_inputhook_name_and_func(gui)
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -914,7 +914,7 @@ class TerminalInteractiveShell(InteractiveShell):
     active_eventloop = None
     def enable_gui(self, gui=None):
         if self._inputhook is not None and gui is not None:
-            warn(
+            print(
                 f"Shell was already running a gui event loop for {self.active_eventloop}; switching to {gui}."
             )
         if gui and (gui not in {"inline", "webagg"}):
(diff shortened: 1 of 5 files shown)


shaperilio:
@Carreau @ccordoba12 FYI

IPython/terminal/interactiveshell.py
```diff
@@ -913,10 +913,19 @@ def inputhook(self, context):
...
+        if self._inputhook is None and gui is None:
+            print("No event loop hook running.")
+            return
+
         if self._inputhook is not None and gui is not None:
-            warn(
-                f"Shell was already running a gui event loop for {self.active_eventloop}; switching to {gui}."
+            print(
+                f"Shell is already running a gui event loop for {self.active_eventloop}. "
+                "Call with no arguments to disable current loop."
```
ccordoba12, line 923:
```suggestion
                "Call with no arguments to disable the current loop."
```

IPython/terminal/pt_inputhooks/__init__.py
```diff
@@ -110,7 +111,11 @@ def set_qt_api(gui):
                 del os.environ["QT_API"]
         else:
             print(f'Unrecognized Qt version: {gui}. Should be "qt5", "qt6", or "qt".')
-            return
+            return None
+        # Import it now so we can figure out which version it is.
```
ccordoba12, line 115:
```suggestion
            return

        # Import it now so we can figure out which version it is.
```

`return` is equivalent to `return None`.

ccordoba12 (commented):
A couple of small suggestions for you @shaperilio, otherwise looks good to me.

shaperilio:
> A couple of small suggestions for you @shaperilio, otherwise looks good to me.

Thanks; I committed your suggestions.

ccordoba12 (approved):
Looks good to me now, thanks @shaperilio!

shaperilio:
@Carreau any comments?

Carreau:
I don't have much experience in this part of the codebase, things looks fine to me, and if both of you are happy with it let's go.

PR #13960, opened 2023-03-04
PR: Add needs_local_scope to debug magic

### LINKED ISSUE #13959: %debug magic doesn't capture the locals
To reproduce:
1 - create a file `test.ipy`:
```python
def fun():
    res = 1
    %debug print(res)
fun()
```
2 - run with `ipython test.ipy`
3 - After pressing continue on the pdb promt, I get:
`NameError: name 'res' is not defined`

This is in contrast with the following code, which works:
```python
def fun():
    res = 1
    %timeit print(res)
fun()
```

The difference seems to be that the `timeit` magic is decorated with `@needs_local_scope`

impact27:
Fixes https://github.com/ipython/ipython/issues/13959

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -422,7 +422,8 @@ class ExecutionMagics(Magics):
     )
     @no_var_expand
     @line_cell_magic
-    def debug(self, line='', cell=None):
+    @needs_local_scope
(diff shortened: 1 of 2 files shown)


Carreau:
Seem reasonable, Thanks.

Issue #13860, opened 2022-12-08
[Feature Request] Support for extending documentation using `?`


nishikantparmariam:
IPython's feature to show documentation using `foo.bar?` is very helpful. It would be great feature addition to provide support for  dynamically showing documentation when say `bar` in not an actually property (defined with `@property` and having docstring) or attribute on `foo`, but can be set on `foo`.

##### Solution for showing dynamic tab completions

As an example, for showing tab suggestions `__dir__` can be defined on a class as per https://github.com/ipython/ipython/issues/7245#issuecomment-67199587. But, there does not seem an equivalent to make docstring using `?` also work.

##### Possible solution

A custom hook can be provided (may be with better name) which users can use something like -

```python3
def show_documentation(self, obj, attrname): 
    """
    :return:
         Docstring of attribute `attrname` on object `obj`
    :rtype: 
         str
    """
```

So, if someone does `foo[2].bar.baz?` and if `baz` is not found with `@property` on `bar`, then `show_documentation(self, <bar object (preferably) or class of bar object>, 'baz')` is invoked. 

Additionally, it would be great if after doing say `foo[2].bar.baz = 5` it continues to show the above documentation and not the documentation for `int`.

This is just one solution, open for better ideas!

Carreau:
Apologies for the late reply. 

If my understanding is correct, you want o be able to define 

```python
class Bar:

    def show_documentation(self, obj, attrname): 
        """
        :return:
             Docstring of attribute `attrname` on object `obj`
        :rtype: 
             str
        """
        if obj is self and attrname is 'baz':
           return "this is a baz it is the bazified value of the Bar instance
        
```

Ir you you want the hook to be external:

```python

ip = get_ipython()
ip.register_doc_hook(...)
```

I do have some concerns about the first instance, similar things have longed been discussed, and the consensus is that we do not want to have object-controlled code execution when showing documentation. This is a security risk, and if documentation can start to have side effects it will create other problems, like impossibility to see documentation when the kernel is busy. This is already a problem when even with just `__doc__`.

For the second This is a larger project (https://github.com/jupyter/papyri) I am working on that completely rework how current documentation work, and would allow more dynamic (and html browsable) documentation within notebook. Though this is not a small project. 

I think you are more meaning the second one, but I'm slightly worried of having it be a cusotmizable hook. Can you give me an example of why you would like it to be different on a per object basis ? 

Would something like simply looking up `__<attrname>_doc__` be sufficient for you / why not ?

[...]

PR #13975, opened 2023-03-14
Allow to dispatch getting documentation on objects.

Carreau:
Base for #13860, so that object can be queried for documentation on their fields/properties.

Typically this allows the following, to extend the doc documentation when requesting information on a field.

```python
class DictLike:
    def __getitem__(self, k):
        if k.startswith('f'):
            return "documentation for k"
        else:
            raise KeyError

class Bar:
    __custom_documentations__ = DictLike()

    faz = 1


    @property
    def foo(self):
        return 1
b = Bar()
b.faz?
```


krassowski:
I guess I am late to notice this, but should we also allow `__custom_documentation__` (no `s`) since in common usage documentation is uncountable? Any native speakers with an opinion?

PR #13986, opened 2023-03-24
MAINT/TEST: remove sleep.

Carreau:
Try to remove a sleep that is supposed to fix pyc modification time, though it's old code, and I'm unsure this is still relevant.


Carreau:
Maybe a better way woudl be to look at https://docs.python.org/3/library/py_compile.html and https://docs.python.org/3/library/py_compile.html#py_compile.PycInvalidationMode

PR #13989, opened 2023-03-25
Improve API documentation around configuration of embedded IPython

phil-blain:
Hi ! 

This PR tries to improve the API documentation of `embed()`, `embed_kernel()`, `start_ipython()` and `start_kernel()` with respect to the configuration of IPython by adding links to other sections of the documentation. It also reorganizes the API docs so that `embed()` is listed under `IPython`, just as it's presented in the narrative documentation.

Here is a commit-by-commit breakdown:

First some preparatory patches:

- [PATCH 1/8] docs/autogen_shortcuts.py: support Python 3.8
- [PATCH 2/8] .gitignore: ignore  'docs/source/config/shortcuts/table.tsv'
- [PATCH 3/8] Revert "Update README to use optional dependency over requirements.txt" 
- [PATCH 4/8] docs: refer to 'embed_kernel' in 'start_kernel's docstring

Then the meat of the changes: 

- [PATCH 5/8] docs: add more pointers to configuration of embedded IPython
- [PATCH 6/8] docs: embed: add parameters to docstring
- [PATCH 7/8] docs: add pointers to IPython and kernel options
- [PATCH 8/8] docs: document embed() under IPython

Full justification for each changes are in the commits messages.

### DIFF
--- a/IPython/terminal/embed.py
+++ b/IPython/terminal/embed.py
@@ -369,8 +369,8 @@ def embed(*, header="", compile_flags=None, **kwargs):
         d = 40
         embed()
 
-    Full customization can be done by passing a :class:`Config` in as the
-    config argument.
+    Full customization can be done by passing a traitlets :class:`Config` in as the
+    `config` argument (see :ref:`configure_start_ipython`).
     """
     config = kwargs.get('config')
     if config is None:
--- a/IPython/terminal/embed.py
+++ b/IPython/terminal/embed.py
@@ -350,7 +350,7 @@ class InteractiveShellEmbed(TerminalInteractiveShell):
 def embed(*, header="", compile_flags=None, **kwargs):
     """Call this to embed IPython at the current point in your program.
 
-    The first invocation of this will create an :class:`InteractiveShellEmbed`
+    The first invocation of this will create a :class:`terminal.embed.InteractiveShellEmbed`
     instance and then call it.  Consecutive calls just call the already
     created instance.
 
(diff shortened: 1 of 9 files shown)


IPython/terminal/embed.py
```diff
@@ -350,7 +350,7 @@ def mainloop(
 def embed(*, header="", compile_flags=None, **kwargs):
     """Call this to embed IPython at the current point in your program.
 
-    The first invocation of this will create an :class:`InteractiveShellEmbed`
+    The first invocation of this will create a :class:`terminal.embed.InteractiveShellEmbed`
```
Carreau, line 353: That should be unnecessary, and it seem to work as is with a cross ref, but I have no objections. 

Dit it not work for you ?

docs/README.rst
```diff
@@ -26,7 +26,7 @@ the following tools are needed to build the documentation:
 In a conda environment, or a Python 3 ``venv``, you should be able to run::
 
   cd ipython
-  pip install .[doc] -U
+  pip install -U -r docs/requirements.txt
```
Carreau, line 29: I think this file is here just for readthedocs, user should likely not use it.
Though they may need quotes. Maybe:

```suggestion
  pip install --upgrade '.[doc]' 
```

Carreau:
Thanks, look good, some questions I'll leave for a few days or so and merging after that. 
I'm travelling this week, so apologies if I'm slow to respond/merge.

docs/README.rst
phil-blain: As I note in af5f1156ee8fbff4e30d6f81f5f6f5ff08673d89, if one follows the instructions in `docs/README.rst`, and then to build the API docs, makes changes, and then builds the docs again, the new changes do not show up in the docs build because the package is not installed in editable mode. That is why I think using the `requirements.txt` is better, since it has `-e`.

IPython/terminal/embed.py
phil-blain: It did not work for me under `IPython`, only under `IPython.terminal.embed` (after 853291bb3104475994842f3ea9fda44ee938177d, which tweaks the API docs so that `embed()` appears under the `IPython` module in addition to under `IPython.terminal.embed`).

phil-blain:
Thank you for merging! :)

PR #13990, opened 2023-03-26
Add `delete` binding for clearing auto-suggestion

### LINKED ISSUE #13987: Proposal: discard autosuggestion with `delete` key
Auto-suggestions can be discarded using <kbd>Esc</kbd> key (which works slow as there is a time delay allowing to press a combination of <kbd>Esc</kbd> with another key):

existing-workaround

When an auto-suggestion is partially accepted e.g. using <kbd>Ctrl</kbd> + <kbd>Right</kbd> it would appear intuitive to use <kbd>delete</kbd> key to clear the remaining part of the suggestion, but currently this does not work:

current-behaviour

the proposed behaviour is depicted on the GIF below:

proposed-behaviour

IPython 3.11 users can opt-in for the proposed behaviour by defining a custom shortcut:

```python
custom_shortcuts = [
    {
        "command": "IPython:auto_suggest.discard",
        "new_keys": ["delete"],
        "new_filter": "navigable_suggestions & has_suggestion & default_buffer_focused"
    }
]
%config TerminalInteractiveShell.shortcuts = custom_shortcuts
```

but there may be a value in making this a default behaviour (opt-out rather than opt-in).

Can I ask anyone interested to indicate if they would be happy with an addition of such a shortcut by default (:+1: or :-1: )?

krassowski:
Closes #13987 assuming that we have a consensus.

### DIFF
--- a/IPython/terminal/shortcuts/__init__.py
+++ b/IPython/terminal/shortcuts/__init__.py
@@ -208,6 +208,11 @@ AUTO_SUGGEST_BINDINGS = [
         "has_suggestion & default_buffer_focused & emacs_insert_mode",
     ),
     Binding(
+        auto_suggest.discard,
+        ["delete"],
+        "has_suggestion & default_buffer_focused & emacs_insert_mode",
+    ),
+    Binding(
         auto_suggest.swap_autosuggestion_up,
         ["up"],
         "navigable_suggestions"


krassowski:
Are you in vi or emacs mode? Your `del` removes character and thus behaves like `backspace`.

Carreau:
I should be in emacs mode. Mostly default config for everything.
It behave like so for me both in iTerm, and Terminal.app. Could it be prompt toolkit ? 
I tried with both 3.0.38 and 3.0.37 and got the same.

Carreau:
I get the same behavior in a fresh docker container... This is weird.

Carreau:
Ah! yes, sure. My bad. Delete does not indeed delete on main but does here.

PR #13999, opened 2023-03-29
Update completer documentation

hammer498:
The docstring for suppression says that `do_not_suppress` should be a list when it should actually be a set


hammer498:
@meeseeksdev tag documentation

Carreau:
Thanks !

PR #14010, opened 2023-04-10
Try to fix traceback Code speed

Carreau:
Instead of using the heuristic of where the function is defined we will now try to count the number of line in the files where the function are defined. 

Unfortunately the way stackdata parses things, the time depends on the size of the file as it does not do partial highlighting. 

So if the functions we are using are defined before the threshold we still go through the slow path.

Now we try to :

1) not look at the IPython stacks
2) for files we can, open and count the number of lines as a proxy (and cache it). 
3) if any of those _files_ are too long, go through fast code.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -89,20 +89,22 @@ Inheritance diagram:
 #*****************************************************************************
 
 
+import functools
 import inspect
 import linecache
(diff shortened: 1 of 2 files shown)


Carreau:
meeseeksdev backport to 8.12.x

Issue #14005, opened 2023-04-06
LimitOverrunError: Separator is not found, and chunk exceed the limit


garlandz-db:
I verified that IPython==7.32.0 works fine for the curl command
```
curl -X GET \
     "https://datasets-server.huggingface.co/first-rows?dataset=tatsu-lab%2Falpaca&config=tatsu-lab--alpaca&split=train"
```
but not in IPython==8.10.0 and instead get this error

```
Task exception was never retrieved
future: <Task finished name='Task-342' coro=<ScriptMagics.shebang.<locals>._handle_stream() done, defined at /Users/garland.zhang/mambaforge/envs/testenv/lib/python3.10/site-packages/IPython/core/magics/script.py:211> exception=ValueError('Separator is not found, and chunk exceed the limit')>
Traceback (most recent call last):
  File "/Users/garland.zhang/mambaforge/envs/testenv/lib/python3.10/asyncio/streams.py", line 524, in readline
    line = await self.readuntil(sep)
  File "/Users/garland.zhang/mambaforge/envs/testenv/lib/python3.10/asyncio/streams.py", line 602, in readuntil
    raise exceptions.LimitOverrunError(
asyncio.exceptions.LimitOverrunError: Separator is not found, and chunk exceed the limit

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/garland.zhang/mambaforge/envs/testenv/lib/python3.10/site-packages/IPython/core/magics/script.py", line 213, in _handle_stream
    line = (await stream.readline()).decode("utf8", errors="replace")
  File "/Users/garland.zhang/mambaforge/envs/testenv/lib/python3.10/asyncio/streams.py", line 533, in readline
    raise ValueError(e.args[0])
ValueError: Separator is not found, and chunk exceed the limit

```

jasongrout-db:
For me:

IPython 7.33.0 works fine.

<details><summary>IPython 8.0.1 does not work</summary>

```
In [3]: import IPython
   ...: 
   ...: IPython.__version__
Out[3]: '8.0.1'

In [4]: %%sh
   ...: curl -X GET "https://datasets-server.huggingface.co/first-rows?dataset=tatsu-lab%2Falpaca&config=tatsu-lab--alpaca&split=train"
   ...: 
   ...: 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #14019, opened 2023-04-14
%%bash: read from stream until separator found or buffer is full

garlandz-db:
Original issue: https://github.com/ipython/ipython/issues/14005
TLDR before we would read line by line however this is not necessary and we should just continue reading until we reach EOF. 

The case we patch here is if we read a line larger than the allowed buffer size, we will return that chunk (aka the entire buffer) to be written out and still continue reading 

before:
image

now:
```
 %%bash
   python -c "print('*' * 100000)"
************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************
...
*********************************************************************************************************************************
```

### DIFF
--- a/IPython/core/magics/script.py
+++ b/IPython/core/magics/script.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import asyncio
+import asyncio.exceptions
 import atexit
 import errno
 import os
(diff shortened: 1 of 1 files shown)


IPython/core/magics/script.py
```diff
@@ -208,15 +209,23 @@ def in_thread(coro):
             """Call a coroutine on the asyncio thread"""
             return asyncio.run_coroutine_threadsafe(coro, event_loop).result()
 
+        async def _readchunk(stream):
+            try:
+                return await stream.readuntil(b"\n")
```
garlandz-db, line 214: \n is by default but i think its more obvious what we're doing here by just specifying it. though not hard opinion

jasongrout (approved):
Thanks @garlandz-db! This code looks good to me, and it also fixes the issue for me when I try it out.

I'll leave the decision about merging up to others since we work at the same company, but +1 from me!

It does look like the doc building test failure is unrelated.

PR #14022, opened 2023-04-18
Fix failing docs build.

Carreau:
Add documentation to two __init__ that otherwise fail the docs build as the parent __init__ reference a section in pygments docs.


Carreau:
meeseeksdev backport to 8.12.x

PR #14029, opened 2023-04-22
Allow safe access to the `__getattribute__` method of modules

### LINKED ISSUE #14028: IPython can't provide completions for second level modules when Jedi completions are disabled
This problem was reported in [Spyder](https://github.com/spyder-ide/spyder/issues/20381#issuecomment-1419243534) and I was able to reproduce it locally. Basically, if you run

```python
In [1]: import numpy as np

In [2]: np.linalg.<Tab>
```

you won't get completions for the `linalg` module.

After a bit of testing, I found that this problem doesn't show up if Jedi completions are enabled. But we have them disabled by default in Spyder because they introduce serious lags when working with large dataframes.

It'd be nice if this could be fixed for the next IPython release and I'm able to put some time to do it. @krassowski, could you let me know where to start looking for a possible solution? Thanks!

ccordoba12:
This restores code completions of second-level modules when Jedi completions are disabled.

Fixes #14028

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -637,6 +637,7 @@ set_non_mutating_methods = set(dir(set)) & set(dir(frozenset))
 
 dict_keys: Type[collections.abc.KeysView] = type({}.keys())
 method_descriptor: Any = type(list.copy)
+module = type(builtins)
 
 NUMERICS = {int, float, complex}
 
@@ -686,6 +687,7 @@ BUILTIN_GETATTR: Set[MayHaveGetattr] = {
     *NUMERICS,
     dict_keys,
     method_descriptor,
+    module,
 }
 
 
--- a/IPython/core/tests/test_guarded_eval.py
+++ b/IPython/core/tests/test_guarded_eval.py
@@ -568,3 +568,15 @@ def test_assumption_named_tuples_share_getitem():
         pass
 
     assert A.__getitem__ == B.__getitem__
+
+
+@dec.skip_without("numpy")
+def test_module_access():
+    import numpy
+
+    context = limited(numpy=numpy)
+    assert guarded_eval("numpy.linalg.norm", context) == numpy.linalg.norm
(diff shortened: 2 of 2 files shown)


ccordoba12:
@krassowski, I don't understand very well how to add a test for this, so I'd appreciate your guidance for that.

krassowski:
Maybe something like:

```python
def test_module_access():
    import sys
    context = limited(sys=sys)
    assert guarded_eval("sys.copyright", context) == sys.copyright

    context = minimal(sys=sys)
    with pytest.raises(GuardRejection):
        guarded_eval("sys.copyright", context)
```

in `IPython/core/tests/test_guarded_eval.py`?

ccordoba12:
Thanks @krassowski! I implemented the same test but using Numpy because it's a more important use case.

krassowski (approved):
LGTM, thank you @ccordoba12!

IPython/core/guarded_eval.py
```diff
@@ -637,6 +637,7 @@ def _list_methods(cls, source=None):
 
 dict_keys: Type[collections.abc.KeysView] = type({}.keys())
 method_descriptor: Any = type(list.copy)
+module = type(builtins)
```
Carreau, line 640: Hum, 

I missed that in the previous PR but those should likely be :

```
from types import MethodDescriptorType
from types import ModuleType
```

Not blocking so let's open as another issue.

krassowski: Oh, `MethodDescriptorType` is neat. I missed it being added in 3.7, but since we are not supporting 3.6 for some time now this is good idea indeed.

ccordoba12: I can take care of this.

ccordoba12:
@Carreau, could this fix be backported to the `8.12` branch? We still have a lot of people that uses Python 3.8 in Spyder, so we'd like them to benefit from this.

Issue #13484, opened 2022-01-22
Counters should display from most common to least common


rhettinger:
IPython is not matching Python's decision to display *Counter* objects arranged by descending count. 

What regular Python does for repr and pprint:

```
$ python
Python 3.10.2 (v3.10.2:a58ebcc701, Jan 13 2022, 14:50:16) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import Counter
>>> from pprint import pprint
>>> hist = Counter(red=3813, orange=4213, yellow=2024, green=5821, blue=1181, indigo=2350, violet=1337, white=8675, block=3099)
>>> hist
Counter({'white': 8675, 'green': 5821, 'orange': 4213, 'red': 3813, 'block': 3099, 'indigo': 2350, 'yellow': 2024, 'violet': 1337, 'blue': 1181})
>>> pprint(hist)
Counter({'white': 8675,
         'green': 5821,
         'orange': 4213,
         'red': 3813,
         'block': 3099,
         'indigo': 2350,
         'yellow': 2024,
         'violet': 1337,
         'blue': 1181})
```

What IPython does:

```
$ ipython
Python 3.10.2 (v3.10.2:a58ebcc701, Jan 13 2022, 14:50:16) [Clang 13.0.0 (clang-1300.0.29.30)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.0.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %config TerminalInteractiveShell.autoformatter=None

In [2]: from collections import Counter

In [3]: hist = Counter(red=3813, orange=4213, yellow=2024, green=5821, blue=1181, indigo=2350, violet
   ...: =1337, white=8675, block=3099)
   ...:

In [4]: hist
Out[4]:
Counter({'red': 3813,
         'orange': 4213,
         'yellow': 2024,
         'green': 5821,
         'blue': 1181,
         'indigo': 2350,
         'violet': 1337,
         'white': 8675,
         'block': 3099})

In [5]: from pprint import pprint

In [6]: pprint(hist)
Counter({'white': 8675,
         'green': 5821,
         'orange': 4213,
         'red': 3813,
         'block': 3099,
         'indigo': 2350,
         'yellow': 2024,
         'violet': 1337,
         'blue': 1181})
```

Carreau:
Hum, it's likely because they are subclasses of dict.
[...]

PR #14032, opened 2023-04-24
Added sorting on counter and passed all tests

jonotassia:
Creating a pull request to address changes requested in #13484 to update logic for printing Counter objects in terminal. This is now matching the standard Python console and pprint outputs which sort by default in a descending order.

image

The following function was updated to add sorting:

```
def _counter_pprint(obj, p, cycle):
    cls_ctor = CallExpression.factory(obj.__class__.__name__)
    if cycle:
        p.pretty(cls_ctor(RawText("...")))
    elif len(obj):
        p.pretty(cls_ctor(dict(sorted(obj.items(), key=lambda x: x[1], reverse=True))))
    else:
        p.pretty(cls_ctor())
```

Additionally, I have added an additional test case to test_collections_counter in test_pretty.py to test for descending order:

```
def test_collections_counter():
    class MyCounter(Counter):
        pass
    cases = [
        (Counter(), 'Counter()'),
        (Counter(a=1), "Counter({'a': 1})"),
        (MyCounter(a=1), "MyCounter({'a': 1})"),
        (Counter(a=1, c=22), "Counter({'c': 22, 'a': 1})")
    ]
    for obj, expected in cases:
        assert pretty.pretty(obj) == expected
```

======================== 34 passed, 1 skipped in 0.13s ========================

Finally, all tests have been passed after making this change.

### DIFF
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -918,7 +918,7 @@ def _counter_pprint(obj, p, cycle):
     if cycle:
         p.pretty(cls_ctor(RawText("...")))
     elif len(obj):
-        p.pretty(cls_ctor(dict(obj)))
+        p.pretty(cls_ctor(dict(sorted(obj.items(), key=lambda x: x[1], reverse=True))))
     else:
         p.pretty(cls_ctor())
 
--- a/IPython/lib/pretty.py
+++ b/IPython/lib/pretty.py
@@ -918,7 +918,7 @@ def _counter_pprint(obj, p, cycle):
     if cycle:
         p.pretty(cls_ctor(RawText("...")))
     elif len(obj):
-        p.pretty(cls_ctor(dict(sorted(obj.items(), key=lambda x: x[1], reverse=True))))
+        p.pretty(cls_ctor(dict(obj.most_common())))
     else:
         p.pretty(cls_ctor())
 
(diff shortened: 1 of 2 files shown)


IPython/lib/pretty.py
```diff
@@ -918,7 +918,7 @@ def _counter_pprint(obj, p, cycle):
     if cycle:
         p.pretty(cls_ctor(RawText("...")))
     elif len(obj):
-        p.pretty(cls_ctor(dict(obj)))
+        p.pretty(cls_ctor(dict(sorted(obj.items(), key=lambda x: x[1], reverse=True))))
```
Carreau, line 921: Ho ! That was an easy update ! 

Isn't `sorted(obj.items(), key=lambda x: x[1], reverse=True)` simply `obj.most_common()` ? which might be simpler or more readable ?

```suggestion
        p.pretty(cls_ctor(dict(obj.most_common())))
```

I'm guessing this will also get rig of the linting failure that is likely to complain line is too long.

jonotassia: Great suggestion (good one for me to know for the future as well)! 

I appreciate you looking at it. I've just added that change and re-run the tests and it appears to all be working still.

jonotassia:
Not urgent, but I'm noticing that the formatting checks were failing because of "Lint with Darker" when I don't have a comma at the end of the last element in the list. I went ahead and made that change, but I was just curious why we needed that comma there?

image

Carreau:
Thanks.

jonotassia:
Great, thanks for that explanation. Makes sense to me!

PR #14039, opened 2023-04-26
Display exception notes in tracebacks

EtiennePelletier:
[PEP 678](https://peps.python.org/pep-0678/) introduced the ability to add notes to exception objects. This has been [released in Python 3.11](https://docs.python.org/3/library/exceptions.html#BaseException.add_note) and is currently not implemented in IPython. These changes are fully compatible with older Python versions that don't include PEP 678.

Here's a sample test that shows the consistency in Python's stdlib traceback module (test 1) and the difference between Python and IPython's runtimes (test 2):
```python
import traceback

print('--- test 1 ---')
try:
    raise Exception('Testing notes')
except Exception as e:
    e.add_note('Does this work?')
    e.add_note('Yes!')
    traceback.print_exc()

print('\n--- test 2 ---')
try:
    raise Exception('Testing notes')
except Exception as e:
    e.add_note('Does this work?')
    e.add_note('No!')
    raise
```

When executed with Python 3.11, both notes are displayed in both tracebacks:
```
$ python test.py 
--- test 1 ---
  File "/app/test.py", line 5, in <module>
    raise Exception('Testing notes')
Exception: Testing notes
Does this work?
Yes!

--- test 2 ---
  File "/app/test.py", line 13, in <module>
    raise Exception('Testing notes')
Exception: Testing notes
Does this work?
No!
```

In IPython's VerboseTB does not yet handle exception notes:
```
$ ipython test.py 
--- test 1 ---
  File "/app/test.py", line 5, in <module>
    raise Exception('Testing notes')
Exception: Testing notes
Does this work?
Yes!

--- test 2 ---
File /app/test.py:13
     11 print('\n--- test 2 ---')
     12 try:
---> 13     raise Exception('Testing notes')
     14 except Exception as e:
     15     e.add_note('Does this work?')

Exception: Testing notes
```

The changes I am suggesting are inspired from implementation of [Lib/traceback.py](https://github.com/python/cpython/blob/main/Lib/traceback.py) (search for `__notes__`) and improvements for dealing with edge cases more nicely in [cpython#103897](https://github.com/python/cpython/pull/103897).

Although notes are meant to be strings only, I kept some inspiration from the existing exception handling to ensure that the notes are uncolored and bytes decoded, if there are any. I am definitely open to using a different color if deemed better. For context, `bpython` keeps the notes uncolored, and [Python's tutorial](https://docs.python.org/3/tutorial/errors.html#enriching-exceptions-with-notes) puts them in light gray, like the line numbers.

Here's how the test 2 looks like after these changes:
image

## :snake: :man_juggling:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ __pycache__
 *.swp
 .pytest_cache
 .python-version
+.venv*/
 venv*/
 .mypy_cache/
 
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -51,24 +51,24 @@ def recursionlimit(frames):
 class ChangedPyFileTest(unittest.TestCase):
     def test_changing_py_file(self):
         """Traceback produced if the line where the error occurred is missing?
-        
+
         https://github.com/ipython/ipython/issues/1456
         """
         with TemporaryDirectory() as td:
             fname = os.path.join(td, "foo.py")
             with open(fname, "w", encoding="utf-8") as f:
                 f.write(file_1)
-            
+
             with prepended_to_syspath(td):
                 ip.run_cell("import foo")
-            
+
             with tt.AssertPrints("ZeroDivisionError"):
                 ip.run_cell("foo.f()")
-            
+
             # Make the file shorter, so the line of the error is missing.
             with open(fname, "w", encoding="utf-8") as f:
                 f.write(file_2)
-            
+
             # For some reason, this was failing on the *second* call after
             # changing the file, so we call f() twice.
             with tt.AssertNotPrints("Internal Python error", channel='stderr'):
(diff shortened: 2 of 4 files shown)


Zac-HD:
Oops, I checked for existing issues a few hours before this was opened and then created https://github.com/ipython/ipython/pull/14040 - it has tests and a changelog, but is a bit less careful about edge cases.  Steal whatever you want and then I'll close it?

Carreau:
I'm guessing you both are at PyCon sprints !? Enjoy ! I'll likely get that in during May :-).

EtiennePelletier:
Changes cherry-picked, good catch Zac on the Plain traceback format!

.gitignore
```diff
@@ -27,6 +27,7 @@ __pycache__
 *.swp
 .pytest_cache
 .python-version
+.venv*/
```
Carreau, line 30: You likely want that in your [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) to not have to do it for all your repos.

Carreau:
Let's try this. Sorry I was mostly travelling last month.

PR #14046, opened 2023-04-28
MAINT: update imports and make run slow test only on latest Python

Carreau:
This should reduce the total CI time a bit


IPython/core/tests/test_completer.py
```diff
@@ -28,6 +30,8 @@
     completion_matcher,
     SimpleCompletion,
     CompletionContext,
+    _unicode_name_compute,
+    _UNICODE_RANGE,
```
Carreau, line 34:
```suggestion
    _UNICODE_RANGES,
```

PR #14052, opened 2023-04-28
Only enable "resume hinting" with right arrow if at the EOL

### LINKED ISSUE #14051: Right arrow broken in 8.12.1
The -> arrow doesn't work. When I press the <- arrow to go back and edit the current command, Ipython ignores it.

I don't think my explanation is very clear, so here is an example. If I type:

```
In [1]: print(A
```
And realise I forgot to open the quotes, I go back,

```
In [1]: print("|A
```
Now I am stuck, and cannot "hop" over A. I can keep moving backwards, though. Pressing `end` does work.
 
I have replicated it in an empty virtualenv under both Python 3.9 and 3.10.  The problem disappears downgrading to Ipython 8.12.0.
I am using Fedora Linux, with Qterminal 1.1.0.

krassowski:
Fixes #14051

### DIFF
--- a/IPython/terminal/shortcuts/__init__.py
+++ b/IPython/terminal/shortcuts/__init__.py
@@ -277,7 +277,9 @@ AUTO_SUGGEST_BINDINGS = [
     Binding(
         auto_suggest.resume_hinting,
         ["right"],
-        "default_buffer_focused & emacs_like_insert_mode",
+        "is_cursor_at_the_end_of_line"
+        " & default_buffer_focused"
+        " & emacs_like_insert_mode",
     ),
 ]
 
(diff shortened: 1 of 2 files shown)


2sn:
@krassowski How can I disable this new hinting and right-arrow feature?  My apologies, whereas it may look nice, I find it rather unsuitable for practical work.

krassowski:
What specifically do you find unsuitable for practical work? If this is about https://github.com/ipython/ipython/issues/14051 issue, this PR fixes it and I believe @Carreau will look at releasing a bug fix soon. I apologise but there is not much I can do here.

sebcampos (commented):
sweet

krassowski:
Every shortcut can be changed to your liking since https://github.com/ipython/ipython/pull/13928, with documented [`TerminalInteractiveShell.shortcuts`](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-TerminalInteractiveShell.shortcuts) traitlet, [list of shortcuts](https://ipython.readthedocs.io/en/stable/config/shortcuts/index.html#terminal-shortcuts-list).

PR #14056, opened 2023-04-28
Add option to `%autoreload` to hide errors when reloading code

ccordoba12:
* We have `%autoreload` enabled by default in Spyder and think it's a bit annoying to show its error messages to users because they make little sense to them. These errors are not uncommon when you are working with some code that is slightly different between two git branches.
* However, I didn't change the current behavior (i.e. showing errors) because it's been like that for as long as I can remember. We'd simply use the new `--hide-errors` option in our kernel.


Carreau (approved):

Carreau:
Make sens ! Thanks !

Issue #14006, opened 2023-04-06
Installed qt5 event loop hook.


forallsunday:
When I use `plt.show()` , and run the script,
``` bash
❯ & D:/Anaconda3/python.exe -m IPython --no-autoindent d:/Documents/C-Project/GeoLocOptim/scripts/estimate.py
```

The following message showed:
```
Installed qt5 event loop hook.
Shell is already running a gui event loop for qt5. Call with no arguments to disable the current loop.
```

What does it mean and how to disable this?

nodatapoints:
I think this is due to #13957.

francescoboc:
I agree with @nodatapoints , I would like to suppress the message `Installed qt5 event loop hook.`.

ccordoba12:
@shaperilio, how do you think we should address this issue?

shaperilio:
Is there a global "debug" flag?

The print statements were added precisely because silent failures were happening before; i.e. you run something which is forcing a `qt` event loop hook without asking, and then you muck around with some `tk` stuff and nothing works.

Happy to help make this quieter, and maybe work on something that allows multiple event loop hooks to run in sequence, which I believe should be possible as well.

ccordoba12:
I think the problem is that Matplotlib sets the Qt backend by default in some cases, which generates this message in IPython when users try to set the backend again.

So, perhaps the message shouldn't be shown when the backend you're trying to set is the same one that you already have.

shaperilio:
> So, perhaps the message shouldn't be shown when the backend you're trying to set is the same one that you already have.

This was certainly the intent. Let me take a look at what's going on.

@forallsunday can you post relevant package versions?

nodatapoints:
> @forallsunday can you post relevant package versions?

I run IPython 8.12.0 with Qt5 on Gentoo. In my configuration I set
```c.TerminalIPythonApp.matplotlib = 'qt5'```
otherwise I use default settings except for some prompt configs.

shaperilio:
> I run IPython 8.12.0 with Qt5 on Gentoo. In my configuration I set `c.TerminalIPythonApp.matplotlib = 'qt5'` otherwise I use default settings except for some prompt configs.

What is `c`?

ccordoba12:
It's the IPython config object. See here for more details:

https://ipython.readthedocs.io/en/stable/config/intro.html

ccordoba12:
I think ignoring warnings for the same event loop is fine for now.

[...]

PR #14067, opened 2023-05-02
Shaperilio/gui msg

shaperilio:
Addresses the annoying extra messages as detailed in #14006.

# Behavior
The intent is to:
- Communicate when the user installs a hook
- Communicate when the user tries to install a different hook and cannot.
- Communicate when the user uninstalls a hook
- *Not* communicate when the user tries to install the same hook multiple times, with special care for the "latest version" `'qt'` hook request.

Take a system with both `PyQt5` and `pyside6` installed:
```
(venv) > ipython
Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
IPython 8.14.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: %gui qt
Installed qt6 event loop hook.

In [2]: %gui qt5
Cannot switch Qt versions for this session; will use qt6.
Installed qt6 event loop hook.

In [3]: %gui qt

In [4]: %gui qt6

In [5]: %gui tk
Shell is already running a gui event loop for qt6. Call with no arguments to disable the current loop.

In [6]: %gui
GUI event loop hook for tk removed.

In [7]: %gui tk
Installed tk event loop hook.

In [8]: %gui qt
Shell is already running a gui event loop for tk. Call with no arguments to disable the current loop.

In [9]: %gui qt6
Shell is already running a gui event loop for tk. Call with no arguments to disable the current loop.

In [10]: %gui qt5
Shell is already running a gui event loop for tk. Call with no arguments to disable the current loop.

In [11]: %gui
GUI event loop hook for qt5 removed.

In [12]: %gui tk
Installed tk event loop hook.

In [13]: %gui
GUI event loop hook for tk removed.

In [14]: %gui qt5
Cannot switch Qt versions for this session; will use qt6.
Installed qt6 event loop hook.

In [15]: 
```

This means when you run `ipython --matplotlib=tk`, for example, you'll see this:
```
Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
IPython 8.14.0.dev -- An enhanced Interactive Python. Type '?' for help.
Installed tk event loop hook.

In [1]:
```

# Notes
Working on this has uncovered some interesting dynamics; `qt` is taken to mean "the latest Qt available" and the current code appears to force, in some cases, what is effectively a `%gui qt` command before command line or config args are processed (see [this issue](https://github.com/ipython/matplotlib-inline/issues/25)), so that executing

```
ipython --matplotlib=qt5
```

effectively runs
```
%gui qt
%gui qt5
```

and if both `pyside6` and `PyQt5` are installed, the `%gui qt5` will trigger a message about not being able to change Qt versions in the session.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3607,6 +3607,9 @@ class InteractiveShell(SingletonConfigurable):
 
         # Now we must activate the gui pylab wants to use, and fix %run to take
         # plot updates into account
+
+        # NOTE: the call to `enable_gui` below does not seem necessary. `pt.activate_matplotlib`
+        # calls `pyplot.install_repl_displayhook()` when switching backends.
         self.enable_gui(gui)
         self.magics_manager.registry['ExecutionMagics'].default_runner = \
             pt.mpl_runner(self.safe_execfile)
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3585,7 +3585,9 @@ class InteractiveShell(SingletonConfigurable):
             make sense in all contexts, for example a terminal ipython can't
             display figures inline.
         """
-        from matplotlib_inline.backend_inline import configure_inline_support
+        # NOTE: this import triggers `enable_gui('qt')` and occurs before the users' config (e.g.
+        # `--matplotlib=qt5`) is procssed.
+        # from matplotlib_inline.backend_inline import configure_inline_support
 
         from IPython.core import pylabtools as pt
         gui, backend = pt.find_gui_and_backend(gui, self.pylab_gui_select)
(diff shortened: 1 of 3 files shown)


IPython/core/interactiveshell.py
```diff
@@ -3585,7 +3585,9 @@ def enable_matplotlib(self, gui=None):
             make sense in all contexts, for example a terminal ipython can't
             display figures inline.
         """
-        from matplotlib_inline.backend_inline import configure_inline_support
+        # NOTE: this import triggers `enable_gui('qt')` and occurs before the users' config (e.g.
+        # `--matplotlib=qt5`) is procssed.
+        # from matplotlib_inline.backend_inline import configure_inline_support
```
shaperilio, line 3590: Removing this import (and the function call that requires it) prevents the unnecessary installation of a `qt` event loop hook, and as far as I can tell, has no negative effects - I still get inline figures in a Jupyter notebook, for example. I'll need someone with more expertise to chime in on this.

Carreau:

PR #14080, opened 2023-05-17
Add pass-through filter for shortcuts

### LINKED ISSUE #14070: cannot move cursor right
when I enter ipython and type something

I found that I cannot move cursor right using the `->` key unless I press `Esc`

but I need to press 1 `Esc` and move 1 letter right and so on...

and the function of pressing `->` and use the code completion is also broken.

krassowski:
This is one way to fix https://github.com/ipython/ipython/issues/14070.

The alternatives are:
- try to change upstream prompt-toolkit to fire all matching keybindings, not only the last one [as it does currently](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/07412a37b38dd70a3d63f2966a21f0a645133264/src/prompt_toolkit/key_binding/key_processor.py#L187)
    - this would be breaking so we would need to wait for a major release if this would be accepted
    - _unless_ instead of firing all matching keybindings only keybindings which oped in via a flag would be additionally fired (there is a precedent of `eager` flag)
- add a tailored function with logic that figures out what to do based on state of application - I am not a fan of this as these are hard to test already.

### DIFF
--- a/IPython/terminal/shortcuts/__init__.py
+++ b/IPython/terminal/shortcuts/__init__.py
@@ -279,7 +279,8 @@ AUTO_SUGGEST_BINDINGS = [
         ["right"],
         "is_cursor_at_the_end_of_line"
         " & default_buffer_focused"
-        " & emacs_like_insert_mode",
+        " & emacs_like_insert_mode"
+        " & pass_through",
     ),
 ]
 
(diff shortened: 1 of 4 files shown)


albireox:
I can confirm this fixes the issue for me with Python 3.11.3.

ashtuchkin:
Nice! Is it possible to backport this to 8.12.x? We're stuck with Python 3.8 for now.

Carreau:
> Nice! Is it possible to backport this to 8.12.x? We're stuck with Python 3.8 for now.

let's see ...

@meeseeksdev backport to 8.12.x

Carreau:
Backport seem to be clean, so I'll likely make a release by end of month.

In general feel free to backport and send a PR on older branches. I'm likely to miss notification on closed issues/PRs.

PR #14084, opened 2023-05-22
Try fixing the nightly wheels action

martinfleis:
Trying to fix the syntax of the nightly wheels action. The linter was also complaining about missing `jobs:` so added that as well.

cc @Carreau


martinfleis:
Ah, it seems @Carreau was faster and done the same in dde5a72

PR #14085, opened 2023-05-22
fix nightly action repo name

martinfleis:
Fixing the action repo name.


martinfleis:
@Carreau this should make it work.

PR #14091, opened 2023-06-02
Fix syntax error 3.12

Carreau:


pablogsal:
@Carreau I have been investigating if we can do something upstream to help here. I did some experiments and seems that if we keep the same error types and we adapt some of the text of the errors (as you are comparing the text of the error themselves in several parts) we can reduce substantially the failures. Also, there are some parts of IPython that are not testing what they should be (they are assuming that invalid code is valid) so after fixing that this is what's failing:

```
FAILED IPython/core/tests/test_inputsplitter.py::InputSplitterTestCase::test_check_complete - AssertionError: Tuples differ: ('incomplete', 2) != ('incomplete', 0)
FAILED IPython/core/tests/test_inputsplitter.py::IPythonInputTestCase::test_syntax_multiline - AssertionError: "get_ipython().run_line_magic('pinfo', 'foo')\nbar'''" != "'''foo?\nbar'''"
FAILED IPython/core/tests/test_inputsplitter.py::IPythonInputTestCase::test_syntax_multiline_cell - AssertionError: "get_ipython().run_line_magic('pinfo', 'foo')\nbar'''" != "'''foo?\nbar'''"
FAILED IPython/core/tests/test_inputsplitter.py::test_find_next_indent - AssertionError: 5 != 4 (expected)
FAILED IPython/core/tests/test_inputtransformer2.py::test_check_make_token_by_line_never_ends_empty - IndexError: list index out of range
FAILED IPython/core/tests/test_inputtransformer2.py::test_find_assign_system - AttributeError: 'NoneType' object has no attribute 'start_line'
FAILED IPython/core/tests/test_inputtransformer2.py::test_check_complete_param[(\n))-incomplete-0] - AssertionError: assert ('invalid', None) == ('incomplete', 0)
FAILED IPython/core/tests/test_inputtransformer2.py::test_check_complete_param[a \\ -invalid-None] - AssertionError: assert ('incomplete', 0) == ('invalid', None)
FAILED IPython/core/tests/test_inputtransformer2.py::test_check_complete - AssertionError: assert ('complete', None) == ('incomplete', 2)
FAILED IPython/core/tests/test_interactiveshell.py::InteractiveShellTestCase::test_var_expand_local - KeyError: 'test'
FAILED IPython/core/tests/test_interactiveshell.py::InteractiveShellTestCase::test_var_expand_self - KeyError: 'cTest'
```

The last two seem unrelated to the tokenizer (I tested by reverting the `tokenize` changes and they still fail).

For the rest, seems that some are just positions of the errors due to the fact that we add some `NL` token before the `EOF` error is raised while the other tokenizer does not (that we cannot fix, but should be easy enough to adapt).

The others probably require some investigation.

Carreau:
Many thinks for pitching in.

I would appreciate any help, and I was not able to spend much time investigating so I'll trust what you are saying. My main problem is the time I can spend on IPython is becoming a bit more rare every month.

Some of that code also likely pre-date me joining the IPython team, so I would have to also figure out what was meant.

pablogsal:
@Carreau probably we should close this as https://github.com/ipython/ipython/pull/14107 addresses these issues already (and more)

PR #14107, opened 2023-06-29
Fix issues due to breaking tokenize changes in 3.12

lysnikolaou:
Hey 👋 

I'm opening this PR to start a discussion on which of these fixes are valid or not. I may have cut some corners in some places, mostly because of lack of knowledge or context.

After we've fixed the issues and CI is green, we can have a bigger discussion on whether to go foward with the `tokenize` module or some other tokenizer.

### DIFF
--- a/IPython/core/inputsplitter.py
+++ b/IPython/core/inputsplitter.py
@@ -44,6 +44,7 @@ from IPython.core.inputtransformer import (leading_indent,
                                            assign_from_system,
                                            assemble_python_lines,
                                            )
+from IPython.utils import tokenutil
 
 # These are available in this module for backwards compatibility.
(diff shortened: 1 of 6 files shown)


Carreau:
Many thanks for this PR, I'll try to have a look soon. My time is a bit limited but I'll try to carve some time out.

pablogsal:
@Carreau I would recommend reviewing this PR before beta 4 so we can address anything else before the release candidate is released so we can iterate if something is missing.

Carreau:
I dont' see anything obviously wrong, let's try to get that in to see if any user complains. 

Thanks, and sorry I only have a few hours on IPython here and there.

PR #14108, opened 2023-06-30
fall back to native when printing an ExceptionGroup

### LINKED ISSUE #13753: `ultratb` hides the contents of `ExceptionGroup` errors
This probably isn't too bad yet, but Python 3.11 finaly will be out next month, and `ExceptionGroup` is already seeing growing usage across the ecosystem - for example in Pytest, Hypothesis, and Trio.  Ideally we'd get this working soon, so that users who only rarely update are likely to have a compatible set of packages before they hit this interaction.

As a simple reproducing example:

```python
try:
    ExceptionGroup
except NameError:
    from exceptiongroup import ExceptionGroup

# Realistically these errors would have their own nested tracebacks, but for simplicity...
raise ExceptionGroup("Multiple errors", [ValueError("msg"), TypeError("msg")])
```
In the standard Python REPL, this shows:
```
>>> raise ExceptionGroup("Multiple errors", [ValueError("msg"), TypeError("msg")])
  + Exception Group Traceback (most recent call last):
  |   File "<stdin>", line 1, in <module>
  | exceptiongroup.ExceptionGroup: Multiple errors (2 sub-exceptions)
  +-+---------------- 1 ----------------
    | ValueError: msg
    +---------------- 2 ----------------
    | TypeError: msg
    +------------------------------------
```
but in IPython, or a Jupyter notebook, we only see the container exception:
```python-traceback
In [2]: try:
   ...:     ExceptionGroup
   ...: except NameError:
   ...:     from exceptiongroup import ExceptionGroup
   ...: 
   ...: raise ExceptionGroup("Multiple errors", [ValueError("msg"), TypeError("msg")])
ExceptionGroup                            Traceback (most recent call last)
Input In [3], in <cell line: 6>()
      3 except NameError:
      4     from exceptiongroup import ExceptionGroup
----> 6 raise ExceptionGroup("Multiple errors", [ValueError("msg"), TypeError("msg")])

ExceptionGroup: Multiple errors (2 sub-exceptions)
```

jakkdl:
Fixes #13753 

I started looking into delving deeper into ultratb to add support for Exception Groups, but felt that this would be enough of an improvement over the status quo that it'd probably be worth implementing on it's own.
I also looked quite a bit at ipyparallel's `CompositeError` https://github.com/ipython/ipyparallel/blob/main/ipyparallel/error.py#L169, which seems like the equivalent of exception groups.

Remaining TODO:
- [x] add tests
- [x] add documentation

Will go ahead with those if the basic approach is deemed acceptable

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -112,6 +112,8 @@ try:
 except ImportError:
     sphinxify = None
 
+if sys.version_info[:2] < (3, 11):
+    from exceptiongroup import BaseExceptionGroup
 
 class ProvisionalWarning(DeprecationWarning):
     """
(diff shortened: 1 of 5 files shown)


Carreau:
Apologies for the delay in reviewing I don't have much time on IPython.


Don't worry about the downstream failure.

jakkdl:
No worries! I noticed you said the same in the other PRs so realized I should just go full steam ahead and implement the full PR so you only needed a single review pass :)

PR #14119, opened 2023-07-22
Fix: change parso logging level

### LINKED ISSUE #10946: Unexpected logging outputs after setting log level to debug.
I got unexcepted outputs to console after set log level to debug while using tab to autocomplete, and it breaks my typing.
Here's my screenshot:
ipython-logging

Reproduce:
```
In [1]: import logging

In [2]: logging.getLogger().setLevel(logging.DEBUG)

In [3]: logging.info("xxxx")
INFO:root:xxxx

In [4]: logging.<TAB>DEBUG:parso.python.diff:diff parser start
DEBUG:parso.python.diff:diff parser calculated
DEBUG:parso.python.diff:diff: line_lengths old: 1, new: 1
DEBUG:parso.python.diff:diff replace old[1:1] new[1:1]
DEBUG:parso.python.diff:parse_part from 1 to 1 (to 0 in part parser)
DEBUG:parso.python.diff:diff parser end
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/collections/__init__.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/io.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/os.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/string.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/threading.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/traceback.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/warnings.py
DEBUG:parso.cache:pickle loaded: /Users/uushuo/.pyenv/versions/3.6.3/lib/python3.6/weakref.py
```

Version info
```
{'commit_hash': 'ca5443062',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/uushuo/.pyenv/versions/3.6.3/envs/next/lib/python3.6/site-packages/IPython',
 'ipython_version': '6.2.1',
 'os_name': 'posix',
 'platform': 'Darwin-17.0.0-x86_64-i386-64bit',
 'sys_executable': '/Users/uushuo/.pyenv/versions/next/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '3.6.3 (default, Dec 15 2017, 20:17:37) \n'
                '[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]'}
```

drunkwcodes:
This PR closes #10946.

I was wondering where to put this line.
I think the logger module is the best place.

The issue won't break the interactive shell now.

### DIFF
--- a/IPython/core/logger.py
+++ b/IPython/core/logger.py
@@ -15,9 +15,14 @@
 # Python standard modules
 import glob
 import io
+import logging
 import os
 import time
 
+
+
+# prevent jedi/parso's debug messages pipe into interactiveshell
+logging.getLogger("parso").setLevel(logging.WARNING)
 
 #****************************************************************************
 # FIXME: This class isn't a mixin anymore, but it still needs attributes from


IPython/core/logger.py
```diff
@@ -15,10 +15,15 @@
...
 import glob
 import io
+import logging
 import os
 import time
 
 
+
+# prevent jedi/parso's debug messages pipe into interactiveshell
+logging.getLogger("parso").setLevel(logging.WARNING)
```
intgr, line 25: I'm not a maintainer here, just a lurker. But I think imports with side-effects are bad. Better put this in a function and call it from some point in IPython initialization.

drunkwcodes: @intgr Would you suggest a place to call the function with line number in interactiveshell.py?

Look at history.py & aliases.py, there are variables assignments in the module level, too.
I think this is acceptable in ipython.
And this change is not that important to change the flow of interactiveshell.py, so IMO this way is more elegant and won't mess with interactiveshell.py.

intgr: I don't know, sorry.

wyatthoho (approved):

drunkwcodes:
Thanks @WyattHo. I'm appreciate your review!

IPython/core/logger.py
Carreau: Thanks for the review and the discussion, in general we don't like side effect, but as IPython is a shell/interpreter often we don't have too much choices to give a good user experience. I think that is fine here for the time being unless we find a better place.

Carreau:
Thanks for you contribution and the team effort.

PR #9854, opened 2016-08-10
run CFRunLoopRun if NSApp:run finishes on its own

minrk:
This seems to happen when the last Window is closed, but the eventloop needs to run a little longer in order to finish closing the Window.

closes #9734

Issue #9734, opened 2016-07-11
Cannot close last figure window


crosset:
After updating ipython to 5.0 (from macports), I cannot close last figure window (neither with button in window bar, nor with 'close()'). The mouse cursor becomes spinning wheel inside the window, but I can still type command in the terminal window.

However, if I type clf(), the figure window closes and a new one appears.

This happens only when there is only one figure window remaining. 

Here's my system (OS X  El Capitan, with python/ipython etc. installed via macports) :
```
[~] $ python -c "import IPython; print(IPython.sys_info())"
{'commit_hash': u'<not found>',
 'commit_source': '(none found)',
 'default_encoding': 'UTF-8',
 'ipython_path': '/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/IPython',
 'ipython_version': '5.0.0',
 'os_name': 'posix',
 'platform': 'Darwin-15.5.0-x86_64-i386-64bit',
 'sys_executable': '/opt/local/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python',
 'sys_platform': 'darwin',
 'sys_version': '2.7.12 (default, Jun 29 2016, 12:52:38) \n[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]'}
```

Carreau:
Do you use the `%matplotlib` magic ? And/or which matplotlib backend are you using to plot ?

crosset:
Oh, sorry I forgot to mention : I am using the OS X backend (as defined in my matplotlibrc file, so that it is selected automatically when calling "ipython --pylab"). I also tried starting with "ipython" and then doing "%matplotlib osx" but I get the same behavior.

Carreau:
Hum, that might be a macport issue, I'm unable to reproduce with other Python install like homebrew. Whether on py2 or py3, though I'm on Yosemite... 

Can you try a pip-installed version ?

therockmandolinist:
@Carreau I'm experiencing the same issue with py3 on OSX 10.11.5 (EL Capitan) with ipython pip-installed, python3 installed through homebrew.

crosset:
Same issue with python27 installed with macports and ipython-5.0 installed via pip.

However, it works fine with Tk backend (macports install, starting with ipython --pylab-tk), I can close the figure window normally.

Carreau:
hum.. what about pure python shell ?

[...]

PR #14125, opened 2023-07-26
Remove the Event '_triggered' in osx.py to fix hangs.

culler:
This is my proposed fix for issue 

closes #14072


ccordoba12:
@minrk, could you review these changes since you were the one that introduced them in PR #9854? Thanks!

minrk:
tested that this fixes the crash and doesn't restore the original #9734 bug.

I believe the Intermittent test failure is https://github.com/ipython/ipykernel/pull/1137

Issue #13953, opened 2023-02-26
Syntax highlighting of traceback context always uses default Pygments style


li-dan:
We have `TerminalInteractiveShell.highlighting_style` to pick a Pygments style for normal syntax highlighting, but tracebacks always use the default style: https://github.com/ipython/ipython/blob/3edbe3bd10c434af6458bdbe02269880b10b9adf/IPython/core/ultratb.py#L985-L988

Similar to #13756 for mitigating #13446 and #13486, would it be possible to put in a stop-gap to allow the user to change the style?

```diff
$ git diff
diff --git a/IPython/core/ultratb.py b/IPython/core/ultratb.py
index 18eff2708..030dea717 100644
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -617,6 +617,7 @@ class VerboseTB(TBTools):
     would appear in the traceback)."""

     _tb_highlight = "bg:ansiyellow"
+    _tb_highlight_style = "default"

     def __init__(
         self,
@@ -841,7 +842,7 @@ class VerboseTB(TBTools):
         after = context // 2
         before = context - after
         if self.has_colors:
-            style = get_style_by_name("default")
+            style = get_style_by_name(self._tb_highlight_style)
             style = stack_data.style_with_executing_node(style, self._tb_highlight)
             formatter = Terminal256Formatter(style=style)
         else:
```

li-dan:
Closing as the patch has been merged to main.

PR #14138, opened 2023-08-14
Add hack to change traceback syntax highlighting

li-dan:
Similar to #13756, add a hack to configure the Pygments style for syntax highlighting of tracebacks.

This does not change default behavior.

To use the hack, add to ipython_config.py:

    from IPython.core.ultratb import VerboseTB
    VerboseTB._tb_highlight_style = "pastie"

Mitigates #13953.


Carreau:
Thanks, that seems reasonable.

PR #14139, opened 2023-08-18
Fix DeprecationWarning on 3.12 with a datetime adapter for sqlite3

skirpichev:
After python/cpython#90016, default adapters and converters are deprecated.  See also
https://docs.python.org/3.12/library/sqlite3.html#sqlite3-adapter-converter-recipes

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -9,7 +9,6 @@ import datetime
 from pathlib import Path
 import re
 import sqlite3
-import sys
 import threading
 
 from traitlets.config.configurable import LoggingConfigurable
@@ -29,14 +28,6 @@ from traitlets import (
     default,
     observe,
 )
-
-
-if sys.version_info >= (3, 12):
-
-    def _adapt_datetime(val):
-        return val.isoformat(" ")
-
-    sqlite3.register_adapter(datetime.datetime, _adapt_datetime)
 
 #-----------------------------------------------------------------------------
 # Classes and functions
@@ -583,7 +574,7 @@ class HistoryManager(HistoryAccessor):
             cur = conn.execute(
                 """INSERT INTO sessions VALUES (NULL, ?, NULL,
                             NULL, '') """,
-                (datetime.datetime.now(),),
+                (datetime.datetime.now().isoformat(" "),),
             )
             self.session_number = cur.lastrowid
 
@@ -591,9 +582,15 @@ class HistoryManager(HistoryAccessor):
         """Close the database session, filling in the end time and line count."""
         self.writeout_cache()
         with self.db:
-            self.db.execute("""UPDATE sessions SET end=?, num_cmds=? WHERE
-                            session==?""", (datetime.datetime.now(),
-                            len(self.input_hist_parsed)-1, self.session_number))
+            self.db.execute(
+                """UPDATE sessions SET end=?, num_cmds=? WHERE
+                            session==?""",
+                (
+                    datetime.datetime.now().isoformat(" "),
+                    len(self.input_hist_parsed) - 1,
+                    self.session_number,
+                ),
+            )
         self.session_number = 0
 
     def name_session(self, name):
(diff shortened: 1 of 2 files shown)


IPython/core/history.py
```diff
@@ -29,6 +30,14 @@
     observe,
 )
 
+
+if sys.version_info >= (3, 12):
+
+    def _adapt_datetime(val):
+        return val.isoformat(" ")
+
+    sqlite3.register_adapter(datetime.datetime, _adapt_datetime)
```
Carreau, line 39: Do you believe it would be safer to not register adapter and call `isoformat(" ")` on the actual places where we insert data into sqlite ? 

The fact that the upstream issue tell us there are issues with the default adapters is maybe a hint we should not re-register the same as it has global effects.

In the long term, I'm also wondering why `isoformat(" ")`, and not just `isoformat()`, and if there would be any way we could migrate to just `isoformat()`...

skirpichev: > Do you believe it would be safer to not register adapter and call isoformat(" ") on the actual places where we insert data into sqlite ?

I'm not sure.  I hit this warning while running ``pytest diofant/tests/interactive`` for the [diofant](https://github.com/diofant/diofant/) project and it seems to be related to the statement in the HistoryManager.new_session() method.

But maybe there are more such cases...  The default adapter

> The fact that the upstream issue tell us there are issues with the default adapters

The default adapters now are explicit rather than implicit.  This is the only difference.  I don't think this means there is an issue with the notion of default adapters or something like that...

> I'm also wondering why isoformat(" "), and not just isoformat(), and if there would be any way we could migrate to just isoformat()

That's a literal copy of the [current default adapter](https://github.com/python/cpython/blob/5d936b64796261373429c86cdf90b1d8d8acefba/Lib/sqlite3/dbapi2.py#L67-L69) from the sqlite3 module.  " " is because the default value for ``sep`` kwarg is "T", see [docs](https://docs.python.org/3/library/datetime.html?highlight=isoformat#datetime.datetime.isoformat).

Carreau: > he default adapters now are explicit rather than implicit. This is the only difference. I don't think this means there is an issue with the notion of default adapters or something like that...

I was more refering to the text in https://github.com/python/cpython/issues/90016 saying "the default converters in SQLite3 have several bugs"



> That's a literal copy of the [current default adapter](https://github.com/python/cpython/blob/5d936b64796261373429c86cdf90b1d8d8acefba/Lib/sqlite3/dbapi2.py#L67-L69) from the sqlite3 module. " " is because the default value for `sep` kwarg is "T", see [docs](https://docs.python.org/3/library/datetime.html?highlight=isoformat#datetime.datetime.isoformat).

Yes, I got that as well, I'm wondering if we should use the default with `T`, obviously we want to be careful as we don't want to mess up existing history.

Carreau:
I'll try an alternative where I call .isoformat(" ") explicitely were necessary and don't change the default adapter.

IPython/core/history.py
skirpichev: > I was more refering to the text in https://github.com/python/cpython/issues/90016 saying "the default converters in SQLite3 have several bugs"

The problem is that in the alternative fix you are using *same* default adapter, but reinvent this code in several places instead.  Default adapters are exactly for this type of things...

> Yes, I got that as well, I'm wondering if we should use the default with T

Perhaps, but this will change the db format.  So, probably this will require a deprecation period, etc.

Carreau:
Thanks !

IPython/core/history.py
Carreau: Yes, but I am using the fix _only_ in IPython internal code. If we reinject the same default adapter with `sqlite3.register_adapter` this mean we affect code that runs in IPython.  Code that will work in IPython may not work in pure Python, in particular is code that use sql and datetime. Once CPython remove the default adapter this code will work in IPython but fail in CPython

PR #14141, opened 2023-08-22
Update %timeit documentation: replace best time with average time

komo-fr:
This PR addresses outdated documentation for the `%timeit` magic command.

As of version 6.0.0rc1, `%timeit` measures the average time instead of the best time.
- Related PR at the time: https://github.com/ipython/ipython/pull/9984

However, the documentation still reflects the old behavior.

### 1. Description of the `-r` option
The current documentation states the following:
https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-timeit

> -r<R>: number of repeats <R>, each consisting of <N> loops, and take the best result. Default: 7

This seems to be the correct `average result`.

The actual code calculates the average:
https://github.com/ipython/ipython/blob/f49a797f1488a3f54a0030de12d92393f878f335/IPython/core/magics/execution.py#L112-L119

### 2. The format of the `%timeit` results in the sample code for `%alias_magic` is outdated.

Current documentation:
https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-alias_magic
> In [2]: %t -n1 pass
> 1 loops, best of 3: 954 ns per loop

Actual result:
> 107 ns ± 43.6 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)

The following pages also have outdated notation:
- Overview: https://ipython.readthedocs.io/en/stable/overview.html#main-features-of-the-interactive-shell
- Introducing Tutorial: https://ipython.readthedocs.io/en/stable/interactive/tutorial.html#magic-functions
- IPython reference: https://ipython.readthedocs.io/en/stable/interactive/reference.html#magic-command-system
- Python vs IPython: https://ipython.readthedocs.io/en/stable/interactive/python-ipython-diff.html#magics

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -109,12 +109,12 @@ class BasicMagics(Magics):
           Created `%%t` as an alias for `%%timeit`.
 
           In [2]: %t -n1 pass
-          1 loops, best of 3: 954 ns per loop
+          107 ns ± 43.6 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)
 
           In [3]: %%t -n1
              ...: pass
              ...:
-          1 loops, best of 3: 954 ns per loop
+          107 ns ± 58.3 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)
 
           In [4]: %alias_magic --cell whereami pwd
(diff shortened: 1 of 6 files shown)


Carreau:
Thanks ! That's great !

Issue #12895, opened 2021-04-07
Markdown rendering in output


linceaerian:
Hello maintainers,

Sorry to bother you, I'm not sure if my problem is due to either IPython / Jupyter / Gitlab, so maybe it will be closed soon.
I'm encountering the same problem than in some issues ex: https://github.com/jupyter/nbconvert/issues/145  (and a lot of other on stackoverflow which use markdown to render math functions).

I want to dynamically render some markdown output in a Python cell but all I get when using `display(Markdown(<string>)` is `<IPython.core.display.Markdown object>` as shown below:

![image](https://user-images.githubusercontent.com/1318470/113896741-8249e180-97ca-11eb-95df-3b43f9f01920.png)

The code used is:

```
        name = '## Working on %s' % host["host"]
        display({'text/plain': name, 'text/markdown': name}, raw=True)
        display(Markdown(name))
```

I'm using the text/plain workaround for the moment which display the text but I would like to have the output rendered if possible :)

I would expect that the process for display.Markdown would render the markdown string into HTML but it seems it does not... or am I missing some call ?

For a markdown object in display I get:

![image](https://user-images.githubusercontent.com/1318470/113899763-698efb00-97cd-11eb-93b8-098132169447.png)
with outpout source for the notebook as:
![image](https://user-images.githubusercontent.com/1318470/113900026-ae1a9680-97cd-11eb-8241-6bd4e3708f4f.png)

While a panda table in display is rendered as;

![image](https://user-images.githubusercontent.com/1318470/113899846-7f9cbb80-97cd-11eb-8f91-e626e35fe174.png)
![image](https://user-images.githubusercontent.com/1318470/113900096-c4c0ed80-97cd-11eb-85cc-ebb28c5e7e66.png)

Sorry to bother you and thanks for reading until the end (I cannot seems to find any relevant information from stackoverflow or others T.T).

Regards

mizoru:
Hello, everybody!

I encounter a similar issue with this code:
```python
from IPython.display import Markdown, display
display(Markdown(md))
```
It renders fine in the notebook, but since the `text/plain` representation is just `<IPython.core.display.Markdown object>`, one sees no actual data in the terminal (an equivalent of `print(md)` would be sufficient).
I tried using `display({'text/markdown': md, 'text/plain':md}, raw=True)`, but that just leads to one seeing the `dict`'s `repr` in the console.
[...]

PR #14150, opened 2023-09-02
Add a 'text/plain' repr to Markdown and JSON

mizoru:
Addresses [this issue](https://github.com/ipython/ipython/issues/12895#issuecomment-1698263807).

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -602,6 +602,9 @@ class JSON(DisplayObject):
         if kwargs:
             self.metadata.update(kwargs)
         super(JSON, self).__init__(data=data, url=url, filename=filename)
+    
+    def __repr__(self):
+        return self.data
 
     def _check_data(self):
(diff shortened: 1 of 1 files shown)


IPython/core/display.py
```diff
@@ -599,6 +602,9 @@ def __init__(self, data=None, url=None, filename=None, expanded=False, metadata=
         if kwargs:
             self.metadata.update(kwargs)
         super(JSON, self).__init__(data=data, url=url, filename=filename)
+    
+    def __repr__(self):
+        return self.data
```
Carreau, line 607: Thanks for this !

Shouldn't the JSON text representation return something more like `Json.dumps(self.data, indent=2)` ?

westurner: IDK how to make that optional for jsonlines or similar; so this is fine as-is imho

westurner: Unless it could be `__str__` instead of `__repr__`?

PR #14161, opened 2023-09-12
Fix wrong u-prefix for strings in magics docs.

anntzer:
`%precision` doesn't output `u'%r'` on Python 3, but `'%r'`.

(Note that the point of this PR is explicitly not to remove u-prefixes on all strings -- something that would likely be better done by automated tooling anyways -- but only on those occurring within some docstrings that end up in the rendered docs, which would more likely not have been found by tooling.)


Carreau:
Thanks ! 

And yes, I'm avoiding major refactoring/rewriting of IPython, but will probably get to it at some point.

PR #14163, opened 2023-09-12
fix semicolon detection with no history

### LINKED ISSUE #14162: Suppress output with semicolon does not work with store_history=False
Steps to reproduce:
```
In [4]: ip.run_cell('1;')
Out[4]: 1
Out[4]: <ExecutionResult object at ...
```

idanpa:
fixes #14162

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -51,7 +51,7 @@ class DisplayHook(Configurable):
 
         # we need a reference to the user-level namespace
         self.shell = shell
-        
+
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks for both the issue and the fix.

rawlins:
Just to flag this here more directly, in hopes that someone who is capable of fixing it sees this: it appears that this PR breaks line magics in an unfortunate way, see https://github.com/ipython/ipython/issues/14195.

Issue #14164, opened 2023-09-14
Example of creating InteractiveShellEmbed in docs is against SingletonConfigurable guidelines and crashes completion


Novakov:
IPython docs (https://ipython.readthedocs.io/en/stable/interactive/reference.html#embedding-ipython) provides an example for embedding IPython shell inside application:
```python
ipshell = InteractiveShellEmbed(config=cfg,
                       banner1 = 'Dropping into IPython',
                       exit_msg = 'Leaving Interpreter, back to program.')
```

Since version 8.10 (or a bit later) this results in crashes when using tab completion
```
Unhandled exception in event loop:
  File "C:\Python39\lib\asyncio\events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\input\win32.py", line 604, in ready
    callback()
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\application\application.py", line 712, in read_from_input_in_context
    context.copy().run(read_from_input)
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\application\application.py", line 692, in read_from_input
    self.key_processor.process_keys()
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\key_binding\key_processor.py", line 272, in process_keys
    self._process_coroutine.send(key_press)
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\key_binding\key_processor.py", line 187, in _process
    self._call_handler(matches[-1], key_sequence=buffer[:])
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\key_binding\key_processor.py", line 322, in _call_handler
    handler.call(event)
  File "D:\tmp\debug-ipython\python-prompt-toolkit\src\prompt_toolkit\key_binding\key_bindings.py", line 124, in call
    result = self.handler(event)
  File "D:\tmp\debug-ipython\ipython\IPython\terminal\shortcuts\__init__.py", line 405, in handle_return_or_newline_or_execute
    return newline_or_execute_outer(shell)(event)
  File "D:\tmp\debug-ipython\ipython\IPython\terminal\shortcuts\__init__.py", line 428, in newline_or_execute
    status, indent = shell.check_complete(check_text)

Exception 'NoneType' object has no attribute 'check_complete'
[...]

PR #14165, opened 2023-09-16
try fix for #14164

Carreau:
See #14164

a proper fix should be in traitlets in the constructor to make sure we can either use the constructor, or "instance()"/


Carreau:
Same as #14169 I forgot I had already send a PR

PR #14169, opened 2023-09-19
Set instance on singleton configurable when created via constructor.

Carreau:
See #14164


Carreau:
@meeseeksdev backport to 8.12.x

PR #14172, opened 2023-09-25
Make `ipapp.py` unrunnable as one py file, due to relative import

### LINKED ISSUE #14159: Unable to run `ipapp.py`
There is a problem with the `ipython/terminal/ipapp.py` file.

On the first line is this:
```python
#!/usr/bin/env python
```
this is an mistake, because further down in the code there is a relative import that makes the file not executable with `./ipapp.py`.
So please remove `#!/usr/bin/env python`.

Steps to reproduce:

```
# download repo
cd /tmp/
git clone https://github.com/ipython/ipython.git

# run file
cd ipython/IPython/terminal/
./ipapp.py
```

Example of error:
```
  File "/tmp/ipython/IPython/terminal/./ipapp.py", line 36, in <module>
    from .interactiveshell import TerminalInteractiveShell
ImportError: attempted relative import with no known parent package
```

marostelka:
fixes #14159

### DIFF
--- a/IPython/terminal/ipapp.py
+++ b/IPython/terminal/ipapp.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # encoding: utf-8
 """
 The :class:`~traitlets.config.application.Application` object for the command
@@ -337,7 +336,3 @@ def load_default_config(ipython_dir=None):
     return app.config
 
 launch_new_instance = TerminalIPythonApp.launch_instance
-
-
-if __name__ == '__main__':
-    launch_new_instance()


IPython/terminal/ipapp.py
```diff
@@ -335,9 +334,3 @@ def load_default_config(ipython_dir=None):
     app.config_file_paths.append(profile_dir)
     app.load_config_file()
     return app.config
-
-launch_new_instance = TerminalIPythonApp.launch_instance
-
-
-if __name__ == '__main__':
-    launch_new_instance()
```
Carreau, line 343: You need to keep the `launch_new_instance = TerminalIPythonApp.launch_instance`, it is imported in other places.

aueam: Thanks for the report, it should be fixed now.

Carreau:
Thanks !

Issue #14186, opened 2023-09-30
look into deprecated utils.text columnize. It is not used anymore.


Carreau:

Carreau:
#14187 does the deprecation, but for each of the warning an issue should be opened and corresponding warning updated to link to the newly open issue about removal.

PR #14187, opened 2023-09-30
Deprecation and removal for 8.17

Carreau:
See #14186

I think for each of those warning we should:

1) Open an issue for removal, and update the corresponding warnings text to point to the corresponding issue.

### DIFF
--- a/IPython/core/inputsplitter.py
+++ b/IPython/core/inputsplitter.py
@@ -91,7 +91,13 @@ def num_ini_spaces(s):
     -------
     n : int
     """
-
+    warnings.warn(
+        "`num_ini_spaces` is Pending Deprecation since IPython 8.17."
+        "It is considered fro removal in in future version. "
+        "Please open an issue if you believe it should be kept.",
+        stacklevel=2,
+        category=PendingDeprecationWarning,
+    )
     ini_spaces = ini_spaces_re.match(s)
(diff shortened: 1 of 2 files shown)


bollwyvl:
I like what pyodide does:

https://pyodide.org/en/stable/project/deprecation-timeline.html#pyodide-deprecation-timeline

Maybe such a thing can be gathered with some plugin like:\

https://deprecation.readthedocs.io/en/latest/

bollwyvl:
There are also some fun tools that can consume/profile such decorators:

https://github.com/QuantStack/memestra

Carreau:
Or likely something like @takluyver https://pypi.org/project/astsearch/ to find functions with `warnings.warn(??, DeprecationWarning)` inside...

Carreau:
In general I find like `@deprecated` decorator make things more complicated than just emitting a warning. You need to learn them and they force you in a framework that always need to more options. Like sometime deprecated mean the function/argument already does nothing and it is already safe to not use it. Sometime you need to describe the replacement... etc.

bollwyvl:
Sure, that works too... end of the day, having a clear, statically-verified collection of _What Will Break In IPython X.Y_ in the docs is win. A sphinx directive that could collect them via `ast` or whatever would be win.

PR #14188, opened 2023-10-01
Fix temporary folder cleanup on exit

### LINKED ISSUE #14099: atexit_operations fails to delete tempdir due to vim undofile leftovers (`*.un~`)
Steps to reproduce:

0. Set EDITOR to `vim`  
1. define function `def foo(): return 1`
3. edit it by `%edit foo`, do some stuff in file while vim set to use undofile `:set undofile` (this is the default setting on Vim 8.2 for windows) and save
5. exit IPython

```
In [1]: def foo():
   ...:     return 1
   ...: 
In [2]: edit foo
Editing In[1]
IPython will make a temporary file named: C:\Users\idank\AppData\Local\Temp\ipython_edit_lfo_g8nb\ipython_edit_4yzyx_uz.py
done. Executing edited code...
Out[2]: 'def foo():\n\n    return 2\n'
In [3]: exit
Exception ignored in atexit callback: <bound method InteractiveShell.atexit_operations of <IPython.terminal.interactiveshell.TerminalInteractiveShell object at 0x0000024205566010>>
  File "C:\Users\idank\mambaforge\envs\py311\Lib\site-packages\IPython\core\interactiveshell.py", line 3887, in atexit_operations
    tdir.rmdir()
  File "C:\Users\idank\mambaforge\envs\py311\Lib\pathlib.py", line 1156, in rmdir
    os.rmdir(self)
OSError: [WinError 145] The directory is not empty: 'C:\\Users\\idank\\AppData\\Local\\Temp\\ipython_edit_lfo_g8nb'
```

Windows 10, IPython 8.14.0

The problem is that vim would leave a `*.un~` file with the undo history in the temp directory which won't get deleted by IPython, then pathlib.Path.rmdir would fail in [interactiveshell.py:3887](https://github.com/ipython/ipython/blob/cf7c266435424f43ed48c91239f895e8ad739266/IPython/core/interactiveshell.py#LL3887C24-L3887C29), (`shutil.rmtree` the entire directory?)

sgaist:
When using mkdtemp, the folder content cleanup must be done by the user.

The current call to `rmdir` fails in case additional files are generated while editing the code. For example vim with undofile activated does this.

Fixes #14099

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -21,6 +21,7 @@ import inspect
 import os
 import re
 import runpy
+import shutil
 import subprocess
 import sys
 import tempfile
@@ -3902,7 +3903,7 @@ class InteractiveShell(SingletonConfigurable):
         del self.tempfiles
         for tdir in self.tempdirs:
             try:
-                tdir.rmdir()
+                shutil.rmtree(tdir)
                 self.tempdirs.remove(tdir)
             except FileNotFoundError:
                 pass


Carreau:
Thanks !

PR #14190, opened 2023-10-01
Cleanup Python 2 compact from Lexers

### LINKED ISSUE #14021: Cleanup Python 2 compact from IPython/lib/lexers.py
There should still be Python 2 lexers that are not useful

sgaist:
Python 2 has reached end of life and its support has been dropped already. This patch removes the lexer dedicated to Python 2.

Fixes #14021

### DIFF
--- a/IPython/lib/lexers.py
+++ b/IPython/lib/lexers.py
@@ -4,13 +4,13 @@ Defines a variety of Pygments lexers for highlighting IPython code.
 
 This includes:
 
-    IPythonLexer, IPython3Lexer
-        Lexers for pure IPython (python + magic/shell commands)
+    IPython3Lexer
+        Lexer for pure IPython (python + magic/shell commands)
 
     IPythonPartialTracebackLexer, IPythonTracebackLexer
-        Supports 2.x and 3.x via keyword `python3`.  The partial traceback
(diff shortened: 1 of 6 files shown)


IPython/lib/lexers.py
```diff
@@ -201,19 +239,11 @@ def __init__(self, **options):
...
-        else:
-            self.aliases = ['ipython2tb', 'ipythontb']
 
-        if self.python3:
-            IPyLexer = IPython3Lexer
-        else:
-            IPyLexer = IPythonLexer
+        DelegatingLexer.__init__(
+            self, IPython3Lexer, IPythonPartialTracebackLexer, **options
+        )
```
Carreau, line 245: can we use `super()` directly here now ?

Carreau:
One nitpick with `super()` and I think we are good.

IPython/lib/lexers.py
sgaist: Yes, it's also mentioned in the DelegatingLexer documentation.

Carreau (approved):

Carreau:
Ok, lets get that in

Carreau:
I'm going to revert as this is actually used in a couple of downstream packages.

Issue #14181, opened 2023-09-29
mypy findings with traitlets 5.10.1


bollwyvl:
Downstream on [conda-forge](https://github.com/conda-forge/ipython-feedstock/pull/197#issuecomment-1740791008), `mypy` found the following issues when run under the `pyproject.toml` settings vs `traitlets 5.10.1`. Of course, some of these might be [false positives](https://github.com/ipython/traitlets/issues/871).

```py
/core/magics/ast_mod.py:244: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
/core/extensions.py:89: error: Item "None" of "Optional[Any]" has no attribute "builtin_trap"  [union-attr]
/terminal/prompts.py:100: error: Item "None" of "Optional[Any]" has no attribute "mime_renderers"  [union-attr]
/terminal/prompts.py:102: error: Item "None" of "Optional[Any]" has no attribute "mime_renderers"  [union-attr]
/core/formatters.py:56: error: Need type annotation for "ipython_display_formatter"  [var-annotated]
/core/formatters.py:61: error: Need type annotation for "mimebundle_formatter"  [var-annotated]
/core/interactiveshell.py:1785: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
/core/interactiveshell.py:2380: error: Item "None" of "Optional[Any]" has no attribute "lazy_magics"  [union-attr]
/core/interactiveshell.py:2403: error: Item "None" of "Optional[Any]" has no attribute "lazy_magics"  [union-attr]
/core/interactiveshell.py:2431: error: Item "None" of "Optional[Any]" has no attribute "__enter__"  [union-attr]
/core/interactiveshell.py:2431: error: Item "None" of "Optional[Any]" has no attribute "__exit__"  [union-attr]
/core/interactiveshell.py:3184: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
/core/interactiveshell.py:3232: error: Item "None" of "Optional[Any]" has no attribute "store_inputs"  [union-attr]
/core/interactiveshell.py:3246: error: "object" not callable  [operator]
/core/interactiveshell.py:3250: error: Item "None" of "Optional[Any]" has no attribute "__enter__"  [union-attr]
/core/interactiveshell.py:3250: error: Item "None" of "Optional[Any]" has no attribute "__exit__"  [union-attr]
[...]

PR #14193, opened 2023-10-01
MAINT: partial type fixes.

Carreau:
See #14181

This fixes the first few errors.

### DIFF
--- a/IPython/core/extensions.py
+++ b/IPython/core/extensions.py
@@ -86,14 +86,15 @@ class ExtensionManager(Configurable):
 
         from IPython.utils.syspathcontext import prepended_to_syspath
 
-        with self.shell.builtin_trap:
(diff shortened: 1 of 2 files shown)


IPython/core/extensions.py
```diff
@@ -86,14 +86,15 @@ def _load_extension(self, module_str: str):
...
-                return "no load function"
+        if self.shell and getattr(self.shell, "builtin_trap", None) is not None:
+            with self.shell.builtin_trap:
+                if module_str not in sys.modules:
+                    mod = import_module(module_str)
+                mod = sys.modules[module_str]
+                if self._call_load_ipython_extension(mod):
+                    self.loaded.add(module_str)
+                else:
+                    return "no load function"
```
Carreau, line 97: I think I'm going to make builtin_trap non optional.

PR #14197, opened 2023-10-01
displayhook - get reference to the actual executed cell

idanpa:
raw_cell might need transformation and cannot be tokenized.
Fixes #7256


Carreau:
Just to let you know this is still on my todo list to review. 
I'll try to understand better and get that in when possible.

idanpa:
I have no access to a laptop so I can't check that.
I would be able to continue once the situation in Israel stabilize.
I don't remember if this commit ready, the fix itself is nice to have.

Carreau:

PR #14212, opened 2023-10-17
[perf] optimize object inspect

zolyfarkas-fb:
Do not call signature(obj) for non-callables. Signature, will check and raise a exception where the message will contain `obj.__repr__` and that can be very expensive (take minutes) on certain objects.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -416,6 +416,8 @@ class Inspector(Colorable):
 
         If any exception is generated, None is returned instead and the
         exception is suppressed."""
+        if not callable(obj):
+            return None        
         try:
             return _render_signature(signature(obj), oname)
         except:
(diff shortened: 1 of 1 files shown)


IPython/core/oinspect.py
```diff
@@ -416,6 +416,8 @@ def _getdef(self,obj,oname='') -> Union[str,None]:
 
         If any exception is generated, None is returned instead and the
         exception is suppressed."""
+        if not callable(obj):
+            return None
         try:
             return _render_signature(signature(obj), oname)
```
zolyfarkas-fb, line 422: the implementation of signature(obj) starts with:

```
    if not callable(obj):
        raise TypeError('{!r} is not a callable object'.format(obj))
```

The current implementation catches this exception and returns None.
constructing the exception error message can be very expensive (minutes) depending on the obj instance.

Carreau:
Thanks, make sens.

PR #14218, opened 2023-10-25
refactor: Use pathlib in tools/

mmtj:
Since script `release` is meant to be run only when doing the actual release of ipython, I could not easily verify it works correctly.
So please double check if it still works the same way as before.

Related to: #12515


Carreau (approved):
Let's try ! Thanks !

PR #14221, opened 2023-10-27
refactor: use pathlib in sysinfo.py

mmtj:
* Replace `os.path` with `pathlib.Path`
* Add type hints for the functions

Related to: #12515


Carreau:
Great ! Lets give this a go as well !

PR #6662, opened 2014-10-09
Proposal: remove dependencies from IPython.external

minrk:
This removes IPython dependencies from IPython.external, and expresses them as regular dependencies.

The one we can't remove is pickleshare, which currently doesn't install on Python 3. It's a trivial fix, but the package doesn't seem to be maintained.

closes #3994

PR #14238, opened 2023-11-09
Remove unused dependency `appnope`

zachborboacryptofi:
Remove the dependency `appnope` as it is no longer being used, but still being installed on macOS.

Last seen in version `3.2.3`.
Removed in https://github.com/ipython/ipython/pull/6662.

`IPython/external/appnope` exists in `3.2.3`:
https://github.com/ipython/ipython/tree/3.2.3/IPython/external

Gone in the next version (`4.0.0`):
https://github.com/ipython/ipython/tree/4.0.0/IPython/external

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,7 +29,6 @@ packages = find:
 python_requires = >=3.9
 zip_safe = False
 install_requires =
-    appnope; sys_platform == "darwin"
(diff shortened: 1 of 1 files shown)


Carreau:
Oh, yeah, only used by ipykernel indeed.

PR #14245, opened 2023-11-19
Fix list pygments styles example

jpoppe:
Styles needs to be imported to be able to access the `get_all_styles()` function or `get_all_styles()` could be imported directly from `styles`.

This fixes the example.


Carreau:
Thanks !

PR #14254, opened 2023-11-24
Drop 3.9

Carreau:
According to SPEC-0 this is about time,
we thus bump minimal python, shift CI to test from 3.10 to 3.12, mark a
curio test as x-fail as curio is not yet compatible with 3.12 but the
fix in in trio main.

Remove all conditionals that will never execute on 3.10+


Carreau:
3.13 failures seem unrelated, merging

PR #14268, opened 2023-12-18
attempt to unpin pytest

Carreau:


Carreau:
It seem to work, let's test.

PR #14275, opened 2023-12-27
Misc linter cleanups

Carreau:


ivanov (approved):
I guess the 3.13 failures have been there for a while, it was silly of me to try to rerun them

Carreau:
yeah, the failure on 3.13 are regular. if by next alpha it is not fixed,
maybe we need to investigate and open a bug upstream

Carreau:
❤️

PR #14278, opened 2023-12-31
Misc fix for Python 3.13 testing (unicode range and skip tests)

Carreau:
Update unicode range for python 3.13, and skip jedi test while in alpha as there are no grammar available.


Carreau:
remaining failures are unrelated.

PR #14279, opened 2023-12-31
Fix docstring leading space.

Carreau:
Since https://github.com/python/cpython/pull/106411, it look like cpython is removing leading space and indent.

We therefore need conditional check, until we decide Wether this is a problem we want to deal with or not.


Carreau:
This reduce the number of failures on 3.13 to 4

PR #14283, opened 2024-01-04
Enable starting `ipython` from `python -m ipython`

SamuelMarks:
Similar to how `pip` is now recommended to be called via `python -m pip`. Makes warning like this impossible:
> UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.

EDIT0: Second commit from running `python -m black --config 'pyproject.toml' 'setup.py' && isort "$_"`
EDIT1: Ohhh interesting. Because Windows isn't neccesarily case sensitive (depending on filesystem) this doesn't work neatly; specifically `sdist` but also the regular command. Hmm. `python -m IPython` might work? - [PEP235](https://peps.python.org/pep-0235/) applies.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,6 +9,8 @@ include .flake8
 include .pre-commit-config.yaml
 include long_description.rst
 
+recursive-include ipython *.py
+
 recursive-exclude tools *
 exclude tools
 exclude CONTRIBUTING.md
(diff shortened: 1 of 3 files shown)


Carreau:
As it It already works with the proper case (`python -m IPython`), I would prefer to _not_ allow lower case. Yes IPython should have been a lower case import, but this project is more than 20 year old, so I think that ship has sailed. 

As you discovered there are enough issues with casing and not on windows only that I don't think we should try to make it work.

I'm going to close, but we can still discuss if you wish, and we can reopen if you convince me otherwise.

PR #14286, opened 2024-01-08
Prevent crash on StackData NotOneValueFound.

Carreau:
This is a workaround for #13598,

It does not fix the root cause, but should prevent IPython crash.

Worst case it should just not show the relevant frame


ivanov (approved):
:medal_military:

Carreau:
@ivanov you are the 🥇

PR #14287, opened 2024-01-09
Try to fix Pytest error on Python 3.13

Carreau:


ivanov (approved):
ZOMG! it passes! :cake:

Carreau:
Thanks @ivanov , you should feel free to merge, I belive you still have commit rights ! ❤️

ivanov:
@Carreau indeed I do, but I forget if we're a self-merge kind of project here, or not.

Carreau:
> @Carreau indeed I do, but I forget if we're a self-merge kind of project here, or not.

We we are not supposed to be a self-merge, but as in practice I appear to be the last active maintainer left, I end up self-merging.

PR #14289, opened 2024-01-11
ipdb: accept header= argument

MrMino:
This improves symmetry between Ipdb and vanilla old Pdb.

The `header` argument to `pdb.set_trace()` [has been added in Python 3.7](https://docs.python.org/3.9/whatsnew/3.7.html#pdb).

_(This still requires changes in [gotcha/ipdb](https://github.com/gotcha/ipdb) to be fully useful - I'll get to that later :wink:)_

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -1111,10 +1111,13 @@ class InterruptiblePdb(Pdb):
                 raise
 
 
-def set_trace(frame=None):
+def set_trace(frame=None, header=None):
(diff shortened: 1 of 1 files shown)


MrMino:
A bit of trivia: funnily enough, the original reason for adding the `header=` argument to Pdb was actually IPython: https://bugs.python.org/issue31389 :smile:

Carreau:
Haha ! Awesome !

Issue #14263, opened 2023-12-10
doesn't easily load_extension() or %load_ext from ~/.ipython/extensions/


blubberdiblub:
I've got a custom extension in `~/.ipython/extensions/` (the path reflected in `get_ipython().extension_manager.ipython_extension_dir` by default).

I expect all extensions in there to load simply with a `%load_ext name` or `get_ipython().extension_manager.load_extension('name')`.

However, they aren't found. I have to resort to own ugly `sys.path` manipulation to get them to load.

I tracked it down to `IPython/core/extensions.py` where [`ExtensionManager._load_extension()`](https://github.com/ipython/ipython/blob/49914f93892640d150cededca16b14b20ce8663d/IPython/core/extensions.py#L87) prepares for searching in the extensions dir by importing `prepended_to_syspath`, but fails to actually make use of that.

Note that [`ExtensionManager.reload_extension()`](https://github.com/ipython/ipython/blob/49914f93892640d150cededca16b14b20ce8663d/IPython/core/extensions.py#L136) imports and **makes use of** `prepended_to_syspath`, giving me a strong feeling that it missing from `_load_extension()` is just an oversight.

```python
{'commit_hash': '49914f938',
 'commit_source': 'installation',
 'default_encoding': 'utf-8',
 'ipython_path': '/home/nboehm/venv3.12/lib/python3.12/site-packages/IPython',
 'ipython_version': '8.18.1',
 'os_name': 'posix',
 'platform': 'Linux-6.5.0-14-lowlatency-x86_64-with-glibc2.38',
 'sys_executable': '/home/nboehm/venv3.12/bin/python3.12',
 'sys_platform': 'linux',
 'sys_version': '3.12.0 (main, Oct  4 2023, 06:27:34) [GCC 13.2.0]'}
```

ivanov:
Hi @blubberdiblub - your intuition is on the right track but going the wrong direction.  What's actually happened is that the `prepend_to_syspath` wasn't removed from enough places when this functionality was removed from IPython in version 8.6.

See #13744

Your simplest workaround is to add that directory to your PYTHONPATH.

An alternative is to install your extensions as regular python modules.

ivanov:
I'm going to go ahead and close this issue, since we've got the vestigial remnants of `ipython_extensions_dir` removed in #14310

Happy hacking! :bowtie:

PR #14310, opened 2024-01-29
Remove old mentions and uses of ipython_extension_dir

ivanov:
The bulk of this functionality was removed in IPython 8.6.0, but there were still mentions of it in the docs and in the reload code.

The extant mentions of `.ipython/extensions` in code came up as a source of confusion in #14263.


Carreau (commented):
Thanks ! It's good to see PRs from you ! ❤️

Carreau (approved):

PR #14312, opened 2024-01-30
`IPython.utils._process_posix`: import `pexpect` only when needed

mkoeppe:
Delaying the import of `pexpect` to the first use.

This is needed to make `import IPython` work on Pyodide; see https://github.com/pyodide/pyodide/pull/4452

### DIFF
--- a/IPython/utils/_process_posix.py
+++ b/IPython/utils/_process_posix.py
@@ -20,8 +20,6 @@ import os
 import subprocess as sp
 import sys
 
-import pexpect
-
 # Our own
 from ._process_common import getoutput, arg_split
 from IPython.utils.encoding import DEFAULT_ENCODING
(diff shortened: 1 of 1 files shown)


hoodmane:
I'd suggest adding a new file making a `_process_emscripten`:
```py
def system(cmd):
    raise OSError("Not available")

def getoutput(cmd):
    raise OSError("Not available")

def check_pid(cmd):
    raise OSError("Not available")

def arg_split(s, posix=False, strict=True):
    """This one could be made to work but it's not clear if it would be useful..."""
    raise OSError("Not available")
```
and then apply this patch:

```patch
--- a/IPython/utils/process.py
+++ b/IPython/utils/process.py
@@ -15,6 +15,8 @@
     from ._process_win32 import system, getoutput, arg_split, check_pid
 elif sys.platform == 'cli':
     from ._process_cli import system, getoutput, arg_split, check_pid
+elif sys.platform == "emscripten":
+    from ._process_emscripten import system, getoutput, arg_split, check_pid
 else:
     from ._process_posix import system, getoutput, arg_split, check_pid
 ```

Carreau:
I'm going to make a release soon, so I'm going to merge this, but +1 to extend with adding a `_process_emscripten.py`

hoodmane:
Thanks @Carreau and @mkoeppe!

PR #14313, opened 2024-01-30
pyproject.toml, setup.py: Declare backend-path

mkoeppe:
Replacing `sys.path.insert(0, ".")` in `setup.py` by declaring a custom PEP-517 build system.


Carreau:
Ohh, interesting, let's try !

PR #14318, opened 2024-02-01
IPython/core/history.py: Handle RuntimeError on Thread.start

mkoeppe:
On the emscripten (Pyodide), `HistorySavingThread.start` raises `RuntimeError`. 
We handle this exception here.

It also reveals a misplaced `atexit` call, which we also fix here.

### DIFF
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -6,15 +6,12 @@
 
 import atexit
 import datetime
-from pathlib import Path
 import re
 import sqlite3
 import threading
+from pathlib import Path
 
-from traitlets.config.configurable import LoggingConfigurable
 from decorator import decorator
-from IPython.utils.decorators import undoc
-from IPython.paths import locate_profile
 from traitlets import (
     Any,
     Bool,
@@ -22,12 +19,16 @@ from traitlets import (
     Instance,
     Integer,
     List,
+    TraitError,
     Unicode,
     Union,
-    TraitError,
     default,
     observe,
 )
+from traitlets.config.configurable import LoggingConfigurable
+
+from IPython.paths import locate_profile
+from IPython.utils.decorators import undoc
 
 #-----------------------------------------------------------------------------
 # Classes and functions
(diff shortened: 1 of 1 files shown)


IPython/core/history.py
```diff
@@ -880,10 +888,10 @@ def __init__(self, history_manager):
         super(HistorySavingThread, self).__init__(name="IPythonHistorySavingThread")
         self.history_manager = history_manager
         self.enabled = history_manager.enabled
-        atexit.register(self.stop)
 
     @only_when_enabled
     def run(self):
+        atexit.register(self.stop)
```
Carreau, line 894: Hum, interesting. I haven't looked at this code in while, and not too familiar with threading.

Would it also be more correct to `atexit.unregister(self.stop)` in `def stop(self)` itself ?

IPython/core/history.py
```diff
@@ -554,7 +555,14 @@ def __init__(self, shell=None, config=None, **traits):
...
             self.save_thread = HistorySavingThread(self)
-            self.save_thread.start()
+            try:
+                self.save_thread.start()
+            except RuntimeError:
+                self.log.error(
+                    "Failed to start history saving thread. History will not be saved.",
+                    exc_info=True,
+                )
+                self.hist_file = ":memory:"
```
Carreau, line 565: +1. Should we _also_ have a condition that check if we are im emscripten to be explicit ?

Carreau (approved):

IPython/core/history.py
mkoeppe: At the moment, I have to say that I don't have a clear picture of what codepaths in IPython have a chance to be useful within emscripten (cf. https://github.com/pyodide/pyodide/pull/4452#discussion_r1475158851). So I'd be reluctant to propose specific checks for the emscripten platform in places that do not already have platform-specific code (in contrast to #14316, thanks for merging that!).

IPython/core/history.py
mkoeppe: Seems reasonable, done in 047ab6e

Carreau:
Thanks !

PR #14325, opened 2024-02-08
more types, everywhere, and remove some ignored files from mypy.

Carreau:


Carreau:
@mkoeppe I have some issues interpreting the sage failure, is that something in this PR, or something else ?

Carreau:
It's all good, failing only on 3.13 so far.

PR #14327, opened 2024-02-11
Retire commands `setup.py symlink`, `setup.py unsymlink`

mkoeppe:
Apparently introduced in #4473 (over 10 years ago), undocumented.


Carreau (approved):
Thanks, that's great ! I was hopping to completely remove setup.py at some point !

Issue #14329, opened 2024-02-12
Multiple calls to matplotlib.pyplot.switch_backend


ianthomas23:
Whilst looking at #14311 I discovered that in many circumstances we are unnecessarily calling `matplotlib.pyplot.switch_backend` twice when we only need to do so once.

Reproducer (using current `main` branches of `ipython`, `matplotlib` and `matplotlib-inline`):
```python
$ ipython
<snip>

In [1]: import matplotlib.pyplot as plt; plt.set_loglevel("debug")

In [2]: %matplotlib tk
DEBUG:matplotlib.pyplot:Loaded backend qtagg version 5.15.9.
DEBUG:matplotlib.pyplot:Loaded backend TkAgg version 8.6.
```
and in JupyterLab:

![Screenshot from 2024-02-12 12-13-35](https://github.com/ipython/ipython/assets/580326/da93bc4c-8a5e-4fe1-84c1-87780b7e3e0e)

The cause is importing `matplotlib_inline.backend_inline.configure_inline_support` https://github.com/ipython/ipython/blob/3d02f18733c0bbf062e8852fdc417609fbf29e51/IPython/core/interactiveshell.py#L3654
before calling `activate_matplotlib`
https://github.com/ipython/ipython/blob/3d02f18733c0bbf062e8852fdc417609fbf29e51/IPython/core/interactiveshell.py#L3669

There is a side-effect of importing `backend_inline` which is to call `_enable_matplotlib_integration`:
https://github.com/ipython/matplotlib-inline/blob/0f419e31f2c3ab70d6bf0f1cd5950f993e01e360/matplotlib_inline/backend_inline.py#L228
which can call `matplotlib.pyplot.switch_backend` with the wrong backend before we call it with the correct one.

The solution is to move the import of `backend_inline` to after the correct `activate_matplotlib(backend)` call and before `configure_inline_support` is called. This works locally for me manually testing in Jupyter and IPython. I'll submit a PR with this to run the full CI suite on it.

An alternative would be to remove the import side-effect in `matplotlib-inline` and lazily call `_enable_matplotlib_integration`, but this will change more code and is potentially more dangerous as other libraries may be relying on the existing call.

[...]

PR #14330, opened 2024-02-12
Delay import from matplotlib_inline.backend_inline

ianthomas23:
Possible fix for #14329.

With this change, the reproducers from the issue become:
```python
$ ipython
<snip>

In [1]: import matplotlib.pyplot as plt; plt.set_loglevel("debug")

In [2]: %matplotlib tk
DEBUG:matplotlib.pyplot:Loaded backend TkAgg version 8.6.
```
and
Screenshot from 2024-02-12 12-44-30

It is still possible to call `matplotlib.pyplot.switch_backend` twice but only if you request `%matplotlib inline`, and the `switch_backend` is called with same correct backend twice. This isn't ideal but it doesn't cause any problems apart from being a little slower than necessary.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3667,6 +3667,7 @@ class InteractiveShell(SingletonConfigurable):
         pt.activate_matplotlib(backend)
 
         from matplotlib_inline.backend_inline import configure_inline_support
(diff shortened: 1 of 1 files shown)


Carreau:
Let's try.

PR #14340, opened 2024-02-17
pyproject.toml (project.entrypoints, project.scripts): Define here

mkoeppe:
... instead of in setup.py, setupbase.py

Another step toward removing setup.py, as requested by @Carreau in https://github.com/ipython/ipython/pull/14327#pullrequestreview-1884755776


Carreau:
Thanks !

PR #14342, opened 2024-02-17
Improve typing and MIME hook API for inspector

### LINKED ISSUE #14339: More complete mime hook API for inspector?
Right now mime hooks get two arguments: `obj` and `info`.

https://github.com/ipython/ipython/blob/e60c06a57de9f1e53fdd4b3a16c6a16c530e4f2a/IPython/core/oinspect.py#L764-L772

This feels redundant because `info` includes `info.obj == obj`. There is no `info_dict` but downstream could just call `.info()`, but there is no `detail_level` so all mime hooks have to return the same result for `obj?` and `obj??`. I am not sure if `omit_sections` should be passed to, it does not appear crucial though.

I think passing `detail_level` is uncontroversial. What do you think about passing `info_dict` too?

krassowski:
Fixes https://github.com/ipython/ipython/issues/14339

### Additions

Adds `InfoDict` type to improve the typing of `info()` result.

Adds missing `"subclasses"` to `info_fields` list (these were added to the field list in https://github.com/ipython/ipython/pull/11486 but we forgot to update `info_fields` variable at the time) - the newly added `InfoDict` type will ensure that this won't happen again.

Adds `InspectorHookData` dataclass which is passed to the MIME hooks which now should expect a single argument. Having a single dataclass argument enables us to deprecate individual fields, or add new fields without breaking the existing hooks. The old hooks will still work (if any are out there since this mechanism got just added in the previous point version).

### Deletions

A comment over `info_fields` gets deleted:
- Contrarily to the comment (which is getting deleted in this PR), `info_fields` were not defining the order of display since at least 2015 (https://github.com/ipython/ipython/pull/7903 - I did not feel the need to go further in the history to find when exactly it happened).
- Also contrarily to this comment, current Jupyter messaging spec does not define the contents of `info_fields` (I guess this was lost during IPython/Jupyter split), but the newly added `InfoDict` at least properly annotates their type (if you know where I can find the old IPython messaging spec with the descriptions I can add these as doc comments).

Unused `cast_unicode` import gets deleted. If someone imported it from here... well they really should not have.

### Deprecations

- mime hooks taking two arguments (`obj, info`)

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -145,12 +145,7 @@ class InspectorHookData:
 
 @undoc
 def object_info(**kw):
-    """DEPRECATED: Make an object info dict with all fields present."""
-    warnings.warn(
-        "IPython.core.oinspect.object_info has been deprecated since IPython 8.22 and will be removed in future versions",
-        DeprecationWarning,
-        stacklevel=2,
-    )
+    """Make an object info dict with all fields present."""
     infodict = {k: None for k in info_fields}
     infodict.update(kw)
     return infodict
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -24,9 +24,7 @@ import os
 import types
 import warnings
 
-from typing import Any, Optional, Dict, Union, List, Tuple
-
-from typing import TypeAlias
+from typing import cast, Any, Optional, Dict, Union, List, TypedDict, TypeAlias, Tuple
 
 import traitlets
 
@@ -42,7 +40,6 @@ from IPython.utils.text import indent
 from IPython.utils.wildcard import list_namespace
 from IPython.utils.wildcard import typestr2type
 from IPython.utils.coloransi import TermColors
-from IPython.utils.py3compat import cast_unicode
 from IPython.utils.colorable import Colorable
 from IPython.utils.decorators import undoc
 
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, that is great, it is something I wanted to do for quite some time but needed to find the time to do it. 

I was envisioning a DataClass, but a typedict is great !

I'll try to get to that and merge it this week

IPython/core/oinspect.py
```diff
@@ -106,21 +103,55 @@ def pylight(code):
...
+    info_dict: InfoDict
+    detail_level: int
+    omit_sections: list[str]
 
 
+@undoc
 def object_info(**kw):
-    """Make an object info dict with all fields present."""
-    infodict = {k:None for k in info_fields}
+    """DEPRECATED: Make an object info dict with all fields present."""
```
krassowski, line 148: Maybe it should not get deprecated because it is used on an unhappy path in IPython codebase:

https://github.com/ipython/ipython/blob/6f35f84e95c16700f9c185a962e5006a584581bb/IPython/core/interactiveshell.py#L1862-L1871

I still _think_ it is a utility which does not need to be documented/public.

krassowski: Resolved in https://github.com/ipython/ipython/pull/14342/commits/6154002c46c92b0b8750930cc4411ceaf1facc8d

Carreau: When I'm unsure about deprecating something I do a `PendingDeprecationWarning` without a date/version and point to an issue for discussing (non)-removal. Like [this]
(https://github.com/ipython/ipython/blob/b479c4a03bb4112c6aeea262a99c0bcbb18bdfa0/IPython/utils/coloransi.py#L123-L131) for example.

Carreau (approved):

Carreau:
Ok, let's get that in and maybe tweak the behavior in subsequent PRs if we decide to.

PR #14348, opened 2024-02-19
IPython/utils/path.py: Deprecate unused `target_outdated`, `target_update`

mkoeppe:
These functions are unused. I propose to deprecate them

(`setupbase.py` brings its own copy of these functions.)


Carreau:
thanks !

Issue #14364, opened 2024-03-06
Documentation about available options and their structural hierarchy


furushchev:
Hi, I'm looking for a list of the available options and I still cannot find in the latest documentation.

There were the pages for them in docs in previous versions:
https://ipython.org/ipython-doc/dev/config/options/index.html

Is there any page that I'm missing or is there any way to get the equivalent information using IPython shell?

i.e. I found the following magic command tells the available options for the `TerminalPythonApp` class, but I want comprehensive list of available options other than the class as well.

```py
%config TerminalIPythonApp
TerminalIPythonApp(BaseIPythonApplication, InteractiveShellApp) options
---------------------------------------------------------------------
TerminalIPythonApp.add_ipython_dir_to_sys_path=<Bool>
...

%config InteractiveShellApp
UsageError: Invalid config statement: 'InteractiveShellApp'
# No information available for the super classes
```

I found a part of the information is described in the latest docs:
https://ipython.readthedocs.io/en/stable/config/intro.html

But I still feel it difficult to customize options because:
- There is no list of available options for each class
- The structural hierarchy of classes is unclear (i.e. I have no idea where is the best place to put option values because the relationship of the similar classes:`TerminalIPythonApp`, `InteractiveShellApp`, `InteractiveShell`)

Thank you!

krassowski:
Indeed a page like this: https://jupyter-server.readthedocs.io/en/latest/other/full-config.html could be useful.

It gets created with a [custom script in `conf.py`](https://github.com/jupyter-server/jupyter_server/blob/49d8cab8f466698d475356b53b8c8ba9218c46f1/docs/source/conf.py#L341-L385C29) using traitlet's `document_config_options()` which used to be used in IPython too but is currently commented out:

https://github.com/ipython/ipython/blob/4f6eb6f21fea6d0e5a55cad7e6fc462dd669e3a6/docs/autogen_config.py#L101-L127

Ouch commented out for last 7 years. I guess maybe worth sending a PR trying to uncomment it?

furushchev:
@krassowski Thank you for the insightful comment!

I've just made a PR to re-enable the pages.

PR #14365, opened 2024-03-06
Uncomment app.document_config_options() for adding IPython options pages to Configuring IPython section

furushchev:
This change, discussed at #14364, adds `IPython options` pages to the `Configuring IPython` section, which lists available options for the classes.


furushchev:
Seems like it is already generated in different style?
https://ipython.readthedocs.io/en/stable/config/options/terminal.html

PR #14366, opened 2024-03-11
Disable scheduled CI runs on forks

ianthomas23:
The three scheduled CI runs are running on forks of the main repo, e.g. https://github.com/ianthomas23/ipython/actions. They can be disabled by each fork owner, but this PR disables them for all forks.


Carreau:
Thanks, sounds reasonable.

PR #14371, opened 2024-03-19
Move Matplotlib backend mapping to Matplotlib

### LINKED ISSUE #14311: Move backend mapping to Matplotlib
I wanted to draw your attention to matplotlib/matplotlib#27663, about moving the Matplotlib backend mappings out of IPython and into Matplotlib.

The primary use case it so support Matplotlib widgets (`ipympl` and `matplotlib-inline`) registering themselves as Matplotlib backends without requiring additional code in IPython and/or Matplotlib. The secondary use case is to support backends in IPython using Matplotlib's `module://name.of.the.backend` syntax, e.g.
```
%matplotlib module://mplcairo.backend
```
which one can already do using `matplotlib.use(...)` but not directly via the `%matplotlib` magic.

Whilst doing this it seems sensible to bring all of the backend registering and mapping together in one place, and that should be Matplotlib rather than IPython. I am not sure how easy (or even possible!) it will be to remove all the related hard-coded stuff in IPython, but I am willing to start and see how it goes.

ianthomas23:
This is WIP to move IPython's backend mapping into Matplotlib and extends it to allow backends to register themselves via [entry points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins). It is a coordinated effort across Matplotlib, IPython, matplotlib-inline and ipympl. Closes #14311. Most of the work is in Matplotlib (matplotlib/matplotlib#27948) but I will repeat the relevant points here.

When using a Matplotlib magic command of the form `%matplotlib something` the identification of the Matplotlib backend and GUI loop are now performed in Matplotlib not IPython. This supports:

1. Matplotlib backends that IPython already supports such as `qtagg` and `tkagg`.
2. Other built-in Matplotlib backends such as `qtcairo`.
3. Backends that use `module://something` syntax such as `module://mplcairo.qt`.
4. Backends that self-register using entry points, currently `inline` and `widget`/`ipympl`.

Implementation details:

1. The magic command is now explicitly `%matplotlib gui_or_backend` rather than `%matplotlib gui`. This is already effectively the case as e.g. `%matplotlib ipympl` is really a backend not GUI name. Within Matplotlib the `gui_or_backend` is checked first to see if it is a GUI name and then falls back to checking if it is a backend name.
2. If you select the `inline` backend the corresponding GUI is now `None` not `inline`. All backends which do not have a GUI loop return `None`, otherwise we have to keep explicit checks within IPython for particular backends.
3. `backends` and `backend2gui` are now deprecated but are still exposed, with a deprecation warning, if required. If using Matplotlib, ipympl, etc releases that include the corresponding changes to this PR then they are not needed as Matplotlib deals with it all. But for backward compatibility they must still be available for a while along with the remainder of the existing backend-handling code.
4. I haven't yet updated the documentation but we need to keep information about valid GUI frameworks and I propose that we should remove all lists of valid Matplotlib backends, replacing them with instructions on how to obtain the current list (pointing to the Matplotlib docs and using `%matplotlib --list`). If we keep any lists then they will inevitably become out of date. This extends to the `backend_keys` in IPython/core/shellapp.py.

Because the four related projects are loosely coupled without direct dependencies on each other (except for `ipython` and `matplotlib-inline`), backward compatibility requires all possible combinations of projects before and after the new functionality (I will call these "old" and "new" from now on) to continue to work. I have tested these all locally, and the CI of this PR will test new IPython against old Matplotlib for example, but I need to add one or more new temporary CI runs to test new IPython against new Matplotlib etc. The identification of new versus old depends on version checks on the other libraries, so here is a table that I will update showing the current status of progress in the 4 projects:

| Project | Relevant PRs | Possible release version |
| --- | --- | --- |
| matplotlib-inline | ipython/matplotlib-inline#34, ipython/matplotlib-inline#35 | 0.1.7 |
| ipympl | matplotlib/ipympl#549 | 0.9.4 |
| Matplotlib | matplotlib/matplotlib#27948 | 3.9.0 |
| IPython | #14371 (this) | 8.24.0 |

The two widget projects can be released soon, once we are happy with the entry point approach. The other two projects' PRs will have to be synchronised as each includes version checks on each other.

To do

- [ ] Add CI runs against the new PR branches of the other projects.
- [ ] Add comments for conditions required for backward-compatibility code blocks to be removed.
- [ ] Update documentation, including removal of lists of valid backends.
- [ ] Update version checks before merging.

### DIFF
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -93,17 +93,12 @@ class PylabMagics(Magics):
         """
         args = magic_arguments.parse_argstring(self.matplotlib, line)
         if args.list:
-            from IPython.core.pylabtools import _matplotlib_manages_backends
+            from IPython.core.pylabtools import _list_matplotlib_backends_and_gui_loops
 
-            if _matplotlib_manages_backends():
-                from matplotlib.backends.registry import backend_registry
-
-                backends_list = backend_registry.list_all()
-            else:
-                from IPython.core.pylabtools import backends
-
-                backends_list = list(backends.keys())
-            print("Available matplotlib backends: %s" % backends_list)
+            print(
+                "Available matplotlib backends: %s"
+                % _list_matplotlib_backends_and_gui_loops()
+            )
         else:
             gui, backend = self.shell.enable_matplotlib(args.gui.lower() if isinstance(args.gui, str) else args.gui)
             self._show_matplotlib_backend(args.gui, backend)
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -24,12 +24,13 @@ from warnings import warn
 #-----------------------------------------------------------------------------
 
 magic_gui_arg = magic_arguments.argument(
-        'gui', nargs='?',
-        help="""Name of the matplotlib backend to use such as 'qt' or 'widget'.
+    "gui",
+    nargs="?",
+    help="""Name of the matplotlib backend to use such as 'qt' or 'widget'.
         If given, the corresponding matplotlib backend is used,
         otherwise it will be matplotlib's default
         (which you can set in your matplotlib config file).
-        """
+        """,
 )
 
 
@@ -93,11 +94,14 @@ class PylabMagics(Magics):
         args = magic_arguments.parse_argstring(self.matplotlib, line)
         if args.list:
             from IPython.core.pylabtools import _matplotlib_manages_backends
+
             if _matplotlib_manages_backends():
                 from matplotlib.backends.registry import backend_registry
+
                 backends_list = backend_registry.list_all()
             else:
                 from IPython.core.pylabtools import backends
+
                 backends_list = list(backends.keys())
             print("Available matplotlib backends: %s" % backends_list)
         else:
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -18,7 +18,6 @@ from IPython.core import magic_arguments
 from IPython.core.magic import Magics, magics_class, line_magic
 from IPython.testing.skipdoctest import skip_doctest
 from warnings import warn
-from IPython.core.pylabtools import backends
 
 #-----------------------------------------------------------------------------
 # Magic implementation classes
@@ -26,11 +25,11 @@ from IPython.core.pylabtools import backends
 
 magic_gui_arg = magic_arguments.argument(
         'gui', nargs='?',
-        help="""Name of the matplotlib backend to use %s.
+        help="""Name of the matplotlib backend to use such as 'qt' or 'widget'.
         If given, the corresponding matplotlib backend is used,
         otherwise it will be matplotlib's default
         (which you can set in your matplotlib config file).
-        """ % str(tuple(sorted(backends.keys())))
+        """
 )
 
 
@@ -93,7 +92,13 @@ class PylabMagics(Magics):
         """
         args = magic_arguments.parse_argstring(self.matplotlib, line)
         if args.list:
-            backends_list = list(backends.keys())
+            from IPython.core.pylabtools import _matplotlib_manages_backends
+            if _matplotlib_manages_backends():
+                from matplotlib.backends.registry import backend_registry
+                backends_list = backend_registry.list_all()
+            else:
+                from IPython.core.pylabtools import backends
+                backends_list = list(backends.keys())
             print("Available matplotlib backends: %s" % backends_list)
         else:
             gui, backend = self.shell.enable_matplotlib(args.gui.lower() if isinstance(args.gui, str) else args.gui)
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -44,7 +44,9 @@ _deprecated_backends = {
 # GUI support to activate based on the desired matplotlib backend.  For the
 # most part it's just a reverse of the above dict, but we also need to add a
 # few others that map to the same GUI manually:
-_deprecated_backend2gui = dict(zip(_deprecated_backends.values(), _deprecated_backends.keys()))
+_deprecated_backend2gui = dict(
+    zip(_deprecated_backends.values(), _deprecated_backends.keys())
+)
 # In the reverse mapping, there are a few extra valid matplotlib backends that
 # map to the same GUI support
 _deprecated_backend2gui["GTK"] = _deprecated_backend2gui["GTKCairo"] = "gtk"
@@ -279,7 +281,7 @@ def select_figure_formats(shell, formats, **kwargs):
 
     [ f.pop(Figure, None) for f in shell.display_formatter.formatters.values() ]
     mplbackend = matplotlib.get_backend().lower()
-    if mplbackend in ('nbagg', 'ipympl', 'widget', 'module://ipympl.backend_nbagg'):
+    if mplbackend in ("nbagg", "ipympl", "widget", "module://ipympl.backend_nbagg"):
         formatter = shell.display_formatter.ipython_display_formatter
         formatter.for_type(Figure, _reshow_nbagg_figure)
 
@@ -330,15 +332,16 @@ def find_gui_and_backend(gui=None, gui_select=None):
     """
 
     import matplotlib
+
     if _matplotlib_manages_backends():
         backend_registry = matplotlib.backends.registry.backend_registry
 
         # gui argument may be a gui event loop or may be a backend name.
         if gui in ("auto", None):
-            backend = matplotlib.rcParamsOrig['backend']
+            backend = matplotlib.rcParamsOrig["backend"]
             backend, gui = backend_registry.resolve_backend(backend)
         else:
-             backend, gui = backend_registry.resolve_gui_or_backend(gui)
+            backend, gui = backend_registry.resolve_gui_or_backend(gui)
 
         return gui, backend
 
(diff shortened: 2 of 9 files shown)


Carreau:
Sorry I was a bit overwhelmed this month, I'll try to push this forward for next release.

ianthomas23:
> Sorry I was a bit overwhelmed this month, I'll try to push this forward for next release.

That's OK, there are still things I need to do to this before it is reviewable. Maybe you should just ignore it until I confirm when it is ready for review.

IPython/core/magics/pylab.py
```diff
@@ -18,19 +18,19 @@
...
-        'gui', nargs='?',
-        help="""Name of the matplotlib backend to use %s.
+    "gui",
+    nargs="?",
+    help="""Name of the matplotlib backend to use such as 'qt' or 'widget'.
         If given, the corresponding matplotlib backend is used,
         otherwise it will be matplotlib's default
         (which you can set in your matplotlib config file).
-        """ % str(tuple(sorted(backends.keys())))
+        """,
```
Carreau, line 33: 👍

IPython/core/pylabtools.py
```diff
@@ -12,9 +12,12 @@
 from IPython.core.display import _pngxy
 from IPython.utils.decorators import flag_calls
 
-# If user specifies a GUI, that dictates the backend, otherwise we read the
-# user's mpl default from the mpl rc structure
-backends = {
+
+# Matplotlib backend resolution functionality moved from IPython to Matplotlib
+# in IPython 8.23 and Matplotlib 3.9. Need to keep `backends` and `backend2gui`
```
Carreau, line 17:
```suggestion
# in IPython 8.24 and Matplotlib 3.9. Need to keep `backends` and `backend2gui`
```

IPython/core/pylabtools.py
```diff
@@ -41,29 +44,40 @@
...
+del _deprecated_backend2gui["pdf"]
+del _deprecated_backend2gui["ps"]
+del _deprecated_backend2gui["module://matplotlib_inline.backend_inline"]
+del _deprecated_backend2gui["module://ipympl.backend_nbagg"]
+
+
+# Deprecated attributes backends and backend2gui mostly following PEP 562.
+def __getattr__(name):
+    if name in ("backends", "backend2gui"):
+        warnings.warn(f"{name} is deprecated", DeprecationWarning)
```
Carreau, line 77:
```suggestion
        warnings.warn(f"{name} is deprecated since IPython 8.24, backends are managed in matplotlib and can be externally registered.", DeprecationWarning)
```

IPython/core/pylabtools.py
```diff
@@ -319,7 +333,23 @@ def find_gui_and_backend(gui=None, gui_select=None):
 
     import matplotlib
 
-    has_unified_qt_backend = getattr(matplotlib, "__version_info__", (0, 0)) >= (3, 5)
+    if _matplotlib_manages_backends():
+        backend_registry = matplotlib.backends.registry.backend_registry
```
Carreau, line 337: Does that need explicit import of `matplotlib.backends` ? and/or other hasattr checks ?

```
In [1]: import matplotlib

In [2]: matplotlib.backends
...
AttributeError: module 'matplotlib' has no attribute 'backends'

In [3]: import matplotlib.backends
```

IPython/core/pylabtools.py
```diff
@@ -346,6 +377,10 @@ def find_gui_and_backend(gui=None, gui_select=None):
             gui = gui_select
             backend = backends_[gui]
 
+    # Since IPython 8.23.0 use None for no gui event loop rather than "inline".
```
Carreau, line 380:
```suggestion
    # Since IPython 8.24.0 use None for no gui event loop rather than "inline".
```

IPython/core/pylabtools.py
```diff
@@ -346,6 +377,10 @@ def find_gui_and_backend(gui=None, gui_select=None):
...
+    if gui == "inline":
+        gui = None
+
```
Carreau, line 383: Should we emit a warning here, to make sure that things like backend_to_gui never return `"inline"` or similar ?

IPython/core/pylabtools.py
```diff
@@ -431,3 +466,10 @@ def configure_inline_support(shell, backend):
     )
 
     configure_inline_support_orig(shell, backend)
+
+
+def _matplotlib_manages_backends():
+    import matplotlib
+
+    mpl_version_info = getattr(matplotlib, "__version_info__", (0, 0))
+    return mpl_version_info >= (3, 9)
```
Carreau, line 475: :+1:

IPython/core/tests/test_pylabtools.py
```diff
@@ -268,3 +268,87 @@ def test_figure_no_canvas():
...
+        "inline"
+        if pt._matplotlib_manages_backends()
+        else "module://matplotlib_inline.backend_inline"
+    )
+    assert backend == expected_backend
+
+
+def test_backend_unknown():
+    with pytest.raises(RuntimeError if pt._matplotlib_manages_backends() else KeyError):
+        pt.find_gui_and_backend("name-does-not-exist")
```
Carreau, line 354: :+1: thanks for all the tests.

Carreau (approved):

Carreau:
Thanks for your patience.

Misc questions, if you tell me this is ready we can merge as is and we can resolve the questions in subsequent iterations.

ccordoba12:
Quick question: does some work also need to be done in IPykernel to support this? I say it because it has its own logic to handle backends, which is independent of IPython.

Carreau:
I think this is fine, I can't find any direct reference to this in IPykernel, and I believe this is why we have matplotlib-inline; to not have this cross-dependency. It might create issues with much older ipykernel though, but I think it's fine to require more recent ones.

ianthomas23:
I can confirm that no changes are required in current IPyKernel for this. The backend handling hasn't changed, just where the mapping from matplotlib magic to backend and gui loop occurs. I have been testing with Jupyter as I go along; this work is sponsored by QuantStack so Jupyter support is very much at the forefront of my mind.

There are a couple of places in IPyKernel that use backend to gui loop mapping and hardcoded backend names (`matplotlib-inline`) that could be improved in future as this information could come from Matplotlib. I will deal with these.

ccordoba12:
Ok, thanks for the confirmation @ianthomas23 and @Carreau!

IPython/core/pylabtools.py
ianthomas23: Done

IPython/core/pylabtools.py
ianthomas23: Done

IPython/core/pylabtools.py
ianthomas23: Done

IPython/core/pylabtools.py
ianthomas23: The `_matplotlib_manages_backends()` call checks that the version of Matplotlib is late enough to include the required `backed_registry` functions, so the import *should* always work. But now you have drawn my attention to this I realise it would be better to be explicit and use `hasattr` checks for `backend_registry` and its function `resolve_gui_or_backend`. This would avoid us having to have a hardcoded Matplotlib version and would allow us to merge this work before we are entirely sure what version of Matplotlib will include the corresponding changes.

Carreau:
I think you may have forgotten to push ? You replied "Done" in a few place but I don't see the changes...

IPython/core/pylabtools.py
ianthomas23: We only get to this code if we are using "old" Matplotlib (i.e. before the `backend_registry` changes), in which case we will have `gui == "inline"` for the inline backend. There will already have been a `DeprecationWarning` issued for accessing `backend2gui` (a few lines above) and a second warning here might be more confusing than helpful.

IPython/core/pylabtools.py
ianthomas23: As I mentioned in https://github.com/ipython/ipython/pull/14371#discussion_r1555393766 above, I am inclined to change this check to use `hasattr` rather than version number.

IPython/core/pylabtools.py
ianthomas23: I will add an extra comment in the code here though.

ianthomas23:
> I think you may have forgotten to push ? You replied "Done" in a few place but I don't see the changes...

Yes, I got distracted :smile:.

ianthomas23:
I've added a fix for `ipython --matplotlib=something` and `--pylab=something`. These previously used a hardcoded list derived from the now-deprecated `backends` module-level attribute, and now they get the list from Matplotlib. I am lazily-evaluating this list as people using IPython without Matplotlib don't need the information. To do this I have subclassed the `CaselessStrEnum` traitlet:

https://github.com/ianthomas23/ipython/blob/b1d9d89eda82337fdc7f47689b024a93a38342c0/IPython/core/shellapp.py#L114

so that initially the list of possible `values` is `None` and it is only correctly populated when first accessed (via `__getattribute__`). It is not much code but I am not sure if this is the best solution. I am not very familiar with `traitlets` and I couldn't find a better approach.

Carreau:
Ok, let's get that in.

ianthomas23:
> Ok, let's get that in.

Thanks. I'll update the docs separately.

When you want to do the next release I should probably look over the code again to check it is not about to break anything.

Carreau:
> When you want to do the next release I should probably look over the code again to check it is not about to break anything.

I usually do the releases on last Friday of each month, so next release should be in about 10 days. If there is a critical issue I can yank and we "only" have to wait another month. So it's ok.

PR #14378, opened 2024-03-30
add __repr__ to MagicsDisplay

iloveitaly:
Here's more details on the `rich` side:

https://github.com/Textualize/rich/issues/3317

I don't see a good reason not to overload `__repr__` in this use case


Carreau:
Seem reasonable.

Issue #14382, opened 2024-04-02
Slow debugger in spyder


Carreau:
see https://github.com/spyder-ide/spyder/issues/20571

Issue #13972, opened 2023-03-13
The debugger calls f_locals on all frames


impact27:
To reproduce, debug the following code line by line, pressing next on each line:

```
import pandas as pd
import plotly.express as px
import plotly.io as pio
pio.renderers.default='svg'  # this line required for plots to show in Spyder Plots panel

df = pd.DataFrame([
    dict(Task="Job A", Start='2009-01-01', Finish='2009-02-28'),
    dict(Task="Job B", Start='2009-03-05', Finish='2009-04-15'),
    dict(Task="Job C", Start='2009-02-20', Finish='2009-05-30')
])

fig = px.timeline(df, x_start="Start", x_end="Finish", y="Task")
fig.update_yaxes(autorange="reversed") # otherwise tasks are listed from the bottom up
fig.show()
```

This results in: `AttributeError: 'NoneType' object has no attribute 'layout'`

The problem is that the IPython debugger calls `frame.f_locals` on every step on every call. Calling `frame.f_locals` will reset the locals of the frame, and plotly does this:
https://github.com/plotly/plotly.py/blob/a4b9887bc7d5f5151b803221fa650291840d150e/packages/python/plotly/plotly/express/_chart_types.py#L424

The second issue is that looping on all frames on each python instruction is slow:
https://github.com/spyder-ide/spyder/issues/20571#issuecomment-1454087578

mlucool:
> The second issue is that looping on all frames on each python instruction is slow:
https://github.com/spyder-ide/spyder/issues/20571

Another test to try:
```
import time

def f():
    g()

def g():
    start = time.time()
    x = 0
    for i in range(10_000_000):
        x += i*i
    print(time.time() - start)
```

![image](https://github.com/ipython/ipython/assets/1813603/aeb27f97-7a29-4174-9bb3-f32f41016806)

ccordoba12:
@mlucool, our solution was to disable the IPython `debuggerskip` functionality (see https://github.com/spyder-ide/spyder-kernels/pull/458). That's not that useful by itself because users need to mark what frames they want to skip first.

mlucool:
Thanks @ccordoba12. I deleted my comment in part because it turns out debuggerskip was not adding that much for that specific test (that is, it matters, but not as much as something I am yet to figure out)

PR #14386, opened 2024-04-02
Cache debugger skip frame predicate

Carreau:
I'm not 100% sure this is correct as technically the value of
__debugger_skip__ could change in the current frame while we are
stepping into it, but that is likely super rare, and the slowdown
that this create is problematic.

There is still a small overhead for me, but this should make the
experience much better.

See https://github.com/spyder-ide/spyder-kernels/pull/458, https://github.com/ipython/ipython/issues/13972, https://github.com/spyder-ide/spyder/issues/20571, https://github.com/ipython/ipython/issues/14382

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -100,6 +100,8 @@ All the changes since then are under the same license as IPython.
 #
 #*****************************************************************************
 
+from __future__ import annotations
+
 import inspect
 import linecache
 import os
@@ -111,6 +113,12 @@ from functools import lru_cache
 from IPython import get_ipython
 from IPython.core.excolors import exception_colors
 from IPython.utils import PyColorize, coloransi, py3compat
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # otherwise circular import
+    from IPython.core.interactiveshell import InteractiveShell
 
 # skip module docstests
 __skip_doctest__ = True
@@ -189,6 +197,8 @@ class Pdb(OldPdb):
     See the `skip_predicates` commands.
 
     """
+
+    shell: InteractiveShell
 
     if CHAIN_EXCEPTIONS:
         MAX_CHAINED_EXCEPTION_DEPTH = 999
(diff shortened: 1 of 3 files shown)


Carreau:
Here is a quick way to test:
```python
$ ipython
Python 3.11.0 | packaged by conda-forge | (main, Oct 25 2022, 06:21:25) [Clang 14.0.4 ]
IPython 8.24.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: !cat test.py
from time import perf_counter

now = perf_counter()
for x in range(1000000):
    if x % 100_000 == 0:
        print(x)
print("done", perf_counter() - now)

In [2]: %run -d test.py
Breakpoint 1 at /Users/bussonniermatthias/test.py:1
> /Users/bussonniermatthias/test.py(1)<module>()
1---> 1 from time import perf_counter
      2
      3 now = perf_counter()
      4 for x in range(1000000):
      5     if x % 100_000 == 0:

ipdb> skip_predicates debuggerskip true
ipdb> c
0
100000
200000
300000
400000
500000
600000
700000
800000
900000
done 3.4291042499244213

In [3]: %run -d test.py
Breakpoint 1 at /Users/bussonniermatthias/test.py:1
> /Users/bussonniermatthias/test.py(1)<module>()
1---> 1 from time import perf_counter
      2
      3 now = perf_counter()
      4 for x in range(1000000):
      5     if x % 100_000 == 0:

ipdb> skip_predicates debuggerskip false
ipdb> c
0
100000
200000
300000
400000
500000
600000
700000
800000
900000
done 3.189767000032589
```

On 8.23, the version with `skip_predicates debuggerskip false`, will be unbearably slow (18sec vs 3.5sec on my machine)

You can try to change the values of debugger skip before entering the debugger.

```python
ip = get_ipython()
if not ip.InteractiveTB.pdb:
    ip.InteractiveTB.pdb = ip.InteractiveTB.debugger_cls()

ip.InteractiveTB.pdb._predicates['debuggerskip'] = False # True by default.
```

PR #14403, opened 2024-04-13
Fix use of --pylab=auto and --matplotlib=auto

### LINKED ISSUE #14401: matplotlib plots not showing when using "ipython --pylab"
In versions of ipython up to 8.21.0, when invoked with `ipython --pylab`, matplotlib plots are displayed to the screen automatically, and without blocking. In later versions (8.22.* and 8.23.0), they are not, and can only be displayed using `plt.show(block=True)`.

Here are steps to reproduce. This uses `pip` installation on Ubuntu for a simple example (where `/usr/bin/python3` is python 3.10.12), although I have also observed the issue in conda environments on both Ubuntu and CentOS, based on python 3.11 and 3.12.

----

**1) Working example -- with ipython 8.21.0:**
```
$  python3 -m venv /tmp/myenv
$  . /tmp/myenv/bin/activate
(myenv) $  pip install pyqt5 numpy matplotlib ipython==8.21.0
(myenv) $  ipython --pylab
In [1]:  plt.plot(np.arange(10)**2)
```
A plot appears on the screen automatically (without needing to call `plt.show` explicitly) -- and in non-blocking mode, so you can continue to type at the ipython `In [2]:` prompt.

<details>
   <summary>Screenshot of plot window</summary>
   
</details>

----

**2) Broken example with ipython 8.23.0:**
- use the same commands as above, except `ipython==8.23.0`

In this case, a matplotlib window is opened, but it is black (with no plot). 

<details>
   <summary>Screenshot of plot window</summary>
   
</details>

If `plt.show()` or `plt.show(block=False)` is typed, neither of these succeed in displaying the plot -- and if `plt.show(block=False)` is typed then not long afterwards a message from GNOME appears saying "matplotlib is not responding".  The plot is only shown if `plt.show(block=True)` is typed, in which case, it blocks as requested (so no futher prompt is issued until the plot is closed).

----

In these examples, matplotlib was version 3.8.4, although further experimentation shows that it does not seem to depend on the matplotlib version.

Please can you restore the behaviour that is obtained in earlier versions.

ianthomas23:
Fixes #14401 which has been a bug in the 8.22.x and 8.23.x releases.

When I removed the multiple initialisation of Matplotlib backends in #14330 it broke use of the following:
```bash
ipython --matplotlib
ipython --matplotlib=auto
ipython --pylab
ipython --pylab=auto
```
by failing to display Matplotlib plot. If you specify a particular GUI event loop such as using
```bash
ipython --pylab=qt
```
then it was and is fine. So for anyone finding this, the workaround until the next release is to specify a GUI loop rather than relying on the auto selection.

I didn't notice this as I've been concentrating on moving the Matplotlib backend logic from IPython to Matplotlib, and with those changes (matplotlib/matplotlib#27948) the above use cases all work OK.

The fix is to reintroduce the early import of `matplotlib-inline` but only if both the gui loop is not specified and the Matplotlib version is before the movement of the backend logic across to it.

There are no explicit tests for this. In the future I will try to think of some tests for some of this IPython-Matplotlib functionality that don't involve installing complicated backend dependencies or adding image comparison tests.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3654,6 +3654,13 @@ class InteractiveShell(SingletonConfigurable):
             make sense in all contexts, for example a terminal ipython can't
             display figures inline.
         """
+        from .pylabtools import _matplotlib_manages_backends
+
+        if not _matplotlib_manages_backends() and gui in (None, "auto"):
+            # Early import of backend_inline required for its side effect of
+            # calling _enable_matplotlib_integration()
+            import matplotlib_inline.backend_inline
+
         from IPython.core import pylabtools as pt
         gui, backend = pt.find_gui_and_backend(gui, self.pylab_gui_select)
 


Carreau:
Thanks, 

I'm marking for next release (in two weeks), if we have more complaints I can backport and make patches releases.

PR #14413, opened 2024-04-20
Fix compatibility with pytest 8

frenzymadness:
setup/teardown nose methods are no longer supported by pytest 8. https://docs.pytest.org/en/latest/deprecations.html#setup-teardown


Carreau:
Oh, interesting. Thanks for taking the time !

SamyDgz (approved):

PR #14420, opened 2024-04-26
Convert "osx" gui framework to/from "macosx" in Matplotlib

ianthomas23:
This is a correction to the code that moves backend handling to Matplotlib, which came to light during the review of matplotlib/matplotlib#27948. The GUI framework of macOSX is called `osx` in IPython but `macosx` in Matplotlib.

It would be possible to allow passing a GUI framework of `osx` to Matplotlib and internally converting it to `macosx`, but the reverse direction is problematic as we would like the answer to be `osx` if we are using IPython and `macosx` if using pure Matplotlib. Therefore the simplest solution is to do the translation in IPython. It is not ideal as we want to minimise such Matplotlib-related implementation details in IPython, but the changes here are small and simple and hence I think they are acceptable.

There are two new private functions `_convert_gui_to_matplotlib` and `_convert_gui_from_matplotlib` to do the required checking and conversion, and these are called whenever a GUI framework name is passed to or from Matplotlib. There are only 3 places in the code where this is currently required.

Inevitably this comes to light just after the release of IPython 8.24.0! But it is not a problem for end users until the next Matplotlib release which contains matplotlib/matplotlib#27948. If that occurs really quickly (sometime in May?) perhaps we could release an IPython 8.24.1 just beforehand, otherwise the usual planned release at the end of next month would be fine.

### DIFF
--- a/IPython/core/pylabtools.py
+++ b/IPython/core/pylabtools.py
@@ -345,8 +345,10 @@ def find_gui_and_backend(gui=None, gui_select=None):
             backend = matplotlib.rcParamsOrig["backend"]
             backend, gui = backend_registry.resolve_backend(backend)
         else:
+            gui = _convert_gui_to_matplotlib(gui)
             backend, gui = backend_registry.resolve_gui_or_backend(gui)
 
+        gui = _convert_gui_from_matplotlib(gui)
         return gui, backend
 
     # Fallback to previous behaviour (Matplotlib < 3.9)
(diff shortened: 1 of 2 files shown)


ianthomas23:
I've pushed a couple more commits to this to be consistent with changes made following review of the corrresponding Matplotlib PR. They are:

- There are no explicit `qt6agg` and `qt6cairo` backends in Matplotlib, they are the unversioned `qtagg` and `qtcairo` instead.
- When requesting a `"headless"` GUI framework, Matplotlib returns `None` rather than `"headless"`.

Carreau:
Thanks !

ianthomas23:
The Matplotlib PR for this (matplotlib/matplotlib#27948) has been merged and will be in 3.9.0 which is at RC stage and should be released later this month. There are no other changes required at either end for this all to work as intended, once that and IPython 8.25 are released.

PR #14424, opened 2024-04-29
Fix downstream ipykernel tests

ianthomas23:
Locally I can get the downstream `ipykernel` tests to pass just by removing the explicit downgrade of `pytest` and `pytest_asyncio`, so here trying it in CI.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -21,7 +21,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.10"]
         include:
-          - os: macos-latest
+          - os: macos-13
             python-version: "3.10"
 
     steps:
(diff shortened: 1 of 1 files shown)


ianthomas23:
This does fix the `ipykernel` tests. The subsequent `sagemath` tests also pass on ubuntu but fail on macos. I've tried switching to a specific mac Intel runner (`macos-latest` recently changing to be ARM) and playing with the `sagemath` pins but it hasn't helped. I am not sure how to continue.

Carreau:
Thanks, merging. I also gave you triage right on the repo, as you've been contributing a lot. Hopefully that will make your life a bit easier. You don't have to use them though. Just recognition.

ianthomas23:
> I also gave you triage right on the repo, as you've been contributing a lot. Hopefully that will make your life a bit easier. You don't have to use them though. Just recognition.

Thanks!

PR #14427, opened 2024-05-01
Moving configuration to sphinx.toml file

melissawm:
This is a PR to move all _static_ docs configuration from the conf.py file to a toml file.

I did things in the simplest way possible since I figure it's no use trying to optimize this now.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,8 +15,11 @@
 # All configuration values have a default value; values that are commented out
 # serve to show the default value.
 
+import toml
 import sys, os
 from pathlib import Path
+
+config = toml.load("./sphinx.toml")
 
 # https://read-the-docs.readthedocs.io/en/latest/faq.html
 ON_RTD = os.environ.get('READTHEDOCS', None) == 'True'
@@ -37,9 +40,6 @@ if ON_RTD:
                 },
             )
 import sphinx_rtd_theme
-
-html_theme = "sphinx_rtd_theme"
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
 # Allow Python scripts to change behaviour during sphinx run
 os.environ["IN_SPHINX_RUN"] = "True"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -72,6 +72,7 @@ doc = [
     "typing_extensions",
     "exceptiongroup",
     "ipython[test]",
+    "toml",
```
bollwyvl, line 75: recommend `tomli ; python_version<'3.11'`

bollwyvl:
On the main, reducing `conf.py` complexity is great. Putting another namespacing layer (without a schema), and a another file, on top of it... not so sure.

A crazy play I've recently considered is to take DRY, declarative project definition to a silly extreme, and unilaterally declare `pyproject.toml#/tool/sphinx-j2` A Thing:

```toml
# pyproject.toml
[tool.sphinx-j2]
author = """{{ project['authors'][0]['name'] }}"""
```

...with the nickel workflow:

```py
from pathlib import Path
import json
try:
    import tomllib
except ImportError:
    import tomli as tomllib
import jinja2

HERE = Path(__file__).parent
ROOT = HERE.parent
PPT = ROOT / "pyproject.toml"
PPT_DATA = tomllib.loads(PPT.read_text(encoding="utf-8"))

globals().update(
    json.loads(
        jinja2.Template(json.dumps(PPT_DATA["tool"]["sphinx-j2"], indent=2)).render(PPT_DATA)
    )
)   

# carry on with actual dynamic things... if needed
```

A complaint can be made about _oh no, `pyproject.toml`'s getting long_... but it doesn't get shipped in `.whl` files, so I don't see much of a problem.

Carreau:
> On the main, reducing `conf.py` complexity is great. Putting another namespacing layer (without a schema), and a another file, on top of it... not so sure.

Thanks @bollwyvl, this is for another project @melissawm and I are working on. We'd love to manage to move the configurations files of many projects to be declarative so that we can more generally analyse and send automatic update. 

You can see that as the same as moving from setup.py to pyproject.toml, but for docs, and yes in the short term there will be another layer of indirection.

We also thinking of https://github.com/sphinx-toolbox/sphinx-pyproject and have seen that sphinx might also consider static configuration files.

bollwyvl:
Great! Doers get stuff done.

The sphinx issue (don't even want to link to it) is reminding me of the flake8 discussion around moving config to `pyproject.toml`.

Taking all the jinja stuff out of that stdlib-only strawman, so one _did_ have to repeat some content, I'd still rather see that optionally managed in `pyproject.toml#/tool/sphinx`, with a proper schema, and as close as possible to the existing flat data model as possible. 

```
# ... stdlib-only imports and parsing
globals().update(ppt['tool']['sphinx])
name = ppt['project']['name']
```

That third-party tool (which, while small, brings in a raft of _other_ third party tools) seems like a mighty big hammer to achieve the same thing.

The sticky DX wickets I see are: 
- path resolution
  - writing in a file in `/` but all the paths would be resolved relative to `/docs`
- the 5% left in python would have no type hints
  - but then we get all-but-no typehints in `conf.py`, so it's hardly _worse_

docs/source/conf.py
```diff
@@ -68,44 +68,125 @@
...
+copyright = config["sphinx"]["copyright"]
+# ghissue config
+github_project_url = config["sphinx"]["github_project_url"]
+# The suffix of source filenames.
+source_suffix = config["sphinx"]["source_suffix"]
+# Exclude these glob-style patterns when looking for source files. They are
+# relative to the source/ directory.
+exclude_patterns = config["sphinx"]["exclude_patterns"]
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = config["sphinx"]["pygments_style"]
```
Carreau, line 86: I'm wondering if `locals().update(config["sphinx"])` would work for most (all?) of that stuff.

docs/source/conf.py
```diff
@@ -68,44 +68,125 @@
...
+extensions = config["sphinx"]["extensions"]
+# Set the default role so we can use `foo` instead of ``foo``
+default_role = config["sphinx"]["default_role"]
+modindex_common_prefix = config["sphinx"]["modindex_common_prefix"]
+
+intersphinx_mapping = config["intersphinx_mapping"]
+for k, v in intersphinx_mapping.items():
+    intersphinx_mapping[k] = tuple(
+        [intersphinx_mapping[k]["url"], intersphinx_mapping[k]["fallback"]]
+    )
```
Carreau, line 98: Obviously here we need some workaround, and updating locals won't work.

docs/source/sphinx.toml
```diff
@@ -0,0 +1,70 @@
...
+[latex]
+latex_documents = [
+    ['index', 'ipython.tex', 'IPython Documentation',
+     'The IPython Development Team', 'manual', 'True'],
+    ['parallel/winhpc_index', 'winhpc_whitepaper.tex',
+     'Using IPython on Windows HPC Server 2008',
+     "Brian E. Granger", 'manual', 'True']
+]
+latex_use_modindex = "True"
+latex_font_size = "11pt"
```
Carreau, line 70: I completely forgot we had latex/pdf output. I think we can/should nuke it.

docs/source/conf.py
Carreau: thought on publishing (and maintaining an up to date) intersphinx_mapping python package that just contain standard urls for most of the scientific Python stack ? This has no reasons to differ between proejcts right ?

Carreau: I guess that works only if all the packages always point to latest/stable, but that woudl be a good default I think.

docs/source/sphinx.toml
Carreau: Yeah, it's not building. I think we will be able to nuke all tex-specific configs.

pyproject.toml
Carreau: So something like this ? 
```suggestion
    "toml",
    "tomli ; python_version<'3.11'"
```

bollwyvl: yeah, though with that trailing slash, and probably sorted :blush: . 

Then the matching `import tomllib/import tomli as tomllib` try/except block where it's used.

PR #14437, opened 2024-05-14
DOC: Copy-edit intro to config

wjandrea:
Grammar
Move example configs into same order


Carreau (approved):
Thanks, much appreciated.

PR #14440, opened 2024-05-22
Fix for embedded shell venv check index error

ojamin:
Fixes the issue from #14126 where we get an index out of range error when VIRTUAL_ENV is set to a 1 or 2 part path (eg ".venv").


Carreau (approved):
Thanks for lookign into this.

PR #14445, opened 2024-05-30
try to fix build on 3.13

Carreau:


Carreau:
This goes further on 3.13, so merging.

Issue #12831, opened 2021-02-23
IPython crash when use showtraceback with exception_only=True


MiklosTacs:
Used IPython version: 7.20.0-py38hd4e2768_1

The following exception was raised:
```
Traceback (most recent call last):
  File "C:\Anaconda3\lib\site-packages\IPython\core\ultratb.py", line 1103, in get_records
    return _fixed_getinnerframes(etb, number_of_lines_of_context, tb_offset)
  File "C:\Anaconda3\lib\site-packages\IPython\core\ultratb.py", line 248, in wrapped
    return f(*args, **kwargs)
  File "C:\Anaconda3\lib\site-packages\IPython\core\ultratb.py", line 281, in _fixed_getinnerframes
    records = fix_frame_records_filenames(inspect.getinnerframes(etb, context))
  File "C:\Anaconda3\lib\inspect.py", line 1503, in getinnerframes
    frameinfo = (tb.tb_frame,) + getframeinfo(tb, context)
AttributeError: 'tuple' object has no attribute 'tb_frame'
```

And when this exception is caught in IPython.core.ultratb.VerboseTB.get_records, the return value is None:

    def get_records(self, etb, number_of_lines_of_context, tb_offset):
        try:
            # Try the default getinnerframes and Alex's: Alex's fixes some
            # problems, but it generates empty tracebacks for console errors
            # (5 blanks lines) where none should be returned.
            return _fixed_getinnerframes(etb, number_of_lines_of_context, tb_offset)
        except UnicodeDecodeError:
            # This can occur if a file's encoding magic comment is wrong.
            # I can't see a way to recover without duplicating a bunch of code
            # from the stdlib traceback module. --TK
            error('\nUnicodeDecodeError while processing traceback.\n')
            return None
        **except:
            # FIXME: I've been getting many crash reports from python 2.3
            # users, traceable to inspect.py.  If I can find a small test-case
            # to reproduce this, I should either write a better workaround or
            # file a bug report against inspect (if that's the real problem).
            # So far, I haven't been able to find an isolated example to
            # reproduce the problem.
            inspect_error()
            traceback.print_exc(file=self.ostream)
            info('\nUnfortunately, your original traceback can not be constructed.\n')
            return None**

But the next operation with return value that it is called len() in IPython.core.ultratb.find_recursion, and this causes an exception and IPython crash.

MrMino:
Could you please show us the code needed to reproduce this?

MrMino:
[...]

PR #14453, opened 2024-06-04
bugfix for call to structured_traceback

smacke:
Calls to `structured_traceback` take the exploded 3-tuple of `*sys.exc_info()` as separate arguments, so we don't want to pass the 3-tuple as one argument. This should fix some of the symptoms people are seeing in https://github.com/ipython/ipython/issues/12831

Test plan: local editable install, seems to work OK. Haven't tried to repro the problem from the linked issue but this change probably can't hurt.

From https://github.com/ipython/ipython/pull/14454, which is just an empty commit, I see the same test failures as in this PR's "Run Downstream tests" job: https://github.com/ipython/ipython/actions/runs/9375565881/job/25813901627?pr=14454

So I'm guessing it's not related to the change in this PR.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -2033,7 +2033,7 @@ class InteractiveShell(SingletonConfigurable):
                     print(self.InteractiveTB.stb2text(stb))
                     print("The original exception:")
                     stb = self.InteractiveTB.structured_traceback(
-                                            (etype,value,tb), tb_offset=tb_offset
+                                            etype, value, tb, tb_offset=tb_offset
                     )
                 return stb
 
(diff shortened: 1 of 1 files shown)


jasongrout-db:
> Calls to `structured_traceback` take the exploded 3-tuple of `*sys.exc_info()` as separate arguments, so we don't want to pass the 3-tuple as one argument. This should fix some of the symptoms people are seeing in #12831

Yep, it looks like @minrk accidentally copied the arguments over from self.showtraceback [here](https://github.com/ipython/ipython/commit/09803ad149b1def32ebf12603af50b384eb56efd#diff-f0df2d6942c603793b48dc8ba2b3032e7ebedcf05a2d65883c374b5b60ef2eb0L1502), and showtraceback [takes the tuple instead.](https://github.com/ipython/ipython/blob/a3232b9df60bb5a1c10f2c58ce23e399372e9838/IPython/core/interactiveshell.py#L2115)

jasongrout (approved):

minrk (approved):

Carreau:
Let's try, there should be a release with this in about 2 weeks.

PR #14460, opened 2024-06-12
Implement `IPython.utils.io.Tee.isatty`

Atry:
A lot of libraries would call `sys.stdout.isatty()`, but `IPython.utils.io.Tee` does not implement `isatty`.

Let's add the method to avoid errors.

### DIFF
--- a/IPython/utils/io.py
+++ b/IPython/utils/io.py
@@ -75,6 +75,8 @@ class Tee(object):
         if not self._closed:
             self.close()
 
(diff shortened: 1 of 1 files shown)


Carreau:
No objections, but if you can comment a link to something that fails or give a real life example that would be great for future records.

Atry:
https://github.com/search?q=sys.stdout.isatty&type=code

PR #14468, opened 2024-06-25
MAINT: mark 3.13.dev failing test as xfail

Carreau:


Carreau:
Ok, this makes 2 test xfail, so one more test to go.

SamyDgz (approved):

PR #14475, opened 2024-06-28
Don't force matplotlib backend names to be lowercase

ianthomas23:
This is something I missed as part of moving the backend resolution to Matplotlib. Since version 3.9.0 Matplotlib handles the case-sensitivity of backend names, making them all lowercase for internal use. But since matplotlib/matplotlib#28473 uppercase letters are allowed in backend names of the form `module://soMe_moDule.sOmE_NAme`. Hence we need to not force backend names to be lowercase in IPython when passing them to Matplotlib. It is just a one line change that happens to make the code simpler, plus a new test.

The Matplotlib change will be released in 3.9.1 which is due shortly, possibly today. The new test checks the Matplotlib version to know whether uppercase backend module names are allowed or not.

### DIFF
--- a/IPython/core/magics/pylab.py
+++ b/IPython/core/magics/pylab.py
@@ -100,7 +100,7 @@ class PylabMagics(Magics):
                 % _list_matplotlib_backends_and_gui_loops()
             )
         else:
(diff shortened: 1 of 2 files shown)


Carreau:
Sorry for the delay, merging.

Issue #14464, opened 2024-06-19
Could it be worthy to make `VerboseTB._tb_highlight_style` a public attribute?


dalthviz:
Hi! Over QtConsole we are doing some work to use the changes done at https://github.com/ipython/ipython/pull/14138 to customize traceback syntax highlighting (see https://github.com/jupyter/qtconsole/pull/608).

Following https://github.com/jupyter/qtconsole/pull/608#discussion_r1638413158 and https://github.com/jupyter/qtconsole/pull/608#discussion_r1643501066, wanted to check if making `VerboseTB._tb_highlight_style` a public attribute makes sense/could be possible.

Also, are there any plans related with changing the traceback syntax highlighting logic in the future? Any in info is greatly appreciated!

Carreau:
I think I would be ok to make it public,  but in general it would be good to rethink some of the ultratb code, it could take the style as a configuration option in the constructor, and could use a lot of new CPython new features like actually use annotations, and also implement the "did you mean" when there are typoes.

In general it is also emitting directly ascii code; in a perfect world I'd like to return more structered data at at least a number of token, style pairs so that rendering could be better handled on a per-frontend basis.

Even in html we could make use of hover to see variable values; or collapse frames.

But all that needs time, and I barely have a few hours a month on IPython, so no no short term plan to change. 

If you send a PR that make this value public (with I guess a property to warn of deprecation of the private attribute I'm happy to  merge.

PR #14476, opened 2024-07-02
[PR]: Make values public (_tb_highlight & _tb_highlight_style)

jsbautista:
Fixes: 
#14464

https://github.com/ipython/ipython/issues/14464#issuecomment-2180056302

### DIFF
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -298,6 +298,13 @@ except Exception:
     raise ValueError("Yikes") from None
     """
 
+    SYS_EXIT_WITH_CONTEXT_CODE = """
+try:
+    1/0
+except Exception as e:
+    raise SystemExit(1)
+    """
+
     def test_direct_cause_error(self):
         with tt.AssertPrints(["KeyError", "NameError", "direct cause"]):
             ip.run_cell(self.DIRECT_CAUSE_ERROR_CODE)
@@ -305,6 +312,11 @@ except Exception:
     def test_exception_during_handling_error(self):
         with tt.AssertPrints(["KeyError", "NameError", "During handling"]):
             ip.run_cell(self.EXCEPTION_DURING_HANDLING_CODE)
+
+    def test_sysexit_while_handling_error(self):
+        with tt.AssertPrints(["SystemExit", "to see the full traceback"]):
+            with tt.AssertNotPrints(["another exception"], suppress=False):
+                ip.run_cell(self.SYS_EXIT_WITH_CONTEXT_CODE)
 
     def test_suppress_exception_chaining(self):
         with tt.AssertNotPrints("ZeroDivisionError"), \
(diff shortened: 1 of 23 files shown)


jsbautista:
Hi @dalthviz can you review this?

IPython/core/tests/test_ultratb.py
```diff
@@ -442,3 +442,48 @@ def eggs(f, g, z=globals()):
...
+
+# ----------------------------------------------------------------------------
+
+# module Test deprecated '_tb_highlight' & '_tb_highlight_style'
+
+
+def test_getter_deprecation_warning_tb_highlight(recwarn):
+    VerboseObj = VerboseTB()
+    value = VerboseObj._tb_highlight
+    w = recwarn.pop(DeprecationWarning)
```
dalthviz, line 455: Just in case, another syntax for this: https://docs.pytest.org/en/latest/reference/reference.html#pytest-warns

dalthviz (commented):
Thanks @jsbautista ! Left some initial suggestions, however giving this a check I thing we need to do some further changes. The thing is that I realized that the attribute of the class needs to be accessed from the class definition (without instanciating the class). The way this is working enables you to do `VerboseTB()._tb_highlight_style` but we want to be able to do `VerboseTB._tb_highlight_style` (which doesn't return a value but a `property` itself with the changes done here). Checking maybe could it be worthy to explore some things like creating a `property` subclass like:

```python
class classproperty(property):
    def __get__(self, owner_self, owner_cls):
        return self.fget(owner_cls)
``` 
https://stackoverflow.com/a/13624858

And replace `@property` for `@classproperty`. Some other alternative could be something like creating a metaclass (an idea on how to do that can be found at https://dev.to/hckjck/python-deprecation-2mof)

What do you think @ccordoba12 ?

Carreau:
Thanks for your patience, 

the problem is that those classes are Configurable, so already have a metaclass from traitlets (MetaHastrait, which inherit from MetaHasDescriptor), and already turn some of the class attribute into instance attributes. So another metaclass will be hard. 

As this is technically private and QtConsole is the only user I know if I'm ok just changing without deprecation if you like. Just le me know that QtConsole can handle both public and private. (I guess you could have a `getattr(..., 'public_name', getattr(..., '_private_name'))`, and once it's done and some time has passed, I can release a version with public name.

dalthviz:
Hi @Carreau thank you for the input! If not having to mark the attribute as deprecated makes sense then we will try to simplify the change here. Also, we already did a change as the one you mentioned over QtConsole at https://github.com/jsbautista/qtconsole/blob/7de610fdf50efe16072f34cc52482f1e77f8c481/qtconsole/mainwindow.py#L813-L823 so I would say things are covered over that side :+1:

Carreau:
Thanks for the updated branch. It's on my todo list, but I have a few travel in the next few weeks, so I may not get to it.

PR #14478, opened 2024-07-08
fix docs of custommagics

chalggg:
While trying to implement custom magic I stumbled upon error while trying to import decorators. I found that wrong module is mentioned and decided to fix that.


Carreau:
thanks.

PR #14497, opened 2024-08-15
Only copy files in startup dir if we just created it.

### LINKED ISSUE #14496: Only create the `README_STARTUP` file the first time the `startup_dir` is created
Every time I start IPython the following file is created, regardless of whether I've previously deleted it.

```
.ipython/profile_default/startup/README
```

This can be seen here as part of the `check_startup_dir()` method.

https://github.com/ipython/ipython/blob/1ec91ebf328bdf3450130de4b4604c79dc1e19d9/IPython/core/profiledir.py#L120-L121

Could we make the copying of the `readme` file into `startup_dir` conditional, only if the `startup_dir` has been created? This would mean if I delete the file (but not the dir) it won't come back again each time I start IPython.

The reason (albeit small) is because I often search for "README" in my IDE and now I have two results coming up, one of which isn't relevant to the project (the IPython one).

I'm happy to put a PR together for this if you're open to the idea?

Thanks

Carreau:
Closes #14496

### DIFF
--- a/IPython/core/profiledir.py
+++ b/IPython/core/profiledir.py
@@ -14,6 +14,8 @@ from ..paths import get_ipython_package_dir
 from ..utils.path import expand_path, ensure_dir_exists
 from traitlets import Unicode, Bool, observe
 
+from typing import Optional
+
 #-----------------------------------------------------------------------------
 # Module errors
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


IPython/core/profiledir.py
```diff
@@ -68,18 +70,31 @@ def _location_changed(self, change):
...
-    
-    def _mkdir(self, path, mode=None):
+
+    def _mkdir(self, path: str, mode: Optional[int] = None) -> bool:
         """ensure a directory exists at a given path
 
         This is a version of os.mkdir, with the following differences:
 
-        - returns True if it created the directory, False otherwise
+        - returns wether the directory has been created or not.
```
jamesbeith, line 79:
```suggestion
        - returns whether the directory has been created or not.
```

IPython/core/profiledir.py
```diff
@@ -109,16 +124,20 @@ def check_log_dir(self, change=None):
...
+            )
+
+            if not os.path.exists(src):
+                self.log.warning(
+                    "Could not copy README_STARTUP to startup dir. Source file %s does not exist.",
+                    src,
+                )
+
+            if os.path.exists(src) and not os.path.exists(readme):
+                shutil.copy(src, readme)
```
jamesbeith, line 140: Small suggestion to avoid multiple `os.path.exists()` calls.

```suggestion
            if os.path.exists(src):
                if not os.path.exists(readme):
                    shutil.copy(src, readme)
            else:
                self.log.warning(
                    "Could not copy README_STARTUP to startup dir. Source file %s does not exist.",
                    src,
                )
```

jamesbeith (approved):
Thanks.

jamesbeith (approved):

SamyDgz (approved):

PR #14500, opened 2024-08-20
Autoreload reliability improvements

### LINKED ISSUE #14395: Defining custom torch function causes "maximum recursion depth" in autoreload
Python 3.12.2
ipython 8.22.2
torch 2.2.1

I have a reproduction that is 100% reliable below.

Create a script with the following contents named `test_module.py`:

```
import torch

class CustomTorchFunc(torch.autograd.Function):
    @staticmethod
    def forward(ctx, input1, input2, weights):
        return 1

    @staticmethod
    def backward(ctx, grad_output):
        return 1


def main():
    var = 'something to edit 0'
    print("called main " + var)
```

Start ipython in the directory with the module

Execute:

```
%load_ext autoreload
%autoreload 2
import test_module
test_module.main()
```

It should print out

`called main something to edit 0`

So far so good. Now, edit `test_module.py` and increment 0 to 1. In ipython execute:

`test_module.main()`

The following error is printed out

```
[autoreload of test_module failed: Traceback (most recent call last):
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 276, in check
    superreload(m, reload, self.old_objects)
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 500, in superreload
    update_generic(old_obj, new_obj)
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 397, in update_generic
    update(a, b)
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 349, in update_class
    if update_generic(old_obj, new_obj):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 397, in update_generic
    update(a, b)
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 349, in update_class
    if update_generic(old_obj, new_obj):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 397, in update_generic
    update(a, b)
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 349, in update_class
    if update_generic(old_obj, new_obj):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 397, in update_generic
    update(a, b)
  File "/Users/will/miniforge3/envs/ml0524/lib/python3.12/site-packages/IPython/extensions/autoreload.py", line 349, in update_class
    if update_generic(old_obj, new_obj):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded
]
called main something to edit 1
```

It seems like the module was successfully reloaded, but the error says it failed. So I'm not sure if I'm in an undefined state or not.

jaysarva:
Introducing Deduperreload, an autoreload implementation that does not produce duplicate versions of objects upon reload. This work was done as part of a Summer 2024 internship at Databricks. We hope the community can also benefit from this!

The goal of this autoreload implementation is to allow for any modification to a function/method and have those changes be implemented directly without having to refresh the repl. 
Here’s our pipeline:
1. Check whether we are able to use deduperreload by analyzing the changes to the AST of each module
2. If we cannot use deduperreload, we should default back to the original autoreload. 
3. If we can use deduperreload, we do.

Note that this functionality only takes effect for CPython.

# What types of changes are supported by deduperreload?
Modifications to any function/method are supported. If some other change occurs (e.g. modification to a class member variable, or some module-level variable), we fall back to superreload. 
Furthermore, if the AST is not modified (in a meaningful way), reloading will not occur, even if the file has been modified (e.g. adding a new line, etc.)

# How does deduperreload work?

Instead of calling `reload()` to generate a new copy of the entire module (and everything in it), we intelligently find functions/methods that have been modified and just update the attributes of these existing function/method object. We also support adding/removing new functions/methods in a similar manner. 

# Why is dedupereload important?

With the original implementation of autoreload, function objects get duplicated due to directly calling the `reload()` function. This can cause many problems, especially with decorators, enums, etc. as it is possible for both the old function object and the new reloaded function object to exist at the same time. 

Fixes #14395


smacke:
cc @jasongrout 

Very very cool stuff!

smacke:
Hi Matthias, I was Jay's intern mentor for this project, for which we were fortunate to have advising from @jasongrout (will make sure to secure his +1 at some point :P ). I can help out with shepherding this PR thru as well as adding the relevant release notes to the IPython docs. Internally we're also discussing a potential blog post that goes into a few more details -- if that ends up happening, I'll be sure to link it as well. Regardless, definitely no rush on the release -- I'll follow up here around the end of September.

Carreau:
No worries. If you tell me Jason's was involved and think it's ok, I trust you. My guess is a coordinated blog post/release would be the best, so I can make a release out of the usual schedule if that helps you. I'm currently travelling but should have some time again at the end of next week.

phihung:
Hi.
I've been playing with the new reload. It works really well. Thanks a lot for the efforts.  
I found a potential issue: Sometime, `update_sources` function may take up to 10s to read 2K python files (from pandas, pytorch, transformers, etc). It happens randomly, maybe my disk was busy doing some things else at the time.

One potential fix is an option to ignore anything from /site-packages/ directory 
```python
if "/site-packages/" in fname:
    self.source_by_modname[new_modname] = ""
    continue
```

smacke:
Hi @phihung, thanks a ton for the report. I'll take a look to see what's going on and I agree that excluding /site-packages/ from deduperreload as a workaround may be the right move

smacke:
@jaysarva opened https://github.com/jaysarva/ipython/pull/1 to address @phihung's [above comment](https://github.com/ipython/ipython/pull/14500#issuecomment-2395060764)

IPython/extensions/autoreload.py
```diff
@@ -233,7 +238,7 @@ def filename_and_mtime(self, module):
 
         return py_filename, pymtime
 
-    def check(self, check_all=False, do_reload=True):
+    def check(self, check_all=False, do_reload=True, use_deduper_reload=True):
```
Carreau, line 241: Ok.

I'd like to see if there is a way to turn off use_duper_reload at runtime; just in case it breaks something maybe have the default as in instance varaible that cn be change with a parameter to the magic.

I'm og with true being the default, but scare we'd break someone without options to turn it off without changing version

IPython/extensions/autoreload.py
```diff
@@ -285,6 +297,8 @@ def check(self, check_all=False, do_reload=True):
                             file=sys.stderr,
                         )
                     self.failed[py_filename] = pymtime
+        if use_deduper_reload:
+            self.deduper_reloader.update_sources()
```
Carreau, line 301: ok. (I put ok in there for myself to know how far I've read).

IPython/extensions/deduperreload/deduperreload.py
```diff
@@ -0,0 +1,477 @@
...
+
+if TYPE_CHECKING:
+    from IPython.core.interactiveshell import ExecutionResult
+
+    TDefinitionAst = ast.FunctionDef | ast.AsyncFunctionDef
+
+DefinitionAst = (ast.FunctionDef, ast.AsyncFunctionDef)
+
+
+def get_module_file_name(module: ModuleType | str) -> str:
```
Carreau, line 22: As a note (not a request to change). Those `|` fails on older version of Python when using runtime types extraction of type (https://docs.python.org/3/library/typing.html#typing.get_type_hints). Som in some case  `from __future__ import annotations` is not sufficient to use the new annotation syntax.

IPython/extensions/deduperreload/deduperreload.py
```diff
@@ -0,0 +1,477 @@
...
+    children: classes inside given scope, maps class name to autoreload tree for that class's scope
+    funcs_to_autoreload: list of function names that can be autoreloaded in given scope.
+    new_nested_classes: Classes getting added in new autoreload cycle
+    """
+
+    def __init__(self) -> None:
+        self.children: dict[str, AutoreloadTree] = {}
+        self.defs_to_reload: dict[str, ast.AST] = {}
+        self.defs_to_delete: set[str] = set()
+        self.new_nested_classes: dict[str, ast.AST] = {}
```
Carreau, line 89: I think usually you want to the type annotation at class level for proper type checking in subclasses I believe; it used to be that mypy was not able to infer type annotation of attribute defined in __init__. (or emit a `AutoreloadTree.children` attribute might not be set/ None does not have method XXX).

I don't think that's a big problem here, but something to keep in mind depending on what

IPython/extensions/deduperreload/deduperreload.py
```diff
@@ -0,0 +1,477 @@
...
+        return cur
+
+
+class DeduperReloader(DeduperReloaderPatchingMixin):
+    """
+    This version of autoreload detects when we can leverage targeted recompilation of a subset of a module and patching existing function/method objects to reflect these changes.
+
+    Detects what functions/methods can be reloaded by recursively comparing the old/new AST of module-level classes, module-level classes' methods, recursing through nested classes' methods.
+    If other changes are made, original autoreload algorithm is called directly.
+    """
```
Carreau, line 109: Maybe wrap at 80 chars wide.

IPython/extensions/deduperreload/deduperreload.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def detect_autoreload(
+        self,
+        old_node: ast.Module | ast.ClassDef,
+        new_node: ast.Module | ast.ClassDef,
+        prefixes: list[str] | None = None,
+    ) -> bool:
+        """
+        Returns `true` if we can run our targeted autoreload algorithm safely.
+        Returns `false` if we should instead use IPython's original autoreload implementation.
```
Carreau, line 215:
```suggestion
        Returns
        -------
        `True` if we can run our targeted autoreload algorithm safely.
        `False` if we should instead use IPython's original autoreload implementation.
```

We try to follow [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) standard when we can.

IPython/extensions/deduperreload/deduperreload.py
```diff
@@ -0,0 +1,477 @@
...
+                )
+        return True
+
+    def build_dependency_graph(self, new_ast: ast.Module | ast.ClassDef) -> bool:
+        """
+        Wrapper function for generating dependency graph given some AST.
+        Returns `true` on success. Returns `false` on failure.
+        Currently, only returns `true` as we do not block on failure to build this graph.
+        """
+        return self._gather_dependents(new_ast.body)
```
Carreau, line 477: Ok.

Carreau (commented):
Thanks for your patience I'm slowly reading through this. I will do a release soon, and might not have the time to fully review. I've dropped some comments for myself when I come back to it.

smacke:
> Thanks for your patience I'm slowly reading through this. I will do a release soon, and might not have the time to fully review. I've dropped some comments for myself when I come back to it.

Hey @Carreau really appreciate your comments! No worries on turnaround time, I similarly have been swamped and **finally** managed to scrounge some time to take a look at them.

From what I can see, the main request is that we provide a way to disable the deduperreload algo. The way I've implemented this is that it's enabled by default, but if we want to turn it off, append `-` to the `%autoreload` invocation, e.g. `%autoreload 2-`. The reason for this is that deduperreload is orthogonal to the various autoreload "levels" and could be either enabled or disabled with each of them. I've updated the docstring in the argument parser to cover this; would love to get your feedback on that approach whenever you have time :)

IPython/extensions/deduperreload/deduperreload.py
smacke: I believe that newer versions of mypy handle this properly now (though I could be wrong)

Carreau:
Ok, main is on ti's way to 9.0 let's emrge.

Issue #14350, opened 2024-02-20
magic conda fails in windows (re can not parse path)


gnosys-scrawford:
In a jupyter notebook when trying to run conda or mamba as a magic command, in Windows, it fails.  For example  
`%conda --version
`

The problem is that in IPython\core\magics\packaging.py, in `_get_conda_like_executable` there is
```
    match = re.search(
        rf"^#\s*cmd:\s*(?P<command>.*{executable})\s[create|install]",
        history,
        flags=re.MULTILINE,
    )
```
where `executable` is a `Path` object.  In windows, this gets converted to a string with backslashes.  `re` then tries to treat these backslashes as escape characters and fails. For example, when executable is 'c:\Users\me\mambaforge\envs\myenv\conda' the error will be `error: incomplete escape \U at position 28`

I believe the fix is to use `executable.name` instead of `executable` in the search string.  This will emulate previous versions but I don't know if there are other considerations.

ivanov:
Thanks for your report, @gnosys-scrawford . I think doing `executable.name` would work only in some cases, and it would not work when executable located somewhere not in the PATH directories (or if there are mutliple executables with the same name in different directories on PATH, and the executable desired isn't the first one). 

By the time the code you highlighted is executing, we're in a path that's commented as "Otherwise, attempt to extract the executable from conda history. " 

https://github.com/ipython/ipython/blob/6046a5f941e54dc1a3074550fe2848e074e32217/IPython/core/magics/packaging.py#L46-L57

Can you try instead of `{executable}` to do `{str(executable)}`? I think this would get around the escape issue. 

The [Python docs say](https://docs.python.org/3/library/pathlib.html#operators)

> The string representation of a path is the raw filesystem path itself (in native form, e.g. with backslashes under Windows), which you can pass to any function taking a file path as a string:
```python
>>> p = PurePath('/etc')
>>> str(p)
'/etc'
>>> p = PureWindowsPath('c:/Program Files')
>>> str(p)
'c:\\Program Files'
```

gnosys-scrawford:
Thanks for the response,

My suggestion was based primarily that before the latest commit, the search would take place just on the literal string "conda" (no absolute path), and the latest commit comment was just "Implement magic for mamba and micromamba" so switching to `executable.name` (actually, switching to `command` would be better) would achieve that specific goal of allowing "conda" or "mamba" or "micromamba".  I could be wrong, but I don't think that including the path to conda when searching the environment history was intended as a bug fix.

[...]

Issue #14505, opened 2024-08-26
Downstream tests failing - sagemath-repl


takluyver:
The 'Run Downstream tests' workflow is failing on sagemath. It *appears* that it succeeded a few weeks ago, but looking at the [logs from that run](https://github.com/ipython/ipython/actions/runs/10290708602/job/28481296123), it actually shows a bunch of failures - `sage-runtests` must have exited with status 0 despite the errors, making the CI think it passed (it looks like sage has a custom mechanism for known failures - maybe this is involved, maybe not).

I couldn't find a run where it definitely passed - logs for older runs are deleted.

Given that consistently failing CI isn't useful, is it worth disabling this until someone figures out how to make it work?

takluyver:
Fixed by #14519

PR #14515, opened 2024-09-12
Use environment variable to identify conda / mamba

scw:
Conda and mamba both set an environment variable which refers to the base environment's executable path, use that in preference to less reliable methods, but fall back on the other approaches if unable to locate the executable this way.

Additionally, change the search to look for the bare command name rather than the command within the top level of the active environment, I'm dubious this approach works with any current conda / mamba version which usually place their executable links in a `condabin` directory or elsewhere not at the same level as the Python executable.

I believe this will also address https://github.com/ipython/ipython/issues/14350, which I'm also seeing in a Windows context where the regex fails to parse and causes a traceback.

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -9,6 +9,7 @@
 #-----------------------------------------------------------------------------
 
 import functools
+import os
 import re
 import shlex
 import sys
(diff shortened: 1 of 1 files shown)


scw:
I believe the test failures from this are related to https://github.com/ipython/ipython/issues/14505 rather than something specific to this code change

Carreau:
Seam reasonable, thanks.

PR #14517, opened 2024-09-14
bugfix for oinspect.object_info

smacke:
`oinspect.object_info` loses stuff passed in as kwargs. I suspect it's not intentional (previous versions of `object_info` don't do this) and it seems to be the root cause of some CI failures when trying out new versions of IPython at Databricks. In this PR we preserve the fields passed in as kwargs.

Test plan: let pre-merge tests run.

### DIFF
--- a/IPython/core/oinspect.py
+++ b/IPython/core/oinspect.py
@@ -173,8 +173,8 @@ def object_info(
     **kw,
 ) -> InfoDict:
     """Make an object info dict with all fields present."""
-    infodict = kw
-    infodict = {k: None for k in _info_fields if k not in infodict}
+    infodict = dict(kw)
+    infodict.update({k: None for k in _info_fields if k not in infodict})
     infodict["name"] = name  # type: ignore
     infodict["found"] = found  # type: ignore
     infodict["isclass"] = isclass  # type: ignore


jasongrout-db:
For reference:

The change that overwrote the object info arguments is [here](https://github.com/ipython/ipython/pull/14349/files#diff-f0ad9a3d464ad8cc22bd87cc63abad1ae625ac9c3f8c56b0d639890eee720b3bL149-L151) (and seems to be have a typo - it doesn't make sense to set the object info to `kw`, then immediately overwrite it with a new dict), and references this [proposed change](https://github.com/ipython/ipython/pull/14342/files#diff-f0ad9a3d464ad8cc22bd87cc63abad1ae625ac9c3f8c56b0d639890eee720b3bL123), which does preserve the kw passed into the function.

Looking at these two changes above, plus @smacke's PR, I'm +1 for merging this.

IPython/core/oinspect.py
```diff
@@ -175,8 +175,9 @@ def object_info(
     **kw,
 ) -> InfoDict:
     """Make an object info dict with all fields present."""
-    infodict = kw
-    infodict = {k: None for k in _info_fields if k not in infodict}
+    infodict = {}
+    infodict.update(kw)
```
jasongrout-db, line 179: Thoughts on making this a bit more succinct to make a copy of kw?
```suggestion
    infodict = dict(kw)
```

jasongrout-db (approved):
LGTM

IPython/core/oinspect.py
smacke: SGTM!

PR #14537, opened 2024-10-11
Add uv line magic

akumor:
This adds the ability to install packages using `%uv install [pkgs]` line magic.

### DIFF
--- a/IPython/core/magics/packaging.py
+++ b/IPython/core/magics/packaging.py
@@ -162,3 +162,20 @@ class PackagingMagics(Magics):
         """
         micromamba = _get_conda_like_executable("micromamba")
         self._run_command(micromamba, line)
+
+    @line_magic
+    def uv(self, line):
+        """Run the uv package manager within the current kernel.
+
+        Usage:
+          %uv pip install [pkgs]
+        """
+        python = sys.executable
+        if sys.platform == "win32":
+            python = '"' + python + '"'
+        else:
+            python = shlex.quote(python)
+
+        self.shell.system(" ".join([python, "-m", "uv", line]))
+
+        print("Note: you may need to restart the kernel to use updated packages.")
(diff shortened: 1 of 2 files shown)


dennisobrien:
I think the motivation here is that `!uv install` will usually do the wrong thing and install into a python environment other than the one running the current kernel.  (In the same way that `!pip install` and `!conda install` will also do the wrong thing.)  From [this classic post from Jake VDP](https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/)

> in Jupyter, the shell environment and the Python executable are disconnected.

(I'm sure you know this much better than I do, I just want to include some motivation that is missing from the PR description.)

I can see the counter-argument here: should the ipython project be responsible for including magics for every python package/environment management module?  That's definitely a question for the ipython maintainers.  I don't know what the adoption rates for `uv` are, but our team has been really impressed with the speed and functionality.  And it seems like `uv` is quickly getting traction.

Carreau:
> I think the motivation here is that `!uv install` will usually do the wrong thing and install into a python environment other than the one running the current kernel.

Yes, but for pip I do take special care of actually calling `sys.executable -m pip`, while here you are using shutil.which, and if uv is not in current env, it will get the potentially system global, so this is identical to `!uv`, I guess least maybe do `sys.executable -m uv` (I think it should work, but I haven't use uv enough, especially the multi-env).

Unless I misunderstand how shutil.which works, or missed something else in the PR. (Which is honestly  quite likely I'm just getting out of a few days sick and still brainfoggy).

akumor:
Thank you for the feedback!  I will work on updating the PR to use `sys.executable -m uv`.  I knew you could install uv with pip but I did not realize `python -m uv` would work.

Carreau:
I'm still not 100% convinced it is needed, but let's try it.

Carreau:
Thanks.

PR #14543, opened 2024-10-17
Refactor with Pathlib for setupbase.py

kevin1kevin1k:
Continued efforts for #12515 though it has been closed.

Validated to be okay by running the following.
- `pip install -e ".[all]"`
- `pytest`

Please note that there are also some syntactical modifications, and part of them are from the darker linter.

### DIFF
--- a/setupbase.py
+++ b/setupbase.py
@@ -65,7 +65,7 @@ def check_package_data(package_data):
         for d in data:
             path = pkg_root / d
             if "*" in str(path):
-                assert len(glob(path)) > 0, "No files match pattern %s" % path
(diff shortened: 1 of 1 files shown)


kevin1kevin1k:
Now I see some failing checks. Fixing them.

kevin1kevin1k:
The remaining failing check `test_run_concurrently_sequence` seems to be irrelevant to this change

kevin1kevin1k:
@Carreau Please have a look, thanks. Also let me know if I should first squash the commits into one or it's okay to leave them as is.

Carreau:
Thanks, it's ok to leave as is. I rebased as test were failing for a unrelated reason.

PR #14550, opened 2024-10-20
Remove broken link in tutorial

LiamConnors:
This link seems to be broken


Carreau (approved):
Indeed, thanks for taking the time.

PR #14553, opened 2024-10-20
Fixes newline issue in sphinx docstring build

q-arya-a:
Added a newline under the %%timeit magic command in cell mode in the built-in magic commands section of the tutorial in the documentation


Carreau (approved):
Good catch !

PR #14558, opened 2024-10-24
Fix re pattern pretty print

wimglenn:
`re.TEMPLATE` is gone in Python 3.13 resulting in a crash in the pattern pretty printer

Reproducer:

```
>>> import re
>>> re.compile("")
```

<details>
  <summary>Click me to see the traceback!</summary>

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/.local/lib/python3.13/site-packages/IPython/core/formatters.py:711, in PlainTextFormatter.__call__(self, obj)
    704 stream = StringIO()
    705 printer = pretty.RepresentationPrinter(stream, self.verbose,
    706     self.max_width, self.newline,
    707     max_seq_length=self.max_seq_length,
    708     singleton_pprinters=self.singleton_printers,
    709     type_pprinters=self.type_printers,
    710     deferred_pprinters=self.deferred_printers)
--> 711 printer.pretty(obj)
    712 printer.flush()
    713 return stream.getvalue()

File ~/.local/lib/python3.13/site-packages/IPython/lib/pretty.py:394, in RepresentationPrinter.pretty(self, obj)
    391 for cls in _get_mro(obj_class):
    392     if cls in self.type_pprinters:
    393         # printer registered in self.type_pprinters
--> 394         return self.type_pprinters[cls](obj, self, cycle)
    395     else:
    396         # deferred printer
    397         printer = self._in_deferred_types(cls)

File ~/.local/lib/python3.13/site-packages/IPython/lib/pretty.py:740, in _re_pattern_pprint(obj, p, cycle)
    738 re_compile = CallExpression.factory('re.compile')
    739 if obj.flags:
--> 740     p.pretty(re_compile(RawStringLiteral(obj.pattern), _ReFlags(obj.flags)))
    741 else:
    742     p.pretty(re_compile(RawStringLiteral(obj.pattern)))

File ~/.local/lib/python3.13/site-packages/IPython/lib/pretty.py:408, in RepresentationPrinter.pretty(self, obj)
    406     meth = cls._repr_pretty_
    407     if callable(meth):
--> 408         return meth(obj, self, cycle)
    409 if (
    410     cls is not object
    411     # check if cls defines __repr__
   (...)
    417     and callable(_safe_getattr(cls, "__repr__", None))
    418 ):
    419     return _repr_pprint(obj, self, cycle)

[...]

PR #14559, opened 2024-10-24
Fix re pattern pretty print

wimglenn:
Alternative to #14558


Carreau:
Thanks, I prefer this one.

The other one contains code that is specifically to be compatible between multiple versions, but it's hard to know later on the `, 0` and `TEMPLATE` could be removed once we drop 3.12, so I leaning toward hanging formatting, worse case we revert.

PR #14578, opened 2024-11-11
crashhandler: Cast rptdir to Path

user202729:
Previously `self.app.ipython_dir` is a string, which makes `Path.is_dir(<object>)` fail. Note that the argument is `self` i.e. should be a `Path` instance.

If `ipython_dir` is `None` then this code would just catch it in the `except:` branch, which makes it be set to `Path.cwd()` anyway.

### DIFF
--- a/IPython/core/crashhandler.py
+++ b/IPython/core/crashhandler.py
@@ -157,10 +157,10 @@ class CrashHandler(object):
         # Use this ONLY for developer debugging (keep commented out for release)
         #color_scheme = 'Linux'   # dbg
         try:
-            rptdir = self.app.ipython_dir
+            rptdir = Path(self.app.ipython_dir)
         except:
             rptdir = Path.cwd()
-        if rptdir is None or not Path.is_dir(rptdir):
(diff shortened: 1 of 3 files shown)


IPython/core/crashhandler.py
```diff
@@ -155,21 +163,23 @@ def __call__(self, etype, evalue, etb):
         color_scheme = 'NoColor'
 
         # Use this ONLY for developer debugging (keep commented out for release)
-        #color_scheme = 'Linux'   # dbg
-        try:
-            rptdir = self.app.ipython_dir
-        except:
+        # color_scheme = 'Linux'   # dbg
+        ipython_dir = getattr(self.app, "ipython_dir")
```
user202729, line 167: getattr will raise AttributeError when property doesn't exist. Don't you want to provide None as the last argument?

@Carreau

Carreau:
Thanks for finding this error, I pushed more commits.

I realized that the file was ignored by static checking and did some cleanup as well as using a conditional instead of try/except.

Issue #14585, opened 2024-11-21
IPCompleter doesn't work inside tuples/arrays


divyansshhh:
Consider the following snippet - 

```python
y = "abc"
x = (1, y.<tab>
```

This doesn't return any autocomplete suggestions. This is fixed by enabling `jedi` but can we look into fixing this for the IPCompleter?

This is reproducible on both ipython v8.22.0 and v8.28.0

krassowski:
The case of unclosed parenthesis is already handled, but this case is not. Briefly the relevant code is:

https://github.com/ipython/ipython/blob/d7eebae444b6cf5e8d184591766cf64764e31773/IPython/core/completer.py#L1144-L1150

and then in evaluation we use naive approach of trimming on the left side if there is a syntax error - unless there is not.

https://github.com/ipython/ipython/blob/d7eebae444b6cf5e8d184591766cf64764e31773/IPython/core/completer.py#L1221-L1228

I think that the problem might be that `1, y` is a valid Python expression which evaluates to a tuple (assuming that `y` is defined).

Partial parsing of Python is not easy in the general case. We could possibly consider some heuristics for closing parenthesis for tuples and lists and after guarded evaluation extract the relevant part.

To spell it out we cannot just split on space as in general case it could be `x = (1, a[" "].y.`

krassowski:
Indeed, `<tab>` returns results corresponding to methods available on a tuple:

![Image](https://github.com/user-attachments/assets/70b24a69-5086-4d6d-89f5-336b2ce8240d)

![Image](https://github.com/user-attachments/assets/168b83ee-709b-42fb-b91c-2ff7cdbb7660)

krassowski:
Some more related cases:
- `x = (1, a[" "].y.`
- `x = [1, y.`
- `x = {1, y.`

```
x = (
   1, y.
```

mlucool:
FWIW we turned off jedi because it had historically been too slow, so using Jedi as a solution is unlikely to be ideal (unless its much faster than it used to be)

mlucool:
The following is also a fix
```
ip = get_ipython()
ip.Completer.python_matcher = ip.Completer.python_matches
```

Carreau:
> FWIW we turned off jedi because it had historically been too slow, so using Jedi as a solution is unlikely to be ideal (unless its much faster than it used to be)

[...]

PR #14594, opened 2024-12-02
Fix completion tuple

Carreau:
In progress work toward #14585

guarded eval strip leading characters until it find soemthing, this is problematic as `(1, x`, becomes valid after 1 char strip:  `1, x` is a tuple; 

So now we trim until it is valid an not a tuple. 

This is still imperfect as things like `(1, a[" "].y` will be trimmed to `y`, while it should stop with `a[" "].y` ?

I think maybe we should back-propagate; build back up from `y`, to `a[" "].y`, greedily until we get the last valid expression – skipping any unbalanced parentheses/quotes if we encounter imblanced.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -348,7 +348,7 @@ def provisionalcompleter(action='ignore'):
         yield
 
 
-def has_open_quotes(s):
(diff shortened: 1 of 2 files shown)


Carreau:
cc @krassowski  to not duplicate work.

PR #14598, opened 2024-12-08
Fix pdb issues in Python 3.13.1

Carreau:
For some reason it is not always set, it was/is a bug in IPython to not check.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -550,7 +550,7 @@ class Pdb(OldPdb):
         So if frame is self.current_frame we instead return self.curframe_locals
 
         """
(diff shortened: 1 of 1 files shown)


gabifalk:
I believe, even with the fix it won't work with python 3.14.  See:
https://github.com/python/cpython/pull/124369 and
https://github.com/python/cpython/commit/986a4e1b6fcae7fe7a1d0a26aea446107dd58dd2

gabifalk:
Also this one, https://github.com/python/cpython/pull/125951.  Perhaps we should just wait. :)

Carreau:
Ah thanks for the pointer; Ii have opened an issue.

PR #14629, opened 2024-12-22
Python vs IPython: recommend %cd, %env over !cd, !export

cben:
https://ipython.readthedocs.io/en/latest/interactive/python-ipython-diff.html#shell-assignment
The quick overview mentioned `!cd DIR` among examples of shell commands one might use — but that doesn't work as user would hope (can only affect commands in same ! e.g. `!cd /; pwd`).

That's not as bad now that `!cd` prints a warning, but IMHO better to provide the right mental model up front.

=> Replaced that with minimal motivation why `!cd` and `!export` can't work, recommending some %magic alternatives;  
moved it out of examples to not distract, and right before Magics section.

### DIFF
--- a/docs/source/interactive/python-ipython-diff.rst
+++ b/docs/source/interactive/python-ipython-diff.rst
@@ -152,13 +152,7 @@ This allows to execute simple commands when present in beginning of the line:
     In[1]: !pwd
     /User/home/
 
-Change directory:
-
-.. code-block:: ipython
-
-    In[1]: !cd /var/etc
-
-Or edit file:
+Edit file:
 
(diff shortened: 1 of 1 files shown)


cben:
Should I include an example demonstrating it works?
```
In [1]: %cd /
/

In [2]: !pwd
/
```
imho it'd add little value, their syntaxes are similar enough to be used intuitively (and each has extensive help), just hope users grasp the mental model — and the quick segue from shell commands to python-level magics is more valuable.

Carreau (approved):
Yeah, you are right. Thanks, nice catch.

Issue #14630, opened 2024-12-23
inspect token_at_cursor improvements


jasongrout:
The [token_at_cursor](https://github.com/ipython/ipython/blob/main/IPython/utils/tokenutil.py#L92) method is used to determine what help should be shown to the user for a given cursor position. In JupyterLab, the help can be updated in real time, performing a similar function to the hover popup in other editors like vs code, etc. Right now, the function explicitly prioritizes returning help about a function when the cursor is inside the parens of a call to the function, rather than returning help about the token the cursor is on (see https://github.com/ipython/ipython/pull/7819). At Databricks, we've recently had some discussion about how it could be better. 

For example, using the utility function below so we can make some test cases, where the cursor position is noted with a `|`:

```python
from IPython.utils.tokenutil import token_at_cursor
def cursor(code):
    idx = code.index('|')
    return (code.replace('|',''), idx)
```

```python
# helpful
assert token_at_cursor(*cursor('foo(bar(x, |))')) == 'bar'

# should be 'bar'
assert token_at_cursor(*cursor('foo(ba|r(x, y))')) == 'foo'

# should be foo?
assert token_at_cursor(*cursor('foo(bar(x, y))|')) == 'y'
```

CC @smacke, who pointed out some of the issues above

PR #14631, opened 2024-12-24
more intuitive behavior for token_at_cursor for inspect requests

smacke:
This PR implements the suggested behavior in https://github.com/ipython/ipython/issues/14630 -- i.e., if the cursor is directly over a non-keyword token inside of parameter list, use that as the token_at_cursor rather than the function. Otherwise, if we're not directly over a token and we're inside some function call, use the most recently seen function call name token. If we're not inside a function call, use the most recently-seen outer name token.

This should yield more intuitive behavior for inspect, and is documented in comments and tests.

### DIFF
--- a/IPython/utils/tests/test_tokenutil.py
+++ b/IPython/utils/tests/test_tokenutil.py
@@ -15,6 +15,7 @@ def expect_token(expected: str, cell: str, cursor_pos: int | None = None) -> Non
     Assert that the token at the cursor position is `expected`.
     """
     if cursor_pos is None:
+        assert cell.count("|") == 1, "Cursor position not specified and no | found"
         cursor_pos = cell.index("|")
         cell = cell.replace("|", "")
     token = token_at_cursor(cell, cursor_pos)
--- a/IPython/utils/tests/test_tokenutil.py
+++ b/IPython/utils/tests/test_tokenutil.py
@@ -15,7 +15,11 @@ def expect_token(expected: str, cell: str, cursor_pos: int | None = None) -> Non
     Assert that the token at the cursor position is `expected`.
     """
     if cursor_pos is None:
-        assert cell.count("|") == 1, "Cursor position not specified and no | found"
+        assert (
+            cursor_count := cell.count("|")
+        ) == 1, (
+            f"Cursor position not specified and found {cursor_count} instance(s) of '|'"
+        )
         cursor_pos = cell.index("|")
         cell = cell.replace("|", "")
     token = token_at_cursor(cell, cursor_pos)
--- a/IPython/utils/tests/test_tokenutil.py
+++ b/IPython/utils/tests/test_tokenutil.py
@@ -1,8 +1,10 @@
 """Tests for tokenutil"""
+
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
 import pytest
+import textwrap
 
 from IPython.utils.tokenutil import token_at_cursor, line_at_cursor
 
@@ -53,16 +55,17 @@ def test_function():
     "cell",
     [
         "\n".join(["a = 5", "b = hello('string', there)"]),
-        "\n".join(
-            [
-                '"""\n\nxxxxxxxxxx\n\n"""',
-                '5, """',
-                "docstring",
-                "multiline token",
-                '""", [',
-                '2, 3, "complicated"]',
-                'b = hello("string", there)',
-            ]
+        textwrap.dedent(
+            '''
+            """\n\nxxxxxxxxxx\n\n"""
+            5, """
+            """\n\nxxxxxxxxxx\n\n"""
+            docstring
+            multiline token
+            """, [
+            2, 3, "complicated"]
+            b = hello("string", there)
+        '''
         ),
     ],
 )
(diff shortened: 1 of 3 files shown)


Carreau:
Thanks to you and @jasongrout; I am off for a few weeks; but will look at it when I'm back. 

This seem reasonable to me; on the top of my head; would/could/should the API support cursor range – even if at first pass we only allow start == end;  so that later we may have inspect request for selection ? 

I'm thinking likely if we improve the API; we also want to return the range of the token we found, so that if it is not exactly at cursor position; the UI can display what we are talking about.

Anyway, just food for thoughts, as we would need to think of a migration path. 

We can resume the discussions when I am back; enjoy the holidays, and happy new year.

IPython/utils/tokenutil.py
```diff
@@ -2,20 +2,29 @@
...
+    line: str
 
-def generate_tokens(readline):
+
+def generate_tokens(text_or_readline) -> Generator[TokenInfo, None, None]:
     """wrap generate_tkens to catch EOF errors"""
+    if isinstance(text_or_readline, str):
+        readline = StringIO(text_or_readline).readline
+    else:
+        readline = text_or_readline
```
Carreau, line 27: I have a few hours in a train so I'll do a cursory review. 

The comments bellow are not a request to change (at least for now), I'm laying down my thoughts, and will come back to it. 

I would prefer to not do that; I'm in general not a big fan of function that takes multiple types like this as their input; It's also meant to mimick tokenize.generate_token, which cannot take a str. 

And yes, I hate that it can't take str; I get annoyed at it everytime. 

Also _technically_ renaming the parameter is a breakage of API as technically a positional parameter can be passed as a named argument.

We could audit all places that uses generate_tokens and move to str only as well; I woudl not be against.

IPython/utils/tokenutil.py
```diff
@@ -2,20 +2,29 @@
...
+from typing import Generator, NamedTuple
 
 
-Token = namedtuple('Token', ['token', 'text', 'start', 'end', 'line'])
+class Token(NamedTuple):
+    token: int
+    text: str
+    start: int
+    end: int
+    line: str
```
Carreau, line 19: :+1:

IPython/utils/tokenutil.py
```diff
@@ -25,7 +34,7 @@ def generate_tokens(readline):
 
 
 def generate_tokens_catch_errors(
-    readline, extra_errors_to_catch: Optional[List[str]] = None
+    readline, extra_errors_to_catch: list[str] | None = None
```
Carreau, line 37: I think it's fine here, but be careful, `from __future__ import annotations` is not always enough to use `|` instead of `Optional`, and things like `dict[]`, `list[]`, as runtime [`get_type_hints`](https://docs.python.org/3/library/typing.html#typing.get_type_hints) may fail.

IPython/utils/tokenutil.py
```diff
@@ -104,15 +113,19 @@ def token_at_cursor(cell: str, cursor_pos: int = 0):
...
+    most_recent_outer_name: str | None = None
+
+    offsets = {1: 0}  # lines start at 1
+    intersects_with_cursor = False
+    cur_token_is_name = False
+    tokens: list[Token] = [Token(*tup) for tup in generate_tokens(cell)]
+    if not tokens:
+        return ""
+    next_tokens: list[Token | None] = tokens[1:] + [None]
+    for tok, next_tok in zip(tokens, next_tokens):
```
Carreau, line 128: https://docs.python.org/3/library/itertools.html#itertools.pairwise ?

IPython/utils/tests/test_tokenutil.py
```diff
@@ -6,14 +6,22 @@
...
 
-def expect_token(expected, cell, cursor_pos):
+
+def expect_token(expected: str, cell: str, cursor_pos: int | None = None) -> None:
+    """
+    If cursor_pos is `None`, look for `|` as the cursor position.
+    Assert that the token at the cursor position is `expected`.
+    """
+    if cursor_pos is None:
+        cursor_pos = cell.index("|")
```
Carreau, line 16: maybe an `assert cell.count('|') == 1` ?

IPython/utils/tests/test_tokenutil.py
```diff
@@ -29,65 +37,91 @@ def test_simple():
...
+            [
+                '"""\n\nxxxxxxxxxx\n\n"""',
+                '5, """',
+                "docstring",
+                "multiline token",
+                '""", [',
+                '2, 3, "complicated"]',
+                'b = hello("string", there)',
+            ]
+        ),
```
Carreau, line 66: I'm fine with `\n.join`; have you thought on textwrap.dedent for readability ?

IPython/utils/tests/test_tokenutil.py
```diff
@@ -29,65 +37,91 @@ def test_simple():
...
+    expect_token("foo", "x + foo(a|, b) + y + z")
+    expect_token("foo", "x + foo(a, b)| + y + z")
+    expect_token("foo", "x + foo(a, b) |+ y + z")
+    expect_token("foo", "x + foo(a, b) +| y + z")
+    expect_token("y", "x + foo(a, b) + |y + z")
+    expect_token("y", "x + foo(a, b) + y| + z")
+    expect_token("y", "x + foo(a, b) + y |+ z")
+    expect_token("y", "x + foo(a, b) + y +| z")
+    expect_token("z", "x + foo(a, b) + y + |z")
+
```
Carreau, line 124: probably coudl make those @pytest.mark.parametrize as well.

Carreau (commented):
A small review; but not easy while train is shaking; so takeit with a grain of salt

IPython/utils/tokenutil.py
smacke: I'll leave as the original

IPython/utils/tests/test_tokenutil.py
smacke: This is copied over from another test that had the same assertions but that's a good idea

IPython/utils/tokenutil.py
smacke: This should be fine since ipython requires python >= 3.11 in this branch, right?

IPython/utils/tests/test_tokenutil.py
smacke: True, but I think it's more readable as-is.

taras-bl (approved):

IPython/utils/tests/test_tokenutil.py
Carreau: It's more in case of failure you see only (and all), the failing one; but fair enough.

IPython/utils/tokenutil.py
Carreau: Yeah, here it's fine. with all the project I'm in I often don't remember for which one it's relevant.

Carreau (approved):
Ok, let's try.

PR #14640, opened 2025-01-07
Fixed bug with a warning message incorrectly displayed

alberti42:
Fixed an issue where the warning message:

> Attempting to work in a virtualenv. If you encounter problems please install IPython inside the virtualenv.

was incorrectly triggered in systems involving symlinks.

When the path provided by `$VIRTUAL_ENV` and the corresponding environment path derived from the executable iPython pointed to the same physical directory, but their strings differed because of symlinks, the warning was incorrectly triggered. By using the `resolve()` function, we ensure that we compare fully resolved paths. The warning disappears when the two folders, once resolved, coincide.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -898,7 +898,7 @@ class InteractiveShell(SingletonConfigurable):
             return
 
         p = Path(sys.executable)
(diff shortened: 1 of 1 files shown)


Carreau:
let's try. I might do a 8.32 with that. Not sure.

Issue #14545, opened 2024-10-18
Recent downstream tests for ipykernel fails


kevin1kevin1k:
In recent PRs starting from #14537 last week, `Run Downstream tests` fails on `test_run_concurrently_sequence` of `ipykernel`.
See https://github.com/ipython/ipython/actions/runs/11390216809/job/31691312661?pr=14543 as an example.
The error seems to be independent of the PRs.
Not sure if it was anything in this repo that caused it, or was it due to something in https://github.com/ipython/ipykernel and maybe I should also file an issue there?

ianthomas23:
Hi @kevin1kevin1k, I'm the original author of those tests in `ipykernel`. They are somewhat risky tests as they rely on timing of actions on different threads and they are sensitive to workload on shared runners in CI, and whether coverage information is being recorded, and so on. Probably I need to modify the tests so that they are more robust to the runtime environment.

I don't think there are any changes in `ipython` that are at fault here, so if any of your PRs here are failing CI because of `test_run_concurrently_sequence` only then perhaps mention that in a comment linking to this issue and it shouldn't stop your PRs from being reviewed and accepted.

A separate issue in `ipykernel` would be useful. I will get round to that eventually but if you are inclined to do so earlier than that then please go ahead.

ianthomas23:
I note that the downstream failures here are consistently on `macos-13` runners, i.e. Intel Mac. The `ipykernel` tests run on `macos-latest == macos-14` runners which are Arm Mac.

kevin1kevin1k:
Hey @ianthomas23 thanks for the kind and detailed explanations.

> A separate issue in ipykernel would be useful. I will get round to that eventually but if you are inclined to do so earlier than that then please go ahead.

As you are the author of the tests I guess it would be better for you to file the issue and describe the situation.

> I note that the downstream failures here are consistently on `macos-13` runners, i.e. Intel Mac. The `ipykernel` tests run on `macos-latest == macos-14` runners which are Arm Mac.

That's a nice catch!
I am not quite sure which of the following works better:
- Switch the test of `ipython` repo to `macos-latest` as well to circumvent the errors
- Make sure both `ipython` and `ipykernel` test on Arm and Intel versions

IMHO the latter one should be safer in the long term, but definitely requires more efforts to deal with the failing tests on Intel.

PR #14651, opened 2025-01-16
TST: Fix test_nonascii_path to use the nonascii path

sethtroisi:
This is currently passing because a previous test has left a foo file in the temp directory.

When I run just this single test (or in some parallel runs) it was previously failing

`pytest IPython/core/tests/test_ultratb.py::NonAsciiTest::test_nonascii_path`

### DIFF
--- a/IPython/core/tests/test_ultratb.py
+++ b/IPython/core/tests/test_ultratb.py
@@ -94,10 +94,10 @@ class NonAsciiTest(unittest.TestCase):
                 f.write(file_1)
 
             with prepended_to_syspath(td):
-                ip.run_cell("import foo")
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, that looks good to me, I'll try to investigate which test is leaving a foo and add a proper clenaup.

sethtroisi:
downstream macos has a failed test `tests/test_subshells.py::test_run_concurrently_sequence[True-are_subshells0]`.

This looks to be a know issue (#14545) that's unrelated to a test-only file. so I believe this is safe to merge.

Carreau:
thanks !

Issue #14304, opened 2024-01-24
remove ability to pass colors as kwargs of ColorScheme


Carreau:
I'm going to add a pending deprecation. 

If you are guided to this issue, please comment that you are still using it, so that you can be informed when I will remove it, and potentially delay removal.

Issue #14306, opened 2024-01-24
removal of case sensitivity of ColorSchemeTable


Carreau:
This is an issue to track the removal of ColorSchemeTable's set_active_scheme(case_sentive=....) key word argument. 

If you saw a warning bout this, please comment and subscribe to this issue to be warn of future deprecation and removal, and /or delay it.

PR #14652, opened 2025-01-21
Remove longtime Pending Deprecations

Carreau:


Carreau:
Should close #14304, #14306 once this is done.

PR #14675, opened 2025-01-22
Use AST token constants from token module.

### LINKED ISSUE #14455: Bug in Tokenizer for 3.12
Hey there!

I've discovered a bug with how the tokenizer/with the automatic forward-slash-parenthesization.

Specifically, the following will result in an error when run in IPython 8.25.0 on Python 3.12.3:
```
1| from pathlib import Path
2| 
3| (
4|     Path(".")
5|     / f")"
6|     / "a a a a a a a a a"
7| )
```

Interestingly, the issue is mitigated if the f-string on line 5 is:
- removed/replaced by a plain string
- has any character after the ending parenthesis

From a little digging, the tokenizer starts a new line (in `tokens_by_line`) when it encounters the new line at the end of the f-string.
It looks like the end parenthesis in the f-string is becoming an `FSTRING_MIDDLE` token, and deincrementing `parenlev`.

See here https://github.com/ipython/ipython/blob/1b4607fbee253a718df14419414f624dfde1164e/IPython/core/inputtransformer2.py#L511-L556

sethtroisi:
First part of a two part change I'm proposing to use tokens from the `token` module instead of `tokenize`

After Python 3.7 all tokens are available in both `token` and `tokenize` modules, it makes more sense to me to use them from the token module as tokenize is re-exporting from token. Except that token is a common variable name (e.g. `for token in tokens`) which causes shadowing with `import token` in a few places. It's certainly feels correct to fix it here now that > Python 3.7 is required.
    
in tokenize's doc
> All constants from the token module are also exported from tokenize.
    
See https://github.com/python/cpython/commit/fc354f07855a9197e71f851ad930cbf5652f9160
and https://mail.python.org/pipermail/python-dev/2017-May/148080.html

### DIFF
--- a/IPython/utils/PyColorize.py
+++ b/IPython/utils/PyColorize.py
@@ -61,76 +61,70 @@ Colors = TermColors  # just a shorthand
 
 # Build a few color schemes
 NoColor = ColorScheme(
-    'NoColor',{
-    'header'         : Colors.NoColor,
-    token.NUMBER     : Colors.NoColor,
-    token.OP         : Colors.NoColor,
-    token.STRING     : Colors.NoColor,
-    tokenize.COMMENT : Colors.NoColor,
-    token.NAME       : Colors.NoColor,
-    token.ERRORTOKEN : Colors.NoColor,
-
-    _KEYWORD         : Colors.NoColor,
-    _TEXT            : Colors.NoColor,
-
-    'in_prompt'      : InputTermColors.NoColor,  # Input prompt
-    'in_number'      : InputTermColors.NoColor,  # Input prompt number
-    'in_prompt2'     : InputTermColors.NoColor, # Continuation prompt
-    'in_normal'      : InputTermColors.NoColor,  # color off (usu. Colors.Normal)
-
-    'out_prompt'     : Colors.NoColor, # Output prompt
-    'out_number'     : Colors.NoColor, # Output prompt number
-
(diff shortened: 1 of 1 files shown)


IPython/utils/PyColorize.py
```diff
@@ -61,76 +61,70 @@
 
 # Build a few color schemes
 NoColor = ColorScheme(
-    'NoColor',{
-    'header'         : Colors.NoColor,
-    token.NUMBER     : Colors.NoColor,
-    token.OP         : Colors.NoColor,
-    token.STRING     : Colors.NoColor,
-    tokenize.COMMENT : Colors.NoColor,
```
sethtroisi, line 69: There are three changes of `tokenize.COMMENT` to `token.COMMENT`,  afterwards darker wants me to reformat these. I'm happy to revert the formatting if you feel it's not useful here.

Carreau (approved):
That seem reasonable as they are the same. I need to clean those files anyway, as IIRC all of these tokens are unused now.
``` 
"in_prompt": InputTermColors.NoColor,  # Input prompt
        "in_number": InputTermColors.NoColor,  # Input prompt number
        "in_prompt2": InputTermColors.NoColor,  # Continuation prompt
        "in_normal": InputTermColors.NoColor,  # color off (usu. Colors.Normal)
        "out_prompt": Colors.NoColor,  # Output prompt
        "out_number": Colors.NoColor, 
```

PR #14678, opened 2025-01-23
Delete prompt colors

sethtroisi:
Follow up to #14675,

These colors were once used by a prompt system, but was deleted circa release 5.0
The code referencing 'in_prompt', 'out_prompt' was deleted in 566cfa8349748a97e8dfa590fe7a7c6d981b5b2b part of #9471
All the tests pass and there's no usage of the 'in_number' or 'out_prompt' in the codebase


Carreau:
Oh ! Thanks for looking into this and verifying !

PR #14700, opened 2025-01-30
Improve Qt object management and performance

### LINKED ISSUE #14581: Enabling %matplotlib qt in IPython shell slows it down significantly
To reproduce:
Fire up a terminal, activate conda environment and enter ipython. This is a normal ipython terminal (not jupyter console or jupyter qtconsole).

At first, the shell is responsive even after many imports and trying out different data manipulation operations. As soon as `%matplotlib qt`is enabled, the shell slows down to a crawl, both input and time to show output. It feels as if there is a large roundtrip that is being taken by the input.

I understand that the matplotlib event loop needs to hook into ipython's loop or something along those lines, but it's very slow and the input delay is significant.

Other backends/frontends:
It looks like the issue does not exist with jupyter-qtconsole.
Update 11/17/24:  Performance degradation does not happen with c.InteractiveShellApp.matplotlib = "tkagg"
Looks like something is causing the qt backend to degrade performance.

OS: Windows 11 24H2 Build 26100.2314
Terminal: Windows Terminal 1.21.2911.0 (native, no WSL)
```
Relevant packages:
python            |     3.12.7   |       hce54a09_0_cpython |  conda-forge
ipython           | 8.29.0   | pyh7428d3b_0 | conda-forge
ipython_genutils | 0.2.0  |  pyhd8ed1ab_1 |  conda-forge
prompt-toolkit |   3.0.48 |    pyha770c72_0 |        conda-forge noarch
pygments        |  2.18.0 |    pyhd8ed1ab_0   |      conda-forge noarch
matplotlib        | 3.9.2   | py312h2e8e312_2 | conda-forge
matplotlib-base  |  3.9.2   | py312h90004f6_2 |  conda-forge
matplotlib-inline | 0.1.7    | pyhd8ed1ab_0   |  conda-forge
jupyter                  | 1.1.1   | pyhd8ed1ab_0 | conda-forge
jupyter-lsp              | 2.2.5  |  pyhd8ed1ab_0 | conda-forge
jupyter_client           | 7.4.9  |  pyhd8ed1ab_0 | conda-forge
jupyter_console       |    6.6.3 |   pyhd8ed1ab_0 |  conda-forge
jupyter_core            |  5.7.2   | pyh5737063_1 |  conda-forge
jupyter_events        |    0.10.0 |  pyhd8ed1ab_0  | conda-forge
jupyter_server        |    2.14.2  | pyhd8ed1ab_0 | conda-forge
jupyter_server_terminals  0.5.3  |  pyhd8ed1ab_0 | conda-forge
jupyterlab              |  4.3.0 | pyhd8ed1ab_0 | conda-forge
jupyterlab_pygments  |     0.3.0  |  pyhd8ed1ab_1 | conda-forge
jupyterlab_server       |  2.27.3  | pyhd8ed1ab_0 | conda-forge
jupyterlab_widgets    |    3.0.13  | pyhd8ed1ab_0 | conda-forge
```

I am happy to run any tests on my end to narrow down the issue. I have a fairly beefy workstation and this is really the only situation where I encounter lag with my inputs/outputs in the terminal.

jdranczewski:
This PR tries to address perceived lag in the IPython console when the `%gui qt` eventloop is enabled (https://github.com/ipython/ipython/issues/14581) as well as wider management of Qt objects in said eventloop, initially addressed in https://github.com/ipython/ipython/issues/14240.

The lag is resolved by setting the timer for exiting the Qt eventloop to 10ms instead of 50ms. This is consistent with the Tk implementation and removes the observable lag on Windows.

I also took this opportunity to restructure the handling of Qt objects - the old implementation created a new `QEventLoop` and `QTimer` each time the inputhook was called. Since `context.input_is_ready()` pretty much always returns True after 10ms in my tests, this meant creating these objects anew 20 times per second (or 100 with this proposed timer patch). https://github.com/ipython/ipython/issues/14240 adds garbage collection for the `QEventLoop`, but not for the `QTimer` (which is a feature on Windows only).

I propose creating one instance of `QEventLoop` and `QTimer` initially, and then just reusing them. In my testing this doesn't improve performance (Qt is pretty quick at making new objects it seems), but it does avoid garbage collection issues.

Note that a `QSocketNotifier` _is_ still created every iteration on POSIX platforms. I'm not sure if that is garbage collected or not. The value of `context.fileno()` could conceivably change, so we may need to recreate it as it is done now instead of reusing. I'm not sure, so I leave it as is.

Let me know if you would rather just change the timer to 10ms to fix https://github.com/ipython/ipython/issues/14581 and avoid the larger restructure in this PR. In that case we could only apply this commit: https://github.com/jdranczewski/ipython/commit/e65cd11a98546af5723251257c03fbaf3d35552b


jdranczewski:
(draft as I still need to test this patch on a Linux computer)

Carreau:
Ok ,let's try it.

Carreau:
thanks !

PR #14701, opened 2025-01-30
Use AST token constants from token module.

sethtroisi:
I'm lukewarm on this change, it's nice for all the constants to use the same namespace but it's ugly to have to change the `for token in tokens` loops.

--- 

After Python 3.7 all tokens are available in both `token` and `tokenize` modules

in tokenize's doc
> All constants from the token module are also exported from tokenize.

See https://github.com/python/cpython/commit/fc354f07855a9197e71f851ad930cbf5652f9160 and https://mail.python.org/pipermail/python-dev/2017-May/148080.html

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -189,6 +189,7 @@ import os
 import re
 import string
 import sys
+import token
 import tokenize
 import time
 import unicodedata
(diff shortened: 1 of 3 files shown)


Carreau:
I'm also not a big fan of now having to import both token and tokenize, but I can live with it.
It does seem more cannonical... so why not. Let me sleep on it a bit.

PR #14720, opened 2025-02-04
Better Docs and configurability of LLM.

Carreau:
This update the documentation and hooks in order to allow registering LLM with IPython.

- Add an option to pass value to construcrtor.
- Cleanup documentation that was refering to ipykernel.
- Extend documentation on how to add shortcuts. -


Carreau:
This still import jupyter-ai/numpy by default if installed and should only on first  attempt.
It should also show a spinner while waiting for first repsonse.

Carreau:
@meeseeksdev backport 8.x

Issue #14710, opened 2025-01-31
Have a 1 line tip-of-the-day at startup


Carreau:
With a --no-tip flag, just in case. 
It should pick at random with a hardcoded list of tips in the source code.
plus some tips that always display at some dates.

If we want to we could have a few "encrypted" (like rot 13) tips as easter eggs.

I'm thinking:

```
$ ipython
Python 3.13.0 | packaged by conda-forge | (main, Nov 27 2024, 19:18:26) [Clang 18.1.8 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.0.0.dev -- An enhanced Interactive Python. Type '?' for help.
Tip: You can assign shell commands with !: `files =! ls`

In [1]:
```

```
$ ipython
Python 3.13.0 | packaged by conda-forge | (main, Nov 27 2024, 19:18:26) [Clang 18.1.8 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.0.0.dev -- An enhanced Interactive Python. Type '?' for help.
Tip: IPython 0.1 was release exactly 25 years ago.

In [1]:
```

Carreau:
@Yann-P

ccordoba12:
I really like this idea! IPython has so much cool functionality that it's not easy to find and this could help users with that.

A couple of suggestions:
* Could you also make the list configurable so that we can add our own tips in Spyder?
* IMHO, the tip line should be separate by one blank line from the rest of the banner to be more visible.

Carreau:
> * Could you also make the list configurable so that we can add our own tips in Spyder?

Yes, by hardcoded I mean : by default not read from an external file. 

> IMHO, the tip line should be separate by one blank line from the rest of the banner to be more visible.

I'm worries we already have complaints that 3 lines is a lot.
I do want to support some minimal formatting (bold/code), I'm thinking allowing pygments' list of (token, text) instead of pure text ?

ccordoba12:
> I'm worries we already have complaints that 3 lines is a lot.

I understand, but I'm afraid that if the tip is not separate from the banner, users would see it as part of the banner and skip it.

> I do want to support some minimal formatting (bold/code), I'm thinking allowing pygments' list of (token, text) instead of pure text ?

Good idea too.

Carreau:
Mostly done by 14724

09F911029D74E35BD84156C5635688C0:
[...]

PR #14724, opened 2025-02-05
Add tips on terminal prompt

Yann-P:
#14710

image

I used a few tips from https://ipython.readthedocs.io/en/stable/interactive/tips.html

- [ ] Document this to inform that it may break some pipelines and advise using --no-tip when needed
- [ ] Add support for text formatting in tips

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -78,6 +78,7 @@ from IPython.core.macro import Macro
 from IPython.core.payload import PayloadManager
 from IPython.core.prefilter import PrefilterManager
 from IPython.core.profiledir import ProfileDir
+from IPython.core.tips import pick_tip
 from IPython.core.usage import default_banner
 from IPython.display import display
 from IPython.paths import get_ipython_dir
@@ -356,6 +357,12 @@ class InteractiveShell(SingletonConfigurable):
         """
         Enable magic commands to be called without the leading %.
         """
+    ).tag(config=True)
+
+    enable_tip = Bool(
+        True,
+        help="""
+        Set to show a tip when IPython starts.""",
     ).tag(config=True)
 
     banner1 = Unicode(default_banner,
@@ -926,6 +933,8 @@ class InteractiveShell(SingletonConfigurable):
             banner += '\nIPython profile: %s\n' % self.profile
         if self.banner2:
             banner += '\n' + self.banner2
+        elif self.enable_tip:
+            banner += "Tip: {tip}\n".format(tip=pick_tip())
         return banner
 
     def show_banner(self, banner=None):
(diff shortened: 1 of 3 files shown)


IPython/core/interactiveshell.py
```diff
@@ -926,6 +933,8 @@ def banner(self):
             banner += '\nIPython profile: %s\n' % self.profile
         if self.banner2:
             banner += '\n' + self.banner2
+        elif self.enable_tips:
```
Yann-P, line 936: elif because banner2 is used in the embedded environment; i assume we don't want the tip in this context 

is this assumption correct?

IPython/core/tips.py
```diff
@@ -0,0 +1,28 @@
+from datetime import datetime
+from random import choice
+from typing import Any
+
+_tips: Any = {
+    # (month, day)
+    "every_year": {
+        (1, 1): "Happy new year!",
+        (9, 12): "IPython was created {} years ago".format(datetime.now().year - 2001),
```
Carreau, line 9: TODO, check the real date :-)

IPython/core/tips.py
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "random": [
+        "Use %edit with no arguments to open an empty editor with a temporary file.",
```
Carreau, line 12: `%edit` or `F2` !

Carreau:
nice ! cc @ccordoba12, what is the best way you'd like to add spyder specific tips ?

IPython/core/tips.py
Yann-P: it was reversed, fixed and added source

IPython/core/tips.py
Yann-P: added

IPython/core/tips.py
```diff
@@ -0,0 +1,29 @@
...
+        (1, 1): "Happy new year!",
+        # https://mail.python.org/pipermail/python-list/2001-December/093408.html
+        (12, 9): "IPython was created {} years ago".format(datetime.now().year - 2001),
+    },
+    "random": [
+        "Use F2 or %edit with no arguments to open an empty editor with a temporary file.",
+        "Run your doctests from within IPython for development and debugging. The special %doctest_mode command toggles a mode where the prompt, output and exceptions display matches as closely as possible that of the default Python interpreter.",
+        "You can use `files = !ls *.png`",
+        "Use the IPython.lib.demo.Demo class to load any Python script as an interactive demo.",
+        "Put a ‘;’ at the end of a line to suppress the printing of output.",
```
Carreau, line 17: I'm just going to add some here for fun.
```suggestion
```suggestion
        "Put a ‘;’ at the end of a line to suppress the printing of output.",   
        "You can use Ctrl-O to force a new line in terminal IPython",
        "use `object?` to see the help on `object`, `object??` to view it's source",
        "`?` alone on a line will brings up IPython's help",
        "you can use `%hist` to view history, the the options with `%history?`",
```

Carreau:
I'm likely going to merge and we can iterate.

PR #14728, opened 2025-02-10
Remove v1 version of unicode and latex matchers

### LINKED ISSUE #14516: Tab completion on path with space not working MacOS
I would like to autocomplete path that has space in one of the sub-directory. The issue is, after it reaches the directory with space, the tab completion doesn't work anymore.

I just migrated from Ubuntu with older version of Python (Python 3.7) and iPython – I am pretty sure I did not encounter this issue before.

It would be very helpful to get this working, is there any workaround for this?

```sh
darren@Darrens-MacBook-Pro ~ % ipython --version
8.26.0
darren@Darrens-MacBook-Pro ~ % python --version
Python 3.12.4
```

https://github.com/user-attachments/assets/3504e0a1-9583-4eca-8cc8-efd32d0cea77

Yann-P:
Goal is to eventually remove all `_convert_matcher_v1_result_to_v2` from this file

Context:  I want to fix #14516. The bug is narrowed down to `completer.py`

@Carreau advised me to start removing v1 matchers first, so it's easier to work on the completer.
I'm starting with 2 small changes and will do the others in future MRs.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1615,45 +1615,41 @@ def back_unicode_name_matches(text: str) -> Tuple[str, Sequence[str]]:
 
 
 @context_matcher()
-def back_latex_name_matcher(context: CompletionContext):
-    """Match latex characters back to unicode name
-
-    Same as :any:`back_latex_name_matches`, but adopted to new Matcher API.
-    """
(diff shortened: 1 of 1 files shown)


Carreau:
Awesome !

PR #14735, opened 2025-02-12
more windows shot in the dark for ipython/ipykernel#1291

Carreau:
ipython/ipykernel#1291


IPython/utils/_process_common.py
```diff
@@ -91,12 +91,17 @@ def process_handler(
 
     try:
         out = callback(p)
+        if close_fds:
+            stderr.close()
```
Carreau, line 95:
```
```

PR #14739, opened 2025-02-14
HistoryManager cleanup.

Carreau:
Right now the cleanup was mostly done at process exit time,
but this leads to issues in downstream test that start to have dozen
of threads blocking, and make finding deadlocks harder

This PR try to cleanup the cleaning logic.

  - Try to seprate saving thread from history manager
  - Make sure to not keep reference to history manager in multiple places.
  - att a utility to refuse to create more then N
  concurrent history manager HistorySavingThread
  instances.
  - default to 1 in conftest.py
  - allow to set it to N via pytest fixture.

Some of this should likely be either context managers, or finalisers.

### DIFF
--- a/IPython/conftest.py
+++ b/IPython/conftest.py
@@ -81,6 +81,12 @@ def inject():
     from .core import page
 
     page.pager_page = nopage
+
+    from IPython.core.history import HistoryManager
+
+    # ensure we don't leak History managers
+    if os.name != "nt":
+        HistoryManager._max_inst = 1
     # yield
 
 
--- a/IPython/core/history.py
+++ b/IPython/core/history.py
@@ -14,6 +14,7 @@ import re
 import threading
 from pathlib import Path
 
+from contextlib import contextmanager
 from decorator import decorator
 from traitlets import (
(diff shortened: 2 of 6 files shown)


davidbrochart:
@Carreau You mentioned [here](https://github.com/ipython/ipykernel/pull/1291#issuecomment-2681161973) that this PR cleans the history manager threads at garbage collection.
Would it be possible to have a more explicit and deterministic way of doing so?

Carreau:
Not without a major refactor and breaking _a lot_ of APIs. Basically you would need `InteractiveShell` itself to be a contextmanager.

There are a number of other things that should be context managers. Some objects are doing crazy things like having `reset()` methods instead of having consumer just delete the current and create a new one – which made sens as a resource consumption 10 years ago, but not now anymore.

In conftest.py you can set HistorySavingThread's class attribute `_max_inst` to an integer and it will prevent creating more that that number of instances, so you can track the leaking ones.

See how I'm doing it in this PR, and the fixture I use to temporarily increase the number of HistoryManager for specific tests.

PR #14746, opened 2025-02-17
Ensure tests are available in sdist

bollwyvl:
This restores the `tests` folder for the built `sdist` tarball.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,8 +9,10 @@ include .flake8
 include .pre-commit-config.yaml
 include long_description.rst
 
+recursive-include tests *
+
 recursive-exclude tools *
-recursive-exclude tests *
+
 exclude tools
 exclude CONTRIBUTING.md
 exclude .editorconfig
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -371,7 +371,8 @@ license-files = ["LICENSE"]
 include-package-data = false
 
 [tool.setuptools.packages.find]
-exclude = ["setupext"]
+include = ["IPython*"]
+exclude = ["setupext", "tests*"]
 namespaces = false
 
 [tool.setuptools.package-data]


Carreau:
Ah thanks !

Carreau:
From my back of the napkin calculation, that's 3TB/week less for PyPI

bollwyvl:
Yeah, would be a helluva lot more if wheels used compression algorithms from this decade.

Carreau:
are there better compression methods that can still deflate with ZIP ?

bollwyvl:
Oh, it's a sordid tale. But yes, e.g. `zstd`, used by modern [`.rpm`](https://fedoraproject.org/wiki/Changes/Switch_RPMs_to_zstd_compression) and `.conda` packages, has been a supported feature in the `.zip` spec since ~2020. As of 2025, `zstd` is clearly still not in cpython stdlib, so any change is probably a non-starter until _at least_ that happened, as well as a plan for back-filling wheel installation (which, humorously, is _also_ not part of stdlib) to non-EOL pythons.

Carreau:
Yeah, I was more thinking can we use a compressing algos that is better than what is in stdlib, but that stdlib can decompress ? I know some of the algos are just ways of building the huffman table differently and so could work. But I haven't looked into that.

bollwyvl:
Oh, no: I reckon IPython unilaterally can't make much of a meaningful impact. Dunno what will come of [PEP-777](https://peps.python.org/pep-0777). 

Fully off-topic now: somewhere in the various threads, saw the lovely concept of: 
- keep using `.zip` with `flate`
  - inside, e.g. `IPython.dist.info` (or at least the parts that _can't_ be a long verbatim file) are as-is 
    - reboot `setup-requires: zstd` as a place to stash how to get at the rest of the contents
  - do whatever's clever with the "meat"

So inside the thing:

```
ipython-9.0.0.dev0-py3-none-any.zstd.in.a.not-really.whl
  ipython-9.0.0.dev0.dist-info/
  ipython-9.0.0.dev0.data.tar.zstd
  IPython.tar.zstd
```

Would give something comparable to:

```
400K  ipython-9.0.0.dev0-py3-none-any.tar.zstd
404K  ipython-9.0.0.dev0-py3-none-any.zstd.in.a.not-really.whl
573K  ipython-9.0.0.dev0-py3-none-any.whl
```

... but after paying 0.2-1.5mb to download `zstd`, basically can't be better than the status quo _unless_ everybody's playing the same game.

This is _basically_ what `.conda` packages do... `jedi`'s whl is 1.6mb, while its `.conda` counterpart is 0.82mb, could be as small as 0.7mb... some experiments with settings could get e.g. `jupyterlab` from 11.0mb down to 5.0mb.

Up in the browser, a de-`zstd` wasm build can be as small as 50kb, so in the pyodide case, there may really be something to some of these shenanigans.

PR #14748, opened 2025-02-17
Remove many useless variants.

### LINKED ISSUE #14745: remove some pyproject.toml extra (nbconvert, ipykernel...)

Carreau:
IPython is just terminal IPython now, so no need for those variants.

Should close #14745

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,26 +67,6 @@ doc = [
     "sphinx>=1.3",
     "sphinxcontrib-jquery",
 ]
-kernel = [
-    "ipykernel",
-]
-nbconvert = [
-    "nbconvert",
-]
-nbformat = [
-    "nbformat",
-]
-notebook = [
-    "ipywidgets",
-    "notebook",
-]
-parallel = [
-    "ipyparallel",
-]
-qtconsole = [
-    "qtconsole",
-]
-terminal = []
 test = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -52,9 +52,6 @@ Source = "https://github.com/ipython/ipython"
 Tracker = "https://github.com/ipython/ipython/issues"
 
 [project.optional-dependencies]
-black = [
```
bollwyvl, line 55: hm, as long as this is still in there... https://github.com/ipython/ipython/blob/aedea319404107bfec9f7c70356cbba68790c06e/IPython/terminal/interactiveshell.py#L136

Carreau: I was not sure about this one, it's not enabled by default because broken, and should we really have a `ipython[black]`, if we can just pip install black ?

bollwyvl: Welp, if it's broken, and not tested, a major bump is a great time to take it out!

`[extras]` _are_ a good way to have a declarative compatibility range, but as `pip` mostly disregards them on subsequent solves, they aren't good for much in onesy-twosey interactive package management. :shrug:

Carreau: you are right i'll probably will remove the black hook

Carreau:
Re-added black for now.

Issue #14591, opened 2024-11-29
Super compact error for docs tracebacks


Carreau:
https://github.com/PyWavelets/pywt/pull/741

see https://github.com/PyWavelets/pywt/pull/741#pullrequestreview-2194532812

PR #14752, opened 2025-02-20
Create a new xmode "Docs"

Carreau:
This is a compact mode with a minimal header and no traceback,
mosttly to be used in Documentation, where the context of the traceback
is rarely useful

```
$ ipython --InteractiveShell.xmode=Docs

In [1]: import pywt
    ...:
    ...: a = pywt.idwt([1,2,3,4,5], [1,2,3,4], 'db2' ,'symmetric')
    ...:
    ...: print(a)
Traceback (most recent call last): Cell In[1], line 3

->  a = pywt.idwt([1,2,3,4,5], [1,2,3,4], 'db2' ,'symmetric')

    [... 3 skipped frames]

ValueError: Coefficients arrays must have the same size.
```


Carreau:
This should take care of #14591 

@rgommers, tell me if you find that the following is compact enough, or if you want something even more compact.

```
Traceback (most recent call last): Cell In[1], line 3

->  a = pywt.idwt([1,2,3,4,5], [1,2,3,4], 'db2' ,'symmetric')

    [... 3 skipped frames]

ValueError: Coefficients arrays must have the same size.
```

rgommers:
That looks good to me, I think that should be enough. Thanks for pushing on this!

Carreau:
Thanks, I still need to clean that up, might a few more days though.

Thanks for the quick feedback.

PR #14753, opened 2025-02-20
handle imports and constexpr assigns in deduperreload

smacke:
There are a few types of code changes that cause the deduperreload algorithm to fall back to "OG" autoreload, which tends to behave less reliably:
- assigns
- imports

In particular, there's no reason in principle that we can't execute imports incrementally, so in this PR, we avoid falling back to OG autoreload if the user adds or removes an import statement (i.e., import nodes don't go to the "unfixable" array anymore).

Assignments are not something that we can execute with deduperreload in general since they could have complicated dependencies on other variables or augment attribute / subscript references, which could trigger more complicated side effects during module execution that we can't account for with deduperreload. For a specific subset of assignments, however, it should be safe to reload these incrementally with deduperreload -- in particular, if all of the following are satisfied:
- the assignment is to simple names only
- the RHS only involves literals and builtin names (i.e. no dependencies)
- the assignment does not occur in an `Enum` class which would end up wrapping the name in an enum wrapper different from the RHS type

So we also add a mechanism to detect that case and reload with deduperreload instead of falling back to OG autoreload, when we can.

Because the same statement could assign to multiple variables (e.g. imports that import multiple names, or assignments with multiple targets), we update the bookkeeping data structures to allow for that.

## How is this tested?
Manual + updated unit tests. After this change, deduperreload applies in more cases and autoreload works AMAZINGLY.

### DIFF
--- a/IPython/extensions/autoreload.py
+++ b/IPython/extensions/autoreload.py
@@ -274,11 +274,11 @@ class ModuleReloader:
             if do_reload:
                 self._report(f"Reloading '{modname}'.")
                 try:
-                    # check if deduperreload is viable for this module
-                    if self.deduper_reloader.maybe_reload_module(m):
-                        pass
-                    elif self.autoload_obj:
+                    if self.autoload_obj:
                         superreload(m, reload, self.old_objects, self.shell)
+                    # if not using autoload, check if deduperreload is viable for this module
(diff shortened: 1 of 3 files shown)


Carreau:
All the downstream failures are unrelated.

Carreau:
Great, let's try that in main.

PR #14778, opened 2025-02-24
Update Index.ipynb files to correct links (`%20` for `blanks`)

justineuro:
Blanks in URLS for links were replaced with `%20`.


justineuro:
This branch for the current request should not be `main` but `revert-14190-14021_cleanup_python2_in_lexers`.

PR #14779, opened 2025-02-24
Update `Index.ipynb` files (change `blanks` to `%20`)

justineuro:
`Blanks` in URLS for links were replaced with `%20`.

### DIFF
--- a/examples/IPython Kernel/Custom Display Logic.ipynb
+++ b/examples/IPython Kernel/Custom Display Logic.ipynb
@@ -18,7 +18,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As described in the [Rich Output](Rich Output.ipynb) tutorial, the IPython display system can display rich representations of objects in the following formats:\n",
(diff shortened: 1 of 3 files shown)


Carreau:
Is that not working ? 

Because it should be working, if not it's a bug in the frontend. and `%20` is  valid inside browsers only, this will be incorrect for things like nbterm: https://blog.jupyter.org/nbterm-jupyter-notebooks-in-the-terminal-6a2b55d08b70

justineuro:
Without the `%20`, this is how the page is rendered:  
  
ipython-Screenshot_2025-02-24_18-55-24
  
Maybe I can use the HTML `<a>` tag instead for the list items, e.g., `<a href='Cell Magics.ipynb'>Cell Magics.ipynb</a>` for the first item in the Tutorials list.

Carreau:
Hum, that is weird, it might be a problem with the github rendrer. But anyways this file is not in the main branch anymore; I guess I could accep a fix to 8.x, but right now you are sending the pr to a branch that should not recieve patches ([ipython:revert-14190-14021_cleanup_python2_in_lexers](https://github.com/ipython/ipython/tree/revert-14190-14021_cleanup_python2_in_lexers))

PR #14780, opened 2025-02-24
Enhance `%notebook` to save outputs, including MIME types and exceptions

### LINKED ISSUE #14750: Enhance %notebook to save outputs
Hi,

Can we make it possible for [%notebook](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-notebook) to optionally save output too? E.g.

```
In [1]: 1 + 3
Out[1]: 4

In [2]: %notebook foo.ipynb
```

Should include 4 as an output, but also work for matplotlib charts. More generally, I think it would be nice to remember all the mimetypes (like JupyterLab does).

Darshan808:
### Description:
This PR implements an enhancement to the `%notebook` magic command, enabling it to optionally save the output, including the MIME types and exceptions.

### Changes:
Introduced two new dictionaries, `output_mime_bundles` and `exceptions` in history manager, to store output MIME types and exceptions separately. This prevents inflating the history.

When the `%notebook` command is invoked, the code extracts MIME types and exceptions from the `history_manager` and appends them to the respective output cells in the generated notebook.

### Feedback Request:
Currently, we are using `mime_obj.get_figure()` to retrieve the MIME data for Matplotlib objects. However, this approach only works for the figure object. Are there any other approaches to capture MIME data ?

I am using `VerboseTB` to get a structured traceback. While this works for most errors, it doesn't handle syntax errors and some other edge cases very well. In `interactiveshell.py`, different types of errors are handled by using different traceback formats. Should we take the same approach here to handle different types of exceptions? It might result in some repetition, but it could improve error reporting for edge cases.


IPython/core/interactiveshell.py
```diff
@@ -3309,7 +3312,16 @@ def error_before_exec(value):
...
             # Write output to the database. Does nothing unless
             # history output logging is enabled.
-            self.history_manager.store_output(self.execution_count)
+            hm = self.history_manager
+            hm.store_output(self.execution_count)
+            exec_count = self.execution_count
+
+            # Capture MIME outputs and exceptions
+            if result.result:
+                hm.output_mime_bundles[exec_count] = deepcopy(result.result)
```
krassowski, line 3321: I wonder if we should prefer to:
a) perform the conversion to mime bundle here (call `self.shell.display_formatter.format`), or
b) store a deep copy (as now).

I think (a) would be better because otherwise user with a large data frame (say few GB) will quickly exhaust their RAM, unless.

IPython/core/interactiveshell.py
```diff
@@ -3309,7 +3312,16 @@ def error_before_exec(value):
...
+            if result.error_in_exec:
+                hm.exceptions[exec_count] = result.error_in_exec
```
krassowski, line 3323: Again, I think I would store MIME bundle representation of exceptions instead of exception objects to avoid memory leaks.

Darshan808:
Could you please take a look at the feedback request on my PR when you get a chance? Thanks!

IPython/core/interactiveshell.py
```diff
@@ -3309,12 +3314,54 @@ def error_before_exec(value):
...
+            exec_count = self.execution_count
+
+            if result.result:
+                # Format the result into a MIME bundle
+                try:
+                    mime_obj = (
+                        result.result[0]
+                        if isinstance(result.result, list)
+                        else result.result
+                    )
```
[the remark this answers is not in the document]
krassowski: To capture the result passed to `display()` we might want to also populate `output_mime_bundles` in:

https://github.com/ipython/ipython/blob/5a183ac6d44f1605ad1e11b53f889acefcd186bf/IPython/core/displaypub.py#L132-L138

I am not yet 100% sure if this is the best place, but seems fine for first iteration.

IPython/core/interactiveshell.py
```diff
@@ -3309,12 +3314,54 @@ def error_before_exec(value):
...
+                    mime_fig = mime_obj.get_figure()
+                    mime_bundle, _ = self.display_formatter.format(mime_fig)
+                except:
+                    # In case formatting fails, fallback to text/plain
+                    mime_bundle = {"text/plain": repr(result.result)}
```
krassowski, line 3333: I think we should not expect the formatting to fail; what might fail is calling `.get_figure` if output is not a matplotlib figure, or has this method but is not a matplotlib figure, e.g.:

```python
class MyAccount:
    def get_figure(self, round_to: int):
        """Returns the figure in my account"""
        pass
```

IPython/core/interactiveshell.py
```diff
@@ -3309,12 +3314,54 @@ def error_before_exec(value):
...
+                hm.output_mime_bundles[exec_count] = mime_bundle
```
krassowski, line 3334: I wonder if this should go to `log_output` which already has `format_dict` (result of `self.display_formatter.format`) ready to go. I think you may have wanted to have more control over what is passed to `self.display_formatter.format` but that possibly is not necessary, see my comment above.

https://github.com/ipython/ipython/blob/7a0c4945b59651c9a6f8f38174729ff15a60d0a9/IPython/core/displayhook.py#L243-L251

Keeping it all in the display hook instead would seem better for a some reasons:
1. from the code organization perspective it avoids inflating `interactiveshell.py` which is already huge
2. it keeps writes to `output_mime_bundles` and `output_hist_reprs` in a single place
3. it avoids computing the mime bundle twice (which may be expensive, especially for large data frames and figures)
4. it respects the quiet logic as `log_output` is only called if the user does not respect the output to be quiet:

https://github.com/ipython/ipython/blob/7a0c4945b59651c9a6f8f38174729ff15a60d0a9/IPython/core/displayhook.py#L84-L85

https://github.com/ipython/ipython/blob/7a0c4945b59651c9a6f8f38174729ff15a60d0a9/IPython/core/displayhook.py#L265-L274

That said when I tested your code it already seem to respect the `quiet` logic because `fill_exec_result` which populates `.result` is only called

krassowski (commented):
I think this is going in the right direction, I think the next steps in order would be:
1. Add tests:
   - for returned value gets included
   - the returned value with custom reprs (you can reuse [this dummy object](https://github.com/ipython/ipython/blob/5a183ac6d44f1605ad1e11b53f889acefcd186bf/tests/test_interactiveshell.py#L975-L983) for testing) get included with all these reprs in the MIME bundle
   - outputs displayed with `display()` get included
   - values printed with `print()` get included
   - exceptions get included
   - matplotlib integration (I would say lower priority, let's get the basics fist)
2. Add typing annotations
3. Implement logic for intercepting output displayed using `display()` 
4. Implement interception of stoud and stderr (`print()` and `print(file=sys.stderr)`)
5. Rethink how we capture matplotlib figures (i.e. possibly remove the special-casing, as (3) might allow use to just document that users need to use inline backend).

IPython/core/interactiveshell.py
```diff
@@ -3309,12 +3314,54 @@ def error_before_exec(value):
...
+                # Format the result into a MIME bundle
+                try:
+                    mime_obj = (
+                        result.result[0]
+                        if isinstance(result.result, list)
+                        else result.result
+                    )
+                    mime_fig = mime_obj.get_figure()
+                    mime_bundle, _ = self.display_formatter.format(mime_fig)
+                except:
```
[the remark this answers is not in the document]
krassowski: :+1: but I left this suggestion out because I think we should not be catching any exceptions here at all see https://github.com/ipython/ipython/pull/14780#discussion_r1971229104

IPython/core/interactiveshell.py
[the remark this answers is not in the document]
Darshan808: Yes, the outputs of `display` can be intercepted here, but I'm not sure how we can get `execution_count` in this context to later associate the bundle with the correct `input_cell`.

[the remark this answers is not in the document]
Darshan808: I'm facing a slight issue while capturing `mime_bundles` here.  

When only `display` is called, the MIME output is shown **before** `execution_count` is updated. However, when both an output and `display` are present, the output appears first, then `execution_count` is updated, and only **after** that is the MIME output displayed.  

How can we determine whether the MIME bundle should be associated with the **current** or **previous** `execution_count` in such cases?

In the screenshot below, in both cases `<Figure size ...>` and `1`, `self.shell.execution_count` is 5.

Screenshot from 2025-02-28 21-35-56

tests/test_magic.py
```diff
@@ -19,6 +19,7 @@
 from time import sleep
 from threading import Thread
 from unittest import TestCase, mock
+import nbformat
```
krassowski, line 22: This needs to be imported after `pytest.importorskip("nbformat")` as in:

https://github.com/ipython/ipython/blob/6cc4d06d0154b12a9d6fa36dad1a14920480c423/tests/test_run.py#L396-L399

tests/test_magic.py
```diff
@@ -914,6 +915,53 @@ def test_notebook_export_json():
         _ip.run_line_magic("notebook", "%s" % outfile)
 
 
+def test_notebook_export_json_with_display_and_error():
+    pytest.importorskip("nbformat")
+    _ip = get_ipython()
+    _ip.history_manager.reset()
+    _ip.history_manager.exceptions.clear()
+    _ip.history_manager.output_mime_bundles.clear()
```
krassowski, line 923: These should not be needed if `reset()` gets called, right?
```suggestion
```

tests/test_magic.py
```diff
@@ -914,6 +915,53 @@ def test_notebook_export_json():
...
+    for cell in nb.get("cells", []):
+        if cell.get("source", "").strip() == "display('test')":
+            for output in cell.get("outputs", []):
+                if output.get("output_type") in ("display_data", "execute_result"):
+                    data = output.get("data", {})
+                    text = data.get("text/plain", "").strip()
+                    assert text == "test", f"Expected 'test', got: {text}"
+                elif output.get("output_type") == "stream":
+                    text = output.get("text", "").strip()
+                    assert text == "test", f"Expected 'test', got: {text}"
```
krassowski, line 962: This test will not fail if there are no outputs. Something similar to `error_found` is needed here.

tests/test_magic.py
```diff
@@ -914,6 +915,53 @@ def test_notebook_export_json():
...
 
 
+def test_notebook_export_json_with_display_and_error():
+    pytest.importorskip("nbformat")
+    _ip = get_ipython()
+    _ip.history_manager.reset()
+    _ip.history_manager.exceptions.clear()
+    _ip.history_manager.output_mime_bundles.clear()
+
+    cmds = ["1/0", "display('test')"]
```
krassowski, line 925: We also need something like "1+1", and "print('hello')".

krassowski:
> If I include stdout outputs, everything printed to the console (errors, tracebacks, etc.) gets captured, but we don’t need to capture them again since we already have logic for it. I’m looking into solutions to capture just print statements and others if required, but would really appreciate any feedback or guidance you have!

Can you give an example? It might be best to demonstrate by writing a test.

Darshan808:
The test now effectively checks for display outputs, errors, normal outputs, and stream outputs from `print` statements. However, recording `print` statements isn't implemented yet, and I'm exploring ways to add it. Locally, the test fails due to the missing print output, but on CI, it is failing because normal outputs aren't being recorded. Investigating the issue further.

krassowski:
And the first exception is incorrect as it is stream and the text renderer instead of error renderer will be used:

image

which will mean links to paths would not show up.

Here is an idea to simplify the test: instead of having a bespoke logic for each cell, let's use nbformat to create a notebook with expected code cells first, and then run it and compare the results.

krassowski:
It feels like Tee should not be needed because IPython/IPykernel already has a way to get the `stream` for `print()`, right? Looking into it.

> ```
>  for cmd in cmds:
>     _ip.run_cell(cmd, store_history=True)
> ```

Agree, that's should be enough. I will dig into this.

tests/test_magic.py
```diff
@@ -914,6 +914,60 @@ def test_notebook_export_json():
...
+
+    _ip = get_ipython()
+    _ip.history_manager.reset()
+
+    cmds = [
+        "display('test')",
+        "1+1",
+        "1/0",
+        "print('test')",
+        "1",
```
krassowski, line 930: It looks current implementation does not cover the stderr. Let's add it to the test. Also, 1 and 1+1 test the same thing right?

```suggestion
        "import sys",
        "print('test stderr', file=sys.stderr)"
```

tests/test_magic.py
```diff
@@ -914,8 +912,68 @@ def test_notebook_export_json():
...
+
+    for i, command in enumerate(commands):
+        actual = actual_nb["cells"][i]
+        expected = expected_nb["cells"][i]
+        assert expected["source"] == command
+        assert actual["source"] == expected["source"]
+        if command == "1/0":
+            # remove traceback from comparison, as traceback formatting will vary
+            actual["outputs"][0].pop("traceback")
+            expected["outputs"][0].pop("traceback")
```
krassowski, line 970: @Darshan808 when you get a chance - do you think we should change the traceback logic to make these two match, or keep this exception?

krassowski:
The failure with `AttributeError: 'CapturingTee' object has no attribute '_is_main_process'` confirms my fear that swapping stdout may have unintended side effects. Let's see if we can recover without pulling in the same complex logic as is needed to make it work in ipykernel.

tests/test_magic.py
Darshan808: I previously tried matching the traceback logic, and they seemed to align. In this snapshot, characters like `\x1b{31m` are visible, but for some reason, they aren't appearing during testing.

image

IPython/core/interactiveshell.py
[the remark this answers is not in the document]
Darshan808: @krassowski 
Got the issue!  Ipython is using `nocolor` as color value in the test environment, while nbclient was set to `neutral` coloring. Once I updated ipython to use `neutral` colors, the test started passing smoothly.

tests/test_magic.py
Darshan808: @krassowski 
Got the issue!  Ipython is using `nocolor` as color value in the test environment, while nbclient was set to `neutral` coloring. Once I updated ipython to use `neutral` colors, the test started passing smoothly.

Darshan808:
> Let's see if we can recover without pulling in the same complex logic as is needed to make it work in ipykernel.

Can we do something like storing the previous `sys.stdout` object inside our `CapturingTee` object
```python
self.ostream = getattr(sys, channel)
```
And when attribute lookup fails, we return the original attribute from previously stored `sys.stdout` object
```python
def __getattr__(self, name):
    """Delegate any other attribute access to the original stream."""
    if name in self._original_attrs:
        return getattr(self.ostream, name)
    raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
```

Tried this in https://github.com/Darshan808/ipython/actions/runs/13872664945/job/38821201850?pr=2, The downstream test passed on Ubuntu but stalled on macOS.

Darshan808:
As expected, downstream tests are passing! 👍 
However, Pytest is still failing, similar to what I encountered before.

> the outputs are correctly recorded and included when running the test locally (where I have the IPython version from this branch installed), but they are not included in the CI environment.

I'm looking into this, but haven't yet found the root cause.

krassowski:
So there are three failures:

```python
FAILED tests/test_magic.py::test_notebook_export_json_with_output - AssertionError: Outputs do not match for cell 1 with source "display('test')"
assert [] == [{'data': {'t...isplay_data'}]
  Right contains one more item: {'data': {'text/plain': "'test'"}, 'metadata': {}, 'output_type': 'display_data'}
  Full diff:
    [
  +  ,
  -  {'data': {'text/plain': "'test'"},
  -   'metadata': {},
  -   'output_type': 'display_data'},
    ]
FAILED tests/test_magic_terminal.py::PasteTestCase::test_paste_echo - AssertionError: assert '\n        a ...ted text --\n' == '\n        a ...ted text --\n'
    
  -         a = 100
  +         a = 100
  ?             +++++   +++++
  -         b = 200
  +         b = 200
  ?             +++++   +++++
  + 
    ## -- End pasted text --
FAILED tests/test_oinspect.py::test_pinfo_docstring_dynamic - AssertionError: assert None
 +  where None = <function search at 0x7f7cdd416840>('Docstring:\\s+cdoc for prop', '\x1b[31mType:\x1b[39m        property\n\x1b[31mString form:\x1b[39m <property object at 0x7f7cce2509f0>\n\x1b[31mDocstring:\x1b[39m   cdoc for prop\n')
 +    where <function search at 0x7f7cdd416840> = re.search
 +    and   '\x1b[31mType:\x1b[39m        property\n\x1b[31mString form:\x1b[39m <property object at 0x7f7cce2509f0>\n\x1b[31mDocstring:\x1b[39m   cdoc for prop\n' = CaptureResult(out='\x1b[31mType:\x1b[39m        property\n\x1b[31mString form:\x1b[39m <property object at 0x7f7cce2509f0>\n\x1b[31mDocstring:\x1b[39m   cdoc for prop\n', err='').out
```


`test_paste_echo` swaps stdout so it looks related:

https://github.com/ipython/ipython/blob/0c9949c31e85b7c40e88ee5f2a94951c502e2c9a/tests/test_magic_terminal.py#L182-L197

`test_pinfo_docstring_dynamic` uses `capsys`, so again not surprised that this fails:

https://github.com/ipython/ipython/blob/0c9949c31e85b7c40e88ee5f2a94951c502e2c9a/tests/test_oinspect.py#L508-L531

We could replace capsys with `CapturingTee` if it was somehow exposed on `InteractiveShell`.

krassowski:
> We could replace capsys with `CapturingTee` if it was somehow exposed on `InteractiveShell`.

Or we could try to guess if we are inside a test environment which tries to capture things and then avoid activating `CapturingTee` logic. I guess for `test_paste_echo` we could check `sys.stdout.write` veracity and for `capsys` we could check `sys.stdout` in its entirety. But this would fail inside ipykernel as is also swapping the streams, but in ipykernel case we do want to intercept them.

Darshan808:
> test_paste_echo swaps stdout so it looks related:

These tests swap `sys.stdout`, but `CapturingTee` interacts with `sys.stdout`. 
```python
self.ostream.write(data)
```
Also the CI logs indicated an assertion mismatch rather than no output being captured.
I found that the failure was due to the `colors` attribute of IPython being set to `neutral` in the previous test, `test_notebook_export_json_with_output`. Resetting it to `nocolor` resolved the issue.

Regarding the outputs not being captured, logging in CI revealed that the `execution_count` was not starting from 1. Fixing this allowed the test to pass.

But Downstream Tests `pytest ipykernel` seems to stall sometime

krassowski:
Maybe `IPythonHistorySavingThread` is just a red herring? It might be related to `OutStream` subclass check in: https://github.com/ipython/ipykernel/blob/ff34a51075f7a6baa69124bdeae1974ba9ec4108/ipykernel/ipkernel.py#L741-L776

```
FAILED tests/test_kernel.py::test_print_to_correct_cell_from_thread - AssertionError: assert 'stderr' == 'stdout'
  
  - stdout
  + stderr
FAILED tests/test_kernel.py::test_print_to_correct_cell_from_child_thread - AssertionError: assert 'stderr' == 'stdout'
  
  - stdout
  + stderr
```

krassowski:
> It might be related to `OutStream` subclass check

Yes it was. Should be all green now.

krassowski (approved):
Thank you @Darshan808!

PR #14793, opened 2025-02-25
Allow color_scheme back for TB constructor

Carreau:
For https://github.com/napari/napari/issues/7649

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -433,6 +433,10 @@ class ListTB(TBTools):
             return "<unprintable %s object>" % type(value).__name__
 
 
+_sentinel = object()
+_default = "default"
+
+
 # ----------------------------------------------------------------------------
 class VerboseTB(TBTools):
     """A port of Ka-Ping Yee's cgitb.py module that outputs color text instead
@@ -450,7 +454,7 @@ class VerboseTB(TBTools):
     def __init__(
         self,
         # TODO: no default ?
-        theme_name: str = "linux",
+        theme_name: str = _default,
         call_pdb: bool = False,
         ostream: Any = None,
         tb_offset: int = 0,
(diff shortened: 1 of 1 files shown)


IPython/core/ultratb.py
```diff
@@ -450,21 +453,49 @@ class VerboseTB(TBTools):
...
+        theme_name: str = _default,
         call_pdb: bool = False,
         ostream: Any = None,
         tb_offset: int = 0,
         long_header: bool = False,
         include_vars: bool = True,
         check_cache: Callable[[], None] | None = None,
         debugger_cls: type | None = None,
+        *,
+        color_scheme=_sentinel,
```
Carreau, line 465:
```suggestion
        color_scheme:Any=_sentinel,
```

IPython/core/ultratb.py
```diff
@@ -450,21 +453,49 @@ class VerboseTB(TBTools):
...
     ):
         """Specify traceback offset, headers and color scheme.
 
         Define how many frames to drop from the tracebacks. Calling it with
         tb_offset=1 allows use of this handler in interpreters which will have
         their own code at the top of the traceback (VerboseTB will first
         remove that frame before printing the traceback info)."""
+        if color_scheme is not _sentinel:
+            assert isinstnance(color_scheme, str)
```
Carreau, line 474:
```suggestion
            assert isinstance(color_scheme, str)
```

Carreau:
@jni if you can try that with napari.

jni:
Tested and works fine on napari main. The tests fail because we have warnings as errors, but the functionality is fine and I can confirm that the deprecation warnings are the *only* errors with:

```
pytest -p no:warnings napari/plugins/_tests/test_exceptions.py
```

which passes.

jni:
Thank you! 🙏

Carreau:
I'll try to do a beta2 later today, with one extra windows and unicode fix.

PR #14837, opened 2025-03-13
Test for globals completion failing in methods

krassowski:
- [x] Add a test for globals completions in methods
- [ ] Fix the issue


Darshan808:
https://github.com/ipython/ipython/pull/14838 fixes the issue and handles other edge cases as well.

krassowski:
Fix in https://github.com/ipython/ipython/pull/14838

PR #14840, opened 2025-03-17
remove numfocus funding

Carreau:


pyproject.toml
```diff
@@ -47,7 +47,6 @@ content-type = "text/x-rst"
 [project.urls]
 Homepage = "https://ipython.org"
 Documentation = "https://ipython.readthedocs.io/"
-Funding = "https://numfocus.org/"
```
krassowski, line 50: In many places we changed it to `https://jupyter.org/about#donate`

PR #14846, opened 2025-03-25
Add request number and logging (do not use `print`)

### LINKED ISSUE #14844: All inline completion LLM requests have the request number set to zero
This is set to zero here:

https://github.com/ipython/ipython/blob/c4fbfc2624800d1fab1a353f6ad1954e14328d23/IPython/terminal/shortcuts/auto_suggest.py#L389-L390

and the docstring argues that:

https://github.com/ipython/ipython/blob/c4fbfc2624800d1fab1a353f6ad1954e14328d23/IPython/terminal/shortcuts/auto_suggest.py#L367-L369

However, this is incorrect, because the request ID is not meant to represent a cell number, but instead represents the number of the request. In jupyter-ai it is used to distinguish multiple replies when sent asynchronously over the websocket, but it can also be used by the model provider to track the requests internally.

### LINKED ISSUE #14847: Async tests do not run on CI
Here is an example:

```
tests/test_shortcuts.py s.............................................ss........                                 [ 91%]
```

The first test is an async test I am adding in https://github.com/ipython/ipython/pull/14846 but the other two are

https://github.com/ipython/ipython/blob/e6f0d4243c6e4416f704ad7575851e65281396e1/tests/test_shortcuts.py#L222

https://github.com/ipython/ipython/blob/e6f0d4243c6e4416f704ad7575851e65281396e1/tests/test_shortcuts.py#L274

Three suggestions:
- run tests on CI with `--verbose` so that the reason for skipping tests is displayed prominently
- fix the issue by running with `--asyncio-mode=auto` (not sure if this is a good fix)
- increase min coverage to avoid this

krassowski:
- fixes #14844 
- fixes #14847

### DIFF
--- a/IPython/terminal/shortcuts/auto_suggest.py
+++ b/IPython/terminal/shortcuts/auto_suggest.py
@@ -351,7 +351,7 @@ class NavigableAutoSuggestFromHistory(AutoSuggestFromHistory):
             try:
                 await self._trigger_llm_core(buffer)
             except Exception as e:
-                get_ipython().log.error(f"error {e}")
+                get_ipython().log.error("error %s", e)
                 raise
 
         # here we need a cancellable task so we can't just await the error catched
@@ -389,7 +389,7 @@ class NavigableAutoSuggestFromHistory(AutoSuggestFromHistory):
 
         hm = buffer.history.shell.history_manager
         prefix = self._llm_prefixer(hm)
-        get_ipython().log.debug(f"prefix: {prefix}")
+        get_ipython().log.debug("prefix: %s", prefix)
 
         self._request_number += 1
         request_number = self._request_number
--- a/IPython/terminal/shortcuts/auto_suggest.py
+++ b/IPython/terminal/shortcuts/auto_suggest.py
@@ -185,6 +185,7 @@ class NavigableAutoSuggestFromHistory(AutoSuggestFromHistory):
         self.skip_lines = 0
         self._connected_apps = []
         self._llm_provider = None
+        self._request_number = 0
 
     def reset_history_position(self, _: Buffer) -> None:
         self.skip_lines = 0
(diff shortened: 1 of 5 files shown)


IPython/terminal/shortcuts/auto_suggest.py
```diff
@@ -350,7 +351,7 @@ async def error_catcher(buffer):
             try:
                 await self._trigger_llm_core(buffer)
             except Exception as e:
-                get_ipython().log.error("error")
+                get_ipython().log.error(f"error {e}")
```
Carreau, line 354: You should never format strings in logging: https://docs.astral.sh/ruff/rules/logging-percent-format/, in addition formatting (for %) prevent structured logging.

Carreau (approved):

Darshan808:
Everything looks great! ✅

tests/test_shortcuts.py
```diff
@@ -219,6 +241,7 @@ def test_other_providers():
     assert swap_autosuggestion_down(event) is None
 
 
+@pytest.mark.asyncio
```
Carreau, line 244: If you have time (or as a separate issue/PR), it would be awesome to:

 1. mark pytest.mark.anyio (to make sure we test on asyncio and trio)
 2. make it an error at least on CI to have an async test skipped.

krassowski: I will do so in another PR

> mark pytest.mark.anyio (to make sure we test on asyncio and trio)

I guess also here?

https://github.com/ipython/ipython/blob/46b09de968521d08a2903316090557da547007ee/tests/test_magic.py#L1248-L1268

> make it an error at least on CI to have an async test skipped.

agree a good idea, I guess we could grep pytest output - or any other ideas?

Carreau: I was hopping there was a pytest option...

krassowski:

PR #14876, opened 2025-04-15
Fix syntax warning in Python 3.14 (remove `return` from `finally` block)

ellert:
SyntaxWarning: 'return' in a 'finally' block

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -3152,14 +3152,17 @@ class InteractiveShell(SingletonConfigurable):
         try:
             result = runner(coro)
         except BaseException as e:
-            info = ExecutionInfo(
-                raw_cell, store_history, silent, shell_futures, cell_id
-            )
-            result = ExecutionResult(info)
-            result.error_in_exec = e
-            self.showtraceback(running_compiled_code=True)
-        finally:
-            return result
+            try:
+                info = ExecutionInfo(
+                    raw_cell, store_history, silent, shell_futures, cell_id
+                )
+                result = ExecutionResult(info)
+                result.error_in_exec = e
+                self.showtraceback(running_compiled_code=True)
+            except:
+                pass
+
+        return result
 
     def should_run_async(
         self, raw_cell: str, *, transformed_cell=None, preprocessing_exc_tuple=None


krassowski:
Can you explain how to reproduce the issue and share a traceback?

ellert:
> Can you explain how to reproduce the issue and share a traceback?

Sorry, I should have given more context.
This triggers a failure in the test suite for ipyparallel when using python 3.14.0a7
https://bugzilla.redhat.com/show_bug.cgi?id=2359525

Carreau (commented):
I'm not sure this is correct. the behaviour is changed if we raise a non base exception And None is returned I think.

ellert:
An exception must be either a subclass or an instance of BaseException (https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement).

If the execution of the except block could raise an exception here, that would be silently ignored by having the return statement in the finally block. This silently ignoring of exceptions is the reason for issuing the syntax warning. If this was the intended behaviour here, the except block should be put in a try block so that the ignoring of the exception is explicit.
```.py
        try:
            result = runner(coro)
        except BaseException as e:
            try:
                info = ExecutionInfo(
                    raw_cell, store_history, silent, shell_futures, cell_id
                )
                result = ExecutionResult(info)
                result.error_in_exec = e
                self.showtraceback(running_compiled_code=True)
            except:
                pass

        return result
```

Carreau:
You are right, I was reviewing from my phone, while doing something else. 

+1

krassowski:
Could the test failures be related?

```python
FAILED tests/test_interactiveshell.py::TestShowTracebackAttack::test_set_show_tracebacks_none - TypeError: 'NoneType' object is not callable
FAILED tests/test_magic.py::test_notebook_export_json_with_output - AssertionError: Outputs do not match for cell 2 with source "print('test')"
assert [{'name': 'st...esttest\n\n"}] == [{'name': 'st...t': 'test\n'}]
  At index 0 diff: {'name': 'stdout', 'output_type': 'stream', 'text': "\ndefaultdict(<class 'list'>, {})\n\ntesttest\n\n"} != {'name': 'stdout', 'output_type': 'stream', 'text': 'test\n'}
  Full diff:
    [
     {'name': 'stdout',
      'output_type': 'stream',
  -   'text': 'test\n'},
  ?            ----   --
  +   'text': '\n'
  +           "defaultdict(<class 'list'>, {})\n"
  +           '\n'
  +           'testtest\n'
  +           '\n'},
    ]
```

All tests pass on `main` branch:

image

Carreau:
for context, the pep that makes it a warning https://peps.python.org/pep-0765/

Carreau (approved):

PR #14877, opened 2025-04-15
Fix LLM prefix including the new lines which are used as spaces

krassowski:
- [x] Make the test fail if there are spurious new lines at the end
- [x] Fix the issue by using `text_before_cursor` instead of `text` in prefix


Carreau (approved):

krassowski:

Issue #14878, opened 2025-04-15
Completer tests are failing: `ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'`


krassowski:
Seen in both:
- https://github.com/ipython/ipython/pull/14876
- https://github.com/ipython/ipython/pull/14877
but was passing in:
- https://github.com/ipython/ipython/pull/14874

```
FAILED tests/test_completer.py::TestCompleter::test_abspath_file_completions - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_aimport_module_completer - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_all_completions_dups - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_back_latex_completion - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_back_unicode_completion - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_cell_magics - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_class_key_completion - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_completes_attributes_in_fstring_expressions - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_completes_globals_as_args_of_methods - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_completes_in_dict_expressions - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_completion_have_signature - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
FAILED tests/test_completer.py::TestCompleter::test_completions_have_type - ModuleNotFoundError: No module named 'langchain_core._api.__warningregistry__'
[...]

PR #14879, opened 2025-04-15
Mask issue caused by `langchain-core` release `0.3.52`

krassowski:
Workaround for #14878 until langchain-core is fixed (or we find a better workaround).


krassowski:

PR #14890, opened 2025-04-25
Fix interruption of `%%time` and `%%debug` magics

### LINKED ISSUE #14889: `%%time` does not raise on interruption
When a cell with a `%%time` magic gets interrupted it does not cancel execution of following cells. This can lead to inadvertent execution of following cells which may lead to data loss (e.g. if following cells overwrite a file and user intended to cancel that action).

https://github.com/user-attachments/assets/0959583f-c9f8-4882-b7fd-4e2414c3c69d

### Context
- Originally reported in https://github.com/jupyterlab/jupyterlab/issues/17504
- Similar to https://github.com/ipython/ipython/issues/14649 which concerned `script` magics, not `time`

Edit: it's `%%time` which is problematic, `%%timeit` works fine

krassowski:
- [x] Add a tests interruption of `%%time` and `%%timeit` magics (the former should fail)
- [x] Fix `%%time` magic exception capture
    Note that this fix is fully backward compatible since users would not get timing information if there was an exception anyways due to an early return in line 1379:
   https://github.com/ipython/ipython/blob/aa25535d3765641e31f1d218537540e41b6f1a74/IPython/core/magics/execution.py#L1377-L1379
- [x] Fix `%%debug` magic exception capture

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1656,7 +1656,6 @@ def _format_time(timespan, precision=3):
                 break
         return " ".join(time)
 
-
     # Unfortunately characters outside of range(128) can cause problems in
     # certain terminals.
     # See bug: https://bugs.launchpad.net/ipython/+bug/348466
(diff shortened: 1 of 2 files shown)


Darshan808:
LGTM 👍

PR #14902, opened 2025-05-21
The --config option with absolute paths will be loaded only once.

ryumasai:
Implements #14723

The --config option with absolute paths will be loaded only once.

It may not be necessary, but I hope I contributed!
This is my first contribution to this project, so please let me know if I am missing anything.

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -360,11 +360,14 @@ class BaseIPythonApplication(Application):
                 continue
             self.log.debug("Attempting to load config file: %s" %
                            self.config_file_name)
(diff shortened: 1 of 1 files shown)


krassowski:
It feels like this should be fixed in traitlets instead, this is if `filename` is absolute, the list of paths should not be iterated here:

https://github.com/ipython/traitlets/blob/064480919b68d7c5cd23e74f7e0b4c6f83f8bd0f/traitlets/config/application.py#L889-L906

qraqras:
@krassowski 
Thanks for your feedback.
Indeed it is.
I will re-create the PR in `traitlets`.

PR #14906, opened 2025-05-27
Enhance `%%time` Magic with `--no-raise-error` Flag 

### LINKED ISSUE #14896: `%%time` does not stop execution on exceptions
- https://github.com/ipython/ipython/issues/14889 reported that interruption does not stop execution of `%%time`,
- which was fixed by https://github.com/ipython/ipython/pull/14890

However, non-system exceptions also allow the execution to continue although it should not. Instead, we should probably call `sys.exit(SIGINT)` after any exception gets caught.

Darshan808:
### Summary of Changes:

This PR modifies the behavior of the `%%time` magic to support better execution flow by adding `--no-raise-error` flag.

### New Behavior Introduced:

**Default Behavior (no flag):**

- On keyboard interrupt or exception, the timing up to the interruption is printed.
- Error Traceback is printed
- Execution does not continue to the cells below.

**With `--no-raise-error` Flag:**

- On keyboard interrupt: Same as default—timing and traceback are printed, execution stops. (as it is used by stop button)

- On other exceptions:
   - The timing up to the exception is printed.
   - No traceback is shown.
   - Execution continues to the cells below.

Looking for feedback on whether this behavior is intuitive and desirable before proceeding to add tests.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -1338,12 +1338,14 @@ class ExecutionMagics(Magics):
                 Wall time: 0.00 s
                 Compiler : 0.78 s
         """
+        line_present = False
         # Try to parse --no-raise-error if present, else ignore unrecognized args
         try:
             args = magic_arguments.parse_argstring(self.time, line)
         except UsageError as e:
             # Only ignore UsageError if caused by unrecognized arguments
             # We'll manually check for --no-raise-error and remove it from line
+            line_present = True
             tokens = shlex.split(line)
             no_raise_error = False
             filtered_tokens = []
@@ -1363,6 +1365,9 @@ class ExecutionMagics(Magics):
         else:
             if not hasattr(args, "no_raise_error"):
                 args.no_raise_error = False
+
+        if line_present and cell:
+            raise UsageError("Can't use statement directly after '%%time'!")
 
         if cell:
             expr = self.shell.transform_cell(cell)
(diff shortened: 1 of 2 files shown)


krassowski:
The overall interaction model looks good, but we need to improve traceback handling, there is just too much noise and relevant traceback is not shown at times.

image

Darshan808:
> there is just too much noise and relevant traceback is not shown at times.

Instead of using `sys.exit()`, we could re-raise the exception to preserve the traceback. But still traceback looks a bit noisy.

Darshan808:
The only way we have currently to avoid stopping further execution is either by using `sys.exit()` or by reraising the exception from magic functions. However, `sys.exit()` doesn’t provide a helpful traceback. After reraising, we get the following traceback — and this traceback seems to be same across all magics.

image
image


Is this acceptable, If not I think we should create a separate issue and work on a follow-up PR to handle these tracebacks returned from magic functions more cleanly.

Darshan808:
The `tests/test_magic.py::test_time_no_var_expand` test is failing in the modified `time` function. The variables are being expanded even though we have the `@no_var_expand` decorator in place.  

I suspect this might be due to the order of the decorators — I tried reshuffling them in different combinations, but none of them made the test pass.  

I am looking into this but any insights or suggestions would be much appreciated.

krassowski:
I guess an extra `__tracebackhide__ = True` somewhere... I agree it would be good to open an issue and a separate PR if it has effect on other magics.

Darshan808:
> I agree it would be good to open an issue and a separate PR if it has effect on other magics. 

Opened an issue for this: https://github.com/ipython/ipython/issues/14923

Darshan808:
@krassowski 
Could you please take a moment to review this. Thanks!

krassowski (approved):
Tested locally, looks good!

PR #14908, opened 2025-05-29
Fix typing of `error_before_exec`, enhance `mypy` coverage

### LINKED ISSUE #14907: Incorrect typing for `error_before_exec`
`error_before_exec` is typed as `bool` here:

https://github.com/ipython/ipython/blob/b854379e3b801ce812c648e95f8846a1334e3195/IPython/core/interactiveshell.py#L286-L294

but later it is raised:

https://github.com/ipython/ipython/blob/b854379e3b801ce812c648e95f8846a1334e3195/IPython/core/interactiveshell.py#L305-L310

The tests suggest that both are `BaseException` descendants:

https://github.com/ipython/ipython/blob/b854379e3b801ce812c648e95f8846a1334e3195/tests/test_interactiveshell.py#L112-L118

A PR fixing the issue should start investigating why mypy was not complaining and make CI fail.

samuraikillers:
This PR fixes issue #14907 

Changes made:

1. `pyproject.toml` => changed ignore_error = false, so that we can catch BaseException like errors.

Drawback => Testing with mypy we see 54 errors for `ipython/IPython/core/interactiveshell.py`
Resolution => To flip certain parameter in pyproject.toml so that we get less errors and try to resolve each of them in segments to minimize review efforts.

2. `ipython/IPython/core/interactiveshell.py` =>  Upon making the best changes to `pyproject.toml` we reduce the errors to 3(we will cover the rest of them gradually, this is for starters).
Which were:
image

I had resolved 2 of them :
https://github.com/ipython/ipython/blob/b854379e3b801ce812c648e95f8846a1334e3195/IPython/core/interactiveshell.py#L334
and
https://github.com/ipython/ipython/blob/b854379e3b801ce812c648e95f8846a1334e3195/IPython/core/interactiveshell.py#L3602

Which eventually results in only 1 error, with current settings:
image

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -331,7 +331,7 @@ class InteractiveShell(SingletonConfigurable):
 
     _instance = None
     _user_ns: dict
-    _sys_modules_keys: set[str, AnyType]
+    _sys_modules_keys: set[str]
 
     inspector: oinspect.Inspector
 
@@ -1572,7 +1572,7 @@ class InteractiveShell(SingletonConfigurable):
             if isinstance(variables, str):
                 vlist = variables.split()
             else:
-                vlist = variables
+                vlist = list(variables)
             vdict = {}
             cf = sys._getframe(1)
             for name in vlist:
@@ -2196,7 +2196,7 @@ class InteractiveShell(SingletonConfigurable):
         except KeyboardInterrupt:
             print('\n' + self.get_exception_only(), file=sys.stderr)
 
-    def _showtraceback(self, etype, evalue, stb: str):
+    def _showtraceback(self, etype, evalue, stb: list[str]):
         """Actually show a traceback.
 
         Subclasses may override this method to put the traceback on a different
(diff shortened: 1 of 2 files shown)


samuraikillers:
@krassowski ,
Yepp, its working now, with the below change not seeing any issue:
image


For the rest of 30 errors I will try to suppress them as much as possible via ignore

krassowski:
If there are large groups of the same error code using `disable_error_code` would do over many `# ignore` comments

https://mypy.readthedocs.io/en/stable/error_codes.html#per-module-enabling-disabling-error-codes

samuraikillers:
Hi @krassowski , done with the changes, sorry for the multi-commits at the end.

I see 1 check is failing but don't think its because of my changes.

krassowski:
Yeah, don't worry about that one. Now, which one of these ignores we need to remove to get an error that would highlight the `error_before_exec` issue?

samuraikillers:
We have to remove `misc`.

samuraikillers:
Done

krassowski (approved):
Thank you @samuraikillers!

samuraikillers:
@krassowski glad to help,  I have started only 1 week before to contribute to open source. And very happy that I contributed to this project which is being used by thousands of users worldwide 😀

PR #14922, opened 2025-06-12
fix for deduperreload global env in case of function decorator from imported module

smacke:

In deduperreload, if the function we're patching is decorated by a function coming from another file, we don't want to use the `__globals__` of that function coming from the other file when we exec code, since it won't have the same references as the function we're trying to patch. Actually, I think we always should be using the module's `__dict__` as globals regardless for this case -- so I think we can just delete some code that was trying to be clever around setting the global env.

## How is this tested?
Tested with existing unit tests + added a new one that only passes with the new changes in this PR


smacke:
FYI @jaysarva

krassowski (approved):
Thank you @smacke!

PR #14926, opened 2025-06-24
Support data frames, series and objects with `__len__` in the `%whos` magic

vlulla:
This modification shows the shape of dataframe and series objects instead of printing the dataframe/series objects.

```python
## With this slight modification
[1] a,b,c = 1,2,3
[2] df = pd.DataFrame({"a":range(10),"b":range(10,20)})
[3] s = df["a"]
[4] %whos
Variable   Type         Data/Info
---------------------------------
a          int          1
b          int          2
c          int          3
df         DataFrame    Shape: (10, 2)
s          Series       Shape: (10,)
```

Compare the above to the way it currently prints the output:

```python
## current way 
[1] a,b,c = 1,2,3
[2] df = pd.DataFrame({"a":range(10),"b":range(10,20)})
[3] s = df["a"]
[4] %whos
Variable   Type         Data/Info
---------------------------------
a          int          1
b          int          2
c          int          3
df         DataFrame        a   b\n0   0  10\n1  <...>\n\n[10 rows x 2 columns]
s          Series       0    0\n1    1\n2    2\n <...> Length: 10, dtype: int64
```

### DIFF
--- a/IPython/core/magics/namespace.py
+++ b/IPython/core/magics/namespace.py
@@ -472,7 +472,7 @@ class NamespaceMagics(Magics):
                 print(f"Shape: {var.shape}")
             elif hasattr(var, "__len__"):
                 ## types that can be used in len function
-                print(f"Iterable with n={len(var)}")
+                print(var if isinstance(var, str) else f"n={len(var)}")
             else:
                 try:
                     vstr = str(var)
--- a/IPython/core/magics/namespace.py
+++ b/IPython/core/magics/namespace.py
@@ -361,6 +361,9 @@ class NamespaceMagics(Magics):
           - For numpy arrays, a summary with shape, number of
             elements, typecode and size in memory.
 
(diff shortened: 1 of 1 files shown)


krassowski:
Thank you, it does make sense as it aligns the output with numpy handling:

https://github.com/ipython/ipython/blob/d47970f7097668ec89a47120e14625d44b617d68/IPython/core/magics/namespace.py#L361-L362

I think the docstring should be updated too?

vlulla:
Will do!

vlulla:
I just realized that pyarrow objects still jumble up the formatting. I can rearrange the if conditions to display the len of pyarrow iterables, in addition to `seq_types` used in that function,  instead of actually printing them.

@krassowski would it be alright if i make this change too? This will require a little bit of code rearrangement but nothing too drastic.

krassowski (approved):
This makes sense to me, thank you @vlulla!

Issue #62, opened 2010-05-10
Global variables undefined in interactive use of embedded ipython shell


ipython:
Original Launchpad bug 399627: https://bugs.launchpad.net/ipython/+bug/399627
Reported by: h-fangohr (Hans Fangohr).

The error can be reproduced as follows:

1. Start Python, and start embedded ipython session. Following ipython's manual, we do

```
fangohr@eta:~$ python
Python 2.4.3 (#1, Jun  8 2009, 14:09:06) 
[GCC 4.1.2 20061115 (prerelease) (Debian 4.1.1-21)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from IPython.Shell import IPShellEmbed
>>> ipshell=IPShellEmbed()
>>> ipshell()
```

Within the just started ipython session, global variables are sometimes not visible. Two examples are:

Example 1:

```
In [1]: a=1

In [2]: def f(x):
   ...:     return a*x
   ...: 

In [3]: f(2)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)

/home/fangohr/<ipython console> 

/home/fangohr/<ipython console> in f(x)

NameError: global name 'a' is not defined
```

Example 2:

```
In [4]: b=1

In [5]: (lambda :b)()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)

/home/fangohr/<ipython console> 

/home/fangohr/<ipython console> in <lambda>()

NameError: global name 'b' is not defined
```

There is no error if "b=1; (lambda :b)()" is put in a script, and this script is executed using ipython's "run" command.

There is no error if "b=1; (lambda :b)()" is executed interactively after having started ipython.

The only way the error seems comes up is if an embedded IPython shell is started from a Python prompt AND the commands are executed interactively at the prompt.

I find the same error when trying ipython-0.9.1 (with python2.4).

The bug was reported by Olivier Klein to the nmag team (http://nmag.soton.ac.uk).

philippschw:
Problem still persists in ipython==4.2.0. Interestingly under windows it is no problem but together with ubuntu global variables are not recognized.

Use case:
```
from ipywidgets import interact, FloatSlider, IntSlider,RadioButtons, Dropdown

@interact(sv1 = radio_1, Scenario = drop_1, sv2 = slider_2)
def update_map(sv1,sv2, Scenario):
```

blackball:
Problem remains. IPython 3.1.0, Debian Whezzy.

[...]

PR #14933, opened 2025-06-30
Backport Pdb closure evaluation fix (list comprehension, generators) from Python 3.13

### LINKED ISSUE #256: List comprehension problem
While using `IPython import embed; embed()` in an ipdb session list comprehensions seem to have a broken scope, take the snippet as an example. Confirmed this is not an IPython issue.
image (41)

krassowski:
- [x] add a failing test case
- [x] backport https://github.com/python/cpython/pull/111094 into IPython codebase
- fixes https://github.com/gotcha/ipdb/issues/256
- solves the pdb-related comments on https://github.com/ipython/ipython/issues/62

image

### Context

IPython already includes bits of Pdb code from Python itself, as documented in:

https://github.com/ipython/ipython/blob/1e9a2e8b5a584ffe1973318fdabf2e39395ec367/IPython/core/debugger.py#L92-L110

I am taking a more cautious approach and separating the backport to a new file to make it easy to delete in the (near) future (1 year from now).

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -146,8 +146,10 @@ from pygments.token import Token
 
 
 if sys.version_info < (3, 13):
+
     class OldPdb(PdbClosureBackport, _OldPdb):
         pass
+
 else:
     OldPdb = _OldPdb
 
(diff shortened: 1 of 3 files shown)


krassowski:
We could use the same approach (`_exec_in_closure`) to solve https://github.com/ipython/ipython/issues/62 altogether, but that will need more testing - we could do so in a follow-up PR..

Issue #14900, opened 2025-05-18
Incorrect traceback after `%autoreload` in IPython 9.x with `DeduperReloader` enabled


daeyun:
## Environment

item | version / value
-- | --
IPython | 9.0.0 – 9.3.0.dev0 (main)
Python | 3.12.3
executing | 2.2.0
stack_data | 0.6.3
OS | Ubuntu 24.04.1

(Problem does **not** occur on IPython 8.36 with the legacy reloader.)

## Minimal reproducible example

Run each cell in a fresh kernel exactly as shown, one after the other, waiting for each to finish before executing the next.

### Cell 1: setup and first failure (shows correct filename and context)

```python
import tempfile, os, sys, traceback
from IPython import get_ipython

# Named temporary directory
tmpdir = os.path.join(tempfile.gettempdir(), "autoreload_demo")
os.makedirs(tmpdir, exist_ok=True)
sys.path.insert(0, tmpdir)
mod_path = os.path.join(tmpdir, "mymod.py")

def write(src: str):
    with open(mod_path, "w") as f:
        f.write(src)

write("def buggy():\n    return 1/0\n")

ip = get_ipython()
ip.run_line_magic("load_ext", "autoreload")
ip.run_line_magic("autoreload", "2")

# --- Toggle these two lines and the problem disappears ----
# from IPython.extensions.deduperreload.deduperreload import DeduperReloader
# DeduperReloader.enabled = False
# -----------------------------------------------------------------------

import mymod

try:
    mymod.buggy()
except Exception:
    traceback.print_exc()
```

Traceback (correct):

```
File "/tmp/autoreload_demo/mymod.py", line 2, in buggy
    return 1/0
           ~^~
ZeroDivisionError: division by zero
```

### Cell 2: change the code (triggers autoreload)

```python
write("def buggy():\n    return 42/0\n")
```

### Cell 3: call the function again

```python
try:
    mymod.buggy()
except Exception:
    traceback.print_exc()
```

Traceback (**incorrect**: no filename and context):

```
File "<string>", line 2, in buggy
ZeroDivisionError: division by zero
```

Expected:

```
File "/tmp/autoreload_demo/mymod.py", line 2, in buggy
    return 42/0
           ~~^~
ZeroDivisionError: division by zero
```

### Observations
- Running the three snippets in a single cell, or queuing all three cells at once (e.g., Run All), does not reproduce the problem. Any reload executed from subsequent cells produces the incorrect traceback.
[...]

PR #14945, opened 2025-07-23
Fix deduperreload file name in tracebacks

jaysarva:
This is a half-fix for tracebacks in deduperreload -- filenames are now properly shown in the traceback. Part 1 of fix to #14900
Line numbers are still not rendered, but we hope to send a fix for this in the future.

### DIFF
--- a/IPython/extensions/deduperreload/deduperreload.py
+++ b/IPython/extensions/deduperreload/deduperreload.py
@@ -434,7 +434,15 @@ class DeduperReloader(DeduperReloaderPatchingMixin):
                 global_env = ns.__dict__
                 if not isinstance(global_env, dict):
                     global_env = dict(global_env)
-                exec(func_code, global_env, local_env)  # type: ignore[arg-type]
+
+                # Compile with correct filename to preserve in traceback
+                filename = (
+                    getattr(to_patch_to, "__code__", None)
+                    and to_patch_to.__code__.co_filename
+                    or "<string>"
+                )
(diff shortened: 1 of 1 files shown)


jaysarva:
cc @smacke

smacke:
Thanks for adding this @jaysarva ! Not going to ping Mike since I imagine he's busy and this is not time sensitive, but for when he does see this: I fully vouch for this change as a solid partial fix for tracebacks getting messed up in deduper-reloaded modules (with more improvements to come for line numbers).

krassowski:
Thank you @jaysarva - could you add a test for it?

krassowski:
Just checking back here - if you do not have anything locally I could help with the test - let me know.

jaysarva:
Hi @krassowski I'm almost done with a more complete fix (just testing to make sure there's no bugs right now).  
it might be reasonable to just directly look at that instead of adding this incremental fix. hope to put it out in the next week or two.

krassowski:
No worries - I will be releasing 9.5 on Friday so maybe it can still be squeezed in. If you would still like to get this one in instead I will note it needs a rebase due to conflicts.

smacke:
> No worries - I will be releasing 9.5 on Friday so maybe it can still be squeezed in. If you would still like to get this one in instead I will note it needs a rebase due to conflicts.

I think I squeezed this fix into [this PR](https://github.com/ipython/ipython/pull/14960) btw, so I think we can close this

PR #14960, opened 2025-08-14
Better handling in deduperreload for patching functions with freevars

### LINKED ISSUE #14958: segmentation fault with autoreload due to not updating function __closure__
The code for updating old function references in `IPython/extensions/autoreload.py` is fairly straightforward:
```
func_attrs = [
    "__code__",
    "__defaults__",
    "__doc__",
    "__closure__",
    "__globals__",
    "__dict__",
]


def update_function(old, new):
    """Upgrade the code object of a function"""
    for name in func_attrs:
        try:
            setattr(old, name, getattr(new, name))
        except (AttributeError, TypeError):
            pass
```
But calling setattr on `__closure__` gets `AttributeError('readonly attribute')` so it doesn't actually update it.

We have the following code in reload.py:
```
def decorate(func):
    free_var_y = "y"
    free_var_x = "x"

    def wrapper(*args, **kwargs):
        print("x is:", free_var_x)
        print("y is:", free_var_y)
        return func(*args, **kwargs)
    return wrapper


@decorate
def f():
    return "v1"
```

In IPython we do:
```
In [1]: %load_ext autoreload

In [2]: %autoreload all

In [3]: from reload import f

In [4]: old_f = f

In [5]: old_f()
x is: x
y is: y
Out[5]: 'v1'
```

Now we update reload.py to be:
```
def decorate(func):
    free_var_ay = "ay"
    free_var_bx = "bx"

    def wrapper(*args, **kwargs):
        print("x is:", free_var_bx)
        print("y is:", free_var_ay)
        return func(*args, **kwargs)
    return wrapper


@decorate
def f():
    return "v2"
```
Now the co_freevars of the new code object are in a different order from the old function's `__closure__` attribute so I expected IPython to print the wrong values. But actually now when calling old_f or f again in IPython I get a segmentation fault. Also when a function has changed the number of variables it closes over, as IPython updates `__code__` but not `__closure__`, we will get an error like ValueError('wrapper() requires a code object with 3 free vars, not 6').

I am working on an environment to enable interactive development without restarting python using autoreload, where I addressed this in this commit: https://codeberg.org/sczi/swanky-python/commit/748952d0eda733c7abef5e7e01c1dafdc42e4b43

I just use ctypes.pythonapi.PyFunction_SetClosure which is a CPython only api, as I don't currently have a need for supporting pypy or other implementations. I considered some possibilities that should work on all python implementations:
1) find and update old references with gc.get_referrers
If we update all dict references to use the full new function object, that takes care of top level module references by way of `module.__dict__`, local variables by way of frame.f_locals dict, and class instance variables or methods by way of `class.__dict__`. But we would need special cases to handle classes with slots, and references stored in lists or other container types. But we can't include custom checks for every possible container type, just the containers in the standard library, and we can't update old references in immutable containers like tuples. So it would mostly work but cause occasional bugs by not updating everything properly.
2) use a trampoline function
We could generate a trampoline function with the same number of freevars as the old function object, that just wraps and calls the new version. This would avoid the errors about `ValueError('wrapper() requires a code object with 3 free vars, not 6')`. But it requires wrapping functions which would impact performance and pollute stack traces.

smacke:
This PR should fix https://github.com/ipython/ipython/issues/14958. When patching a decorated function, rather than giving it the union of the old version's and new version's freevars, we give it the old version's for anything callable, and the new version's for everything else. This ensure we use the old `__class__` attribute and avoid errors such as "TypeError: super(type, obj): obj must be an instance or subtype of type". Using old versions of other callables is also better since these references are the ones we maintain in deduperreload, rather than the new references. It does, however, mean that we need to reload changed decorated functions twice: once without decorators (to ensure the old version gets patched), and once with. This PR handles that as well.

Test plan: all the other unit tests pass + additionally added a new one which exercises this case.

### DIFF
--- a/IPython/extensions/deduperreload/deduperreload.py
+++ b/IPython/extensions/deduperreload/deduperreload.py
@@ -4,6 +4,7 @@ import builtins
 import contextlib
 import itertools
 import os
+import pickle
 import platform
 import sys
 import textwrap
@@ -415,7 +416,11 @@ class DeduperReloader(DeduperReloaderPatchingMixin):
         namespace_to_check = ns
         for prefix in prefixes:
             namespace_to_check = namespace_to_check.__dict__[prefix]
+        seen_names: set[str] = set()
         for names, new_ast_def in cur.defs_to_reload:
+            if len(names) == 1 and names[0] in seen_names:
+                continue
+            seen_names.update(names)
             local_env: dict[str, Any] = {}
             if (
                 isinstance(new_ast_def, (ast.FunctionDef, ast.AsyncFunctionDef))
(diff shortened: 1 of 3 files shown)


krassowski (commented):
Verified that this fixes the segfault for me:

image

IPython/extensions/deduperreload/deduperreload.py
```diff
@@ -434,37 +439,54 @@ def _patch_namespace_inner(
...
-                        "adding or removing property decorations not supported"
+                # Compile with correct filename to preserve in traceback
+                filename = (
+                    getattr(to_patch_to, "__code__", None)
+                    and to_patch_to.__code__.co_filename
+                    or "<string>"
+                )
+                func_asts = [ast.parse(func_code)]
+                if len(cast(ast.FunctionDef, func_asts[0].body[0]).decorator_list) > 0:
+                    without_decorator_list = pickle.loads(pickle.dumps(func_asts[0]))
```
krassowski, line 450: Is it preferred over `copy.deepcopy`? It feels like deepcopy would be faster?

krassowski (approved):
This looks good to me, thank you @smacke. I have one question before merging regarding the pickle vs deepcopy usage.

IPython/extensions/deduperreload/deduperreload.py
smacke: Hey @krassowski thanks for the review and sorry I missed this! It turns out that when it is applicable, composing `pickle.loads` and `pickle.dumps` should always be faster than `copy.deepcopy`, because the former uses a fast C implementation while the latter uses a slower Python implementation for recursively traversing objects. The drawback of pickle / unpickle is that it only works for pickle-able objects, but fortunately Python ASTs happen to be one such type of object.

This is coming from my memory so let me know if you'd prefer to see it for yourself via some benchmarks ;)

PR #14973, opened 2025-08-28
Add module ignoring functionality to debugger

ktowen:
This PR adds two new commands to ignore (and unignore) specific modules when navigating frames in the IPython debugger, similar to how the debugger already skips hidden frames. It will work with all the commands that implies moving between frames like `up`, `down`, `step` and `next`.
It is useful when you don't want the debugger to enter third party code.

Closes https://github.com/ipython/ipython/issues/1665

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -1107,6 +1107,21 @@ class Pdb(OldPdb):
                         ]
                     )
                 )
+        if self.skip and self.is_skipped_module(frame.f_globals.get("__name__", "")):
+            print(
+                self.theme.format(
+                    [
+                        (
+                            Token.ExcName,
+                            "    [... skipped 1 ignored module(s)]",
+                        ),
+                        (Token, "\n"),
+                    ]
+                )
+            )
+
+            return False
+
         return super().stop_here(frame)
 
     def do_up(self, arg):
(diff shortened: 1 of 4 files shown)


krassowski:
I think this closes https://github.com/ipython/ipython/issues/1665. The proposal there was to expose `skip` option from https://docs.python.org/3/library/bdb.html but I think this achieves similar end result.

IPython/core/debugger.py
```diff
@@ -1214,6 +1239,63 @@ def do_down(self, arg):
...
+        if self.skip is None:
+            self.skip = set()
+
+        module_name = arg.strip()
+
+        if not module_name:
+            self._show_ignored_modules()
+            return
+
+        self.skip.remove(module_name)
```
krassowski: If I call `unignore_module threading` without first calling `ignore_module threading` this would throw a long and distracting traceback. I would suggest:

```suggestion
        try:
            self.skip.remove(module_name)
        except KeyError:
            print(f"Module {module_name} is not currently ignored")
            self._show_ignored_modules()
```

krassowski (commented):
Thank you for the PR @ktowen!

This looks mostly fine, the only thing missing is tests. Can you add tests for both `ignore_module` and `unignore_module` with and without wildcard usage?

krassowski (approved):
Thank you!

PR #14994, opened 2025-09-18
Bump minimum version (spec-0) and whitespace update.

Carreau:
Note that I'm moving from listing some optional
dependencies explicitely to refering to
`ipython[optional_group]` in `pyproject.toml`.
This avoid redundancies, but I'm unsure about the
support in all tools.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,9 +2,9 @@
 
 ## Reporting a Vulnerability
 
-All IPython and Jupyter security are handled via security@ipython.org. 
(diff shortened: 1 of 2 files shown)


krassowski:
I restarted downstream macos because I think it is flaky downstream test. I will add a retry if it recurs.

Carreau:
It's pyflyby and output interleaving I believe.

krassowski (approved):

Carreau:
Thanks !

PR #14996, opened 2025-09-20
Respect `DisplayFormatter.active_types` trait configuration

### LINKED ISSUE #14991: `DisplayFormatter.active_types` trait configuration should be respected
While `DisplayFormatter.active_types` is trait-configurable:

https://github.com/ipython/ipython/blob/78d65f02c9bd62883f11905bb4977dceea387d38/IPython/core/formatters.py#L90-L110

Any user-set value gets overridden by a hard-coded value in:

https://github.com/ipython/ipython/blob/78d65f02c9bd62883f11905bb4977dceea387d38/IPython/terminal/interactiveshell.py#L766-L769

Which requires users to manually invoke `get_ipython().display_formatter.active_types = ["text/plain", "image/png"]` if they want to publish images from plots for later export in the `%notebook` magic.

Image

Darshan808:
### Description
If `DisplayFormatter.active_types` is set via config, avoid overriding it with `["text/plain"]` in `TerminalInteractiveShell`. 

This allows users to include `"image/png"` in `active_types`, ensuring that plot images are properly published for later export with the `%notebook` magic.

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -765,8 +765,13 @@ class TerminalInteractiveShell(InteractiveShell):
 
     def init_display_formatter(self):
         super(TerminalInteractiveShell, self).init_display_formatter()
-        # terminal only supports plain text
-        self.display_formatter.active_types = ["text/plain"]
+        # terminal only supports plain text if not explicitly configured
+        config = self.display_formatter._trait_values["config"]
+        if not (
+            "DisplayFormatter" in config
+            and "active_types" in config["DisplayFormatter"]
+        ):
+            self.display_formatter.active_types = ["text/plain"]
 
     def init_prompt_toolkit_cli(self):
         if self.simple_prompt:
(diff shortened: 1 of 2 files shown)


Darshan808:
Currently I'm using this to add multiple types to `DisplayFormatter.active_types`:
```python
ipython --DisplayFormatter.active_types="text/plain" --DisplayFormatter.active_types="image/png"
```
Is there an easier way to set multiple values?
@krassowski In the linked issue, I saw you tried:
```python
--DisplayFormatter.active_types='["text/plain", "image/png"]'
```
but that results in `['["text/plain", "image/png"]']` for active_types, which doesn't work.

Any suggestions? Thanks!

krassowski:
> Currently I'm using this to add multiple types to `DisplayFormatter.active_types`:

This is fine!

> Is there an easier way to set multiple values?

Array handling from CLI is a bit annoying it changed between traitlet versions. In any case usually users would add it to [`ipython_config.py`](https://ipython.readthedocs.io/en/stable/config/intro.html) like:

```python
c.DisplayFormatter.active_types = ["text/plain", "image/png"]
```

krassowski:
A test here would be useful!

krassowski (approved):
Thank you @Darshan808!

PR #14997, opened 2025-09-20
Make `%notebook` magic notarise exported notebooks (mark as trusted)

### LINKED ISSUE #14992: `%notebook` magic should notarise notebooks (mark as trusted)
Notebooks created with `%notebook` magic should be trusted on the machine they were created at to allow displaying the SVG plots immediately.

Darshan808:
### Description

Mark notebooks created with %notebook magic as trusted so SVG plots and rich outputs display immediately without manual trust approval.

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -559,6 +559,7 @@ Currently the magic system has the following functions:""",
         outfname = os.path.expanduser(args.filename)
 
         from nbformat import write, v4
+        from nbformat.sign import NotebookNotary
 
         cells = []
         hist = list(self.shell.history_manager.get_range())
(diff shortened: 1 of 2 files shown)


IPython/core/magics/basic.py
```diff
@@ -629,6 +630,9 @@ def notebook(self, s):
                 },
             },
         )
+        # Sign the notebook to make it trusted
+        notary = NotebookNotary()
```
krassowski, line 634:
```suggestion
        notary = NotebookNotary()
        notary.update_config(shell.config)
```

Without the above change it would use the default options for traitlets like `db_file`, `data_dir`, or `algorithm`.

In the test can you invoke `%config NotebookNotary.algorithm = "sha384"` and then in test verification use `NotebookNotary(algorithm="sha384")`.

image

krassowski (approved):
Thank you!

PR #14998, opened 2025-09-21
Fix matplotlib plots displaying in wrong cells during `%notebook` export

### LINKED ISSUE #14990: `%notebook` magic displays the plots published by `matplotlib` with one cell delay
All plots published from `matplotlib`, whether using `inline` or `ipympl` backend are displayed only in the subsequent cell when exporting to a notebook using `%notebook`  magic.

Naively changing `self.shell.execution_count` to `self.shell.execution_count - 1` in `DisplayPublisher.display` does lead to the correct behaviour (but needs more testing):

https://github.com/ipython/ipython/blob/78d65f02c9bd62883f11905bb4977dceea387d38/IPython/core/displaypub.py#L134-L138

Darshan808:
## Fixes #14990

**Problem:** When using `%notebook` magic to export IPython sessions, auto flushed matplotlib plots appeared one cell later than where they were created. This happened because matplotlib's `flush_figures()` runs during the `post_execute` event, after the cell's output capture phase.

**Solution:** Added execution state tracking to the display publisher:
- Track when we're in `post_execute` phase vs normal execution
- When in `post_execute`: use `execution_count - 1` (associate with originating cell)
- When in normal execution: use current `execution_count` (for explicit `plt.show()`)

**Result:** 
- Plots without `plt.show()` now appear in the correct cell
- Plots with explicit `plt.show()` continue to work correctly
- No breaking changes to existing behavior

**Testing:** Added test to verify both auto-flush and manual-flush scenarios place plots in the correct cells.

### DIFF
--- a/tests/test_display_2.py
+++ b/tests/test_display_2.py
@@ -212,6 +212,53 @@ def test_set_matplotlib_formats_kwargs():
     assert formatter_kwargs == expected
 
 
+@dec.skip_without("matplotlib")
+def test_matplotlib_positioning():
+    _ip = get_ipython()
+    _ip.history_manager.reset()
+    _ip.run_line_magic("matplotlib", "inline")
+    _ip.display_formatter.active_types = ["text/plain", "image/png"]
+
+    try:
+        _ip.execution_count = 0
+        _ip.run_cell("'no plot here'", store_history=True)
+
+        # Cell 1: No manual flush
+        _ip.execution_count = 1
+        _ip.run_cell(
+            "import matplotlib.pyplot as plt;plt.plot([0, 1])", store_history=True
+        )
+
+        _ip.execution_count = 2
+        _ip.run_cell("'no plot here'", store_history=True)
+
+        # Cell 3: Manual flush
+        _ip.execution_count = 3
+        _ip.run_cell("plt.plot([1, 0])\nplt.show()", store_history=True)
+
+        _ip.execution_count = 4
+        _ip.run_cell("'no plot here'", store_history=True)
+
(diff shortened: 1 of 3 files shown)


Darshan808:
The tests pass locally but fail in CI. Could this be because the `on_post_execute` function is running after some other functions have already executed?
Looking into this!

Darshan808:
Not sure why tests like `test_underscore_no_overwrite_user` are failing!

tests/test_display_2.py
```diff
@@ -212,6 +212,49 @@ def test_set_matplotlib_formats_kwargs():
     assert formatter_kwargs == expected
 
 
+@dec.skip_without("matplotlib")
+def test_matplotlib_positioning():
+    _ip = get_ipython()
+    _ip.history_manager.reset()
+    _ip.run_line_magic("matplotlib", "inline")
+    _ip.display_formatter.active_types = ["text/plain", "image/png"]
```
krassowski, line 227: Does this need to be reverted after test ends?

In general, I think most of the failing tests are consequence of the plots being emitted in unrelated tests AND this test fails because other tests emit plots. I would suggest to call `plt.close("all")` before `ip.history_manager.reset()`.

Darshan808:
I am not able to reproduce the failing test and I’m not sure why it is failing.  
`_` is being printed as **HELLO WORLD**, and I don’t see the reason behind it.

Darshan808:
The test passes if I remove the newly added test.

krassowski (approved):
Thank you @Darshan808!

PR #15001, opened 2025-09-24
Fix `%notebook` magic creating multiple `display_data` outputs for single widgets

Darshan808:
### Description
The `%notebook` magic was incorrectly creating separate `display_data` outputs for each MIME type from a single widget, resulting in duplicate plots and widgets in exported notebooks.
The fix collects all MIME types from a single `display_data` output into one data dictionary, ensuring that widgets with multiple representations (`text/plain`, `text/html`, `image/png`, etc.) are exported as a single `display_data` entry in the notebook.

*Before*
```json
{
  "outputs": [
    {
      "output_type": "display_data",
      "data": {"text/plain": "test"},
      "metadata": {}
    },
    {
      "output_type": "display_data", 
      "data": {"text/html": "<div>test</div>"},
      "metadata": {}
    }
  ]
}
```
*After*
```json
{
  "outputs": [
    {
      "output_type": "display_data",
      "data": {
        "text/plain": "test",
        "text/html": "<div>test</div>"
      },
      "metadata": {}
    }
  ]
}
```

This resolves the multiple plots issue and makes the exported notebook structure match the expected format where all MIME types for a single output are grouped together.

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -572,7 +572,7 @@ Currently the magic system has the following functions:""",
             cell = v4.new_code_cell(execution_count=execution_count, source=source)
 
             for output in outputs[execution_count]:
-                if output.output_type == "out_stream":
+                if output.output_type in {"out_stream", "err_stream"}:
                     text_data = []
                     for mime_type, data in output.bundle.items():
                         if isinstance(data, list):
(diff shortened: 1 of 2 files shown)


Darshan808:
The above commit demonstrates the current bug where the test fails due to multiple `display_data` outputs being generated instead of a single combined output.
```
FAILED tests/test_magic.py::test_notebook_export_single_display
- AssertionError: Expected 1 display_data output, got 2
assert 2 == 1
 +  where 2 = len([{'data': {'text/plain': ['test']}, 'metadata': {}, 'output_type': 'display_data'}, {'data': {'text/html': ['<div>test</div>']}, 'metadata': {}, 'output_type': 'display_data'}])
```

IPython/core/magics/basic.py
```diff
@@ -567,35 +567,75 @@ def notebook(self, s):
...
+                    normalized_text = []
+                    lines = full_text.split("\n")
+                    for i, line in enumerate(lines):
+                        if i < len(lines) - 1:
+                            normalized_text.append(line + "\n")
+                        elif line:
+                            normalized_text.append(line + "\n")
+                    err_output = v4.new_output("stream", text=normalized_text)
+                    err_output.name = "stderr"
+                    cell.outputs.append(err_output)
```
krassowski:
```suggestion
                if output.output_type in {"out_stream", "err_stream"}:
                    text_data = []
                    for mime_type, data in output.bundle.items():
                        if isinstance(data, list):
                            text_data.extend(data)
                        else:
                            text_data.append(data)
                    full_text = "".join(text_data)
                    # Replace literal \n with actual newlines
                    full_text = full_text.replace("\\n", "\n")
                    normalized_text = []
                    lines = full_text.split("\n")
                    for i, line in enumerate(lines):
                        if i < len(lines) - 1:
                            normalized_text.append(line + "\n")
                        elif line:  # Last line only if it's not empty
                            normalized_text.append(line + "\n")
                    stream_output = v4.new_output("stream", text=normalized_text)
                    if output.output_type == "err_stream":
                        err_output.name = "stderr"
                    cell.outputs.append(stream_output)
```

krassowski (commented):
It looks like `out_stream`/`err_stream` are almost identical, should we merge the code path to avoid accidental drift later?

krassowski (approved):
Thank you @Darshan808!

Issue #14863, opened 2025-04-03
Error tracebacks highlight variable names with a yellow background causes poor contrast in dark terminal themes in iPython 8.x


rowoflo:
In IPython 8.x, error tracebacks highlight variable names with a yellow background, which causes poor contrast in dark terminal themes.

![Image](https://github.com/user-attachments/assets/7ccd677b-9d64-4f7f-9b0f-40d19dadcc72)

This doesn’t happen in iPython 7.x.

![Image](https://github.com/user-attachments/assets/ea409b7f-3fff-4e17-9368-4cd84be9648c)

I’ve tried:
	•	c.InteractiveShell.use_stack_data = False (no effect)
	•	highlighting_overrides (doesn’t apply)
	•	Monkey-patching stack_data, but format_exc isn’t accessible anymore

Please add a config option to disable this behavior — ideally something like:

```
c.InteractiveShell.highlight_traceback_variables = False
```

Carreau:
Duplicate of #13502

```
from IPython.core.ultratb import VerboseTB

VerboseTB.tb_highlight = "" # 8.28+
VerboseTB._tb_highlight = "" # before 8.28

# VerboseTB.tb_highlight = "bg:#ccccff" # bluish
```

Carreau:
(And it's IPython upper case I)

PR #15004, opened 2025-09-27
feat: Gruvbox Dark theme with same colours in exception tracebacks

tfpf:
I hope this is allowed; the documentation does say that users are encouraged to contribute themes!

#### Gruvbox Dark Theme

This theme was available in older versions of IPython, but the exception tracebacks used colours from another theme. This change adds the theme and also uses the colours from it in tracebacks.

ipython-gruvbox-dark-demo-1

#### Other Information

* The colour used to highlight the source of the exception can make the text unreadable, but a solution for the same has been proposed in https://github.com/ipython/ipython/issues/14863#issuecomment-2780603046.
* Although all lines in `docs/source/config/details.rst` are 80 characters or shorter in length, I didn't break the line I changed because that would affect the Git blame for the lines below.
* The names of the variables used to store colour codes for the existing themes in `IPython/utils/PyColorize.py` are actually names of the colours; I used the names of the tokens instead just as a matter of semantics. If you'd like that the names of the colours be used, I can make the change.

### DIFF
--- a/IPython/utils/PyColorize.py
+++ b/IPython/utils/PyColorize.py
@@ -331,6 +331,40 @@ pridel_theme = Theme(
     symbols={"arrow_body": "\u2500", "arrow_head": "\u25b6", "top_line": "\u2500"},
 )
 
+GRUVBOX_VAL_EM = "#D79921"
+GRUVBOX_V_NAME = "#83A598"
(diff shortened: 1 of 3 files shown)


Carreau:
Seem straitforward ! Awesome to see new theme !

PR #15006, opened 2025-10-01
Fix: Restore special underscore variable after deletion.

Aniketsy:
#15000 

Underscore variable (_) not restored as last expression result after deletion when first input is a string.

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thankyou !


Aniketsy:
@krassowski  When you get a chance, could you take a look at this PR?

PR #15014, opened 2025-10-06
IPython now does not include the current working dir in sys.path, when PYTHONSAFEPATH is set or -P is used

algal:
This updates IPython so that it respects the value of the Python flag `sys.flags.safe_path`, which is usually controlled by the environment variable PYTHONSAFEPATH, and which stops Python from adding the current working directory to the `sys.path`.

So now a user can do this by relying on that standard Python env var, or Python's `-P` command line flag, rather than relying on IPython's own `--ignore_cwd` command line flag.

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -260,7 +260,7 @@ class InteractiveShellApp(Configurable):
         .. versionchanged:: X.X
             Respect sys.flags.safe_path (PYTHONSAFEPATH and -P flag)
         """
-        if '' in sys.path or self.ignore_cwd or sys.flags.safe_path:
+        if "" in sys.path or self.ignore_cwd or sys.flags.safe_path:
             return
         for idx, path in enumerate(sys.path):
             parent, last_part = os.path.split(path)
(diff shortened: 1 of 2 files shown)


IPython/core/shellapp.py
```diff
@@ -255,8 +257,10 @@ def init_path(self):
             Try to insert after the standard library, instead of first.
         .. versionchanged:: 8.0
             Allow optionally not including the current directory in sys.path
+        .. versionchanged:: X.X
```
ivanov:
```suggestion
        .. versionchanged:: 9.7
```

ivanov (approved):
Thanks for the PR!

IPython/core/shellapp.py
```diff
@@ -255,8 +257,10 @@ def init_path(self):
...
+            Respect sys.flags.safe_path (PYTHONSAFEPATH and -P flag)
         """
-        if '' in sys.path or self.ignore_cwd:
+        if '' in sys.path or self.ignore_cwd or sys.flags.safe_path:
```
ivanov:
```suggestion
        if "" in sys.path or self.ignore_cwd or sys.flags.safe_path:
```

I know you just added the trailing condition, but `darker` is asking to update the first condition, since we're touching this line.

algal:
@ivanov I committed the suggested changed to the empty str literal and to the version in the docstring.

ivanov:
Thanks for taking the time to improve IPython, @algal

PR #15020, opened 2025-10-15
Fix tab completion for subclasses of trusted classes

Darshan808:
## Solution
Added logic to check the MRO for base classes from trusted modules. Now trusts subclasses if:
- Any parent class comes from a trusted module
- The `__getattr__`  method is inherited (not overridden)

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -180,7 +180,10 @@ def _has_original_dunder_external(
             method = getattr(value_type, method_name, None)
             for base_class in value_type.__mro__[1:]:
                 base_module = getmodule(base_class)
-                if base_module and base_module.__name__ == member_type.__name__:
+                if base_module and (
+                    base_module.__name__ == member_type.__name__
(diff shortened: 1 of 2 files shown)


IPython/core/guarded_eval.py
```diff
@@ -176,6 +176,15 @@ def _has_original_dunder_external(
             member_method = getattr(member_type, method_name, None)
             if member_method == method:
                 return True
+        if isinstance(member_type, ModuleType):
+            method = getattr(value_type, method_name, None)
+            for base_class in value_type.__mro__[1:]:
+                base_module = getmodule(base_class)
+                if base_module and base_module.__name__ == member_type.__name__:
```
Darshan808: I was not sure whether to use `base_module.__name__.startswith(member_type.__name__)` or `==` like done above. But I think prefix check introduces a bug. Opened https://github.com/ipython/ipython/issues/15021

Darshan808: Modified logic to align with #15022

krassowski (approved):
Thank you @Darshan808

PR #15022, opened 2025-10-16
Prevent trusting modules with matching prefix

### LINKED ISSUE #15021: Trusted namespace check allows untrusted submodules with same prefix
### Steps to Reproduce

```python
In [1]: %config Completer.use_jedi = False
      
       # Only trust trusted namespace
In [2]: %config Completer.policy_overrides = {"allowed_getattr_external": {("trusted",),}}

In [3]: import trusted

In [4]: import trusted_fake

     # Create instance from trusted_fake (untrusted_namespace)
In [5]: test = trusted_fake.A()

In [6]:  test.<tab> # shows completions

```

### Expected Behavior

Only modules that exactly match the trusted namespace should be considered trusted.

### Cause
In the trust validation logic, we currently perform a prefix check:
https://github.com/ipython/ipython/blob/1127641a07e48e33ac5e1a4e70c84579e4dcb2ab/IPython/core/guarded_eval.py#L168-L169

Darshan808:
### Fixes #15021 

Modified module trust validation to require exact name match or proper submodule notation (with dot separator), preventing modules like `my_fake` from being trusted when only `my` is in the allowlist.

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -165,7 +165,10 @@ def _has_original_dunder_external(
             value_module = getmodule(value_type)
             if not value_module or not value_module.__name__:
                 return False
-            if value_module.__name__.startswith(member_type.__name__):
+            if (
+                value_module.__name__ == member_type.__name__
+                or value_module.__name__.startswith(member_type.__name__ + ".")
+            ):
                 return True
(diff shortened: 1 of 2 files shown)


Darshan808:
The test fails as expected in the first commit.

krassowski (approved):
Thank you!

PR #15023, opened 2025-10-16
Fix `%time` magic on windows

Darshan808:
### Possible Cause
The Windows-specific `arg_split` uses `CommandLineToArgvW` API which mangles Python expressions with quotes, turning `print("a space")` into `print("a space)`.

### Solution
Use `strict=False` when calling `arg_split` in **partial** parsing mode. This triggers the Python-based fallback splitter instead of the Windows API, correctly preserving quoted strings.

Modified `parse_argstring` to pass `strict=not partial` to `arg_split`


Darshan808:
The test fails as expected in windows runner.

krassowski (approved):
Thank you @Darshan808!

PR #15027, opened 2025-10-19
Infer type from return value and improve attribute completions

Darshan808:
### Description
With:
```
%config Completer.use_jedi=True
```
Enables completions for class methods without return type annotations by evaluating the property body and inferring the type from the actual return value.
Also enables completions for instance attributes assigned in class methods.

image

image

image

image

image

- [x] Added tests

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from copy import copy
 from inspect import isclass, signature, Signature, getmodule
 from typing import (
@@ -396,6 +397,8 @@ class EvaluationContext:
     policy_overrides: dict = field(default_factory=dict)
     #: Transient local namespace used to store mocks
     transient_locals: dict = field(default_factory=dict)
+    #: Tranisents of class level
+    class_transients: dict | None = None
 
     def replace(self, /, **changes):
         """Return a new copy of the context, with specified changes"""
@@ -560,6 +563,7 @@ def _validate_policy_overrides(
 def _handle_assign(node: ast.Assign, context: EvaluationContext):
     value = eval_node(node.value, context)
     transient_locals = context.transient_locals
+    class_transients = getattr(context, "class_transients", None)
     for target in node.targets:
         if isinstance(target, (ast.Tuple, ast.List)):
             # Handle unpacking assignment
@@ -573,6 +577,14 @@ def _handle_assign(node: ast.Assign, context: EvaluationContext):
             # Before starred
             for i in range(star_or_last_idx):
                 transient_locals[targets[i].id] = values[i]
+                # Check for self.x assignment
+                if class_transients is not None and hasattr(targets[i], "ctx"):
+                    if (
+                        isinstance(targets[i], ast.Attribute)
+                        and isinstance(targets[i].value, ast.Name)
+                        and targets[i].value.id == "self"
+                    ):
+                        class_transients[targets[i].attr] = values[i]
 
             # Starred if exists
             if starred:
@@ -580,14 +592,40 @@ def _handle_assign(node: ast.Assign, context: EvaluationContext):
                 transient_locals[targets[star_or_last_idx].value.id] = values[
                     star_or_last_idx:end
                 ]
+                if (
+                    class_transients is not None
+                    and isinstance(targets[star_or_last_idx], ast.Attribute)
+                    and isinstance(targets[star_or_last_idx].value, ast.Name)
+                    and targets[star_or_last_idx].value.id == "self"
+                ):
+                    class_transients[targets[star_or_last_idx].attr] = values[
+                        star_or_last_idx:end
+                    ]
 
                 # After starred
                 for i in range(star_or_last_idx + 1, len(targets)):
                     transient_locals[targets[i].id] = values[
                         len(values) - (len(targets) - i)
                     ]
+                    if (
+                        class_transients is not None
+                        and isinstance(targets[i], ast.Attribute)
+                        and isinstance(targets[i].value, ast.Name)
+                        and targets[i].value.id == "self"
+                    ):
+                        class_transients[targets[i].attr] = values[
+                            len(values) - (len(targets) - i)
+                        ]
         else:
-            transient_locals[target.id] = value
+            if (
+                isinstance(target, ast.Attribute)
+                and isinstance(target.value, ast.Name)
+                and target.value.id == "self"
+            ):
+                if class_transients is not None:
+                    class_transients[target.attr] = value
+            elif hasattr(target, "id"):
+                transient_locals[target.id] = value
     return None
 
 
@@ -643,6 +681,10 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
         return result
     if isinstance(node, ast.FunctionDef):
         # we ignore body and only extract the return type
+        func_locals = context.transient_locals.copy()
+        func_context = context.replace(transient_locals=func_locals)
+        for child_node in node.body:
+            eval_node(child_node, func_context)
         is_property = False
 
         for decorator_node in node.decorator_list:
@@ -662,7 +704,7 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
                     return_type, context
                 )
             else:
-                return_value = _infer_return_value(node, context)
+                return_value = _infer_return_value(node, func_context)
                 context.transient_locals[node.name] = return_value
 
             return None
@@ -673,7 +715,7 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
         if return_type is not None:
             dummy_function.__annotations__["return"] = return_type
         else:
-            inferred_return = _infer_return_value(node, context)
+            inferred_return = _infer_return_value(node, func_context)
             if inferred_return is not None:
                 dummy_function.__inferred_return__ = inferred_return
 
@@ -685,6 +727,7 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
         # TODO support class decorators?
         class_locals = context.transient_locals.copy()
         class_context = context.replace(transient_locals=class_locals)
+        class_context.class_transients = class_locals
         for child_node in node.body:
             eval_node(child_node, class_context)
         bases = tuple([eval_node(base, context) for base in node.bases])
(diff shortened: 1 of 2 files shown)


krassowski:
Do we need to create a duck? My instinct would be that we can return the real value (if it was allowed to be evaluated so far).

Can we also resolve the trivial case of a variable which is in locals or transient locals being returned?

Darshan808:
> Can we also resolve the trivial case of a variable which is in locals or transient locals being returned?

This is addressed by #15030. Once that gets merged, we can update this branch (it may introduce some merge conflicts).
For now, I’ll add the currently failing test here.

IPython/core/guarded_eval.py
```diff
@@ -658,22 +658,33 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
...
+                dummy_function.__annotations__["return"] = inferred_type
+
         dummy_function.__name__ = node.name
         dummy_function.__node__ = node
         context.transient_locals[node.name] = dummy_function
         return None
     if isinstance(node, ast.ClassDef):
         # TODO support class decorators?
-        class_locals = {}
+        class_locals = context.transient_locals.copy()
```
Darshan808: Since `transient_locals` were not included in `class_locals`, completions for `transient_locals` defined outside the class were failing.
```python
In [1]: li = []
   ...: class MyClass2:
   ...:     def test(self):
   ...:         return li
   ...: 
   ...: c = MyClass2()
   ...: c.test().<tab>
```
Including `transient_locals` in `class_locals` resolves this issue.

Darshan808:
<strike>Once that gets merged, we can update this branch</strike>
I believe this PR is now complete on its own.

IPython/core/guarded_eval.py
```diff
@@ -853,6 +864,22 @@ def dummy_function(*args, **kwargs):
     return None
 
 
+def _infer_return_value(node: ast.FunctionDef, context: EvaluationContext):
+    """Execute the function body to infer its return value."""
+
+    for stmt in node.body:
+        if isinstance(stmt, ast.Return):
```
krassowski: This works fine for a trivial case of a single return. I wonder if we should handle some more complex cases too, say:

```python
def string_or_number(flag):
    if flag:
        return "Hello"
    return 42
```

Then we would expect completions to return a union of str and int attributes. There is already a code creating unions.

I think the logic should be:
- collect all `return` statements
- filter out those which evaluate to `None` or are `NOT_EVALUATED`
- if only one `return` remains use the actual value
- if there is more than one meaningful `return`, create a union using a duck
   - we could possibly skip creating a duck if the type is the same, say we saw `return 1` and `return 2`, but if we saw `return {"a": 1}` and `return {"b": 2}` then we would want to use a duck to collect both keys.

tests/test_completer.py
```diff
@@ -2077,6 +2077,30 @@ def _(expected):
...
             "capitalize",
         ],
+        [
+            "\n".join(
+                [
+                    "class NotYetDefined:",
+                    "    def my_method(self):",
+                    "        return []",
+                    "my_instance = NotYetDefined()",
+                    "my_instance.my_method().",
```
krassowski, line 2087: Does it already work with `@property` too?

krassowski (commented):
🚀

tests/test_completer.py
Darshan808: Yes it does.

Darshan808:
Found a case where completion still fails:
```python
In [1]: def foo():
   ...:     li = []
   ...:     return li
   ...: foo().
```
Because we don't parse function body.
Working on this. But I might push this feature in #15030

Darshan808:
> <strike>Found a case where completion still fails:</strike>

This case now passes!

tests/test_completer.py
krassowski: Let's add a test case then :)

Darshan808:
#15030 and this PR are closely related. Since the logic was interconnected, I ended up implementing the enhancement from #15030 here as well (including migrating the related tests).

I’d appreciate a review of my approach.

Because `self.attr` assignments are not limited to the current function scope but belong to the class’s local context (regardless of how deeply nested they are), I added a new attribute to the `EvaluationContext` to maintain `class_locals` in the current context no matter where we are within the class. And on each `self.attr` assignment within the class we update the class locals accordingly.

This solution appears to address the issue, but I’d like feedback on whether this approach makes sense.
@krassowski

IPython/core/guarded_eval.py
```diff
@@ -853,6 +915,69 @@ def dummy_function(*args, **kwargs):
...
+    types = {type(v) for v in return_values}
+    if len(types) == 1:
+        t = next(iter(types))
+        if t is dict:
+            keys = set()
+            for v in return_values:
+                keys.update(v.keys())
+            return _Duck(
+                attributes=dict.fromkeys(dir(dict())), items={k: None for k in keys}
+            )
```
krassowski: The value of items should be preserved to, so that further chained completion can be performed.

IPython/core/guarded_eval.py
```diff
@@ -573,21 +577,55 @@ def _handle_assign(node: ast.Assign, context: EvaluationContext):
...
                 end = len(values) - (len(targets) - star_or_last_idx - 1)
                 transient_locals[targets[star_or_last_idx].value.id] = values[
                     star_or_last_idx:end
                 ]
+                if (
+                    class_transients is not None
+                    and isinstance(targets[star_or_last_idx], ast.Attribute)
+                    and isinstance(targets[star_or_last_idx].value, ast.Name)
+                    and targets[star_or_last_idx].value.id == "self"
+                ):
```
krassowski: - This looks quite repetitive, I would create a function to check this condition
- The class instance argument can have a name other than "self". I think we could extract it from the function signature and store the name on the context when entering function definition. Maybe this should be addressed in a separate PR, but in any case this shows the case for extracting this check logic into a single place.

IPython/core/guarded_eval.py
krassowski: Maybe we need a test for it.

Darshan808:
I believe this covers all the cases I can think of right now.

Darshan808:
Updated title and PR message.
I believe we can close #15030, once this gets merged.

IPython/core/guarded_eval.py
```diff
@@ -396,6 +396,10 @@ class EvaluationContext:
     policy_overrides: dict = field(default_factory=dict)
     #: Transient local namespace used to store mocks
     transient_locals: dict = field(default_factory=dict)
+    #: Tranisents of class level
```
krassowski:
```suggestion
    #: Transients of class level
```

IPython/core/guarded_eval.py
```diff
@@ -654,27 +693,50 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
...
             if decorator is property:
                 is_property = True
+            elif decorator is staticmethod:
+                is_static = True
+            elif decorator is classmethod:
+                is_classmethod = True
+
+        if func_context.class_transients is not None:
+            if not is_static and not is_classmethod:
+                func_context.instance_arg_name = node.args.args[0].arg
```
krassowski: Completion breaks if an instance method is defined incorrectly:

Image

Works fine if there is an argument:

Image

Since only one incorrectly defined method breaks completion for the entire class I suggest we first check if there are any arguments here. Let's also add a test case.

IPython/core/guarded_eval.py
```diff
@@ -853,6 +924,81 @@ def dummy_function(*args, **kwargs):
...
+
+    if dict in types:
+        key_values = {}
+        for v in values:
+            if type(v) is not dict:
+                continue
+            for k, val in v.items():
+                key_values.setdefault(k, []).append(val)
+
+        merged_items = {k: _merge_values(vals) for k, vals in key_values.items()}
```
krassowski: This is a good start.

There will be more cases when we want to merge items, e.g. subclasses of `dict` such as `defaultdict`. In general, these will be all instances of classes which implement `__getitem__` and either `.items()` or `keys()`. We should then check if we are allowed to call `items()` or `keys()`. If they only implement `keys()` then we would set value to `None`.

A good example would be `pandas.DataFrame` since we want to be able to complete column names. Maybe we should merge this block into lines 952-9544 where we collect attributes.

IPython/core/guarded_eval.py
```diff
@@ -853,6 +924,81 @@ def dummy_function(*args, **kwargs):
...
+    if len(types) == 1:
+        t = next(iter(types))
+        if t is not dict:
+            if t in (list, set, tuple):
+                return t()
+            return values[0]
+
+    attributes = set()
+    for v in values:
+        attributes.update(dir(v))
```
krassowski: We may need to use `policy.can_call(v, v.__dir__)` to prevent side effects here, but I don't won't it to end up being too restrictive... The default `object.__dir__` should be allowed and I think it currently is not on the allow-list. This would need a test where we check that a side-effect in a custom `__dir__` does not get invoked.

IPython/core/guarded_eval.py
Darshan808: Just discovered that this has some serious side effects!
```python
a = 1
class B:
    pass

B().<tab> # a is suggested in completion
```

Darshan808: Solved this by storing `transient_locals` into `locals` when entering a class.

Darshan808:
To check for `defaultdict` case, I discovered https://github.com/ipython/ipython/issues/15039 and added a fix for it in <strike>this PR itself.</strike> https://github.com/ipython/ipython/pull/15042

IPython/core/guarded_eval.py
```diff
@@ -853,6 +959,100 @@ def dummy_function(*args, **kwargs):
...
+def _merge_values(values, policy: EvaluationPolicy):
+    """Recursively merge multiple values, combining attributes and dict items."""
+    if len(values) == 1:
+        return values[0]
+
+    types = {type(v) for v in values}
+    merged_items = None
+    key_values = {}
+    for v in values:
+        if hasattr(v, "items") and policy.can_get_item(v, None):
```
Darshan808: Using `hasattr` might trigger side effects. What’s the safest way to check whether an object has a specific method before attempting to call it?

For example, is this approach reasonable:
```python
if policy.can_get_attr(v, "items") and hasattr(v, "items") and policy.can_get_item(v, None):
```

Darshan808:
It still needs some tests, but I think it’s ready for a review first.

IPython/core/guarded_eval.py
```diff
@@ -1277,6 +1305,7 @@ def _list_methods(cls, source=None):
     collections.Counter.most_common,
     object.__dir__,
     type.__dir__,
+    dict.__setitem__,
```
krassowski: But this would mutate the dictionary, wouldn't it? I think we do not want to add this one to allowed calls...

IPython/core/guarded_eval.py
```diff
@@ -607,11 +608,38 @@ def _handle_assign(node: ast.Assign, context: EvaluationContext):
...
+                if container is None:
+                    container = context.globals.get(name)
+                if container is None:
+                    raise NameError(
+                        f"{name} not found in locals, globals, nor builtins"
+                    )
+
+                key = eval_node(target.slice, context)
+                if policy.can_call(container.__setitem__):
+                    container[key] = value
```
krassowski: We should make a copy of the container instead. Except that this would be rather inefficient for pandas DataFrame. I need to think more about this, maybe we could convert it to a duck and add an extra item?

Darshan808:
To avoid adding more enhancements in this PR, I opened https://github.com/ipython/ipython/pull/15042 to specifically focus on https://github.com/ipython/ipython/issues/15039.

krassowski (approved):
Thank you @Darshan808!

PR #15029, opened 2025-10-19
fixed AssertionError while rendering traceback in ipython 9.6.0

sonianuj287:
Issue - [https://github.com/ipython/ipython/issues/15024](https://github.com/ipython/ipython/issues/15024)

## Resolution - 
Restored safe coercion of SyntaxError.text so tracebacks now decode bytes via py3compat.cast_unicode, fall back cleanly to str, and avoid the AssertionError path identified in gh-15024 (ultratb.py). Added a regression in test_ultratb.py that triggers xml.etree.ElementTree.ParseError and confirms we no longer emit “Internal Python error”.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -336,13 +336,18 @@ class ListTB(TBTools):
                     )
                 )
                 if textline == "":
-                    # sep 2025:
-                    # textline = py3compat.cast_unicode(value.text, "utf-8")
-                    if value.text is None:
+                    # Gracefully coerce SyntaxError.text into a printable str.
+                    value_text = getattr(value, "text", None)
+                    if value_text is None:
                         textline = ""
                     else:
-                        assert isinstance(value.text, str)
-                        textline = value.text
+                        try:
+                            textline = py3compat.cast_unicode(value_text, "utf-8")
+                        except Exception:
+                            # As a last resort, fall back to ``str`` for exotic objects.
+                            textline = str(value_text)
+                        if textline is None:
+                            textline = ""
 
                 if textline is not None:
                     i = 0
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -342,7 +342,11 @@ class ListTB(TBTools):
                         textline = ""
                     else:
                         try:
-                            textline = py3compat.cast_unicode(value_text, "utf-8")
+                            # Handle bytes by decoding to utf-8, fallback to str() for other types
+                            if isinstance(value_text, bytes):
+                                textline = value_text.decode("utf-8", "replace")
+                            else:
+                                textline = str(value_text)
                         except Exception:
                             # As a last resort, fall back to ``str`` for exotic objects.
                             textline = str(value_text)
--- a/tests/test_ultratb.py
+++ b/tests/test_ultratb.py
@@ -264,6 +264,12 @@ bar_syntax_error_test_2()
             with tt.AssertPrints("QWERTY"):
                 ip.showsyntaxerror()
 
+    def test_parseerror_non_string_text(self):
+        cell = "from xml.etree import ElementTree\nElementTree.fromstring('hello')"
+        with tt.AssertNotPrints("Internal Python error", channel="stderr"):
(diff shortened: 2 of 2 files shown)


sonianuj287:
@pbx @scw @lebigot @blueyed

bennuttall (approved):
I've tested that this resolves the issue

sonianuj287:
> I've tested that this resolves the issue

Thank you so much.
Whom should I mention to move this PR for merging..

Carreau:
Wasn't #15012 sufficient ? I'd like to avoid catching overly too many errors if we can.

IPython/core/ultratb.py
```diff
@@ -336,13 +336,18 @@ def _format_exception_only(
...
-                    if value.text is None:
+                    # Gracefully coerce SyntaxError.text into a printable str.
+                    value_text = getattr(value, "text", None)
+                    if value_text is None:
                         textline = ""
                     else:
-                        assert isinstance(value.text, str)
-                        textline = value.text
+                        try:
+                            textline = py3compat.cast_unicode(value_text, "utf-8")
```
Carreau: and really I'm trying to _remove_ py3compat, It's from the Python2 days, it shoudl be gone by now, so I'd like to not re-introduce it.

sonianuj287: Thanks for pointing. please review the changes.

sonianuj287:
> Wasn't #15012 sufficient ? I'd like to avoid catching overly too many errors if we can.

wasn't sure about it. I'm looking for contribution for hacktoberfest. Please let me know for any improvement I can do here. Thanks:)

sonianuj287:
Hi @Carreau @minrk @takluyver @jdfreder , please review this PR and suggest me changes. Thanks :)

IPython/core/ultratb.py
```diff
@@ -336,13 +336,22 @@ def _format_exception_only(
...
                         textline = ""
                     else:
-                        assert isinstance(value.text, str)
-                        textline = value.text
+                        try:
+                            # Handle bytes by decoding to utf-8, fallback to str() for other types
+                            if isinstance(value_text, bytes):
+                                textline = value_text.decode("utf-8", "replace")
+                            else:
+                                textline = str(value_text)
```
Carreau, line 349: That is weird... if value_text is a str, what's the point of `str()` of a str ? 
And "utf-8", replace is already a way to avoid most errors...

IPython/core/ultratb.py
```diff
@@ -336,13 +336,22 @@ def _format_exception_only(
...
+                        except Exception:
+                            # As a last resort, fall back to ``str`` for exotic objects.
+                            textline = str(value_text)
```
Carreau, line 352: SO... if `textline = str(value_text)` fails on line 349, you do `textline = str(value_text)` ... Do you see the problem ?

Carreau (changes requested):
I know you you are in a hurry for hacktoberfest, but please you are not the most pressing PR, and please don't ping people, especially those who are not active on this repo anymore. 

- You have not checked that the bug is fixed on main
- You have not check that your added test fails on main – that the minimum at least create a test that fails – so it's not really a regression test.

Maintaining open source code is not like competitive coding, it is writing code that is meant to work, and be maintained for a long time, not throwing anything that stick to the wall.

Please respect other people time; and properly code when you are contributing.

By principle, this won't be merged before end of october.

IPython/core/ultratb.py
```diff
@@ -336,13 +336,22 @@ def _format_exception_only(
...
+                        if textline is None:
```
Carreau, line 353: I'm not sure how textline can be None, you are already in a if textline == ""

Yo need to review your logic or explain how it can be None. 

`str(..)` cannot:

```
In [1]: class Foo:
   ...:     def __str__(self):
   ...:         return None
   ...:

In [2]: str(Foo())
Cell In[2], line 1
----> 1 str(Foo())

TypeError: __str__ returned non-string (type NoneType)
```


I don't think .decode() can.

Carreau:
Closing, this is not up to IPython's standard.

Thanks.

PR #15030, opened 2025-10-19
Infer type for locals and attributes defined in `__init__()`

Darshan808:
### Description
With:
```
%config Completer.use_jedi=True
```
#### Enables completions for locals

image

#### Enables completions for instance attributes assigned in `__init__()`.

image

image

- [x] Added tests

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -677,6 +677,10 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
         class_context = context.replace(transient_locals=class_locals)
         for child_node in node.body:
             eval_node(child_node, class_context)
+        # extract self.attribute assignments
+        init_attributes = _extract_init_attributes(node, class_context)
+        # Merge init attributes into class_locals
+        class_locals.update(init_attributes)
         bases = tuple([eval_node(base, context) for base in node.bases])
         dummy_class = type(node.name, bases, class_locals)
         context.transient_locals[node.name] = dummy_class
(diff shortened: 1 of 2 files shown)


IPython/core/guarded_eval.py
```diff
@@ -853,6 +858,116 @@ def dummy_function(*args, **kwargs):
...
+    init_method = None
+    for node in class_node.body:
+        if isinstance(node, ast.FunctionDef) and node.name == "__init__":
+            init_method = node
+            break
+
+    if not init_method:
+        return attributes
+
+    for stmt in init_method.body:
```
krassowski, line 882: I think we may not need `_extract_init_attributes` no `_extract_variables_from_module`. In particular we may _not_ need to:
- iterate over the class body for the second time (instead we would just handle that in `eval_node → func` case); we already have `transient_locals` which represent the locals of a class. `transient_locals` were a quick fix but a better implementation would be tracking scopes explicitly and each time we enter a class or function definition we would push a new scope onto a list, each time we exit we would pop it; the difference is that it would allow to correctly access variables defined in parent scopes.
- look for `__init__` specifically

krassowski: > each time we enter a class or function definition we would push a new scope onto a list, each time we exit we would pop it;

which is kind of trivial since we are calling `eval_node` recursively, this gets handled by itself if we are making copies of context when entering `eval_node` for class elements

krassowski:
Closing as superseded by https://github.com/ipython/ipython/pull/15027

PR #15032, opened 2025-10-21
Adding check if file not found and cleanups (#14987)

### LINKED ISSUE #14987: Can IPython print a warning if unable to open ipython_config.py?
I have `~/.ipython/profile_default/ipython_config.py` symlinked to somewhere else. Sometimes when I move directories around I break the symlink, which causes random regressions in my IPython set-up. It would be nice if IPython printed a message on start-up if it was unable to open the `ipython_config.py` file, so I can more quickly diagnose the problem.

aijadugar:
what this PR do?

finds the file path, 
if not found then prints warning.

tested with test_application.py file 

clean-up code with better logic.

Fixes #14987

### DIFF
--- a/IPython/core/application.py
+++ b/IPython/core/application.py
@@ -343,18 +343,31 @@ class BaseIPythonApplication(Application):
             if suppress_errors is not None:
                 old_value = Application.raise_config_file_errors
                 Application.raise_config_file_errors = not suppress_errors
-            Application.load_config_file(
-                self,
+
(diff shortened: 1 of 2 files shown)


Carreau:
If anything I think the logic should go up in traitlets; and we should carefully think about the warnings; I think it's ok to not have any config file, but warn if one if unreadable. I need to have a bit more time to wrap my head arround this code.

PR #15036, opened 2025-10-22
Suppress file completions in context of attributes/methods

### LINKED ISSUE #15031: Suppress paths tab completion when completing a method/attribute
- Path completions are great in strings!
- Path completions can be also somewhat useful if you want to name variables based on the names of files/folders in the current directory
- Path completions are not very useful when completing code and often get in the way; this is especially true:
   - when current directory contains many hidden files (starting with dot using Unix convention)
   - when using plain IPython rather than Jupyter UI because then the type of suggestion is not shown which makes it confusing

Image

Image

Darshan808:
### Description
Suppress `file_matcher` when matching `attribute` \\ `methods`.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -2749,7 +2749,7 @@ class IPCompleter(Completer):
                 return matches
             except NameError:
                 # catches <undefined attributes>.<tab>
-                matches = SimpleMatcherResult(completions=[], suppress=False)
+                return SimpleMatcherResult(completions=[], suppress=False)
         else:
             try:
                 matches = self.global_matches(context.token, context=context)
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -2209,6 +2209,15 @@ class IPCompleter(Completer):
         """
         # TODO: add a heuristic for suppressing (e.g. if it has OS-specific delimiter,
         #  starts with `/home/`, `C:\`, etc)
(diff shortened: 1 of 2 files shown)


IPython/core/completer.py
```diff
@@ -2209,6 +2209,15 @@ def file_matcher(self, context: CompletionContext) -> SimpleMatcherResult:
...
         #  starts with `/home/`, `C:\`, etc)
+        last_line = context.full_text.split("\n")[-1]
+        text_before_token = last_line[: last_line.rfind(context.token)]
+
+        # Suppress path completion when completing methods/attributes
+        if text_before_token.endswith((")", "]")):
+            return {
+                "completions": [],
+                "suppress": True,
+            }
```
krassowski: It works the other way: this would suppress all non-file matchers. Rather than duplicate the logic for context detection (i.e. is it an attribute completion) in `file_matcher` I would suggest we add `"suppress": {"file_matcher"}` in the attribute matcher.

krassowski (commented):
This seems to break a couple of useful completions. For example, if there is a directory named `notes`:
- `!ls n<tab>`
- `%ls n<tab>`

| Before | With this PR |
|--|--|
| image | image |

I think we need to be more selective and only set `suppress` if we have matches for attributes/methods.

krassowski (approved):
Let's try this, thank you!

PR #15037, opened 2025-10-24
Fix some ruff issues with import (rule UP35 not activated yet)

Carreau:


Carreau:
Just some coverage issue, merging.

PR #15040, opened 2025-10-25
TST/DEP: add missing test dependency on setuptools

neutrinoceros:
setuptools is directly imported in some tests, but not declared as a requirement.
A test env created with, e.g. `uv`, doesn't include setuptools by default. This fixes it.


neutrinoceros:
closing in favor of #15041

PR #15041, opened 2025-10-25
DEP: update and add missing lower bounds on direct dependencies

neutrinoceros:
Include changes from #15040
This is a best-effort attempt to provide/update lower bounds on direct dependencies in two groups: hard requirements, and test dependencies.
At the time of opening, I still see 7 test failing locally with an env obtained with `uv sync --resolution=lowest-direct` which may indicate that some updates are still needed (although it is not clear what dependencies are causing the failures). It could also be a symptom of *transitive* dependencies incompatibilities. In any case, this is still much better than no explicit lower bounds at all.
The end goal is to enable testing downstream with `--resolution=lowest`.
To guide reviewers I've included small comments to justify my choice of lower bounds. I simply commented "wheel" in places where I picked the oldest version that has a compatible wheel on PyPI (in some cases, older versions might still build from source, but relying on much more recent versions of setuptools is more or less bound to break at some point).

If you're interested, I'd be happy to also setup a CI job to systematically check that lower bounds are sound and up to date.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,15 +22,15 @@ classifiers = [
 ]
 requires-python = ">=3.11"
 dependencies = [
-    'colorama; sys_platform == "win32"',
-    "decorator",
-    "ipython-pygments-lexers",
-    "jedi>=0.16",
-    "matplotlib-inline",
+    'colorama>=0.4.4; sys_platform == "win32"', # lower bound is pure guess (not checked)
+    "decorator>=4.3.2", # wheel
+    "ipython-pygments-lexers>=1.0.0",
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, I've resolve some conflicts.

I tend to be ok with being fairly aggressive with minimum dependencies as IPython follows spec-0 (https://scientific-python.org/specs/spec-0000/)

If you think any other test matrix elements are needed, feel free to send a PR.

neutrinoceros:
> If you think any other test matrix elements are needed, feel free to send a PR.

will do once this one goes in. Thanks !

PR #15048, opened 2025-10-29
Support completions for lambdas and `async` functions

Darshan808:
### Description
Handle `lambda` and `async` functions in `guarded_eval` to enhance completions.

### DIFF
--- a/tests/test_guarded_eval.py
+++ b/tests/test_guarded_eval.py
@@ -552,6 +552,16 @@ def test_mocks_items_of_call_results(data, code, expected_items):
             ),
             str,
         ],
+        [
+            "\n".join(
+                [
+                    "async def async_func():",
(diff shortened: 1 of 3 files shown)


tests/test_guarded_eval.py
```diff
@@ -552,6 +552,25 @@ def test_mocks_items_of_call_results(data, code, expected_items):
             ),
             str,
         ],
+        [
+            "\n".join(
+                [
+                    "async def async_func():",
+                    "    return []",
+                    "(await async_func())",
```
krassowski, line 560: We should also handle the `async_func().<tab>` case. I think we could create a duck based of attributes of `types.CoroutineType`.

krassowski (approved):
Thank you!

Issue #14438, opened 2024-05-15
How to change `IPython.terminal.shortcuts.newline_autoindent` to 2 extra spaces?


sci-42ver:
In the [doc](https://ipython.readthedocs.io/en/stable/api/generated/IPython.terminal.shortcuts.html#IPython.terminal.shortcuts.newline_autoindent), it says
> That is to say, indent by _4 extra space_ after a function definition, class definition, context manager

Then in `ipython`, for example when I write one `for` loop when pressing enter after `for ...:`, then 4 extra spaces will be inserted. Is there one way to change this behavior to 2 spaces?

Darshan808:
As of today (IPython 9.6.0), IPython doesn't provide a built-in configuration option to change the indent width from 4 to 2 spaces. The 4-space indentation is hardcoded in the way IPython handles Python's indentation rules.
https://github.com/ipython/ipython/blob/089818d7518edbf5a38e39a8a37abb518a9202ec/IPython/core/inputtransformer2.py#L757-L758

PR #15049, opened 2025-10-29
Change newline_autoindent default extra indent to 2 spaces

ira-po:
#14438
Changes:
Modified IPython/terminal/shortcuts/__init__.py to set default extra indent to 2 spaces.

### DIFF
--- a/change_indent.py
+++ b/change_indent.py
@@ -1,0 +1,12 @@
+import fileinput
+
+
+file_path = "IPython/terminal/shortcuts/auto_suggest.py"
+
+
+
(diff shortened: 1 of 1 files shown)


Darshan808:
Thanks @ira-po!
IPython doesn't provide a built-in configuration option to change the indent width from 4 to 2 spaces. The 4-space indentation is hardcoded in the way IPython handles Python's indentation rules.
https://github.com/ipython/ipython/blob/089818d7518edbf5a38e39a8a37abb518a9202ec/IPython/core/inputtransformer2.py#L757-L758

So, this PR is not effective in accomplishing the desired effect.

PR #15051, opened 2025-10-31
BLD: adjust license spec to PEP 639

neutrinoceros:
license "OSI Approved" classifiers are deprecated, here's the new standard.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,4 @@
 include README.rst
-include COPYING.rst
-include LICENSE
 include setupbase.py
 include _build_meta.py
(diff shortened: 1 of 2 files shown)


neutrinoceros:
for reference, this resolves the following warning currently seen in CI:
```
********************************************************************************
By 2026-Feb-18, you need to update your project and remove deprecated calls
or your builds will no longer be supported.
See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-files for details.
********************************************************************************
```

Carreau:
Thanks.

PR #15054, opened 2025-10-31
TST: add systematic backward compatibility checks against oldest supported versions of direct dependencies

neutrinoceros:
Follow up to #15041

Since this setup requires uv (pip install doesn't support the `--resolution` argument, crucial to this work), I've written it as a completely separate job instead of a matrix item. It would be possible to reconcile the two, but that would make for a much more invasive change where I'd replace pip with uv everywhere. Of course, if that's desired, I'm also happy to go all-in with uv, either here, or as a follow up PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,3 +107,47 @@ jobs:
       with:
          name: Test
          files: /home/runner/work/ipython/ipython/coverage.xml
+
+  oldest-deps:
+    # pro-actively check backward compatibility
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 15
+    # Disable scheduled CI runs on forks
+    if: github.event_name != 'schedule' || github.repository_owner == 'ipython'
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-22.04
+          # include windows because of platform-specific direct dependencies
+          - windows-2022
+
+    steps:
+    - uses: actions/checkout@v5
+    - name: Set up uv with Python 3.11
+      uses: astral-sh/setup-uv@v7
+      with:
+        python-version: '3.11'
+        enable-cache: true
+        activate-environment: true
+        prune-cache: false
+        cache-dependency-glob: |
+          pyproject.toml
+
+    - name: Install Python dependencies (oldest supported versions)
+      run: uv pip install --resolution=lowest-direct -e .[test]
+
+    - name: Try building with uv build
+      if: runner.os != 'Windows'  # setup.py does not support sdist on Windows
+      run: |
+        uv build
+        shasum -a 256 dist/*
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -105,3 +105,47 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-22.04
+          # include windows because of platform-specific direct dependencies
+          - windows-2022
+
+    steps:
+    - uses: actions/checkout@v5
+    - name: Set up uv with Ptyoh 3.11
```
Carreau:
```suggestion
    - name: Set up uv with Python 3.11
```

neutrinoceros:
I got the number of failed tests from 2 to 1 by bumping the minimum requirement on jedi, but I don't get what's happening with the remaining failure yet, so I don't have a good idea of *which* other direct dependency needs bumping to resolve it. Of course I could just try them all but that'd be wasteful, so in case it might be obvious to others, I'm open to suggestions !

Carreau:
@krassowski ?

neutrinoceros:
@Carreau just curious, did you actually request a review from Copilot, or is GH's UI just pretending you did ? (I heard such reviews could appear un-prompted)

Carreau:
I tried clicking on it to see if it could find or suggest the root cause of the test failure, but it just summarized the PR...

So yes the noise was due to me, sorry about that.

krassowski:
> @krassowski ?

Did you ping me for the completion XFAILs? These are expected. I see a real issue with `tests.test_iplib.doctest_tb_sysexit_verbose_stack_data_05` not sure about it.

neutrinoceros:
It was about that test yes. Do you have any idea what direct dependency would be affect the outcome ?

krassowski:
Well, the failure looks like could be due to formatting of cPython copyright message so maybe `python-version: 3.11`? I don't know what that test is meant to check but it looks like it could be more selective? Also, 3.11 support window ended as per https://scientific-python.org/specs/spec-0000/ should this PR use 3.12?

neutrinoceros:
This is supposed to check the viability of an environment barely matching minimal requirements as declared in pyproject.toml
Dropping support for Python 3.11 is not in scope with what I'm trying to achieve here.

neutrinoceros:
The fact that the job is currently failing necessarily indicates that at least one minimal requirement needs to be updated. I just don't know where to look first.

krassowski:
Well, I would bump python-version first :P

neutrinoceros:
After a couple iterations, I found, empirically, that bumping `stack-data` resolved the failure. Unless something else breaks, I expect this should be ready for review now.

neutrinoceros:
FTR; bumping Python to 3.12 wouldn't have solved this failure.

PR #15056, opened 2025-10-31
Completion in loops and conditionals

### LINKED ISSUE #15050: Static inference for completion in loops
> Hmm it looks like things inside loops do not work well either.
> 
> For full `async` support we would potentially want to support:
> 
> ```python
> import asyncio
> 
> async def async_range():
>     for i in range(10):
>         yield i
>         await asyncio.sleep(0.1)
> 
> async for i in async_range(10):
>     i.<tab>
> ```
> 
> but currently completions on objects defined in `for` loops is not implemented:
> 
> 
> ```python
> for i in range(10):
>     i.<tab>
> ```
> 
> And surprisingly even this does not work, even though it should:
> 
> ```python
> i = 10
> while True:
>     i.<tab>
> ```
> 
>  

 _Originally posted by @krassowski in [#15045](https://github.com/ipython/ipython/issues/15045#issuecomment-3463094952)_

This would be a follow-up to https://github.com/ipython/ipython/pull/14993.

Darshan808:
### Description
Adds blocks of code to handle `ast.While`, `ast.For` and `ast.If` with test cases.


Darshan808:
Few queries!
I’ve added `_Duck.__dir__` to `ALLOWED_CALLS` since we need it to access attributes for merging values. Given that `_Duck` is a controlled, internally created class, there’s no security risk in allowing this call, right?

Regarding the for-loop variable, it currently doesn’t work because the policy does not permit calling `__iter__`. As a result, we cannot infer the type of the loop variable. Should we consider allowing `__iter__` for built-in types to address this limitation?
CC: @krassowski

krassowski:
> Should we consider allowing `__iter__` for built-in types to address this limitation?

Yes, it should be fine for most built-in types but not in general, because generators could be written in a way which exhausts them when iteration starts. Basically, we need to allow-list them individually.

krassowski:
So if we have:

```python
collection: list[int] = [1, 2, 3]
it = iter(collection)
type(it) # <class 'list_iterator'>
```

we should be able to infer the type of the iterator values without ever iterating. Similarly, for `dict`: `items()`, `keys()`, `values()` they have dedicated classes. Not sure if we want to go this route in this PR just saying that this is a possibility.

krassowski:
> but not in general, because generators could be written in a way which exhausts them when iteration starts

For example, we would not want to call `next(iter(f))` on file objects as this would be a serious (slow) side-effect depending on the file system:

```python
with open("example.txt") as f:
    for line in f:
        pass
```

So even though we could recover using `f.seek(0)`, this would be to slow for practical use case.

Other examples of built-ins which would exhaust the data are `map`, `filter`, `zip`, `enumerate`, or any `itertools` object. It might be that dedicated logic would be required to handle the iteration for anything more complex that the plain built-in collections, but maybe it is ok.

In the ideal world we could infer most of it from typing like, for

```python
f = open('example.txt')
for line in f:
```

we should be able to call `inspect.signature(f.__iter__).return_type` - currently we could do it, but it returns `_empty` rather than returning a `Generator` return type. We would need to pull standard library typing stubs which cover it.

Overall, I think for this PR having allow-list of collections for which it is fine to call `next(iter(collection))` to get a value is fine.

IPython/core/guarded_eval.py
```diff
@@ -1326,48 +1424,74 @@ def _list_methods(cls, source=None):
...
 ALLOWED_CALLS = {
     bytes,
     *_list_methods(bytes),
+    bytes.__iter__,
     dict,
     *_list_methods(dict, dict_non_mutating_methods),
+    dict.__iter__,
+    dict.keys,
+    dict.values,
+    dict.items,
```
krassowski: Isn't it (`dict.keys`, `dict.values` and `dict.items`) already covered by `dict_non_mutating_methods`?

IPython/core/guarded_eval.py
```diff
@@ -1326,48 +1424,74 @@ def _list_methods(cls, source=None):
...
     str,
     *_list_methods(str),
+    str.__iter__,
     tuple,
     *_list_methods(tuple),
+    tuple.__iter__,
     bool,
     *_list_methods(bool),
+    enumerate,
+    enumerate.__iter__,
```
krassowski:
```suggestion
```

We cannot have this one, compare

good:

image

bad: 
image

krassowski: Unless we make `deepcopy` of the enumerate object like:

```python
from copy import deepcopy
if isinstance(sth, enumerate):
     sth = deepcopy(sth)
```

IPython/core/guarded_eval.py
Darshan808: Ah, it does. Thanks!

IPython/core/guarded_eval.py
Darshan808: For now, I just removed those from the `allow_list`.

IPython/core/guarded_eval.py
```diff
@@ -817,6 +817,86 @@ def dummy_function(*args, **kwargs):
...
+        loop_locals = context.transient_locals.copy()
+        loop_context = context.replace(transient_locals=loop_locals)
+
+        if sample is not None:
+            try:
+                fake_assign = ast.Assign(
+                    targets=[node.target], value=ast.Constant(value=sample)
+                )
+                _handle_assign(fake_assign, loop_context)
+            except:
```
krassowski:
```suggestion
            except Exception:
```

IPython/core/guarded_eval.py
```diff
@@ -817,6 +817,86 @@ def dummy_function(*args, **kwargs):
...
+        merged_locals = _merge_dicts_by_key(
+            [loop_locals, context.transient_locals.copy()], policy
+        )
+        context.transient_locals.update(merged_locals)
+
+        return result
+    if isinstance(node, ast.For):
+        try:
+            iterable = eval_node(node.iter, context)
+        except:
```
krassowski: We do not want to catch user interrupts or system exceptions (almost never)

```suggestion
        except Exception:
```

krassowski (approved):
Thank you @Darshan808, this is perfect for now - just one comment on exception handling.

PR #15057, opened 2025-11-02
Reduce types in splitinput.py

Carreau:
As far as I can tell, those are always strings, so should remove some overhead.

Also deprecated and ruff-format the file.

### DIFF
--- a/IPython/core/splitinput.py
+++ b/IPython/core/splitinput.py
@@ -141,7 +141,7 @@ class LineInfo:
             Use ``shell._ofind(line_info.ifun)`` directly instead.
         """
         warnings.warn(
-            "LineInfo.ofind() is deprecated since IPython 9.8. "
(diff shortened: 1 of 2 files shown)


IPython/core/splitinput.py
```diff
@@ -135,11 +136,20 @@ def ofind(self, ip) -> OInfo:
 
         Does cache the results of the call, so can be called multiple times
         without worrying about *further* damaging state.
+
+        .. deprecated:: 9.8
+            Use ``shell._ofind(line_info.ifun)`` directly instead.
         """
+        warnings.warn(
+            "LineInfo.ofind() is deprecated since IPython 9.8. "
```
Carreau:
```suggestion
            "LineInfo.ofind() is deprecated since IPython 9.9. "
```

PR #15059, opened 2025-11-03
Add `setuptools-scm` to pyflyby tests

krassowski:
Testing if this fixes CI

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -80,7 +80,7 @@ jobs:
         cd ..
         git clone https://github.com/deshaw/pyflyby
         cd pyflyby
-        pip install meson-python meson ninja pybind11>=2.10.4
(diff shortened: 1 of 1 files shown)


.github/workflows/downstream.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         cd ..
         git clone https://github.com/deshaw/pyflyby
         cd pyflyby
-        pip install meson-python meson ninja pybind11>=2.10.4
+        pip install meson-python meson ninja pybind11>=2.10.4 setuptools-scm
         pip install setuptools wheel # needed for epydoc
         pip install --no-build-isolation -ve .[test]
```
Carreau, line 85: +1, 

Should we try removing --no-build-isolation, and see if that gets rid of having to list explicit dependencies ?

I think also in the end, we should move to released version to not rebuild every time now that it's expensive to rebuild.

Carreau (approved):

PR #15070, opened 2025-11-10
FEATURE: History: allow session number without trailing slash

Carreau:
Prompted by #15069


Carreau:
@pierre-haessig, do you think that would help ?

pierre-haessig:
Hi,
Making the trailing `/` optional is a small time saviour, but not critical I would say. Actually, it even create a small amount of ambiguity in the syntax description. Previously, the input range syntax could be described as:

> `session description`/`line range description`, with the session description being optional.

Now a single number can be either a line or a session if it's relative.

What I find more useful in this PR is the feature (or the documentation if it was preexisting) of the `4-` syntax

Finally, I would suggest to remove the extra doc line about `~4/4-`, to keep the examples a bit more orthogonal between each others.

Carreau:
thanks i've remove allowing without trailing slashes.

PR #15071, opened 2025-11-12
add cell_meta to ExecutionInfo and pass through run_cell

erawn:
This is a PR to propose adding `cell_meta` to `ExecutionInfo` so that IPython extensions can utilize it. This PR builds on [1169](https://github.com/ipython/ipykernel/pull/1169) which passed `cell_meta` to the kernel, and adds it as an *optional* field in `ExecutionInfo` which is then passed to `run_cell`. [Other half on the ipykernel side is here](https://github.com/ipython/ipykernel/pull/1475)

My own personal use case (and the one I would imagine most have for this), is building IPython extensions which pair with extended/customized frontend clients. Personally, I've had to do some rather messy hacks in order to run an IPython and JupyterLab extension which can communicate, and for anything more than a prototype it would be valuable to have a way to append information to the execution requests directly that can be seen by extensions. Since `cell_meta` already extends the `metadata` field in the [Jupyter Messaging protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#metadata), this seemed like the most natural way to go about it. 

As discussed [here](https://github.com/ipython/ipykernel/pull/886) when `cell_id` was introduced, there was talk of refactoring into a more general `metadata` field, but it seems the current decision is to keep both `cell_id` and `cell_meta`. This PR would extend that decision and continue passing cell_meta. 

I'm interested in feedback on what would be most useful to others or if there are decisions we can make now that support the most flexibility long-term. Continuing on with `cell_meta` as I've done here wouldn't introduce any breaking changes, and so seems the most straight forward, but I'm curious what others think. Thanks!


erawn:
Because the default for cell_meta is already None, I've kept that default and changed the existing instance where the default was an empty dictionary. 

Btw, using empty dictionaries as default arguments is [warned against in the docs]('[cell_meta](https://docs.python.org/3/tutorial/controlflow.html#default-argument-values)') so it seemed reasonable to stick with `None` as the default.

rodrigosf672 (approved):
Thank you so much for these changes, @erawn. The more I understood about them, the more appreciate the value of this contribution. 

Non-blocking question: since `cell_meta` becomes part of the public execution path, do you think it’d be worth adding a brief docstring note about its expected use and structure? Even one sentence might help extension authors discover it. If you think that'd be a good idea, I can address it as a separate issue/PR. 

Again, thanks so much for this contribution and for being so open about getting feedback from the Jupyter community. :)

erawn:
Thank you so much for your help! You've made my first IPython pull request a really lovely experience :) 

Happy to add the docstring—is the most recent change I've made what you had in mind? The docstring for `run_cell` seems to be where most of the documentation is that I can see, but if there is a better place just let me know.

rodrigosf672 (commented):
Yes, that's great! The wording is clear and I think it should be helpful for extension authors. Thank you so much, and congrats on your first IPython pull request!! 🙌

Carreau:
rebasing this, and happy to merge if there is still interest.

erawn:
@Carreau yes that would be lovely! Is there anything I can do to help get it over the finish line? Thanks so much!

tests/test_interactiveshell.py
```diff
@@ -1397,3 +1398,42 @@ def test_enable_gui_osx():
     ip.enable_gui()
 
     ip.simple_prompt = simple_prompt
+
+class TestCellMeta(unittest.TestCase):
```
Carreau, line 1404:
```suggestion
import unittest

class TestCellMeta(unittest.TestCase):
```

There was recent refactors to convert this file to pytest only, cleanest fix is to reimport unittest for now.

Carreau:
I'm just going to get the tests to pass due to recent refactor.

PR #15073, opened 2025-11-14
Add system_raise_on_error config option for ! shell operator

### LINKED ISSUE #12264: Shell command via bang doesn't have an error on status code
I ran into a use-case where I wanted to call a small bash script in a notebook, but couldn't halt execution easily based on exit status. This is similar to #11198 but this is about invoking a shell command via `!`. I'll use a similar example:

```
======= Cell 1 =======

shell_result = !command_that_produces_error

======= Cell 2 =======

# I don't want to run this if the bash command didn't return status code 0
print(shell_result)
```

This was solved for bash magics `%%bash` in https://github.com/ipython/ipython/pull/11287 but not for `!` shell command.

adityawasudeo:
Fixes #12264

Add system_raise_on_error configuration option that raises CalledProcessError when shell commands executed via the ! operator return non-zero exit status. This brings similar error-handling behavior to the ! operator as was added to %%bash magic in PR #11287.

Changes:
- Add system_raise_on_error Bool traitlet config (default: False)
- Import CalledProcessError from subprocess
- Modify system_piped() to raise on non-zero exit when enabled
- Modify system_raw() to raise on non-zero exit when enabled
- Modify getoutput() to raise on non-zero exit when enabled (uses get_output_error_code to capture exit status)

When enabled, users can halt notebook execution on command failures:
  get_ipython().system_raise_on_error = True
  !false  # Now raises CalledProcessError

Note: A follow-up PR to ipykernel will be needed to support this in Jupyter notebooks, as ZMQInteractiveShell overrides system_piped().

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -22,6 +22,7 @@ import re
 import runpy
 import shutil
 import subprocess
+from subprocess import CalledProcessError
 import sys
 import tempfile
 import traceback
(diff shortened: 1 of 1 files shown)


adityawasudeo:
Bumping this. @Carreau would you be able to review it if you have a few minutes to spare? Many thanks!

adityawasudeo:
Pinging this, is there any interest in this PR? Would be nice to at least get some feedback on why this didn't make the cut

Carreau:
Sorry, thanks for pinging me.

PR #15074, opened 2025-11-14
Use values for tab completion of variables created using annotated assign.

### LINKED ISSUE #15035: Completion for values created with annotated assign could use the value itself
Currently we extract the type from the `AnnAssign` node e.g. from:

```python
x: str | int = 5
```

we would extract `str | int`. In such case the type annotation has more information than the value (when it comes to tab completion).

However, sometimes the value will have more information. Consider:

```python
y: dict = {'long_key': 1, 'longer_key': 2}
```

For `y` completions the `dict` annotation is redundant with respect to the value and the value contains more information as it will allow us to complete the keys of the dictionary as in:

```
y["<tab>
```

Darshan808:
### Description
In annotated assignments, if the type annotations are generic, use value instead for tab completion.
Considered generic types: `dict, list, set, tuple, frozenset`.


IPython/core/guarded_eval.py
```diff
@@ -499,7 +496,7 @@ def guarded_eval(code: str, context: EvaluationContext):
     ast.Not: ("__not__",),
 }
 
-
+GENERIC_CONTAINER_TYPES = (dict, list, set, tuple, frozenset)
```
krassowski, line 505:
```suggestion
GENERIC_CONTAINER_TYPES = (dict, list, set, tuple, frozenset)


```

krassowski (commented):
From reading the code this will works for:

```python
x: dict | str = {'test': 1}
x.<tab> # shows both dict and str methods
```

```python
x: dict = {'test': 1}
x.<tab> # shows both dict methods
x['t.<tab> # shows test
```

but not for 

```python
x: dict | str = {'test': 1}
x['t.<tab> # does not show test
```

How difficult it would be to handle the last case properly? Would it be a significant expansion to the codebase?

Darshan808:
This PR and https://github.com/ipython/ipython/pull/15078 are related!
In fact, the following code:
```python
x: dict | str = {'test': 1}
x.<tab>
```
doesn't work in this branch. Issue: https://github.com/ipython/ipython/issues/15077

Based on the review, I also handled the following case in https://github.com/ipython/ipython/pull/15078:
```python
x: dict | str = {'test': []}
x.<tab>        # shows both dict and str methods
x['test'].<tab> # shows list methods
```

I guess we first need to finish and merge https://github.com/ipython/ipython/pull/15078

krassowski (approved):
Thank you @Darshan808!

PR #15076, opened 2025-11-17
Fix error on tab completions

### LINKED ISSUE #15075: tab-completion sometimes raise error
IPython Version: 9.0.2
Python version: 3.11

```
In [1]: class A: pass

In [2]: def f():
   ...:     a=A()
   ...:     assert a.
```

position the cursor after the `.` then press `[tab] [tab]`

see a traceback ending with `AttributeError: 'Assert' object has no attribute 'value'` .

Darshan808:
### Description
Added a logic to check whether the node we have is `ast.Expr` before trying to access its `value`.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -1352,8 +1352,6 @@ class Completer(Configurable):
             assert res is not None
             if len(res.body) != 1:
                 continue
-            if not isinstance(res.body[0], ast.Expr):
-                continue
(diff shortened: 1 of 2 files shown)


krassowski:
Thank you! Is it easy to add a test?

Darshan808:
The test fails as expected!
Adding the fix again now.

krassowski (approved):
Thank you!

PR #15078, opened 2025-11-17
Show completions for annotated union types

### LINKED ISSUE #15077: Completion not shown for annotated union types
### Reproduce
```python
In [1]: a: int | str
   ...: a.<tab>
```
No completions are shown!

### Expected
Completions should include the combined attributes of `int` and `str`.

#### IPython Version: latest

Darshan808:
### Description
Added logic for `ast.BinOp` case to return an object with merged attributes when the binary operation represents a union type.

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -567,7 +567,7 @@ def _validate_policy_overrides(
     return all_good
 
 
-def is_type_annotation(obj) -> bool:
+def _is_type_annotation(obj) -> bool:
     """
     Returns True if obj is a type annotation, False otherwise.
     """
@@ -793,7 +793,7 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
 
         if is_property:
             if return_type is not None:
-                if is_type_annotation(return_type):
+                if _is_type_annotation(return_type):
                     context.transient_locals[node.name] = _resolve_annotation(
                         return_type, context
                     )
@@ -809,7 +809,7 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
             pass
 
         if return_type is not None:
-            if is_type_annotation(return_type):
+            if _is_type_annotation(return_type):
                 dummy_function.__annotations__["return"] = return_type
             else:
                 dummy_function.__inferred_return__ = return_type
@@ -933,18 +933,18 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
     if isinstance(node, ast.Assign):
         return _handle_assign(node, context)
     if isinstance(node, ast.AnnAssign):
-        context.current_value = getattr(node, "value", None)
+        context_with_value = context.replace(current_value=getattr(node, "value", None))
         if node.simple:
-            annotation_result = eval_node(node.annotation, context)
-            if is_type_annotation(annotation_result):
+            annotation_result = eval_node(node.annotation, context_with_value)
+            if _is_type_annotation(annotation_result):
                 value = _resolve_annotation(annotation_result, context)
             else:
                 value = annotation_result
             context.transient_locals[node.target.id] = value
         # Handle non-simple annotated assignments only for self.x: type = value
         if _is_instance_attribute_assignment(node.target, context):
-            annotation_result = eval_node(node.annotation, context)
-            if is_type_annotation(annotation_result):
+            annotation_result = eval_node(node.annotation, context_with_value)
+            if _is_type_annotation(annotation_result):
                 value = _resolve_annotation(annotation_result, context)
             else:
                 value = annotation_result
@@ -966,7 +966,11 @@ def eval_node(node: Union[ast.AST, None], context: EvaluationContext):
     if isinstance(node, ast.BinOp):
         left = eval_node(node.left, context)
         right = eval_node(node.right, context)
-        if is_type_annotation(left) and is_type_annotation(right):
+        if (
+            isinstance(node.op, ast.BitOr)
+            and _is_type_annotation(left)
+            and _is_type_annotation(right)
+        ):
             left_duck = (
                 _Duck(dict.fromkeys(dir(left)))
                 if policy.can_call(left.__dir__)
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -30,7 +30,7 @@ from types import MethodDescriptorType, ModuleType, MethodType
 
 from IPython.utils.decorators import undoc
 
-
+import types
 from typing import Self, LiteralString
 
 if sys.version_info < (3, 12):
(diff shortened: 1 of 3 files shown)


Darshan808:
Not sure why it's failing on `oldest-deps` test. I tried locally and  it passes with python 3.11 too.

IPython/core/guarded_eval.py
```diff
@@ -927,6 +960,25 @@ def dummy_function(*args, **kwargs):
     if isinstance(node, ast.BinOp):
         left = eval_node(node.left, context)
         right = eval_node(node.right, context)
+        if is_type_annotation(left) and is_type_annotation(right):
```
krassowski: While at it, any idea why:
```python
def f() -> dict | int: pass
f().<tab>
```
does not work? It does not need to be addressed in this PR but would be good to understand.

IPython/core/guarded_eval.py
```diff
@@ -903,12 +927,21 @@ def dummy_function(*args, **kwargs):
     if isinstance(node, ast.Assign):
         return _handle_assign(node, context)
     if isinstance(node, ast.AnnAssign):
+        context.current_value = getattr(node, "value", None)
```
krassowski: I'm somewhat concerned that this is writing directly rather than making a copy of the context with `context.replace()` because this could lead to a stale value on:

```python
x: int | dict = {'a': []}  # Sets current_value to {'a': []}
def f() -> int | dict: pass
f().<tab>
```

Not a terrible bug, but still might be hard to debug later.

IPython/core/guarded_eval.py
```diff
@@ -927,6 +960,25 @@ def dummy_function(*args, **kwargs):
...
+                _Duck(dict.fromkeys(dir(left)))
+                if policy.can_call(left.__dir__)
+                else _Duck()
+            )
+            right_duck = (
+                _Duck(dict.fromkeys(dir(right)))
+                if policy.can_call(right.__dir__)
+                else _Duck()
+            )
+            value_node = context.current_value
```
krassowski, line 987: An alternative would be to have `.parent` in the context and check if it was an annotated assign. But I guess adding `current_value` is a smaller change.

IPython/core/guarded_eval.py
krassowski: If I recall correctly that type annotations only support `|` binary operation then we can reject more quickly here:


```suggestion
        if node.op is ast.BitOr and is_type_annotation(left) and is_type_annotation(right):
```

krassowski: It looks like there was some discussion on supporting `&` in the future but it died out, and if it was added then we will want to narrow values to a shared subset rather than merge them into a union.

IPython/core/guarded_eval.py
```diff
@@ -564,6 +567,27 @@ def _validate_policy_overrides(
...
+    if hasattr(obj, "__module__") and obj.__module__ == "typing":
+        return True
+    if isinstance(obj, types.GenericAlias):
+        return True
+    if hasattr(types, "UnionType") and isinstance(obj, types.UnionType):
+        return True
+    if type(obj).__name__ in ("_GenericAlias", "_SpecialForm", "_UnionGenericAlias"):
+        return True
+
+    return False
```
krassowski, line 591: I worry that the implementation is brittle and relies on the internal names. I did something similar for generic alias in:

https://github.com/ipython/ipython/blob/2769bb58432dcc1e24a466b9596b918daced9bee/IPython/core/guarded_eval.py#L315-L320

Can you add think a little bit if your implementation can be improved and add more tests to ensure we have full coverage?

IPython/core/guarded_eval.py
```diff
@@ -564,6 +567,27 @@ def _validate_policy_overrides(
     return all_good
 
 
+def is_type_annotation(obj) -> bool:
```
krassowski: Let's make it private:

```suggestion
def _is_type_annotation(obj) -> bool:
```

krassowski (commented):
This approach makes sense!

IPython/core/guarded_eval.py
Darshan808: Thanks for pointing this out! I needed to handle it in functions as well.

IPython/core/guarded_eval.py
Darshan808: Oops, thanks for catching that!

IPython/core/guarded_eval.py
Darshan808: I've tweaked the implementation and added a unit test for this function. Are there any other cases you think we might have missed?

krassowski: Well, at some level I wonder if we need that function. If we knew that we are in annotation context (say if we set an attribute on context) then maybe we would not need to check that in the first place? Or would it make things harder?

Darshan808: Thanks for the suggestion. I explored that approach. It simplifies certain parts but complicates others. For example, with an annotation-aware context we no longer need to check for type annotations when evaluating a bitwise-or node, since the context already tells us we're in annotation mode. But the current logic returns simple types as annotations and `_Duck` types as real objects, and in places like this
```python
if _is_type_annotation(return_type):
    context.transient_locals[node.name] = _resolve_annotation(
        return_type, context
    )
else:
    context.transient_locals[node.name] = return_type
```
and this
```python
def dummy_function(*args, **kwargs):
    pass

if _is_type_annotation(return_type):
    dummy_function.__annotations__["return"] = return_type
else:
    dummy_function.__inferred_return__ = return_type
 ```
 we still need to distinguish whether the returned value is a type annotation or an actual `_Duck` instance.

If there’s a way to make that distinction without `is_type_annotation`, it should work, but I haven’t found one yet.

File: tests/test_completer.py [hunk 6]
```diff
@@ -2527,6 +2527,46 @@ def _(expected):
...
+                [
+                    "class T:",
+                    "   @property",
+                    "   def p(self) -> int | str: pass",
+                    "t = T()",
+                    "t.p.",
+                ]
+            ),
+            ["bit_length", "capitalize"],
+        ],
```
krassowski: It looks like the oldest jedi version we accept does not support some of these:

```
FAILED tests/test_completer.py::test_undefined_variables[t: int | str\nt.-insert_text42-True-minimal] - AssertionError: bit_length not found among []
assert 0 == 1
 +  where 0 = len([])
FAILED tests/test_completer.py::test_undefined_variables[def func() -> int | str: pass\nfunc().-insert_text43-True-minimal] - AssertionError: bit_length not found among ['.editorconfig', '.flake8', '.git-blame-ignore-revs', '.git/', '.gitattributes', '.github/', '.gitignore', '.mailmap', '.meeseeksdev.yml', '.pre-commit-config.yaml', '.readthedocs.yaml', '.venv/']
assert 0 == 1
 +  where 0 = len([])
FAILED tests/test_completer.py::test_undefined_variables[class T:\n   @property\n   def p(self) -> int | str: pass\nt = T()\nt.p.-insert_text44-True-minimal] - AssertionError: bit_length not found among []
assert 0 == 1
 +  where 0 = len([])
```

We could:
- require newer version of jedi (not sure if this is the right time to move it)
- move these test cases down to `test_undefined_variables_without_jedi`
- add another argument with min required jedi version and then in the body call `pytest.skip` conditionally

Darshan808: > move these test cases down to test_undefined_variables_without_jedi

I think this is the simplest solution in this case.

krassowski (approved):
Thank you @Darshan808!

PR #15079, opened 2025-11-17
Fallback to type annotations for attribute completions

### LINKED ISSUE #14917: Tab completer should fallback to type annotations for attributes
When evaluation policy forbids access, the tab completer should fallback to type annotations for attributes.

Give `repro.py`:

```python
class StringMethods:
    def a():
        pass

class Test:
     str: StringMethods
     def __init__(self):
         self.str = StringMethods()
     def __getattr__(self, name):
         raise AttributeError(f"{name} not found")
```

or

```python
class StringMethods:
    def a():
        pass

class Test:
     def str(self) -> StringMethods:
         return StringMethods()
     def __getattr__(self, name):
         raise AttributeError(f"{name} not found")
```

IPython should be able to complete on:

```
%config Completer.use_jedi = False
from repro import Test
r = Test()
r.str.# <tab>
```

Darshan808:
### Description:
Enable IPython tab completer to use type annotations when attribute access is restricted by the evaluation policy.

### DIFF
--- a/IPython/core/guarded_eval.py
+++ b/IPython/core/guarded_eval.py
@@ -31,7 +31,7 @@ from types import MethodDescriptorType, ModuleType, MethodType
 from IPython.utils.decorators import undoc
 
 
-from typing import Self, LiteralString
+from typing import Self, LiteralString, get_type_hints
 
 if sys.version_info < (3, 12):
     from typing_extensions import TypeAliasType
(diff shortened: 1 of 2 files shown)


Darshan808:
The test fails as expected in the first commit.

Carreau (approved):

PR #15081, opened 2025-11-21
Strictly suppress `file_matcher` in attribute completions

### LINKED ISSUE #15031: Suppress paths tab completion when completing a method/attribute

Darshan808:
### Description
The current logic only suppresses matchers if they are in the suppress set and there are matching results. As a result, `file_matcher` can still display file matches when attribute completion fails. This PR updates the logic to always suppress `file_matcher` during attribute completion, regardless of whether any attributes were successfully matched.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -2215,7 +2215,11 @@ class IPCompleter(Completer):
         text = context.token
         code_until_cursor = self._extract_code(context.text_until_cursor)
         completion_type = self._determine_completion_context(code_until_cursor)
-        if completion_type == self._CompletionContextType.ATTRIBUTE:
+        in_cli_context = self._is_completing_in_cli_context(code_until_cursor)
+        if (
+            completion_type == self._CompletionContextType.ATTRIBUTE
+            and not in_cli_context
+        ):
             return {
                 "completions": [],
                 "suppress": False,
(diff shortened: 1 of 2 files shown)


krassowski (changes requested):
This would break completion on `ls` automagic:

| Before | This PR as of now |
|--|--|
| Image | Image |

Can you add a test for the `ls` case?

Darshan808:
Oops, I didn't realize backticks inside double quotes get executed. My git commit -m "fix `ls .` case" ended up expanding the command.
Anyways, cases like `ls file.<tab>` still don't complete or suggest files with the same name but different extensions, since they're treated as attribute completion. This doesn't seem trivial to fix. Should we address it in this PR, or would it be better to open a separate issue?

krassowski:
> Anyways, cases like ls file.<tab> still don't complete or suggest files with the same name but different extensions, since they're treated as attribute completion. This doesn't seem trivial to fix. Should we address it in this PR, or would it be better to open a separate issue?

We should not introduce regressions in this PR. Let's take a step back and think about it differently.

I think there is only a limited set of commands which can be auto-evaluated without `!` which are stored in `get_ipython().alias_manager.aliases`. It feels that we should determine whether we are completing in such CLI alias, line  magics (`%`), or bang expressions (`!`). This would be similar to `_determine_completion_context` but I would probably add a separate method for simplicity. The logic for detecting if we are completing in cli-alias should be along the lines of
> does the line start with a string enumerated in `aliases` followed by a whitespace character and there is no Python variable which overshadows it (as in there is no `ls = 1` or similar defined)

IPython/core/completer.py
```diff
@@ -3651,10 +3695,25 @@ def _complete(self, *, cursor_line, cursor_pos, line_buffer=None, text=None,
...
+                ):
+                    should_suppress = True
+                else:
+                    should_suppress = (
+                        (suppression_config is True)
+                        or (
+                            suppression_recommended
+                            and (suppression_config is not False)
+                        )
+                    ) and has_any_completions(result)
```
krassowski: Looking at how this requires hard-coding matcher-specific behaviour (or breaking backward compatibility) I now think it would be better to go with your initial idea of short-circuiting in `file_matcher` itself.

Darshan808: That makes sense!

krassowski (approved):
Thank you @Darshan808!

PR #15085, opened 2025-11-30
Make debugger work with Python 3.14

ccordoba12:
A Linux user reported in https://github.com/spyder-ide/spyder/issues/25299 that our debugger is failing with their distro packages for Python 3.14 due to an error in IPython 8 (the error is not present in version 9).

After some digging, I applied the minimal changes I could from commit a796b95 to fix that. I also checked that the debugger is working fine for Python 3.11 to 3.13.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -248,13 +248,7 @@ class Pdb(OldPdb):
         docs for more info.
         """
 
-        # Parent constructor:
-        try:
-            self.context = int(context)
-            if self.context <= 0:
-                raise ValueError("Context must be a positive integer")
-        except (TypeError, ValueError) as e:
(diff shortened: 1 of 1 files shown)


ccordoba12:
@Carreau, @krassowski, the failures in the downstream tests are due to a DeprecationWarning in the flaky pytest plugin (so, not related to this). However, I don't know how to fix them.

Carreau:
Thanks, that sounds good, I've restarted the tests and worse case I think it's ok to make a release. 

I'll merge that and do a release at the same time later, so keeping opened for now to remind myself of doing it.

ccordoba12:
Thanks @Carreau!

Carreau:
Should be released as 8.38. Probably soon on conda-forge.

PR #15088, opened 2025-12-01
execution count more atomic, run_cell_async reenterable

drorspei:
This pull request makes `run_cell_async` able to re-enter itself without messing the history causing a new session. This solves issue #15087 

The changes are to increment `execution_count` of `InteractiveShell` _before_ running the user code in a cell, and some compatible changes so that prompts and outputs get the correct `execution_count`.

With this change it's possible to modify frontends so that they can run multiple cells concurrently. I've written a small one-file package on top of ipykernel that shows this nicely and works well

https://github.com/drorspei/bgipykernel

### DIFF
--- a/IPython/core/displayhook.py
+++ b/IPython/core/displayhook.py
@@ -63,7 +63,7 @@ class DisplayHook(Configurable):
 
     @property
     def prompt_count(self):
-        return self.shell.execution_count
+        return self.shell.execution_count - 1
 
     #-------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Carreau:
I don't super like the -1 in a few places, but let's try

drorspei:
Yeah... I tried the alternative of starting the execution count at 0 and having everything show the "current" execution_count, but other projects (e.g. ipykernel) already have their own minus ones, so the projects would have to be modified in tandem

Carreau:
It should be released :-)

PR #15095, opened 2025-12-04
Enable pretty-printing for PEP-649 annotated functions

zware:
This PR detects the possibility of such functions by importing annotationlib.Format from inspect, and wraps inspect.signature in a partial that sets annotation_format=Format.FORWARDREF.

This may be more proof-of-concept than final form; I won't be offended if you want to scrap this and tackle the issue a different way :)

I'm also not sure this is actually enough to support all possibilities, but it at least gets us from

```python
In [1]: def test(p: Annotated): pass

In [2]: test

<huge traceback elided>

NameError: name 'Annotated' is not defined

In [3]:
```

to

```python
In [1]: def test(p: Annotated): pass

In [2]: test
Out[2]: <function __main__.test(p: Annotated)>
```


IPython/lib/pretty.py
```diff
@@ -109,6 +109,18 @@ def _repr_pretty_(self, p, cycle):
...
+# Allow pretty-printing of functions with PEP-649 annotations
+try:
+    # Actually an annotationlib symbol imported by inspect
+    # but if inspect has it that should mean it can use it
+    from inspect import Format
+except ImportError:
+    pass
+else:
+    from functools import partial
+    signature = partial(signature, annotation_format=Format.FORWARDREF)
```
Carreau: I would prefer have a `if sys.version_info>= (3,14):...`, this way once we drop 3.13 we can remove conditional.

Carreau:
Thanks !

PR #15096, opened 2025-12-05
Use Any type in traceback tuple

jenshnielsen:
And not traitlets any trait. Since this is a python standard library type, it seems likely that using the Any trait here was a typo?

Formally I would expect the traceback type to be `types.TracebackType | None` but perhaps there
is a good reason to use Any here?


Carreau:
Thanks !

PR #15099, opened 2025-12-08
Fix filename of CVE test

decorator-factory:
From a local pytest run, and also the most recent CI run on main (https://github.com/ipython/ipython/actions/runs/19927362702/job/57130406121) it looks like `tests/cve.py` is not executed at all because its file name does not start with `test_`. This PR fixes that

(the test does still pass, thankfully)


Carreau:
Thanks

Issue #15098, opened 2025-12-08
Typing string literal with a surrogate crashes IPython


decorator-factory:
Python's `str` objects are allowed to contain "surrogates", values from `\uD800` to `\uDFFF` that are used by UTF-16 to encode certain characters. However, UTF-8 doesn't allow encoding these values. That means that `a_str.encode("utf-8")` can fail.

If I type `"\udead"` (as a string literal) into IPython and press "Enter", it crashes:

<details>

<summary>
Crash
</summary>

```
$ ipython
Python 3.13.5 (main, Jul  4 2025, 23:57:18) [GCC 15.1.1 20250521 (Red Hat 15.1.1-2)]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.8.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can find how to type a Unicode symbol by back-completing it, eg `\Ⅷ<tab>` will expand to `\ROMAN NUMERAL EIGHT`.

In [1]: "\udead"
Traceback (most recent call last):
  File "/home/df/.pyenv/versions/3.13.5/bin/ipython", line 8, in <module>
    sys.exit(start_ipython())
             ~~~~~~~~~~~~~^^
  File "/home/df/.pyenv/versions/3.13.5/lib/python3.13/site-packages/IPython/__init__.py", line 144, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/home/df/.pyenv/versions/3.13.5/lib/python3.13/site-packages/traitlets/config/application.py", line 1075, in launch_instance
    app.start()
    ~~~~~~~~~^^
  File "/home/df/.pyenv/versions/3.13.5/lib/python3.13/site-packages/IPython/terminal/ipapp.py", line 325, in start
    self.shell.mainloop()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/home/df/.pyenv/versions/3.13.5/lib/python3.13/site-packages/IPython/terminal/interactiveshell.py", line 1017, in mainloop
    self.interact()
    ~~~~~~~~~~~~~^^
  File "/home/df/.pyenv/versions/3.13.5/lib/python3.13/site-packages/IPython/terminal/interactiveshell.py", line 1010, in interact
    self.run_cell(code, store_history=True)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #15101, opened 2025-12-14
Fix crash on literal with surrogate (#15098)

decorator-factory:
Fixes issue #15098.

This now works as expected:
```
$ ipython
Python 3.13.5 (main, Jul  4 2025, 23:57:18) [GCC 15.1.1 20250521 (Red Hat 15.1.1-2)]
IPython 9.9.0.dev -- An enhanced Interactive Python. Type '?' for help.
Tip: Use `object?` to see the help on `object`, `object??` to view its source

In [1]: "\ud800"
Out[1]: '\ud800'

In [2]: ord(_)
Out[2]: 55296
```

This raises an exception, as intended in Python 3.13+:
```
In [1]: def f():
   ...:     "\ud800"
UnicodeEncodeError                        Traceback (most recent call last)
File ~/.pyenv/versions/3.13.5/lib/python3.13/codeop.py:118, in Compile.__call__(self, source, filename, symbol, flags, **kwargs)
    116     flags &= ~PyCF_DONT_IMPLY_DEDENT
    117     flags &= ~PyCF_ALLOW_INCOMPLETE_INPUT
--> 118 codeob = compile(source, filename, symbol, flags, True)
    119 if flags & PyCF_ONLY_AST:
    120     return codeob  # this is an ast.Module in this case

UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' in position 0: surrogates not allowed
```

### DIFF
--- a/IPython/core/async_helpers.py
+++ b/IPython/core/async_helpers.py
@@ -136,20 +136,15 @@ def _pseudo_sync_runner(coro):
 
 
 def _should_be_async(cell: str) -> bool:
-    """Detect if a block of code need to be wrapped in an `async def`
+    """Detect if a block of code needs to be wrapped in an `async def`
 
-    Attempt to parse the block of code, it it compile we're fine.
-    Otherwise we  wrap if and try to compile.
-
-    If it works, assume it should be async. Otherwise Return False.
-
-    Not handled yet: If the block of code has a return statement as the top
-    level, it will be seen as async. This is a know limitation.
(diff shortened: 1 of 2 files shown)


Carreau (approved):

decorator-factory:
Looks like the macos-13 runner is decommissioned :thinking:

Carreau:
> Looks like the macos-13 runner is decommissioned 🤔

That's fine, let update that later.

PR #15105, opened 2025-12-19
Add prompt_toolkit's unix_word_rubout to assignable commands for shortcuts

buhanec:
`unix-word-rubout` / `nc.unix_word_rubout` is closer to a lot of folks' shell's Ctrl-W behaviour, so it would be nice to be able to set it up cleanly in config files.

As best as I can tell the set of unassigned allowed commands is neither tested nor documented anywhere; it's not clear what warrants inclusion or exclusion from the group.


krassowski (approved):
Makes sense to me!

PR #15106, opened 2025-12-20
Fix completions for methods starting with `_`

### LINKED ISSUE #15104: Completion no longer shows underscore methods when user types in `_` since IPython 9.1
This appears to have been introduced in https://github.com/ipython/ipython/pull/14838.

```python
# ipython --Completer.use_jedi False
class Test:
    _test1 = 1
    _test2 = 2
t = Test()
```

Before

Image

After

Image

Darshan808:
### Description
The current regex pattern only matches attribute access where the attribute name starts with a letter ([a-zA-Z]). This PR modifies the pattern to handle attributes starting with `_` too.


krassowski (commented):
The change looks good, should we also have end-to-end test?

krassowski (approved):
Thank you @Darshan808!

PR #15110, opened 2025-12-26
fix: bug#15089 - Removing leading indentation 

marcoscoladev:
# Fix doctest paste: strip prompts and dedent (#15089)

## Summary

This PR improves how IPython handles pasting doctest and xdoctest code snippets. When pasting code like:

```
>>>     print(1)
```

IPython now strips the `>>>` prompt and also removes the extra indentation, so the result is:

```
print(1)
```

Previously, the result would have been:

```
    print(1)
```

which can cause an `IndentationError` at top-level.

## What changed
- `inputtransformer2.py`
    - Enhanced PromptStripper to:
        - Detect doctest-style pastes when any line matches `^\s*>>>`.
        - In doctest-mode, strip `^\s*>>>\s?` and `^\s*\.\.\.\s?` (continuation).
        - After stripping, apply `textwrap.dedent()` to the joined block, then return dedented lines.
        - Only apply dedent when doctest prompts were actually stripped (to avoid changing non-doctest pastes).
        - Only treat `...` as a continuation prompt when a `>>>` is present in the same block (avoids confusing Python Ellipsis with doctest continuation).
- `test_inputtransformer2_line.py`
    - Added/updated tests covering single-line doctest, indented `>>>`, multi-line `>>>` + `...` doctest, and standalone ... (unchanged when no `>>>`).

## Description of changes

- Update `IPython/core/inputtransformer2.py` to enhance the `PromptStripper`:
  - Detect doctest-mode for a pasted block when any line matches `^\s*>>>`.
  - In doctest-mode, strip indented doctest prompts:
    - `^\s*>>>\s?` (primary prompt)
    - `^\s*\.\.\.\s?` (continuation prompt)
  - After stripping doctest prompts, apply `textwrap.dedent()` to the joined block and return the dedented lines.
  - Only apply `dedent()` when doctest prompts were actually stripped to avoid changing normal paste semantics.
- Update/add tests in `tests/test_inputtransformer2_line.py` for the new behavior.

## Motivation and context

Users often copy doctest examples from documentation or tools that show doctest prompts and indentation (for example, xdoctest-style indented examples). Previously IPython removed the `>>>` prompt but preserved the remaining indentation, which could turn a pasted, single-line doctest into an indented top-level statement and raise `IndentationError`. This PR makes doctest pastes behave as users expect: prompts are removed and the code block is normalized by dedenting.

### Single-line doctest

Input:

```py
>>>     print(1)
```

Before:

```py
    print(1)
```

After:

```py
print(1)
```

### Indented xdoctest-style prompt

Input:

```py
    >>>     print(1)
```

After:

```py
print(1)
```

### Multi-line doctest with continuation

Input:

```py
>>> for i in range(2):
...     print(i)
```

After:

```py
for i in range(2):
    print(i)
```

### Standalone continuation prompt (unchanged)

Input:

```py
...     print(1)
```

After (unchanged because no `>>>` present):

```py
...     print(1)
```

## Implementation notes

- `dedent()` is applied only after doctest prompts were detected and stripped. This avoids affecting normal paste behavior for non-doctest code, and prevents accidental removal of meaningful indentation in user-pasted code.
- `...` is considered a continuation prompt only when the pasted block contains at least one `>>>`, to avoid confusing Python's `Ellipsis` literal with doctest continuations.
- Regexes used are tolerant of leading whitespace to support indented `>>>` prompts (xdoctest-style).

## Related issues

- Fixes #15089

## Testing performed

- Added and updated tests in `tests/test_inputtransformer2_line.py` to cover:
  - Single-line doctest with extra indentation: `">>>     print(1)\n" -> "print(1)\n"`
  - Leading whitespace before `>>>`: `"    >>>     print(1)\n" -> "print(1)\n"`
  - Multiline doctest using `>>>` and `...` continuation lines.
  - Standalone `...` only lines remain unchanged.

## Doc

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=191372963

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -67,9 +67,18 @@ class PromptStripper:
     If any prompt is found on the first two lines,
     prompts will be stripped from the rest of the block.
     """
-    def __init__(self, prompt_re, initial_re=None):
+    def __init__(self, prompt_re, initial_re=None, *, doctest=False):
         self.prompt_re = prompt_re
         self.initial_re = initial_re or prompt_re
+        self.doctest = doctest
+        if doctest:
+            # Doctest/xdoctest prompts may be indented (e.g. "    >>>").
+            # We only treat "..." as a continuation prompt when the same pasted
+            # block contains at least one ">>>" line, to avoid ambiguity with the
+            # Python Ellipsis literal.
+            self._doctest_initial_re = re.compile(r'^\s*>>>')
+            self._doctest_ps1_re = re.compile(r'^\s*>>>\s?')
+            self._doctest_ps2_re = re.compile(r'^\s*\.\.\.\s?')
 
     def _strip(self, lines):
         return [self.prompt_re.sub('', l, count=1) for l in lines]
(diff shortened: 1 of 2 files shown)


marcoscoladev:
Hi @krassowski,

All required checks are passing.
This PR implements the fix described in “Removing leading indentation when pasting in a doctest snippet” (#15089).
When you have a moment, could you please review it and merge if everything looks good?

Thank you very much for your time and for maintaining IPython.

krassowski:
What happens when pasting into a docstring? Does it distinguish these cases?

marcoscoladev:
Good point — preserving doctest prompts inside docstrings makes sense.

I’ll update this so that when the paste occurs inside a triple-quoted string (docstring), we **skip doctest stripping and dedenting**, and only apply the transformation in normal code contexts.

I’ll add a targeted test case covering pasting doctest snippets inside a docstring to ensure the prompts remain intact.

marcoscoladev:
@krassowski I’ve pushed an update that preserves doctest prompts when pasting inside triple-quoted strings and added a targeted test covering this case.

Carreau:
Thanks, let's try it.

PR #15122, opened 2026-02-05
Fix encoding to utf8 for autoreload

amirebrahimi:
Addresses issue #15116 by setting utf8 as the default encoding since this will not come for all platforms until Python 3.15.


Carreau:
thanks !

Carreau:
meeseeksdev backport to 9.10.x

PR #15144, opened 2026-02-25
Have a static banner when setting SOURCE_DATE_EPOCH

Carreau:
This should help with things like

jupyterlab/jupyterlab#18552, and allow screenshots to be reproducible.


krassowski (approved):
Thanks!

PR #15145, opened 2026-02-26
fix: replace 97 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


Carreau:
Please don't do codebase wide search and replace like that; and please show that you have at lease check each of these; some might have to be BaseExecption maybe, and each have to be carefully considered. Please send at max one or two file; I can't review 40+ files at once.

PR #15156, opened 2026-03-14
solves #14901 illegible yellow string on yellow background issue

### LINKED ISSUE #14901: IPython 9.x produces illegible yellow string on yellow background in tracebacks
Originally reported in https://github.com/jupyterlab/jupyterlab/issues/17576.

In IPython 9.x it is

```
[33;43mhello
```

which means:
- 33: set yellow foreground
- 43: set yellow background

whereas in IPython 8.x it was

```
[38;5;124;43mhello
```

which means:
- 38: set foreground to 124 (red-ish)
- 43: set yellow background

Image

GSAUC3:
Fixes #14901
Before:

After:

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -519,6 +519,15 @@ class VerboseTB(TBTools):
 
         self.skip_hidden = True
 
+        theme_highlight_map = {
+            "linux": "ansiblack bg:ansiyellow",
+            "neutral": "ansiblack bg:ansiyellow", 
+            "lightbg": "bg:ansiblue",
+            "nocolor": "", 
+        }
+
+        self.tb_highlight = theme_highlight_map.get(self._theme_name, "bg:ansiyellow")
+
     def format_record(self, frame_info: FrameInfo) -> str:
         """Format a single stack frame"""
         assert isinstance(frame_info, FrameInfo)
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -450,7 +450,7 @@ class VerboseTB(TBTools):
     traceback, to be used with alternate interpreters (because their own code
     would appear in the traceback)."""
 
-    tb_highlight = "bg:ansiyellow"
+    tb_highlight = "ansiblack bg:ansiyellow"
     tb_highlight_style = "default"
 
     _mode: str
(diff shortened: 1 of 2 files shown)


themavik (commented):
nit: `ansiblack bg:ansiyellow` fixes contrast on yellow-on-yellow; if a theme maps `ansiblack` oddly you get a new edge case — still likely better than the old single-token highlight.

GSAUC3:
Hi @themavik , thank you for your review comment. I understand that this solution isn’t fully theme-independent. Would you prefer that I make it theme-aware, for example, by adjusting the colors based on whether the user is using a light or dark theme?
I also noticed that merging is currently blocked. Could you please let me know what might be causing this? I created a separate branch before opening the PR, so I’d appreciate any guidance on what I might have missed.

Carreau:
unfortunately this works for you, but not on all terminal themes, in some case this is worse:

Screenshot 2026-03-25 at 11 04 09

Maybe update tb_highlight and tb_highlight_style to be settable as part of the theme for flexibilty

GSAUC3:
Hi @Carreau, I’ve updated the implementation to make the traceback highlight theme-dependent for the four built-in themes: Neutral, Linux, LightBG, and NoColor.
```python
theme_highlight_map = {
    "linux": "ansiblack bg:ansiyellow",
    "neutral": "ansiblack bg:ansiyellow",
    "lightbg": "bg:ansiblue",
    "nocolor": "",
}
```
Would this approach be acceptable, or would you prefer a different way of wiring it into the theme system?

Thanks for the review.

Carreau:
You want to update `class Theme` in pycolorize to either:

 - add a `tb_highlight` attribute
 - or allow a `Token.TbHighlight: str` entry in the style dict (I think I prefer this one, but I could be convinced otherwise)

Then in 

```
base_style = theme_table[self._theme_name].as_pygments_style()
style = stack_data.style_with_executing_node(base_style, self.tb_highlight)
```

you can probably do `style_with_executing_node(base_style, theme_table[self._theme_name].get('tb_highlight', self.tb_highlight)` or `theme_table[self._theme_name].as_pygments_style()[Token.TbHighlight]` or similar.

GSAUC3:
Hi @Carreau , I’ve made the requested changes based on your suggestions. I have added `Token.TbHighlight` to the `style_extra` dictionary for each theme. Please let me know if there’s anything else you’d like me to modify. I’d be happy to help. Thanks again for your feedback.

Carreau:
Thanks ! That might not make it to today's release, but probably for the next one.

Carreau (approved):

Carreau (approved):

PR #15158, opened 2026-03-15
fix: preserve sys.exit() exit codes in non-interactive mode

Sarah-2003:
## Summary

Fixes #15132.

When running `ipython -c "import sys;sys.exit(2)"`, IPython returned exit code 1 instead of 2. Regular Python correctly returns 2.

**Root cause:** `TerminalIPythonApp.start()` in `ipapp.py` hardcoded `sys.exit(1)` for all failed executions, regardless of the actual exit code passed to `sys.exit()`.

## Changes

- **`IPython/terminal/ipapp.py`** -- Extract the original exit code from `last_execution_result.error_in_exec` when the error is a `SystemExit`. Falls back to exit code 1 for non-SystemExit failures.
- **`IPython/core/shellapp.py`** -- Add a separate `except SystemExit` clause in the file execution path of `_run_cmd_line_code()` to preserve exit codes when running script files.
- **`tests/test_shellapp.py`** -- Add 8 tests covering exit code preservation for `-c` execution and script file execution (codes 0, 1, 2, 42, no-arg, normal execution).

## How it works

When `sys.exit(N)` is called inside user code, `InteractiveShell.run_code()` catches the `SystemExit` and stores it in `result.error_in_exec`. The fix reads `error_in_exec.code` to extract the original exit code instead of discarding it.

Edge cases handled:
- `sys.exit(0)` and `sys.exit()` -- exit code 0
- `sys.exit(N)` for any integer N -- preserves N
- `sys.exit("message")` -- exit code 1 (matches Python behavior)
- Non-SystemExit failures -- exit code 1 (unchanged)

## Test plan

- [x] `ipython -c "import sys; sys.exit(2)"` returns code 2 (was 1)
- [x] `ipython -c "import sys; sys.exit(42)"` returns code 42 (was 1)
- [x] `ipython -c "import sys; sys.exit(0)"` returns code 0
- [x] `ipython -c "print('hello')"` returns code 0 (unchanged)
- [x] Script file with `sys.exit(2)` returns code 2
- [x] All 3 modified files pass syntax check

### DIFF
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -482,7 +482,6 @@ class InteractiveShellApp(Configurable):
             try:
                 self._exec_file(fname, shell_futures=True)
             except SystemExit as e:
-                self.shell.showtraceback(tb_offset=4)
                 if not self.interact:
                     exit_code = 1
                     if isinstance(e.code, int):
--- a/IPython/core/shellapp.py
+++ b/IPython/core/shellapp.py
@@ -481,6 +481,15 @@ class InteractiveShellApp(Configurable):
                     self.exit(2)
             try:
                 self._exec_file(fname, shell_futures=True)
+            except SystemExit as e:
+                self.shell.showtraceback(tb_offset=4)
+                if not self.interact:
+                    exit_code = 1
+                    if isinstance(e.code, int):
+                        exit_code = e.code
+                    elif e.code is None:
+                        exit_code = 0
+                    self.exit(exit_code)
             except:
                 self.shell.showtraceback(tb_offset=4)
                 if not self.interact:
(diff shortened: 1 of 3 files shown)


themavik (commented):
the new `except SystemExit` in `_run_cmd_line_code` still calls `showtraceback` before exiting — behavior change vs a quiet propagate; might be noisy for intentional `sys.exit` in non-interactive runs.

Sarah-2003:
Good catch @themavik. sys.exit() should exit quietly, not print a traceback. Removed the showtraceback() call from the except SystemExit handler. The bare except: below still prints tracebacks for actual errors.

Carreau:
Thanks for the contribution;  I'm not sure why the tests are failing on CI and passing locally

Carreau:
You may want to convert those subprocess test to simply use mock and thest the function themselves, it will be faster this way to test as well.

Issue #15161, opened 2026-03-18
Flaky test - teardown in test_hist_file_config


frenzymadness:
We build ipython in Fedora Linux as an RPM package, and we've recently hit the following issue in tests:

```
==================================== ERRORS ====================================
__________________ ERROR at teardown of test_hist_file_config __________________

    @pytest.fixture()
    def inner():
        assert HistoryManager._max_inst == 1
        HistoryManager._max_inst = N
        lh = len(HistoryManager._instances)
        try:
            yield
            gc.collect()
>           assert len(HistoryManager._instances) == lh
E           AssertionError: assert 2 == 1
E            +  where 2 = len({<weakref at 0x7f4e89744cc0; to 'IPython.core.history.HistoryManager' at 0x7f4e8bcc96a0>, <weakref at 0x7f4e882850d0; to 'IPython.core.history.HistoryManager' at 0x7f4e88764050>})
E            +    where {<weakref at 0x7f4e89744cc0; to 'IPython.core.history.HistoryManager' at 0x7f4e8bcc96a0>, <weakref at 0x7f4e882850d0; to 'IPython.core.history.HistoryManager' at 0x7f4e88764050>} = HistoryManager._instances

../tests/test_history.py:47: AssertionError
```

The problem is that it's very hard to debug because it fails in our build system only in like 3 out of 10 builds. Do you have any idea what might be causing this?

frenzymadness:
I've created an unrelated PR, and I see that the CI there failed for the same reason on macos-latest with Python 3.12: https://github.com/ipython/ipython/actions/runs/23286987693/job/67712754942?pr=15163

Carreau:
You an probably enable tracemalloc and run it in a loop and try to figure out where was the history manager that is leaking.

Maybe use something like https://objgraph.readthedocs.io/en/stable/ to see the graph of objects

PR #15163, opened 2026-03-19
Fix expected length in test_unicode_range for Python 3.15 alpha 7

frenzymadness:
The range itself had been updated in 63476b2, but the test started to fail with the latest alpha 7 release of Python 3.15.


frenzymadness:
The CI failure on macos-latest is unrelated to this change, and it's already reported as https://github.com/ipython/ipython/issues/15161

PR #15168, opened 2026-03-26
fix: only surface nested repr_mimebundle NotImplementedError

Maa-ly:
Closes #15154

## Summary
- keep swallowing direct `NotImplementedError` opt-outs from repr methods so interactive fallbacks keep working
- surface nested `_repr_mimebundle_` `NotImplementedError`s from helper code instead of silently discarding them
- add regression tests covering both the direct opt-out path and the nested helper-error path

## Validation
- `source .venv/bin/activate && pytest tests/test_formatters.py -q`

### DIFF
--- a/IPython/core/formatters.py
+++ b/IPython/core/formatters.py
@@ -281,7 +281,15 @@ def catch_format_error(method, self, *args, **kwargs):
     try:
         r = method(self, *args, **kwargs)
     except NotImplementedError:
-        # don't warn on NotImplementedErrors
+        if getattr(self, "catch_notimplemented", True):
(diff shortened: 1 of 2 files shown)


Carreau (approved):
+1 in principle (reviewing from my phone) i need to check and think about reaching for the global ipython instance.

Maa-ly:
The previous version made every `_repr_mimebundle_` `NotImplementedError` visible, which is what broke the downstream `pyflyby` debugger test. This update narrows the change to nested helper errors only, while preserving the long-standing direct opt-out path that interactive fallbacks rely on.

Local validation: `source .venv/bin/activate && pytest tests/test_formatters.py -q`.

PR #15176, opened 2026-04-05
fix: resolve #12105

### LINKED ISSUE #12105: How can I disable filename abbreviations in the results of tab-autocompletion?

venturevd:
## Summary
Automated fix for #12105

## Changes
Modified files: IPython/terminal/ptutils.py

## Test Results
All existing tests pass after this change.

---
*This PR was generated by [Agora Farm](https://github.com/venturevd), an autonomous software engineering system. Please review carefully and provide feedback.*

### DIFF
--- a/IPython/terminal/ptutils.py
+++ b/IPython/terminal/ptutils.py
@@ -50,7 +50,7 @@ def _elide_point(string: str, *, min_elide) -> str:
     if file_parts[-1] == '':
         file_parts.pop()
 
-    if len(object_parts) > 3:
+    if len(object_parts) > 3 and len(string) >= min_elide:
         return "{}.{}\N{HORIZONTAL ELLIPSIS}{}.{}".format(
             object_parts[0],
             object_parts[1][:1],
@@ -58,7 +58,7 @@ def _elide_point(string: str, *, min_elide) -> str:
             object_parts[-1],
         )
 
-    elif len(file_parts) > 3:
+    elif len(file_parts) > 3 and len(string) >= min_elide:
         return ("{}" + os.sep + "{}\N{HORIZONTAL ELLIPSIS}{}" + os.sep + "{}").format(
             file_parts[0], file_parts[1][:1], file_parts[-2][-1:], file_parts[-1]
         )


Carreau:
Hi, 

I suppose you are an automated agent, thanks for your contribution.

You description is not useful though, it is factually correct but does not bring any information while reading it. 

Remember that the commit will be read outside of the context of GitHub without access to the original issue you must:

 - Describe the problem and the fix (why, how)
 - Avoid information that are already containd in VCS itself (which file is changed is irrelevant the info is given by git-diff
 - Whether the test are passing is not for you to decide in the commit message, it will be verified by CI.

PR #15184, opened 2026-04-08
Implement support for image/png format in terminal

### LINKED ISSUE #13287: Revive PR : Support for image/png format in the terminal
This issue is opened to remind us that #10610 is still quite interesting but need to be done in a more flexible way.

danielzgtg:
Closes #13287

Now `PIL.Image` and `sympy.init_printing(use_latex="png")` show images. The Kitty graphics protocol is used as [suggested](https://github.com/ipython/ipython/pull/10610#issuecomment-630484667).

Alternative implementations:
* Putting a [mime renderer extension](https://ipython.readthedocs.io/en/stable/config/shell_mimerenderer.html#mime-renderer-extensions) in `ipython_config.py` is inconvenient, and someone requested a [native](https://github.com/ipython/ipython/pull/10610#issuecomment-795792151) feature
* Probing for Kitty graphics support would detect any compatible terminal, but hardcoding terminal executable names avoids startup slowdown
* Adding tmux support via Unicode-based Kitty graphics would be very complicated

Alternatives protocols:
* iTerm2 graphics don't work in Ghostty or st
* Sixel requires complicated rasterization and dithering

# Before

text only output

# After

graphical output

### DIFF
--- a/IPython/core/kitty.py
+++ b/IPython/core/kitty.py
@@ -1,0 +1,73 @@
+# Implements https://sw.kovidgoyal.net/kitty/graphics-protocol/
+
+from base64 import b64encode, b64decode
+import sys
+from typing import Union
+
+def _supports_kitty_graphics() -> bool:
+    import platform
+
+    if platform.system() not in ("Darwin", "Linux"):
+        return False
+
+    if not sys.stdout.isatty():
+        return False
+    # Hardcoding process names instead of using
+    # https://sw.kovidgoyal.net/kitty/graphics-protocol/#querying-support-and-available-transmission-mediums
+    # to avoid startup slowdown
+    supported_terminals = {
+        "ghostty",
+        "iTerm2",
+        "kitty",
+        "konsole",
+        "warp",
+        "wayst",
+        "wezterm-gui",
+    }
+    import psutil
+
+    process = psutil.Process()
+    while process := process.parent():
+        if process.name() in supported_terminals:
+            return True
+    return False
+
+
+supports_kitty_graphics = _supports_kitty_graphics()
+
+
+def png_to_kitty_ansi(png: bytes) -> str:
+    if not png.startswith(b"\x89PNG\r\n\x1a\n"):
+        raise ValueError
+    # This simplicity resembles
+    # https://sw.kovidgoyal.net/kitty/graphics-protocol/#a-minimal-example
+    # but if we need tmux support, we can switch to Unicode like
+    # https://github.com/hzeller/timg/blob/main/src/kitty-canvas.cc
+    result = ["\033_Ga=T,f=100,", "m=1;"]
+    encoded = b64encode(png)
+    for i in range(0, len(encoded), 4096):
+        result.append(encoded[i : i + 4096].decode("ascii"))
+        result.append("\033\\")
+        result.append("\033_G")
+        result.append("m=1;")
+    del result[-2:]
+    result[-3] = "m=0;"
+    return "".join(result)
+
+
+def kitty_png_render(png: Union[bytes, str], _md_dict: object) -> None:
+    if isinstance(png, str):
+        png = png_to_kitty_ansi(b64decode(png))
+    else:
+        png = png_to_kitty_ansi(png)
+    print(png)
+
+
+display_formatter_default_active_types = [
+    "text/plain",
+    *(["image/png"] if supports_kitty_graphics else []),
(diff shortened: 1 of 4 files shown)


danielzgtg:
# Effects on Matplotlib

No Matplotlib changes are required in this version, which uses `%matplotlib inline` not `plt.show()`.

This allows non-GUI terminal backends like the Kitty graphics protocol integration in IPython to work. Previously, there was just a warning and no output. Now, the graph is rendered to Kitty-compatible terminals as if it were a Jupyter Notebook.

Inspiration: https://github.com/Kabilan108/ipython-icat/tree/main

- Why is this change necessary? This graphics had been invisible in terminal, leaving only a NonGuiException warning. Additionally, some people are running Matplotlib in headless environments. This change makes the graphics visible for us.
- What problem does it solve? In the terminal, closing the popup GUI window causes the graph to be lost, unlike Jupyter Notebook where it stayed in the scrollback. I also need a minimal-dependency implementation because some projects have tricky venvs with conflicting Qt and other dependencies.
- What is the reasoning for this implementation? Previous requests were rejected for putting terminal code in "core matplotlib". Forwarding the request to IPython provides the necessary level of abstraction, simplification, and code deduplication.

## Before

matplotlib descirption only no graphics in terminal

## After

matplotlib plot visible in terminal

Carreau:
cursory review from phone, and comment to myself as well for later: that's cool; and thanks for doing it properly using the hooks. Maybe it should be gated behind an option that depends on wether the user set their terminal emulator instead of enabling kitty by default? i'm going to assume support for other emulator may pop up as well. i know there is an issue open to implement that; i need to check it's properly crosslinked. i also don't know which is better the octal escape code or the hex ones.

any way to detect we are in kitty ?

danielzgtg:
> [...] any way to detect we are in kitty ?

I have a check for ancestor process names at https://github.com/ipython/ipython/pull/15184/changes#diff-1ebf983c373f968661eca7f4cd9dbf65708d599818fc59a33131ce14825049a3R20-R25 .

For macOS support, can someone post the output of `ps aux` inside iTerm2 please?

This feature isn't enabled for most users by default. Most people are either on Windows (not right `platform`) or in Jupyter Notebook (not `isatty`), where this gets disabled. Most Linux users use gnome-terminal/ptyxis, where I don't enable this either. Btw, I'm using Konsole instead of Kitty, but I tested on both and it's similar. There might be a false positive if someone starts a non-compatible terminal from a compatible terminal, but querying the terminal for its capabilities via control sequences might be too slow. In the rare event that a problem occurs, users should just put `c.TerminalInteractiveShell.mime_renderers = {}` in `~/.ipython/profile_default/ipython_config.py`.

Carreau:
```
In [12]: while pp:
    ...:     pp = pp.parent()
    ...:     print(pp)
    ...:
psutil.Process(pid=30160, name='zsh', status='running', started='2026-03-25 10:14:37')
psutil.Process(pid=30159, name='login', status='sleeping', started='2026-03-25 10:14:37')
psutil.Process(pid=979, name='iTermServer-3.6.9', status='running', started='2026-03-19 10:52:02')
psutil.Process(pid=679, name='iTerm2', status='running', started='2026-03-18 00:50:59')
psutil.Process(pid=1, name='launchd', status='sleeping', started='1970-01-01 01:05:13')
psutil.Process(pid=0, name='kernel_task', status='idle', started='1970-01-01 01:05:13')
None
```

Carreau:
II was thinking of an ENQ request/reply (so that it work via SSH, but iterm2 does not reply to enq, and we can set do that as a subsequent PR anyway.

danielzgtg:
> `psutil.Process(pid=979, name='iTermServer-3.6.9', status='running', started='2026-03-19 10:52:02')`
> `psutil.Process(pid=679, name='iTerm2', status='running', started='2026-03-18 00:50:59')`

To avoid introducing a separate `.startswith`, I'm checking for "iTerm2". I hope I'm correct in assuming the "iTermServer-..." process never gets orphaned during normal operation.

Carreau:
Just notting something, apparently you can look at $TERM_PROGRAM that a few terminal emulator will set; I don't know which.

danielzgtg:
> $TERM_PROGRAM

That is not set in Konsole and Kitty—the two terminals I'm focusing on.

Carreau:
Let's try; it's self contained enough.

PR #15185, opened 2026-04-10
Add %xmode Doctest mode for doctest friendly tracebacks

Jatin-Shihora:
Adds a new `%xmode Doctest` mode that formats tracebacks for easy copy-paste into Python doctests. The output shows only the traceback header, a literal ellipsis, and the exception line:

It will look like this when set to doctest:
```
    ...
```

This is useful because Python's `doctest` module accepts `...` as a wildcard in expected traceback output, so this format lets users copy-paste IPython output directly into doctests.

#### changes

- Added `Doctest` mode to `FormattedTB.valid_modes` and routing logic in `ultratb.py`
- Added `structured_traceback_doctest` method to `ListTB`
- Updated `xmode` config enum in `interactiveshell.py`
- Updated `%xmode` magic docstring in `basic.py`
- Added test in `test_ultratb.py`
* Fixed `test_xmode` in `test_magic.py` to avoid hardcoding the mode count. So previously, `5` was hardcoded in the range, so updated it to use the mode count based on the size of the list.
- Added whatsnew entry

Closes #14925

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -579,7 +579,7 @@ class InteractiveShell(SingletonConfigurable):
     separate_out2 = SeparateUnicode('').tag(config=True)
     wildcards_case_sensitive = Bool(True).tag(config=True)
     xmode = CaselessStrEnum(
-        ("Context", "Plain", "Verbose", "Minimal", "Docs"),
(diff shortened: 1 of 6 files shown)


Jatin-Shihora:
Hi @Carreau, just a friendly follow-up on the review when you have a chance. Thanks!

Carreau:
Sure,let's try; thanks for the reminder.

PR #15190, opened 2026-04-24
avoid psutil requirement on emscripten

bollwyvl:
Congratulations on 9.13. :tada: 

This builds on #15184 by adding an environment marker to the hard `psutil` dependency for  emscripten environments. To my knowledge, no current approach offers processes.


Carreau (approved):

Carreau:
Thanks !

PR #15194, opened 2026-04-27
Update _tokens_filename with the option to not compress user path

Carreau:
This can be desirable in some case,
in particular in complex deployment where the filesystem of the server and the kernel don't match to be able to open the file path from the UI.

This will later be extended with global IPython configuration options, maybe at runtime.

See Quansight/deshaw#259


Carreau:
I'm also considering having it never compress user and make it the business of the called to compress or not.

krassowski:
The idea of https://github.com/jupyter-server/jupyter_server/pull/1331 is to not have to make these changes in kernels.

PR #15199, opened 2026-05-04
Post link to what's new in zulip. 

Carreau:


jasongrout (approved):
Cool! I didn't realize this is where the zulip post came from. Thanks!

Carreau:
it's org wide credential; it could almost be nice to have that automatically done in jupyter releaser...

PR #15205, opened 2026-05-10
Add singular alias for ipdb exceptions command

ghost:
## Summary
- add `exception` as an alias for the existing `exceptions` ipdb command
- cover the alias with a debugger unit test

Fixes #14870

## Test
- `uv run --extra test pytest tests/test_debugger.py::test_ipdb_exception_command_alias`

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -515,6 +515,12 @@ class Pdb(OldPdb):
                 else:
                     self.error("No exception with that number")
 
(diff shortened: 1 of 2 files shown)


FahadAshraf29:
The failing downstream tests on macOS 14 / Python 3.13 appear to be pre-existing issues unrelated to this change,  specifically traceback format changes in newer IPython versions and a timing-related test. These tests fail on main as well. Happy to help investigate if needed.

PR #15206, opened 2026-05-11
Accept singular pdb exception command

tdashelby-cmyk:
## Summary
- add `exception` as an alias for the existing `exceptions` chained-exception debugger command
- cover the command dispatch path with a regression test

Closes #14870

## Tests
- `.\.venv312\Scripts\python.exe -m pytest tests/test_debugger.py -q`

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -515,6 +515,12 @@ class Pdb(OldPdb):
                 else:
                     self.error("No exception with that number")
 
+        def do_exception(self, arg):
+            """exception [number]
+            Alias for the ``exceptions`` command.
+            """
+            return self.do_exceptions(arg)
+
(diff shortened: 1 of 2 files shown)


tdashelby-cmyk:
The remaining failing check is the downstream `test (macos-14, 3.13)` job. The failure is in pyflyby `tests/test_interactive.py::test_error_during_completion_1[prompt_toolkit]`, where the pty transcript differs around completion/error output.

This PR only adds the singular `exception` debugger command alias under the existing chained-exception command path; that path is skipped on Python 3.13, and the Ubuntu 3.13 downstream job passed. I tried rerunning the failed job, but GitHub requires repository admin permissions for that action.

Carreau:
thanks

PR #15208, opened 2026-05-12
Fix banner customization when SOURCE_DATE_EPOCH is set

### LINKED ISSUE #15207: banner cannot be customized when SOURCE_DATE_EPOCH is set
Normally one can customize the banner via `banner1` parameter, as in:
```
$ cat ipython_banner.py 

shell=InteractiveShell(banner1="MY CUSTOM BANNER", enable_tip=False)

print(shell.banner)
$ python ipython_banner.py 
MY CUSTOM BANNER
```
However, this is ignored when running with `SOURCE_DATE_EPOCH` set:
```
$ SOURCE_DATE_EPOCH=1234567890 python ipython_banner.py 
Python 3.y.z | Packaged with love | (main, Friday, 13 February 2009) [Compiler]
IPython 9.y.z -- An enhanced Interactive Python. Type '?' for help.
Tip: unset SOURCE_DATE_EPOCH to restore dynamic banner.

```

This affects e.g. doctesting in sagemath.

Perhaps the static banner can be used only when `self.banner1` is `default_banner`, so if `banner1` is set it will override the static banner?

xavya-dev:
Fixes #15207

Updated the banner logic to prevent an early return when SOURCE_DATE_EPOCH is set. This ensures that user-defined custom banners are displayed correctly instead of being locked.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1026,7 +1026,7 @@ class InteractiveShell(SingletonConfigurable):
     def banner(self):
         banner = self.banner1
         # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner (default starts with "Python")
-        if banner.startswith("Python") and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
+        if banner is self.banner1 and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
             from datetime import datetime
             date = datetime.fromtimestamp(int(when))
             banner = textwrap.dedent(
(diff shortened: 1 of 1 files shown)


tornaria:
Works for me, thanks!

tornaria:
Another (more robust?) way would be to use `banner is default_banner` as a test, instead of `banner.startswith("Python")`. Either way is ok.

xavya-dev:
@tornaria Thank you for the feedback! I have updated the logic to make it more robust.

Carreau:
Thanks

collares:
I'm confused. The patch that got committed starts with:

```
        banner = self.banner1
        if banner is self.banner1 and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
```

Isn't the first part of the condition always true, given the preceding assignment?

xavya-dev:
"Thanks for pointing this out, @collares! I have opened a new PR to fix this redundant condition: https://github.com/ipython/ipython/pull/15314"

PR #15211, opened 2026-05-14
fix: duplicated words in usage.py and oinspect.py docstrings

lphuc2250gma:
Two one-line typo fixes for duplicated words in docstrings:
- `IPython/core/usage.py` — "IPYTHONDIR defaults to to `$HOME/.ipython`." → "IPYTHONDIR defaults to `$HOME/.ipython`."
- `IPython/core/oinspect.py` — "This is is helper function to skip displaying the docstring if the" → "This is a helper function to skip displaying the docstring if the"

No code/behavior change.


krassowski (approved):
Thanks!

PR #15216, opened 2026-05-21
docs: clarify autoawait example requires running event loop

### LINKED ISSUE #14865: Documentation for autoawait gives example that doesn't work
The documentation for autoawait (ie. allowing automatic use of an event loop) at https://ipython.readthedocs.io/en/stable/interactive/autoawait.html gives some examples like:

    In [1]: import aiohttp
       ...: session = aiohttp.ClientSession()
       ...: result = session.get('https://api.github.com')

    In [2]: response = await result

However, this doesn't actually work as written - the first block will raise `RuntimeError: no running event loop.`

I think this is because no event loop is created unless the code being run actually contains an `await` instruction or similar, so code that requires the event loop to be running, but doesn't itself await will fail.
Eg. modifying the above to include the `response = await result` in the same block as the setup logic **does** work.

Ideally there'd be a way to handle code like this - I'm not sure if always running an event loop (possibly configurable) has problems.  If not, I think at least the examples should be amended to mention this.

SAY-5:
Fixes #14865.

The `aiohttp.ClientSession()` example in `autoawait.rst` raises `RuntimeError: no running event loop` because cell `In [1]` contains no `await`, so IPython has not yet started the loop when `ClientSession()` is constructed.

Per @Carreau's suggestion on the issue ("the easier way to force a loop is to await a dummy thing, like `await sleep(0)`"), this prepends `await asyncio.sleep(0)` to the first cell and adds a short note explaining why it is needed.

### DIFF
--- a/docs/source/interactive/autoawait.rst
+++ b/docs/source/interactive/autoawait.rst
@@ -36,6 +36,14 @@ syntax error in the Python REPL::
                               ^
     SyntaxError: invalid syntax
 
+.. note::
+
+    Libraries like ``aiohttp`` require a running event loop at the moment an
+    object such as ``ClientSession`` is constructed. IPython only starts an
+    event loop when a cell contains an ``await`` expression, so the example
+    below uses ``await asyncio.sleep(0)`` to force the loop to start before
+    ``aiohttp.ClientSession()`` is called.
+
 Should behave as expected in the IPython REPL::
 
     Python 3.12.0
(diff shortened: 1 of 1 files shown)


SAY-5:
The `Build docs` failure is pre-existing: the workflow is also failing on recent `main` branch runs (not caused by this PR). The RST change here only adds a note block and two REPL lines.

PR #15220, opened 2026-05-22
Fix %debug and ipdb with Python 3.15

### LINKED ISSUE #15217: ipdb does not work with Python 3.15 beta 1
AFAIU, Python now by default uses the new PyREPL for pdb, see: https://github.com/python/cpython/commit/234c12c0fc6f9710cee34d033a7f6a808d74bfac

And that seems to be the case why `ipdb` or `%debug` does not work in 3.15 beta 1:

```
Python 3.15.0b1 (main, May 11 2026, 00:00:00) [GCC 16.1.1 20260501 (Red Hat 16.1.1-1)]
IPython 9.14.0.dev -- An enhanced Interactive Python. Type '?' for help.
Tip: Use `object?` to see the help on `object`, `object??` to view its source

In [1]: 2/0
Cell In[1], line 1
----> 1 2/0

In [2]: %debug
Cell In[2], line 1
----> 1 get_ipython().run_line_magic('debug', '')

File /usr/lib64/python3.15/pdb.py:934, in Pdb.interaction(self, frame, tb_or_exc)
    928 # We should print the stack entry if and only if the user input
    929 # is expected, and we should print it right before the user input.
    930 # We achieve this by appending _pdbcmd_print_frame_status to the
    931 # command queue. If cmdqueue is not exhausted, the user input is
    932 # not expected and we will not print the stack entry.
    933 self.cmdqueue.append('_pdbcmd_print_frame_status')
--> 934 self._cmdloop()
    935 # If _pdbcmd_print_frame_status is not used, pop it out
    936 if self.cmdqueue and self.cmdqueue[-1] == '_pdbcmd_print_frame_status':

File /usr/lib64/python3.15/pdb.py:799, in Pdb._cmdloop(self)
    797 self.allow_kbdint = True
    798 with self._maybe_use_pyrepl_as_stdin():
--> 799     self.cmdloop()
    800 self.allow_kbdint = False
    801 break

ValueError: Sorry ipdb does not support use_rawinput=False
```

frenzymadness:
Python 3.15 introduced PyREPL as the default pdb input backend (via _maybe_use_pyrepl_as_stdin()), which temporarily sets use_rawinput=False before calling cmdloop(). IPython's TerminalPdb raises ValueError when it sees use_rawinput=False.

Override _cmdloop() in IPython.core.debugger.Pdb to bypass the PyREPL wrapping — IPython already has its own input handling via prompt_toolkit and doesn't need PyREPL.

Fixes: https://github.com/ipython/ipython/issues/15217

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -540,6 +540,18 @@ class Pdb(OldPdb):
             """
             return self.do_exceptions(arg)
 
+    def _cmdloop(self):
+        # Override to bypass Python 3.15's _maybe_use_pyrepl_as_stdin(), which
+        # sets use_rawinput=False and conflicts with IPython's own input handling.
+        while True:
+            try:
+                self.allow_kbdint = True
+                self.cmdloop()
+                self.allow_kbdint = False
+                break
+            except KeyboardInterrupt:
+                self.message("--KeyboardInterrupt--")
+
     def interaction(self, frame, tb_or_exc):
         try:
             if CHAIN_EXCEPTIONS:


Carreau:
@copilot resolve the merge conflicts in this pull request

PR #15223, opened 2026-05-25
Honor quotes when glob-expanding %run arguments

### LINKED ISSUE #12726: `%run` incorrectly expands glob patterns when quoted
I think I discovered a bug in how `%run` works with argv.  `%run` will expand glob patterns even when quoted.

### Demo:

First, in terminal:

    $ touch foo.txt
    $ touch bar.txt

Then make the program `demo.py`:

```python
import sys
print(sys.argv[1:])
```

From the command line:

    $ python demo.py *.txt
    ['bar.txt', 'foo.txt']

    $ python demo.py "*.txt"
    ['*.txt']

This is ad you'd expect.

From iPython:

```python
In [1]: %run demo.py *.txt
['foo.txt', 'bar.txt']

In [2]: %run demo.py "*.txt"
['foo.txt', 'bar.txt']
```

The latter should *not* be expanded!

### Versions

I didn't have time to test this on multiple platforms but this was on macOS Catalina with Anaconda Python:

    Python 3.8.3 (default, Jul  2 2020, 11:26:31)
    IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

Thanks!

Ijtihed:
Fixes #12726.

## Summary

Wrapping a `%run` argument in single or double quotes now suppresses shell-style glob expansion of that argument matching how real shels behave. the existing backslash-escape (`\*`) and `-G` flag are unchanged. Before this with `foo.txt` and `bar.txt` in the working directory:

```python
In [1]: %run script.py "*.txt"
# sys.argv -> ['foo.txt', 'bar.txt']
```

After:

```python
In [1]: %run script.py "*.txt"
# sys.argv -> ['*.txt']
```

## Motivation

The issue thread already has a maintainer pointer:

> Looks like an oversight in `IPython.core.magics.execution.ExecutionMagics.parse_options`.

-- @MrMino, [comment](https://github.com/ipython/ipython/issues/12726#issuecomment-823672101)

## What changed

I added a new helper `arg_split_with_quotes(commandline, strict=True)` in `IPython/utils/_process_common.py` (re-exported from `IPython.utils.process`). Returns `[(token, was_quoted), ...]`. shlex-based on two passes - `posix=True` and `posix=False` - so the quote semantics are the same on every platform

`%run` consults it after `parse_options` and skips globbing for any `arg_lst` entry that came from a quoted token in `parameter_s`. Docstring updated.

This is a small backwards-incompatible behavior change so ofc anyone depending on the previous "quotes do not suppress globs" behavior wouldd see different output. I added `docs/source/whatsnew/pr/incompat-run-quoted-glob.rst`.

## Tests

- `tests/test_process.py`: parametrized unit tests for `arg_split_with_quotes` covering bare, double-quoted, single-quoted, mixed-with-options and `strict=False` cases.
- `tests/test_run.py`: new integration test `test_run_quoted_glob_arg_is_not_expanded` that runs `%run -i` against a small script in a tempdir with f les that would match `*.txt`, and checks `sys.argv` for bare double-quoted and (POSIX-only) single-quoted forms.
- Updated `doctest_run_option_parser_for_posix` its expected output was locking in the old buggy behavior.

The POSIX-only single-quote doctest is done by CI on the Linux/macOS jobs.

## Open questions

- The Windows doctest `doctest_run_option_parser_for_windows` still describes the existing `'print*.py'` → `["'print*.py'"]` behavior. That case is preserved by the patch (`CommandLineToArgvW` keeps the literal single quotes so they never enter the `quoted_remaining` set) so I left it. 
- Single quotes on Windows stay documented as unsupported.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -61,6 +61,7 @@ from IPython.utils.contexts import preserve_keys
 from IPython.utils.ipstruct import Struct
 from IPython.utils.module_paths import find_mod
 from IPython.utils.path import get_py_filename, shellglob
+from IPython.utils.process import arg_split_with_quotes
 from IPython.utils.timing import clock, clock2
 from IPython.core.magics.ast_mod import ReplaceCodeTransformer
 
@@ -579,11 +580,12 @@ class ExecutionMagics(Magics):
         interactive work, while giving each program a 'clean sheet' to run in.
 
         Arguments are expanded using shell-like glob match.  Patterns
-        '*', '?', '[seq]' and '[!seq]' can be used.  Additionally,
-        tilde '~' will be expanded into user's home directory.  Unlike
-        real shells, quotation does not suppress expansions.  Use
-        *two* back slashes (e.g. ``\\\\*``) to suppress expansions.
-        To completely disable these expansions, you can use -G flag.
+        '*', '?', '[seq]' and '[!seq]' can be used, and tilde '~' is
+        expanded to the user's home directory.  As in real shells,
+        wrapping an argument in single or double quotes suppresses glob
+        expansion for that argument (see #12726).  You can also use
+        *two* back slashes (e.g. ``\\\\*``) outside of quotes, or pass
+        the ``-G`` flag to disable expansion entirely.
 
         On Windows systems, the use of single quotes `'` when specifying
         a file is not supported. Use double quotes `"`.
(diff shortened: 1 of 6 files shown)


Ijtihed:
@Carreau anything needed from my end? 👯

Carreau:
I just need to find time, sit down and think about the consequences and if we actually want to do that, and whether this may break something else.

Carreau:
I think this is fine ; thanks !

PR #15232, opened 2026-06-01
Annotate get_ipython

jenshnielsen:
I am using this in my code and it would be convenient if it was annotated.

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -220,7 +220,7 @@ class Pdb(OldPdb):
 
     """
 
-    shell: InteractiveShell | None
+    shell: InteractiveShell
     _theme_name: str
     _context: int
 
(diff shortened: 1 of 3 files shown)


IPython/core/debugger.py
```diff
@@ -220,7 +220,7 @@ class Pdb(OldPdb):
 
     """
 
-    shell: InteractiveShell
+    shell: InteractiveShell | None
```
jenshnielsen: Allowing this to be None breaks downstream type checking but enforcing this to be not None breaks tests. Happy to look more into this but running out of time right now

jenshnielsen: I think the current solution should work however it looks like type checking is now failing since mypy type checks against numpy 2.5 expecting 3.11 support (numpy2.5 dropped 3.11 support)

jenshnielsen:
@Carreau Issues are resolved and this should be ready to review.

Carreau:
Thanks.

Carreau:
thanks

PR #15235, opened 2026-06-01
Add forward compatibility for pdb.Pdb's mode argument

Carreau:
This PR adds support for the `mode` argument that was introduced to stdlib's `pdb.Pdb` in Python 3.14, ensuring IPython's debugger subclasses remain drop-in replacements across Python versions.

- Updated `IPython.core.debugger.Pdb.__init__` to accept a `mode` keyword argument that specifies how the debugger was invoked (`'inline'`, `'cli'`, or `None`)
  - The argument is forwarded to the underlying `pdb.Pdb` on Python 3.14+
  - On earlier Python versions, it's stored as an instance attribute for compatibility
  - Added docstring documentation for the new parameter

- Added test coverage to verify signature compatibility:
  - `test_pdb_subclass_signature_compatible`: validates that all IPython debugger subclasses (`Pdb`, `InterruptiblePdb`, `TerminalPdb`) accept every argument that stdlib's `pdb.Pdb` accepts (directly or via `**kwargs`)
  - `test_pdb_subclass_accepts_mode`: ensures the `mode` argument is accepted and stored on all supported Python versions. `TerminalPdb` is skipped on Windows for the instantiation case, since it creates a prompt_toolkit session that requires a real console (`NoConsoleScreenBufferError` on headless Windows CI)

Should help with spyder-ide/spyder#25550

Might release that as 9.14.1 (unless 9.15 at end of month is ok).

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -243,6 +243,8 @@ class Pdb(OldPdb):
         stdin=None,
         stdout=None,
         context: int | None | str = 5,
+        *,
+        mode: str | None = None,
         **kwargs,
     ):
         """Create a new IPython debugger.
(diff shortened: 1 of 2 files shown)


IPython/core/debugger.py
ccordoba12:
```suggestion
    def set_trace(self, frame=None, **kwargs):
```

`set_trace` has `commands` as a new kwarg in 3.14+, so we need to take care of that too:

https://github.com/python/cpython/blob/3.14/Lib/pdb.py#L419

ccordoba12:
```suggestion
        return super().set_trace(frame, **kwargs)
```

To go with my suggestion above.

ccordoba12 (commented):
@Carreau, thanks for your help with this!

I tested your changes in Spyder and a couple more are necessary. So, I left suggestions for them.

PR #15242, opened 2026-06-07
Use `print_stack_entry` to print frames in Python 3.14

ccordoba12:
This change is needed to override some changes done in the base `Pdb` class in 3.14, which end up using `print_stack_trace` to print stack entries in IPython:

https://github.com/python/cpython/blob/5b33e9c1763813c459f9857ad3c168692fb95148/Lib/pdb.py#L1258-L1261

It's effect can be seen when using a `breakpoint()` call in Spyder:

**Before**

image

**After**

image

### DIFF
--- a/IPython/core/debugger.py
+++ b/IPython/core/debugger.py
@@ -602,6 +602,19 @@ class Pdb(OldPdb):
         filename = frame.f_code.co_filename
         self.shell.hooks.synchronize_with_editor(filename, lineno, 0)
 
+    def _pdbcmd_print_frame_status(self, arg):
+        """Use print_stack_entry to print frames in Python 3.14+."""
+        if sys.version_info[:2] >= (3, 14):
+            # This is the only line changed from the base class.
+            self.print_stack_entry(self.stack[self.curindex])
+
+            # Same as in 3.14
+            self._validate_file_mtime()
+            self._show_display()
+        else:
+            # 3.13 and 3.12 don't need any changes.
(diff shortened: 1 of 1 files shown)


ccordoba12:
@Carreau, could you consider this for `9.14.2`? I took me some time to understand what was happening, that's why I couldn't submit a PR in time for `9.14.1`.

IPython/core/debugger.py
```diff
@@ -621,6 +602,19 @@ def print_stack_entry(
...
+        if sys.version_info[:2] >= (3, 14):
+            # This is the only line changed from the base class.
+            self.print_stack_entry(self.stack[self.curindex])
+
+            # Same as in 3.14
+            self._validate_file_mtime()
+            self._show_display()
+        else:
+            # 3.13 and 3.12 don't need any changes.
+            super()._pdbcmd_print_frame_status(arg)
```
Carreau:
```suggestion
            super()._pdbcmd_print_frame_status(arg) # type: ignore[misc]
```

Carreau:
I'll see, 9.15 is less then 2 weeks out (monthly release; not sure it's worth doing patches release for 2 weeks max wait, I'll try)

ccordoba12:
> I'll see, 9.15 is less then 2 weeks out (monthly release; not sure it's worth doing patches release for 2 weeks max wait, I'll try)

Sure, no problem with that because this is not so serious.

PR #15243, opened 2026-06-07
Restore underscore tracking after deletion

MUFFANUJ:
Fixes #15000 

Fixes displayhook underscore ownership detection so automatic `_` tracking resumes after `del _`, while still preserving active user-assigned `_`, `__`, and `___` values.


Carreau:
I'd like to have a test, and want to understand how `self.shell.user_ns[unders]` fails if `unders` not in `self.shell.user_ns`

PR #15248, opened 2026-06-10
Fix: ipython directive skips execution inside excluded `only` blocks

### LINKED ISSUE #9339: ipython Sphinx directive executes code when excluded by only, ifconfig
The ``ipython`` directive for Sphinx provided by ``IPython.sphinxext.ipython_directive`` executes its code block at the parsing stage. As a result, even if an ``ipython`` directive is surrounded by an ``only`` or ``ifconfig`` directive with a false-y condition, the code block is still executed. I have posted a [minimum working example](https://gist.github.com/cgranade/3cc6b87c8efef89d54b5) for this issue that includes a Sphinx configuration file and a reStructuredText file to reproduce the problem.

w1ndcn:
Closes #9339

## Problem

The `ipython` Sphinx directive executes code at the **parse stage**. When wrapped in a conditional directive like `only` with a false condition for the current builder, the code still runs:

```rst
.. only:: latex

   .. ipython::

      In [1]: dangerous_operation()
```

Building with `html` builder — the `only:: latex` condition is false, yet `dangerous_operation()` still executes. This is because Sphinx's `only` directive evaluates conditions at the **post-transform stage**, after all child directive `run()` methods have already been called.

## Fix

Add `_is_inside_excluded_only()` to `IPythonDirective`. It walks up the parent node chain via `self.state.parent`, and if any ancestor is an `only` node whose expression evaluates to false for the current builder's tags, returns `True`. When true, `run()` returns `[]` early without executing code or calling setup/teardown.

Key insight: during `nested_parse` inside `Only.run()`, the `only` node is set as the parent in the docutils state machine, so `self.state.parent` is directly the `only` node — no fragile heuristics needed.

## Changes

**`IPython/sphinxext/ipython_directive.py`**

- `IPythonDirective._is_inside_excluded_only()`: walks parent chain, evaluates `only` node expressions against `env.app.tags`
- `IPythonDirective.run()`: early return when inside an excluded `only` block

**`tests/test_ipython_directive.py`** (new)

- `test_only_excluded_skips_execution`: `only:: latex` under html builder → code not executed
- `test_only_included_executes_normally`: `only:: html` under html builder → code executed
- `test_nested_only_skips_execution`: nested `only:: html` > `only:: latex` → inner block skipped

### DIFF
--- a/IPython/sphinxext/ipython_directive.py
+++ b/IPython/sphinxext/ipython_directive.py
@@ -1000,8 +1000,29 @@ class IPythonDirective(Directive):
                                       store_history=False)
         self.shell.clear_cout()
 
+    def _is_inside_excluded_only(self):
+        """Return True if inside an ``only`` node whose condition is false."""
+        try:
+            env = self.state.document.settings.env
+            tags = env.app.tags
+        except AttributeError:
+            return False
+
+        node = self.state.parent
+        while node is not None:
+            if getattr(node, 'tagname', '') == 'only':
+                expr = node.get('expr', '')
+                if expr and not tags.eval_condition(expr):
+                    return True
+            node = getattr(node, 'parent', None)
+        return False
+
     def run(self):
         debug = False
+
+        # gh-9339: skip execution when inside an excluded only block.
+        if self._is_inside_excluded_only():
+            return []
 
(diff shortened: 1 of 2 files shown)


w1ndcn:
Renaming branch — recreating

PR #15263, opened 2026-06-15
Preserve NameError/AttributeError suggestions in traceback formatting

### LINKED ISSUE #13445: NameError/AttributeError suggestions in Python 3.10 error messages

shincho-1000:
Fixes #13445

IPython currently formats exception messages using str(exception), which drops Python's "Did you mean" suggestions for NameError and AttributeError.

This change preserves that suggestion text whenever traceback information is available.

Tests also added for both Verbose and Plain traceback modes.

### DIFF
--- a/IPython/core/ultratb.py
+++ b/IPython/core/ultratb.py
@@ -210,7 +210,7 @@ class ListTB(TBTools):
             )
             out_list.extend(self._format_list(elist))
         # The exception info should be a single entry in the list.
-        lines = "".join(self._format_exception_only(etype, evalue))
+        lines = "".join(self._format_exception_only(etype, evalue, etb))
         out_list.append(lines)
 
         # Find chained exceptions if we have a traceback (not for exception-only mode)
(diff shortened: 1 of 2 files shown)


shincho-1000:
Umm, my own tests are failing for versions 3.11 and 3.13. I'll work on it further.

Edit: Actually this only fails for 3.11 and works for 3.13.

shincho-1000:
While investigating these failures, I found that on Python 3.11 the suggestion text is not present in str(exception), traceback.format_exception(), traceback.print_exception() but it is present when using the built-in sys.excepthook.

The current implementation does preserve the message returned by the traceback module, but that is insufficient to reproduce suggestions on  3.11 since it is probably being done by a C-level helper function.

I think to implement this we need access to the CPython suggestion logic itself. Is that still in the scope of this issue?

Carreau:
Just don't implement suggestion on 3.11 with a version gate, and any specific test you can mark as `skipif(version<3.11)`

PR #15265, opened 2026-06-19
fix: clarify tk simple-prompt gui message

nightcityblade:
## Summary
- avoid the contradictory generic event-loop warning for Tk when `--simple-prompt` is active
- explain that Tk is natively supported and no event loop hook is installed
- add a regression test for the Tk simple-prompt message

## Testing
- `python -m pytest tests/test_interactiveshell.py::test_enable_gui_tk_simple_prompt_message -q`
- `python -m pytest tests/test_interactiveshell.py::test_enable_gui_tk_simple_prompt_message tests/test_interactiveshell.py::test_enable_gui_osx -q`

Fixes #14904


Carreau:
thnkas

PR #15275, opened 2026-06-26
Fix memory leak and handle errors gracefully in LLM autosuggester

aakritiieee7:
### Summary
This PR resolves a memory leak in the auto-suggestion subsystem of the terminal interactive shell, fixes an incorrect default value in the prefixer, and improves robustness against cancellations and API errors.

### Changes
1. **Memory Leak Fix**: The `disconnect` method of `NavigableAutoSuggestFromHistory` was updated to remove `_dismiss` from the buffer's `on_cursor_position_changed` event. Previously, only the `on_text_insert` handler was removed, resulting in a leak where the auto-suggester remained referenced by the event.
2. **Default Prefixer Fix**: Changed `_llm_prefixer` default implementation from returning `"wrong"` to `""`.
3. **Execution Robustness**: Wrapped `await self._llm_task` execution in a try-except block to catch `asyncio.CancelledError` and `Exception`. This ensures that natural task cancellations (during cursor movement) or LLM provider errors do not bubble up to prompt_toolkit and crash the prompt.

### Testing Checklist
- [x] Verify `on_cursor_position_changed` handler is removed upon disconnect.
- [x] Verify trigger routine gracefully handles task cancellation without throwing `CancelledError`.
- [x] Verify trigger routine gracefully catches and suppresses provider exceptions.

### DIFF
--- a/IPython/terminal/shortcuts/auto_suggest.py
+++ b/IPython/terminal/shortcuts/auto_suggest.py
@@ -177,7 +177,7 @@ class NavigableAutoSuggestFromHistory(AutoSuggestFromHistory):
     _init_llm_provider: Callable | None
 
     _llm_provider_instance: Any | None
-    _llm_prefixer: Callable = lambda self, x: "wrong"
+    _llm_prefixer: Callable = lambda self, x: ""
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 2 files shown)


Carreau (approved):

Carreau:
thanks.

PR #14526, opened 2024-09-30
ENH: add webp support to IPython.display.Image and complete identification of JPEG,PNG,GIF,WEBP image types by initial bytes

westurner:
- So that `Image('./example.webp')` does not raise a `"Cannot embed the 'webp' image format"` error.
- So that matplotlib animations saved as .webp images work (e.g. when ffmpeg+h264 isn't working).

PR #15276, opened 2026-06-28
fix: make Image(retina=True) work with WebP

flying-sheep:
The code exists, seems like it was forgotten in #14526.

cc @Carreau and @westurner

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1047,8 +1047,9 @@ class Image(DisplayObject):
             w, h = _jpegxy(self.data)
         elif self.format == self._FMT_GIF:
             w, h = _gifxy(self.data)
+        elif self.format == self._FMT_WEBP:
+            w, h = _webpxy(self.data)
         else:
-            # retina only supports png
             return
         self.width = w // 2
         self.height = h // 2


westurner:
Could be:
```
DPI_HI = 192
#DPI_RETINA = 

Image(,dpi=DPI_HI)

```

westurner:
From Arch wiki > HiDPI: https://wiki.archlinux.org/title/HiDPI :
> 96 DPI screens are regarded as comfortable for most people to read ~12pt font on, and is about where most "low-DPI" monitors fall. HiDPI screens are around 192 DPI and greater, with some screens falling in the middle ("medium PPI").

flying-sheep:
@westurner I think it’s a great idea to make this more generic, and I’d happily help, but this PR is just a very small fix and I don’t want to increase its scope.

Carreau:
thanks !

flying-sheep:
Great! I have a branch in which I refactor the image format handling – should I do a PR separately with that or should I do the DPI stuff in there?

The idea is that it becomes an enum which centrally handles format name, extension, mime type, and shape deriving. That way when modifying the code, it’s no longer possible to forget a patch: https://github.com/ipython/ipython/pull/15287

Issue #13587, opened 2022-03-12
[new future] Add postfix completion for ipython


fishandsheep:
Hi 
   I just learned python not long ago, ipython is fantastic! I'm a java coder.
   Is the team considering adding **postfix completion**？
   
   About **postfix completion** : Add var (or other word) to the end of your method and press tab, and the method will intelligently complete.
   For a sample  example by java:

   `                   String.valueOf(100).var`

   After you enter tab:

`          String s = String.valueOf(100);`

   Thanks

Carreau:
That would be possible, so the team is mostly me these days, so this will likely be only if someone contributes the feature. I think that maybe another trigger than `.var` as this may be valid python ? Or it may be possible to bind another shortcut than tab.

fishandsheep:
Thanks for the reply，Yes , It is very convenient to use **postfix completion** in jetbrains Idea, and you can create custom postfix templates, see more [postfix-completion](https://www.jetbrains.com/help/idea/settings-postfix-completion.html)

MrMino:
I think it would be super useful, and if only I had more time on my hands I would be starting the work on this right now.

Happy to review PRs and pull this if someone has the time to code it.

zishiwu123:
Hi, I am interested in working on this. I did a quick check on `.var` postfix idea to see it conflicts with Python reserved words, but I have not found any conflicts yet. My search shows that Python has:
- a function `vars()` at applies to objects that implement the `__dict__` attribute: https://docs.python.org/3/library/functions.html
- 33 keywords, none of which are named `var`: https://docs.python.org/3.10/reference/lexical_analysis.html#keywords
- Soft keywords since Python 3.10, of which none are named `var`: https://docs.python.org/3.10/reference/lexical_analysis.html#soft-keywords

I think it would be good 1) to start with the idea of a built-in postfix completion trigger such as `.var` as a proof of concept. After that is done, we can think about 2) how to creating the custom postfix competion triggers. Here is an example of part 1), based on the example in Java provided by @fishandsheep.
```py
str(100).var
s = str(100)
```

I think the string to number example is better for Java than for Python, since Java requires a user to type more characters typed than Python does for that example. In fact, a user would type the same number of characters (13) for both expressions `s = str(100)` and `str(100).var`. To justify the postfix completion feature, we need more complex examples in Python where the feature actually saves users a considerable amount of typing. Maybe something like generating a list comprehension with a `.lc` trigger that only works on `Iterable` variables. For example:
```py
[...]

PR #15280, opened 2026-07-04
Add opt-in terminal postfix completions

fishandsheep:
## Summary

This adds opt-in postfix completions to the terminal prompt. The feature is disabled by default and only participates when `TerminalInteractiveShell.postfix_completion` is enabled.

It supports both direct postfix expansion and menu-driven postfix completion:

- `x.print<Tab>` expands to `print(x)`.
- `x.pri<Tab>` can accept the `print` postfix item directly and expand to `print(x)`.
- `x.<Tab>` shows normal attribute/method completions plus available postfix templates.
- Postfix menu entries use configurable prompt_toolkit styles so they can be visually distinguished from regular completions.

Refs #13587

## How to enable

In `ipython_config.py`:

```python
c.TerminalInteractiveShell.postfix_completion = True
```

Optional configuration:

```python
c.TerminalInteractiveShell.postfix_completion_trigger = "."
c.TerminalInteractiveShell.postfix_completion_templates = [
    "not",
    "par",
    "return",
    "if",
    "while",
    "print",
    "len",
    "raise",
    "yield",
    "str",
    "list",
    "set",
    "dict",
    "tuple",
]
c.TerminalInteractiveShell.postfix_completion_style = "bg:#44475a #f8f8f2"
c.TerminalInteractiveShell.postfix_completion_selected_style = "bg:#6272a4 #ffffff"
```

## Supported templates

| Postfix | Example input | Expansion |
| --- | --- | --- |
| `.not` | `x.not` | `not x` |
| `.par` | `x.par` | `(x)` |
| `.return` | `x.return` | `return x` |
| `.if` | `x.if` | `if x:` followed by an indented line |
| `.while` | `x.while` | `while x:` followed by an indented line |
| `.print` | `x.print` | `print(x)` |
| `.len` | `items.len` | `len(items)` |
| `.raise` | `err.raise` | `raise err` |
| `.yield` | `x.yield` | `yield x` |
| `.str` | `x.str` | `str(x)` |
| `.list` | `x.list` | `list(x)` |
| `.set` | `x.set` | `set(x)` |
| `.dict` | `x.dict` | `dict(x)` |
| `.tuple` | `x.tuple` | `tuple(x)` |

## Behavior details

- Postfix expansion only applies to the current input line before the cursor.
- The expression before the postfix trigger must parse as a Python expression.
- Lines containing comments, tokenizer errors, or incomplete expressions are ignored and fall back to normal Tab behavior.
- Normal IPython completions are yielded first; postfix entries are appended after them.
- Selecting a postfix completion from the menu inserts the expanded form immediately, rather than first completing the postfix key and requiring another Tab.
- The trigger is configurable, so users can avoid conflicts with dot-based attribute completion if desired.

## Validation

```bash
python -m pytest tests/test_shortcuts.py tests/test_interactivshell.py tests/test_completer.py -q
```

Result:

```text
374 passed, 3 skipped, 3 xfailed
```

### DIFF
--- a/IPython/terminal/interactiveshell.py
+++ b/IPython/terminal/interactiveshell.py
@@ -302,6 +302,47 @@ class TerminalInteractiveShell(InteractiveShell):
         """,
     ).tag(config=True)
 
+    postfix_completion = Bool(
+        False,
+        help="Enable postfix completion in the terminal prompt.",
+    ).tag(config=True)
+
+    postfix_completion_trigger = Unicode(
+        ".",
+        help="Character sequence used to trigger postfix completion templates.",
+    ).tag(config=True)
+
+    postfix_completion_templates = List(
+        Unicode(),
+        default_value=[
+            "not",
+            "par",
+            "return",
+            "if",
+            "while",
+            "print",
+            "len",
+            "raise",
+            "yield",
+            "str",
+            "list",
+            "set",
+            "dict",
+            "tuple",
+        ],
+        help="Postfix completion templates enabled in the terminal prompt.",
+    ).tag(config=True)
+
+    postfix_completion_style = Unicode(
+        "bg:#44475a #f8f8f2",
+        help="Style used for postfix completion menu entries.",
+    ).tag(config=True)
+
+    postfix_completion_selected_style = Unicode(
+        "bg:#6272a4 #ffffff",
+        help="Style used for the selected postfix completion menu entry.",
+    ).tag(config=True)
+
     mouse_support = Bool(False,
         help="Enable mouse support in the prompt\n(Note: prevents selecting text with the mouse)"
     ).tag(config=True)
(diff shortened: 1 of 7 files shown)


fishandsheep:
Enable it quickly from the terminal
```
%config TerminalInteractiveShell.postfix_completion = True
```

Four years later, AI helped me finally make an old idea real.

Carreau:
That's neat; now that you have you AI, do you think we could look at whether this is doable as an IPython extension; and if there are minimal changes to make it _possible_ get those changes in, and keep that as a separate package instead of in-core ?

fishandsheep:
Thanks for the feedback and the suggestion. That approach makes sense. I'll try implementing it as an IPython extension first and come back with a smaller PR if core changes are needed.

fishandsheep:
Update: I created a small external IPython extension to prototype this idea without changing IPython core:

[ipython-postfix-completion](https://github.com/fishandsheep/ipython-postfix-completion)

Install into the current Python environment with `uv`:

```bash
uvx --with ipython-postfix-completion ipython
```

It provides configurable postfix completion for IPython. After loading the extension:

```python
%load_ext ipython_postfix_completion
```

Carreau:
Love this ! super useful for "await" imho; i find that .await is more readable and chainable.

PR #15288, opened 2026-07-06
Limit file completions to path contexts

### LINKED ISSUE #14501: Only trigger filename completion in special syntax or strings
Right now filename completions trigger in all sorts of situations, for example:

```python
def f(x): pass
f(x=<TAB>
```
If there are a lot of files in a directory, this can overwhelm the completions.

As mentioned [here](https://github.com/ipython/ipython/issues/10926#issuecomment-515949953), it would be nice if completers had more semantic context about the match request automatically available to them (for example, if we were in a string, etc.). In this case, it would be great if the filename matcher only matched if we were in special syntax (e.g., lines starting with ! or a %magic line), or if we were in a string. https://github.com/ipython/ipython/pull/12128 implemented an "are we in a string" check for Jedi.

My proposal here is that we:
1. abstract out the "are we in a string" check from https://github.com/ipython/ipython/pull/12128 to a function 
2. Use that function in the [file matcher](https://github.com/ipython/ipython/blob/78fea5f8d7bf6ca55796e82ed341e7fd291878f0/IPython/core/completer.py#L2009) to only return matches if we are in a string or special syntax.

(FYI, another path completion issue is https://github.com/ipython/ipython/issues/11530)

cjc0013:
## Summary

- limit filename completions to string literals, shell/magic contexts, and explicit path-looking tokens
- keep direct path completion for prefixes such as `./`, `../`, `~`, absolute paths, and Windows drive paths
- include `name = !command` as shell syntax by reusing IPython's system-assignment tokenizer

Fixes #14501.

## Local checks

- `python -m pytest tests/test_completer.py::TestCompleter::test_file_completions_only_in_path_contexts tests/test_completer.py::TestCompleter::test_local_file_completions -q`
- `python -m pytest tests/test_completer.py -q`
- `python -m ruff check IPython/core/completer.py`
- `python -m compileall -q IPython/core/completer.py tests/test_completer.py`
- `git diff --check`

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -216,7 +216,11 @@ from IPython.core.guarded_eval import (
     _validate_policy_overrides,
 )
 from IPython.core.error import TryNext, UsageError
-from IPython.core.inputtransformer2 import ESC_MAGIC
+from IPython.core.inputtransformer2 import (
+    ESC_MAGIC,
+    SystemAssign,
+    make_tokens_by_line,
+)
 from IPython.core.latex_symbols import latex_symbols, reverse_latex_symbol
 from IPython.testing.skipdoctest import skip_doctest
 from IPython.utils import generics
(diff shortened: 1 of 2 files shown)


Carreau:
@copilot resolve the merge conflicts in this pull request

PR #15292, opened 2026-07-07
Remove deprecated py3compat module and inline its utilities

Carreau:
This PR removes the deprecated `IPython/utils/py3compat.py` module and inlines its utility functions throughout the codebase. The py3compat module was primarily designed for Python 2/3 compatibility, which is no longer needed as IPython now requires Python 3.

## Summary
The `py3compat` module contained several compatibility utilities that are no longer necessary in a Python 3-only codebase. Rather than maintaining this deprecated module, its functionality has been directly integrated into the modules that use it.

## Key Changes

- **Removed** `IPython/utils/py3compat.py` entirely
- **Inlined `safe_unicode()` logic** in `IPython/core/interactiveshell.py` - replaced with direct try/except handling of `str()`, `repr()`, and fallback string
- **Inlined `execfile()` logic** in `IPython/core/interactiveshell.py` and `tests/test_debugger.py` - replaced with direct file reading and `exec()` calls
- **Replaced `cast_unicode()` calls** in `IPython/lib/clipboard.py` and `IPython/core/tbtools.py` with inline conditional checks and `.decode()` calls
- **Replaced `decode()` calls** in `IPython/lib/clipboard.py`, `IPython/testing/tools.py`, and `IPython/utils/_process_common.py` with direct `.decode(DEFAULT_ENCODING, "replace")` calls
- **Replaced `encode()` calls** in `IPython/testing/tools.py` with direct `.encode(DEFAULT_ENCODING, "replace")` calls
- **Replaced `input()` wrapper** in `IPython/lib/demo.py`, `IPython/core/page.py`, `IPython/lib/editorhooks.py`, and `IPython/terminal/magics.py` with builtin `input()`
- **Replaced `PYPY` constant** in `IPython/lib/pretty.py` with inline `platform.python_implementation() == "PyPy"` check
- **Updated imports** across all affected modules to remove `py3compat` references and import `DEFAULT_ENCODING` from `IPython.utils.encoding` where needed

## Implementation Details

All inlined code maintains the same error handling and fallback behavior as the original utility functions. The `DEFAULT_ENCODING` constant is imported from `IPython.utils.encoding` where needed for encoding/decoding operations.

https://claude.ai/code/session_01NyXpP8SsNDxTCEKnZ5MpF8


IPython/core/tbtools.py
```diff
@@ -202,9 +201,10 @@ def _tokens_filename(
...
             ]
     else:
-        name = util_path.compress_user(
-            py3compat.cast_unicode(file or "", util_path.fs_encoding)
-        )
+        file_str = file or ""
+        if isinstance(file_str, bytes):
+            file_str = file_str.decode(util_path.fs_encoding, "replace")
+        name = util_path.compress_user(file_str)
         if lineno is None:
```
Carreau, line 207: @claude it look like `file` can never be bytes; remove the conditional; add an assertion at the start of the function.

PR #15314, opened 2026-07-15
Refactor banner property logic

xavya-dev:
Refactor banner property to check SOURCE_DATE_EPOCH environment variable directly.

### DIFF
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1034,7 +1034,7 @@ class InteractiveShell(SingletonConfigurable):
     @property
     def banner(self):
         banner = self.banner1
-        # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner (default starts with "Python")
+        # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner
         if banner is default_banner and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
             from datetime import datetime
             date = datetime.fromtimestamp(int(when))
--- a/IPython/core/interactiveshell.py
+++ b/IPython/core/interactiveshell.py
@@ -1035,7 +1035,10 @@ class InteractiveShell(SingletonConfigurable):
     def banner(self):
         banner = self.banner1
         # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner
-        if banner is default_banner and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
+        if (
+            banner is default_banner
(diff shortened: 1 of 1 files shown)


collares:
This PR removes a redundant condition in the current code, but the redundant condition was due to a typo in #15208. I believe the correct fix is to replace `banner is self.banner1` by `banner is default_banner`, which wouldn't be redundant anymore.

cc @tornaria

collares:
Or perhaps just `banner == default_banner`, since I believe `default_banner` is [copied into `self.banner1` instead of assigned](https://github.com/xavya-dev/ipython/blob/fbdfcc70fb9c7ad67746f83fe881090cda6c5c1d/IPython/core/interactiveshell.py#L432).

xavya-dev:
Thanks for the suggestion! Just to confirm, should I update the if condition to check if banner == default_banner instead of removing it entirely? I can update this PR with that change.

collares:
I just tested and `banner is default_banner` seems to work, and it should be faster. So, yes, the suggestion I'm making is to replace the condition `banner is self.banner1` by `banner is default_banner`, which is essentially what @tornaria suggested in #15208.

collares:
Wait, wait, I was testing `banner == default_banner`, which explains why it's working. Let me test the other variant.

collares:
Seems like `banner is default_banner` indeed works as well.

IPython/core/interactiveshell.py
```diff
@@ -1032,7 +1032,7 @@ def restore_sys_module_state(self):
     def banner(self):
         banner = self.banner1
         # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner (default starts with "Python")
-        if banner is self.banner1 and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
+        if (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
```
tornaria:
```suggestion
        if banner is default_banner and (when := os.environ.get("SOURCE_DATE_EPOCH", None)) is not None:
```

You need some test here.

IPython/core/interactiveshell.py
```diff
@@ -1032,7 +1032,7 @@ def restore_sys_module_state(self):
     def banner(self):
         banner = self.banner1
         # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner (default starts with "Python")
```
tornaria:
```suggestion
        # Only use SOURCE_DATE_EPOCH if the user hasn't set a custom banner
```
Also maybe adjust the comment here

PR #15315, opened 2026-07-16
 Fix `%edit` failing when path contains spaces (editor hook double-quoting)

### LINKED ISSUE #14961: %edit editorhook fails if path has spaces
Notepad++ editorhook doesn't work for %edit if you try to edit an object from a file whose path has spaces.

----

The `%edit` magic single-quotes the path if it has spaces:
https://github.com/ipython/ipython/blob/main/IPython/core/magics/code.py#L724-L725

The default editor expects a safe filename, and joins it into a string to use in a Popen call:
https://github.com/ipython/ipython/blob/main/IPython/core/hooks.py#L76
(Notice that it DOES quote the editor path, conditionally, so that is an inconsistency.)

The editor hooker makes an editor callback which expects the filename to be an unescaped string, and then uses `shlex.quote` to substitute it into a template and do the call:
https://github.com/ipython/ipython/blob/main/IPython/lib/editorhooks.py#L49

----

In my opinion, the responsibility of escaping should be moved down to the callee, because taking a quoted filepath is weird. This will break people's custom hooks which correct for this bug.

But there is already a test which calls a custom editor hook with unquoted spaces:
https://github.com/ipython/ipython/blob/cbac235320fe91056a0fb6dba74cd8bef7fc21d9/tests/test_editorhooks.py#L25

And `%editmacro` also calls the editor hook without quoting:
https://github.com/ipython/ipython/blob/main/IPython/core/magics/code.py#L539-L540
I suspect that breaks if the temp folder is on a path with spaces.

----

Also, %edit takes the absolute path, which might lengthen error messages unnecessarily.
https://github.com/ipython/ipython/blob/main/IPython/core/magics/code.py#L722

Also, there should be more consistency between how the default and the custom process calls are done.

shreyabyte:
Fixes #14961

This PR standardises filename handling by moving quoting responsibility into the editor hook.

### Problem

`%edit` single-quoted the target filename before passing it to `self.shell.hooks.editor()` if the path contained spaces. The default editor hook (`IPython/core/hooks.py`) then built a shell command by string-formatting `editor`, `linemark`, and `filename` together, without quoting `filename` at all — relying on `%edit` having already done it. Custom hooks installed via `IPython/lib/editorhooks.py` also call `shlex.quote(filename)` on their own, assuming a *raw* filename — so any path with spaces reaching those hooks got double-quoted and broken (e.g. the Notepad++ hook).

In short, callers and hooks disagreed about who owns quoting, and only one side actually did it.

### Fix

Established a clear contract: **`hooks.editor()` always receives a raw, unquoted `filename`. Quoting/escaping for shell execution is the responsibility of the hook implementation.**

- `IPython/core/magics/code.py`: removed the pre-quoting logic in `%edit` (`quoted = "'%s'" % quoted`); now passes the raw filename to `self.shell.hooks.editor()`, matching what `_edit_macro` already did. Added comments at both call sites documenting the contract.
- `IPython/core/hooks.py`: added a small `_quote_if_needed()` helper, applied consistently to both `editor` and `filename` before building the `Popen` command string, so the default hook now correctly quotes filenames with spaces (previously only the editor binary path was quoted).
- `IPython/lib/editorhooks.py`: unchanged — its `shlex.quote(filename)` calls already assumed a raw filename; this was previously being fed an already-quoted string, causing the double-quoting bug. No code change needed there, it now works as originally intended.

### Tests

- `IPython/core/tests/test_hooks.py`: added tests for the default `editor()` hook verifying filenames with spaces are quoted in the resulting command, filenames without spaces are left untouched, the editor binary path quoting behavior is preserved, and the full command (with a line number) is assembled correctly.
- `IPython/core/tests/test_magics_code.py`: fixed `test_edit_filename_with_space_is_quoted` (renamed to `test_edit_filename_with_space_passed_raw`), which previously asserted the *quoted* form was received and stripped quotes in the `EditorStub` fixture — masking the real bug. It now asserts `%edit` passes the hook a fully raw filename with no surrounding quote characters. Removed the now-unneeded `filename.strip("'")` in `EditorStub` and the `@skip_win32` marker, since the test no longer depends on shell-quoting semantics.

### Compatibility note

This changes the editor hook contract so that `hooks.editor()` now receives a raw filename instead of a pre-quoted one.

### Manual verification

Verified `%edit` on a path containing spaces opens correctly via the default hook and via a custom hook (Notepad++), with no stray quote characters and no "file not found" errors.


shreyabyte:
Hi @Carreau, just updated the branch with the latest main; checks are passing. Wanted to bump this since it's been open for a bit. Please take a look when you get a chance, and let me know if anything needs changing before it can be merged. Thanks!

PR #15316, opened 2026-07-16
fix: recognize cell magics after comments

### LINKED ISSUE #15312: Comment lines above a cell magic make it parse as a line magic named `%name`, with a confusing error
Putting a comment above a cell magic fails like this:

```
In [1]: # setup
   ...: %%time
   ...: pass
UsageError: Line magic function `%%time` not found.
```

Blank lines above a cell magic are fine, since the `leading_empty_lines` cleanup transform drops them. Comment lines are the one thing that breaks.

From reading `inputtransformer2`, what seems to happen is: cell magic detection only looks at the first line of the cell, so the `%%time` line falls through to the escaped-command transform, which strips one `%` as the escape character and takes `%time` as the magic name. Names registered through the documented APIs come from Python function names, which can't contain `%`, so the lookup always fails. As far as we can tell the current parse serves no reachable purpose, with one exception: `register_magic_function` doesn't validate names, so a magic literally named `%foo` can be registered programmatically, and a mid-cell `%%foo` line does invoke it. We doubt anyone depends on that, but it's why we're not suggesting a hard error.

Two things seem worth improving, assuming we're not missing some history here:

1. The error message. The user wrote a cell magic, and the message says a line magic wasn't found. The doubled `%%` in the message also hides that the failed lookup was for `%time`. A message saying that a cell magic must be the first line of the cell would point at the actual mistake.

2. Possibly the parse itself. A `leading_comment_lines` cleanup transform in the spirit of `leading_empty_lines`, dropping leading blank and comment lines when the first real line starts with `%%`, would make the cell mean what the author obviously intended. A line starting with `%%` can't begin a Python statement, and a string continuation can't be the first real line of a cell, so we couldn't find currently valid code whose meaning would change, apart from the pathological registration above.

We ran into this through notebook tooling where both humans and LLM agents tend to put a narration comment at the top of a cell. We've added the transform to our own `InteractiveShell` subclass ([execnb](https://github.com/AnswerDotAI/execnb)) and it has behaved well so far. Happy to send a PR for either or both parts, and equally happy to be told what nuance we've missed.

Tested on IPython 9.15.0.

tianrking:
## Summary
- recognize a cell magic after leading comments and blank lines
- add a regression test for a commented `%%time` cell
- document the corrected input transformation behavior

## Root cause
Cell-magic detection only examined the first physical line. A leading comment left `%%time` to the escaped-command transformer, which then attempted to invoke an invalid `%time` line magic.

Fixes #15312

## Validation
- `python -m pytest tests/test_inputtransformer2_line.py tests/test_inputtransformer2.py -q` (66 passed)
- `python -m compileall -q IPython/core/inputtransformer2.py`

🤖🍆

### DIFF
--- a/IPython/core/inputtransformer2.py
+++ b/IPython/core/inputtransformer2.py
@@ -36,6 +36,23 @@ def leading_empty_lines(lines):
             return lines[i:]
     return lines
 
+
+def leading_comment_lines(lines):
+    """Remove leading comments before a cell magic.
+
+    Cell magics must be the first meaningful line in a cell. Comments and
+    blank lines preceding a cell magic do not affect its execution, so remove
+    them before detecting the magic.
+    """
+    for i, line in enumerate(lines):
+        if not line.strip() or line.lstrip().startswith('#'):
+            continue
+        if line.startswith('%%'):
+            return lines[i:]
+        break
+    return lines
+
+
 def leading_indent(lines):
     """Remove leading indentation.
 
@@ -668,6 +685,7 @@ class TransformerManager:
         self.cleanup_transforms = [
             leading_empty_lines,
             leading_indent,
+            leading_comment_lines,
             classic_prompt,
             ipython_prompt,
         ]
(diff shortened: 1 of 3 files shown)


Carreau:
@tianrking if this is an automated PR can you run autoformatting ? See Ci failure.

Carreau:
This branch needs rebase as it conflicts with main now.

PR #15318, opened 2026-07-19
Fix bare except clauses in tests/test_formatters.py and tests/simpleerr.py

koteshyelamati:
Replace bare `except:` clauses with specific exception types.

**Changes:**
- `tests/test_formatters.py`: `except ImportError:` for the `import numpy` guard
- `tests/simpleerr.py`: `except ValueError:` when parsing `int(sys.argv[2])`

Bare `except:` catches `BaseException` subclasses (including `KeyboardInterrupt`, `SystemExit`). These specific types match the actual exceptions raised by the guarded code.

### DIFF
--- a/tests/simpleerr.py
+++ b/tests/simpleerr.py
@@ -21,7 +21,7 @@ def bar(mode):
     elif mode == "exit":
         try:
             stat = int(sys.argv[2])
-        except:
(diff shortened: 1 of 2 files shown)


Carreau:
can you please look at the test failures ?

PR #15333, opened 2026-07-25
Lazily import many imports to speed up startup. 

Carreau:
This is a global effort to speedup both import time and startup of IPython (and ipykernel). 

```
IPYKERNEL_BENCHMARK_STARTUP_SHUTDOWN=1 python tools/importtime_average.py -n 20 -- python -X importtime -m ipykernel  > totlog.log && tuna totlog.log
```

With `tuna` installed to view the report. 

Beyond making things lazy, it also add an option to avoid autodetection of kitty support at startup that can take some time. This can be controled with `IPYTHON_KITTY_GRAPHICS`, which is an env variable as it need to be detected before traitlets configuration options.

This did use Opus to split the work into multiple commits in case of regression to partially revert and write some of the commit messages

### DIFF
--- a/IPython/__init__.py
+++ b/IPython/__init__.py
@@ -34,16 +34,84 @@ from typing import Any
 # rely on the transitive side effects: they do `import IPython` and then
 # access attribute chains like `IPython.terminal.ipapp.TerminalIPythonApp`,
 # which only resolve because the imports below load those submodules.
-from .core.getipython import get_ipython
+#
+# `embed`, `Application` and `get_ipython` are the exceptions, and are
+# deferred via module `__getattr__` below:
+#
+#  - `embed` drags in the whole terminal / prompt_toolkit stack, by far
+#    the most expensive of these imports, and is only needed by code that
+#    calls `IPython.embed()`;
+#  - `Application` is only a re-export of `traitlets.config.application
+#    .Application`, but importing it pulled in `IPython.core.application`
+#    and with it the crash handler; no known downstream imports it from
+#    here (ipykernel imports `BaseIPythonApplication` from
+#    `IPython.core.application` directly);
+#  - `get_ipython` costs nothing to defer -- `IPython.core.getipython`
+#    ends up imported anyway via `IPython.core.magic` -- but is kept
+#    alongside the others so all three top-level names resolve the same
+#    way.
+#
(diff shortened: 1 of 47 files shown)


Carreau:
## Startup : 200ms -> 150ms

```
IPYTHON_KITTY_GRAPHICS=1 python profile_tty_startup.py -n 10
$ /Users/bussonniermatthias/miniconda3/envs/arm64/bin/python -m IPython   (10 runs, under a pty)
  first byte (banner) : median  197.5 ms   min  193.6 ms   max  459.6 ms   [10/10]
  first prompt (In[1]): median  201.8 ms   min  197.9 ms   max  465.9 ms   [10/10]
```
[...SNIP...]
```
Switched to branch 'lazy2'
$ IPYTHON_KITTY_GRAPHICS=1 python profile_tty_startup.py -n 10
$ /Users/bussonniermatthias/miniconda3/envs/arm64/bin/python -m IPython   (10 runs, under a pty)
  first byte (banner) : median  125.4 ms   min  123.7 ms   max  196.2 ms   [10/10]
  first prompt (In[1]): median  142.4 ms   min  140.7 ms   max  213.4 ms   [10/10]
```


Import is of course more impressive:

##  import  : 150ms -> 50ms

```
IPYTHON_KITTY_GRAPHICS=1 hyperfine 'python -c "import IPython"'
Benchmark 1: python -c "import IPython"
  Time (mean ± σ):     167.9 ms ±   2.0 ms    [User: 144.3 ms, System: 22.2 ms]
  Range (min … max):   163.6 ms … 170.9 ms    17 runs
```

vs 
```
IPYTHON_KITTY_GRAPHICS=1 hyperfine 'python -c "import IPython"'
Benchmark 1: python -c "import IPython"
  Time (mean ± σ):      57.2 ms ±   7.9 ms    [User: 48.9 ms, System: 7.5 ms]
  Range (min … max):    53.6 ms …  97.9 ms    29 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (97.9 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
```

PR #15334, opened 2026-07-26
escape urls and file names interpolated into display html attributes

sage-mode-hunter:
Values callers pass to the display objects land unescaped inside quoted HTML attributes, so a quote in one closes the attribute and the rest is parsed as markup:

- `Image` and `Video` interpolate url/filename into `src="..."`; `Image._repr_html_` already escapes `alt` but not the url beside it
- `IFrame` does the same for `src`, `width` and `height`, reachable through the `YouTubeVideo`/`VimeoVideo`/`ScribdDocument` id argument
- `Audio` for its url and `element_id`, and the `FileLinks` notebook formatter for names read off disk
- `YouTubeVideo('abc"><script>')` closes the iframe and injects a tag into the notebook output

Left raw: `Video.html_attributes`, `IFrame.extras` and the `FileLink` html prefix/suffix, which are documented as HTML; `IFrame` params are already percent-encoded by urlencode. For a valid url the only change is `&` rendering as `&amp;`.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1065,7 +1065,7 @@ class Image(DisplayObject):
             if self.alt:
                 alt = ' alt="%s"' % html.escape(self.alt)
             return '<img src="{url}"{width}{height}{klass}{alt}/>'.format(
-                url=self.url,
+                url=html.escape(self.url or ""),
                 width=width,
                 height=height,
                 klass=klass,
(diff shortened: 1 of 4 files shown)


Carreau:
It look like this need some fixes on windows. But +1

sage-mode-hunter:
Fixed. The only windows failure was my own `test_escaped_names_FileLinks`, which built a file called `a'><script>.txt` and `<`/`>` aren't legal in windows names, so the open() blew up with EINVAL before the test got anywhere near the escaping.

Swapped the payload for `a' onmouseover='alert(1)&.txt`, which is valid everywhere. It's arguably the more accurate repro anyway since FileLinks emits `href='%s'` with single quotes, so the apostrophe is the character that actually breaks out of the attribute. Still fails on the unpatched tree and passes with the fix.

PR #14331, opened 2024-02-13
Add json option to lsmagic

s-kai273:
# Summary
This PR is based on the discussion in issue[#11793](https://github.com/ipython/ipython/issues/11793). 
When executing lsmagic in a Lab environment, the results are returned in JSON format, and following that, I have made it possible to obtain similar ones by adding a `--json` option.

# Change
- [x] Add json option to lsmagic
implementation code reffers to this [comment](https://github.com/ipython/ipython/issues/11793#issuecomment-1650723671)

PR #15335, opened 2026-07-26
Make %lsmagic return plain text by default

### LINKED ISSUE #11793: lsmagic returns JSON which makes for awkward UX in JupyterLab
I'm not sure why `lsmagic` needs to return a JSON data structure by default - but the end result is that in JupyterLab, it becomes kind of useless... In Classic calling `%lsmagic` produces (clipped for simplicity):

image

while in Lab, this is the output:

image

Lab defaults to showing JSON reprs when available, which typically makes sense, so while this is a change in behavior re. Classic, I'm inclined to think it's a good step forward.

Users can get the text via

image

but that's  a pretty awkward experience.  Do we really need `lsmagic` to produce JSON by default? It seems to me that generating plain text would be fine most of the time, perhaps with a `--json/-j` option for JSON output when desired?

utkarshalpha:
Closes #11793.

`%lsmagic` now returns its human-readable text by default so frontends such as JupyterLab do not prefer the JSON MIME representation. Users that need the machine-readable mapping can request it explicitly with `%lsmagic --json`.

This adds coverage for both output modes and an incompatibility news fragment. It follows up on the approach in the now-closed #14331, with the default path returning text instead of the original rich-display object.

### DIFF
--- a/IPython/core/magics/basic.py
+++ b/IPython/core/magics/basic.py
@@ -181,10 +181,21 @@ class BasicMagics(Magics):
                 magic_escapes['cell'], name,
                 magic_escapes['cell'], target, params_str))
 
+    @magic_arguments.magic_arguments()
+    @magic_arguments.argument(
+        "-j", "--json", action="store_true", help="Return the magic list as JSON."
+    )
     @line_magic
     def lsmagic(self, parameter_s=''):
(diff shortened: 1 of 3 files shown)


Carreau:
Why not... thanks !

PR #15337, opened 2026-07-29
Resolve attribute annotations under the policy in eval_node

sage-mode-hunter:
**Arbitrary code execution from class annotations in eval_node**

The attribute branch falls back to `typing.get_type_hints()` after the policy has already refused the attribute, and that call resolves stringized annotations with `eval()`, so completing `obj.attr` runs whatever the annotation contains under the default `limited` policy. Collecting the raw annotations with `inspect.get_annotations(eval_str=False)` and resolving them through `eval_node` keeps annotation completion working while routing the expression through the same policy checks as the rest of the input.


krassowski (approved):
Good catch, the diff looks good.

PR #15338, opened 2026-07-30
Use guarded_eval for the callable in python_func_kw_matches

sage-mode-hunter:
Keyword completion is the one evaluation site left in the completer that never consulted the policy:

- `python_func_kw_matches` resolves the callable with a bare `eval()`, while `global_matches`, `_evaluate_expr` and the dict key matcher all go through `guarded_eval`
- `python_func_kw_matcher` is in the default matcher list, so `obj.attr(` plus Tab runs a property getter or a custom `__getattr__` with no cell executed
- `guarded_eval` refuses that same attribute access under the default `limited` policy, so the two currently disagree about the same expression

Routed it through the `EvaluationContext` the sibling matchers already build. Plain functions and ordinary methods still complete their keywords, and the eager behaviour stays available under `evaluation="unsafe"`.

### DIFF
--- a/IPython/core/completer.py
+++ b/IPython/core/completer.py
@@ -2975,8 +2975,18 @@ class IPCompleter(Completer):
         argMatches = []
         try:
             callableObj = '.'.join(ids[::-1])
-            namedArgs = self._default_arguments(eval(callableObj,
-                                                    self.namespace))
+            namedArgs = self._default_arguments(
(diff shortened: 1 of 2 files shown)


krassowski (approved):
LGMT, thank you!

sage-mode-hunter:
gentle ping

PR #15348, opened 2026-08-03
Restore pathlib.Path support in iFrame

### LINKED ISSUE #15345: IPython 9.16.0 no longer allows passing a pathlib.Path as an IFrame's `src`
At least as far back as 2022, IPython allowed you to pass a `pathlib.Path` as the `src` parameter to the `IPython.display.IFrame` constructor. As of 9.16.0, that fails with a `TypeError: Path.replace() takes 2 positional arguments but 3 were given`, because the `IFrame` constructor now calls `html_escape(self.src)`. It looks like it's #15334 that changed this.

If it was never supported to pass a `Path` and this only ever worked by coincidence, feel free to close this, but if this is an unintentional regression perhaps consider calling `html_escape(str(self.src))` instead.

I noticed this because it breaks [Memray's Jupyter integration](https://bloomberg.github.io/memray/jupyter_magic.html).

itsdaxen:
Fixes #15345

PR #15334 broke pathlib.Path support in IFrame._repr_html_() because html.escape() doesn't handle Path objects (it hits Path.replace() instead of converting to a string).

### Changes:

- Explicitly convert src to a string before passing it to html.escape().
- Add a regression test using a path with & to verify both conversion and escaping work as expected.

### DIFF
--- a/IPython/lib/display.py
+++ b/IPython/lib/display.py
@@ -292,7 +292,7 @@ class IFrame:
         else:
             params = ""
         return self.iframe.format(
-            src=html_escape(self.src),
+            src=html_escape(str(self.src)),
             width=html_escape(str(self.width)),
             height=html_escape(str(self.height)),
             params=params,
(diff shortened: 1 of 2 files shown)


Carreau:
thanks.

PR #15351, opened 2026-08-03
ci: try ubuntu-slim runner for lint-only jobs

Carreau:
## Summary
- Switch the lint-only CI jobs to `runs-on: ubuntu-slim` as a test: `ruff.yml`, `mypy.yml`, `zizmor.yml`, and the `formatting` job in `python-package.yml`.
- These jobs only install Python packages via `pip`/`uv` and don't need the full `ubuntu-latest` toolcache (browsers, docker, extra language runtimes, etc.), so they're candidates for a lighter/cheaper runner.
- Left the actual test matrices (`test.yml`, `downstream.yml`), `docs.yml` (needs `apt-get install graphviz`), and sdist/build jobs on `ubuntu-latest` since they rely more on the full image or apt packages.

## Test plan
- [ ] Confirm the `ruff`, `mypy`, `zizmor`, and `formatting` CI jobs succeed on `ubuntu-slim` in this PR's checks
- [ ] Compare job duration/cost vs. `ubuntu-latest` runs on `main`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hh6BALpqMaCX9Qn5yHsfgv)_

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   formatting:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     timeout-minutes: 5
     steps:
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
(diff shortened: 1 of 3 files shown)


Carreau:
Actually not worth it as it makes some job slower and thus hurt us on concurency.

PR #15362, opened 2026-08-09
Preserve source line numbers in deduperreload reloaded functions

### LINKED ISSUE #15359: deduperreload gives invalid tracebacks in reloaded functions

```python

In [1]: %load_ext autoreload

In [2]: %autoreload -p 1

In [3]: %%writefile module.py
   ...: """
   ...: """
   ...: def func():
   ...:
   ...:

In [4]: %aimport module

In [5]: module.func()
Cell In[5], line 1
----> 1 module.func()

      4 def func():

AssertionError: before reload

In [6]: %%writefile module.py
   ...: """
   ...: """
   ...: def func():
   ...:
   ...:
Overwriting module.py

In [7]: module.func()
Reloading 'module'.
Cell In[7], line 1
----> 1 module.func()

      1 """
      3 """
      4 def func():

AssertionError: after reload
```

fudianchn:
## Problem

`DeduperReloader._patch_namespace_inner` unparsed the new function AST to source (`textwrap.dedent(ast.unparse(new_ast_def))`) and re-parsed it before `exec`. That round-trip discards the original source line numbers, so reloading a function resets its line information and tracebacks point at the wrong line — the offset within the unparsed text rather than the source file (#15359).

## Fix

Operate on the AST node directly instead of unparse→reparse:

```diff
-                func_code = textwrap.dedent(ast.unparse(new_ast_def))
                 if is_method := (len(prefixes) > 0):
-                    func_code = "class __autoreload_class__:\n" + textwrap.indent(func_code, "    ")
+                    new_ast_def = ast.ClassDef(name="__autoreload_class__", body=[new_ast_def])
                 ...
-                func_asts = [ast.parse(func_code)]
+                func_asts = [ast.Module(body=[new_ast_def], type_ignores=[])]
+                ast.fix_missing_locations(func_asts[0])
```

The function node keeps its original `lineno`, so tracebacks point at the right line. The `ast.Module` wrap keeps the structure `ast.parse` returns (so the existing decorator-list check on `func_asts[0].body[0]` is unchanged and `compile()` accepts it).

This builds on the direction sketched by @yut23 in #15359, but that snippet set `func_asts = [new_ast_def]`, which `compile(..., mode="exec")` rejects (`TypeError: expected Module node, got FunctionDef/ClassDef`); wrapping in `ast.Module` is what makes it actually run. Also drops the now-unused `textwrap` import.

## Verification

The unparse→reparse reset is reproducible directly:

```
src: def line 3, raise line 5
OLD unparse:    func lineno=1, raise lineno=3   (renumbered — wrong)
NEW Module-wrap: compile OK, traceback -> "line 5, in f"   (preserved)
```

`squish_text` keeps the leading comment + blank line, so the `def` is on line 3 of the reloaded source; added `test_deduperreload_preserves_source_line_numbers` asserting the reloaded `foo.__code__.co_firstlineno == 3` (it would be `1` with the old unparse path).

(The repo's editable `[test]` build failed in my local venv, so the new test wasn't executed locally — the mechanism is verified via the AST round-trip above and the line-number is pinned by `squish_text`. CI will run it.)

Closes #15359

### DIFF
--- a/IPython/extensions/deduperreload/deduperreload.py
+++ b/IPython/extensions/deduperreload/deduperreload.py
@@ -8,7 +8,6 @@ import os
 import pickle
 import platform
 import sys
-import textwrap
 import tokenize
 from types import ModuleType
 from typing import TYPE_CHECKING, Any, NamedTuple, cast
@@ -446,10 +445,12 @@ class DeduperReloader(DeduperReloaderPatchingMixin):
                 if isinstance(to_patch_to, (staticmethod, classmethod)):
                     to_patch_to = to_patch_to.__func__
                 # exec new source code using old function's (obj) globals environment.
-                func_code = textwrap.dedent(ast.unparse(new_ast_def))
                 if is_method := (len(prefixes) > 0):
-                    func_code = "class __autoreload_class__:\n" + textwrap.indent(
-                        func_code, "    "
+                    # Wrap in a class so the method binds `self`. Operate on the AST
+                    # node directly (rather than unparse + reparse) below to keep the
+                    # original source line numbers in tracebacks. See #15359.
+                    new_ast_def = ast.ClassDef(
+                        name="__autoreload_class__", body=[new_ast_def]
                     )
                 global_env = ns.__dict__
                 if not isinstance(global_env, dict):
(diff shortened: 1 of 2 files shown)


fudianchn:
Closing this — the change introduces a regression. CI fails an existing `test_deduperreload` case with a cascade through the autoreload fallback (`<lambda>() requires a code object with 1 free vars, not 0`), i.e. wrapping the AST node in `ast.Module` + `ast.fix_missing_locations` breaks reload for functions whose body contains a lambda with freevars.

I couldn't reproduce locally (the repo's editable `[test]` build errors in my venv), so I'm closing rather than leaving red CI. The direction (operate on the AST node instead of unparse→reparse to keep line numbers) is still sound, but it needs the freevar/lambda case handled correctly — I'll revisit when I can run the suite locally. Apologies for the noise.

PR #15363, opened 2026-08-09
BUG: Preserve deduperreload source line numbers 🤖🍆

### LINKED ISSUE #15359: deduperreload gives invalid tracebacks in reloaded functions

aryansk:
Fixes #15359

Keep reloaded function AST nodes intact when compiling replacements instead of round-tripping them through `ast.unparse`. Method definitions are wrapped directly in an AST `ClassDef`, and decorated functions use a copied AST without decorators so source line locations remain available in tracebacks. Add regression coverage for the reloaded function's source line number. 🤖🍆

### DIFF
--- a/IPython/extensions/deduperreload/deduperreload.py
+++ b/IPython/extensions/deduperreload/deduperreload.py
@@ -8,7 +8,6 @@ import os
 import pickle
 import platform
 import sys
-import textwrap
 import tokenize
 from types import ModuleType
 from typing import TYPE_CHECKING, Any, NamedTuple, cast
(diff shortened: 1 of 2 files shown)


Carreau:

PR #15364, opened 2026-08-09
quote the filename in the default editor hook

sage-mode-hunter:
The default editor hook runs the editor through the shell without quoting the filename:
- editor() in core/hooks.py builds the command with shell=True and interpolates the filename raw
- %edit only wraps the path in quotes when it has a space, so a filename with ';', '$(...)' or backticks runs before the editor opens
- the path is not always user-typed: %edit on an object resolves it from the object's source file via inspect.getabsfile

install_editor already shlex-quotes the filename inside the hook; this makes the default hook do the same and drops the now-redundant quoting in %edit.

### DIFF
--- a/IPython/core/hooks.py
+++ b/IPython/core/hooks.py
@@ -76,10 +76,16 @@ def editor(self, filename, linenum=None, wait=True):
     if ' ' in editor and os.path.isfile(editor) and editor[0] != '"':
         editor = '"%s"' % editor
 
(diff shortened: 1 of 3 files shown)


sage-mode-hunter:
pushed a fix for the mypy win32 failure (cmd was reassigned from str to list on the windows branch)

sage-mode-hunter:
any update?

PR #15366, opened 2026-08-12
Use dependency groups (PEP 735)

mwtoews:
This PR moves some dependency lists from `project.optional-dependencies` (aka "extras") to `dependency-groups`, as described by [PEP 735](https://peps.python.org/pep-0735/). Some details:

- Keep `black`, `matplotlib` and `all` as "extras" intended for end-users
- Move `docs` (renamed from `doc`), `test`, and `test_extra` to dependency groups, intended for developers
  - Note that `pip install --dryrun ipython[test]` will show a warning that the package does not have an extra named test.
  - Developers that need this dependency will require pip [v25.1 (2025-04-26)](https://pip.pypa.io/en/stable/news/#v25-1) (or `uv`) and use the `--group test` option
- Add a `build` group with tools needed to build this package
- Add a `dev` group, which adds pre-commit and the build and test groups; `uv` enables this group by default.
- Merge docs/requirements.txt into the `docs` group, and remove this file, as it is unnecessary duplication
- Rewrite a few "uv pip install" commands into one line, as it often works best to solve a larger group of dependencies at once


.github/workflows/downstream.yml
```diff
@@ -52,11 +52,11 @@ jobs:
         cd ..
         git clone https://github.com/ipython/ipykernel
         cd ipykernel
-        pip install -e .[test]
+        pip install -e . --group test
```
mwtoews, line 55: See https://github.com/ipython/ipykernel/pull/1547

PR #15368, opened 2026-08-13
Fix local image URL when not embedding

### LINKED ISSUE #13971: Bug in `Image` when using local file with`embed=False`
The following code should show an Image when using the `Image` object when the embed parameter is set to false.
However, it does not do that.
```python
from IPython.display import Image
Image("assets/cute_dog.jpg", embed=False, width=300)
```
I could fix the problem by replacing
`src=\"None\"` with  `src=\"assets/cute_dog.jpg\"`

in my local notebook file:

image

image

I think that this line needs to be modified:

https://github.com/ipython/ipython/blob/3339fad91de3e3da1f02b0b209afef20f021ca33/IPython/core/display.py#L958

fff122:
Fixes #13971

## Summary

When `Image` is constructed with a local `filename` and `embed=False`, the path is stored in `filename` while `url` is unset. Use the local filename as the HTML image source when no URL is present.

Add regression coverage for the non-embedded local-file path.

### DIFF
--- a/IPython/core/display.py
+++ b/IPython/core/display.py
@@ -1067,8 +1067,9 @@ class Image(DisplayObject):
                 klass = ' class="unconfined"'
             if self.alt:
                 alt = ' alt="%s"' % html.escape(self.alt)
+            url = self.url if self.url is not None else self.filename
             return '<img src="{url}"{width}{height}{klass}{alt}/>'.format(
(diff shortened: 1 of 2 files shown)


Carreau:
your commit doe not carry your contibution information, maybe you want to fix that with an --amnend ?

PR #15369, opened 2026-08-16
Preserve source line numbers when deduperreload patches functions

### LINKED ISSUE #15359: deduperreload gives invalid tracebacks in reloaded functions

aryansk:
## What

Fixes #15359 — after `%autoreload` patches a function via deduperreload, tracebacks point at the real source lines instead of the top of the file.

## Why

`DeduperReloader._patch_namespace_inner` rebuilt the replacement function's source text (`ast.unparse` → dedent → re-indent → `ast.parse`), which resets every line number to 1. The patched function's code object then reported `co_firstlineno == 1`, so a traceback pointed at the file's first lines — typically a docstring — rather than the failing statement:

```text
      1 """
      3 """
      4 def func():
```

## How

Instead of round-tripping through text, wrap the **freshly-parsed AST node** directly:

- top-level functions: `ast.Module(body=[new_ast_def])`
- methods: wrap in a synthetic `ast.ClassDef(name="__autoreload_class__", ...)` (so `super()` and zero-arg `__class__` closures keep working), then wrap that in a module

Both keep the same `func_asts[0].body[0]` shape the downstream decorator handling expects, so the decorated-function path is unchanged. `compile()` requires position info on every node, so the synthetic `ClassDef` gets `lineno`/`col_offset` from the wrapped method. The `textwrap` import is no longer needed.

## Tests

Added `test_deduperreloader_preserves_traceback_lines` in `IPython/extensions/tests/test_deduperreload.py`: reloads a module whose function raises, then asserts the captured traceback frame is the real `assert` line (line 6) inside `func` — fails before the fix (points at the docstring line 2), passes after. Full deduperreload suite: 72 passed (was 71). Ruff and darker clean.

🤖 Generated with Codebuff

### DIFF
--- a/IPython/extensions/deduperreload/deduperreload.py
+++ b/IPython/extensions/deduperreload/deduperreload.py
@@ -8,7 +8,6 @@ import os
 import pickle
 import platform
 import sys
-import textwrap
 import tokenize
 from types import ModuleType
 from typing import TYPE_CHECKING, Any, NamedTuple, cast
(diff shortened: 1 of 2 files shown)


Carreau:
While I appreciate you trying to contribute, you send way to many PRs, and comment on too many issues on Jupyter, you will just go down to the bottom of the list and I limited the number of PR new contributor can open. Sometime you do not even check for prior art.

I can ask AI to generate most of those fixes myself; so the problem just asking AI to do it and send a PR.

Carreau:

PR #15371, opened 2026-08-16
Only treat exact -m token as %run module option

### LINKED ISSUE #14849: Some error in %run ,when I use %run "{path}" if path has start with "-m"
the code in execution.py line 688, has a bug. when the path has a string start wit "-m" , the path cannot be exe. for example path "C:/test-main/test.py"

if '-m' in parameter_s and '--' not in parameter_s:
            argv = shlex.split(parameter_s, posix=(os.name == 'posix'))
            for idx, arg in enumerate(argv):
                if arg and arg.startswith('-') and arg != '-':
                    if arg == '-m':
                        argv.insert(idx + 2, '--')
                        break
                else:
                    # Positional arg, break
                    break
            parameter_s = ' '.join(shlex.quote(arg) for arg in argv)

aryansk:
🤖🍆

Fixes #14849

## Summary
- only run the special `%run -m` rewrite when `-m` is an exact parsed token
- leave ordinary script paths containing `-m` unchanged
- add a regression test that fails if an ordinary path is rewritten

## Testing
- Not run locally; added focused regression coverage for the reported path case.

### DIFF
--- a/IPython/core/magics/execution.py
+++ b/IPython/core/magics/execution.py
@@ -690,15 +690,16 @@ class ExecutionMagics(Magics):
         # Add '--' after '-m <module_name>' to ignore additional args passed to a module.
         if '-m' in parameter_s and '--' not in parameter_s:
             argv = shlex.split(parameter_s, posix=(os.name == 'posix'))
-            for idx, arg in enumerate(argv):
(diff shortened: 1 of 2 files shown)


Carreau:

PR #15373, opened 2026-08-16
fix: allow pastebin options after arguments

### LINKED ISSUE #13059: `%pastebin` arguments get silently dropped if they are out of order
```
%pastebin 1 -d "This is a test"
```

Intuitively, the above should result in either:
 - An error, since `-d "This is a test"` should come first.
 - A dpaste.org entry where the description is set to "This is a test".

Instead, it silently drops the arguments.

- [ ] Find for which builtin magics do this sort of thing (`Magics.parse_options` is used for the arguments in `%pastebin` - look for the ones that use it).
- [ ] Fix them - preferably by enabling unordered arguments, whenever possible.

Found out during the review of #13056.

aryansk:
Fixes #13059

Adds opt-in interspersed option parsing and enables it for %pastebin, so options such as -d work before or after the history/file argument without changing other magics' parsing behavior.

Tests: focused pytest regression (2 passed); changed Python modules compile cleanly.

### DIFF
--- a/IPython/core/magic.py
+++ b/IPython/core/magic.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 import os
 import re
 import sys
-from getopt import getopt, GetoptError
+from getopt import GetoptError, getopt, gnu_getopt
 
 from traitlets.config.configurable import Configurable
 from .error import UsageError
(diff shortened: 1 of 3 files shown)


Carreau:

PR #15381, opened 2026-08-31
Have builin-lazy magics have a lower priority than registered on.

### LINKED ISSUE #15380: IPython 9.17.0 breaks `%edit` magic from IPykernel
Due to the new, lazy way magics are loaded here, it's no longer possible for other projects to override core IPython magics. That's because core magics are loaded on demand whereas other magics are loaded at init time of their respective shells.

I'm trying to come up with a PR for this problem, but I wanted to bring it to your attention so a fix could be included in a possible 9.17.1.

Carreau:
Otherwise we break people overriding magics

closes #15380


ccordoba12 (approved):
Tested locally and works as expected. Thanks @Carreau for your help with this!

